PR #6185, opened 2018-05-15
[4.0] RavenDB-8113

aviv86:


src/Raven.Server/Smuggler/Documents/StreamSource.cs
```diff
@@ -719,6 +738,20 @@ private IEnumerable<DocumentItem> ReadDocuments(INewDocumentActions actions = nu
                         continue;
                     }
 
+                    if (metadata != null &&
```
ayende: Why are you removing this?

src/Raven.Server/Smuggler/Documents/StreamSource.cs
```diff
@@ -100,7 +100,18 @@ public DatabaseItemType GetNextType()
                 SkipArray();
                 type = ReadType();
                 if (type == null)
-                    break;
+                    return DatabaseItemType.None;
+            }
+
+            if (type.Equals("Counters", StringComparison.OrdinalIgnoreCase))
```
ayende: What happens when we have TS? 
We need to make this so this will be _ignoring_ any unknown properties, not ignoring well known ones.

src/Raven.Server/Smuggler/Documents/StreamSource.cs
```diff
@@ -803,6 +840,17 @@ private IEnumerable<DocumentTombstone> ReadTombstones(INewDocumentActions action
                         data.TryGet(nameof(DocumentTombstone.Collection), out tombstone.Collection) &&
                         data.TryGet(nameof(DocumentTombstone.LastModified), out tombstone.LastModified))
                     {
+                        if (type == "Counter")
```
ayende: Need to ignore unknown types

ayende (changes requested):

PR #6207, opened 2018-05-16
[4.1] RavenDB-11116 - Error getting license support info

grisha-kotler:
http://issues.hibernatingrhinos.com/issue/RavenDB-11116

### DIFF
--- a/src/Raven.Server/Commercial/LicenseManager.cs
+++ b/src/Raven.Server/Commercial/LicenseManager.cs
@@ -387,31 +387,19 @@ namespace Raven.Server.Commercial
             }
         }
 
-        public async Task Activate(License license, bool skipLeaseLicense, bool ensureNotPassive = true)
+        public async Task Activate(License license, bool skipLeaseLicense, bool ensureNotPassive = true, bool forceActivate = false)
         {
             var newLicenseStatus = GetLicenseStatus(license);
-
-            var licenseExpiration = newLicenseStatus.Expiration;
-            if (licenseExpiration.HasValue == false)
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Commercial/LicenseManager.cs
```diff
@@ -129,7 +130,7 @@ public void Initialize(StorageEnvironment environment, TransactionContextPool co
             {
                 _leaseLicenseTimer = new Timer(state =>
                     AsyncHelpers.RunSync(ExecuteTasks), null,
-                    (int)TimeSpan.FromMinutes(1).TotalMilliseconds,
+                    (int)TimeSpan.FromMinutes(0).TotalMilliseconds,
```
ayende: Is this intentional, and if so , why?

src/Raven.Server/Commercial/LicenseManager.cs
```diff
@@ -129,7 +130,7 @@ public void Initialize(StorageEnvironment environment, TransactionContextPool co
             {
                 _leaseLicenseTimer = new Timer(state =>
                     AsyncHelpers.RunSync(ExecuteTasks), null,
```
ayende: Who handles errors here?

src/Raven.Server/Commercial/LicenseManager.cs
grisha-kotler: no, i reverted this change

src/Raven.Server/Commercial/LicenseManager.cs
grisha-kotler: it is handled in ExecuteTasks

PR #6220, opened 2018-05-17
RavenDB-11109 CSV import to skip badly formatted lines

talweiss1982:


ayende:
We need to report errors properly.

PR #6221, opened 2018-05-17
Raven db 11109

talweiss1982:


src/Raven.Server/Smuggler/Documents/CsvStreamSource.cs
```diff
@@ -153,13 +155,26 @@ public DatabaseRecord GetDatabaseRecord()
...
                 var context = actions.GetContextForNewDocument();
-                yield return ConvertRecordToDocumentItem(context, _csvReader.Context.Record, _csvReaderFieldHeaders, _collection);
+                DocumentItem item = null;
+                try
+                {
+                    item = ConvertRecordToDocumentItem(context, _csvReader.Context.Record, _csvReaderFieldHeaders, _collection);
+                }
+                catch (Exception e)
+                {
+                    _result.AddError($"Fail to parse CSV line {line}, Error:{e.Message}");
```
ayende: Include the full error message?

src/Raven.Server/Smuggler/Documents/CsvStreamSource.cs
```diff
@@ -153,13 +155,26 @@ public DatabaseRecord GetDatabaseRecord()
...
                     continue;
 
                 var context = actions.GetContextForNewDocument();
-                yield return ConvertRecordToDocumentItem(context, _csvReader.Context.Record, _csvReaderFieldHeaders, _collection);
+                DocumentItem item = null;
+                try
+                {
+                    item = ConvertRecordToDocumentItem(context, _csvReader.Context.Record, _csvReaderFieldHeaders, _collection);
+                }
+                catch (Exception e)
```
ayende, line 172: This will throw index out of range error, which likely cause the user to think the fault is with us and not the source data.
We need to handle this explicitly.

src/Raven.Server/Smuggler/Documents/CsvStreamSource.cs
```diff
@@ -153,13 +155,26 @@ public DatabaseRecord GetDatabaseRecord()
...
+                {
+                    _result.AddError($"Fail to parse CSV line {line}, Error:{e.Message}");
+                    item = new DocumentItem();
```
ayende: Why are you returning an empty document instead of calling `continue` ?

ayende (changes requested):

src/Raven.Server/Smuggler/Documents/CsvStreamSource.cs
talweiss1982: https://github.com/ravendb/ravendb/blob/v4.0/src/Raven.Server/Smuggler/Documents/DatabaseSmuggler.cs#L495

src/Raven.Server/Smuggler/Documents/CsvStreamSource.cs
talweiss1982: "Fail to parse CSV line {line}, Error:{e.Message}"
Can't make users smart, it's behind my capabilities

src/Raven.Server/Smuggler/Documents/CsvStreamSource.cs
talweiss1982: The call stack won't give the user any additional info and will make huge messages in the studio, give a line number the user can check his input.

src/Raven.Server/Smuggler/Documents/CsvStreamSource.cs
redknightlois: "The CSV provided to import does not conform to standard rules. Check line {line}. Parse Error: {e.Message}"
That would do it.

PR #6222, opened 2018-05-17
Raven db 11109 v4.1

talweiss1982:


ml054:
what about bumping skip count?

talweiss1982:
@ml054 returning null document bumps the error count which is what i believe we want.

ml054:
that's great! I just wanted to verify that.

ayende:
@ml054 We _can_ cancel import, by throwing an error out.
And we need to decide when to actually abort

ml054:
Since we expose import stats, then maybe user should decide to stop import in studio, once he spot too many unexpected errors? auto abort on 20% might not be good for everyone.

ayende:
Okay, good point, given that the user is likely watching this, they can abort as they want

PR #6225, opened 2018-05-17
RavenDB-11041 properly reporting Mapping Function in static index per…

talweiss1982:
…formance stats


src/Raven.Studio/typescript/viewmodels/database/indexes/indexPerformance.ts
```diff
@@ -169,6 +169,7 @@ class indexPerformance extends viewModelBase {
             "Map": "#0b4971",
             "Storage/DocumentRead": "#0077b5",
             "Linq": "#008cc9",
+            "Jint": "#008cc9",
```
ayende, line 172: @ml054 / @kwiato Can you look at that and make sure that this make sense from the UI?

ml054: sure, I will get this issue and check this.

PR #6228, opened 2018-05-17
[4.1] RavenDB-10692 test secured connection

iftahbe:


src/Raven.Server/RavenServer.cs
```diff
@@ -1298,6 +1298,11 @@ internal void SetCertificate(X509Certificate2 certificate, byte[] rawBytes, stri
                 return true;
             }
 
+            if (tcp.Operation == TcpConnectionHeaderMessage.OperationTypes.TestConnection)
+            {
+                return true;
```
ayende, line 1303: Huh? This is for a specific database, but you are returning true for server wide?

PR #6240, opened 2018-05-21
[4.0] RavenDB-11083 return success in ForceRenew

iftahbe:


src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
```diff
@@ -785,14 +786,22 @@ public Task ForceRenew()
 
             try
             {
-                Server.RefreshClusterCertificate(true);
+                var success = Server.RefreshClusterCertificate(true);
+                using (ServerStore.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
+                using (var writer = new BlittableJsonTextWriter(context, ResponseBodyStream()))
+                {
+                    writer.WriteStartObject();
```
ppekrol, line 793: Wouldn't it be better to use HTTP status codes for this and throw when you cannot refresh the certificate? The exception would contain the exact reason?

ayende: But this isn't an exceptional situation, it is (sort of) expected.

PR #6251, opened 2018-05-22
[4.0] RavenDB-9643 Ensure an unique name of a temp export file for the offl…

arekpalinski:
…ine migration

### DIFF
--- a/src/Raven.Client/ServerWide/Operations/Migration/OfflineMigrationConfiguration.cs
+++ b/src/Raven.Client/ServerWide/Operations/Migration/OfflineMigrationConfiguration.cs
@@ -2,6 +2,7 @@
 using System.IO;
 using System.Linq;
 using System.Text;
+using Raven.Client.Util;
 
 namespace Raven.Client.ServerWide.Operations.Migration
 {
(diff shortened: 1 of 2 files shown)


src/Raven.Client/ServerWide/Operations/Migration/OfflineMigrationConfiguration.cs
```diff
@@ -57,7 +57,7 @@ internal void Validate()
             if (OutputFilePath == null)
             {
                 var rempDir = Path.GetTempPath();
-                OutputFilePath = tmpFile = Path.Combine(rempDir, "export.ravendump");
+                OutputFilePath = tmpFile = Path.Combine(rempDir, $"export-{Guid.NewGuid()}.ravendump");
```
ayende: Use db name and date as well here, I think

arekpalinski: I've dropped the usage of Guid in favor of `$"export-{DatabaseRecord.DatabaseName}-{SystemTime.UtcNow:HH_mm_ss.ffff_MM-dd-yyyy}.ravendump"`

ml054: Isn't better to have yyyy-dd-MM at the beginning? (allows to sort by date)

arekpalinski: Yeah, I will fix it and make it more consistent with RavenDB-11126

PR #6253, opened 2018-05-22
RavenDB-11135 Cluster transactions should be supported in backup, sna…

karmeli87:
…pshot and smuggler

### DIFF
--- a/src/Raven.Client/Documents/Session/ClusterTransactionSession.cs
+++ b/src/Raven.Client/Documents/Session/ClusterTransactionSession.cs
@@ -85,6 +85,8 @@ namespace Raven.Client.Documents.Session
             _storeCompareExchange = null;
         }
 
+        public bool HasCommands => _deleteCompareExchange != null || _storeCompareExchange != null;
+
         protected Task<CompareExchangeValue<T>> GetCompareExchangeValueAsyncInternal<T>(string key)
         {
             return _session.Operations.SendAsync(new GetCompareExchangeValueOperation<T>(key));
--- a/src/Raven.Client/Documents/Session/ClusterTransactionSession.cs
+++ b/src/Raven.Client/Documents/Session/ClusterTransactionSession.cs
@@ -85,7 +85,7 @@ namespace Raven.Client.Documents.Session
             _storeCompareExchange = null;
         }
 
-        public bool HasCommands => _deleteCompareExchange != null || _storeCompareExchange != null;
+        internal bool HasCommands => _deleteCompareExchange != null || _storeCompareExchange != null;
 
         protected Task<CompareExchangeValue<T>> GetCompareExchangeValueAsyncInternal<T>(string key)
         {
--- a/src/Raven.Client/Documents/Smuggler/DatabaseItemType.cs
+++ b/src/Raven.Client/Documents/Smuggler/DatabaseItemType.cs
@@ -18,6 +18,7 @@ namespace Raven.Client.Documents.Smuggler
         LegacyDocumentDeletions = 1 << 9,
         LegacyAttachmentDeletions = 1 << 10,
         DatabaseRecord = 1 << 11,
-        Counters = 1 << 12
+        Counters = 1 << 12,
+        PendingClusterTransactions = 1 << 13
     }
 }
--- a/src/Raven.Client/Documents/Smuggler/DatabaseSmuggler.cs
+++ b/src/Raven.Client/Documents/Smuggler/DatabaseSmuggler.cs
@@ -125,9 +125,12 @@ namespace Raven.Client.Documents.Smuggler
             // as the previous files can hold indexes and identities which were deleted and shouldn't be imported
             var oldOperateOnTypes = options.OperateOnTypes;
             options.OperateOnTypes = options.OperateOnTypes &
-                                     ~(DatabaseItemType.Indexes | DatabaseItemType.CompareExchange | DatabaseItemType.Identities);
+                                     ~(DatabaseItemType.Indexes | ClusterStateMachineImportOptions);
             return oldOperateOnTypes;
         }
+
+        public static DatabaseItemType ClusterStateMachineImportOptions =>
+            DatabaseItemType.CompareExchange | DatabaseItemType.Identities | DatabaseItemType.PendingClusterTransactions;
 
         public async Task<Operation> ImportAsync(DatabaseSmugglerImportOptions options, string fromFile, CancellationToken cancellationToken = default)
         {
--- a/src/Raven.Client/Documents/Smuggler/DatabaseSmuggler.cs
+++ b/src/Raven.Client/Documents/Smuggler/DatabaseSmuggler.cs
@@ -129,7 +129,7 @@ namespace Raven.Client.Documents.Smuggler
             return oldOperateOnTypes;
         }
 
-        public static DatabaseItemType ClusterStateMachineImportOptions =>
+        internal static DatabaseItemType ClusterStateMachineImportOptions =>
             DatabaseItemType.CompareExchange | DatabaseItemType.Identities | DatabaseItemType.PendingClusterTransactions;
 
         public async Task<Operation> ImportAsync(DatabaseSmugglerImportOptions options, string fromFile, CancellationToken cancellationToken = default)
--- a/src/Raven.Server/Documents/DatabasesLandlord.cs
+++ b/src/Raven.Server/Documents/DatabasesLandlord.cs
@@ -183,7 +183,7 @@ namespace Raven.Server.Documents
                     var currentRaftIndex = ChangeVectorUtils.GetEtagById(global, dbGrpId);
                     var firstCommandIndex = ClusterTransactionCommand.ReadFirstIndex(serverContext, database.Name);
                     var first = Math.Max(firstCommandIndex, currentRaftIndex + 1);
-                    foreach (var command in ClusterTransactionCommand.ReadCommandsBatch(serverContext, database.Name, first))
+                    foreach (var command in ClusterTransactionCommand.ReadCommandsBatch(serverContext, docContext, database.Name, first))
                     {
                         var mergedCommands = new BatchHandler.ClusterTransactionMergedCommand(database, command);
                         database.TxMerger.Enqueue(mergedCommands).ContinueWith(t =>
@@ -206,6 +206,13 @@ namespace Raven.Server.Documents
                                         Array = mergedCommands.Reply,
                                         IndexTask = indexTask,
                                     };
+
+                                    if (mergedCommands.Options?.TaskId == null)
+                                    {
+                                        database.RachisLogIndexNotifications.NotifyListenersAbout(index, null);
+                                        return;
+                                    }
+
                                     database.ClusterTransactionWaiter.SetResult(mergedCommands.Options.TaskId, index, result);
                                     return;
                                 }
@@ -505,6 +512,15 @@ namespace Raven.Server.Documents
             try
             {
                 var task = new Task<DocumentDatabase>(() => ActuallyCreateDatabase(databaseName, config), TaskCreationOptions.RunContinuationsAsynchronously);
+                task.ContinueWith(t =>
+                {
+                    if (t.IsCompletedSuccessfully)
+                    {
+                        // we don't care about the raft index so we pass 0. 
+                        NotifyPendingClusterTransaction(databaseName, t, 0);
+                    }
+                });
+
                 var database = DatabasesCache.GetOrAdd(databaseName, task);
                 if (database == task)
                     task.Start(); // the semaphore will be released here at the end of the task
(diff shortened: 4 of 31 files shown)


src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -206,6 +206,13 @@ private void NotifyPendingClusterTransaction(string databaseName, Task<DocumentD
                                         Array = mergedCommands.Reply,
                                         IndexTask = indexTask,
                                     };
+
+                                    if (mergedCommands.Options?.TaskId == null)
```
ayende: How can this be null? Because if so, in line 216 we have NRE

src/Raven.Server/Documents/DocumentPutAction.cs
```diff
@@ -115,6 +115,12 @@ public DocumentPutAction(DocumentsStorage documentsStorage, DocumentDatabase doc
 
                 var result = BuildChangeVectorAndResolveConflicts(context, id, lowerId, newEtag, document, changeVector, expectedChangeVector, flags, oldValue);
 
+                if (flags.Contain(DocumentFlags.FromClusterTransaction) && nonPersistentFlags.Contain(NonPersistentDocumentFlags.FromSmuggler))
+                {
+                    var index = ChangeVectorUtils.GetEtagById(context.LastDatabaseChangeVector, _documentDatabase.DatabaseGroupId);
+                    changeVector = $"RAFT:{index}-{_documentDatabase.DatabaseGroupId}";
```
ayende: This is bad, because it doesn't take into account the cluster level index.
A cluster tx on this may very well have a raft index that is smaller than the etag, resulting in us expecting that this is already committed and ignoring the change.

src/Raven.Client/Documents/Session/ClusterTransactionSession.cs
```diff
@@ -85,6 +85,8 @@ public void Clear()
             _storeCompareExchange = null;
         }
 
+        public bool HasCommands => _deleteCompareExchange != null || _storeCompareExchange != null;
```
ppekrol: internal?

src/Raven.Client/Documents/Smuggler/DatabaseSmuggler.cs
```diff
@@ -125,10 +125,13 @@ public static DatabaseItemType ConfigureOptionsForIncrementalImport(DatabaseSmug
             // as the previous files can hold indexes and identities which were deleted and shouldn't be imported
             var oldOperateOnTypes = options.OperateOnTypes;
             options.OperateOnTypes = options.OperateOnTypes &
-                                     ~(DatabaseItemType.Indexes | DatabaseItemType.CompareExchange | DatabaseItemType.Identities);
+                                     ~(DatabaseItemType.Indexes | ClusterStateMachineImportOptions);
             return oldOperateOnTypes;
         }
 
+        public static DatabaseItemType ClusterStateMachineImportOptions =>
```
ppekrol: private or internal

src/Raven.Server/ServerWide/Commands/CompareExchangeCommands.cs
```diff
@@ -22,7 +22,7 @@ public abstract class CompareExchangeCommandBase : CommandBase
 
         public static string GetActualKey(string database, string key)
         {
-            return ("db/" + database + "/" + key).ToLowerInvariant();
+            return (database + "/" + key).ToLowerInvariant();
```
ppekrol, line 25: isn't this breaking backward-compatibility?

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
```diff
@@ -254,11 +259,62 @@ private void FinishBatchOfDocuments()
             }
         }
 
+        private class DatabasePendingClusterTransacinsActions : IValueActions<ClusterTransactionCommand.SingleClusterDatabaseCommand>
```
ppekrol: typo :)

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
```diff
@@ -254,11 +259,62 @@ private void FinishBatchOfDocuments()
...
+                if (_commands.Count == 0)
+                    return;
+
+                SendTransactionCommands();
+            }
+
+            private void SendTransactionCommands()
+            {
+                var batch = new ClusterTransactionBatchCommand(_commands);
+                AsyncHelpers.RunSync(async () => await _database.ServerStore.SendToLeaderAsync(batch));
```
ppekrol: no need to add async/await here?

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
```diff
@@ -268,25 +324,27 @@ public DatabaseCompareExchangeActions(DocumentDatabase database, JsonOperationCo
...
 
-                SendCommands(_context);
+                SendCommands();
             }
 
-            private void SendCommands(JsonOperationContext context)
+            private void SendCommands()
             {
-                AsyncHelpers.RunSync(async () => await _database.ServerStore.SendToLeaderAsync(new AddOrUpdateCompareExchangeBatchCommand(_compareExchangeCommands, context)));
+                AsyncHelpers.RunSync(async () => await _database.ServerStore.SendToLeaderAsync(new AddOrUpdateCompareExchangeBatchCommand(_compareExchangeCommands, _context)));
```
ppekrol: can remove async/await from here?

src/Sparrow/Json/BlittableJsonTextWriter.cs
```diff
@@ -137,6 +138,11 @@ public override string ToString()
             return Encodings.Utf8.GetString(_pinnedBuffer.Buffer.Array, _pinnedBuffer.Buffer.Offset, _pos);
         }
 
+        public void WriteIDynamicJsonWithContext(IDynamicJsonWithContext obj)
```
ppekrol: WriteObject?

src/Raven.Server/Smuggler/Documents/StreamSource.cs
```diff
@@ -225,6 +228,43 @@ public IDisposable GetCounterValues(out IEnumerable<CounterDetail> counters)
...
+        }
+
+        private IEnumerable<ClusterTransactionCommand.SingleClusterDatabaseCommand> InternalGetPendingClusterTransactions(DisposableScope scope)
+        {
+            foreach (BlittableJsonReaderObject reader in ReadArray(disposableScope: scope))
+            {
+                scope.EnsureDispose(reader);
+                if (reader.TryGet(nameof(ClusterTransactionCommand.SingleClusterDatabaseCommand.Commands), out BlittableJsonReaderArray commandsArray) == false)
+                {
+                    _result.CompareExchange.ErroredCount++;
```
ppekrol: CompareExchange?

src/Raven.Client/Documents/Smuggler/DatabaseItemType.cs
```diff
@@ -18,6 +18,7 @@ public enum DatabaseItemType
         LegacyDocumentDeletions = 1 << 9,
         LegacyAttachmentDeletions = 1 << 10,
         DatabaseRecord = 1 << 11,
-        Counters = 1 << 12
+        Counters = 1 << 12,
+        PendingClusterTransactions = 1 << 13
```
ppekrol, line 22: just 'ClusterTransactions' ?

ppekrol:
@karmeli87 please check the http://jenkins.hibernatingrhinos.com/job/PR_Conventions/1535/console

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1140,7 +1150,8 @@ public IEnumerable<(string Key, long Index)> GetCompareExchangeIndexes(Transacti
         private static unsafe string ReadCompareExchangeKey(TableValueReader reader, string dbPrefix)
         {
             var ptr = reader.Read((int)UniqueItems.Key, out var size);
-            return Encodings.Utf8.GetString(ptr, size).Substring(dbPrefix.Length + 4);
+            // we need to read only the key from the format: 'databaseName/key'
+            return Encodings.Utf8.GetString(ptr, size).Substring(dbPrefix.Length + 1);
```
ayende, line 1154: why this change now?

src/Raven.Client/Documents/Session/ClusterTransactionSession.cs
karmeli87: done

src/Raven.Server/ServerWide/Commands/CompareExchangeCommands.cs
ayende, line 25: This is to match 4.0 behavior, right?

src/Raven.Client/Documents/Smuggler/DatabaseSmuggler.cs
karmeli87: done

src/Raven.Server/Smuggler/Documents/DatabaseSource.cs
```diff
@@ -248,6 +252,18 @@ public IDisposable GetCompareExchangeValues(out IEnumerable<(string key, long in
             }
         }
 
+        public IDisposable GetPendingClusterTransactions(out IEnumerable<ClusterTransactionCommand.SingleClusterDatabaseCommand> transaction)
+        {
+            using (var scope = new DisposableScope())
+            {
+                scope.EnsureDispose(_database.ServerStore.ContextPool.AllocateOperationContext(out TransactionOperationContext context));
+                scope.EnsureDispose(context.OpenReadTransaction());
+                // if we export the transaction without having the document, we will not check if the transaction was already applied.
```
ayende, line 261: What is the meaning here?

ayende (changes requested):

src/Raven.Server/ServerWide/Commands/CompareExchangeCommands.cs
karmeli87: The previous implementation, broke it. This should fix it :)

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
karmeli87: done

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
karmeli87: done

src/Raven.Server/Documents/DatabasesLandlord.cs
karmeli87: done

src/Raven.Server/ServerWide/ClusterStateMachine.cs
karmeli87: Because we changed the key format. (removed the 'db/' prefix)

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
karmeli87: done

src/Raven.Server/Smuggler/Documents/DatabaseSource.cs
karmeli87: When we export transaction _with_ documents we will check whether the transaction was already executed and if so, we will skip exporting it.
If we export _without_ the documents we will always take the stored transaction. (Not sure if there is a real case when we want to do something like this, but it is necessary for the tests)

src/Raven.Server/Smuggler/Documents/StreamSource.cs
karmeli87: done

src/Sparrow/Json/BlittableJsonTextWriter.cs
karmeli87: done

src/Raven.Server/Documents/DocumentPutAction.cs
karmeli87: This is indeed bad, because two document will have the same change vector.

the RAFT here is confusing, it doesn't reflect the raft index of the cluster, rather it is an incremental etag for each PUT/DELETE command that was passed through the cluster transaction.
With the unique Id of the database group.

src/Raven.Client/Documents/Smuggler/DatabaseItemType.cs
ayende: That is actually pending ones, that haven't been applied yet

src/Raven.Server/Documents/DocumentPutAction.cs
```diff
@@ -115,6 +115,17 @@ public DocumentPutAction(DocumentsStorage documentsStorage, DocumentDatabase doc
 
                 var result = BuildChangeVectorAndResolveConflicts(context, id, lowerId, newEtag, document, changeVector, expectedChangeVector, flags, oldValue);
 
+                if (flags.Contain(DocumentFlags.FromClusterTransaction) && nonPersistentFlags.Contain(NonPersistentDocumentFlags.FromSmuggler))
```
ayende, line 118: Strip `FromClusterTransaction` during import

ayende (changes requested):

karmeli87:
Instead of this complex commit, we simply should wait on the export for all pending transactions to complete.

PR #6260, opened 2018-05-22
RavenDB-9629 Update images for sample data

kwiato:


ayende:
Revisions are gone in there.

gregolsky:
Right, we need to enable those before making sample data.

ayende:
Note that the revisions we have there are already for _building_ the orders.
So sample data have 30 revisions for `orders/830-A`

PR #6292, opened 2018-05-23
[4.1] RavenDB-10925 added a surpassed auto index cleanup task

ppekrol:


src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -1005,6 +1003,61 @@ public IEnumerable<Index> GetIndexes()
         public void RunIdleOperations()
         {
             AsyncHelpers.RunSync(HandleUnusedAutoIndexes);
+            AsyncHelpers.RunSync(DeleteSurpassedAutoIndexes);
+        }
+
+        private async Task DeleteSurpassedAutoIndexes()
+        {
+            if (_serverStore.IsLeader() == false)
```
ayende: This is wrong. 
What you are doing here is asking for the _cluster_, not the db group.
If we have a db that is not on the leader, it will never run this.
Better to say something like: "dbGroup[0] == this"

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -1005,6 +1003,61 @@ public IEnumerable<Index> GetIndexes()
...
+                    }
+                }
+            }
+
+            if (indexesToRemove.Count == 0)
+                return;
+
+            foreach (var indexName in indexesToRemove)
+            {
+                await TryDeleteIndexIfExists(indexName);
```
ayende: Error handling?

ayende (changes requested):

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -1002,15 +1003,25 @@ public IEnumerable<Index> GetIndexes()
...
         public void RunIdleOperations()
         {
-            AsyncHelpers.RunSync(HandleUnusedAutoIndexes);
-            AsyncHelpers.RunSync(DeleteSurpassedAutoIndexes);
+            DatabaseRecord record;
+            using (_serverStore.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
+            using (context.OpenReadTransaction())
+                record = _serverStore.Cluster.ReadDatabase(context, _documentDatabase.Name);
+
+            if (record.Topology.Members[0] != _serverStore.NodeTag)
```
ppekrol: @ayende Is there a possibility to have `null` here or 0 members? I need to apply similar fix to v4.0

ayende: No, it can't be null / empty

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -1054,17 +1065,22 @@ public void RunIdleOperations()
...
+                }
+                catch (Exception e)
+                {
+                    if (_logger.IsOperationsEnabled)
+                        _logger.Operations($"Could not delete surpassed index '{indexName}'.", e);
+                }
             }
         }
 
         private async Task HandleUnusedAutoIndexes()
```
ayende: I'm not sure how correct this is going to be.  What happens if queries are made on a _different_ node, and not this one?

ayende: Might actually be better to move this entirely to the cluster observer?

ppekrol: Created a separate issue for that: http://issues.hibernatingrhinos.com/issue/RavenDB-11206

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -1004,14 +1003,84 @@ public IEnumerable<Index> GetIndexes()
...
         }
 
-        private async Task HandleUnusedAutoIndexes()
+        private async Task RunIdleOperationsAsync()
+        {
+            await HandleUnusedAutoIndexes();
+            await DeleteSurpassedAutoIndexes();
+        }
+
+        private async Task DeleteSurpassedAutoIndexes()
```
ayende: Let's run this IFF there have been changes in the indexes since we last run this.

ayende: Can just check the index of the db record to see if it changed

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -1004,14 +1003,84 @@ public IEnumerable<Index> GetIndexes()
...
+                record = _serverStore.Cluster.ReadDatabase(context, _documentDatabase.Name);
+
+            if (record.Topology.Members[0] != _serverStore.NodeTag)
+                return;
+
+            AsyncHelpers.RunSync(RunIdleOperationsAsync);
         }
 
-        private async Task HandleUnusedAutoIndexes()
+        private async Task RunIdleOperationsAsync()
```
ayende: Technically speaking, this may actually run on two different machines at the same time, even with the check.
I wonder if we should handle this in some manner? Maybe do the operation optimistically, so send it to the leader and it will only actually run if the db record didn't change?

ayende (changes requested):

src/Raven.Server/Documents/Indexes/IndexStore.cs
ppekrol: Done

src/Raven.Server/Documents/Indexes/IndexStore.cs
ppekrol: I have some ideas, let's apply them as a part of http://issues.hibernatingrhinos.com/issue/RavenDB-11206

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -1004,14 +1005,120 @@ public IEnumerable<Index> GetIndexes()
...
+            long etag;
+            DatabaseRecord record;
+            using (_serverStore.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
+            using (context.OpenReadTransaction())
+                record = _serverStore.Cluster.ReadDatabase(context, _documentDatabase.Name, out etag);
+
+            if (record.Topology.Members[0] != _serverStore.NodeTag)
+                return;
+
+            AsyncHelpers.RunSync(() => RunIdleOperationsAsync(etag));
```
ayende, line 1017: Can there be errors from here? Don't we need to handle this?

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -1004,14 +1005,120 @@ public IEnumerable<Index> GetIndexes()
...
+                if (string.Equals(definition.Name, kvp.Key, StringComparison.Ordinal))
+                    continue;
+
+                try
+                {
+                    await CreateIndex(definition);
+
+                    indexesToRemove.Add(kvp.Key);
+                }
+                catch (Exception e)
```
ayende, line 1097: shouldn't we abort if there are errors?

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -1004,14 +1005,120 @@ public IEnumerable<Index> GetIndexes()
...
+                        _logger.Operations($"Could not create extended index '{definition.Name}'.", e);
+                }
+            }
+
+            foreach (var indexName in indexesToRemove)
+            {
+                try
+                {
+                    await TryDeleteIndexIfExists(indexName);
+                    if (_logger.IsInfoEnabled)
```
ayende, line 1109: Shouldn't we abort if there are errors?

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -1004,14 +1005,120 @@ public IEnumerable<Index> GetIndexes()
...
+            foreach (var kvp in indexesToExtend)
+            {
+                var definition = kvp.Value.CreateAutoIndexDefinition();
+
+                if (string.Equals(definition.Name, kvp.Key, StringComparison.Ordinal))
+                    continue;
+
+                try
+                {
+                    await CreateIndex(definition);
```
ayende, line 1093: I wonder if it isn't safer to do this one index at a time, instead of all the matches.

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -1004,14 +1005,120 @@ public IEnumerable<Index> GetIndexes()
...
+            await HandleUnusedAutoIndexes();
+            await DeleteOrMergeSurpassedAutoIndexes(databaseRecordEtag);
+        }
+
+        private async Task DeleteOrMergeSurpassedAutoIndexes(long databaseRecordEtag)
+        {
+            if (_lastSurpassedAutoIndexesDatabaseRecordEtag >= databaseRecordEtag)
                 return;
 
+            _lastSurpassedAutoIndexesDatabaseRecordEtag = databaseRecordEtag;
```
ayende, line 1031: If there have been errors, we won't run until the next DatabaseRecord update.
What about moving this to the end?

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -1081,9 +1188,17 @@ orderby lastQueryingTime
                 {
                     if (age <= ageThreshold || lastQuery >= timeToWaitBeforeDeletingAutoIndexMarkedAsIdle.AsTimeSpan)
                     {
-                        await TryDeleteIndexIfExists(item.Index.Name);
-                        if (_logger.IsInfoEnabled)
-                            _logger.Info($"Deleted index '{item.Index.Name}' due to idleness. Age: {age}. Last query: {lastQuery}.");
+                        try
+                        {
+                            await TryDeleteIndexIfExists(item.Index.Name);
```
ayende, line 1193: I think we should do this one index at a time.

ayende (commented):
We still need to ensure that this doesn't run concurrently, which means that we need to add the original database record etag to the requests.

PR #6294, opened 2018-05-23
RavenDB-10561: add some info to dump on case of PeepingTomTest failure

aviviadi:


src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -121,7 +121,7 @@ public unsafe class ManagedPinnedBuffer : IDisposable
             public int Valid, Used;
             public byte* Pointer;
 
-            private GCHandle _handle;
+            public GCHandle _handle; // TODO: return to private after solving RavenDB-10561
```
ppekrol, line 124: Can we use `internal` instead?

aviviadi: Yes, I will change that, but it is temporary code anyhow and will be removed once the issue is solved

PR #6297, opened 2018-05-23
RavenDB-11135 Cluster transactions should be supported in backup, sna…

karmeli87:
…pshot and smuggler

### DIFF
--- a/src/Raven.Server/Documents/DocumentPutAction.cs
+++ b/src/Raven.Server/Documents/DocumentPutAction.cs
@@ -116,6 +116,12 @@ namespace Raven.Server.Documents
 
                 }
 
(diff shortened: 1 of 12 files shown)


src/Raven.Server/Documents/DocumentPutAction.cs
```diff
@@ -116,6 +116,12 @@ public DocumentPutAction(DocumentsStorage documentsStorage, DocumentDatabase doc
 
                 }
 
+                if (flags.Contain(DocumentFlags.FromClusterTransaction) && nonPersistentFlags.Contain(NonPersistentDocumentFlags.FromSmuggler))
```
ayende, line 119: Let's move this to `DatabaseSource`, instead. We are already too busy here.

ayende: Also need this on Tombstone, for example

PR #6317, opened 2018-05-24
[4.0] RavenDB-11053 set passive state when not in cluster

iftahbe:


src/Raven.Server/Documents/Handlers/Debugging/NodeDebugHandler.cs
```diff
@@ -54,7 +54,20 @@ public Task ListRecentEngineLogs()
...
             return Task.CompletedTask;
         }
-        
+
+        [RavenAction("/admin/debug/node/state-change-history", "GET", AuthorizationStatus.Operator, IsDebugInformationEndpoint = true)]
+        public Task GetStateChangeHistory()
+        {
+            using (ServerStore.ContextPool.AllocateOperationContext(out JsonOperationContext context))
+            using (var write = new BlittableJsonTextWriter(context, ResponseBodyStream()))
+            {
```
ayende, line 63: Use object, not direct array to write this.

PR #6326, opened 2018-05-24
[4.0] RavenDB-11181

karmeli87:


src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -239,6 +244,14 @@ public long LastRaftCommitIndex
                 }
                 catch (Exception e)
                 {
+                    if (e is WebSocketException && 
+                        e.InnerException?.InnerException?.Message != null && 
+                        e.InnerException.InnerException.Message.Contains("A certificate is required to complete client authentication"))
```
ayende, line 249: This is a bad idea. It won't work with localized exception messages. 
Better mode here would be to handle this with exponential backoff strategy.
`UpdateTopologyChangeNotification` should wait 500 ms after first failure, then 1 second, then 2 seconds, then 3, etc. 
Up to a check per every 30 seconds.

PR #6329, opened 2018-05-24
[4.1] RavenDB-11181

karmeli87:
…ed notification

### DIFF
--- a/src/Raven.Server/ServerWide/ServerStore.cs
+++ b/src/Raven.Server/ServerWide/ServerStore.cs
@@ -223,15 +223,16 @@ namespace Raven.Server.ServerWide
                                         if (notification == null)
                                             break;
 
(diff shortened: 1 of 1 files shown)


src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -223,15 +223,16 @@ public long LastRaftCommitIndex
                                         if (notification == null)
                                             break;
 
-                                        if (notification.TryGet("Id",out string id) && id != nameof(ClusterTopologyChanged))
+                                        if (notification.TryGet(nameof(ClusterTopologyChanged.Type), out NotificationType notificationType) == false ||
```
ayende, line 226: Here too, we need the backoff strategy when calling `UpdateTopologyChangeNotification`

PR #6330, opened 2018-05-24
RavenDB-11078 spatial JavaScript index fields

talweiss1982:


src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/JintLuceneDocumentConverter.cs
```diff
@@ -69,11 +72,59 @@ protected override int GetFields<T>(T instance, LazyStringValue key, object docu
...
+
+                        SpatialField spatialField = null;
+                        IEnumerable<AbstractField> spatial = null;
+                        switch (type.AsString())
+                        {
+                            case "Coordinates":
+                                if (sf.TryGetValue("Coordinates", out var coordinates) == false || coordinates.IsObject() == false)
+                                    throw new InvalidSpatialArgument($"Property {field.Name} is a spatial field of Coordinates type but has invalid or missing Coordinates field.");
+                                var coordinatesObj = coordinates.AsObject();
+                                if (coordinatesObj.TryGetValue("Lat", out var lat) == false || coordinatesObj.TryGetValue("Lng", out var lng) == false)
```
ayende, line 112: In C#, we just ignore this scenario, I think this might be a better option here.

src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/JintLuceneDocumentConverter.cs
```diff
@@ -69,11 +72,59 @@ protected override int GetFields<T>(T instance, LazyStringValue key, object docu
...
+
+                        var sf = inner.AsObject();
+                        if (sf.TryGetValue("Type", out var type) == false || type.IsString() == false)
+                        {
+                            throw new InvalidSpatialArgument($"Property {field.Name} is a spatial field but it doesn't have a valid type.");
+                        }
+
+                        SpatialField spatialField = null;
+                        IEnumerable<AbstractField> spatial = null;
+                        switch (type.AsString())
```
ayende, line 106: Do we need a type, we can just tell by the precence of fields, no?

src/Raven.Server/Documents/Indexes/Static/JavaScriptIndex.cs
```diff
@@ -223,7 +224,19 @@ private JsValue LoadDocument(JsValue self, JsValue[] args)
         reduce.aggregateBy = reduceFunction;
     }
     return reduce;
-}";
+}
+
+function createSpatialPoint(lat, lng) {
```
ayende, line 229: I don't like this:
`createSpatialField(createSpatialPoint(l.Latitude, l.Longitude))}`

Let's do "overloading" here by checking number of args and their types.
Using `arguments`, we have the following signatures:

* `createSpatialField(name, lat,lng)`
* `createSpatialField(name, wkt)`
* `createSpatialField(lat, lng)`
* `createSpatialField(wkt)`

We can use the types to decide how to deal with this directly

src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/JintLuceneDocumentConverter.cs
```diff
@@ -69,11 +72,59 @@ protected override int GetFields<T>(T instance, LazyStringValue key, object docu
                 object value;
                 if (obj.IsObject() && obj.IsArray() == false)
                 {
+                    //Note that for dynamic fields 'field.Name' will contain the real name of the field and not 'property'
```
ayende, line 75: I don't see where this is actually happening, though?

src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/JintLuceneDocumentConverter.cs
talweiss1982: what do you mean by "this scenario"? bad spatial field needs to be ignored and not error?

src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/JintLuceneDocumentConverter.cs
talweiss1982: I guess we can avoid this extra field, but I'll still need to have the same number of parameters since there is no overloading in JS

src/Raven.Server/Documents/Indexes/Static/JavaScriptIndex.cs
talweiss1982: No overloading in JS, the 'createSpatialPoint' is just an helper function you can invoke it like so: 
createSpatialField({$spatial:{Lat:x.lat, Lng:x.lng}})      // (assuming i'll remove the Type parameters)

src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/JintLuceneDocumentConverter.cs
talweiss1982: This is just a comment to explain the "magic" since i understand that it won't be clear to somebody reading the code. I have actually already implemented the logic more than a month ago when dealing with dynamic fields and there was very little i had to change to support the spatial fields. The value pased to the dynamic field is the createSpatialField(...) and it is later handled in the same code path as regular spatial fields.

src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/JintLuceneDocumentConverter.cs
ayende: Yes

src/Raven.Server/Documents/Indexes/Static/JavaScriptIndex.cs
ayende: No, overloading won't work in JS, but _varaible number of args_ work

src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/JintLuceneDocumentConverter.cs
ayende: I'm not following this from the comment and the code, can you extend the comment?

src/Raven.Server/Documents/Indexes/Static/JavaScriptIndex.cs
talweiss1982: I have looked at the JS community and asked and this was the answer I got, I don't claim to know what makes sense in JS. I'll check the number of arguments, I don't really have an opinion about this...

src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/JintLuceneDocumentConverter.cs
talweiss1982: i'll try to make the comment clearer

src/Raven.Server/Documents/Indexes/Static/JavaScriptIndex.cs
ml054: @talweiss1982 maybe this will be useful for you:
https://www.typescriptlang.org/docs/handbook/functions.html#overloads

Ex.

function pickCard(x: {suit: string; card: number; }[]): number;
function pickCard(x: number): {suit: string; card: number; };
function pickCard(x): any {
    // Check to see if we're working with an object/array
    // if so, they gave us the deck and we'll pick the card
    if (typeof x == "object") {
        let pickedCard = Math.floor(Math.random() * x.length);
        return pickedCard;
    }
    // Otherwise just let them pick the card
    else if (typeof x == "number") {
        let pickedSuit = Math.floor(x / 13);
        return { suit: suits[pickedSuit], card: x % 13 };
    }
}

PR #6332, opened 2018-05-24
RavenDB-11089 related changes

myarichuk:
* Make sure queries, index definitions, MoreLikeThis, Patches and Facets adhere to custom json serializers.
* add convention for property name changes
* relevant tests
(RavenDB-11089)

### DIFF
--- a/src/Raven.Client/Documents/Conventions/DocumentConventions.cs
+++ b/src/Raven.Client/Documents/Conventions/DocumentConventions.cs
@@ -38,6 +38,8 @@ namespace Raven.Client.Documents.Conventions
 
         internal static readonly DocumentConventions Default = new DocumentConventions();
 
+        public Func<MemberInfo, string> PropertyNameConverter = info => info.Name;
+
         private static Dictionary<Type, string> _cachedDefaultTypeCollectionNames = new Dictionary<Type, string>();
 
         private readonly Dictionary<MemberInfo, CustomQueryTranslator> _customQueryTranslators = new Dictionary<MemberInfo, CustomQueryTranslator>();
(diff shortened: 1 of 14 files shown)


src/Raven.Client/Documents/Conventions/DocumentConventions.cs
```diff
@@ -38,6 +38,8 @@ public class DocumentConventions : Client.Conventions
 
         internal static readonly DocumentConventions Default = new DocumentConventions();
 
+        public Func<MemberInfo, string> PropertyNameConverter = info => info.Name;
```
ppekrol, line 41: if public then it needs to be frozen after DocumentConventions.Freeze

src/Raven.Client/Documents/Indexes/ExpressionStringBuilder.cs
```diff
@@ -125,8 +125,9 @@ private void Out(string s)
 
         private void OutMember(Expression instance, MemberInfo member, Type exprType)
         {
-            bool isId = false;
-            var name = GetPropertyName(member.Name, exprType);
+            var isId = false;
+            var name = _conventions.PropertyNameConverter(member) ?? GetPropertyName(member.Name, exprType);
```
ppekrol, line 129: The convention is not null (or does not return null by default, which means that it changes the old behavior and introduces a breaking change.

src/Raven.Client/Extensions/PropertyNameConventionJavaScriptCompilationExtension.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.Diagnostics;
+using System.Linq.Expressions;
+using System.Reflection;
+using Lambda2Js;
+using Raven.Client.Documents.Conventions;
+
+namespace Raven.Client.Extensions
+{
+    public class PropertyNameConventionJSMetadataProvider : JavascriptMetadataProvider
```
ppekrol, line 9: please make it internal

src/Raven.Client/Documents/Indexes/ExpressionStringBuilder.cs
ayende: Good point, let's make it null / noop by default

PR #6336, opened 2018-05-25
RavenDB-11220 Fixing "Lucene.Net.Store.AlreadyClosedException: this D…

arekpalinski:
…irectory is closed" during index recovery on db startup

### DIFF
--- a/Raven.Database/Indexing/IndexStorage.cs
+++ b/Raven.Database/Indexing/IndexStorage.cs
@@ -253,7 +253,22 @@ namespace Raven.Database.Indexing
                         throw new InvalidOperationException("Could not open / create index" + indexName + ", reset already tried", e);
 
                     if (indexImplementation != null)
+                    {
                         indexImplementation.Dispose();
+
+                        // index disposal just closed the lucene directory, let's try to reopen it
+                        try
+                        {
+                            luceneDirectory = OpenOrCreateLuceneDirectory(indexDefinition, createIfMissing: resetTried, forceFullExistingIndexCheck: forceFullIndexCheck);
+                        }
+                        catch (Exception ex)
+                        {
+                            if (log.IsWarnEnabled)
+                                log.Warn("Could not reopen index lucene directory after disposing index instance. Will reset the index", ex);
+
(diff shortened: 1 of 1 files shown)


Raven.Database/Indexing/IndexStorage.cs
```diff
@@ -253,8 +253,23 @@ private Index OpenIndex(string indexName, bool onStartup, bool forceFullIndexChe
...
                     if (indexImplementation != null)
+                    {
                         indexImplementation.Dispose();
 
+                        // index disposal just closed the lucene directory, let's try to reopen it
+                        try
+                        {
+                            luceneDirectory = OpenOrCreateLuceneDirectory(indexDefinition, createIfMissing: resetTried, forceFullExistingIndexCheck: forceFullIndexCheck);
+                        }
+                        catch (Exception ex)
```
ayende, line 264: What will cause this error here?

arekpalinski: It's very unlikely we will get an error here as we already succeeded in opening it once, so I don't see why we could fail on the second attempt. Just defensive coding.

ayende: But that was the user's issue, no? Any idea what caused it?

arekpalinski: No, the user issue  was that we succeeded in opening the lucene directory and creating an index instance:

https://github.com/ravendb/ravendb/blob/8936425e9eda9975a189890cae73d16eaba78b85/Raven.Database/Indexing/IndexStorage.cs#L201-L202

but failed in:

https://github.com/ravendb/ravendb/blob/8936425e9eda9975a189890cae73d16eaba78b85/Raven.Database/Indexing/IndexStorage.cs#L204

In result `indexImplementation.Dispose()` was called which closes the lucene directory. But we were still using it in ` TryRecoveringIndex`
https://github.com/ravendb/ravendb/blob/8936425e9eda9975a189890cae73d16eaba78b85/Raven.Database/Indexing/IndexStorage.cs#L278

PR #6346, opened 2018-05-27
RavenDB-11078 PR comments

talweiss1982:


src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/JintLuceneDocumentConverter.cs
```diff
@@ -72,13 +66,14 @@ protected override int GetFields<T>(T instance, LazyStringValue key, object docu
                 object value;
                 if (obj.IsObject() && obj.IsArray() == false)
                 {
-                    //Note that for dynamic fields 'field.Name' will contain the real name of the field and not 'property'
+                    //In case TryDetectDynamicFieldCreation finds a dynamic field it will populate 'field.Name' witht the actual property name
```
ayende, line 69: `witht`

ayende: Will this mutate the actual `field.Name` instance?

talweiss1982: yes, this is why we pass the field to the method.

talweiss1982: see https://github.com/ravendb/ravendb/blob/v4.1/src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/JintLuceneDocumentConverter.cs#L137

ayende: Isn't this the field that we just pulled from the cache, so we are mutating a global value here?

talweiss1982: it shouldn't because the field name is originaly '_' and we modify it to something else. but I'll need to verify this.

ayende: I'm pretty sure that we aren't actually verifying that the field name is `_` (and we shouldn't)

PR #6352, opened 2018-05-28
[4.0] RavenDB-11152 Auto Indexes - 'Lock Mode' options are not relevant

Danielle9897:


src/Raven.Client/Documents/Operations/Indexes/SetIndexesLockOperation.cs
```diff
@@ -34,6 +37,18 @@ public SetIndexesLockOperation(Parameters parameters)
...
             _parameters = parameters;
+
+            FilterAutoIndexes();
+        }
+
+        private void FilterAutoIndexes()
+        {
+            _parameters.IndexNames = _parameters.IndexNames.Where(indexName => indexName.StartsWith("Auto/", StringComparison.OrdinalIgnoreCase) == false).ToArray();
+            
+            if (_parameters.IndexNames.Length == 0)
```
ppekrol: Better to throw when there any any auto indexes in the list instead of silently filtering them out:
```
var anyAutoIndexes = _parameters.IndexNames
.Where(indexName => indexName.StartsWith("Auto/", StringComparison.OrdinalIgnoreCase)
.Any();
```

src/Raven.Server/Documents/Handlers/IndexHandler.cs
```diff
@@ -500,6 +500,15 @@ public Task Status()
                 var json = await context.ReadForMemoryAsync(RequestBodyStream(), "index/set-lock");
                 var parameters = JsonDeserializationServer.Parameters.SetIndexLockParameters(json);
 
+                if (parameters.IndexNames == null || parameters.IndexNames.Length == 0)
+                    throw new ArgumentNullException(nameof(parameters.IndexNames));
+                
+                // Apply setLock only for static indexes
+                parameters.IndexNames = parameters.IndexNames.Where(indexName => indexName.StartsWith("Auto/", StringComparison.OrdinalIgnoreCase) == false).ToArray();
```
ppekrol: Let's not ignore them. Throw here.

ppekrol (changes requested):

ppekrol:
We should also add explicit exception in Index.SetLock when we have Auto-Index.

src/Raven.Client/Documents/Operations/Indexes/SetIndexesLockOperation.cs
Danielle9897: If I throw - as you request - then when the list that is passed contains a mixture of both static & auto indexes then work will Not be done at all for the static indexes

Is that the wanted flow?
The way it is now - work is still done for the static-indexes

What do you think ?

src/Raven.Server/Documents/Handlers/IndexHandler.cs
Danielle9897: same question as above

src/Raven.Client/Documents/Operations/Indexes/SetIndexesLockOperation.cs
ppekrol: Answered on Slack

Danielle9897: done

src/Raven.Server/Documents/Handlers/IndexHandler.cs
Danielle9897: done

PR #6353, opened 2018-05-28
RavenDB-10903 Rename Testing.ParentProcessId to Embedded.ParentProcessId

IdanHaim:


src/Raven.Server/Config/Categories/EmbeddedConfiguration.cs
```diff
@@ -0,0 +1,47 @@
...
+namespace Raven.Server.Config.Categories
+{
+    public class EmbeddedConfiguration : ConfigurationCategory
+    {
+        private int? _legacyParentProcessId;
+        private int? _parentProcessId;
+
+        [Description("Watch the parent process id and exit when it exited as well (for tests)")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Testing.ParentProcessId", ConfigurationEntryScope.ServerWideOnly)]
```
ppekrol, line 13: I think that we should add support for multiple keys in configuration entry. We do not really care from which key it was set probably or drop 'Testing.ParentProcessId' entirely.

ayende: Good idea

IdanHaim: Fixed

ayende: No, we want a _single_ property, not LegacyParentProcessId and ParentProcessId

IdanHaim: dropping the Testing.ParentProcessId?

PR #6362, opened 2018-05-28
RavenDB-10903 Rename Testing.ParentProcessId to Embedded.ParentProcessId

IdanHaim:


src/Raven.Server/Web/System/DatabaseHelper.cs
```diff
@@ -23,7 +25,7 @@ private static string[] GetServerWideOnlyConfigurationKeys()
 
                 foreach (var categoryProperty in configurationProperty.PropertyType.GetProperties(BindingFlags.Instance | BindingFlags.Public))
                 {
-                    var configurationEntryAttribute = categoryProperty.GetCustomAttribute<ConfigurationEntryAttribute>();
+                    var configurationEntryAttribute = categoryProperty.GetCustomAttributes<ConfigurationEntryAttribute>().FirstOrDefault();
```
ppekrol: Both of the keys should be inserted, not only the first.

Also check 'RavenConfiguration.AllConfigurationKeys'. We need to make changes there also.

src/Raven.Server/Config/Attributes/ConfigurationEntryAttribute.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace Raven.Server.Config.Attributes
 {
-    [AttributeUsage(AttributeTargets.Property)]
+    [AttributeUsage(AttributeTargets.Property, AllowMultiple = true)]
```
ppekrol, line 12: If you want to go that way, then please validate (ConfigurationCategory.Initialize) that all of the attributes have the same scope.

IdanHaim: Fixed

PR #6365, opened 2018-05-28
[4.0] RavenDB-11152 Auto Indexes - 'Lock Mode' options are not relevant - P…

Danielle9897:
…revent setting the Database Record


ppekrol:
There will be more changes here. Already discussed with @Danielle9897

PR #6366, opened 2018-05-28
[v4.0]RavenDB-11200: Fixing test: we had parallel usage of JsonOperationCon…

maximburyak:
…text, due to parallel blittable field access

### DIFF
--- a/test/SlowTests/Client/Subscriptions/CriteriaScript.cs
+++ b/test/SlowTests/Client/Subscriptions/CriteriaScript.cs
@@ -1,7 +1,9 @@
 ﻿using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Security.Cryptography.X509Certificates;
+using System.Threading;
 using System.Threading.Tasks;
 using FastTests.Client.Subscriptions;
 using Raven.Client.Documents.Operations;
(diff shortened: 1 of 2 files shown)


test/SlowTests/Client/Subscriptions/CriteriaScript.cs
```diff
@@ -128,32 +132,27 @@ select project(d)
                     };
 
                     var subsId = subscriptionManager.Create(subscriptionCreationParams);
-                    using (var subscription = subscriptionManager.GetSubscriptionWorker<BlittableJsonReaderObject>(new SubscriptionWorkerOptions(subsId)))
+                    using (var subscription = subscriptionManager.GetSubscriptionWorker<Thing>(new SubscriptionWorkerOptions(subsId)))
```
ayende, line 135: This looks like something that user would do?

maximburyak: Yeah, that's an interesting question, because "formally", we say that in blittable and contextes we do not support parallel usage, but "intuitively" it seems "ok" to be able to read blittables parallelly, the problem is that when we access strings in a blittable, we use the context's LazyStringValue object pool, and therefore it's not possible to do that, unless we disconnect blittable from the context..

ayende: The problem is that this is one case of very obvious pitfall at the _user_ level.
I'm asking if we want to prevent that scenario.
Most blittable scenarios are on our code, and we know that it should be single threaded.

maximburyak: First, we dont expect the average user to use blittables. Second, there will always be pitfalls like that, for example, if we had another context.read outside the subscription task, in that case there is no way we can accept that. We could add flag to blittable reader and to context.read that will tell them not to use object pooling and other non mandatory context services, but we have to remember that it won't prevent all pitfalls and we should probably document common ones..

PR #6368, opened 2018-05-28
Raven db 11097

talweiss1982:


src/Raven.Server/Documents/Handlers/IndexHandler.cs
```diff
@@ -751,6 +756,117 @@ public Task SuggestIndexMerge()
...
+        [RavenAction("/databases/*/indexes/try", "POST", AuthorizationStatus.ValidUser)]
+        public async Task TestJavaScriptIndex()
+        {
+            using (ContextPool.AllocateOperationContext(out DocumentsOperationContext context))
+            {
+                var input = await context.ReadForMemoryAsync(RequestBodyStream(), "TestJavaScriptIndex");
+                if (input.TryGet("Definition", out BlittableJsonReaderObject index) == false)
+                    ThrowRequiredPropertyNameInRequest("Definition");
+
+                var indexDefinition = JsonDeserializationServer.IndexDefinition(index);
```
ayende, line 768: Why aren't you just using the root object here?

src/Raven.Server/Documents/Handlers/IndexHandler.cs
```diff
@@ -751,6 +756,117 @@ public Task SuggestIndexMerge()
...
+
+                if (indexDefinition.Maps == null || indexDefinition.Maps.Count == 0)
+                    throw new ArgumentException("Index must have a 'Maps' fields");
+
+                indexDefinition.Type = indexDefinition.DetectStaticIndexType();
+
+                if (indexDefinition.Type.IsJavaScript() == false)
+                    throw new UnauthorizedAccessException("Testing indexes is only allowed for JavaScript indexes.");
+
+                if (indexDefinition.Name.StartsWith(Constants.Documents.Indexing.SideBySideIndexNamePrefix, StringComparison.OrdinalIgnoreCase))
```
ayende, line 778: Why do we care about the index name?

src/Raven.Server/Documents/Handlers/IndexHandler.cs
```diff
@@ -751,6 +756,117 @@ public Task SuggestIndexMerge()
...
+                {
+                    throw new ArgumentException(
+                        $"Index name must not start with '{Constants.Documents.Indexing.SideBySideIndexNamePrefix}'. Provided index name: '{indexDefinition.Name}'");
+                }
+
+                var compiledIndex = new JavaScriptIndex(indexDefinition, Database.Configuration);
+
+                var pageSize = GetIntValueQueryString("pageSize", false) ?? defaultPageSizeForTestingJavaScriptIndex;
```
ayende, line 786: That might not be a good idea, if they are filtering in the index.
Why not just get an array of document ids, instead?

src/Raven.Server/Documents/Handlers/IndexHandler.cs
```diff
@@ -751,6 +756,117 @@ public Task SuggestIndexMerge()
...
+                    }
+
+                    var mapRes = new List<ObjectInstance>();
+                    //all maps
+                    foreach (var ListOfFunctions in compiledIndex.Maps)
+                    {
+                        //multi maps per collection
+                        foreach (var mapFunc in ListOfFunctions.Value)
+                        {
+                            foreach (var res in mapFunc(docsPerCollection[ListOfFunctions.Key]))
```
ayende, line 804: How would this handle a multi map index on different collections?

src/Raven.Server/Documents/Handlers/IndexHandler.cs
talweiss1982: Because nobody responded to my request to define the controller requirement for the studio. Below you want to pass ids where should they reside, inside the index definitions?
I expected a million comments on my work I planned ahead.

src/Raven.Server/Documents/Handlers/IndexHandler.cs
talweiss1982: This is a test controller we should warn the user he is doing something wrong if we can.

src/Raven.Server/Documents/Handlers/IndexHandler.cs
ayende: Okay, make sense. We'll need an array of document ids there, for sure.

src/Raven.Server/Documents/Handlers/IndexHandler.cs
ayende: No, this is a test for JS only, not for everything else. We shouldn't care about the name here.

src/Raven.Server/Documents/Handlers/IndexHandler.cs
talweiss1982: This is the best way to go by default since you don't need to define anything and is more likely to run into unexpected document structure than to define a set of documents.
I would much rather write the index and maybe set the "pageSize" which might be a misleading name since it isn't the result's pageSize but the input's.
We can support both but I still think this should be the default.
I'm still waiting for requirements, without any I just implemented what I thought is the best approach, I expect @ml054 to come up with more UI stuff later on.

src/Raven.Server/Documents/Handlers/IndexHandler.cs
talweiss1982: I don't follow your question, this is the code that is doing exactly what you're asking about.

talweiss1982: docsPerCollection is input per collection, this is why an array of ids is going to be annoying since i'll have to read them check their metadata bucket them into their respected collection...

PR #6380, opened 2018-05-29
RavenDB-8956 fixing lucene

ppekrol:


src/Raven.Server/Documents/Queries/QueryMetadata.cs
```diff
@@ -67,7 +67,8 @@ public QueryMetadata(string query, BlittableJsonReaderObject parameters, ulong c
                                   && (OrderBy[0].OrderingType == OrderByFieldType.Implicit || OrderBy[0].OrderingType == OrderByFieldType.String)
                                   && HasExplanations == false
                                   && HasHighlightings == false
-                                  && IsDistinct == false;
+                                  && IsDistinct == false
+                                  && IsGroupBy == false;
```
ayende, line 71: why group by matters here? Shouldn't we still be able to run over the indexed fields?

ppekrol: The problem is that GetStringIndex returns invalid result (not complete) when we are grouping by array content. This occurs e.g. in RavenDB_8761.Can_group_by_array_content

ayende: I'n not sure that I'm following on that. 
What is not complete? It is supposed to give you (any of) single value from the array, typically the first one, but not guaranteed.

PR #6385, opened 2018-05-30
RavenDB-8956 fixed invalid order by

ppekrol:
I will create separate issue to detect (and throw) when we are trying to order by group by field that is grouping by array content


ppekrol:
http://issues.hibernatingrhinos.com/issue/RavenDB-11245

PR #6389, opened 2018-05-30
RavenDB-1127: RavenDB-11137: Allow failure before 'Send' success afte…

aviviadi:
…r DB unload

### DIFF
--- a/src/Raven.Client/Documents/Operations/Operation.cs
+++ b/src/Raven.Client/Documents/Operations/Operation.cs
@@ -60,7 +60,14 @@ namespace Raven.Client.Documents.Operations
 
         private async void ConnectionStatusEvent(object sender, EventArgs e)
         {
-            await FetchOperationStatus().ConfigureAwait(false);
+            try
+            {
+                await FetchOperationStatus().ConfigureAwait(false);
+            }
+            catch
+            {
+                // ignore
+            }
         }
 
         protected virtual void StopProcessing()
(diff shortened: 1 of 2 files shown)


src/Raven.Client/Documents/Operations/Operation.cs
```diff
@@ -54,12 +54,25 @@ public Operation(RequestExecutor requestExecutor, Func<IDatabaseChanges> changes
             var observable = changes.ForOperationId(_id);
             _subscription = observable.Subscribe(this);
             await observable.EnsureSubscribedNow().ConfigureAwait(false);
-
+            changes.ConnectionStatusChanged += ConnectionStatusEvent;
             await FetchOperationStatus().ConfigureAwait(false);
         }
 
+        private async void ConnectionStatusEvent(object sender, EventArgs e)
```
ayende, line 61: Let's avoid `async void` and friends.
And we need to call to `TrySetException` in this case, no?

ppekrol: @aviviadi what do you think about this approach: https://github.com/ppekrol/ravendb/commits/RavenDB-11137 ?

aviviadi: @ayende - if we TrySetException, we might report failure on a successful operation. Hunged client (with hopefully timeout set) is more reasonable for operation with no knowledge if it went all the way or not (and the connection is probably "dead" anyhow at this point)

aviviadi: @ppekrol - obviously that's better (fixes the above remark)

ayende: Yes, let's go ahead and merge this

PR #6393, opened 2018-05-30
RavenDB-11253 Failed test - CanImportExportAndBackupWithClusterTransa…

karmeli87:
…ctions

### DIFF
--- a/src/Raven.Server/Documents/DatabasesLandlord.cs
+++ b/src/Raven.Server/Documents/DatabasesLandlord.cs
@@ -171,14 +171,19 @@ namespace Raven.Server.Documents
                 }
                 return;
             }
+
+            RavenTransaction rtnCtx = null;
             try
             {
                 using (_serverStore.ContextPool.AllocateOperationContext(out TransactionOperationContext serverContext))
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -171,14 +171,19 @@ private void NotifyPendingClusterTransaction(string databaseName, Task<DocumentD
...
             try
             {
                 using (_serverStore.ContextPool.AllocateOperationContext(out TransactionOperationContext serverContext))
                 using (database.DocumentsStorage.ContextPool.AllocateOperationContext(out DocumentsOperationContext docContext))
-                using (serverContext.OpenReadTransaction())
                 using (docContext.OpenReadTransaction())
                 {
-                    ExecutePendingClusterTransactions(database, index, docContext, serverContext);
+                    rtnCtx = serverContext.OpenReadTransaction();
+                    var tasks = ExecutePendingClusterTransactions(database, index, docContext, serverContext);
```
ayende, line 183: This looks like it is wrong to me. You are _not_ allowed to use a single transaction concurrently from multiple threads.
But you are spawning off multiple tasks here?

PR #6406, opened 2018-05-31
[v4.1] Raven db 11234 various memory utilization improvements

maximburyak:


src/Sparrow/ByteString.cs
```diff
@@ -590,7 +590,31 @@ public void LowMemoryOver()
...
+                        var segment = current.Value;
+                        current = current.Next;
+                        if (segment == null)
+                            continue;
+                        if (!segment.InUse.Raise())
+                            continue;
+                        segment.Dispose();
+                    }
+                }
+                catch (Exception)
```
ayende: This is wrong, we should _never_ throw from finalizer.

ayende (changes requested):

src/Sparrow/ByteString.cs
maximburyak: There is no difference between this finalizer and all other finalizers in that file. Other finalizer prevent throwing "ObjectDisposedException" but other exception won't be treated. And frankly, if you have memory management exception during finalization, you better loose the process then continue working in invalid memory state.

PR #6415, opened 2018-06-01
Feature/custom default order

ryanheath:
As discussed at https://groups.google.com/forum/#!topic/ravendb/Pq_U9Ch-d3E
I have implemented a way to be able to have the same order and filter semantics a timespan has 
but now for a custom type.

I needed to have a QueryConverter that could convert to string but also to long type.
So I ended up with a converter to object, we probably need a better API here ...

// Ryan

### DIFF
--- a/src/Raven.Client/Documents/Conventions/DocumentConventions.cs
+++ b/src/Raven.Client/Documents/Conventions/DocumentConventions.cs
@@ -35,6 +35,7 @@ namespace Raven.Client.Documents.Conventions
         public delegate LinqPathProvider.Result CustomQueryTranslator(LinqPathProvider provider, Expression expression);
 
         public delegate bool TryConvertValueForQueryDelegate<in T>(string fieldName, T value, bool forRange, out string strValue);
+        public delegate bool TryConvertValueToObjectForQueryDelegate<in T>(string fieldName, T value, bool forRange, out object objValue);
 
         internal static readonly DocumentConventions Default = new DocumentConventions();
 
@@ -44,6 +45,11 @@ namespace Raven.Client.Documents.Conventions
 
         private readonly List<(Type Type, TryConvertValueForQueryDelegate<object> Convert)> _listOfQueryValueConverters =
             new List<(Type, TryConvertValueForQueryDelegate<object>)>();
+
+        private readonly List<(Type Type, TryConvertValueToObjectForQueryDelegate<object> Convert)> _listOfQueryValueToObjectConverters =
+            new List<(Type, TryConvertValueToObjectForQueryDelegate<object>)>();
+
+        private readonly Dictionary<Type, RangeType> _customRangeTypes = new Dictionary<Type, RangeType>();
 
         private readonly List<Tuple<Type, Func<string, object, Task<string>>>> _listOfRegisteredIdConventionsAsync =
             new List<Tuple<Type, Func<string, object, Task<string>>>>();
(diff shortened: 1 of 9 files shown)


src/Raven.Client/Documents/Conventions/DocumentConventions.cs
```diff
@@ -837,6 +848,51 @@ public bool TryConvertValueForQuery(string fieldName, object value, bool forRang
...
+                var entry = _listOfQueryValueToObjectConverters[index];
+                if (entry.Type.IsAssignableFrom(typeof(T)))
+                    break;
+            }
+
+            _listOfQueryValueToObjectConverters.Insert(index, (typeof(T), Actual));
+
+            bool Actual(string name, object value, bool forRange, out object objValue)
+            {
+                if (value is T)
```
ayende: `value is T t` is shorter

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
```diff
@@ -837,6 +848,51 @@ public bool TryConvertValueForQuery(string fieldName, object value, bool forRang
...
+                objValue = null;
+                return false;
+            }
+        }
+
+        public void RegisterQueryValueConverter<T>(TryConvertValueToObjectForQueryDelegate<T> converter, RangeType rangeType)
+        {
+            RegisterQueryValueConverter(converter);
+
+            if (_customRangeTypes.ContainsKey(typeof(T)) == false)
```
ayende, line 878: This might be confusing to the user, if they try to register two value converters, they'll get the first one?

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
```diff
@@ -837,6 +848,51 @@ public bool TryConvertValueForQuery(string fieldName, object value, bool forRang
...
+                _customRangeTypes.Add(typeof(T), rangeType);
+        }
+
+        internal bool TryConvertValueToObjectForQuery(string fieldName, object value, bool forRange, out object objValue)
+        {
+            foreach (var queryValueConverter in _listOfQueryValueToObjectConverters)
+            {
+                if (queryValueConverter.Type.IsInstanceOfType(value) == false)
```
ayende, line 886: Just call them until you find someone that can process it, no need for the checks

src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
```diff
@@ -1546,6 +1546,9 @@ private object TransformValue(WhereParams whereParams, bool forRange = false)
             if (_conventions.TryConvertValueForQuery(whereParams.FieldName, whereParams.Value, forRange, out var strVal))
                 return strVal;
 
+            if (_conventions.TryConvertValueToObjectForQuery(whereParams.FieldName, whereParams.Value, forRange, out var objValue))
```
ayende, line 1549: `TryConvertValueToObjectForQuery` vs `TryConvertValueForQuery` - Why do we have both, and how do we explain to users the diff?

ayende: We can't change `TryConvertValueForQuery`, because of API compact.
But we can use a single location for this and wrap the `TryConvertValueForQuery` callback with one that outputs an object, so we'll only have to check for it once.

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
ryanheath: I can change it, but it was copied from the TryConvertValueForQuery version.
Shall I change both implementations?

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
ryanheath: I can change it, but it was copied from the TryConvertValueForQuery version.
Shall I change both implementations?

ryanheath: It seems those checks are here for fast shortcutting ...

src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
ryanheath: Yes, this is my concern as well. But I wanted some what more pointers before impl it.

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
ryanheath: Yes, so how I want to register a converter that can strongly typed convert to string (the existing API)
but I also want to register a version that can convert to object. (a new API).

ayende:
See my changes on top of yours

PR #6421, opened 2018-06-03
[4.0] RavenDB-11269

aviv86:


src/Raven.Client/Documents/Linq/RavenQueryProviderProcessor.cs
```diff
@@ -1975,6 +1980,17 @@ private void VisitSelect(Expression operand)
             }
         }
 
+        private string RemoveTransperentIdentifiersIfNeeded(MemberExpression memberExpression)
+        {
+            var selectPath = GetSelectPath(memberExpression);
+            while (selectPath.StartsWith("<>h__TransparentIdentifier"))
+            {
+                selectPath = selectPath.Substring(28);
```
ayende, line 1988: Are you sure this is `28` ? `<>h__TransparentIdentifier` is just 26 chars?

aviv86: yes, 26 + 1 for the number suffix and +1 for "."

ppekrol: Better to define const next time and use const.Length + 1 + 1 and add relevant comment explaining what does +1 are, so future maintainers will better understand what is going on.

PR #6423, opened 2018-06-04
[4.0] RavenDB-11133 Support mix version nodes cluster - WIP

karmeli87:


src/Raven.Server/Rachis/Candidate.cs
```diff
@@ -150,12 +151,19 @@ private void Run()
                             StateChange();
 
                             var connections = new Dictionary<string, RemoteConnection>();
+                            var versions = new List<int>();
                             foreach (var candidateAmbassador in _voters)
                             {
                                 connections[candidateAmbassador.Tag] = candidateAmbassador.Connection;
+                                if (candidateAmbassador.ClusterCommandsVersion > 0)
```
ayende, line 159: If this is 0, that is also information, it means the client is too old

src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
```diff
@@ -45,7 +45,16 @@ public class RachisAdminHandler : RequestHandler
                 HttpContext.Response.Headers["Reached-Leader"] = "true";
 
                 var commandJson = await context.ReadForMemoryAsync(RequestBodyStream(), "external/rachis/command");
-                var command = CommandBase.CreateFrom(commandJson);
+                CommandBase command;
+                try
+                {
+                    command = CommandBase.CreateFrom(commandJson);
+                }
+                catch (InvalidOperationException e)
```
ayende, line 53: What can cause this?

src/Raven.Server/Rachis/Candidate.cs
ayende: You need to distinguish between "don't know yet (other node is down)" to "The other node is too old to tell me"

src/Raven.Server/Rachis/RequestVote.cs
```diff
@@ -24,6 +26,7 @@ public class RequestVoteResponse
         public long Term { get; set; }
         public bool VoteGranted { get; set; }
         public bool NotInTopology { get; set; }
+        public int ClusterCommandsVersion { get; set; } = ClusterCommandsVersionManager.DefaultVersion;
```
ayende: Better to have a fixed version for this.

src/Raven.Server/ServerWide/ClusterCommandsVersionManager.cs
```diff
@@ -0,0 +1,131 @@
...
+
+namespace Raven.Server.ServerWide
+{
+    public static class ClusterCommandsVersionManager
+    {
+        public const int DefaultVersion = 400;
+
+        public static readonly int MyCommandsVersion;
+
+        public static int Max(IEnumerable<int> values)
```
ayende: What does this gives over `Enumerable.Max`?

src/Raven.Server/ServerWide/ClusterCommandsVersionManager.cs
```diff
@@ -0,0 +1,131 @@
...
+            foreach (var value in values)
+            {
+                if (value > max)
+                    max = value;
+            }
+
+            return max;
+        }
+
+        public static int Min(IEnumerable<int> values)
```
ayende: Same

src/Raven.Server/ServerWide/ClusterValidator.cs
```diff
@@ -0,0 +1,73 @@
...
+namespace Raven.Server.ServerWide
+{
+    public class ClusterValidator : RachisVersionValidation
+    {
+        public override void AssertPutCommandToLeader(CommandBase cmd)
+        {
+            var commandName = cmd.GetType().Name;
+            if (ClusterCommandsVersionManager.CanPutCommand(commandName) == false)
+            {
+                throw new RejectPutClusterCommandException($"Cannot accept the command '{commandName}', " +
```
ayende: Move to separate method

src/Raven.Server/ServerWide/ClusterValidator.cs
```diff
@@ -0,0 +1,73 @@
...
+            if (entry.TryGet(nameof(RachisEntry.Flags), out RachisEntryFlags flag) && flag != RachisEntryFlags.StateMachineCommand)
+            {
+                return;
+            }
+
+            string type = null;
+            if ((entry.TryGet(nameof(RachisEntry.Entry), out BlittableJsonReaderObject blittableEntry) &&
+                 blittableEntry.TryGet("Type", out type)) == false)
+            {
+                throw new RejectSendToFollowerException("Rachis entry is invalid!");
```
ayende: Because?

src/Raven.Server/ServerWide/ClusterValidator.cs
```diff
@@ -0,0 +1,73 @@
...
+            }
+
+            var myCommandVersion = ClusterCommandsVersionManager.ClusterCommandsVersions[type];
+
+            if (myCommandVersion > version)
+            {
+                throw new RejectSendToFollowerException($"The command '{type}' with the version {myCommandVersion} is not supported on follower {follower}.");
```
ayende: Move throws to dedicated method to allow inlining of this one

test/FastTests/Server/Cluster/VersionValidation.cs
```diff
@@ -0,0 +1,39 @@
...
+                var typeInfo = type.GetTypeInfo();
+                if (typeInfo.IsAbstract)
+                    continue;
+
+                if (typeInfo.IsSubclassOf(typeof(CommandBase)) == false)
+                    continue;
+
+                if (ClusterCommandsVersionManager.ClusterCommandsVersions.TryGetValue(type.Name, out int _))
+                    continue;
+                Console.WriteLine($"[nameof({type.Name})] = 404,");
```
ayende: Remove this.

ayende (changes requested):

src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
karmeli87: When the deserializer for the command doesn't exists.

src/Raven.Server/Rachis/Candidate.cs
karmeli87: OK, in any case I think we need to use a median here instead of taking the minimum of the voters.

src/Raven.Server/ServerWide/ClusterCommandsVersionManager.cs
karmeli87: Oops..  I've searched the Max function in Math.

src/Raven.Server/Rachis/Candidate.cs
karmeli87: Actually, the distinguish between old version and "don't know" happened when the candidate ambassador reads the 'RequestVoteResponse'. 
https://github.com/karmeli87/ravendb/blob/7eb891a9b7e09cda1b301160fe7cd3317a821fe7/src/Raven.Server/Rachis/RequestVote.cs#L29

src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
karmeli87: More concrete example is when a 4.1 follower sends 4.1 command to his 4.0 leader.

src/Raven.Server/Rachis/Candidate.cs
ayende: And it is at that you can take this into account.
But we need the minimum number among the majority that we know of.

src/Raven.Server/ServerWide/ClusterCommandsVersionManager.cs
```diff
@@ -12,32 +13,18 @@ namespace Raven.Server.ServerWide
...
-            {
-                if (value > max)
-                    max = value;
-            }
-
-            return max;
-        }
-
-        public static int Min(IEnumerable<int> values)
+        public static int Median(int[] values)
```
ayende: Why use the median here?

ayende (changes requested):

src/Raven.Server/ServerWide/ClusterCommandsVersionManager.cs
karmeli87: Cluster version is determined by the highest common version of the majority.
Which is simply the median :-)

ayende: no, the minimum versions of all the nodes that are there

src/Raven.Server/Config/Categories/ClusterConfiguration.cs
```diff
@@ -71,5 +71,9 @@ public class ClusterConfiguration : ConfigurationCategory
         [ConfigurationEntry("Cluster.HardDeleteOnReplacement", ConfigurationEntryScope.ServerWideOnly)]
         public bool HardDeleteOnReplacement { get; set; }
 
+        [Description("If exceeded, clamp the cluster to the specified version.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Cluster.MaximalVersion", ConfigurationEntryScope.ServerWideOnly)]
```
ayende, line 76: This needs to have a note on `EXPERT ONLY`

ayende: Also `MaximalVersion` ?

PR #6435, opened 2018-06-05
[4.0] RavenDB-11133 Support mix version nodes cluster

karmeli87:
Better naming


ayende:
BTW, we need these changes in 4.1 as well, for when we hit 4.2

PR #6437, opened 2018-06-05
[4.0] RavenDB-11277

aviv86:


ppekrol:
@aviv86 Let's also mark as obsolete the DocumentChange.TypeName property

ppekrol:
@aviv86 Can you also check the compilation?

PR #6438, opened 2018-06-05
[4.1] RavenDB-11277

aviv86:


src/Raven.Client/Documents/Changes/DatabaseChanges.cs
```diff
@@ -211,33 +211,22 @@ public IChangesObservable<DocumentChange> ForDocumentsInCollection<TEntity>()
...
-            var encodedTypeName = Uri.EscapeDataString(typeName);
-
-            var counter = GetOrAddConnectionState("types/" + typeName, "watch-type", "unwatch-type", encodedTypeName);
-
-            var taskedObservable = new ChangesObservable<DocumentChange, DatabaseConnectionState>(
-                counter,
-                notification => string.Equals(typeName, notification.TypeName, StringComparison.OrdinalIgnoreCase));
-
-            return taskedObservable;
+            throw new NotSupportedException($"Method {nameof(ForDocumentsOfType)} is obsolete.");
```
ayende, line 217: This is a breaking change for someone who might be calling this.
Might be better to return noop to continue the current behavior.

PR #6442, opened 2018-06-05
[v4.1] RavenDB-11250: Make sure that store is initialized when getting subsc…

maximburyak:
…ription worker


src/Raven.Client/Documents/DocumentStoreBase.cs
```diff
@@ -170,7 +170,7 @@ protected void EnsureNotClosed()
                 throw new ObjectDisposedException(GetType().Name, "The document store has already been disposed and cannot be used");
         }
 
-        protected void AssertInitialized()
+        public void AssertInitialized()
```
ppekrol, line 173: I will change this to internal

PR #6443, opened 2018-06-05
[v4.0] RavenDB-11250: Make sure that store is initialized when getting subsc…

maximburyak:
…ription worker


src/Raven.Client/Documents/IDocumentStore.cs
```diff
@@ -171,5 +171,7 @@ public interface IDocumentStore : IDisposalNotification
         DatabaseSmuggler Smuggler { get; }
 
         IDisposable SetRequestTimeout(TimeSpan timeout, string database = null);
+
+        void AssertInitialized();
```
ayende: I don't like it on the root interface. The user will never call it.
Better to do downcast when needed.

ayende (changes requested):

PR #6450, opened 2018-06-05
RavenDB-11295 Stepping down in a 2 nodes cluster will break the cluster

karmeli87:


ayende:
don't we need this in 4.1 as well?

karmeli87:
Yes, This bug I introduced with RavenDB-11133, which is not implemented yet in 4.1.

PR #6455, opened 2018-06-06
Raven db 8647

ml054:


src/Raven.Server/SqlMigration/MsSQL/MsSqlDatabaseMigrator.cs
```diff
@@ -105,9 +105,11 @@ private ColumnType MapColumnType(string type)
                     return ColumnType.Boolean;
                 case "binary":
                 case "varbinary":
+                case "image":
```
ayende, line 108: See here:
https://www.w3schools.com/sql/sql_datatypes.asp

missing more things (tinytext, MEDIUMBLOB, etc)

ml054: ok updated ticket - thanks

PR #6458, opened 2018-06-06
[4.1] RavenDB 11133 Support mix version nodes cluster

karmeli87:


src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -206,7 +206,7 @@ public abstract partial class InMemoryDocumentSessionOperations : IDisposable
             MaxNumberOfRequestsPerSession = _requestExecutor.Conventions.MaxNumberOfRequestsPerSession;
             GenerateEntityIdOnTheClient = new GenerateEntityIdOnTheClient(_requestExecutor.Conventions, GenerateId);
             EntityToBlittable = new EntityToBlittable(this);
-            SessionInfo = new SessionInfo(_clientSessionId, false, _documentStore.LastTransactionIndex);
+            SessionInfo = new SessionInfo(_clientSessionId, false, () => _documentStore.GetLastTransactionIndex(DatabaseName));
```
ayende: Let's avoid the lamda here and just return the current value

src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
```diff
@@ -44,6 +45,13 @@ public void SetResult(BatchCommandResult result)
                 return;
             }
 
+            if (_session.TransactionMode == TransactionMode.ClusterWide)
+            {
+                if (result.TransactionIndex > 0 == false)
```
ayende: `result.TransactionIndex <= 0` ?

src/Raven.Client/Documents/Session/SessionInfo.cs
```diff
@@ -1,16 +1,18 @@
...
+﻿using System;
+
+namespace Raven.Client.Documents.Session
 {
     public class SessionInfo
     {
         public int? SessionId { get;}
 
-        public long? LastClusterTransaction { get;}
+        public Func<long?> GetLastClusterTransactionFunc;
```
ayende: Should switch that.

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -683,6 +684,16 @@ protected virtual bool TryLoadFromCache(JsonOperationContext context)
                                 }
 
                                 response = await preferredTask.ConfigureAwait(false);
+                                if (sessionInfo?.GetLastClusterTransactionFunc?.Invoke() != null)
```
ayende: This is a bad idea. You are relying on mutable global state that doesn't belong to you.
You might be reading some _other_ command state, for example.

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -227,6 +234,10 @@ public abstract class RachisConsensus : IDisposable
     {        
         internal abstract RachisStateMachine GetStateMachine();
 
+        internal abstract RachisVersionValidation GetValidator();
```
ayende, line 237: Why both `GetValidator` and `Validator`?

src/Raven.Server/RavenServerStartup.cs
```diff
@@ -154,6 +156,7 @@ public static bool IsHtmlAcceptable(HttpContext context)
             {
                 context.Response.StatusCode = (int)HttpStatusCode.OK;
                 context.Response.Headers["Content-Type"] = "application/json; charset=utf-8";
+                context.Response.Headers[Constants.Headers.ServerVersion] = RavenVersionAttribute.Instance.Build;
```
ayende, line 159: Not sure if this is a good idea, for dev build, that will always be 13, for example.
Means that we can't inter-operate between dev and official builds.
Better to use a fixed version, such as `4.0`, etc

src/Raven.Server/ServerWide/ClusterCommandsVersionManager.cs
```diff
@@ -0,0 +1,140 @@
...
+using Raven.Server.ServerWide.Commands.Subscriptions;
+using Sparrow.Logging;
+
+namespace Raven.Server.ServerWide
+{
+    public static class ClusterCommandsVersionManager
+    {
+        public static readonly int MyCommandsVersion;
+
+        public static int Median(int[] values)
```
ayende, line 20: Remove this

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -956,6 +961,11 @@ private void UpdateDatabase(TransactionOperationContext context, string type, Bl
             }
         }
 
+        public override RachisVersionValidation Validator()
+        {
+            return new ClusterValidator();
```
ayende, line 966: Do we really need to create a new instance every time?

ayende (changes requested):

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
karmeli87: The problem is, if we do several save changes in the same session.

ayende: See later, your usage of this in RequestExecuter isn't good

src/Raven.Server/RavenServerStartup.cs
karmeli87: The dev build for 4.0 is 40 and for 4.1 is 41.

karmeli87:
Rest of the comments are common to 4.0 version. So I'll address them there and cherry-pick.

src/Raven.Server/Routing/RequestRouter.cs
```diff
@@ -129,6 +135,22 @@ public RouteInformation GetRoute(string method, string path, out RouteMatch matc
             return reqCtx.Database?.Name;
         }
 
+        public static void AssertClientVersion(HttpContext context, Exception innerException)
+        {
+            // client in this context could be also a follower sending a command to his leader.
+            if (context.Request.Headers.TryGetValue(Constants.Headers.ClientVersion, out var versionHeader) && 
+                Version.TryParse(versionHeader, out var clientVersion))
+            {
+                if (clientVersion.Build > ServerVersion.Build)
```
ppekrol, line 144: As a continuation of:
https://github.com/ravendb/ravendb/pull/6478#discussion_r193503446

Is this correct now? We are sending here strings like `4.1.5.0`. Which means that we are comparing only number `5` here? Which is a semver 'Patch' with `ServerVersion.Build` which is a build number e.g. `40038`.

ayende: That should be fine here, we explicitly state that we want to allow servers from the same minor version to cooperate together

ppekrol: Shouldn't this be changed to:
```csharp
if (clientVersion.Major != RavenVersionAttribute.Instance.MajorVersion || clientVersion.Minor != RavenVersionAttribute.Instance.MinorVersion)
```

where `AssemblyVersion` is:

```csharp
    [AttributeUsage(AttributeTargets.Assembly)]
    public class RavenVersionAttribute : Attribute
    {
        public string CommitHash { get; set; }
        public string Build { get; set; }
        public string Version { get; set; }
        public string FullVersion { get; set; }

        private static int? _buildVersion;
        private static readonly Version _assemblyVersion;

        private static RavenVersionAttribute _instance;

        public static RavenVersionAttribute Instance => _instance ?? (_instance = (RavenVersionAttribute)typeof(RavenVersionAttribute).GetTypeInfo().Assembly.GetCustomAttributes(typeof(RavenVersionAttribute)).Single());

        static RavenVersionAttribute()
        {
            _assemblyVersion = typeof(RavenVersionAttribute).GetTypeInfo().Assembly.GetName().Version;
        }

        public RavenVersionAttribute()
        {
            MajorVersion = _assemblyVersion.Major;
            MinorVersion = _assemblyVersion.Minor;
            PatchVersion = _assemblyVersion.Build;
            AssemblyVersion = $"{MajorVersion.ToInvariantString()}.{MinorVersion.ToInvariantString()}.{PatchVersion.ToInvariantString()}.{BuildVersion.ToInvariantString()}";
        }

        public readonly string AssemblyVersion;

        public readonly int MajorVersion;

        public readonly int MinorVersion;

        public readonly int PatchVersion;

        public int BuildVersion
        {
            get
            {
                if (_buildVersion == null)
                {
                    if (int.TryParse(Build, out var buildVersion) == false)
                    {
                        _buildVersion = 41;
                    }
                    else
                    {
                        _buildVersion = buildVersion;
                    }
                }

                return _buildVersion.Value;
            }
        }
    }
```

?

ayende: That works

PR #6461, opened 2018-06-06
RavenDB-11089 (4.1)

myarichuk:


src/Raven.Client/Documents/Linq/RavenQueryProvider.cs
```diff
@@ -31,18 +32,22 @@ internal class RavenQueryProvider<T> : IRavenQueryProvider
         private readonly QueryStatistics _queryStatistics;
         private readonly LinqQueryHighlightings _highlightings;
         private readonly bool _isMapReduce;
+        private DocumentConventions _conventions;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="RavenQueryProvider{T}"/> class.
         /// </summary>
         public RavenQueryProvider(
+#if FEATURE_HIGHLIGHTING
```
ayende, line 41: Why is this under #if def?

ppekrol: It needs to be removed entirely. Probably git cherry-pick issue.

PR #6462, opened 2018-06-06
RavenDB-11089 -> additional fix for edgecase

myarichuk:
* make sure not to apply serialization convention on system types in the Linq provider
* some more tests

### DIFF
--- a/src/Raven.Client/Documents/Indexes/ExpressionStringBuilder.cs
+++ b/src/Raven.Client/Documents/Indexes/ExpressionStringBuilder.cs
@@ -129,7 +129,12 @@ namespace Raven.Client.Documents.Indexes
 
             string name = null;
             if (_conventions.PropertyNameConverter != null)
-                name = _conventions.PropertyNameConverter(member);
+            {   
+                //do not use convention for types in system namespaces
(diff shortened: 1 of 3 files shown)


src/Raven.Client/Documents/Indexes/ExpressionStringBuilder.cs
```diff
@@ -129,7 +129,12 @@ private void OutMember(Expression instance, MemberInfo member, Type exprType)
 
             string name = null;
             if (_conventions.PropertyNameConverter != null)
-                name = _conventions.PropertyNameConverter(member);
+            {   
+                //do not use convention for types in system namespaces
+                if(member.DeclaringType?.Namespace?.StartsWith("System") == false &&
+                   member.DeclaringType?.Namespace?.StartsWith("Microsoft") == false)
+                    name = _conventions.PropertyNameConverter(member);
```
ppekrol, line 136: @myarichuk Do we need to apply this in PropertyNameConventionJSMetadataProvider also?

myarichuk: Yeah, I think we should apply it there too.

PR #6463, opened 2018-06-06
[4.0] RavenDB-11151 Terms of a Faulty Index - Error is received but is Not …

Danielle9897:
…presented to user

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Debugging/IndexDebugExtensions.cs
+++ b/src/Raven.Server/Documents/Indexes/Debugging/IndexDebugExtensions.cs
@@ -436,7 +436,7 @@ namespace Raven.Server.Documents.Indexes.Debugging
                 case IndexType.AutoMapReduce:
                     return ((AutoMapReduceIndex)self).Definition.IndexFields.Keys.ToArray();
                 case IndexType.Faulty:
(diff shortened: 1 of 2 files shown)


src/Raven.Server/Documents/Indexes/Debugging/IndexDebugExtensions.cs
```diff
@@ -435,6 +435,8 @@ public static string[] GetEntriesFields(this Index self)
                     return ((AutoMapIndex)self).Definition.IndexFields.Keys.ToArray();
                 case IndexType.AutoMapReduce:
                     return ((AutoMapReduceIndex)self).Definition.IndexFields.Keys.ToArray();
+                case IndexType.Faulty:
+                    throw new ArgumentException($"Index '{self.Name}' is Faulty. Can't return index entries for a faulty index.");
```
ppekrol: InvalidOperatonException would be better here

ppekrol (changes requested):

src/Raven.Server/Documents/Indexes/Debugging/IndexDebugExtensions.cs
Danielle9897: done

PR #6474, opened 2018-06-06
[4.1] RavenDB-11010

aviv86:
add support for selecting counter in session.Query


src/Raven.Client/Documents/Queries/RavenQuery.cs
```diff
@@ -34,6 +34,21 @@ public static IMetadataDictionary Metadata<T>(T instance)
...
+        {
+            throw new NotSupportedException("This method is here for strongly type support of server side call during Linq queries and should never be directly called");
+        }
+
+        public static long? Counter(string docId, string name)
+        {
+            throw new NotSupportedException("This method is here for strongly type support of server side call during Linq queries and should never be directly called");
+        }
+
+        public static long? Counter<T>(T path, string name)
```
ayende, line 49: `T path` should be `object documentInstance`

PR #6478, opened 2018-06-06
[4.1] RavenDB 11133

karmeli87:


src/Raven.Server/RavenServerStartup.cs
```diff
@@ -156,7 +156,7 @@ public static bool IsHtmlAcceptable(HttpContext context)
             {
                 context.Response.StatusCode = (int)HttpStatusCode.OK;
                 context.Response.Headers["Content-Type"] = "application/json; charset=utf-8";
-                context.Response.Headers[Constants.Headers.ServerVersion] = RavenVersionAttribute.Instance.Build;
+                context.Response.Headers[Constants.Headers.ServerVersion] = RavenVersionAttribute.Instance.Version;
```
ppekrol, line 159: I think it would be 'safer' (for future purposes) to put the same value as we put in Raven-Client-Version
```
typeof(RavenServerStartup).GetTypeInfo().Assembly.GetName().Version.ToString();
```

ayende: The problem here is that this changes for dev.
We don't usually have issues for dev client to server, but we'll have server to server a lot, and existing client to dev servers.

ppekrol: TBH, till this day I thought that we are sending either:
a) `major.minor.patch.build` e.g. `4.0.3.40038`
b) `major.minor.patch-buildtype-build` e.g. `4.0.3-patch-40038`

but the truth is that `RequestExecutor.ClientVersion` is returning only `major.minor.patch.0` e.g. `4.0.3.0`, so it does not contain build version anyway, but at least there is a `patch` number.

Using the `RavenVersionAttribute.Instance.Version` will yield only `4.1` for all `4.1` servers. I'm just afraid that this might not be enough someday. We had this issue in the past where certain tools (IIRC smuggler) were modifying its behavior depending on the server version.

karmeli87: So we could use the FullVersion? Since we interested `major.minor.patch` this should be enough, no?

ppekrol: For best flexibility we should go with: `major.minor.patch.build` and apply this change also to the client. But `typeof(RavenServerStartup).GetTypeInfo().Assembly.GetName().Version.ToString();` could be enough. @ayende ?

ayende: I'm fine with either option, go ahead

ppekrol: @karmeli87 let's go with `major.minor.patch.build` for both client and server (4.0 and 4.1)

PR #6486, opened 2018-06-07
v4.0 Raven db 10692

ml054:


ayende:
why closed?

ml054:
Pawel wants to release new version, I will PR after release.

PR #6495, opened 2018-06-07
Raven db 10450

ayende:


src/Raven.Server/NotificationCenter/SlowWriteNotifications.cs
```diff
@@ -0,0 +1,135 @@
...
+                {
+                    Path = path,
+                    DataWrittenInMb = dataSizeInMb,
+                    DurationInSec = durationInSec,
+                    SpeedInMbPerSec = dataSizeInMb / durationInSec,
+                    Date = now
+                };
+            }
+
+            _slowWrites.AddOrUpdate(path, info, (s, i) => info);
```
ayende, line 58: We already did the check, why are we calling add or update here?

arekpalinski: Right, we can call `TryAdd` if key isn't there. If it is there we are updating in-memory reference anyway - no need to call `AddOrUpdate`. Thanks

PR #6509, opened 2018-06-08
v4.1 Raven db 10450b

ml054:


src/Raven.Server/NotificationCenter/SlowWriteNotifications.cs
```diff
@@ -0,0 +1,135 @@
...
+
+            using (_notificationsStorage.Read(id, out var ntv))
+            {
+                if (ntv == null || ntv.Json.TryGet(nameof(PerformanceHint.Details), out BlittableJsonReaderObject detailsJson) == false || detailsJson == null)
+                {
+                    details = new SlowWritesDetails();
+                }
+                else
+                {
+                    details = (SlowWritesDetails)EntityToBlittable.ConvertToEntity(
```
maximburyak, line 111: Not a fan of using EntityToBlittable on server side, in that case it wont be too frequent though..

arekpalinski: It's easy to use. Is there any problem with that? We use it quite commonly when dealing with aggregated notifications (as you pointed, those aren't frequently running actions).

maximburyak: As long as it is not frequently called, it's bearable, but in general, I would feel more comfortable if we would use our "compiled" deserialization, which is more performant.

arekpalinski: Oh, yeah but it's not so important in those scenarios

PR #6518, opened 2018-06-08
[4.0] RavenDB-11317 all dates in logs should be local

ppekrol:


src/Sparrow/Logging/Logger.cs
```diff
@@ -48,7 +48,7 @@ public void Operations(FormattableString msg, Exception e = null)
 
         public void Info(string msg, Exception ex = null)
         {
-            _logEntry.At = DateTime.UtcNow;
+            _logEntry.At = DateTime.Now;
```
ayende, line 51: The dates in the logs are `UtcNow` for a reason.
In particular, `UtcNow` is _much_ faster than `Now`.
And this matters a lot in the logs. See:
https://blogs.msdn.microsoft.com/kirillosenkov/2012/01/10/datetime-utcnow-is-generally-preferable-to-datetime-now/

PR #6536, opened 2018-06-08
v4.1 RavenDB-11188 Consider removing phantomjs from studio

ml054:


ml054:
I will regenerate package-lock.json with VS as it is running very very old version.

PR #6537, opened 2018-06-08
[4.0] RavenDB-11317

ppekrol:
- logs should be in utc by default
- added the option to log in local server time zone


src/Raven.Server/Config/Categories/LogsConfiguration.cs
```diff
@@ -14,5 +14,9 @@ public class LogsConfiguration : ConfigurationCategory
         [DefaultValue(LogMode.Operations)]
         [ConfigurationEntry("Logs.Mode", ConfigurationEntryScope.ServerWideOnly)]
         public LogMode Mode { get; set; }
+
+        [DefaultValue(true)]
```
ppekrol, line 18: @ayende not sure about that default. Your call.

src/Sparrow/Logging/Logger.cs
```diff
@@ -13,6 +13,13 @@ public class Logger
         [ThreadStatic]
         private static LogEntry _logEntry;
 
+        private static readonly long LocalToUtcOffsetInTicks;
+
+        static Logger()
+        {
+            LocalToUtcOffsetInTicks = TimeZoneInfo.Local.GetUtcOffset(DateTime.Now).Ticks;
```
ayende, line 20: Might need to update it every now and then for things like daylight savings changes?

ayende: Can probably get away with doing that only once every file, or when we are about to go idle?

ayende (approved):

src/Sparrow/Logging/Logger.cs
ayende: I'll fix that in a bit

PR #6539, opened 2018-06-08
RavenDB-11317 - Handle updates of date time offsets on the fly

ayende:


src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -222,6 +230,16 @@ private FileStream GetNewStream(long maxFileSize)
             }
         }
 
+        private void UpdateLocalDateTimeOffset()
+        {
+            if (_updateLocalTimeOffset == false || UseUtcTime == false)
```
ppekrol: Shouldn't this be `if (_updateLocalTimeOffset == false || UseUtcTime)` ?

src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -405,6 +423,9 @@ private void BackgroundLogger()
                                     // we don't want to have fsync here, we just
                                     // want to send it to the OS
                                     currentFile.Flush(flushToDisk: false);
+                                    if (_hasEntries.IsSet == false)
```
ppekrol, line 426: Wondering if we will be able to hit this on a live huge system?

src/Sparrow/Logging/LoggingSource.cs
ayende: Yes

ppekrol:
LGTM

src/Sparrow/Logging/LoggingSource.cs
ayende: What do you mean? We want to avoid this on an ongoing basic, and we are, at this point, going to call the `Wait` method anyway.
Since we are about to go idle, we can afford to check if this is still up to date, no?

ppekrol: Ok

PR #6541, opened 2018-06-08
[4.0] RavenDB-11317 minor optimization

ppekrol:


src/Raven.Server/Program.cs
```diff
@@ -69,7 +69,7 @@ public static int Main(string[] args)
             configuration.Initialize();
 
             LoggingSource.Instance.SetupLogMode(configuration.Logs.Mode, configuration.Logs.Path.FullPath);
-            LoggingSource.UseUtcTime = configuration.Logs.UseUtcTime;
+            LoggingSource.UseUtcTime = configuration.Logs.UseUtcTime || LoggingSource.UseUtcTime == false && LoggingSource.LocalToUtcOffsetInTicks == 0;
```
ayende, line 72: I'm not sure that I'm following on the optimization here? 
`configuration.Logs.UseUtcTime` defaults to `true`, no?

ppekrol: I wanted to optimize a case when someone sets UseUtcTime to false but is in UTC timezone. But now I'm not sure if my assumption is correct, that if LocalToUtcOffsetInTicks is 0 then it will always stay 0.

ppekrol: And the answer to that is no, my assumption was invalid e.g. UK uses BST during the summer.

ayende: UTC != GMT. And offset 0 doesn't stay that way.
Also, humanity's time handling it stupid. I vote for using `.beat` from now on.
At `@989.9`

ppekrol: I still do not understand why we do not support Stardate :)

ayende: That is even more broken, actually.
It is only ever consistent with a _single episode_.

PR #6545, opened 2018-06-10
[v4.1]Raven db 11233 4.1

maximburyak:


src/Raven.Client/Documents/Changes/DatabaseChanges.cs
```diff
@@ -343,17 +343,14 @@ private DatabaseConnectionState GetOrAddConnectionState(string name, string watc
             catch (OperationCanceledException e)
             {
                 NotifyAboutError(e);
+                _tcs.TrySetException(e);
```
ayende, line 346: Shouldn't this be `TrySetCancelled`?

ayende:
I'm not sure how this is solving the race condition on the `OnError` method?

ppekrol:
It covers an issue where the exception happenes during the first initialization of the Changes API. I think this should be enough. We cannot have a consistent behavior when `OnError` would be passed to the `Changes()` method, due to caching of the `DatabaseChanges`. Which means that second call to that method would be racy again.

src/Raven.Client/Documents/Changes/DatabaseChanges.cs
maximburyak: I'm not sure, It's possible that we want to track the exception stack trace, or we just want to know that it was canceled?

ppekrol: I think we should go with `TrySetCanceled`. You will not get much more details here anyway.

PR #6547, opened 2018-06-10
[4.0] RavenDB-11065 ObjectDisposedException: Cannot access a disposed object. Object name: 'RemoteConnection'.

karmeli87:
Some small fixes are also included:
RavenDB-11068 The connection with node A was suddenly broken.
RavenDB-10510 Failed to talk with A, message: Cannot access a disposed object. Object name: 'RemoteConnection'.

RavenDB-11319 Cluster observer logs do not indicate the node related
RavenDB-11023 KeyNotFoundException: current.Length == 0 but we try to access the 'A' node

### DIFF
--- a/src/Raven.Server/Rachis/FollowerAmbassador.cs
+++ b/src/Raven.Server/Rachis/FollowerAmbassador.cs
@@ -9,13 +9,13 @@ using System.Threading;
 using Raven.Client.Exceptions;
 using Raven.Client.Http;
 using Raven.Client.ServerWide;
-using Raven.Server.ServerWide.Commands;
 using Raven.Server.ServerWide.Context;
 using Raven.Server.Utils;
 using Sparrow;
 using Sparrow.Binary;
 using Sparrow.Json;
 using Sparrow.Threading;
+using Sparrow.Utils;
 using Voron;
 using Voron.Data;
 using Voron.Data.Tables;
@@ -31,6 +31,7 @@ namespace Raven.Server.Rachis
         FailedToConnect,
         Disconnected,
         Closed,
+        Error,
     }
 
     public class FollowerAmbassador : IDisposable
@@ -153,6 +154,7 @@ namespace Raven.Server.Rachis
                         _engine.Log.Info($"{ToString()} was unable to set the thread priority, will continue with the same priority", e);
                     }
                 }
+
                 var hadConnectionFailure = false;
                 var needNewConnection = _connection == null;
                 while (_leader.Running && _running)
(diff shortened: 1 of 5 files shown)


src/Raven.Server/Rachis/FollowerAmbassador.cs
```diff
@@ -317,11 +319,20 @@ private unsafe void Run()
...
 
                                 if (aer.CurrentTerm != _term)
-                                    ThrowInvalidTermChanged(aer);
+                                {
+                                    var msg = $"The current engine term has changed " +
+                                              $"({aer.CurrentTerm:#,#;;0} -> {_term:#,#;;0}), " +
+                                              $"this ambassador term is no longer valid";
+                                    if (_engine.Log.IsInfoEnabled)
+                                    {
+                                        _engine.Log.Info($"{ToString()}: failure to append entries to {_tag} because: " + msg);
```
ayende, line 332: Can do string interpolation there all the way

src/Raven.Server/Rachis/FollowerAmbassador.cs
```diff
@@ -347,15 +358,22 @@ private unsafe void Run()
                             _debugRecorder.Start();
                         }
                     }
+                    catch (LockAlreadyDisposedException)
```
ayende, line 361: Better to unify all these throwing exception blocks and add a comment on why you are "handling" it in this way

src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
```diff
@@ -88,6 +88,12 @@ class ClusterObserver : IDisposable
                     if (Suspended == false)
                     {
                         _iteration++;
+                        if (_iteration == 1)
+                        {
+                            prevStats = _maintenance.GetStats();
```
ayende, line 93: Why aren't you doing that before the `while`?

src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
```diff
@@ -248,8 +271,27 @@ private void RaiseNoLivingNodesAlert(string alertMsg, string dbName)
...
+                    if (previous.TryGetValue(member, out _))
+                    {
+                        // if we found this node in the previous report, we will ingore it this time and wait for the next report.
+                        continue;
+                    }
+
+                    var msg =
+                        $"The member node {member} was not found in both current and previuos reports of the cluster observer. " +
+                        $"If this error continue to raise, check the latancy between the cluster nodes.";
+                    if (_logger.IsInfoEnabled)
```
ayende, line 286: What about if there is an error? And how can we avoid spamming the log about this?

ayende (approved):

src/Raven.Server/Rachis/FollowerAmbassador.cs
karmeli87: done

src/Raven.Server/Rachis/FollowerAmbassador.cs
karmeli87: done

src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
karmeli87: done. 
Thought to have it inside the try/catch but `GetStats()` can't really throw so moved it before the `while`.

src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
karmeli87: Since we generate an empty report for a node that haven't sent any info yet.
So this actually should be a very rare case when there is a race between adding the nodes to `ClusterMaintenanceSupervisor` and analyzing the stats in the observer.

PR #6549, opened 2018-06-10
[4.1] RavenDB-11296 Track Counters in session API

aviv86:
CancellationTokens will be handled as a part of RavenDB-11326
http://issues.hibernatingrhinos.com/issue/RavenDB-11326

Still need to support includes with counters.


src/Raven.Client/Documents/Session/AsyncSessionDocumentCounters.cs
```diff
@@ -0,0 +1,147 @@
...
+    {
+        public AsyncSessionDocumentCounters(InMemoryDocumentSessionOperations session, string documentId) : base(session, documentId)
+        {
+        }
+
+        public AsyncSessionDocumentCounters(InMemoryDocumentSessionOperations session, object entity) : base(session, entity)
+        {
+        }
+
+        public async Task<Dictionary<string, long>> GetAllAsync()
```
ayende: CancellationToken here? 
Do we have CT for all our methods?

src/Raven.Client/Documents/Session/AsyncSessionDocumentCounters.cs
```diff
@@ -0,0 +1,147 @@
...
+            else
+            {
+                cache = new InMemoryDocumentSessionOperations.CountersCache
+                {
+                    Values = new Dictionary<string, long?>()
+                };
+            }
+
+            cache.GotAll = true;
+            Session.CountersByDocId[DocId] = cache;
```
ayende: This should be add _after_ the call, because an exception may be thrown here.

src/Raven.Client/Documents/Session/AsyncSessionDocumentCounters.cs
```diff
@@ -0,0 +1,147 @@
...
+                }
+
+            }
+            else
+            {
+                cache = new InMemoryDocumentSessionOperations.CountersCache
+                {
+                    Values = new Dictionary<string, long?>()
+                };
+                Session.CountersByDocId.Add(DocId, cache);
```
ayende: Only it add if after the call

src/Raven.Client/Documents/Session/AsyncSessionDocumentCounters.cs
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        public async Task<Dictionary<string, long?>> GetAsync(IEnumerable<string> counters)
+        {
+            var countersToGetFromServer = new List<string>();
+            var result = new Dictionary<string, long?>();
+
+            if (Session.CountersByDocId.TryGetValue(DocId, out var cache))
+            {
+                if (cache.GotAll)
```
ayende: Merge the `GotAll` with the other call.
Then call to the server is didn't call `GotAll`

src/Raven.Client/Documents/Session/DocumentSession.CountersFor.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// Implements Unit of Work for accessing the RavenDB server
+    /// </summary>
+    public partial class DocumentSession
+    {
+        public SessionDocumentCounters CountersFor(string documentId)
+        {
+            return new SessionDocumentCounters(this, documentId);
+        }
+
+        public SessionDocumentCounters CountersFor(object entity)
```
ayende: Is there a  reason we aren't resolving this to the `documentId` directly here?

src/Raven.Client/Documents/Session/ICountersSessionOperationsAsync.cs
```diff
@@ -18,50 +17,18 @@ public interface ICountersSessionOperationsAsync : ICountersSessionOperationsBas
...
-        Task<Dictionary<string, long>> GetAsync(string documentId, params string[] counters);
-
-        /// <summary>
-        /// Returns CountersDetail on all the specified counters
-        /// <param name="entity">instance of entity of the document which holds the counter</param>
+        /// Returns the a dictionary of counter values by counter names
         /// <param name="counters">counters names</param>
         /// </summary>
-        Task<Dictionary<string, long>> GetAsync(object entity, params string[] counters);
+        Task<Dictionary<string, long?>> GetAsync(IEnumerable<string> counters);
```
ayende: I live how the interface shrank

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -120,6 +120,15 @@ public abstract partial class InMemoryDocumentSessionOperations : IDisposable
...
             new Dictionary<object, DocumentInfo>(ObjectReferenceEqualityComparer<object>.Default);
 
+        /// <summary>
+        /// hold the data required to manage Counters tracking for RavenDB's Unit of Work
+        /// </summary>
+        protected internal readonly Dictionary<string, CountersCache> CountersByDocId =
+            new Dictionary<string, CountersCache> (StringComparer.OrdinalIgnoreCase);
+
+        protected internal readonly HashSet<(string docId, string counter)> CountersToDelete = new HashSet<(string docId, string Name)>();
+        protected internal readonly HashSet<(string docId, string counter)> CountersToIncrement = new HashSet<(string docId, string Name)>();
```
ayende: What is this used for, and where is the actual amount to increment?

ayende: Can we lazily allocate all of these? In many cases, they will not be used, and it will be extra garbage.

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -749,6 +760,9 @@ internal SaveChangesData PrepareForSaveChanges()
 
             PrepareCompareExchangeEntities(result);
 
+            RemoveCountersFromCache(CountersToIncrement);
```
ayende: Why are you removing it? The result of the Inc operation will already tell you what the new value is, no?

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -820,6 +834,24 @@ private void PrepareCompareExchangeEntities(SaveChangesData result)
...
 
+        private void RemoveCountersFromCache(HashSet<(string docId, string counter)> countersToRemove)
+        {
+            foreach (var (docId, counter) in countersToRemove)
+            {
+                if (counter == null)
+                {
+                    // remove all
+                    CountersByDocId.Remove(docId);
+                    return;
```
ayende: I think you meant `continue;` here?

src/Raven.Client/Documents/Session/SessionCountersBase.cs
```diff
@@ -13,16 +13,31 @@ namespace Raven.Client.Documents.Session
...
             if (string.IsNullOrWhiteSpace(documentId))
                 throw new ArgumentNullException(nameof(documentId));
+            DocId = documentId;
+        }
+
+        protected SessionCountersBase(InMemoryDocumentSessionOperations session, object entity)
+        {
+            Session = session;
+
+            if (Session.DocumentsByEntity.TryGetValue(entity, out DocumentInfo document) == false)
```
ayende: Okay, so this is resolved once and we aren't holding on to it. Make sense.
However, I wonder if we can take advantage of this to skip the `GotAll` note. If we know that we have the metadata and are missing counters, we can check with the metadata in the session to know if we should get it.

ayende (changes requested):

src/Raven.Client/Documents/Session/AsyncDocumentSession.cs
```diff
@@ -101,9 +101,6 @@ protected override Task<string> GenerateIdAsync(object entity)
         public IRevisionsSessionOperationsAsync Revisions => _revisions ?? (_revisions = new DocumentSessionRevisionsAsync(this));
         private IRevisionsSessionOperationsAsync _revisions;
 
-        public ICountersSessionOperationsAsync Counters => _counters ?? (_counters = new DocumentSessionCountersAsync(this));
```
ppekrol, line 104: We do not want to have counters in .Advanced? It is a bit inconsistent with the rest of the session API.

src/Raven.Client/Documents/Session/AsyncDocumentSession.CountersFor.cs
```diff
@@ -0,0 +1,25 @@
...
+//-----------------------------------------------------------------------
+
+namespace Raven.Client.Documents.Session
+{
+    /// <summary>
+    /// Implements Unit of Work for accessing the RavenDB server
+    /// </summary>
+    public partial class AsyncDocumentSession
+    {
+        public AsyncSessionDocumentCounters CountersFor(string documentId)
```
ppekrol: Just in case I would go with Interface here.

src/Raven.Client/Documents/Session/AsyncSessionDocumentCounters.cs
ppekrol: We don't. Wrote a Roslyn analyzer to check that.

src/Raven.Client/Documents/Session/AsyncDocumentSession.cs
ayende: Yes, because it got really verbose and awkward when we started using it.
We also want to make this a 1st class item.
We'll probably also add `AttachmentsFor` to make this more discoverable

src/Raven.Client/Documents/Session/AsyncSessionDocumentCounters.cs
ppekrol: http://issues.hibernatingrhinos.com/issue/RavenDB-11326

src/Raven.Client/Documents/Session/AsyncDocumentSession.CountersFor.cs
aviv86: we have one already https://github.com/ravendb/ravendb/pull/6549/files#diff-eefe503b561793713fc89e176ec41b71

src/Raven.Client/Documents/Session/SessionCountersBase.cs
aviv86: done

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
aviv86: i did :)

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
aviv86: fixed, now we use the inc op result to update counters cache

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
aviv86: removed both

src/Raven.Client/Documents/Session/AsyncSessionDocumentCounters.cs
aviv86: fixed

src/Raven.Client/Documents/Session/AsyncSessionDocumentCounters.cs
aviv86: fixed

src/Raven.Client/Documents/Session/AsyncSessionDocumentCounters.cs
aviv86: fixed

src/Raven.Client/Documents/Session/AsyncDocumentSession.CountersFor.cs
ppekrol: I mean as a return type of this method.

aviv86: done

src/Raven.Client/Documents/Commands/Batches/CountersBatchCommandData.cs
```diff
@@ -30,6 +30,7 @@ public CountersBatchCommandData(string documentId, CounterOperation counterOpera
         public string Id { get; }
         public string Name { get; }
         public string ChangeVector { get; }
+        public bool HasIncrementOperation { get; set; }
```
ayende: Why do we need this property when we have `HasIncrement`?

src/Raven.Client/Documents/Linq/LinqPathProvider.cs
```diff
@@ -468,7 +463,7 @@ public static string RemoveTransparentIdentifiersIfNeeded(string path)
         public static bool IsCounterCall(MethodCallExpression mce)
         {
             return mce.Method.DeclaringType == typeof(RavenQuery) && mce.Method.Name == "Counter"
-                   || mce.Object?.Type == typeof(ICountersSessionOperations) && mce.Method.Name == "Get";
+                   || mce.Object?.Type == typeof(SessionDocumentCounters) && mce.Method.Name == "Get";
```
ayende: Shouldn't this be the interface here?

src/Raven.Client/Documents/Session/AsyncSessionDocumentCounters.cs
```diff
@@ -0,0 +1,153 @@
...
+            else
+            {
+                cache = new InMemoryDocumentSessionOperations.CountersCache();
+            }
+
+            Session.IncrementRequestCount();
+
+            Dictionary<string, long> values;
+
+            if (cache.MissingCounters != null)
```
ayende: What is this for?
An in general, if we are already going to the server to get all, we might as well get them all and simplify the code.

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -120,6 +120,12 @@ public abstract partial class InMemoryDocumentSessionOperations : IDisposable
         protected internal readonly Dictionary<object, DocumentInfo> DocumentsByEntity =
             new Dictionary<object, DocumentInfo>(ObjectReferenceEqualityComparer<object>.Default);
 
+        /// <summary>
+        /// hold the data required to manage Counters tracking for RavenDB's Unit of Work
+        /// </summary>
+        protected internal readonly Dictionary<string, CountersCache> CountersByDocId =
```
ayende: Use lazy creation for this, so we won't create a dictionary we won't need most times

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -249,17 +255,17 @@ public void SetTransactionMode(TransactionMode mode)
...
             var documentInfo = GetDocumentInfo(instance);
 
-            if (documentInfo.Metadata.TryGet(Constants.Documents.Metadata.Counters, out BlittableJsonReaderArray counters))
-                return counters.Select(x=>x.ToString()).ToList();
-
-            return null;
+            if (documentInfo.Metadata.TryGet(Constants.Documents.Metadata.Counters, 
+                out BlittableJsonReaderArray counters) == false)
+                return null;
+            return new HashSet<string>(counters.Select(x=>x.ToString()), StringComparer.OrdinalIgnoreCase);
```
ayende: Is there a scenario where we _can_ have duplicate counter names?
And are counter names really case insensitive?

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -820,6 +830,22 @@ private void PrepareCompareExchangeEntities(SaveChangesData result)
             clusterTransactionOperations.Clear();
         }
 
+        private void PrepareForCountersIncrement(SaveChangesData result)
+        {
+            foreach (var (id, type, name) in result.DeferredCommandsDictionary.Keys)
```
ayende: Just iterate over the dictionary directly, not on the `Keys`, you can also then directly check the type of the value and not check the type.

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -1474,5 +1508,112 @@ public void OnBeforeQueryInvoke(BeforeQueryEventArgs beforeQueryEventArgs)
...
+                {
+                    _gotAll = value;
+
+                    if (value)
+                        _missingCounters = null;
+                }
+            }
+
+            private bool _gotAll { get; set; }
+            private HashSet<string> _missingCounters;
```
ayende: What is the purpose of this?
Why not just put `Values[counter] = null` ?

src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
```diff
@@ -102,6 +112,24 @@ public void SetResult(BatchCommandResult result)
...
                 _session.OnAfterSaveChangesInvoke(afterSaveChangesEventArgs);
             }
+
+        }
+
+        private void UpdateCounterValuesInCache(BlittableJsonReaderArray counters, string docId)
+        {
+            if (_session.CountersByDocId.TryGetValue(docId, out InMemoryDocumentSessionOperations.CountersCache cache) == false)
+            {
+                cache = new InMemoryDocumentSessionOperations.CountersCache();
```
ayende: You didn't _add_ it back to the `CountersByDocId` ?

src/Raven.Client/Documents/Session/Operations/LoadOperation.cs
```diff
@@ -115,6 +115,13 @@ public void SetResult(GetDocumentsResult result)
 
                 var newDocumentInfo = DocumentInfo.GetNewDocumentInfo(document);
                 _session.DocumentsById.Add(newDocumentInfo);
+
+                if (_session.CountersByDocId.TryGetValue(newDocumentInfo.Id, out var cache) == false)
+                    continue;
+
+                newDocumentInfo.Metadata.TryGet(Constants.Documents.Metadata.Counters, out BlittableJsonReaderArray bjra);
+                var metadataCounters = new HashSet<string>(bjra.Select(c => c.ToString()), StringComparer.OrdinalIgnoreCase);
+                cache.RegistarMissingCounters(metadataCounters);
```
ayende: What is the purpose of this code?

src/Raven.Client/Documents/Session/SessionCountersBase.cs
```diff
@@ -13,16 +14,48 @@ namespace Raven.Client.Documents.Session
...
+
+            if (Session.CountersByDocId.TryGetValue(DocId, out var cache) == false)
+            {
+                cache = new InMemoryDocumentSessionOperations.CountersCache();
+            }
+
+            if (cache.GotAll == false && cache.MissingCounters == null)
+            {
+                var metadataCounters = Session.GetCountersFor(entity);
+                cache.RegistarMissingCounters(metadataCounters);
```
ayende: Why?

ayende (changes requested):

src/Raven.Client/Documents/Operations/CountersOperationExecutor.cs
```diff
@@ -81,14 +81,14 @@ public Task<CountersDetail> IncrementAsync(string docId, string name, long delta
             return details.Counters[0].TotalValue;
         }
 
-        public Dictionary<string, long> Get(string docId, IEnumerable<string> counters)
+        public Dictionary<string, long?> Get(string docId, IEnumerable<string> counters)
```
ppekrol: Do we need CountersOperationExecutor at all? All now is a part of the session and we do not have equivalents for the rest. If the answer is 'yes' then we are missing there `SessionInfo sessionInfo = null` argument in each of those methods.

ppekrol: Also all are OperationExecutors are based on IOperation or IServerOperation concept, but this one isnt.

ayende: The one thing that we should be concerned here is the fact that:
* We need a lower level than the session
* Counters have different load balancing than other stuff, we likely will want to configure them separately.

ppekrol: We can always use `store.Operations`.
If we want to have something different then I suggest renaming this to `DocumentStoreCounters` instead.

ayende: @aviv86 Currently we _don't_ have a different behavior for load balancing on counters, right?

aviv86: right, we don't have a different behavior for load balancing on counters yet

src/Raven.Client/Documents/Commands/Batches/CountersBatchCommandData.cs
aviv86: removed this property

src/Raven.Client/Documents/Linq/LinqPathProvider.cs
aviv86: fixed

src/Raven.Client/Documents/Session/AsyncSessionDocumentCounters.cs
aviv86: removed cache.MissingCounters,
if going to server will ask for all

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
aviv86: done

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
aviv86: reverted back to List

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
aviv86: removed this method

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
aviv86: removed _missingCounters

src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
aviv86: fixed

src/Raven.Client/Documents/Session/Operations/LoadOperation.cs
aviv86: removed

src/Raven.Client/Documents/Session/SessionCountersBase.cs
aviv86: removed cache.MissingCounters

src/Raven.Client/Documents/Operations/CountersOperationExecutor.cs
aviv86: removed store.Counters
will recreate this in 4.2 with new policy
http://issues.hibernatingrhinos.com/issue/RavenDB-11348

src/Raven.Client/Documents/Session/AsyncSessionDocumentCounters.cs
```diff
@@ -0,0 +1,170 @@
...
+                    out BlittableJsonReaderArray metadataCounters) &&
+                metadataCounters.BinarySearch(counter, StringComparison.OrdinalIgnoreCase) >= 0))
+                
+            {
+                // we either don't have the document in session and GotAll = false,
+                // or we do and it's metadata contains the counter name
+
+                Session.IncrementRequestCount();
+
+                var details = await Session.DocumentStore.Operations.ForDatabase(Session.DatabaseName)
```
ppekrol, line 51: I think that we have RequestExecutor and OperationExecutor in session already. Try using:
```
Session.Operations.SendAsync
```

Also remember to put `Session.SessionInfo` as a parameter to `SendAsync`.

src/Raven.Client/Documents/Session/AsyncSessionDocumentCounters.cs
```diff
@@ -0,0 +1,170 @@
...
+
+            if (missingCounters)
+            {
+                // we either don't have the document in session and GotAll = false,
+                // or we do and cache doesn't contain all metadata counters
+
+                Session.IncrementRequestCount();
+
+                var details = await Session.DocumentStore.Operations.ForDatabase(Session.DatabaseName)
+                    .SendAsync(new GetCountersOperation(DocId, new string[0]), token: token).ConfigureAwait(false);
```
ppekrol, line 155: I think it would be better from UX perspective to add a second constructor for this case.

Passing `null` or empty array should not be allowed. Check how it is done here: https://github.com/ravendb/ravendb/blob/43b7e45576d02df20df375eefc5b2da044bfb174/src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs#L30-L51

ppekrol (approved):

src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
```diff
@@ -102,6 +103,44 @@ public void SetResult(BatchCommandResult result)
                 var afterSaveChangesEventArgs = new AfterSaveChangesEventArgs(_session, documentInfo.Id, documentInfo.Entity);
                 _session.OnAfterSaveChangesInvoke(afterSaveChangesEventArgs);
             }
+
+            for (int i = _sessionCommandsCount; i < result.Results.Length; i++)
+            {
+                if (!(result.Results[i] is BlittableJsonReaderObject batchResult))
+                    throw new ArgumentNullException();
```
ayende, line 110: You don't dictate what goes here, just ignore what you don't understand.

src/Raven.Client/Documents/Session/AsyncSessionDocumentCounters.cs
ayende: done

src/Raven.Client/Documents/Session/AsyncSessionDocumentCounters.cs
ayende: done

PR #6550, opened 2018-06-11
[4.0] RavenDB-11065 fix PR comments

karmeli87:


src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
```diff
@@ -79,7 +79,12 @@ class ClusterObserver : IDisposable
 
         public async Task Run(CancellationToken token)
         {
-            var prevStats = new Dictionary<string, ClusterNodeStatusReport>();
+            // we give some time to populate the stats.
+            await TimeoutManager.WaitFor(SupervisorSamplePeriod, token);
+            var prevStats = _maintenance.GetStats();
```
ayende, line 84: _can_ this fail?

ayende (approved):

src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
karmeli87: This was initially my concern. But then I realized that this simply can't fail. :)

PR #6558, opened 2018-06-11
[4.0] RavenDB-11068 The Read method cannot be called when another read oper…

karmeli87:
…ation is pending.


src/Raven.Server/Rachis/CandidateAmbassador.cs
```diff
@@ -166,12 +168,16 @@ private void Run()
                             while (_candidate.Running)
                             {
                                 RequestVoteResponse rvr;
+                                
+                                // clear the current published connection
+                                Interlocked.Exchange(ref PublishedConnection, null)?.Dispose();
```
ayende: In this case, this is the same as `_connection`, no?

ayende:
We need a different approach, talked with @karmeli87 about this already, will have new PR

PR #6564, opened 2018-06-11
RavenDB-11329 Allow to call WaitForUserToContinueTheTest more than on…

ayende:
…ce per test


talweiss1982:
I have suggested this a year ago.

PR #6569, opened 2018-06-11
RavenDB-10034 SQL ETL - Support additional databases 

arekpalinski:


ppekrol (approved):

ppekrol:
I'm wondering at what other DBs we should look at?

src/Raven.Client/Documents/Operations/ETL/SQL/SqlConnectionStringParser.cs
```diff
@@ -24,11 +23,70 @@ internal static (string Database, string Server) GetDatabaseAndServerFromConnect
...
 
-                    var port = GetConnectionStringValue(connectionString, new[] { "Port" });
+                    var postgrePort = GetConnectionStringValue(connectionString, new[] { "Port" });
 
-                    if (string.IsNullOrEmpty(port))
-                        server += $":{port}";
+                    if (string.IsNullOrEmpty(postgrePort))
+                        server += $":{postgrePort}";
                     break;
+                case SqlProvider.MySqlClient:
```
ayende, line 31: Do we really need to parse the connection strings? Why not pass them as is to the `AdoConnection`?

ayende:
I think that this close the loop in terms of what we support until  we get a user request.
I don't think we want to pre-preemptively support DB2, for example.

ayende:
I left my mainframe on my other pants

src/Raven.Client/Documents/Operations/ETL/SQL/SqlConnectionStringParser.cs
arekpalinski: This class (and its friends that use it) are in Raven.Client where we don't have ADO dependencies. The values that we parse here are displayed in the studio (ongoing task view) and go into the logs (`dbname@server`).  When we do the actual connection then we pass the entire connection string.

PR #6570, opened 2018-06-11
[v4.1] RavenDB-8322: Introduced query cache debug endpoint

maximburyak:


src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
```diff
@@ -93,5 +95,112 @@ public Task RunningQueries()
...
+        {
+            using (ContextPool.AllocateOperationContext(out DocumentsOperationContext context))
+            using (var writer = new BlittableJsonTextWriter(context, ResponseBodyStream()))
+            {
+                var queryCache = Database.QueryMetadataCache.GetQueryCache();
+
+                var responseDjv = new DynamicJsonValue();
+                var queriesList = new DynamicJsonArray();
+
+                responseDjv["CachedQueryMetadataItems"] = queriesList;
```
ppekrol: We are using "Results" most often. Let's stay consistent.

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
```diff
@@ -93,5 +95,112 @@ public Task RunningQueries()
...
+                foreach (var item in queryCache)
+                {   
+                    if (item != null)
+                    {
+                        var curDjvItem = new DynamicJsonValue();
+                        queriesList.Add(curDjvItem);
+
+                        if (item.IsGroupBy)
+                        {
+                            curDjvItem["IsGroupBy"] = true;
```
ppekrol: Let's add .ToJson() method inside QueryMetadata and handle that there. Remember to use `nameof`.

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
```diff
@@ -93,5 +95,112 @@ public Task RunningQueries()
...
+                        if (string.IsNullOrEmpty(item.IndexName) == false)
+                        {
+                            curDjvItem["IndexName"] = item.IndexName;
+                        }                        
+                    
+                        curDjvItem["QueryText"] = item.QueryText;                        
+                    }
+                }
+
+                var blittableResult = context.ReadObject(responseDjv, "debug/queries/cache/list");
```
ppekrol: IMO we are missing some stuff:
1. We should add CreatedAt
2. We should add LastQueriedAt
3. We need to add total number of queries 
4. We need to add current number of queries

ppekrol (changes requested):

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
ppekrol: IIRC writer has a method for that `WriteArray`

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
maximburyak: I'm not sure that ToJson is appropriate here, because I don't think that we need "all" the cache properties, we want to minimize it to the most important info to the user, that's why I don't "always" write the properties, only when it's interesting

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
maximburyak: I agree about 3,4 . I've discussed with @ayende similiar issue question to 1,2 yesterday and I'm not sure how relevant this will be for the ever-changing nature of the cache (512 items etc)

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
ppekrol: You can apply same logic there

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
ayende: What is 3 & 4? I get total number of queries, but what is current number of queries?
1) Is easy, 2) will be best effort.

ppekrol: I wouldn't say ever changing, I know that they can overlap but in decent amount of cases cache will (more or less) stabilize after few minutes IMO.

ppekrol: @ayende 
Total is = 'how many times we took this Query from cache' 
and 
Current is = 'how many queries for that particular Query are done at the moment'

ppekrol: This could help us to know which ones are actively used.

ayende: Current probably means `Interlocked.Increment` all over the place. 
And will likely create hot spot for concurrency.

maximburyak: @ppekrol This will be confusing. You don't want to use the cache as metrics, because queries go in and out, so the data will not be complete

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
maximburyak: Then ToJson won't be the ToJson that we know as the function for full serialization. I think that the convention in QueriesDebugHandler right now is to create the serialization "by hand", I think that we better be consistent

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
ppekrol: @ayende I'm not sure if Interlocked.Increment would be a big problem here? But since current is not a big deal here, then we can drop it. But Total would be good (even if we know that this might be dropped from cache).

@maximburyak as I said, I think that most of the time the cache will stabilize. I would not use this as a metric. This gives you cache state and traffic.

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
ppekrol: Use `nameof` then

maximburyak: Yes, it is important

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
maximburyak: @ayende?

ayende: Go ahead, yes. I don't think we'll see a lot of churn in the cache, and if we do, that is a valuable insight all on its own

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
maximburyak: nice

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
```diff
@@ -93,5 +96,143 @@ public Task RunningQueries()
 
             return Task.CompletedTask;
         }
+
+        private void WritePropertyNameAndScalarValue(BlittableJsonTextWriter writer, string propertyName,object propertyValue)
```
ppekrol: It is not used. Can be deleted?

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
```diff
@@ -93,5 +96,143 @@ public Task RunningQueries()
...
+            using (ContextPool.AllocateOperationContext(out DocumentsOperationContext context))
+            using (var writer = new BlittableJsonTextWriter(context, ResponseBodyStream()))
+            {
+                var queryCache = Database.QueryMetadataCache.GetQueryCache();
+
+                var responseDjv = new DynamicJsonValue();
+                var queriesList = new List<DynamicJsonValue>();
+
+                writer.WritePropertyName("TotalCachedQueries");
+                writer.WriteValue(BlittableJsonToken.Integer,(long)queryCache.Length);
```
ppekrol: I think we have a `writer.WriteNumber` or `write.WriteInteger` for this. No need to do this like that.

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
```diff
@@ -93,5 +96,143 @@ public Task RunningQueries()
...
+                writer.WriteComma();
+                                               
+                foreach (var item in queryCache)
+                {
+                    List<(string propertyName, object propertyValue)> properties = new List<(string propertyName, object propertyValue)>();
```
ppekrol: This is not used?

ppekrol (changes requested):

ppekrol:
Beside cosmetics. LGTM

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
maximburyak: yeah, correct, it's write integer

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
maximburyak: sure

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
maximburyak: yes

PR #6599, opened 2018-06-13
[4.1] RavenDB-11326 go over client public API and check if we are not missing cancellation tokens

ppekrol:


src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1077,7 +1075,7 @@ private static Task HandleConflict(JsonOperationContext context, HttpResponseMes
 
             OnFailedRequest(url, e);
 
-            await ExecuteAsync(currentNode, currentIndex, context, command, sessionInfo: sessionInfo).ConfigureAwait(false);
+            await ExecuteAsync(currentNode, currentIndex, context, command, shouldRetry: true, sessionInfo: sessionInfo, token: token).ConfigureAwait(false);
```
arekpalinski: minor: maybe a bit safer would be to use `shouldRetry: default`

src/Raven.Server/Documents/Handlers/BatchRequestParser.cs
```diff
@@ -741,7 +741,7 @@ private static void ThrowUnexpectedToken(JsonParserToken jsonParserToken, JsonPa
         {
             // Although we using here WithCancellation and passing the token,
             // the stream will stay open even after the cancellation until the entire server will be disposed.
-            var read = await stream.ReadAsync(buffer.Buffer.Array, buffer.Buffer.Offset, buffer.Buffer.Count, token).WithCancellation(token);
+            var read = await stream.ReadAsync(buffer.Buffer.Array, buffer.Buffer.Offset, buffer.Buffer.Count, token);
```
arekpalinski: the above comment is no longer valid - we don't use WithCancellation anymore. why did we need to add it here as we pass `token` anyway

src/Raven.Client/Documents/DocumentStore.cs
```diff
@@ -405,10 +406,10 @@ public override CountersOperationExecutor Counters
             }
         }
 
-        public override BulkInsertOperation BulkInsert(string database = null)
+        public override BulkInsertOperation BulkInsert(string database = null, CancellationToken token = default)
```
ayende, line 409: This is a breaking binary change, so only for 4.1, maintaining source compatability, right?

ppekrol: Yes, I planned not do do a 4.0 PR at all for this issue.

PR #6612, opened 2018-06-13
RavenDB-10923: Linux Container under Windows Host with cifs Mount

aviviadi:


src/Voron/Platform/Posix/PosixMemoryMapPager.cs
```diff
@@ -35,7 +35,24 @@ public sealed unsafe class PosixMemoryMapPager : PosixAbstractPager
             if (_fd == -1)
             {
                 var err = Marshal.GetLastWin32Error();
-                Syscall.ThrowLastError(err, "when opening " + file);
+                if ((Errno)err == Errno.ENOENT)
+                {
+                    // unique case, when file was previously deleted, but still exists. This can happen on cifs mount, see RavenDB-10923
+                    // if this is a temp file we can try recreate it in a different name
+                    if (Path.GetExtension(file.FullPath ?? "").Equals("buffers"))
```
arekpalinski, line 42: why `?? ""` usage here  ? I don't hink `file.FullPath` can return null

PR #6614, opened 2018-06-13
[4.0] RavenDB-11354 High CPU after hitting leader stepdown several times

karmeli87:
The issue is related to code for sending topology change notification from the leader.
We waited to leave 'Follower' state in order to reconnect to a new leader, but it could happened that the leader was changed without leaving the follower state, and so we were connected to the wrong node.
This could caused a circular connection between the nodes which sent the same notification again and again.


src/Raven.Server/NotificationCenter/Notifications/Server/ClusterTopologyChanged.cs
```diff
@@ -53,7 +53,8 @@ public override DynamicJsonValue ToJson()
                 NodeTag = nodeTag,
                 Status = status,
                 CurrentTerm = term,
-                NodeLicenseDetails = nodeLicenseDetails
+                NodeLicenseDetails = nodeLicenseDetails,
+                IsPersistent = true
```
ppekrol, line 57: Why do we want to persist this?

src/Raven.Server/NotificationCenter/NotificationsBase.cs
```diff
@@ -17,18 +17,30 @@ public abstract class NotificationsBase : IDisposable
         private TaskCompletionSource<object> _newWebSocket = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
         private TaskCompletionSource<object> _allWebSocketsRemoved = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
 
-        public Task WaitForAllRemoved => _allWebSocketsRemoved.Task;
+        public Task WaitForRemoveAllWebSocketClients
+        {
+            get
+            {
+                lock (_watchersLock)
```
ayende: Can we avoid this lock?

src/Raven.Server/NotificationCenter/Notifications/Server/ClusterTopologyChanged.cs
karmeli87: 1. We need to update the topology after reelection even if some nodes are connected after the initial message was sent.
2. Anyway we keep only the recent topology. (the Id defined by the Type in this case)

src/Raven.Server/NotificationCenter/NotificationsBase.cs
```diff
@@ -17,18 +17,30 @@ public abstract class NotificationsBase : IDisposable
...
+        private volatile int _websocketClients;
 
-        public Task WaitForNew()
+        public Task WaitForAnyWebSocketClient
         {
-            lock (_watchersLock)
+            get
             {
-                return _websocketClients > 0 ? Task.CompletedTask : _newWebSocket.Task;
+                lock (_watchersLock)
```
ayende: can we avoid this lock?

ayende: We can use `Exchange` to do this atomically, instead

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -255,7 +255,7 @@ private set
 
         public event EventHandler LeaderElected;
 
-        private string _tag;
+        private volatile string _tag;
```
ayende: Under what scenario is this desirable?

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -484,6 +484,21 @@ private void SwitchToSingleLeader(TransactionOperationContext context)
...
 
+        public async Task WaitForLeaderChange(CancellationToken cts)
+        {
+            var currentLeader = LeaderTag;
+            while (cts.IsCancellationRequested == false)
+            {
+                // we setup the wait _before_ checking the state
+                var task = _stateChanged.Task;
+
+                if (currentLeader != LeaderTag)
```
ayende: Better to have the hit here, using `Volatile.Read`, than everywhere.

src/Raven.Server/NotificationCenter/NotificationsBase.cs
```diff
@@ -69,7 +68,7 @@ public IDisposable TrackActions(AsyncQueue<DynamicJsonValue> notificationsQueue,
 
                 if (watcher.Writer is NotificationCenterWebSocketWriter)
                 {
-                    if (_websocketClients == 0)
+                    if (Interlocked.CompareExchange(ref _haveClients, 1, 0) == 0)
```
ayende: You have the wrong order here

ayende:
See how this behaves https://github.com/ravendb/ravendb/pull/6614/commits/c87cbc54d56672ae7d75c9b3626edf20f6a77d61

PR #6617, opened 2018-06-13
RavenDB-10923: Linux Container under Windows Host with cifs Mount bec…

ayende:
…ause of file system caching issues

- Instead of reading the transaction data from the journal file, we'll keep it in our own memory and use that for sync purposes.
  This allows us to avoid stale read after write when using CIFS system (such as Docker using Linux container on Windows with external volume).
  There is no impact to the safety guarantees since we not relying on the sync validating the status of the journal file and we are known to be okay WRT the state of writes ot the journal.
- Will properly handle FileNotFound errors when we create scratch files

### DIFF
--- a/src/Sparrow/Platform/Posix/Syscall.cs
+++ b/src/Sparrow/Platform/Posix/Syscall.cs
@@ -304,6 +304,8 @@ namespace Sparrow.Platform.Posix
             {
                 case Errno.ENOMEM:
                     throw new OutOfMemoryException("ENOMEM on " + msg);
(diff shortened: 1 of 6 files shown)


ppekrol:
We missed something here, because this is generating a lot of tests failures.

ayende:
Yes, checking now

ayende:
FastTest run fine now. Now running full test suite

ayende:
Slow tests run as well

aviviadi (approved):

PR #6623, opened 2018-06-14
Raven db 11361 4.0

ayende:


src/Raven.Server/Documents/Indexes/Static/IndexCompiler.cs
```diff
@@ -32,6 +32,47 @@ public static class IndexCompiler
...
+        private static Lazy<(string Path, AssemblyName AssemblyName, MetadataReference Reference)[]> KnownManagedDlls = new Lazy<(string,AssemblyName,MetadataReference)[]>(DiscoverManagedDlls);
+
+        static IndexCompiler()
+        {
+            AssemblyLoadContext.Default.Resolving += (AssemblyLoadContext ctx, AssemblyName name) =>
+            {
+                if (KnownManagedDlls.IsValueCreated == false)
+                    return null; // this also handles the case of failure
+
+                foreach (var item in KnownManagedDlls.Value)
```
ryanheath, line 44: Why not using a dictionary here? There might be quite a few dlls to loop through.

ayende: The expected number of elements is low (single digits, most often).
The number of calls is similarly low (once per assembly resolution).
Not worth it

ryanheath: Huh, are you not looping through the directory of raven.server.exe? On my machine that directory is full of dlls.

ryanheath: But number of calls to load a dll would be low, that's true.

ppekrol (approved):

src/Raven.Server/Documents/Indexes/Static/IndexCompiler.cs
ayende: @ryanheath That depends on what the deployment model is, yes. In particular, in hosted setup, we'll have a lot of dlls, but when we use the system `dotnet` version, we have very few.

PR #6636, opened 2018-06-14
RavenDB-10923: Rename CreateTemporaryBufferPager if cannot be created…

aviviadi:
… due to ENOENT although exists

### DIFF
--- a/src/Voron/StorageEnvironmentOptions.cs
+++ b/src/Voron/StorageEnvironmentOptions.cs
@@ -632,9 +632,9 @@ namespace Voron
                     File.Delete(file);
             }
 
-            public override AbstractPager CreateScratchPager(string name, long initialSize)
+            private AbstractPager GetMemoryMapPagerWithPossibleFileRename(string name, long initialSize, Func<StorageEnvironmentOptions, long?, VoronPathSetting, bool, bool, AbstractPager> getMemoryMapPagerFunc)
             {
-                // here we can afford to rename the file if needed because this is a scratch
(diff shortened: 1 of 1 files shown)


src/Voron/StorageEnvironmentOptions.cs
```diff
@@ -675,18 +676,19 @@ void Rename()
...
             // This is used for special pagers that are used as temp buffers and don't 
             // require encryption: compression, recovery, lazyTxBuffer.
             public override AbstractPager CreateTemporaryBufferPager(string name, long initialSize)
             {
-                var scratchFile = TempPath.Combine(name);
-                if (File.Exists(scratchFile.FullPath))
-                    File.Delete(scratchFile.FullPath);
-
-                var pager = GetMemoryMapPagerInternal(this, initialSize, scratchFile, deleteOnClose: true);
+                var pager = GetMemoryMapPagerWithPossibleFileRename(name, initialSize, GetMemoryMapPagerInternal);
```
ayende, line 691: Given that they are so similar, is there a reason not to just call CreateScratchPager here and avoid `GetMemoryMapPagerWithPossibleFileRename` ?

ayende: going to implement this in this manner.

aviviadi: https://github.com/ravendb/ravendb/pull/6648

ayende: Wasn't aware of that, good point

PR #6647, opened 2018-06-15
[4.1] RavenDB-10038

ppekrol:


src/Raven.Server/Documents/Handlers/StatsHandler.cs
```diff
@@ -41,6 +41,14 @@ public Task Stats()
                 var attachments = Database.DocumentsStorage.AttachmentsStorage.GetNumberOfAttachments(context);
                 stats.CountOfAttachments = attachments.AttachmentCount;
                 stats.CountOfUniqueAttachments = attachments.StreamsCount;
+
+                using (ServerStore.ContextPool.AllocateOperationContext(out TransactionOperationContext serverContext))
+                using (serverContext.OpenReadTransaction())
+                {
+                    stats.CountOfIdentities = ServerStore.Cluster.GetNumberOfIdentities(serverContext, Database.Name);
```
ppekrol, line 48: @ayende I'm wondering if we should not move it from DatabaseStatistics to:
a) separate endpoint per each of those
b) to endpoint like /databases/stats/detailed

ayende: I think `/detailed` is better yes.
In 3.5 we had a stats endpoint just for the studio to hit, because it `/stats` was getting expensive.

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1347,6 +1347,45 @@ public IEnumerable<(string Prefix, long Value)> ReadIdentities<T>(TransactionOpe
...
+
+        public long GetNumberOfCompareExchange(TransactionOperationContext context, string databaseName)
+        {
+            var items = context.Transaction.InnerTransaction.OpenTable(CompareExchangeSchema, CompareExchange);
+            var compareExchange = items.GetTree(CompareExchangeSchema.Key);
+            var prefix = CompareExchangeCommandBase.GetActualKey(databaseName, null);
+            return GetNumberOf(compareExchange, prefix, context);
+        }
+
+        private static long GetNumberOf(Tree tree, string prefix, TransactionOperationContext context)
```
ayende, line 1366: Probably better to implement this explicitly in the Tree as `GetNumberOfEntriesWithPrefix` 
Like `GetNumberOfEntriesAfter` in Fixed Size Tree.

ayende (approved):

PR #6663, opened 2018-06-18
RavenDB-8993 Expose index explain in the studio

ml054:


ml054:
Please review server part.

PR #6670, opened 2018-06-19
[4.1] RavenDB-11296 : session include counters

aviv86:


src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs
```diff
@@ -33,20 +35,22 @@ public GetDocumentsCommand(int start, int pageSize)
             _pageSize = pageSize;
         }
 
-        public GetDocumentsCommand(string id, string[] includes, bool metadataOnly)
+        public GetDocumentsCommand(string id, string[] includes, string[] counters, bool metadataOnly)
```
ppekrol: This brakes the source-level compatibility. You need to add a separate constructor with that.

src/Raven.Client/Constants.cs
```diff
@@ -103,6 +103,15 @@ private Configuration()
             public const string ClientId = "Configuration/Client";
         }
 
+        public class Counters
+        {
+            private Counters()
+            {
+            }
+
+            public const string All = "@allCounters";
```
ppekrol: Let's match the `@all_docs` and change this to `@all_counters`

src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs
ppekrol: Assert in this new ctor if this is not null.

src/Raven.Client/Documents/Session/DocumentSession.IncludeCounters.cs
```diff
@@ -0,0 +1,46 @@
...
+using Raven.Client.Documents.Session.Loaders;
+
+namespace Raven.Client.Documents.Session
+{
+    /// <summary>
+    /// Implements Unit of Work for accessing the RavenDB server
+    /// </summary>
+    public partial class DocumentSession
+    {
+        public ILoaderWithInclude<object> IncludeCounter(string name)
```
ppekrol: I do not think that we want those overloads without generic `T` here. Let's drop them.

src/Raven.Client/Documents/Session/Loaders/MultiLoaderWithInclude.cs
```diff
@@ -67,14 +68,52 @@ public ILoaderWithInclude<T> Include<TInclude>(Expression<Func<T, IEnumerable<st
             return Include(IncludesUtil.GetPrefixedIncludePath<TInclude>(path.ToPropertyPath(), _session.Conventions));
         }
 
+        /// <summary>
+        /// Includes a counter of the document to be loaded, by counter name
+        /// <param name="name">Name of the counter to include.</param>
+        /// </summary>
+        public ILoaderWithInclude<T> IncludeCounter(string name)
+        {
```
ppekrol: assert if name is not null or whitespace?

src/Raven.Client/Documents/Session/Loaders/MultiLoaderWithInclude.cs
```diff
@@ -67,14 +68,52 @@ public ILoaderWithInclude<T> Include<TInclude>(Expression<Func<T, IEnumerable<st
...
+        /// Includes counters of the document to be loaded, by counters names
+        /// <param name="names">Names of the counters to include.</param>
+        /// </summary>
+        public ILoaderWithInclude<T> IncludeCounters(string[] names)
+        {
+            if (_counters == null)
+                _counters = new List<string>();
+            if (names != null)
+            {
+                foreach (var name in names)
```
ppekrol: Let's assert if this is not null or whitespace.

src/Raven.Server/Json/BlittableJsonTextWriterExtensions.cs
```diff
@@ -1192,6 +1193,38 @@ public static void WriteObjects(this BlittableJsonTextWriter writer, JsonOperati
...
+                writer.WritePropertyName(nameof(CounterDetail.CounterName));
+                writer.WriteString(counter.CounterName);
+                writer.WriteComma();
+
+                writer.WritePropertyName(nameof(CounterDetail.TotalValue));
+                writer.WriteDouble(counter.TotalValue);
+
+                writer.WriteEndObject();
+
+                await writer.MaybeOuterFlsuhAsync();
```
ppekrol: typo :)

I do not think we need to flush that?

ppekrol (changes requested):

src/Raven.Server/Json/BlittableJsonTextWriterExtensions.cs
arekpalinski: BTW there is typo there: Flsuh -> Flush

src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs
aviv86: done

src/Raven.Client/Constants.cs
aviv86: done

src/Raven.Client/Documents/Session/DocumentSession.IncludeCounters.cs
aviv86: done

src/Raven.Client/Documents/Session/Loaders/MultiLoaderWithInclude.cs
aviv86: done

src/Raven.Client/Documents/Session/Loaders/MultiLoaderWithInclude.cs
aviv86: done

src/Raven.Server/Json/BlittableJsonTextWriterExtensions.cs
aviv86: fixed

src/Raven.Client/Constants.cs
```diff
@@ -103,6 +103,15 @@ private Configuration()
             public const string ClientId = "Configuration/Client";
         }
 
+        public class Counters
```
ayende: Use static class here, instead.

src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs
```diff
@@ -95,6 +103,18 @@ public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, Serve
                 }
             }
 
+            if (_counters != null)
+            {
+                if (_counters.Length == 0)
+                {
+                    pathBuilder.Append("&counter=").Append(Constants.Counters.All);
```
ayende: Why does sending empty array means all?

src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs
```diff
@@ -95,6 +103,18 @@ public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, Serve
...
+                }
+                else foreach (var counter in _counters)
```
ayende: Please use `{}` here

src/Raven.Client/Documents/Session/DocumentSession.IncludeCounters.cs
```diff
@@ -0,0 +1,31 @@
...
+        }
+
+        public ILoaderWithInclude<T> IncludeCounters<T>(string[] names)
+        {
+            return new MultiLoaderWithInclude<T>(this).IncludeCounters(names);
+        }
+
+        public ILoaderWithInclude<T> IncludeCounters<T>()
+        {
+            return new MultiLoaderWithInclude<T>(this).IncludeCounters(new string[0]);
```
ayende: Let's create an explicit value here, instead of using empty array.
Note that we might call `IncludeCounters` and then another call with specific value.

src/Raven.Client/Documents/Session/IDocumentSession.IncludeCounters.cs
```diff
@@ -0,0 +1,36 @@
...
+//-----------------------------------------------------------------------
+
+using Raven.Client.Documents.Session.Loaders;
+
+namespace Raven.Client.Documents.Session
+{
+    /// <summary>
+    ///     Interface for document session
+    /// </summary>
+    public partial interface IDocumentSession
```
ayende: This I don't like.
We already have a top level "taint" with `CountersFor`, I don't want to add `IncludeCounters` there at this level.
@ppekrol Any suggestions about the kind of API we should strive for?

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -1278,6 +1279,34 @@ public void RegisterMissingIncludes(BlittableJsonReaderArray results, BlittableJ
             }
         }
 
+        internal void RegisterCounters(BlittableJsonReaderArray counters)
+        {
+            if (counters == null)
+                return;
+
```
ayende: You need to account for counters that you _tried_ to load and wasn't able to.
Also, don't you need to account for `GotAll` too?

src/Raven.Client/Documents/Session/Loaders/MultiLoaderWithInclude.cs
```diff
@@ -67,14 +68,59 @@ public ILoaderWithInclude<T> Include<TInclude>(Expression<Func<T, IEnumerable<st
...
+
         /// <summary>
         /// Loads the specified ids.
         /// </summary>
         /// <param name="ids">The ids.</param>
         /// <returns></returns>
         public Dictionary<string, T> Load(params string[] ids)
         {
-            return _session.LoadInternal<T>(ids, _includes.ToArray());
+            return _session.LoadInternal<T>(ids, _includes.ToArray(), _counters?.ToArray());
```
ayende: Why another allocation here? Why not pass the list?

src/Raven.Client/Documents/Session/Loaders/MultiLoaderWithInclude.cs
```diff
@@ -67,14 +68,59 @@ public ILoaderWithInclude<T> Include<TInclude>(Expression<Func<T, IEnumerable<st
...
+            }
+            return this;
+        }
+
+        /// <summary>
+        /// Includes all the counters of the document to be loaded
+        /// </summary>
+        public ILoaderWithInclude<T> IncludeCounters()
+        {
+            return IncludeCounters(new string[0]);
```
ayende: don't like this, will fail in funny ways. `IncludeCounters` and then `IncludeCounter` will only load the specific counter we wanted.

test/SlowTests/Client/Counters/SessionCounters.cs
```diff
@@ -1034,5 +1034,247 @@ public void SessionShouldRemoveCountersFromCacheAfterDocumentDeletion()
...
+
+                    session.SaveChanges();
+                }
+
+                using (var session = store.OpenSession())
+                {
+                    var user = session
+                        .IncludeCounter<User>("likes")
+                        .Load("users/1-A");
+
```
ayende: What happens if I add `IncludeCounter("dances")` and then try to get it?
It should remember that it is missing

ayende (changes requested):

src/Raven.Client/Constants.cs
ppekrol: This is ok actually.

src/Raven.Client/Documents/Session/DocumentSession.IncludeCounters.cs
ppekrol: maybe `IncludeAllCounters`?

src/Raven.Client/Documents/Session/Loaders/MultiLoaderWithInclude.cs
ppekrol: `IncludeCounter(name)`
`IncludeCounters(names)`
`IncludeAllCounters`
?

src/Raven.Client/Documents/Session/Loaders/MultiLoaderWithInclude.cs
ppekrol: Probably rest of the API is done in the same way e.g. we are doing that with `_includes`

src/Raven.Client/Documents/Session/IDocumentSession.IncludeCounters.cs
ppekrol: Let me think about it...

src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs
ppekrol: If we will have a method `IncludeAllCounters` then we can create a ctor that covers that (bool variable `includeAllCounters`) and in other ctors check if array is not null or empty. Which will make this part of the code more clear I think.

src/Raven.Client/Documents/Session/IDocumentSession.IncludeCounters.cs
ppekrol: I do not think that we have other choice here. We are in the context of the document here and if we want to download the counters in a single call then there is no other option since `Load` is materializing the entity right away.

src/Raven.Client/Constants.cs
ayende: Why, this is using a private ctor to avoid instantiation. A static class is clearer in intent.
This is there as a namespace scoping mechanism only, after all.

src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs
ayende: Yes, I like that

src/Raven.Client/Documents/Session/IDocumentSession.IncludeCounters.cs
ayende: How about `Includes` property? 
Then we hang everything on top of that?

ayende: Or:
```
session.Includes<Order>("orders/1-A")
    .Counter("likes")
    .Document(x=>x.Employee)
    .Load();
```

ppekrol: Maybe something like that?

```csharp
session.Load<Order>(
    "orders/1-A", 
    includes => includes
        .Counter("likes")
        .Documents(x => x.Employee));
```

ayende: Ohhhh. I _like_ this, yes.

ayende: Would we do the same in `Query`?

ppekrol:
```csharp
session.Query<Company>()
    .Include(includes => includes
        .Counter("likes")
        .Documents(x => x.Employee))
    .Where(x => x.Name == "John")
    .ToList();
```

ayende: Looks good to me, but might be hard to retrofit with what we already have, no?

ppekrol: Should be doable, but we cannot remove old 'Include' methods until 5.0.

ayende: Yes, but I'm not talking about removing the `session.Include`, I'm talking about the `Query().Include` and the ability to do this compare to what we have right now

ppekrol:
```csharp
        public static IRavenQueryable<TResult> Include<TResult>(this IQueryable<TResult> source, Action<IFacetBuilder<TResult>> builder)
        {
            return null;
        }
```

```csharp
                    var orders = session.Query<Order>()
                        .Customize(x => x.WaitForNonStaleResults())
                        .Include(x => x.AllResults())
                        .Include(x => x.CustomerId)
                        .ToList();
```

compiles and works. Although ReSharper is going crazy here with Intellisense.

ayende: I'm in awe. Poor C#.

ayende: Although, we can't use fluent type here, or can we use overload based on the `Func<IncludeBuilder>` ?

ppekrol: I think it could be `Action<IncludeBuilder<TResult>>`.

src/Raven.Client/Constants.cs
aviv86: done

src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs
aviv86: fixed

src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs
aviv86: fixed

src/Raven.Client/Documents/Session/DocumentSession.IncludeCounters.cs
aviv86: fixed

src/Raven.Client/Documents/Session/IDocumentSession.IncludeCounters.cs
aviv86: changed to 
```
session.Load<Order>(
    "orders/1-A", 
    includes => includes
        .Counter("likes")
        .Documents(x => x.Employee));
```

aviv86: create a ticket for 
```
session.Query<Company>()
    .Include(includes => includes
        .Counter("likes")
        .Documents(x => x.Employee))
    .Where(x => x.Name == "John")
    .ToList();
``` 
?

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
aviv86: fixed

src/Raven.Client/Documents/Session/Loaders/MultiLoaderWithInclude.cs
aviv86: leave it with ``string []`` for consistency, or change to List?

src/Raven.Client/Documents/Session/Loaders/MultiLoaderWithInclude.cs
aviv86: changed it to :
```
IIncludeBuilder.Counter(name)
IIncludeBuilder.Counters(names)
IIncludeBuilder.AllCounters()
```

test/SlowTests/Client/Counters/SessionCounters.cs
aviv86: this is handled now 
https://github.com/aviv86/ravendb/blob/9e082d0dd9a3ba4f65bb36d9eda0c6b545a870e7/test/SlowTests/Client/Counters/SessionCounters.cs#L1273

src/Raven.Client/Documents/Session/Loaders/MultiLoaderWithInclude.cs
ayende: keep it consistent

src/Raven.Client/Documents/Session/DocumentSession.Load.cs
```diff
@@ -67,11 +85,12 @@ private void LoadInternal(string[] ids, LoadOperation operation, Stream stream =
             }
         }
 
-        public Dictionary<string, T> LoadInternal<T>(string[] ids, string[] includes)
+        public Dictionary<string, T> LoadInternal<T>(string[] ids, string[] includes, string[] counters = null, bool includeAllCounters = false)
         {
             var loadOperation = new LoadOperation(this);
             loadOperation.ByIds(ids);
             loadOperation.WithIncludes(includes);
+            loadOperation.WithCounters(counters, includeAllCounters);
```
ppekrol: Probably we will want to split it to `WithCounters` and `WithAllCounters`

src/Raven.Client/Documents/Session/Loaders/IIncludeBuilder.cs
```diff
@@ -0,0 +1,100 @@
+using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using Raven.Client.Extensions;
+using Raven.Client.Documents.Queries.Facets;
+
+namespace Raven.Client.Documents.Session.Loaders
+{
+    public interface IIncludeOperations<T>
```
ppekrol: IIncludeBuilder just in case can return IIncludeOperations, but I'm not sure if we will need it here. Maybe drop this and return IIncludeBuilder inside it?

src/Raven.Client/Documents/Session/Loaders/IIncludeBuilder.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        IIncludeOperations<T> Documents(string path);
+
+        IIncludeOperations<T> Documents(Expression<Func<T, string>> path);
+
+        IIncludeOperations<T> Documents(Expression<Func<T, IEnumerable<string>>> path);
+
+    }
+
+    public class IncludeBuilder<T> : IIncludeBuilder<T>, IIncludeOperations<T>
```
ppekrol: change to internal

src/Raven.Client/Documents/Session/Operations/LoadOperation.cs
```diff
@@ -55,6 +63,14 @@ public LoadOperation WithIncludes(string[] includes)
             return this;
         }
 
+        public LoadOperation WithCounters(string[] counters, bool includeAll = false)
```
ppekrol: Let's split this to `WithCounters` and `WithAllCounters`

src/Raven.Client/Documents/Session/Loaders/IIncludeBuilder.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        IIncludeOperations<T> Counters(string[] names);
+
+        IIncludeOperations<T> AllCounters();
+
+        IIncludeOperations<T> Documents(string path);
+
+        IIncludeOperations<T> Documents(Expression<Func<T, string>> path);
+
+        IIncludeOperations<T> Documents(Expression<Func<T, IEnumerable<string>>> path);
```
ppekrol: We need to cover all cases from https://github.com/ravendb/ravendb/blob/v4.0/src/Raven.Client/Documents/Session/IDocumentSession.Include.cs

src/Raven.Client/Documents/Session/IDocumentSession.Load.cs
```diff
@@ -22,7 +24,28 @@ public partial interface IDocumentSession
...
+
+        /// <summary>
+        ///     Loads the specified entities with the specified ids, 
+        ///     and includes other Documents and/or Counters.
+        /// </summary>
+        /// <param name="includes">
+        ///     An action that specifies which documents and\or counters 
+        ///     to include, by using the IIncludeBuilder interface.
+        /// </param>
+        Dictionary<string, T> Load<T>(IEnumerable<string> ids, Action<IIncludeBuilder<T>> includes);
```
ppekrol, line 48: Don't forget about the AsyncSession and add at least one test for it

src/Raven.Client/Documents/Session/Loaders/IIncludeBuilder.cs
```diff
@@ -0,0 +1,100 @@
...
+            return Documents(path.ToPropertyPath());
+        }
+
+        public IIncludeOperations<T> Documents(Expression<Func<T, IEnumerable<string>>> path)
+        {
+            return Documents(path.ToPropertyPath());
+        }
+
+        public IIncludeOperations<T> Counter(string name)
+        {
```
ppekrol: Let's add small validation here, when IncludeAllCounters is `true` then we can throw. Thoughts?

src/Raven.Client/Documents/Session/Loaders/IIncludeBuilder.cs
```diff
@@ -0,0 +1,100 @@
...
+            {
+                if (string.IsNullOrWhiteSpace(name))
+                    throw new InvalidOperationException("Counters(string[] names) : 'names' should not contain null or whitespace elements");
+                CountersToInclude.Add(name);
+            }
+            return this;
+        }
+
+        public IIncludeOperations<T> AllCounters()
+        {
```
ppekrol: Let's add small validation here, when CountersToInclude is not `null` then we can throw. Thoughts?

src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs
```diff
@@ -95,6 +109,19 @@ public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, Serve
                 }
             }
 
+            if (_includeAllCounters)
+            {
+                pathBuilder.Append($"&counter={Constants.Counters.All}");
+            }
+
```
ppekrol: something went wrong with formatting here I think?

src/Raven.Client/Documents/Session/DocumentSession.Load.cs
```diff
@@ -52,6 +54,22 @@ public T Load<T>(string id)
...
         }
 
+        public T Load<T>(string id, Action<IIncludeBuilder<T>> includes)
+        {
+            return Load(new []{id}, includes).Values.FirstOrDefault();
+        }
+
+        public Dictionary<string, T> Load<T>(IEnumerable<string> ids, Action<IIncludeBuilder<T>> includes)
+        {
+            var includeBuilder = new IncludeBuilder<T>();
```
ppekrol: Let's not create IncludeBuilder when action is null

src/Raven.Client/Documents/Session/Loaders/IIncludeBuilder.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        IIncludeOperations<T> Documents(Expression<Func<T, string>> path);
+
+        IIncludeOperations<T> Documents(Expression<Func<T, IEnumerable<string>>> path);
+
+    }
+
+    public class IncludeBuilder<T> : IIncludeBuilder<T>, IIncludeOperations<T>
+    {
+        public List<string> DocumentsToInclude;
```
ppekrol: Maybe we want to use hashset here?

src/Raven.Client/Documents/Session/IDocumentSession.IncludeCounters.cs
ppekrol: When you will address those few cosmetic comments that I just wrote, then we will be able to merge it and you can continue work on the query under the same issue number.

src/Raven.Client/Documents/Session/Loaders/IIncludeBuilder.cs
aviv86:
```
ILoaderWithInclude<T> Include<T, TInclude>(Expression<Func<T, string>> path);
```
and
```
ILoaderWithInclude<T> Include<T, TInclude>(Expression<Func<T, IEnumerable<string>>> path);
```
are obsolete  in 4.0, since we enforce Id property to be ``string`` now

add those anyway?

ppekrol: This is a different thing, check this test:
https://github.com/ravendb/ravendb/blob/v4.0/test/SlowTests/MailingList/TestRavenIncludes.cs
and this issue:
http://issues.hibernatingrhinos.com/issue/RavenDB-10968

src/Raven.Client/Documents/Session/IDocumentSession.Load.cs
aviv86: done

src/Raven.Client/Documents/Session/DocumentSession.Load.cs
aviv86: done

src/Raven.Client/Documents/Session/Loaders/IIncludeBuilder.cs
aviv86: done

src/Raven.Client/Documents/Session/Loaders/IIncludeBuilder.cs
aviv86: done

src/Raven.Client/Documents/Session/Operations/LoadOperation.cs
aviv86: done

src/Raven.Client/Documents/Session/Loaders/IIncludeBuilder.cs
aviv86: done

src/Raven.Client/Documents/Session/Loaders/IIncludeBuilder.cs
aviv86: added

src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs
aviv86: what do you mean?

src/Raven.Client/Documents/Session/DocumentSession.Load.cs
aviv86: fixed

src/Raven.Client/Documents/Session/Loaders/IIncludeBuilder.cs
aviv86: done

ppekrol (approved):

src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs
```diff
@@ -95,6 +109,19 @@ public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, Serve
                 }
             }
 
+            if (_includeAllCounters)
+            {
+                pathBuilder.Append($"&counter={Constants.Counters.All}");
```
ayende: It's actually better to use: `pathBuilder.Append("&counter=").Append(Constants.Counters.All);`

src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs
```diff
@@ -95,6 +109,19 @@ public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, Serve
...
+            }
+
+            else if (_counters != null && _counters.Length > 0)
+            {
+                foreach (var counter in _counters)
+                {
+                    pathBuilder.Append($"&counter={counter}");
```
ayende: Same here: `pathBuilder.Append("&counter=").Append(counter);`

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -1278,6 +1279,76 @@ public void RegisterMissingIncludes(BlittableJsonReaderArray results, BlittableJ
...
+        internal void RegisterCounters(BlittableJsonReaderArray counters, bool gotAll, string[] ids)
+        {
+            if (CountersByDocId == null)
+            {
+                CountersByDocId = new Dictionary<string, (bool GotAll, Dictionary<string, long?> Values)>(StringComparer.OrdinalIgnoreCase);
+            }
+
+            if (counters == null || counters.Length == 0)
+            {
+                if (gotAll == false || ids.Length == 0)
```
ayende: I think you have `||` but you meant `&&` here? what are you trying to check here?

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -1278,6 +1279,76 @@ public void RegisterMissingIncludes(BlittableJsonReaderArray results, BlittableJ
             }
         }
 
+        internal void RegisterCounters(BlittableJsonReaderArray counters, bool gotAll, string[] ids)
```
ayende: what are the `ids` here for, documents ids, or counter names?

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -1278,6 +1279,76 @@ public void RegisterMissingIncludes(BlittableJsonReaderArray results, BlittableJ
...
+            {
+                CountersByDocId = new Dictionary<string, (bool GotAll, Dictionary<string, long?> Values)>(StringComparer.OrdinalIgnoreCase);
+            }
+
+            if (counters == null || counters.Length == 0)
+            {
+                if (gotAll == false || ids.Length == 0)
+                    return;
+                
+                foreach (var id in ids)
```
ayende: Why do you need to create an empty dictionary here?

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -1278,6 +1279,76 @@ public void RegisterMissingIncludes(BlittableJsonReaderArray results, BlittableJ
...
+                cache.Values[name] = value;
+                if (gotAll)
+                {
+                    cache.GotAll = true;
+                }
+
+            }
+        }
+
+        internal void RegisterMissingCounters(string[] ids, string[] counters)
```
ayende: Why have both register counters and register missing? Isn't that basically the same operation?

src/Raven.Client/Documents/Session/Operations/LoadOperation.cs
```diff
@@ -107,6 +128,9 @@ public void SetResult(GetDocumentsResult result)
                 return;
 
             _session.RegisterIncludes(result.Includes);
+            
+            _session.RegisterCounters(result.Counters, _includeAllCounters, _ids);
+            _session.RegisterMissingCounters(_ids, _counters);
```
ayende: Pretty sure that this can be merged into a single operation. You just need to pass the `_counters` to the `RegisterCounters`.

src/Raven.Server/Documents/Handlers/CountersHandler.cs
```diff
@@ -79,6 +80,22 @@ public void Add(string id, CounterOperation op)
                 }
             }
 
+            public void Gather(DocumentsOperationContext context, string id, StringValues counters)
```
ayende: Why this roundabout way of doing this? Instead of creating a command, then executing it, let's do this directly.

src/Raven.Server/Documents/Handlers/DocumentHandler.cs
```diff
@@ -208,10 +211,13 @@ public Task GetDocSize()
 
                 documents.Add(document);
                 includeDocs.Gather(document);
+                counterBatchCmd?.Gather(context, id, counters);
             }
 
             includeDocs.Fill(includes);
 
+            counterBatchCmd?.Execute(context);
```
ayende: can be simplified so there is no `Gather` stage

test/FastTests/Client/Includes.cs
```diff
@@ -51,6 +51,51 @@ public void Can_Load_With_Include()
...
+
+                    session.SaveChanges();
+                }
+
+                using (var session = store.OpenSession())
+                {
+                    var user = session.Load<User>(
+                        "users/1-A",
+                        includes => includes
+                            .Documents(x => x.AddressId));
```
ayende: lambda usually have shorter names.
How about if we do it like:
```
       var user = session.Load<User>(
                        "users/1-A",
                        i=> i.IncludeDocuments(x => x.AddressId));
```

ayende: And I'm not sure about `Documents` or `Document`

ayende (changes requested):

test/FastTests/Client/Includes.cs
ppekrol: The `Documents` are loaded by a path, so it is hard to say if one or many will be loaded. I would stay with `Documents`.

src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs
aviv86: What did you mean in yesterday's comment 'Please use {} here', then?

ppekrol: IIRC you missed brackets in `if else` statement and the C# formatting was incorrect.

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
aviv86: No, this is intentional actually. 
If I have no counters in result, and gotAll is true, and ids is not empty, then I wanna remember that by setting GotAll=true in cache for all the requested documents.

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
aviv86: Same here, if I have no counters in result but asked to get all, then I want to remember that and not go to server again, no?

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
aviv86: Not exactly, RegisterCounters iterates over counter result and adds to cache, RegisterMissing iterates over countersToInclude and checks if in cache.
But this can be merged, yes

src/Raven.Client/Documents/Session/Operations/LoadOperation.cs
aviv86: Okay will merge

src/Raven.Server/Documents/Handlers/CountersHandler.cs
aviv86: Just for consistency.. We use includes. Gather. Will remove this and build CountersBatch object directly, then pass to command ctr and execute

test/FastTests/Client/Includes.cs
aviv86: So, IncludeDocuments or just Documents?

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
aviv86: Document ids

src/Raven.Server/Documents/Handlers/CountersHandler.cs
ppekrol: @ayende this actually depends, if we are going to use this in more than 1 place, then it would be better to wrap it in a command as we do with regular includes, so the code would be more reusable and maintainable. And I guess we will use it everywhere we use regular includes?

test/FastTests/Client/Includes.cs
ppekrol: add `Include` prefix to all methods there.

src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs
aviv86: Oh okay yea I see it now.. Thanks

src/Raven.Server/Documents/Handlers/CountersHandler.cs
ayende: Yes, having all of the include behavior in one location is better, now that we are doing more of that (and we'll continue to do more)

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
ayende: In this case, you can skip the `ids.Length` check, because the `foreach` already has this behavior.

src/Raven.Server/Documents/Handlers/CountersHandler.cs
```diff
@@ -213,6 +208,16 @@ public Task Get()
             return Task.CompletedTask;
         }
 
+        public static void GetInternal(DocumentDatabase database, DocumentsOperationContext context, StringValues counters, string docId, bool full, CountersDetail countersDetail)
```
ppekrol, line 211: I'm wondering if there is a need to pass and fill the CounterDetails and not return them instead of void?

ppekrol:
LGTM, @ayende ?

src/Raven.Server/Documents/Handlers/CountersHandler.cs
aviv86: fixed

PR #6677, opened 2018-06-19
[4.0] RavenDB-11384 & RavenDB-11383

ppekrol:


src/Raven.Server/Documents/ETL/FilterTombstonesEnumerator.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        public bool MoveNext()
+        {
+            Current = null;
+
+            while (_tombstones.MoveNext())
+            {
+                var current = _tombstones.Current;
+                if (current.Type == Tombstone.TombstoneType.Document)
```
ayende, line 26: Wouldn't you need attachment / counter tombstones here as well?

ppekrol: @arekpalinski ?

arekpalinski: It's good for now. I've got an issue for deletion of attachments - http://issues.hibernatingrhinos.com/issue/RavenDB-11379. Counters - work in progress in 4.1

PR #6679, opened 2018-06-19
[4.0] RavenDB-11366 InvalidDataException Expected to get type of 'LogLength…

karmeli87:
…NegotiationResponse' message, but got 'Error' message.

The issue is probably due to a race between disposing a follower in term X, while creating a new one in the same term.
This could occur during a short period split-brain (network delay) so the follower went to elections but was immediately forced by the leader to becamce his follower again.

### DIFF
--- a/src/Raven.Server/Rachis/Follower.cs
+++ b/src/Raven.Server/Rachis/Follower.cs
@@ -19,6 +19,8 @@ namespace Raven.Server.Rachis
 {
     public class Follower : IDisposable
     {
+        private static int _uniqueId;
+
         private readonly RachisConsensus _engine;
         private readonly long _term;
         private readonly RemoteConnection _connection;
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Rachis/Follower.cs
```diff
@@ -32,8 +33,9 @@ public Follower(RachisConsensus engine, long term, RemoteConnection remoteConnec
             _engine = engine;
             _connection = remoteConnection;
             _term = term;
-            
-            _debugName = $"Follower in term {_term}";
+
+            // this will give us a unique identifier for this follower
+            _debugName = $"Follower in term {_term} started at {SystemTime.UtcNow:s}";
```
ayende: This isn't actually going to generate unique identifier. This is already racing, so this will happen if you have this in the same second. That is pretty much _guaranteed_ to be the case. You need to use an actual unique value, best to use something like `Interlocked.Increment` on a static value here.

ayende (changes requested):

PR #6685, opened 2018-06-19
RavenDB-11390 - force invariant culture

ayende:


ppekrol:
I will apply this to 4.1

PR #6713, opened 2018-06-22
RavenDB-11157 Raven ETL - Counters 

arekpalinski:


src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -203,15 +283,42 @@ public void PutCounterFromReplication(DocumentsOperationContext context, string
...
+                var tombstoneTable = context.Transaction.InnerTransaction.OpenTable(TombstonesSchema, CountersTombstonesSlice);
+
+                if (tombstoneTable.ReadByKey(keyPerfix, out var existingTombstone))
+                {
+                    tombstoneTable.Delete(existingTombstone.Id);
+                }
+            }
+        }
+
+        private readonly HashSet<string> _tableCreated = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
ayende, line 294: why do we care, it is safe to call `CountersSchema.Create` twice, no?

src/Raven.Server/Documents/ETL/EtlProcess.cs
```diff
@@ -291,8 +348,19 @@ private void EnterFallbackMode()
...
         protected abstract void LoadInternal(IEnumerable<TTransformed> items, JsonOperationContext context);
 
-        public bool CanContinueBatch(EtlStatsScope stats)
+        public bool CanContinueBatch(EtlStatsScope stats, TExtracted currentItem)
         {
+            if (currentItem.Type == EtlItemType.Counter)
+            {
+                // we have special counters enumerator which ensures that we iterate counters up to last processed doc etag
+                // although as long as it returns counters we need to ETL all of them
+
```
ayende: Why is this the case?

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -142,51 +198,75 @@ private static ReplicationBatchItem CreateReplicationBatchItem(DocumentsOperatio
                 Name = name,
                 ChangeVector = TableValueToString(context, (int)CountersTable.ChangeVector, ref result.Reader),
                 Value = TableValueToLong((int)CountersTable.Value, ref result.Reader),
+                Collection = TableValueToId(context, (int)CountersTable.Collection, ref result.Reader),
                 TransactionMarker = TableValueToShort((int)CountersTable.TransactionMarker, nameof(ReplicationBatchItem.TransactionMarker), ref result.Reader),
             };
         }
 
-        public void PutCounterFromReplication(DocumentsOperationContext context, string documentId, string name, string changeVector, long value)
+        public enum PutCounterMode
```
ayende: Don't really like it. This is putting high level concern at the storage level.

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlItem.cs
```diff
@@ -1,15 +1,28 @@
...
            
         }
 
-        public RavenEtlItem(Tombstone tombstone, string collection) : base(tombstone, collection)
+        public RavenEtlItem(Tombstone tombstone, string collection, EtlItemType type) : base(tombstone, collection, type)
         {
            
         }
+
+        public RavenEtlItem(CounterDetail counter, string collection)
```
maximburyak, line 17: I think that we need here another, RavenCounterEtlItem that inherits from RavenEtlItem.

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -130,6 +168,24 @@ private static (LazyStringValue Doc, LazyStringValue Name) ExtractDocIdAndName(J
...
+            var size = counterTombstoneId.Size;
+                
+            int sizeOfDocId = 0;
+            for (; sizeOfDocId < size; sizeOfDocId++)
+            {
+                if (p[sizeOfDocId] == SpecialChars.RecordSeparator)
+                    break;
+            }
+
+            var doc = Encoding.UTF8.GetString(p, sizeOfDocId);
```
maximburyak: In case of big amount of changes and counters, we may end up materializing a lot strings here. I think that we can actually stay with LazyStringValue through all the process. For the dictionary, for an instance, we use the LazyStringValueComparer without having significant changes in our code base.

src/Raven.Server/Documents/CountersStorage.cs
arekpalinski: It's it. It was meant to prevent from calling it on every operation that modifies a counter. Actually I borrowed this code from revisions storage - should it be removed from both?

src/Raven.Server/Documents/ETL/EtlProcess.cs
arekpalinski: The comment isn't correct actually. We _can_ send all counters that we have but we _don't have to_. Also see the TODO comment below - we should prevent from creating counters huge batch here

src/Raven.Server/Documents/CountersStorage.cs
arekpalinski: You don't like the mode switch here or dislike the values of enum that are high level terms? Do you prefer to have separate methods?

ayende: Don't like the intermixing of responsability. The storage shouldn't make decisions based on the source.
The storage should offer options that match what the source need.

src/Raven.Server/Documents/CountersStorage.cs
ayende: It's probably cheaper not to.

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlItem.cs
arekpalinski: I can do that but not sure if it's really worth. The logic is based on `EtlItemType` anyway (the same concept as we have for `ReplicationBatchItem`). What advantages do you see of introducing a new type here?

src/Raven.Server/Documents/CountersStorage.cs
arekpalinski: Good idea. Thanks

src/Raven.Server/Documents/CountersStorage.cs
arekpalinski: I'll stay with current code then.

src/Raven.Server/Documents/CountersStorage.cs
arekpalinski: The point is passing or not the change vector. In ETL we intentionally don't use it while the replication and import need to pass it. I've exposed the relevant overloads instead of having the notion of the mode:

https://github.com/arekpalinski/ravendb/commit/aa4aea3a39f872b7eb995d5231265b1afec19eb6

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlItem.cs
maximburyak: I think that because we are going to add other sub entities like time series etc, we might want to add some better seperation between records, other than one general entity for everything.

arekpalinski: I'll try to make `RavenEtlItem` an abstract class so we could have separate implementations like: `RavenDocumentEtlItem` `RavenCounterEtlItem` etc

ayende: Let's wait here until we actually need it. Time Series is about 6 months away

arekpalinski: OK

src/Raven.Server/Documents/ETL/EtlProcess.cs
arekpalinski: Please check this commit: https://github.com/ravendb/ravendb/pull/6713/commits/d3891d8a661356f5b

src/Raven.Server/Documents/CountersStorage.cs
arekpalinski: Changed in: https://github.com/ravendb/ravendb/pull/6713/commits/9d8132eff09b20562c4b88ed74

src/Raven.Server/Documents/CountersStorage.cs
arekpalinski: Done for document id in https://github.com/ravendb/ravendb/pull/6713/commits/19390245f79 . We need the counter name as a string anyway

arekpalinski:
It can be reviewed and merged. Summary of dealing with counters by Raven ETL can be found here: http://issues.hibernatingrhinos.com/issue/RavenDB-11157#comment=67-29744

PR #6718, opened 2018-06-24
RavenDB-7787a Add Certificate (Security) and DatabaseRecord configuration

IdanHaim:


src/Raven.Embedded/RavenEmbedded.cs
```diff
@@ -35,6 +39,29 @@ public void StartServer(ServerOptions options = null)
...
+            if (options.Security != null)
+            {
+                _certificate = options.Security.ClientCertificate;
+
+                try
+                {
+                    var thumbprint = options.Security.ServerCertificiateThumbprint;
+                    RequestExecutor.ServerCertificateCustomValidationCallback +=
+                        (message, certificate2, chain, errors) =>
+                            certificate2.Thumbprint == thumbprint;
```
maximburyak, line 51: Shoud be configurable. What if we have several thumbprints etc? ServerCertificateCustomValidationCallback is static

src/Raven.Embedded/DatabaseOptions.cs
```diff
@@ -1,9 +1,13 @@
-﻿namespace Raven.Embedded
+﻿using Raven.Client.ServerWide;
+
+namespace Raven.Embedded
 {
     public class DatabaseOptions
     {
         public string DatabaseName { get; set; }
```
ppekrol, line 7: Now when we have DatabaseRecord then I'm not sure if we need this. We should drop this or change DatabaseRecord to `Action<DatabaseRecord>` to modify the record which database name was set using the `DatabaseName` property.

src/Raven.Embedded/RavenEmbedded.cs
ppekrol: Not sure if this is something that we need to support with embedded? Only one instance of server will be started by application.

IdanHaim: The server still needs to be trusted

ayende: And this add trust to the server.
User can also configure this separately from us, which will overwrite this logic.

src/Raven.Embedded/DatabaseOptions.cs
ayende: I like the action better, yes.

ayende:
@ppekrol ` certificate2.Thumbprint == thumbprint;` is required when running HTTPS, otherwise, how would we know that the server is trustworthy?
By definition, almost, we aren't going to be using a valid cert, since we are bound to `127.0.0.1`.

ppekrol:
I was answering to @maximburyak

PR #6723, opened 2018-06-24
[4.0] RavenDB-11194 Prevent from putting a database on a node that is down.

karmeli87:


ayende:
Did you miss the promotable fix?

PR #6724, opened 2018-06-24
[4.1] RavenDB-11194 Prevent from putting a database on a node that is down.

karmeli87:


src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -1473,6 +1474,63 @@ private static bool DatabaseNeedsToRunIdleOperations(DocumentDatabase database)
...
         }
 
+        public void AssignNodesToDatabase(ClusterTopology clusterTopology, DatabaseRecord record)
+        {
+            var topology = record.Topology;
+
+            Debug.Assert(topology != null);
+
+            var clusterNodes = clusterTopology.Members.Keys
+                .Concat(clusterTopology.Promotables.Keys)
```
ayende: We shouldn't add it to promotables. They may take a while to come up and actually figure that one out.
Only members / watchers should be there.

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -1473,6 +1474,63 @@ private static bool DatabaseNeedsToRunIdleOperations(DocumentDatabase database)
...
+                .Concat(clusterTopology.Watchers.Keys)
+                .ToList();
+
+            if (record.Encrypted)
+            {
+                clusterNodes.RemoveAll(n => AdminDatabasesHandler.NotUsingHttps(clusterTopology.GetUrlFromTag(n)));
+                if (clusterNodes.Count == 0)
```
ayende: Need to compare to the replication factor here, no?

ayende (changes requested):

src/Raven.Server/ServerWide/ServerStore.cs
karmeli87: done

src/Raven.Server/ServerWide/ServerStore.cs
karmeli87: OK, done.

PR #6727, opened 2018-06-25
RavenDB-11292: Fix pwrite call for 32bit linux

aviviadi:


src/Sparrow/Platform/Posix/Syscall.cs
```diff
@@ -193,9 +193,18 @@ public static long read(int fd, void* buf, ulong count)
         [DllImport(LIBC_6, SetLastError = true)]
         private static extern IntPtr pwrite(int fd, IntPtr buf, UIntPtr count, IntPtr offset);
 
+        [DllImport(LIBC_6, SetLastError = true)]
+        private static extern IntPtr pwrite(int fd, IntPtr buf, UIntPtr count, long offset);
+
         public static long pwrite(int fd, void* buf, ulong count, long offset)
         {
-            if(PlatformDetails.RunningOnMacOsx)
+            if (PlatformDetails.Is32Bits)
```
ayende, line 201: This is wrong.
We _have_ to call `pwrite64` in 32 bits, otherwise we can't handle > 4GB files.
We call `pwrite` on MacOS because that always handles this, but the `pwrite` for macos should be long for the offset, at least.

aviviadi: i've verified this, pwrite does write in the Pi > 4GB.   With pwrite64 it fails

ayende: You are correct.

aviviadi: note that pwrite with `long long int` for offset signature is used for pwrite under 32bit

PR #6728, opened 2018-06-25
RavenDB-11292: Fix pwrite call for 32bit linux

aviviadi:


src/Sparrow/Platform/Posix/Syscall.cs
```diff
@@ -266,7 +275,7 @@ public static int AllocateFileSpace(int fd, long size, string file, out bool usi
                         // fallocate is not supported, we'll use lseek instead
                         usingWrite = true;
                         byte b = 0;
-                        if (pwrite(fd, &b, 1, size - 1) != 1)
+                        if (pwrite(fd, &b, 1, size - 1) != 1L)
```
ayende, line 278: This is wrong on the return value, no?
`-1 on error`, not `1L`

aviviadi: it is 1.  not -1..  we check we actually wrote 1 byte

ayende: ok

PR #6743, opened 2018-06-25
RavenDB-11409 Debug code that will output if the request was aborted

arekpalinski:


src/Raven.Server/RavenServerStartup.cs
```diff
@@ -176,7 +176,10 @@ public static bool IsHtmlAcceptable(HttpContext context)
                     LogTrafficWatch(context, 0, database ?? "N/A");
 
                 if (context.RequestAborted.IsCancellationRequested)
+                {
+                    Console.WriteLine($"Aborted request. Request: {context.Request.Path} {context.Request.QueryString}.{Environment.NewLine}Exception: {e}");
```
ayende, line 180: This should not be merged. We don't want to write to the console.
If this is required, let's send this to the log. 

For the purpose of verification, we do want this on Jenkins, so let's at least wrap this in `#if DEBUG`.

Also, we need to handle this on the client side as well, providing timing information for the error.

ppekrol: We will remove it soon. This only happens in release.
If really needed, we can add some other constant to the jenkins so this will be hit.

arekpalinski: We got the socket error in some PRs but did't see the console output added here. Reverted in https://github.com/ravendb/ravendb/pull/6746

PR #6744, opened 2018-06-25
RavenDb-7787 Remove DatabaseName property from DatabaseOptions

IdanHaim:


src/Raven.Embedded/DatabaseOptions.cs
```diff
@@ -1,13 +1,22 @@
...
     public class DatabaseOptions
     {
-        public string DatabaseName { get; set; }
-
         public bool SkipCreatingDatabase { get; set; }
 
-        public DatabaseRecord DatabaseRecord { get; set; }
+        public DatabaseRecord DatabaseRecord { get; private set; }
+
+        public DatabaseOptions(string databaseName) : this(new DatabaseRecord(databaseName))
```
ayende, line 12: I like this!

PR #6758, opened 2018-06-26
Raven db 11425

talweiss1982:
** This PR include some minor fixes to javascript indexes examples that are unrelated to the issue (commit - RavenDB-11239 adding Map-Reduce example)

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Static/JavaScriptReduceOperation.cs
+++ b/src/Raven.Server/Documents/Indexes/Static/JavaScriptReduceOperation.cs
@@ -85,10 +85,21 @@ namespace Raven.Server.Documents.Indexes.Static
 
                     if (xCalCulated == false)
                         _xKey.Process(_allocator, xVal);
+
                     if (yCalculated == false)
                         _yKey.Process(_allocator, yVal);
                 }
 
+                var xIsNotAllNulls = _xKey.IsBufferSet;
+                var yIsNotAllNulls = _yKey.IsBufferSet;
+                // null == null
+                if (xIsNotAllNulls == false && yIsNotAllNulls == false)
+                    return true;
+                // x == null and y != null or y != null and y == null
+                if (xIsNotAllNulls == false || yIsNotAllNulls == false)
+                    return false;
+
+                //At this point both buffer should be populated
                 var xBuffer = _xKey.GetBuffer();
                 var yBuffer = _yKey.GetBuffer();
                 _lastUsedBucket = x;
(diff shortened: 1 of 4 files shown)


ayende (approved):

src/Raven.Server/Documents/Indexes/Static/JavaScriptReduceOperation.cs
```diff
@@ -85,10 +85,21 @@ public unsafe bool Equals(BlittableJsonReaderObject x, BlittableJsonReaderObject
...
+                var xIsNotAllNulls = _xKey.IsBufferSet;
+                var yIsNotAllNulls = _yKey.IsBufferSet;
+                // null == null
+                if (xIsNotAllNulls == false && yIsNotAllNulls == false)
+                    return true;
+                // x == null and y != null or y != null and y == null
+                if (xIsNotAllNulls == false || yIsNotAllNulls == false)
+                    return false;
+
+                //At this point both buffer should be populated
```
arekpalinski, line 102: Just wondering why do we need to compare the buffers instead of using `ReduceKeyProcessor.Hash`?

talweiss1982: that would work but it will calculate the hash, the memory compare is faster I believe.

talweiss1982: wait, i recall, we need to actually avoid hash colusions here, that is why.

arekpalinski: Hash calculation is pretty optimized as I know and didn't showed up to be expensive part when profiling static indexes. I don't think we should bother too much on this at this point - especially that we didn't profile JS indexes yet.

arekpalinski: okay, I'll check how static c# indexes deals with that - don't remember exactly now

talweiss1982: It doesn't matter if this is really rare we still need to handle it correctly.
I did profile JS indexes and Map-Reduce and this was the result of the performance optimization.
The bottom line is that the code is working, working fast and most importantly correct.

arekpalinski: Great then :)

PR #6764, opened 2018-06-26
[4.0] RavenDB-11391: Calling `put()` in patch should not include metadata a…

haludi:
…ttachments property

### DIFF
--- a/src/Raven.Server/Documents/Patch/ScriptRunner.cs
+++ b/src/Raven.Server/Documents/Patch/ScriptRunner.cs
@@ -408,7 +408,13 @@ namespace Raven.Server.Documents.Patch
                 {
                     reader = JsBlittableBridge.Translate(_jsonCtx, ScriptEngine, args[1].AsObject(), usageMode: BlittableJsonDocumentBuilder.UsageMode.ToDisk);
 
-                    var put = _database.DocumentsStorage.Put(_docsCtx, id, _docsCtx.GetLazyString(changeVector), reader);
+                    var put = _database.DocumentsStorage.Put(
+                        _docsCtx, 
+                        id, 
+                        _docsCtx.GetLazyString(changeVector), 
+                        reader, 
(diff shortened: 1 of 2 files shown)


src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -408,7 +408,13 @@ public JsValue PutDocument(JsValue self, JsValue[] args)
                 {
                     reader = JsBlittableBridge.Translate(_jsonCtx, ScriptEngine, args[1].AsObject(), usageMode: BlittableJsonDocumentBuilder.UsageMode.ToDisk);
 
-                    var put = _database.DocumentsStorage.Put(_docsCtx, id, _docsCtx.GetLazyString(changeVector), reader);
+                    var put = _database.DocumentsStorage.Put(
+                        _docsCtx, 
+                        id, 
+                        _docsCtx.GetLazyString(changeVector), 
+                        reader, 
+                        nonPersistentFlags: NonPersistentDocumentFlags.ResolveAttachmentsConflict
```
ayende, line 416: what does this do?

haludi:
This flag will cause DocumentPutAction to recreate attachment metadata
property according to attachment metadata table

```csharp
private bool ShouldRecreateAttachments(DocumentsOperationContext context,
Slice lowerId, BlittableJsonReaderObject oldDoc,
            BlittableJsonReaderObject document, ref DocumentFlags flags,
NonPersistentDocumentFlags nonPersistentFlags)
        {
            if ((nonPersistentFlags &
NonPersistentDocumentFlags.ResolveAttachmentsConflict) ==
NonPersistentDocumentFlags.ResolveAttachmentsConflict)
            {
                document.TryGet(Constants.Documents.Metadata.Key, out
BlittableJsonReaderObject metadata);
               * RecreateAttachments(context, lowerId, document, metadata,
ref flags);*
                return true;
            }
}
```

ayende:
In this case, let's add a comment there explaining that. In 6 months, that might be a puzzle.

PR #6769, opened 2018-06-26
[4.1] RavenDB-10614

aviv86:


ayende (approved):

src/Raven.Server/Smuggler/Documents/DatabaseSmuggler.cs
```diff
@@ -600,6 +600,14 @@ private SmugglerProgressBase.Counts ProcessDocuments(SmugglerResult result, Buil
                     item.Document.Flags = item.Document.Flags.Strip(DocumentFlags.FromClusterTransaction);
                     item.Document.NonPersistentFlags |= NonPersistentDocumentFlags.FromSmuggler;
 
+                    if (_destination is DatabaseDestination
+                        && _options.SkipRevisionCreation
+                        && _options.OperateOnTypes.HasFlag(DatabaseItemType.RevisionDocuments)
+                        && item.Document.Flags.HasFlag(DocumentFlags.HasRevisions))
```
ppekrol: I think that last condition is not necessary?

src/Raven.Server/Smuggler/Documents/DatabaseSmuggler.cs
```diff
@@ -600,6 +600,14 @@ private SmugglerProgressBase.Counts ProcessDocuments(SmugglerResult result, Buil
                     item.Document.Flags = item.Document.Flags.Strip(DocumentFlags.FromClusterTransaction);
                     item.Document.NonPersistentFlags |= NonPersistentDocumentFlags.FromSmuggler;
 
+                    if (_destination is DatabaseDestination
+                        && _options.SkipRevisionCreation
+                        && _options.OperateOnTypes.HasFlag(DatabaseItemType.RevisionDocuments)
```
ppekrol: I'm wondering if this is necessary (I'm aware that we've agreed to add it). Maybe if someone does not want to create revisions then we should just add the 'SkipRevisionCreation' without checking for this? @ayende what is your opinion?

ppekrol (approved):

src/Raven.Server/Smuggler/Documents/DatabaseSmuggler.cs
ayende: Yes, we can just always add that.
For that matter, since this is non persistent flag, we don't even care if this goes to a db destination.
Stream destination doesn't care.

aviv86: removed all conditions except `_options.SkipRevisionCreation`

PR #6771, opened 2018-06-26
Raven db 8354 fix

ml054:


src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -524,7 +524,24 @@ internal IndexCreationOptions GetIndexCreationOptions(object indexDefinition, In
             return IndexCreationOptions.Update;
         }
 
-        private void ValidateIndexName(string name, bool isStatic)
+        public static bool IsValidResourceName(string name, bool isStatic, out string errorMessage)
```
ppekrol: `IsValidIndexName` would be better

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -524,7 +524,24 @@ internal IndexCreationOptions GetIndexCreationOptions(object indexDefinition, In
...
+        {
+            errorMessage = null;
+            
+            try
+            {
+               ValidateIndexName(name, isStatic);
+            }
+            catch (Exception e)
+            {
+                errorMessage = e.Message.Split('\n')[0];
```
ppekrol, line 537: what are we splitting here?

src/Raven.Server/Web/Studio/StudioTasksHandler.cs
```diff
@@ -7,14 +8,65 @@
...
+using Raven.Server.Utils;
+using Raven.Server.Web.System;
 using Sparrow.Json;
 using Sparrow.Json.Parsing;
 using Voron.Util.Settings;
+using IndexStoreStatic = Raven.Server.Documents.Indexes.IndexStore;
 
 namespace Raven.Server.Web.Studio
 {
+    public enum ElementType
```
ppekrol: Can we put it inside StudioTasksHandler? Rename to `ItemType` ?

src/Raven.Server/Web/Studio/StudioTasksHandler.cs
```diff
@@ -7,14 +8,65 @@
 using Raven.Server.Config;
 using Raven.Server.Routing;
 using Raven.Server.ServerWide;
+using Raven.Server.Utils;
+using Raven.Server.Web.System;
 using Sparrow.Json;
 using Sparrow.Json.Parsing;
 using Voron.Util.Settings;
+using IndexStoreStatic = Raven.Server.Documents.Indexes.IndexStore;
```
ppekrol: Can be removed?

src/Raven.Server/Web/Studio/StudioTasksHandler.cs
```diff
@@ -7,14 +8,65 @@
...
     {
+        [RavenAction("/studio-tasks/is-valid-name", "GET", AuthorizationStatus.ValidUser)]
+        public Task IsValidName()
+        {
+            if (Enum.TryParse(GetQueryStringValueAndAssertIfSingleAndNotEmpty("type").Trim(), out ElementType elementType) == false)
+            {
+                throw new ArgumentException($"Type {elementType} is not supported");
+            }
+            
+            var name = GetQueryStringValueAndAssertIfSingleAndNotEmpty("name").Trim();
```
ppekrol, line 30: Do we need those `Trims` ?

src/Raven.Server/Web/Studio/StudioTasksHandler.cs
```diff
@@ -7,14 +8,65 @@
...
+            var path = GetStringQueryString("dataPath", false);
+            
+            using (ServerStore.ContextPool.AllocateOperationContext(out JsonOperationContext context))
+            {
+                HttpContext.Response.StatusCode = (int)HttpStatusCode.OK;
```
ppekrol: FYI, OK is set by default.

ppekrol (changes requested):

src/Raven.Server/Web/Studio/StudioTasksHandler.cs
Danielle9897: ok - I will remove

src/Raven.Server/Web/Studio/StudioTasksHandler.cs
Danielle9897: removed

Danielle9897: The current regex:
```public const string ValidResourceNameCharacters = @"([_\-\.]+)";```
doesn't allow for spaces

src/Raven.Server/Web/Studio/StudioTasksHandler.cs
Danielle9897: removed

src/Raven.Server/Documents/Indexes/IndexStore.cs
Danielle9897: done

src/Raven.Server/Web/Studio/StudioTasksHandler.cs
Danielle9897: done

src/Raven.Server/Documents/Indexes/IndexStore.cs
Danielle9897: Because e.message contains:

"Index name 'abc?' is not permitted. Only letters, digits and characters that match regex '([_\/\-\.]+)' are allowed. *\r\n* Parameter name: name"


with the parameter name added

PR #6803, opened 2018-06-27
[4.0] RavenDB-11421

aviv86:


src/Raven.Client/Documents/LinqExtensions.cs
```diff
@@ -861,6 +861,20 @@ public static IOrderedQueryable<T> OrderByScore<T>(this IQueryable<T> self)
             return (IOrderedQueryable<T>)queryable;
         }
 
+        /// <summary>
+        /// Perform an initial sort by lucene score.
+        /// </summary>
+        public static IOrderedQueryable<T> ThenByScore<T>(this IQueryable<T> self)
```
ppekrol: We need to change this to `IOrderedQueryable` IIRC

src/Raven.Client/Documents/LinqExtensions.cs
```diff
@@ -875,6 +889,20 @@ public static IOrderedQueryable<T> OrderByScoreDescending<T>(this IQueryable<T>
             return (IOrderedQueryable<T>)queryable;
         }
 
+        /// <summary>
+        /// Perform an initial sort by lucene score descending.
+        /// </summary>
+        public static IOrderedQueryable<T> ThenByScoreDescending<T>(this IQueryable<T> self)
```
ppekrol: We need to change this to IOrderedQueryable IIRC

ppekrol (changes requested):

src/Raven.Client/Documents/LinqExtensions.cs
ppekrol: I'm referring to `this IQueryable<T> self`

aviv86: fixed

src/Raven.Client/Documents/LinqExtensions.cs
aviv86: fixed

PR #6808, opened 2018-06-27
RDoc-1214 rename configuration property Hours->Hrs

iftahbe:


ppekrol:
Let's leave it as in in 4.0 and support both in 4.1. Unless we want to migrate the support for multiple ConfigurationEntry attributes.

ppekrol:
@iftahbe since I've made that call, I will handle this :)

iftahbe:
Oh sorry, it's a mistake.
I meant to PR only for 4.1 but accidentally did it for 4.0 
(and yes, I should have kept the old entry too..)

PR #6809, opened 2018-06-27
RavenDB-11248 Failed SqlEtlTests.WillLog

ayende:
Avoid race condition between listening to the log and starting the operation


ayende:
Green build :-)

PR #6811, opened 2018-06-27
[4.0] RavenDB-11015 Removed node will truncate his log to the index of the …

karmeli87:
…last topology.

- Also prevent from the process to crash on maintenance exception.

### DIFF
--- a/src/Raven.Server/ServerWide/Maintenance/ClusterMaintenanceWorker.cs
+++ b/src/Raven.Server/ServerWide/Maintenance/ClusterMaintenanceWorker.cs
@@ -42,7 +42,30 @@ namespace Raven.Server.ServerWide.Maintenance
 
         public void Start()
         {
-            _collectingTask = PoolOfThreads.GlobalRavenThreadPool.LongRunning(x => CollectDatabasesStatusReport(), null, _name);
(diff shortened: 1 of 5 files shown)


src/Raven.Server/ServerWide/Maintenance/ClusterMaintenanceWorker.cs
```diff
@@ -42,7 +42,30 @@ public ClusterMaintenanceWorker(TcpConnectionOptions tcp, CancellationToken exte
...
+                {
+                    // expected
+                }
+                catch (Exception e)
+                {
+                    if (_logger.IsInfoEnabled)
+                    {
+                        _logger.Info($"Exception occurred while collecting info from {_server.NodeTag}. Task is closed.", e);
+                    }
+                    // we don't want to crash the process so we don't propagate this exception.
```
maximburyak, line 65: But it means that now we don't have a cluster maintenance worker at all. If we do that, we need a watchdog that will revive it if it's dead

ayende: That shouldn't be an `Info`. It should be `Operations` and it should be a `Notifications`.

This way, it is user visible. 
I don't want to create another thread here, might get into a dangerous loop

PR #6813, opened 2018-06-27
[4.0] RavenDB-11440 Provide an option to change logging level with the usage of client API

ppekrol:


src/Raven.Server/Documents/Handlers/Admin/AdminLogsHandler.cs
```diff
@@ -1,12 +1,54 @@
...
+        [RavenAction("/admin/logs/configuration", "POST", AuthorizationStatus.ClusterAdmin)]
+        public async Task SetConfiguration()
+        {
+            using (ServerStore.ContextPool.AllocateOperationContext(out JsonOperationContext context))
+            {
+                var json = await context.ReadForMemoryAsync(RequestBodyStream(), "logs/configuration");
+
+                var configuration = JsonDeserializationServer.Parameters.SetLogsConfigurationParameters(json);
+
+                LoggingSource.Instance.SetupLogMode(configuration.Mode, Server.Configuration.Logs.Path.FullPath);
```
ayende, line 46: Validate arguments here ?

ppekrol: I'm not changing the path (it is the default one). Only the mode.

PR #6818, opened 2018-06-28
[4.0] RavenDB-11446 Can't save index with whitespace

Danielle9897:


src/Raven.Client/Documents/Indexes/PutIndexResult.cs
```diff
@@ -4,13 +4,40 @@
...
+            return new DynamicJsonValue
+            {
+                [nameof(Index)] = Index
+            };
+        }
+    }
+    
+    public class PutIndexesResult : IDynamicJson
+    {
+        public List<PutIndexResult> indexes { get; set; }
```
ppekrol: Should be `Indexes`

src/Raven.Server/Documents/Handlers/Admin/AdminIndexHandler.cs
```diff
@@ -48,24 +48,25 @@ public class AdminIndexHandler : DatabaseRequestHandler
...
                 HttpContext.Response.StatusCode = (int)HttpStatusCode.Created;
 
+                var indexesResult = new PutIndexesResult
+                {
+                    indexes = createdIndexes
+                };
+                
                 using (var writer = new BlittableJsonTextWriter(context, ResponseBodyStream()))
                 {
-                    writer.WriteStartObject();
```
ppekrol: This is a breaking change. Old clients will not work properly. You need to stay with the JSON that we had.

ppekrol (changes requested):

src/Raven.Server/Documents/Handlers/Admin/AdminIndexHandler.cs
```diff
@@ -21,14 +20,15 @@ public class AdminIndexHandler : DatabaseRequestHandler
...
-                var createdIndexes = new List<string>();
+                var createdIndexes = new List<PutIndexResult>();
                 var input = await context.ReadForMemoryAsync(RequestBodyStream(), "Indexes");
                 if (input.TryGet("Indexes", out BlittableJsonReaderArray indexes) == false)
                     ThrowRequiredPropertyNameInRequest("Indexes");
 
                 foreach (var indexToAdd in indexes)
                 {
                     var indexDefinition = JsonDeserializationServer.IndexDefinition((BlittableJsonReaderObject)indexToAdd);
+                    indexDefinition.Name = indexDefinition.Name.Trim();
```
ppekrol: This can throw NRE and name validation is done later in code.

ppekrol: You can change that to: `indexDefinition.Name?.Trim();`

Danielle9897: ok

src/Raven.Client/Documents/Indexes/PutIndexResult.cs
Danielle9897: class removed

src/Raven.Server/Documents/Handlers/Admin/AdminIndexHandler.cs
Danielle9897: used ``` context.Write(writer, new DynamicJsonArray(createdIndexes.Select(x => x.ToJson())));``` instead

PR #6825, opened 2018-06-28
[4.1] RavenDB-11452

ppekrol:
- AsyncDocumentSession.LoadAsync should return default(T) when id is null to match sync session behavior
- added various null argument checks
- synchronized AsyncDocumentSession.Load and DocumentSession.Load implementations

### DIFF
--- a/src/Raven.Client/Documents/Session/AsyncDocumentSession.Load.cs
+++ b/src/Raven.Client/Documents/Session/AsyncDocumentSession.Load.cs
@@ -4,7 +4,6 @@ using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
-using Raven.Client.Documents.Commands;
 using Raven.Client.Documents.Session.Loaders;
 using Raven.Client.Documents.Session.Operations;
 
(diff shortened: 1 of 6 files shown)


arekpalinski (approved):

src/Raven.Client/Documents/Session/AsyncDocumentSession.Load.cs
```diff
@@ -28,25 +30,35 @@ public partial class AsyncDocumentSession
...
+                return default;
+
+            var result = await LoadAsync(new[] { id }, includes, token).ConfigureAwait(false);
+
             return result.Values.FirstOrDefault();
         }
 
-        public Task<Dictionary<string, T>> LoadAsync<T>(IEnumerable<string> ids, Action<IIncludeBuilder<T>> includes, CancellationToken token = default(CancellationToken))
+        /// <inheritdoc />
+        public Task<Dictionary<string, T>> LoadAsync<T>(IEnumerable<string> ids, Action<IIncludeBuilder<T>> includes, CancellationToken token = default)
```
ayende, line 57: Need to check `ids` for null

PR #6866, opened 2018-07-01
[4.1] RavenDB-11369 RachisTests.DisableNodeOnClusterTest.BackToFirstNodeAft…

karmeli87:
…erRevive failed

### DIFF
--- a/src/Raven.Client/Http/RequestExecutor.cs
+++ b/src/Raven.Client/Http/RequestExecutor.cs
@@ -1111,6 +1111,27 @@ namespace Raven.Client.Http
                 nodeStatus.StartTimer();
         }
 
+        internal Task CheckNodeStatusNow(string tag)
+        {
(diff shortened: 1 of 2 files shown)


src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1111,6 +1111,27 @@ private void SpawnHealthChecks(ServerNode chosenNode, int nodeIndex)
                 nodeStatus.StartTimer();
         }
 
+        public Task CheckNodeStatusNow(string tag)
```
ppekrol: internal?

karmeli87: Wasn't sure myself. 
You tell me :)

ppekrol: Make it internal for now.

PR #6889, opened 2018-07-02
RavenDB-11477

ayende:
- Better error if test fails
 - Avoid checking directory permissions outside a lock
 - Simplify database path names in tests
 - Fixes an issue where there was a race condition when updating the non passive state of the cluster node before the cluster transaction has been committed

RavenDB-11403
 - Fixing bad subscription mode causing the test to fail because it assumed drop connection is synchronous
 - Avoid race condition in the test

### DIFF
--- a/test/Tests.Infrastructure/RavenTestHelper.cs
+++ b/test/Tests.Infrastructure/RavenTestHelper.cs
@@ -37,7 +37,7 @@ namespace FastTests
         {
             testName = testName?.Replace("<", "").Replace(">", "");
 
-            var newDataDir = Path.GetFullPath($".\\Databases\\{testName ?? "TestDatabase"}_{serverPort}-{DateTime.Now:yyyy-MM-dd-HH-mm-ss-fff}-{Interlocked.Increment(ref _pathCount)}");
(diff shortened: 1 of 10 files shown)


test/Tests.Infrastructure/RavenTestHelper.cs
```diff
@@ -37,7 +37,7 @@ public static string NewDataPath(string testName, int serverPort, bool forceCrea
         {
             testName = testName?.Replace("<", "").Replace(">", "");
 
-            var newDataDir = Path.GetFullPath($".\\Databases\\{testName ?? "TestDatabase"}_{serverPort}-{DateTime.Now:yyyy-MM-dd-HH-mm-ss-fff}-{Interlocked.Increment(ref _pathCount)}");
+            var newDataDir = Path.GetFullPath($".\\Databases\\{testName ?? "TestDatabase"}@{serverPort}-{Interlocked.Increment(ref _pathCount)}");
```
ppekrol, line 40: @ayende the `@` char is causing some test failures. Can we switch back to `_` ?

ayende: Yes, we can switch it back, I'll do so shortly.

PR #6894, opened 2018-07-02
[4.1] RavenDB-11296  include counters in query 

aviv86:


src/Raven.Client/Documents/Queries/QueryResultBase.cs
```diff
@@ -15,6 +18,17 @@ public abstract class QueryResultBase<TResult, TInclude>
         /// </summary>
         public TInclude Includes { get; set; }
 
+        /// <summary>
+        /// Gets or sets the Counters included in the result.
+        /// </summary>
+        public BlittableJsonReaderObject CounterIncludes { get; set; }
```
ppekrol, line 24: I think it would be better to be consistent with `Load` here, so this should be named `Counters`.

src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
```diff
@@ -492,6 +498,33 @@ public void Include(Expression<Func<T, object>> path)
...
+            {
+                foreach (var doc in includeBuilder.DocumentsToInclude)
+                {
+                    DocumentIncludes.Add(doc);
+                }
+            }
+
+            if (includeBuilder.AllCounters)
+            {
+                CounterIncludes = new HashSet<string>();
```
ppekrol: Better to wrap it inside CounterIncludesToken

src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
```diff
@@ -492,6 +498,33 @@ public void Include(Expression<Func<T, object>> path)
...
+            }
+
+            else if (includeBuilder.CountersToInclude?.Count > 0)
+            {
+                CounterIncludes = includeBuilder.CountersToInclude;
```
ppekrol: Better to wrap it inside CounterIncludesToken

src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
```diff
@@ -1143,6 +1180,30 @@ private void BuildInclude(StringBuilder queryText)
                 }
             }
 
+            if (CounterIncludes != null)
+            {
+                if (first == false)
+                    queryText.Append(",");
+
+                queryText.Append("counters(");
```
ppekrol: This should be handled by CounterIncludesToken

src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
```diff
@@ -1143,6 +1180,30 @@ private void BuildInclude(StringBuilder queryText)
...
+
+                first = true;
+                foreach (var counter in CounterIncludes)
+                {
+                    if (first == false)
+                        queryText.Append(",");
+                    first = false;
+
+                    queryText.Append("'");
```
ppekrol: in QueryToken you have a `WriteField` method

src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
ppekrol: I'm wondering if we should not support parameters here? @ayende ?

src/Raven.Server/Documents/Queries/QueryMetadata.cs
```diff
@@ -486,6 +502,55 @@ private void HandleSelectFunctionBody(BlittableJsonReaderObject parameters)
...
+                }
+            }
+
+            if (start == expression.Arguments.Count)
+            {
+                counterIncludes.Counters[sourcePath ?? string.Empty] = new HashSet<string>();
+                return;
+            }
+
+            for (var index = start; index < expression.Arguments.Count; index++)
```
ppekrol, line 527: IIRC `QueryBuilder.GetValues` is doing the same thing.

src/Raven.Client/Documents/Session/Operations/LoadOperation.cs
```diff
@@ -169,7 +169,18 @@ public void SetResult(GetDocumentsResult result)
...
+            if (_includeAllCounters || _counters != null)
+            {
+                var counterToInclude = new Dictionary<string, string[]>();
+                var counters = _counters ?? new string[0];
+                foreach (var id in _ids)
+                {
+                    counterToInclude[id] = counters;
+                }
+
+                _session.RegisterCounters(result.Counters, counterToInclude);
```
ppekrol: I'm not sure that I'm following this code here. Why are we doing this like that? We cannot include different counters per document ID. Seems redundant and a lot of additional work?

src/Raven.Server/Documents/Includes/IncludeCountersCommand.cs
```diff
@@ -10,23 +12,54 @@ public class IncludeCountersCommand
...
+            : this(database, context)
+        {
+            _countersBySourcePath = countersBySourcePath.ToDictionary(kvp => kvp.Key, kvp => kvp.Value.ToArray());
+        }
+
+        public void Fill(Document document)
+        {
+            var docId = document.Id;
+
+            foreach (var kvp in _countersBySourcePath)
```
ppekrol, line 48: What is that? I think we do not support this in API?

ppekrol (changes requested):

src/Raven.Server/Documents/Includes/IncludeCountersCommand.cs
aviv86: yes, there is no support for this in client api (yet ?)
but it is supported in rql:
https://github.com/aviv86/ravendb/blob/e58118da3ae828e0400b1138c3df52e9ee3ce1f7/test/SlowTests/Client/Counters/QueryOnCounters.cs#L2250

src/Raven.Client/Documents/LinqExtensions.cs
```diff
@@ -81,6 +82,28 @@ public static IRavenQueryable<TResult> Include<TResult>(this IQueryable<TResult>
...
+        /// <summary>
+        /// Includes the specified documents and/or counters in the query
+        /// </summary>
+        /// <typeparam name="TResult">The type of the object that holds the id that you want to include.</typeparam>
+        /// <param name="source">The source for querying</param>
+        /// <param name="includes">Specifies the documents and/or counters to include </param>
+        /// <returns></returns>
+        public static IRavenQueryable<TResult> Include<TResult>(this IQueryable<TResult> source, Action<IIncludeBuilder<TResult>> includes)
+        {
+#if CURRENT
```
ayende: compilation error if this sin't defined, no?

src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
ayende: And yes, we want to handle this via parameters, not direct injection of string into the query

src/Raven.Client/Documents/Queries/QueryResultBase.cs
maximburyak: Yeah, but in the context of the query results, it's iportant to be clear and state that it's counters received by includes, I guess the name could be "IncludedCounters" or something, but the "include" part is significant.

ppekrol: Works for me also, but it should be in sync with `Load` and any other operation.

src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
aviv86: done

src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
aviv86: done

src/Raven.Client/Documents/Session/AbstractDocumentQuery.CounterIncludes.cs
```diff
@@ -0,0 +1,24 @@
...
+            if (includeAll == false)
+            {
+                if (counters?.Count > 0 == false)
+                    return;
+
+                CounterIncludesToken = CounterIncludesToken.Create(counters, QueryParameters);
+                return;
+            }
+
+            CounterIncludesToken = CounterIncludesToken.Create(null, null);
```
ppekrol: You can substitute this this another method called e.g. `.All` that will do that under the hood.

src/Raven.Client/Documents/Session/Tokens/CounterIncludesToken.cs
```diff
@@ -0,0 +1,43 @@
...
+using Raven.Client.Extensions;
+
+namespace Raven.Client.Documents.Session.Tokens
+{
+    public class CounterIncludesToken : QueryToken
+    {
+        private readonly Parameters _queryParameters;
+        private readonly HashSet<string> _counters;
+
+        private CounterIncludesToken(HashSet<string> countres, Parameters queryParameters)
```
ppekrol: typo :)

src/Raven.Client/Documents/Session/Tokens/CounterIncludesToken.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            _counters = countres;
+            _queryParameters = queryParameters;
+        }
+
+        public static CounterIncludesToken Create(HashSet<string> countres, Parameters queryParameters)
```
ppekrol: pass only queryParameter name. Check how we do it in other places.

e.g.
```
var whereToken = WhereToken.Create(WhereOperator.AllIn, fieldName, AddQueryParameter(array));
```

src/Raven.Client/Documents/Session/Tokens/CounterIncludesToken.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public override void WriteTo(StringBuilder writer)
+        {
+            writer
+                .Append("counters(");
+
+            if (_queryParameters != null)
+            {
+                var parameterName = $"p{_queryParameters.Count.ToInvariantString()}";
+                _queryParameters.Add(parameterName, _counters);
```
ppekrol: if we are including `All` counters then you do not need to create a new query parameter just `counters()` will be enough

src/Raven.Server/Documents/Queries/StreamDocumentQueryResult.cs
```diff
@@ -42,6 +44,16 @@ public override void AddExplanation(ExplanationResult explanationResult)
             throw new NotSupportedException();
         }
 
+        public override void AddCounterIncludes(IncludeCountersCommand includeCountersCommand)
+        {
+            throw new NotSupportedException();
```
ppekrol: This is something that we will want to support

ppekrol (changes requested):

src/Raven.Client/Documents/Session/Tokens/CounterIncludesToken.cs
aviv86: actually there is a reason why i delayed adding the parameter to the point when we actually write the RQL.
I did like you suggested at first, the problem with that is that these kinds of queries:
```
session.Query<Order>()
	.Include(i => i
		.IncludeCounter("Downloads")
		.IncludeDocuments(x => x.Company))
	.Where(x => x.ShipTo.Country == "Brazil");
```
will generate :
```
from Orders where ShipTo.Country = $p1 
include Company,counters($p0)
```
because `Include` is computed before `Where`, but `where` comes first in RQL.

ppekrol: It is not a problem that the order of parameters does not match.

ayende: Yes, that isn't an issue.

src/Raven.Client/Documents/Session/Tokens/CounterIncludesToken.cs
aviv86: done

src/Raven.Client/Documents/Session/Tokens/CounterIncludesToken.cs
aviv86: fixed

src/Raven.Client/Documents/Session/AbstractDocumentQuery.CounterIncludes.cs
aviv86: done

src/Raven.Client/Documents/Queries/QueryResultBase.cs
aviv86: changed from `Counters` to `CounterIncludes` in `Load`

src/Raven.Server/Documents/Includes/IncludeCountersCommand.cs
aviv86: client support for this will be in a separate PR
```
session.Query<Order>()
    .Include(i => i
        .IncludeCounter("Downloads")
        .IncludeCounter(o => o.Employee, "Downloads")
        .IncludeDocuments(x => x.Company))
```

src/Raven.Server/Documents/Queries/QueryMetadata.cs
aviv86: fixed

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -1314,6 +1314,32 @@ public void RegisterMissingIncludes(BlittableJsonReaderArray results, BlittableJ
...
+            if (NoTracking)
+                return;
+
+            if (resultCounters == null || resultCounters.Count == 0)
+            {
+                if (gotAll)
+                {
+                    foreach (var id in ids)
+                    {
+                        SetGotAllCountersForDocumet(id);
```
ppekrol: typo

ppekrol (approved):
LGTM, @ayende ?

PR #6897, opened 2018-07-03
[4.0] RavenDB-11478 Check if we are not leaking servers or client tasks tha…

karmeli87:
…t are issuing request in a periodic manner

The issue was in `Round_robin_load_balancing_with_failing_node_should_work` (which had the same database name as the `Round_robin_load_balancing_should_work` test), that we haven't disposed the request executor.


ppekrol (approved):
Waiting for CI

PR #6906, opened 2018-07-03
[4.0] RavenDB-11288 retrying queries when ODE is caught

ppekrol:


ayende (commented):
I don't like the infinite recursion part.
If for some reason we do get a case where an index is disposed, we'll recurse until we hit stack overflow.
Better to do that only once, then throw if we get it twice in a row.

ayende (approved):

ppekrol:
Switched to loop (3x)

PR #6916, opened 2018-07-04
[4.0] RavenDB-11486

ppekrol:


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1019,14 +1020,22 @@ protected void ExecuteIndexing()
...
-                                        // this can fail if the indexes lock is currently held, so we'll retry
-                                        // however, we might be requested to shutdown, so we want to skip replacing
-                                        // in this case, worst case scenario we'll handle this in the next batch
-                                        while (_indexingProcessCancellationTokenSource.IsCancellationRequested == false)
+                                        try
+                                        {
+                                            // this can fail if the indexes lock is currently held, so we'll retry
+                                            // however, we might be requested to shutdown, so we want to skip replacing
+                                            // in this case, worst case scenario we'll handle this in the next batch
+                                            while (_indexingProcessCancellationTokenSource.IsCancellationRequested == false)
```
ayende, line 1030: Can this ever be an infinite loop?

ppekrol: I'm pretty sure that it can't.

ayende: As an aside, I looked at `TryReplaceIndexes`, it _never_ returns false?!

PR #6920, opened 2018-07-04
Raven db 10953

talweiss1982:


src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -341,11 +348,31 @@ void HandleSocketException(SocketException e)
                 if (_cts.IsCancellationRequested)
                     return;
 
+                DynamicJsonValue returnValue;
+
+                if (e is AggregateException ag && ag.InnerExceptions.Count == 1 && ag.InnerExceptions[0] is MissingAttachmentException mae)
```
arekpalinski: You can use `ExtractSingleInnerException` to simplify that

src/Raven.Server/Documents/AttachmentsStorage.cs
```diff
@@ -467,6 +469,30 @@ public IEnumerable<Attachment> GetAttachmentsForDocument(DocumentsOperationConte
...
+                        continue;
+
+                    attachment.Size = GetAttachmentStreamLength(context, attachment.Base64Hash);
+
+                    yield return attachment;
+                }
+            }
+        }
+
+        private static ByteStringContext.InternalScope From(DocumentsOperationContext context, LazyStringValue lowerDocumentId, out Slice lowerDocumentIdSlice)
```
arekpalinski: Either rename this method or drop it and call `Slice.From` directly

src/Raven.Server/Documents/AttachmentsStorage.cs
```diff
@@ -558,6 +589,11 @@ public Stream GetAttachmentStream(DocumentsOperationContext context, Slice hashS
             return tree.ReadStream(hashSlice);
         }
 
+        public bool VerifyAttachmentStreamExist(DocumentsOperationContext context, Slice hashSlice)
```
arekpalinski: Let's make it private

src/Raven.Server/Documents/AttachmentsStorage.cs
```diff
@@ -522,12 +548,17 @@ public Attachment GetAttachment(DocumentsOperationContext context, string docume
...
+                        return _documentsStorage.AttachmentsStorage.GetAttachment(context, documentId, name, AttachmentType.Document, null, withoutStream);
                     }
                 }
 
                 return null;
             }
+            // We were invoked just for the purpose of checking the existence of the attachment, no need to actually read the stream from disk only to verify its there.
+            if (withoutStream)
+            {
+                return VerifyAttachmentStreamExist(context, attachment.Base64Hash)?attachment:null;
```
arekpalinski: Why is it possible to have an attachment without the stream?

arekpalinski: BTW it isn't obvious for the caller of this method that `withoutStream` parameter changes the logic and calls  `VerifyAttachmentStreamExist` - it looks the storage should have a dedicated method for that check

src/Raven.Server/Documents/AttachmentsStorage.cs
```diff
@@ -997,5 +1033,26 @@ public static void AssertAttachments(BlittableJsonReaderObject document, Documen
             }
         }
 #endif
+        public void AssertAttachmentsFromReplication(BlittableJsonReaderObject document, AttachmentType type,string id, DocumentsOperationContext context)
+        {
+            var changeVector = type == AttachmentType.Document ? null : document.GetChangeVector();
```
arekpalinski: Is it supposed to work for attachments of revisions as well (`AttachmentType.Revision`)?

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -525,7 +554,17 @@ private unsafe void ReceiveSingleDocumentsBatch(DocumentsOperationContext docume
             catch (Exception e)
             {
                 if (_log.IsInfoEnabled)
-                    _log.Info("Failed to receive documents replication batch. This is not supposed to happen, and is likely a bug.", e);
+                {
+                    //This is the case where we had a missing attachment, it is rare but expected.
+                    if (e is AggregateException ag && ag.InnerExceptions.Count == 1 && ag.InnerExceptions[0] is MissingAttachmentException mae)
```
arekpalinski: Please use `ExtractSingleInnerException `

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
```diff
@@ -240,6 +244,22 @@ public bool ExecuteReplicationOnce(OutgoingReplicationStatsScope stats, ref Date
 
                             _stats.Storage.RecordInputAttempt();
 
+                            //Here we add missing attachments in the same batch as the document that contains them without modifying the last etag or transaction boundry
+                            if (MissingAttachmentsInLastBatch && item.Type == ReplicationBatchItem.ReplicationItemType.Document &&
+                                (item.Flags & DocumentFlags.HasAttachments) == DocumentFlags.HasAttachments)
+                            {
+                                var type = (item.Flags & DocumentFlags.Revision) == DocumentFlags.Revision ? AttachmentType.Revision: AttachmentType.Document;
+                                foreach (var attachment in _parent._database.DocumentsStorage.AttachmentsStorage.GetAttachmentsForDocument(documentsContext, type, item.Id)
+                                    .Where(a => a.Etag <= firstBatchEtag).Select(a => ReplicationBatchItem.From(a)))
```
arekpalinski: We can avoid Linq here and filtering it explicitly in the loop - it might be slightly more optimized way

src/Voron/Data/BTrees/Tree.Stream.cs
```diff
@@ -257,6 +257,27 @@ public VoronStream ReadStream(Slice key)
             return new VoronStream(tree.Name, chunksDetails, _llt);
         }
 
+        /// <summary>
+        /// This method just verify that a stream exist under the given key
+        /// </summary>
+        /// <param name="key"> The key of the stream</param>
+        /// <returns>'true' if there is a stream under the given 'key' and 'false' otherwise</returns>
+        public bool SeekStream(Slice key)
```
arekpalinski: As the purpose of this method is to check the existence  of the stream we'll be better of by renaming it to `Exists` or something like that

arekpalinski:
Please check failing replication tests

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
talweiss1982: okay

src/Raven.Server/Documents/AttachmentsStorage.cs
talweiss1982: its unsafe method and you can't call unsafe from IEnumerable

src/Raven.Server/Documents/AttachmentsStorage.cs
talweiss1982: okay

src/Raven.Server/Documents/AttachmentsStorage.cs
talweiss1982: This is an optimization so we won't need to fetch the stream, it is invoked just to verify the existence of a stream without having to read the stream from the disk.

src/Raven.Server/Documents/AttachmentsStorage.cs
talweiss1982: it suppose to unless you know something i don't, what is the issue here?

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
talweiss1982: okay

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
talweiss1982: this is a rare case and will only invoke if there are missing attachments and if this document has attachments so the performance penalty is neglectable.

src/Voron/Data/BTrees/Tree.Stream.cs
talweiss1982: okay

src/Raven.Server/Documents/AttachmentsStorage.cs
ayende: Since we just want to verify existence, why not just have an `AttachmentExists` method?

ayende: I agree with @arekpalinski that the boolean arg changing the meaning so drastically is better done with a different method entirely.

src/Raven.Server/Documents/AttachmentsStorage.cs
```diff
@@ -997,5 +1033,26 @@ public static void AssertAttachments(BlittableJsonReaderObject document, Documen
...
+            var changeVector = type == AttachmentType.Document ? null : cv;
+            if (document.TryGet(Constants.Documents.Metadata.Key, out BlittableJsonReaderObject metadata) &&
+                metadata.TryGet(Constants.Documents.Metadata.Attachments, out BlittableJsonReaderArray attachments))
+            {
+                foreach (BlittableJsonReaderObject attachment in attachments)
+                {
+                    if (attachment.TryGet(nameof(AttachmentName.Name), out LazyStringValue name) &&
+                        attachment.TryGet(nameof(AttachmentName.Hash), out LazyStringValue hash))
+                    {
+                        if (GetAttachment(context, id, name, type, changeVector, withoutStream:true) == null)
```
ayende: Here you don't need any of the validation that happens in `GetAttachment`, can probably call `VerifyAttachmentStreamExist` directly

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -341,11 +348,31 @@ void HandleSocketException(SocketException e)
                 if (_cts.IsCancellationRequested)
                     return;
 
+                DynamicJsonValue returnValue;
+
+                if (e.ExtractSingleInnerException() is MissingAttachmentException mae)
+                {
+                    if (_cts.IsCancellationRequested)
```
ayende: We just checked this, no need to do so again

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -1080,6 +1119,8 @@ public override int Execute(DocumentsOperationContext context)
                                         //the other side will receive negative ack and will retry sending again.
                                         document = new BlittableJsonReaderObject(_buffer + item.Position, item.DocumentSize, context);
                                         document.BlittableValidation();
+                                        var attachmentType = item.Flags.Contain(DocumentFlags.Revision) ? AttachmentType.Revision : AttachmentType.Document;
```
ayende: Pretty sure that we don't really care if this is revision or document.
We only care if the attachment hash exists

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
ayende: I don't think we need to do any of that. 
You already _have_ all the attachments you need, in the metadata of the document, no?

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
```diff
@@ -300,10 +322,17 @@ public bool ExecuteReplicationOnce(OutgoingReplicationStatsScope stats, ref Date
                             _log.Info("Failed to send document replication batch", e);
                         throw;
                     }
+
+                    if (hadMissingAttachment)
```
ayende: why not just set unconditionally

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
```diff
@@ -441,7 +470,8 @@ private bool AddReplicationItemToBatch(ReplicationBatchItem item, OutgoingReplic
             }
 
             // destination already has it
-            if (ChangeVectorUtils.GetConflictStatus(item.ChangeVector, _parent.LastAcceptedChangeVector) == ConflictStatus.AlreadyMerged)
+            if (MissingAttachmentsInLastBatch == false &&
```
ayende: When will this happen?

src/Voron/Data/BTrees/Tree.Stream.cs
```diff
@@ -257,6 +257,27 @@ public VoronStream ReadStream(Slice key)
...
+        /// <returns>'true' if there is a stream under the given 'key' and 'false' otherwise</returns>
+        public bool StreamExist(Slice key)
+        {
+            var tree = FixedTreeFor(key, ChunkDetails.SizeOf);
+            var numberOfChunks = tree.NumberOfEntries;
+
+            if (numberOfChunks <= 0)
+                return false;
+
+            using (var it = tree.Iterate())
```
ayende: Why? This must be true, you already checked `NumberOfEntries`

ayende (changes requested):

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
talweiss1982: The type effects the key we search for and it is importent, so we do care.

src/Raven.Server/Documents/AttachmentsStorage.cs
talweiss1982: i have not changed the logic of the method even by a bit, was just trying to save the actual read of the stream from the disk where only its existence i care about.
I see that effort is not been appreciated next time I'll just invoke code like /databases/*/attachments HEAD

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
talweiss1982: no

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
talweiss1982: Code refactoring, leftovers...

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
talweiss1982: This is the code that filters items that have lower CV, this will filter the missing attachments if we ignore MissingAttachmentsInLastBatch == false

src/Voron/Data/BTrees/Tree.Stream.cs
talweiss1982: I was not familiar with the code enough to realize reading the Tree is enough

src/Raven.Server/Documents/AttachmentsStorage.cs
arekpalinski: ok, makes sense

src/Raven.Server/Documents/AttachmentsStorage.cs
arekpalinski: Just asking. I wondered why we pass `null` here for revisions - I saw your other comment about the key that we search on and understand now

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
ayende: Why do you need a type? All you need to do is check if the hash is there, and you already have this from the metadata.

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
ayende: why not?

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
talweiss1982: the type defines if the key of the attachment is constructed by the document id or it's change vector if it is a revision the change vector replace the id, why? I don't know I haven't written it, this is whats implemented.

talweiss1982:
var type = (item.Flags & DocumentFlags.Revision) == DocumentFlags.Revision ? AttachmentType.Revision: AttachmentType.Document;
                                foreach (var attachment in _parent._database.DocumentsStorage.AttachmentsStorage.GetAttachmentsForDocument(documentsContext, type, item.Id)
                                    .Where(a => a.Etag <= firstBatchEtag).Select(a => ReplicationBatchItem.From(a)))

This is the code that actually brings the attachments into the batch why don't we need it ?
where do expect me to fetch the attachments from?
all we have are the descriptor of attachments in the document's metadata not the actual attachments

ppekrol:
@talweiss1982 can you check failing tests?

talweiss1982:
@ppekrol the majority of them are failing because of conflicts not been created (not a bug but a feature) need to go over them one by one and verify whether a bug or a feature

ppekrol:
Ok, great.

src/Raven.Server/Documents/AttachmentsStorage.cs
```diff
@@ -467,6 +470,30 @@ public IEnumerable<Attachment> GetAttachmentsForDocument(DocumentsOperationConte
...
+                        continue;
+
+                    attachment.Size = GetAttachmentStreamLength(context, attachment.Base64Hash);
+
+                    yield return attachment;
+                }
+            }
+        }
+
+        private static ByteStringContext.InternalScope GetSliceFromLazyString(DocumentsOperationContext context, LazyStringValue lowerDocumentId, out Slice lowerDocumentIdSlice)
```
ayende: Not sure if this is worth a separate method.

src/Raven.Server/Documents/AttachmentsStorage.cs
```diff
@@ -997,5 +1030,25 @@ public static void AssertAttachments(BlittableJsonReaderObject document, Documen
             }
         }
 #endif
+        public void AssertAttachmentsFromReplication(DocumentsOperationContext context, string id, BlittableJsonReaderObject document)
+        {
+            if (document.TryGet(Constants.Documents.Metadata.Key, out BlittableJsonReaderObject metadata) &&
+                metadata.TryGet(Constants.Documents.Metadata.Attachments, out BlittableJsonReaderArray attachments))
+            {
```
ayende: Let's get everything about the document itself out of here, we shouldn't have the document's metadata details in attachment storage

src/Raven.Server/Documents/AttachmentsStorage.cs
ayende: It isn't worth it, but this is required by the compiler, cannot use `yield` and `unsafe` together.

ayende: Fixed this by adding a new `Slice.From` overload

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
ayende: The reason we require it here is that we need to send all the attachment's data.
That include the change vector, content-type, etc. These are stored on a per attachment, not at the hash level.

ayende: However, we don't need to have the `Where` here, we can rely on the fact that attachments are _always_ modified before the document.

ayende: Therefor, we can remove all the `Linq` in here completely

ayende:
Pushed the wrong branch, forcing pushing.

PR #6924, opened 2018-07-04
[4.1] RavenDB-11282 If the replacement index wasn't running then we shouldn…

arekpalinski:
…'t start it after renaming it. Making sure nobody will start the index if we stop it during the replacement. Changing the order of stop indexing / put index operations to test what we actually want - we want to test that we can get back to the original index def if the replacement index still exists

### DIFF
--- a/src/Raven.Server/Documents/Indexes/IndexStore.cs
+++ b/src/Raven.Server/Documents/Indexes/IndexStore.cs
@@ -1237,23 +1237,26 @@ namespace Raven.Server.Documents.Indexes
 
             _indexes.ReplaceIndex(oldIndexName, oldIndex, newIndex);
 
-            var needToStop = PoolOfThreads.LongRunningWork.Current != newIndex._indexingThread;
+            using (newIndex.DrainRunningQueries()) // to ensure nobody will start index meanwhile if we stop it here
+            {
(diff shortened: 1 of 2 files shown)


maximburyak:
This commit description is very hard to read

arekpalinski:
Tried to do my best :) In particular, wanted to include as much details as possible to describe the issue. What changes would you suggest to make it clearer / easier to read for you?

maximburyak:
@arekpalinski Sorry for my late response! 

I guess that if you didn't write anything detailed, one (me, for example) would say: "well, it's not clear, I'll move on", but because you already started writing something detailed, it draws the eye, and it could be great if it could be more clear. 

I would start from splitting it into three separate sentences, because the two first sentences looks from the first glance as continuation of one another, and it's not: 

1) If the replacement index wasn't running then we shouldn't start it after renaming it. 
2) Making sure nobody will start an index if we stop it during the replacement. 
3) Changing the order of stop indexing / put index operations to test what we actually want - we want to test that we can get back to the original index def if the replacement index still exists

Then, maybe I would split sentence 3 into a different commit, because, if I understood correctly, it's not directly about the first two changes. On the other hand, if it would stay, I would add a "Fixing test by.." prefix, so it will be clear to what it refers.

arekpalinski:
Thanks for the hints!

PR #6926, opened 2018-07-04
RavenDB-11477 Move all pre-processing of commands on the leader to an…

ayende:
… event to avoid race conditions during elections

- Fix proper dispose order
- Ensure we can wait properly for database deletion even if we aren't deleting just specific nodes
- Better name for data path in test
- Increase timeout undre stress test

### DIFF
--- a/src/Raven.Server/ServerWide/ServerStore.cs
+++ b/src/Raven.Server/ServerWide/ServerStore.cs
@@ -528,6 +528,7 @@ namespace Raven.Server.ServerWide
             }
 
             _engine = new RachisConsensus<ClusterStateMachine>(this);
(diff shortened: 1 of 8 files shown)


src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -537,6 +538,21 @@ public void Initialize()
             ConfigureAuditLog();
         }
 
+
+        private void BeforeAppendToRaftLog(object sender, CommandBase cmd)
+        {
```
karmeli87: To be even more safe, maybe we should pass the context? So we get the cluster topology (and assign nodes) with the same context as we put the command.

PR #6955, opened 2018-07-06
RavenDB-11491 removed duplicate tests

ppekrol:


maximburyak:
@ppekrol I don't think that those tests are entirely duplicate.. The goal of those tests is to make sure that  the behavior of the indexes that inherit from AbstractJavaScriptIndexCreationTask is as those that inherit from AbstractJavascriptIndexCreationTask, but has all the flags set manually. In the bottom line, those tests checks if all the properties like 'Maps', 'Reduce', 'Fields' etc. are passed correctly to the index definition creation. Simply removing those checks is not good, but those tests don't serve their goal either, I guess that we better have a more specific tests that checks this transformation between the AbstractJavaScriptIndexCreationTask  and IndexDefinition. What do you think?

ppekrol:
@efratshenhar reverted this.

ppekrol:
Now we have only tests that inherit from `AbstractJavaScriptIndexCreationTask` and I think that is the way to go, because they are converting to IndexDefinition at the end, so we are checking both things at once.

maximburyak:
Agreed, makes more sense!

PR #6957, opened 2018-07-07
[4.0] RavenDB-11379 Fixed handling of attachment deletions in Raven ETL by…

arekpalinski:
… tracking attachment tombstones and sending AttachmentDELETE command

### DIFF
--- a/src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtl.cs
+++ b/src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtl.cs
@@ -41,6 +41,14 @@ namespace Raven.Server.Documents.ETL.Providers.Raven
             return new TombstonesToRavenEtlItems(tombstones, collection);
         }
 
+        protected override bool ShouldTrackAttachmentTombstones()
+        {
+            // if script isn't empty we relay on addAttachment() calls and detect that attachments needs be deleted
+            // when this call gets an attachment reference marked as null ($attachment/{attachment-name}/$null)
+
+            return string.IsNullOrEmpty(Transformation.Script);
+        }
+
         protected override EtlTransformer<RavenEtlItem, ICommandData> GetTransformer(DocumentsOperationContext context)
         {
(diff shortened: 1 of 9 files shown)


src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtl.cs
```diff
@@ -41,6 +41,14 @@ protected override IEnumerator<RavenEtlItem> ConvertTombstonesEnumerator(IEnumer
             return new TombstonesToRavenEtlItems(tombstones, collection);
         }
 
+        protected override bool ShouldTrackAttachmentTombstones()
+        {
+            // if script isn't empty we relay on addAttachment() calls and detect that attachments needs be deleted
```
ayende, line 46: `relay` -> `rely`

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtl.cs
```diff
@@ -41,6 +41,14 @@ protected override IEnumerator<RavenEtlItem> ConvertTombstonesEnumerator(IEnumer
...
+            // when this call gets an attachment reference marked as null ($attachment/{attachment-name}/$null)
```
ayende, line 47: What happens if they do something like:

```
for(var attachment in this["@metadata"]["@attachments"])
{
   if(attachment.Name.endsWith(".png"))
       addAttachment(...);
}
```

In this manner, you don't _see_ that there are attachments to be deleted on the other side.

arekpalinski: Correct. Also it wound't work for renaming attachments. It means we always have to track them.

ayende: So, we need to delete all the other attachments that exists on destination when we send it, no?

arekpalinski: True, that would make it easier.

PR #6962, opened 2018-07-08
 [4.0] RavenDB-10697 add netcoreapp2.1 target in Raven.Client

iftahbe:


src/Raven.Client/Documents/Changes/DatabaseChanges.cs
```diff
@@ -56,11 +58,26 @@ public DatabaseChanges(RequestExecutor requestExecutor, string databaseName, Act
             _task = DoWork();
         }
 
+        public static event Func<object, X509Certificate, X509Chain, SslPolicyErrors, bool> WebSocketsRemoteCertificateValidationCallback;
```
ppekrol: Probably we want to support multiple of these. Check how it is done in RequestExecutor.

src/Raven.Client/Documents/Changes/DatabaseChanges.cs
```diff
@@ -56,11 +58,26 @@ public DatabaseChanges(RequestExecutor requestExecutor, string databaseName, Act
...
+            return WebSocketsRemoteCertificateValidationCallback(sender, certificate, chain, errors);
+        }
+
         public static ClientWebSocket CreateClientWebSocket(RequestExecutor requestExecutor)
         {
             var clientWebSocket = new ClientWebSocket();
             if (requestExecutor.Certificate != null)
                 clientWebSocket.Options.ClientCertificates.Add(requestExecutor.Certificate);
+
+#if NETCOREAPP2_1
```
ppekrol: change to NETCOREAPP, we will not have to remember to change this when 2_2 or 3_0 will be out.

src/Raven.Client/Documents/DocumentQueryExtensions.cs
```diff
@@ -1,4 +1,4 @@
-﻿#if NETSTANDARD2_0
+﻿#if NETSTANDARD2_0 || NETCOREAPP2_1
```
ppekrol: Change to NETCOREAPP

src/Raven.Client/Documents/LinqExtensions.cs
```diff
@@ -4,7 +4,7 @@
 // </copyright>
 //-----------------------------------------------------------------------
 
-#if NETSTANDARD2_0
+#if NETSTANDARD2_0 || NETCOREAPP2_1
```
ppekrol: Change to NETCOREAPP

src/Raven.Client/Raven.Client.csproj
```diff
@@ -72,4 +72,7 @@
     <PackageReference Include="System.Threading.ThreadPool" Version="4.3.0" />
     <PackageReference Include="System.ValueTuple" Version="4.5.0" />
   </ItemGroup>
+  <ItemGroup Condition="$(TargetFramework) == 'netcoreapp2.1'">
+    <PackageReference Include="Microsoft.AspNetCore.WebSockets" Version="2.1.1" />
```
ppekrol: This is a server-side package? Why do we need this?

src/Raven.Client/Raven.Client.csproj
```diff
@@ -3,7 +3,7 @@
     <Version></Version>
     <Description>RavenDB Client is the client library for accessing RavenDB</Description>
     <Authors>Hibernating Rhinos</Authors>
-    <TargetFrameworks>netstandard1.3;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netstandard1.3;netstandard2.0;netcoreapp2.1</TargetFrameworks>
```
ppekrol, line 6: Check if we are including Sparrow.dll when NuGet package is created. This is done in this csproj (line 33).

src/Raven.Client/Documents/Changes/DatabaseChanges.cs
ppekrol: I'm wondering if this is the right place. Now it is the only place in client using web-sockets, but if this will change? Maybe it would be best to keep it in RequestExecutor like the other callback?

ppekrol: When this is changed after DatabaseChanges was created then we need to `ReconnectClient` probably.

ppekrol (changes requested):

src/Raven.Client/Documents/Changes/DatabaseChanges.cs
ayende: Let's reuse the `RequestExecuter` impl here. This way, a user only need to configure it once globally.
There is no sense in defining the same logic once for all calls and once for websockets.

iftahbe: I don't believe we need multiple user validation callbacks. The implementation in RequestExecutor was done like that because of the server, registering its own callback at startup (and had a race with the callback used in tests which we register in TestBase.

I don't mind moving it to RE but I think each feature should have its own callback. So if we add a new feature in the future which uses websockets, it will not share the same callback as DatabaseChanges

ppekrol: I do not see a good reason for having separate websocket callbacks per feature. Having separate callbacks in DatabaseChanges and some other feature would only be confusing. Let's try using the RE implementation if possible.

src/Raven.Client/Raven.Client.csproj
iftahbe: Yes you are right, it's not required

src/Raven.Client/Raven.Client.csproj
```diff
@@ -38,6 +38,11 @@
       <PackagePath>lib/netstandard2.0/</PackagePath>
       <Pack>true</Pack>
     </Content>
+    <!-- this is required for the nuget build to properly include the right files -->
+    <Content Include="..\Sparrow\bin\Release\netcoreapp2.1\Sparrow.dll">
```
iftahbe, line 42: @ppekrol I'm not sure about this one...

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1324,12 +1324,50 @@ private static void ValidateClientKeyUsages(X509Certificate2 certificate)
...
 
-        private static Func<HttpRequestMessage, X509Certificate2, X509Chain, SslPolicyErrors, bool>[] _serverCertificateCustomValidationCallback = Array.Empty<Func<HttpRequestMessage, X509Certificate2, X509Chain, SslPolicyErrors, bool>>();
+        private static RemoteCertificateValidationCallback[] _serverCertificateCustomValidationCallback = Array.Empty<RemoteCertificateValidationCallback>();
         private static ConcurrentDictionary<WeakReference<Action>, object> _liveClients = new ConcurrentDictionary<WeakReference<Action>, object>();
         private Action _updateHttpHandlerDelegate;// we need this to hold a reference to the action as long as the executer is alive
         private static readonly ReaderWriterLockSlim _serverCallbackRWLock = new ReaderWriterLockSlim();
 
+        // HttpClient and ClientWebSocket use certificate validation callbacks with different signatures.
+        // We need this translator for backward compatibility to allow the user to supply any of the two signatures.
+        public class CallbackTranslator
```
ppekrol: can be made private?

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1383,7 +1421,7 @@ private static void ForceUpdateOfAllHttpClients()
             }
         }
 
-        private static bool OnServerCertificateCustomValidationCallback(HttpRequestMessage msg, X509Certificate2 cert, X509Chain chain, SslPolicyErrors errors)
+        public static bool OnServerCertificateCustomValidationCallback(HttpRequestMessage msg, X509Certificate2 cert, X509Chain chain, SslPolicyErrors errors)
```
ppekrol: can be made internal?

ppekrol: we need to change signature to the one matching RemoteValidationCallback I think?

test/Tests.Infrastructure/TestBase.cs
```diff
@@ -85,7 +86,7 @@ static TestBase()
 
             var maxNumberOfConcurrentTests = Math.Max(ProcessorInfo.ProcessorCount / 2, 2);
 
-            RequestExecutor.ServerCertificateCustomValidationCallback += (message, certificate2, arg3, arg4) => true;
+            RequestExecutor.RemoteCertificateValidationCallback += (message, certificate2, arg3, arg4) => true;
```
ppekrol: message -> sender

src/Raven.Server/RavenServer.cs
```diff
@@ -350,19 +350,25 @@ private void RedirectsHttpTrafficToHttps()
...
                                 // It worked, let's register this callback globally in the RequestExecutor
-                                if(RequestExecutor.HasServerCertificateCustomValidationCallback == false)
-                                    RequestExecutor.ServerCertificateCustomValidationCallback += CertificateCallback;
+                                if (RequestExecutor.HasServerCertificateCustomValidationCallback == false)
+                                {
+                                    var callbackTranslator = new RequestExecutor.CallbackTranslator
+                                    {
+                                        Callback = CertificateCallback
+                                    };
+                                    RequestExecutor.RemoteCertificateValidationCallback += callbackTranslator.Translate;
```
ppekrol: Should not be translating here, change the signature of CertificateCallback

src/Raven.Client/Http/RequestExecutor.cs
iftahbe: No, we're also using it in DatabaseChanges

src/Raven.Client/Http/RequestExecutor.cs
iftahbe: We can but then we'll need to write a second translator to use in CreateHttpMessageHandler() where we assign the callback to the httpMessageHandler

src/Raven.Server/RavenServer.cs
iftahbe: yes

src/Raven.Client/Raven.Client.csproj
ppekrol: Should be good

PR #6969, opened 2018-07-08
[4.1] RavenDB-11516

aviv86:


test/SlowTests/Issues/RavenDB-9727.cs
```diff
@@ -30,7 +30,7 @@ public void Can_Load_with_Argument_that_has_String_Interpolation()
                 {
                     var query = from u in session.Query<User>()
                                 where u.LastName == "Garcia"
-                                let detail = session.Load<Detail>($"details/{u.DetailShortId}")
+                                let detail = session.Load<Detail>("details/"+u.DetailShortId)
```
maximburyak, line 33: @aviv86 Do we need this test, if we no longer test string interpolation anymore?

PR #6978, opened 2018-07-09
[v4.1] RavenDB-11523 List document ID value

fitzchak:


src/Raven.Studio/typescript/common/queryCompleter.ts
```diff
@@ -599,37 +599,25 @@ class queryCompleter {
                 if (lastKeyword.dividersCount === 2) {
                     return callback(null, queryCompleter.whereOperators);
                 }
-                
-                if (true) { // TODO: refactor this
+
+                if (lastKeyword.dividersCount === 3) {
```
ml054, line 612: can you add some comment why 3 is magic number here?

fitzchak: It's the number of spaces (or actually dividers - a divider can be more than a space/new line) from the last keyword. 
`where OrderedAt = com|` in this example last keyword is `where` and | is 3 contexts after it.

PR #6987, opened 2018-07-09
Raven db 10953 v4.0

talweiss1982:
+++++++++++
| DON'T MERGE |
+++++++++++
This PR includes the port of the missing attachment fix.
It also includes the start of trying to correct the handling of TCP operation versioning.
I think that the TCP versioning is something that is going to grow in complexity and we need to handle now before we go live with v4.1 since it will put constraints we will find hard to break later on.
In my simple approach, I'm planning to state for each server which protocol version he supports for each operation.
What I'm still missing (WIP) is the destination handling of an unmatching operation version.
It is possible that the destination is able to work with a lower protocol version but the source isn't able to handle the destination version E.G. replication from 4.0 to 4.1, there is no way 4.0 could talk with a 4.1 protocol but 4.1 could respond as if it was 4.0.
I have started to modify the accept TCP connection and the idea is that if I support your version I'll accept the connection and pass the accepted protocol version down the line to be handled by the "bundles".
This means that each "bundle" will have to deal with the protocol giving to it with the TCP stream, for incoming replication that means not responding with missing attachment to a v4.0 server but raising a flag instead. for outgoing replication that may mean not sending counters to a v4.0 server.
This is not a complete solution nor is this set in stone, your feedback is appreciated.

### DIFF
--- a/src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
+++ b/src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
@@ -210,7 +210,7 @@ namespace Raven.Client.Documents.Subscriptions
             }
         }
 
-        private int ReadServerRespondAndGetVersion(JsonOperationContext context, BlittableJsonTextWriter writer)
+        private int ReadServerRespondAndGetVersion(JsonOperationContext context, BlittableJsonTextWriter writer, Stream stream, string url)
         {
             //Reading reply from server
             using (var response = context.ReadForMemory(_stream, "Subscription/tcp-header-response"))
@@ -219,7 +219,7 @@ namespace Raven.Client.Documents.Subscriptions
                 switch (reply.Status)
                 {
                     case TcpConnectionStatus.Ok:
-                        break;
+                        return reply.Version;
                     case TcpConnectionStatus.AuthorizationFailed:
                         throw new AuthorizationException($"Cannot access database {_dbName} because " + reply.Message);
                     case TcpConnectionStatus.TcpVersionMismatch:
--- a/src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
+++ b/src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
@@ -550,7 +550,7 @@ namespace Raven.Client.Documents.Subscriptions
         }
 
         private DateTime? LastConnectionFailure;
-        private TcpFeaturesSupported _supportedFeatures;
+        private TcpConnectionHeaderMessage.SupportedFeatures _supportedFeatures;
 
         private void AssertLastConnectionFailure()
         {
--- a/src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
+++ b/src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
@@ -195,7 +195,7 @@ namespace Raven.Client.Documents.Subscriptions
                     Version = TcpConnectionHeaderMessage.SubscriptionTcpVersion,
                     ReadRespondAndGetVersion = ReadServerRespondAndGetVersion
                 };
-                _protocolVersion = TcpNegotiation.NegotiateProtocolVersion(context, _stream, parameters);
+                _supportedFeatures = TcpNegotiation.NegotiateProtocolVersion(context, _stream, parameters);
 
                 var options = Encodings.Utf8.GetBytes(JsonConvert.SerializeObject(_options));
                
@@ -550,7 +550,7 @@ namespace Raven.Client.Documents.Subscriptions
         }
 
         private DateTime? LastConnectionFailure;
-        private int _protocolVersion;
+        private TcpFeaturesSupported _supportedFeatures;
 
         private void AssertLastConnectionFailure()
         {
--- a/src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
+++ b/src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
@@ -23,6 +23,7 @@ using Raven.Client.ServerWide.Tcp;
 using Raven.Client.Util;
 using Sparrow;
 using Sparrow.Json;
+using Sparrow.Json.Parsing;
 using Sparrow.Logging;
 using Sparrow.Utils;
 
@@ -186,44 +187,18 @@ namespace Raven.Client.Documents.Subscriptions
                 _stream = await TcpUtils.WrapStreamWithSslAsync(_tcpClient, command.Result, _store.Certificate, requestExecutor.DefaultTimeout).ConfigureAwait(false);
 
                 var databaseName = _dbName ?? _store.Database;
-                var serializeObject = JsonConvert.SerializeObject(new TcpConnectionHeaderMessage
+
+                var parameters = new TcpNegotiateParamaters
                 {
+                    Database = databaseName,
                     Operation = TcpConnectionHeaderMessage.OperationTypes.Subscription,
-                    DatabaseName = databaseName,
-                    OperationVersion = TcpConnectionHeaderMessage.SubscriptionTcpVersion
-                });
-                var header = Encodings.Utf8.GetBytes(serializeObject);
+                    Version = TcpConnectionHeaderMessage.SubscriptionTcpVersion,
+                    ReadRespondAndGetVersion = ReadServerRespondAndGetVersion
+                };
+                _protocolVersion = TcpNegotiation.NegotiateProtocolVersion(context, _stream, parameters);
 
                 var options = Encodings.Utf8.GetBytes(JsonConvert.SerializeObject(_options));
-
-                await _stream.WriteAsync(header, 0, header.Length).ConfigureAwait(false);
-                await _stream.FlushAsync().ConfigureAwait(false);
-                //Reading reply from server
-                using (var response = context.ReadForMemory(_stream, "Subscription/tcp-header-response"))
-                {
-                    var reply = JsonDeserializationClient.TcpConnectionHeaderResponse(response);
-                    switch (reply.Status)
-                    {
-                        case TcpConnectionStatus.Ok:
-                            break;
-                        case TcpConnectionStatus.AuthorizationFailed:
-                            throw new AuthorizationException($"Cannot access database {databaseName} because " + reply.Message);
-                        case TcpConnectionStatus.TcpVersionMismatch:
-                            //Kindly request the server to drop the connection
-                            serializeObject = JsonConvert.SerializeObject(new TcpConnectionHeaderMessage
-                            {
-                                Operation = TcpConnectionHeaderMessage.OperationTypes.Drop,
-                                DatabaseName = databaseName,
-                                OperationVersion = TcpConnectionHeaderMessage.SubscriptionTcpVersion,
-                                Info = $"Couldn't agree on subscription tcp version ours:{TcpConnectionHeaderMessage.SubscriptionTcpVersion} theirs:{reply.Version}"
-                            });
-                            header = Encodings.Utf8.GetBytes(serializeObject);
-                            await _stream.WriteAsync(header, 0, header.Length).ConfigureAwait(false);
-                            await _stream.FlushAsync().ConfigureAwait(false);
-                            throw new InvalidOperationException($"Can't connect to database {databaseName} because: {reply.Message}");
-                    }
-
-                }
+               
                 await _stream.WriteAsync(options, 0, options.Length).ConfigureAwait(false);
 
                 await _stream.FlushAsync().ConfigureAwait(false);
@@ -233,6 +208,40 @@ namespace Raven.Client.Documents.Subscriptions
 
                 return _stream;
             }
+        }
+
+        private int ReadServerRespondAndGetVersion(JsonOperationContext context, BlittableJsonTextWriter writer)
+        {
+            //Reading reply from server
+            using (var response = context.ReadForMemory(_stream, "Subscription/tcp-header-response"))
+            {
+                var reply = JsonDeserializationClient.TcpConnectionHeaderResponse(response);
+                switch (reply.Status)
+                {
+                    case TcpConnectionStatus.Ok:
+                        break;
+                    case TcpConnectionStatus.AuthorizationFailed:
+                        throw new AuthorizationException($"Cannot access database {_dbName} because " + reply.Message);
+                    case TcpConnectionStatus.TcpVersionMismatch:
+                        if (reply.Version != -1)
+                        {
+                            return reply.Version;
+                        }
+                        //Kindly request the server to drop the connection
+                        context.Write(writer, new DynamicJsonValue
+                        {
+                            [nameof(TcpConnectionHeaderMessage.DatabaseName)] = _dbName,
+                            [nameof(TcpConnectionHeaderMessage.Operation)] = TcpConnectionHeaderMessage.OperationTypes.Drop.ToString(),
+                            [nameof(TcpConnectionHeaderMessage.OperationVersion)] = TcpConnectionHeaderMessage.GetOperationTcpVersion(TcpConnectionHeaderMessage.OperationTypes.Drop),
+                            [nameof(TcpConnectionHeaderMessage.Info)] = $"Couldn't agree on subscription tcp version ours:{TcpConnectionHeaderMessage.SubscriptionTcpVersion} theirs:{reply.Version}"
+                        });
+                        writer.Flush();
+                        throw new InvalidOperationException($"Can't connect to database {_dbName} because: {reply.Message}");
+                }
+                return reply.Version;
+            }
+
+            
         }
 
         private void AssertConnectionState(SubscriptionConnectionServerMessage connectionStatus)
@@ -541,6 +550,7 @@ namespace Raven.Client.Documents.Subscriptions
         }
 
         private DateTime? LastConnectionFailure;
+        private int _protocolVersion;
 
         private void AssertLastConnectionFailure()
         {
--- a/src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
+++ b/src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
@@ -27,21 +27,21 @@ namespace Raven.Client.ServerWide.Tcp
 
         public string Info { get; set; }
         
-        public static readonly int PingBaseLine4000 = -1;
-        public static readonly int NoneBaseLine4000 = -1;
-        public static readonly int DropBaseLine4000 = -2;
-        public static readonly int ClusterBaseLine4000 = 10;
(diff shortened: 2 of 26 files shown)


src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
```diff
@@ -29,10 +30,29 @@ public enum OperationTypes
...
 
+        public static bool OperationVersionSupported(OperationTypes operationType, int version)
+        {
+            if (_operationsToSupportedProtocolVersions.TryGetValue(operationType, out var supportedProtocols) == false)
+                return false;
+            return supportedProtocols.Contains(version);
+        }
+
+        private static readonly Dictionary<OperationTypes,HashSet<int>> _operationsToSupportedProtocolVersions
+        = new Dictionary<OperationTypes, HashSet<int>>
```
ayende: Probably better to use `List<int>` here, since order matters

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
```diff
@@ -463,15 +467,22 @@ private void WriteHeaderToRemotePeer()
             using (_database.DocumentsStorage.ContextPool.AllocateOperationContext(out DocumentsOperationContext documentsContext))
             using (var writer = new BlittableJsonTextWriter(documentsContext, _stream))
             {
-                documentsContext.Write(writer, new DynamicJsonValue
+                WriteHeaderInternal(documentsContext, writer, TcpConnectionHeaderMessage.ReplicationTcpVersion);
+                var destinationProtocol = ReadHeaderResponseAndThrowIfUnAuthorized(documentsContext, writer);
+                if (destinationProtocol != TcpConnectionHeaderMessage.ReplicationTcpVersion)
```
ayende: Here we need to do this in a loop.
The logic goes like this:

10 Src sends its highest version. If dest accepts it, great. 
20 If not, the src sends its highest version. If the dest knows of it, great.
30 If not, src sends its next highest version, repeat until there is a version that both of them understand.

The idea is that we'll use the highest version that both endpoints understand.
Src will try them in order.
Dest will check if it understand the protocol version and _always_ reply with its own preferred version.

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
```diff
@@ -529,8 +556,17 @@ private void ReadHeaderResponseAndThrowIfUnAuthorized(DocumentsOperationContext
                         throw new InvalidOperationException($"{Destination.FromString()} replied with unknown status {headerResponse.Status}, message:{headerResponse.Message}");
                 }
             }
+
+            return TcpConnectionHeaderMessage.ReplicationTcpVersion;
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private bool CanCommunicateWithReplicationProtocol(int headerResponseVersion)
```
ayende: I don't think this is helpful. You are going to need to check _what properties_ are valid here.

ayende (changes requested):

src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
talweiss1982: Yes, i agree.

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
talweiss1982: Yes, this needs to change according to our talk yesterday.
This ping pong will be handled at the same level on the accepting side as now but each sender will have to modify the logic to support this. This will actually not break anything since we already allow 3 streaks and since we don't have more than 3 version for any of our protocol increasing the retry won't break anything :)
I'll add the agreed upon version with the TCP stream and each logic will have to behave accordingly.

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
talweiss1982: This is going to move, this is wrong and not according to what we talked about.

src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
```diff
@@ -186,44 +187,18 @@ private Task Run(CancellationToken ct)
...
+                    Database = databaseName,
                     Operation = TcpConnectionHeaderMessage.OperationTypes.Subscription,
-                    DatabaseName = databaseName,
-                    OperationVersion = TcpConnectionHeaderMessage.SubscriptionTcpVersion
-                });
-                var header = Encodings.Utf8.GetBytes(serializeObject);
+                    Version = TcpConnectionHeaderMessage.SubscriptionTcpVersion,
+                    ReadRespondAndGetVersion = ReadServerRespondAndGetVersion
+                };
+                _protocolVersion = TcpNegotiation.NegotiateProtocolVersion(context, _stream, parameters);
```
ayende: `NegotiateProtocolVersion` should return an object, not an int.
That will be basically full of properties that say what operations are supported.

ayende: Example:
* `31 - BaseLine40` - 4.0
* `32 - BaseLine40, Counters` - 4.1
* `33 - BaseLine40, AttachmentMissing` - 4.0
* `34 - BaseLine40, AttachmentMissing, Counters` - 4.1

src/Raven.Client/ServerWide/Tcp/TcpNegotiation.cs
```diff
@@ -0,0 +1,47 @@
...
+                        [nameof(TcpConnectionHeaderMessage.Operation)] = parameters.Operation.ToString(),
+                        [nameof(TcpConnectionHeaderMessage.SourceNodeTag)] = parameters.NodeTag,
+                        [nameof(TcpConnectionHeaderMessage.OperationVersion)] = currentVersion
+                    });
+                    writer.Flush();
+                    var version = parameters.ReadRespondAndGetVersion(documentsContext, writer);
+                    var (supported, prevSupported) = TcpConnectionHeaderMessage.OperationVersionSupported(parameters.Operation, version);
+                    if (supported)
+                        return version;
+                    if (prevSupported == -1)
```
ayende: Just to note, you have some `-2` there that might cause issues

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -1147,5 +1240,19 @@ public void Dispose()
                 }
             }
         }
+
+        public bool SupportMissingAttachments
```
ayende: Need to turn that into an options object, and make sure its use is consistent.

ayende (changes requested):

src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
talweiss1982: I think we can skip 34 since nobody is using v4.1 yet

ayende: @talweiss1982 Yes, that was an example to explain the concept.

src/Raven.Client/ServerWide/Tcp/TcpNegotiation.cs
```diff
@@ -0,0 +1,47 @@
...
+    public class TcpNegotiation
+    {
+        public static int NegotiateProtocolVersion(JsonOperationContext documentsContext, Stream stream, TcpNegotiateParamaters parameters)
+        {
+            using (var writer = new BlittableJsonTextWriter(documentsContext, stream))
+            {
+                var currentVersion = parameters.Version;
+                while (true)
+                {
+                    documentsContext.Write(writer, new DynamicJsonValue
```
ppekrol, line 21: I think it would be better to 'invert' the responsibility and send list of all supported protocol versions for this particular operations so destination can make a decision and send back the version that should be used.

src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
```diff
@@ -25,14 +26,44 @@ public enum OperationTypes
...
+        }
+
+        private static readonly Dictionary<OperationTypes, List<int>> _operationsToSupportedProtocolVersions
+        = new Dictionary<OperationTypes, List<int>>
+            {
+                [OperationTypes.Ping] = new List<int> { -1},
+                [OperationTypes.None] = new List<int> { -1 },
+                [OperationTypes.Drop] = new List<int> { -2 },
+                [OperationTypes.Subscription] = new List<int> { 40 },
+                [OperationTypes.Replication] = new List<int> { 33, 31 },
```
ppekrol: I suggest bumping those to:
- 40xxx for 4.0 supported protocols
- 41xxx for 4.1 supported protocols

This way it would be much more clear which protocol is related to which 4.x version.

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -144,7 +152,7 @@ private void ReceiveReplicationBatches()
 
                             using (var msg = interruptibleRead.ParseToMemory(
                                 _replicationFromAnotherSource,
-                                "IncomingReplication/read-message",
+                                "IncomingReplicationStr/read-message",
```
ppekrol, line 155: You wanted to probably use `$"{IncomingReplicationStr}/read-message"` here

src/Raven.Client/ServerWide/Tcp/TcpNegotiation.cs
talweiss1982: Indeed

src/Raven.Client/ServerWide/Tcp/TcpNegotiation.cs
ayende: Take into account, we have `4.0` deployed already. I don't think we can make it work with that.

talweiss1982: it is both sides responsibility since you can't know who know better

src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
talweiss1982: Trying to bind protocol version to server version will only be confusing and it is not related. Moreover, this will require to add workaround code just to deal with backward compatibility with v4.0

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
talweiss1982: This is the original message I have not modified any of the internal logic.

src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
ppekrol: Not binding it to a server version. Just making clear which protocol is supported where.

Image a list of protocols like that:
`32, 33, 34, 35, 36, 37, 38, 39, 40, 41, 42, 43`

Which ones are supported by 4.0, which ones are supported by 4.1 and which ones are supported by 4.2?

If this list would look like:

`40032, 40033, 41001, 42010, 41002, 40035, 42015, 41032`

Then it would be more clear which protocol is from which version.

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -1147,5 +1240,19 @@ public void Dispose()
...
             }
         }
+
+        public bool SupportMissingAttachments
+        {
+            get
+            {
+                switch (_protocolVersion)
+                {
+                    case 33:
```
ppekrol: If we will 'split' the versioning then we can assume that all `_protocolVersion >= 33 && _protocolVersion <= 40999` will support this?

ayende: We are better off with checking only explicitly known version, not ranges

src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
talweiss1982: I agree it is not clear, please review the pushed modifications.

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
ppekrol: Then we should put everything related to those versions in one place (one file), so in the future we would have one place/file to change. Preferably method that returns a class with set of bools for supported features.

src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
```diff
@@ -52,17 +61,17 @@ public static (bool Supported, int PrevSupported) OperationVersionSupported(Oper
             return (false, prev);
         }
 
-        private static readonly Dictionary<OperationTypes, List<int>> _operationsToSupportedProtocolVersions
+        private static readonly Dictionary<OperationTypes, List<int>> OperationsToSupportedProtocolVersions
```
ayende: `Dictionary<OperationTypes, List<SupportFeatures>>`

Where:

````
public class SupportedFeatures
{
     public bool BaseLine40, Counters, MissingAttachments;
     public int ProtocolVersion;
}
````

And then init this using:

```
[Operation.Replication] = new List<SupportedOperations>
{
   new SupportedOperations{ ProtocolVersion = 32, BaseLine40 = true },
   new SupportedOperations{ ProtocolVersion = 33, BaseLine40 = true , Counters = true},
   new SupportedOperations{ ProtocolVersion = 34, BaseLine40 = true, MissingAttachments = true },
   new SupportedOperations{ ProtocolVersion = 35, BaseLine40 = true, Counters = true , MissingAttachments = true},
}
```

This keep all the knowledge in one location

File: src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs [hunk 12]
```diff
@@ -26,17 +26,26 @@ public enum OperationTypes
...
-        public static readonly int SubscriptionTcpVersion = 40;
-        public static readonly int TestConnectionTcpVersion = 50;
+        
+        public static readonly int PingBaseLine40 = -1;
+        public static readonly int NoneBaseLine40 = -1;
+        public static readonly int DropBaseLine40 = -2;
+        public static readonly int ClusterBaseLine40 = 10;
+        public static readonly int HeartbeatsBaseLine40 = 20;
+        public static readonly int ReplicationBaseLine40 = 31;
+        public static readonly int ReplicationAttachmentMissing = 33;
```
ppekrol: Bump it to 40000?

src/Raven.Client/ServerWide/Tcp/TcpNegotiation.cs
```diff
@@ -25,16 +25,50 @@ public static int NegotiateProtocolVersion(JsonOperationContext documentsContext
...
+        private int _version;
+        private TcpConnectionHeaderMessage.OperationTypes _type;
+
+        public TcpFeaturesSupported(TcpConnectionHeaderMessage.OperationTypes type, int protocolVersion)
+        {
+            _version = protocolVersion;
+            _type = type;
+        }
+
+        public bool IsMissingAttachmentSupported {
```
ppekrol: Wrap it inside some class that can be access via 'Replication' property. It would be more clear what bool is for what feature. e.g. `features.Replication.SupportsMissingAttachments` ?

src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
talweiss1982: your example is ReplicationSupportedOperations (which is currently the only tcp operation we have versions for).
We need to make sure we don't mix protocol types...
I want to have a single class to handle supported features and not have to have a base class and derived classes...

src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
```diff
@@ -27,30 +27,149 @@ public enum OperationTypes
...
         
-        public static readonly int PingBaseLine40 = -1;
-        public static readonly int NoneBaseLine40 = -1;
-        public static readonly int DropBaseLine40 = -2;
-        public static readonly int ClusterBaseLine40 = 10;
-        public static readonly int HeartbeatsBaseLine40 = 20;
-        public static readonly int ReplicationBaseLine40 = 31;
+        public static readonly int PingBaseLine4000 = -1;
+        public static readonly int NoneBaseLine4000 = -1;
+        public static readonly int DropBaseLine4000 = -2;
```
ayende: Why the negative numbers here?

ayende (approved):

src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
talweiss1982: backward compatibility :(

src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
```diff
@@ -25,13 +26,160 @@ public enum OperationTypes
...
 
         public string Info { get; set; }
+        
+        public static readonly int PingBaseLine4000 = -1;
+        public static readonly int NoneBaseLine4000 = -1;
+        public static readonly int DropBaseLine4000 = -2;
+        public static readonly int ClusterBaseLine4000 = 10;
+        public static readonly int HeartbeatsBaseLine4000 = 20;
+        public static readonly int ReplicationBaseLine4000 = 31;
+        public static readonly int ReplicationAttachmentMissing = 33;
```
ppekrol: What about bumping that?

ppekrol: @talweiss1982 ping

talweiss1982:
This break competabilty I'm not going to break something and then add a workaround for it without good enough justification.

ppekrol:
@talweiss1982 Haven't you added this number in this PR?

talweiss1982:
Nope, this has been around since RTM.
Only replication was actually doing something with it. V4.0 is running version 31.

talweiss1982:
Note that those version are per operation and not per server version

ppekrol:
I'm aware of that, so changing this `33` protocol version to `40000` will give you instant information that this is supported in 4.0+

talweiss1982:
@ppekrol bumping the new values to 4XXX is okay with me but we also need to tell the difference between replication and clustering this is why the version was not identical in the first place (was different so people won't do a mistake and it will show the wrong error)
what do you think about:
40303 instead of 33 where 40 is for v40 3 is for replication and 03 is for the actual version?

ppekrol (approved):

PR #6988, opened 2018-07-10
[4.0] RavenDB-11533 register OnServerCertificateCustomValidationCallback always

ppekrol:


src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1241,18 +1220,8 @@ public static HttpClientHandler CreateHttpMessageHandler(X509Certificate2 certif
...
-                {
-                    throw new PlatformNotSupportedException("On Mac OSX, is it not possible to register to ServerCertificateCustomValidationCallback because of https://github.com/dotnet/corefx/issues/9728");
-                }
-                else
-                {
-                    httpMessageHandler.ServerCertificateCustomValidationCallback += OnServerCertificateCustomValidationCallback;
-                }
-            }
+            if (PlatformDetails.RunningOnMacOsx == false)
+                httpMessageHandler.ServerCertificateCustomValidationCallback += OnServerCertificateCustomValidationCallback;
```
ayende: Needs error handling here. 
In particular, need to see if an exception is raised here. If so, need to raise it up if there are any registered callbacks and if there aren't, remember it and raise it when someone try to register.

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1361,53 +1337,24 @@ public bool Translate(object sender, X509Certificate cert, X509Chain chain, SslP
...
-                    ForceUpdateOfAllHttpClients();
-                }
-                finally
-                {
-                    _serverCallbackRWLock.ExitWriteLock();
-                }
-            }
-        }
+                if (PlatformDetails.RunningOnMacOsx)
+                    throw new PlatformNotSupportedException("On Mac OSX, is it not possible to register to RemoteCertificateValidationCallback because of https://github.com/dotnet/corefx/issues/9728");
```
ayende: I don't like this approach. It means that when OSX does support it, we'll have to do something special for this.

ayende (changes requested):

src/Raven.Client/Http/RequestExecutor.cs
ppekrol: Can you check now?

src/Raven.Client/Http/RequestExecutor.cs
ppekrol: Can you check now?

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1337,8 +1352,8 @@ public bool Translate(object sender, X509Certificate cert, X509Chain chain, SslP
         {
             add
             {
-                if (PlatformDetails.RunningOnMacOsx)
-                    throw new PlatformNotSupportedException("On Mac OSX, is it not possible to register to RemoteCertificateValidationCallback because of https://github.com/dotnet/corefx/issues/9728");
+                if (ServerCertificateCustomValidationCallbackRegistrationException != null)
+                    throw ServerCertificateCustomValidationCallbackRegistrationException;
```
ayende: Please wrap it. I'm not sure whatever this can cause issue, but throwing the same exception concurrently sounds like potential race condition

ppekrol: Wrapped :)

ayende (approved):

PR #6997, opened 2018-07-10
[4.0] RavenDB-11444 Wake the database for backups

karmeli87:
Always wake up the database for a full backup.
But for incremental, wake the database only if changes were made.

- Prevent the database to unload during the actual backup process.
- Unloading a database due to idleness will generate a wakeup timer, dependent on the nearest backup task.

PR #6998, opened 2018-07-10
[4.1] RavenDB-11444 Wake the database for backups

karmeli87:
Always wake up the database for a full backup.
But for incremental, wake the database only if changes were made.

- Prevent the database to unload during the actual backup process.
- Unloading a database due to idleness will generate a wakeup timer, dependent on the nearest backup task.

### DIFF
--- a/src/Raven.Server/Documents/DatabasesLandlord.cs
+++ b/src/Raven.Server/Documents/DatabasesLandlord.cs
@@ -32,6 +32,7 @@ namespace Raven.Server.Documents
         private readonly ReaderWriterLockSlim _disposing = new ReaderWriterLockSlim(LockRecursionPolicy.SupportsRecursion);
         public readonly ConcurrentDictionary<StringSegment, DateTime> LastRecentlyUsed =
             new ConcurrentDictionary<StringSegment, DateTime>(CaseInsensitiveStringSegmentEqualityComparer.Instance);
+        private readonly ConcurrentDictionary<string, Timer> _wakeupTimers = new ConcurrentDictionary<string, Timer>();
 
         public readonly ResourceCache<DocumentDatabase> DatabasesCache = new ResourceCache<DocumentDatabase>();
         private readonly TimeSpan _concurrentDatabaseLoadTimeout;
@@ -401,6 +402,18 @@ namespace Raven.Server.Documents
             {
                 var exceptionAggregator = new ExceptionAggregator(_logger, "Failure to dispose landlord");
 
+                // we don't want to wake up database during dispose.
+                var handles = new List<WaitHandle>();
+                foreach (var timer in _wakeupTimers.Values)
+                {
+                    var handle = new ManualResetEvent(false);
+                    timer.Dispose(handle);
+                    handles.Add(handle);
+                }
+
+                if (handles.Count > 0)
+                    WaitHandle.WaitAll(handles.ToArray());
+
                 // shut down all databases in parallel, avoid having to wait for each one
                 Parallel.ForEach(DatabasesCache.Values, new ParallelOptions
                 {
@@ -464,6 +477,11 @@ namespace Raven.Server.Documents
         {
             try
             {
+                if (_wakeupTimers.TryRemove(databaseName.Value, out var timer))
+                {
+                    timer.Dispose();
+                }
+
                 if (_disposing.TryEnterReadLock(0) == false)
                     ThrowServerIsBeingDisposed(databaseName);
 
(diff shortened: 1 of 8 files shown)


src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -401,6 +402,14 @@ public void Dispose()
             {
                 var exceptionAggregator = new ExceptionAggregator(_logger, "Failure to dispose landlord");
 
+                // we don't want to wake up database during dispose.
+                Parallel.ForEach(_wakeupTimers, (timer) =>
```
maximburyak: Timer.Dispose is blocking, I think that you should call it serially, have an array of all MREs and then call WaitHandle.WaitAll(mreArray);

src/Raven.Server/Documents/PeriodicBackup/PeriodicBackupRunner.cs
```diff
@@ -493,12 +536,11 @@ public void UpdateConfigurations(DatabaseRecord databaseRecord)
 
             if (databaseRecord.PeriodicBackups == null)
             {
-                foreach (var periodicBackup in _periodicBackups)
+                Parallel.ForEach(_periodicBackups, (periodicBackup) =>
```
maximburyak: In that case, if we had a DisposeOnce implementation with an async Dispose, we could write something much nicer

karmeli87:
I hope that all the nice comments from this PR and from https://github.com/ravendb/ravendb/pull/6997 are taken care of :-)

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -776,10 +794,48 @@ public enum ClusterDatabaseChangeType
...
+            if (_wakeupTimers.TryRemove(name, out var timer))
+            {
+                timer.Dispose();
+            }
+
+            if(wakeup.HasValue == false || wakeup.Value == DateTime.MaxValue)
+                return true;
+
+            // if we have a small value or even a negative one, simply don't dispose the database.
+            dueTime = (int)(wakeup - DateTime.Now).Value.TotalMilliseconds;
```
ppekrol, line 874: Should be SystemTime.UtcNow?

ppekrol (approved):

src/Raven.Server/Documents/DatabasesLandlord.cs
karmeli87: the `wakeup` is local time

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -401,6 +402,18 @@ public void Dispose()
...
+                var handles = new List<WaitHandle>();
+                foreach (var timer in _wakeupTimers.Values)
+                {
+                    var handle = new ManualResetEvent(false);
+                    timer.Dispose(handle);
+                    handles.Add(handle);
+                }
+
+                if (handles.Count > 0)
+                    WaitHandle.WaitAll(handles.ToArray());
```
ayende: Is it possible that we have more than 64 items here?

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -776,10 +794,48 @@ public enum ClusterDatabaseChangeType
...
+        }
+
+        private void ScheduleDatabaseWakeup(string name, long milliSeconds)
+        {
+            if (milliSeconds == 0)
+                return;
+
+            _wakeupTimers.TryAdd(name, new Timer(_ =>
+            {
+                if (_disposing.IsWriteLockHeld)
```
ayende: I don't like this. Better to handle this _inside_ the read lock.

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -776,10 +794,48 @@ public enum ClusterDatabaseChangeType
...
+            {
+                timer.Dispose();
+            }
+
+            if(wakeup.HasValue == false || wakeup.Value == DateTime.MaxValue)
+                return true;
+
+            // if we have a small value or even a negative one, simply don't dispose the database.
+            dueTime = (int)(wakeup - DateTime.Now).Value.TotalMilliseconds;
+            return dueTime > 1000;
```
ayende: Cost of loading / starting the db is quite high.
Let's skip this if we have a backup scheduled for the next five _minutes_

src/Raven.Server/Documents/PeriodicBackup/PeriodicBackupRunner.cs
```diff
@@ -99,6 +99,26 @@ public PeriodicBackupRunner(DocumentDatabase database, ServerStore serverStore)
             return taskStatus == TaskStatus.Disabled ? null : GetNextBackupDetails(configuration, backupStatus, skipErrorLog: true);
         }
 
+        private DateTime? GetNextWakeupTime(long lastEtag, PeriodicBackupConfiguration configuration, PeriodicBackupStatus backupStatus)
+        {
+            // we will always wake up the database for a full backup.
+            // but for incremental we will wake the database only if there were changes made.
+            
+            var now = SystemTime.UtcNow;
```
ayende, line 107: This is wrong, back up times are always local, not UTC

ayende (changes requested):

src/Raven.Server/Documents/DatabasesLandlord.cs
karmeli87: Theoretically, if we have more than 64 backup tasks, then yes.

src/Raven.Server/Documents/PeriodicBackup/PeriodicBackupRunner.cs
karmeli87: We have to keep this time as utc (cluster wide) and only later, on the node itself we translate it to local.

src/Raven.Server/Documents/DatabasesLandlord.cs
ayende: IIRC, `WaitAll` can accept a maximum number, need to handle this here.

src/Raven.Server/Documents/PeriodicBackup/PeriodicBackupRunner.cs
ayende: I don't follow, this is used only on the local node.

karmeli87: The `backupStatus` is in UTC

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -811,10 +828,22 @@ private void ScheduleDatabaseWakeup(string name, long milliSeconds)
 
             _wakeupTimers.TryAdd(name, new Timer(_ =>
             {
-                if (_disposing.IsWriteLockHeld)
-                    return;
+                _disposing.EnterReadLock();
```
ayende: This line may throw, need to have that in the `try`

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -811,10 +828,22 @@ private void ScheduleDatabaseWakeup(string name, long milliSeconds)
...
-                TryGetOrCreateResourceStore(name);
+                    TryGetOrCreateResourceStore(name);
+                }
+                catch (ObjectDisposedException)
+                {
+                    // expected 
+                }
+                finally
+                {
+                    _disposing.ExitReadLock();
```
ayende: This line may also throw, if the `EnterReadLock` threw

ayende (changes requested):

PR #7001, opened 2018-07-10
RavenDB-11485 Add ServerUrl to ServerOptions

IdanHaim:


src/Raven.Embedded/RavenServerRunner.cs
```diff
@@ -34,24 +34,30 @@ public static Process Run(ServerOptions options)
 
             if (options.Security != null)
             {
-                options.CommandLineArgs.Add("--ServerUrl=https://127.0.0.1:0");
+                if (string.IsNullOrWhiteSpace(options.ServerUrl))
+                    options.ServerUrl = "https://127.0.0.1:0";
+
+                options.CommandLineArgs.Add($"--ServerUrl={options.ServerUrl}");
```
ayende: This can be moved outside the `if` now

ayende (changes requested):

PR #7004, opened 2018-07-10
RavenDB-11528 Return terms for field in dynamic index when there is a…

fitzchak:
…n index to answer it

### DIFF
--- a/src/Raven.Server/Documents/Handlers/IndexHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/IndexHandler.cs
@@ -17,8 +17,9 @@ using Raven.Server.Documents.Indexes.Debugging;
 using Raven.Server.Documents.Indexes.Errors;
 using Raven.Server.Documents.Indexes.MapReduce.Static;
 using Raven.Server.Documents.Indexes.Static;
-using Raven.Server.Documents.Indexes.Static.Extensions;
 using Raven.Server.Documents.Patch;
+using Raven.Server.Documents.Queries;
+using Raven.Server.Documents.Queries.Dynamic;
 using Raven.Server.Json;
 using Raven.Server.Routing;
 using Raven.Server.ServerWide;
@@ -600,14 +601,15 @@ namespace Raven.Server.Documents.Handlers
         [RavenAction("/databases/*/indexes/terms", "GET", AuthorizationStatus.ValidUser)]
         public Task Terms()
         {
-            var name = GetQueryStringValueAndAssertIfSingleAndNotEmpty("name");
             var field = GetQueryStringValueAndAssertIfSingleAndNotEmpty("field");
-            var fromValue = GetStringQueryString("fromValue", required: false);
 
             using (var token = CreateTimeLimitedOperationToken())
             using (Database.DocumentsStorage.ContextPool.AllocateOperationContext(out DocumentsOperationContext context))
             using (context.OpenReadTransaction())
             {
+                var name = GetIndexNameFromCollectionAndField(field, context) ?? GetQueryStringValueAndAssertIfSingleAndNotEmpty("name");
+
+                var fromValue = GetStringQueryString("fromValue", required: false);
                 var existingResultEtag = GetLongFromHeaders("If-None-Match");
 
                 var result = Database.QueryRunner.ExecuteGetTermsQuery(name, field, fromValue, existingResultEtag, GetPageSize(), context, token, out var index);
@@ -628,7 +630,7 @@ namespace Raven.Server.Documents.Handlers
                         field.EndsWith("__maxY"))
                     {
                         if (index.Definition.IndexFields != null &&
-                        index.Definition.IndexFields.TryGetValue(field.Substring(0, field.Length - 6), out var indexField) == true)
+                            index.Definition.IndexFields.TryGetValue(field.Substring(0, field.Length - 6), out var indexField) == true)
                         {
                             if (indexField.Spatial?.Strategy == Client.Documents.Indexes.Spatial.SpatialSearchStrategy.BoundingBox)
                             {
@@ -640,6 +642,7 @@ namespace Raven.Server.Documents.Handlers
                                     var num = Lucene.Net.Util.NumericUtils.PrefixCodedToDouble(item);
                                     readableTerms.Add(NumberUtil.NumberToString(num));
                                 }
+
                                 result.Terms = readableTerms;
                             }
                         }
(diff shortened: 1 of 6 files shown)


src/Raven.Server/Documents/Handlers/IndexHandler.cs
```diff
@@ -652,6 +660,19 @@ public Task Terms()
             }
         }
 
+        private string GetIndexNameFromCollectionAndField(string field, DocumentsOperationContext context)
+        {
+            var collection = GetStringQueryString("collection", false);
```
ppekrol, line 660: I think that you can use  `GetQueryStringValueAndAssertIfSingleAndNotEmpty` ?

src/Raven.Server/Documents/Handlers/IndexHandler.cs
```diff
@@ -652,6 +660,19 @@ public Task Terms()
...
+            if (string.IsNullOrEmpty(collection))
+                ThrowRequiredMember("name");
+            var query = new IndexQueryServerSide(new QueryMetadata(collection, field));
+            var dynamicQueryToIndex = new DynamicQueryToIndexMatcher(Database.IndexStore);
+            var match = dynamicQueryToIndex.Match(DynamicQueryMapping.Create(query), context);
+            if (match.MatchType == DynamicQueryMatchType.Failure)
```
ppekrol: I think that it needs to be a complete match?

src/Raven.Server/Web/RequestHandler.cs
```diff
@@ -252,7 +252,7 @@ protected string GetStringQueryString(string name, bool required = true)
             return val[0];
         }
 
-        private static void ThrowRequiredMember(string name)
+        protected static void ThrowRequiredMember(string name)
```
ppekrol: Can be made private after using `GetQueryStringValueAndAssertIfSingleAndNotEmpty`

src/Raven.Server/Documents/Handlers/IndexHandler.cs
```diff
@@ -600,14 +601,20 @@ public Task GetErrors()
...
-            var name = GetQueryStringValueAndAssertIfSingleAndNotEmpty("name");
             var field = GetQueryStringValueAndAssertIfSingleAndNotEmpty("field");
-            var fromValue = GetStringQueryString("fromValue", required: false);
 
             using (var token = CreateTimeLimitedOperationToken())
             using (Database.DocumentsStorage.ContextPool.AllocateOperationContext(out DocumentsOperationContext context))
             using (context.OpenReadTransaction())
             {
+                var name = GetStringQueryString("name", false) ?? GetIndexNameFromCollectionAndField(field, context);
+                if (name == null)
```
ppekrol: This is a breaking change. We were throwing in that case.

Two exceptions here:
1. The one from GetQueryStringValueAndAssertIfSingleAndNotEmpty
2. When 'name' is specified (or 'collection') and we do not have an index then we are throwing IndexDoesNotExistException

src/Raven.Server/Documents/Handlers/IndexHandler.cs
```diff
@@ -652,6 +660,19 @@ public Task Terms()
             }
         }
 
+        private string GetIndexNameFromCollectionAndField(string field, DocumentsOperationContext context)
+        {
+            var collection = GetStringQueryString("collection", false);
+            if (string.IsNullOrEmpty(collection))
+                ThrowRequiredMember("name");
+            var query = new IndexQueryServerSide(new QueryMetadata(collection, field));
```
ppekrol: Maybe pass the the 'old' constructor `from <collection> select <field>` and parse it using the same flow?

ppekrol (changes requested):

ml054 (approved):

src/Raven.Server/Documents/Handlers/IndexHandler.cs
fitzchak: This is logic is like this: if `name` parameter is not there, than we check for the `collection` parameter, and if it isn't there than we throw that the `name` parameter is mandatory.

src/Raven.Server/Documents/Handlers/IndexHandler.cs
fitzchak: The behavior is observed. We still throwing for the `name` parameter. The only case that it can reach the `null` case is when there is no index for the specified collection and field. In that case I return 404 instead of throwing.

src/Raven.Server/Documents/Handlers/IndexHandler.cs
fitzchak: I can change it, yes. But it would do more work than required.

src/Raven.Server/Web/RequestHandler.cs
fitzchak: See other comments.

src/Raven.Server/Documents/Handlers/IndexHandler.cs
fitzchak: Why? In partial match I see we clone the index... so it means the index does not have the requested field?

ayende: In particular, I don't think you _can_ get partial here.
You are only ever sending a single field, after all.

ppekrol: @fitzchak so it means that you need to have a complete match, right?

fitzchak: I'll change this to `Complete` and `CompleteButIdle` than.

src/Raven.Server/Documents/Handlers/IndexHandler.cs
ppekrol: First check is 'collection' is specified. If not use 'GetQueryStringValueAndAssertIfSingleAndNotEmpty' on 'name'.

fitzchak: ok 💯

src/Raven.Server/Documents/Handlers/IndexHandler.cs
ppekrol: True, but from the other side, at some point we will add additional logic in first ctor, and 2nd one will be omitted. We can leave as is, but I thought that the change is not that complex?

src/Raven.Server/Documents/Handlers/IndexHandler.cs
ppekrol: Check my other comment for point 1. and I'm good with point 2.

src/Raven.Server/Documents/Handlers/IndexHandler.cs
fitzchak: Done

src/Raven.Server/Documents/Handlers/IndexHandler.cs
fitzchak: OK, I'll throw IndexDoesNotExistException instead of returning 404.

src/Raven.Server/Documents/Handlers/IndexHandler.cs
fitzchak: It is a bit more complex but can be done. I'll need to create an empty blittable for the `parameters` and an cache key, which I'll look how to generate.

fitzchak: Done, the query won't be cached.

PR #7012, opened 2018-07-11
[4.0] RavenDB-11480 : RQL load with index stored fields

aviv86:


src/Raven.Server/Documents/Queries/FieldsToFetch.cs
```diff
@@ -135,7 +135,15 @@ ref ignored
                 }
             }
 
-            var extract = indexDefinition.MapFields.TryGetValue(selectFieldName, out var value) && value.Storage == FieldStorage.Yes;
+            var key = selectFieldName.Value.Length == 0 &&
```
arekpalinski: Can we _try_ to put this condition into a method with some meaningful name? Something that would describe what happened that all those conditions are satisfied.

src/Raven.Server/Documents/Queries/Results/QueryResultRetrieverBase.cs
```diff
@@ -399,6 +399,13 @@ private bool TryGetValue(FieldsToFetch.FieldToFetch fieldToFetch, Document docum
                     }
                 }
 
+                else if (fieldToFetch.CanExtractFromIndex)
+                {
+                    var field = luceneDoc.GetField(fieldToFetch.QueryField.SourceAlias);
```
arekpalinski, line 404: Maybe check it `field == null` and throw exception then (unless we are guaranteed that the field exists at this point)

aviv86: yes we know we that the field exists at this point

src/Raven.Server/Documents/Queries/FieldsToFetch.cs
aviv86: done

src/Raven.Server/Documents/Queries/Results/QueryResultRetrieverBase.cs
ayende: How do you know that the field must be here?

arekpalinski: @aviv86 ping

aviv86: Actually the field can be missing if the index has dynamic fields. fixing that

PR #7016, opened 2018-07-11
[4.0] RavenDB-10799 workaround for multiple localKeyId in certificate

iftahbe:
Workaround for https://github.com/dotnet/corefx/issues/30946


ppekrol (approved):

src/Raven.Server/ServerWide/PkcsStore.cs
```diff
@@ -0,0 +1,488 @@
...
+namespace Raven.Server.ServerWide
+{
+    // Workaround for https://github.com/dotnet/corefx/issues/30946
+    // This class is a partial copy of BouncyCastle's Pkcs12Store which doesn't throw the exception: "attempt to add existing attribute with different value".
+
+    // Explanation: Certificates which were exported in Linux were changed to include a multiple bag attribute (localKeyId).
+    // When using BouncyCastle's Pkcs12Store to load the cert, we got an exception: "attempt to add existing attribute with different value",
+    // but we don't care about that. We just need to extract the private key when using Load().
+
+    public class PkcsStore
```
ayende: Let's give it a clearer name, `PkcsStoreWorkaroundFor30946` or something like that.

ayende (changes requested):
Beyond the class name, looks good

src/Raven.Server/ServerWide/PkcsStore.cs
ayende: Also, make this class private

ayende (approved):

PR #7017, opened 2018-07-11
[4.1] RavenDB-10799 workaround for multiple localKeyId in certificate

iftahbe:


src/Raven.Server/ServerWide/SecretProtection.cs
```diff
@@ -683,5 +693,476 @@ public static void ValidateKeyUsages(string source, X509Certificate2 loadedCerti
                 throw new EncryptionException(msg);
             }
         }
+
+        public class PkcsStoreWorkaroundFor30946
```
ayende: This is public, not private?

iftahbe: Sorry about that. Fixed.

PR #7020, opened 2018-07-11
[4.0] RavenDB-11124 RavenTestBase fix

karmeli87:
- Add cancellation token to various WaitForState in Rachis.
- Don't change the urls of the store during the test.

### DIFF
--- a/test/Tests.Infrastructure/RavenTestBase.cs
+++ b/test/Tests.Infrastructure/RavenTestBase.cs
@@ -175,7 +175,6 @@ namespace FastTests
                         }
 
                         Assert.True(result.RaftCommandIndex > 0); //sanity check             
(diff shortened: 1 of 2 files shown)


test/Tests.Infrastructure/RavenTestBase.cs
```diff
@@ -175,7 +175,6 @@ protected DocumentStore GetDocumentStore(Options options = null, [CallerMemberNa
                         }
 
                         Assert.True(result.RaftCommandIndex > 0); //sanity check             
-                        store.Urls = result.NodesAddedTo.SelectMany(UseFiddler).ToArray();
```
ppekrol, line 178: Shouldn't we forbid changing the URLs after store is initialized?

ayende: Yes, I can't think of any reason why we would want to allow that.

PR #7028, opened 2018-07-12
[4.1] RavenDB-11545 JavaScript index fail when using Date

efratshenhar:


src/Raven.Server/Utils/TypeConverter.cs
```diff
@@ -166,6 +167,12 @@ private static object ToBlittableSupportedType(object root, object value, bool f
             if (value is DateTime || value is DateTimeOffset || value is TimeSpan)
                 return value;
 
+            if (value is JsValue js)
```
ppekrol, line 170: We should add a test for this

talweiss1982 (commented):
Need to to check if we can move this to the jint lucene document converter it has a get value method that could handle it at that level. It is still possible that the map reduce index doesn't go through the converter and then we must handle it at this level.
I would suggest testing this vs a map index and a map reduce index

ppekrol:
cc @ppekrol @arekpalinski

arekpalinski:
We are arriving there from:

```
	Raven.Server.dll!Raven.Server.Utils.TypeConverter.ToBlittableSupportedType(object root, object value, bool flattenArrays, int recursiveLevel) Line 176	C#
 	Raven.Server.dll!Raven.Server.Utils.TypeConverter.ToBlittableSupportedType(object value, bool flattenArrays) Line 141	C#
 	Raven.Server.dll!Raven.Server.Documents.Indexes.MapReduce.Static.MapReduceIndex.AnonymousObjectToBlittableMapResultsEnumerableWrapper.Enumerator.MoveNext() Line 432	C#
 	Raven.Server.dll!Raven.Server.Documents.Indexes.MapReduce.MapReduceIndexBase<Raven.Server.Documents.Indexes.MapReduce.Static.MapReduceIndexDefinition, Raven.Server.Documents.Indexes.IndexField>.PutMapResults(Sparrow.Json.LazyStringValue lowerId, Sparrow.Json.LazyStringValue id, System.Collections.Generic.IEnumerable<Raven.Server.Documents.Indexes.MapReduce.MapResult> mappedResults, Raven.Server.ServerWide.Context.TransactionOperationContext indexContext, Raven.Server.Documents.Indexes.IndexingStatsScope stats) Line 120	C#
 	Raven.Server.dll!Raven.Server.Documents.Indexes.MapReduce.Static.MapReduceIndex.HandleMap(Sparrow.Json.LazyStringValue lowerId, Sparrow.Json.LazyStringValue id, System.Collections.IEnumerable mapResults, Raven.Server.Documents.Indexes.Persistence.Lucene.IndexWriteOperation writer, Raven.Server.ServerWide.Context.TransactionOperationContext indexContext, Raven.Server.Documents.Indexes.IndexingStatsScope stats) Line 285	C#
```

It appears we need to have that check in `TypeConverter.cs ` although it seems to me that we'll be better of by calling `JintLuceneDocumentConverter.GetValue` then (would require to make it `public static`) as that would be a single transformation point of JsValues:

https://github.com/ravendb/ravendb/blob/v4.1/src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/JintLuceneDocumentConverter.cs#L175

@talweiss1982 Thoughts?

talweiss1982:
The converter get value is not generic enough. There is the blittable to js bridge that holds similar functionality but it was also lacking some of the conversion I needed. @maximburyak wants to push all jint related functionality to a single place and I agree with that. The other converters should invoke that single source of conversion this will also ensure we are consistent.
As for now i would leave Efrat fix because calling the converter would be wrong, we flatten arrays for indexes and it is not what we want for patches and subscriptions and future features.

maximburyak:
@ppekrol, can we merge that for now, until we have that significant SingleRun integration refactoring?

maximburyak:
@efratshenhar, I see that the previously reviewed code has been removed and the tests that related to it was change. Can you confirm one of the 2: the previous change was redundant, or it's just a "work in progress" for now?

efratshenhar:
It was by mistake

ppekrol:
I'm merging this, but the we should have one place for those kind of conversions like @arekpalinski wrote.

There can even be a switch to 'flattenArrays' or not. Like we have in TypeConverter. Maybe it would be better to move the method from 'JintLuceneDocumentConverter' to 'TypeConverter'.

ppekrol:
http://issues.hibernatingrhinos.com/issue/RavenDB-11555

PR #7037, opened 2018-07-13
RavenDB-11058 added attachment copy and move

ppekrol:


ayende (approved):

src/Raven.Client/Documents/Session/DocumentSessionAttachmentsBase.cs
```diff
@@ -126,7 +126,7 @@ public void Rename(string documentId, string name, string newName)
                 return; // no-op
 
             if (DocumentsById.TryGetValue(documentId, out DocumentInfo documentInfo) && DeletedEntities.Contains(documentInfo.Entity))
-                throw new InvalidOperationException($"Can't rename attachment {name} of document {documentId}, the document was already deleted in this session.");
+                throw new InvalidOperationException($"Can't rename attachment '{name}' of document '{documentId}', the document was already deleted in this session.");
```
maximburyak: We want to encapsulate the throw statement by a separate method (repeats later on as well)

ppekrol: Applied here https://github.com/ravendb/ravendb/pull/7056

PR #7040, opened 2018-07-13
[4.0] RavenDB-11409 If database is shutting down let's wrap any exception t…

arekpalinski:
…hrown in StateChanged into DatabaseDisabledException. This way we won't expose exceptions like ObjectDisposedExceptions, which are expected during db shut down, directly to the user.

### DIFF
--- a/src/Raven.Server/Documents/DocumentDatabase.cs
+++ b/src/Raven.Server/Documents/DocumentDatabase.cs
@@ -328,7 +328,12 @@ namespace Raven.Server.Documents
 
         internal void ThrowDatabaseShutdown(Exception e = null)
         {
-            throw new DatabaseDisabledException("The database " + Name + " is shutting down", e);
+            throw CreateDatabaseShutdownException(e);
+        }
+
+        internal DatabaseDisabledException CreateDatabaseShutdownException(Exception e = null)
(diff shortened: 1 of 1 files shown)


ayende (approved):

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -806,13 +811,18 @@ public void StateChanged(long index)
...
+
                 RachisLogIndexNotifications.NotifyListenersAbout(index, e);
 
                 if (_logger.IsInfoEnabled)
                     _logger.Info($"Got exception during StateChanged({index}).", e);
 
-                if (_databaseShutdown.IsCancellationRequested)
-                    ThrowDatabaseShutdown(e);
+                if (throwShutDown != null)
+                    throw throwShutDown;
```
maximburyak, line 825: We want the "throw" code to be encapsulated in a separate function in order to reduce generated function code size and complexity

arekpalinski: I know but:
- we have `throw` here anyway as we rethrow the exception
- I don't think that it's hot path because `StateChanged` is called really occasionally

PR #7051, opened 2018-07-16
RavenDB-10953 support TCP negotiation

talweiss1982:


ppekrol:
@karmeli87 can you check this?

PR #7055, opened 2018-07-16
RavenDB-11494 object not indexed in JS because 'TryGetValue' generate…

talweiss1982:
…s the property

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/JintLuceneDocumentConverter.cs
+++ b/src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/JintLuceneDocumentConverter.cs
@@ -87,7 +87,8 @@ namespace Raven.Server.Documents.Indexes.Persistence.Lucene.Documents
                             }
                         }
 
-                        if (actualValue.AsObject().TryGetValue("$spatial", out var inner))
+                        var objectValue = actualValue.AsObject();
+                        if (objectValue.HasOwnProperty("$spatial") && objectValue.TryGetValue("$spatial", out var inner))
                         {
 
                             SpatialField spatialField;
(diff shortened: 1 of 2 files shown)


src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/JintLuceneDocumentConverter.cs
```diff
@@ -207,7 +216,7 @@ private object GetValue(JsValue jsValue)
             }
             else if (jsValue.IsObject())
             {
-                return JavaScriptIndexUtils.StringifyObject(jsValue);
+                return JavaScriptIndexUtils.StringifyObject(jsValue).ToString();
```
ppekrol, line 219: @talweiss1982 shouldn't we convert this to BlittableObject? Check how they are handled in:
https://github.com/ravendb/ravendb/blob/a0760baf73a63bdbb4a37f0e04f43d46bdd5e8a4/src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/LuceneDocumentConverterBase.cs#L318-L349

maximburyak: @ppekrol I see your point, but with the current implementation of BlittableObjectReader, we end up creating a lot of arrays, that the memory stream uses, meaning, it's more or less the same as strigify, in my opinion. 
In addition to that, I think that in the buttom line, lucene indexes a string.

ppekrol: Then it would be better to pass the ObjectInstance to the converter and handle that there.

ppekrol: There = inside LuceneDocumentConverterBase like other types of object.

maximburyak: Looking at the problem again, I think that we should do that, although I don't like the fact that we should convert ObjectInstance to blittable to StreamReader to a string just for that, but I don't like having to different approaches performing the same stringification even more :/

ppekrol: We can pass `JsObject` and handle that inside `LuceneDocumentConverterBase`. No need for conversion to BlittableObject.

maximburyak: I'm not sure I follow.. how will we strigify the jsObject there?

maximburyak: we are going to create a BlittableJsonReaderObject here, probably using JsBlittableBridge

PR #7056, opened 2018-07-16
RavenDB-11552 Deferred commands do not affect session state

ppekrol:


src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
```diff
@@ -77,21 +80,154 @@ CommandType GetCommandType(BlittableJsonReaderObject batchResult)
...
+                ThrowMissingField(CommandType.PATCH, nameof(PatchStatus));
+
+            if (Enum.TryParse(statusAsString, ignoreCase: true, out PatchStatus status) == false)
+                ThrowMissingField(CommandType.PATCH, nameof(PatchStatus));
+
+            switch (status)
+            {
+                case PatchStatus.Created:
+                case PatchStatus.Patched:
+                    HandleDeleteInternal(batchResult, CommandType.PATCH); // deleting because it is impossible to know what was in the patch
```
ppekrol: @ayende @maximburyak not sure what to do here. What do you think?

ppekrol:
@maximburyak Can you review?

src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
```diff
@@ -77,21 +80,154 @@ CommandType GetCommandType(BlittableJsonReaderObject batchResult)
...
+
+            var attachments = documentInfo.Metadata.TryGet(Constants.Documents.Metadata.Attachments, out BlittableJsonReaderArray attachmentsJson)
+                ? new DynamicJsonArray(attachmentsJson)
+                : new DynamicJsonArray();
+
+            documentInfo.Metadata.Modifications[Constants.Documents.Metadata.Attachments] = attachments;
+
+            attachments.Add(new DynamicJsonValue
+            {
+                [nameof(AttachmentDetails.ChangeVector)] = GetStringField(batchResult, type, nameof(AttachmentDetails.ChangeVector)),
```
maximburyak: We can avoid the string allocation here, we could just use the LazyStringValue

src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
```diff
@@ -77,21 +80,154 @@ CommandType GetCommandType(BlittableJsonReaderObject batchResult)
...
-            // DeferredCommands
-            for (var i = _sessionCommandsCount; i < result.Results.Length; i++)
+            if (_session.DocumentsById.TryGetValue(id, out var documentInfo) == false)
+                return;
+
+            if (documentInfo.Metadata.TryGet(Constants.Documents.Metadata.Attachments, out BlittableJsonReaderArray attachmentsJson) == false || attachmentsJson == null ||
+                attachmentsJson.Length == 0)
+                return;
+
+            var name = GetStringField(batchResult, type, attachmentNameFieldName);
```
maximburyak: We could avoid string materialization here, by directly comparing the attachmentNameFieldName with attachmentName, as they are both LazyStringValues

src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
```diff
@@ -154,6 +295,27 @@ private void UpdateCounterValuesInCache(BlittableJsonReaderArray counters, strin
             }
         }
 
+        private static string GetStringField(BlittableJsonReaderObject json, CommandType type, string fieldName)
```
maximburyak: I recommend to see where we can avoid using that function

src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
maximburyak: We could have a direct comparison between this and the name from Attachment in line 150

src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
maximburyak: We could use LazyStringValue directrly here

src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
maximburyak: It makes sense, don't think we should do anything else here

ppekrol:
@maximburyak nice catch!

src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
```diff
@@ -20,13 +22,17 @@ public BatchOperation(InMemoryDocumentSessionOperations session)
 
         private List<object> _entities;
         private int _sessionCommandsCount;
+        private int _allCommandsCount;
 
         public BatchCommand CreateRequest()
         {
             var result = _session.PrepareForSaveChanges();
             _sessionCommandsCount = result.SessionCommands.Count;
```
ayende, line 31: Why is this needed now?

src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
```diff
@@ -64,48 +70,192 @@ CommandType GetCommandType(BlittableJsonReaderObject batchResult)
                         $"Cluster transaction was send to a node that is not supporting it. So it was executed ONLY on the requested node on {_session.RequestExecutor.Url}.");
             }
 
-            for (var i = 0; i < _sessionCommandsCount; i++)
+            for (var i = 0; i < _allCommandsCount; i++)
```
ayende: Better to have two `for` statements, one from `0 .. _sessionCommandsCount` and one from `_sessionCommandsCount` to `_allCommandsCount`

src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
ayende: This will mean that loading the document again will _return null_, no?
In this case, probably should evict it.

Note that I'm pretty sure that we already check that the document isn't already in the session / modified.

ppekrol: This will remove it from the session (evict). So next 'Load' will issue a new request to the server to get it.

Downside of this is that following code fill not work:
```csharp
using (session)
{
    var x = load
    patch(x)
    saveChanges

    x.Name = 'Name' // not tracked anymore
    saveChanges // nothing will happen
}
```

ayende: _should_ such code work?
That is a very strange thing to do, I think.
I also think we check if there are modifications to a patched entity.
I'm fine with keeping that check across `SaveChanges` calls.

Maybe just mark this as patched, meaning that unless user explicitly evict, we can't user the existing instance?

ppekrol: What if... just before sending the DeferredPatchCommand... we will check if entity exists on the session. And if it does, then issue this command with additional parameter to return the document after patch in the 'replay'. We can then update the session properly?

ayende: Sounds good to me. Yes.

ppekrol: Done

src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
ppekrol: Done

ayende (approved):

PR #7057, opened 2018-07-16
[4.0] RavenDB-11422

aviv86:


test/SlowTests/Issues/RavenDB-11422.cs
```diff
@@ -0,0 +1,451 @@
...
+                    });
+
+                    commands.RequestExecutor.Execute(command, commands.Context);
+
+                    var results = new DynamicArray(command.Result.Results);
+                    Assert.Equal(2, results.Count);
+
+                    foreach (dynamic r in results)
+                    {
+                        Assert.NotNull(r[Constants.Documents.Metadata.Key][Constants.Documents.Metadata.IndexScore]);
```
ppekrol: @aviv86 I think we should verify that this score is not 0?

PR #7059, opened 2018-07-16
Raven db 10988

ml054:


src/Raven.Server/Documents/PeriodicBackup/Restore/RestoreUtils.cs
```diff
@@ -77,13 +84,45 @@ public static class RestoreUtils
...
                     NodeTag = folderDetails.NodeTag
                 });
             }
         }
 
+        private static bool CheckIfSnapshotIsEncrypted(string filePath, TransactionOperationContext context)
+        {
+            using (var zip = ZipFile.Open(filePath, ZipArchiveMode.Read, System.Text.Encoding.UTF8))
+            {
+                foreach (var zipEntries in zip.Entries.GroupBy(x => x.FullName.Substring(0, x.FullName.Length - x.Name.Length)))
```
ppekrol: I think that we do not need to group here? Just find an entry which name equals to `RestoreSettings.SettingsFileName` ?

ppekrol (changes requested):
Check my comment.

ml054:
@ppekrol I've force pushed my changes. Please review.

PR #7072, opened 2018-07-17
[4.1] RavenDB-11555 move JS conversions for indexing purposes to one place

efratshenhar:


src/Raven.Server/Utils/TypeConverter.cs
```diff
@@ -142,6 +144,47 @@ private static object ToBlittableSupportedType(object root, object value, bool f
...
                 NestingLevelTooDeep(root);
             }
+             if (value is JsValue js)
+            {
+                if (js.IsNull() || js.IsUndefined())
+                    return null;
+                if (js.IsString())
+                    return js.AsString();
+                if (js.IsBoolean())
+                    return js.AsBoolean().ToString(); // avoid boxing the boolean
```
ppekrol: This should be just a bool. Without .ToString()

src/Raven.Server/Utils/TypeConverter.cs
```diff
@@ -142,6 +144,47 @@ private static object ToBlittableSupportedType(object root, object value, bool f
...
+                        case LazyNumberValue lnv:
+                            return lnv; //should be already blittable supported type.
+                    }
+                    ThrowInvalidObject(js);
+                }
+                //Array is an object in Jint
+                else if (js.IsArray())
+                {
+                    var arr = js.AsArray();
+                    return EnumerateArray(arr);
```
ppekrol: You need to take into account the value of 'flattenArrays' and return DynamicJsonArray. Check how we do it with non-js objects.

ppekrol (changes requested):

src/Raven.Server/Utils/TypeConverter.cs
```diff
@@ -142,6 +144,47 @@ private static object ToBlittableSupportedType(object root, object value, bool f
...
+                }
+                else if (js.IsObject())
+                {
+                    return JavaScriptIndexUtils.StringifyObject(js);
```
ppekrol: Check this comment also: https://github.com/ravendb/ravendb/pull/7055#discussion_r202927174
Consult with @talweiss1982

src/Raven.Server/Utils/TypeConverter.cs
maximburyak: Ok, we should do that

src/Raven.Server/Utils/TypeConverter.cs
maximburyak: we are going to create a BlittableJsonReaderObject here, probably using JsBlittableBridge

src/Raven.Server/Utils/TypeConverter.cs
ayende: This is done intentionally. When using this for indexing, we index `true` as the string `True`
Because this method returns an `object`, if we return a `Boolean` here it will be boxed. So we cut the middleman and avoid the boxing allocation. The end result is that the string `True` is indexed.
Note that this is an optimization only, and intended only for indexing.

ppekrol: Ok, although I do not like the idea of having a different flow inside LuceneDocumentConverterBase for bools from JSIndex and any other type of index.

ayende: I'm fine with making this change if this code is now reused, mind.

ppekrol: Ok. Then let's remove it.

PR #7074, opened 2018-07-17
[4.0] RavenDB-11497 register WellKnownCertificates.Admin thumbprints during…

iftahbe:
… startup

This way, the well known certificates are in the server store along with all the other certificates. All the existing certificate infrastructure will just work (including displaying them in the Studio)

The disadvantage of this approach is that when I want to revoke trust for a certificate, I have to delete it both from the settings.json configuration **and** from the certificate view in the Studio (this can be documented).

Will open a PR for 4.1 when/if this one is approved.


ayende:
I don't like this approach. It is better to just return it from the handler. The risk here is indeed that values in the settings.json (or command line) which are expected to be transient, will become permanent.

PR #7090, opened 2018-07-18
RavenDB-10415 Removing lock between read/write transactions

ayende:


src/Voron/Impl/StorageEnvironmentState.cs
```diff
@@ -1,32 +1,47 @@
...
-                {
-                    Root = Root?.Clone(),
-                    NextPageNumber = NextPageNumber,
-                    Options = Options
-                };
+            {
+                Root = Root?.Clone(),
+                NextPageNumber = NextPageNumber,
+                Options = Options,
+                SnapshotCache = SnapshotCache,
```
talweiss1982, line 41: At the cost of asking a dumb question, is it safe to shallow copy a dictionary and a list?
The low-level transaction is cloning the state it should not be able to modify the snapshot state but correct me if I'm wrong the snapshot cache and the PagerStatesAllScratchesCache  modification will reflect on to the snapshot wouldn't they?

ayende: That is _not_ a dumb question.
And the usual answer is that it is _not_ safe.

In this case, however, the list / dictionary are readonly and are never mutated.

talweiss1982: Maybe use ReadOnlyDictionary<TKey, TValue> and friends to make sure this doesn't happen?

ayende: I looked into doing that, and I worry that this may cause perf issues.
I know that `List<T>` has special treatment and is highly efficient. Turning that to `ReadOnlyCollection` may have impact that we don't like. 

I'm going to make the change anyway, because it it relatively self contained. But it is something to be aware off and we'll need to see if this pops up in the profiler output for 4.1

ayende:
Trying this out shows about 80% _more_ time to run the fast tests.
This is not a good thing, so we'll test this out more and probably do a few profiler runs first.

PR #7098, opened 2018-07-18
RavenDB-11592 Fix Embedded to redirect input to the console

IdanHaim:


src/Raven.Embedded/RavenServerRunner.cs
```diff
@@ -71,6 +71,7 @@ public static Process Run(ServerOptions options)
                 CreateNoWindow = true,
                 RedirectStandardOutput = true,
                 RedirectStandardError = true,
+                RedirectStandardInput = true,
```
ppekrol, line 74: Should we do the same in TestDriver?

IdanHaim:
DON'T we want to change the test drive?

ppekrol:
Can you elaborate what do you mean?

PR #7099, opened 2018-07-18
v4.1 Raven db 11497 41 well known

ml054:


src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
```diff
@@ -472,12 +468,13 @@ public Task GetCertificates()
                         {
                             using (context.OpenReadTransaction())
                             {
+                                // Since we're not passive, the server cert is part of the cluster certificates
```
ayende, line 471: Is this true? We don't add that to the local state any longer.

ayende (changes requested):

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
iftahbe: Yes, we register the server certificate in EnsureNotPassive()

ayende: Why are we doing that?

iftahbe: It's for the manual setup case where we run a cluster with different certificates. We need all the cluster certificates registered in the store so that nodes trust eachother.

When thinking about it more... When building such a cluster, we first register A's cert in B. Then when node A adds node B, A registers B's certificate in the cluster. But also: node A's certificate (which was in node B's local state) now propagates to the server store. So there is no actual need for node A to register its own certificate during EnsureNotPassive().

ayende: I'm not sure that I'm following. In manual setup, before you get to this stage, you already have to manually setup the trust, no?

iftahbe: Yes, the original motivation was for manual setup but because we manually setup trust anyway, it is not actually needed.

I though of another scenario which relies on the server certificate being registered:
When we renew a let's encrypt cert, each node replaces its own cert. There will be a window of time where one server replaced it while the others still run on the older cert. This server still needs to trust the older certificate because now it will not accept connections from the other nodes because the server certificates are not equal.

I guess we can register the server cert only for the duration of the renew process and then delete it afterwards.

iftahbe: Please merge this, and I will make the changes in another PR

PR #7101, opened 2018-07-18
[4.0] RavenDB-11396 Better error for wrong host name

iftahbe:


src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1396,6 +1402,25 @@ internal static bool OnServerCertificateCustomValidationCallback(object sender,
             return false;
         }
 
+        private static void ThrowNameMismatchException(object sender, X509Certificate cert)
+        {
+            var hostname = ((HttpRequestMessage)sender).RequestUri.DnsSafeHost;
+            var cert2 = new X509Certificate2(cert.Handle);
```
iftahbe: @ayende This is a difference between 4.0 and 4.1
Netstandard1.3 doesn't have a ctor which accepts x509certificate so using the handle instead.

src/Raven.Client/Exceptions/Security/NameMismatchException.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+
+namespace Raven.Client.Exceptions.Security
+{
+    public class NameMismatchException : AuthenticationException
```
ppekrol: CertificateNameMismatchException maybe?

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1396,6 +1402,29 @@ internal static bool OnServerCertificateCustomValidationCallback(object sender,
             return false;
         }
 
+        private static void ThrowNameMismatchException(object sender, X509Certificate cert)
+        {
+            var hostname = ((HttpRequestMessage)sender).RequestUri.DnsSafeHost;
```
ppekrol: It is not always HttpRequestMessage. We are using this callback also in Changes API with WebSockets. Then the sender is SslStream

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1396,6 +1402,29 @@ internal static bool OnServerCertificateCustomValidationCallback(object sender,
             return false;
```
ppekrol, line 1407: Should we also check and throw here?

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1396,6 +1402,29 @@ internal static bool OnServerCertificateCustomValidationCallback(object sender,
...
         }
 
+        private static void ThrowNameMismatchException(object sender, X509Certificate cert)
+        {
+            var hostname = ((HttpRequestMessage)sender).RequestUri.DnsSafeHost;
+#if NETSTANDARD1_3
+            var cert2 = new X509Certificate2(cert.Handle);
+#else
+            var cert2 = new X509Certificate2(cert);
```
ppekrol: var cert2 = cert as X509Certificate2 ?? new X509Certificate2(cert);

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1396,6 +1402,29 @@ internal static bool OnServerCertificateCustomValidationCallback(object sender,
             return false;
         }
 
+        private static void ThrowNameMismatchException(object sender, X509Certificate cert)
+        {
+            var hostname = ((HttpRequestMessage)sender).RequestUri.DnsSafeHost;
+#if NETSTANDARD1_3
+            var cert2 = new X509Certificate2(cert.Handle);
```
ppekrol: var cert2 = cert as X509Certificate2 ?? new X509Certificate2(cert.Handle);

It looks like not much is happening in ctor, but let's do this.

ppekrol (changes requested):

src/Raven.Client/Http/RequestExecutor.cs
ayende: I think we better. The major issue is that if we don't, we get a pretty cryptic (pun intended) error.

src/Raven.Client/Http/RequestExecutor.cs
ayende: We already have code that does this, IIRC

ppekrol (approved):

PR #7107, opened 2018-07-19
[4.0] RavenDB-11598 fixing test

ppekrol:


test/FastTests/Client/LoadAsync.cs
```diff
@@ -60,12 +62,26 @@ public class LoadAsync : RavenTestBase
...
+                            {
+                                await Task.WhenAll(load1, load2, load3, load4, load5);
+                            }
+                            catch (InvalidOperationException)
+                            {
+                                throw;
+                            }
+                            catch (Exception)
+                            {
+                                // we do not have a choice, we are accessing session in a concurrent way
```
ayende, line 81: So this might still fail, right?

ppekrol: If all 5 of them will throw NRE, yes...

ayende: I wonder if we can make it _not_ race by hitting a slow proxy?

ayende: Probably not worth it

ppekrol: I think we can go with this... and when it fails... remove this test entirely? There are too many places inside Session that can blow when accessed concurrently.

ayende (approved):

PR #7114, opened 2018-07-19
RavenDB-11589 support for string[] in settings.json

ppekrol:


ayende (commented):
Interesting approach, how would that handle specifying the same value multiple times in the command line?

ppekrol:
CommandLine provider takes last value only. Just checked.

ppekrol:
Refactored a bit the code there. Added string[] support for nested json also.

ayende (approved):

PR #7116, opened 2018-07-19
[4.1] RavenDB-11577 Import database from another server 4.1 is detected as 4.0

efratshenhar:


src/Raven.Server/Smuggler/Migration/MigrationConfiguration.cs
```diff
@@ -35,6 +35,8 @@ public abstract class MigrationConfigurationBase
 
         public int BuildVersion { get; set; }
 
+        public string ProductVersion { get; set; }
```
ml054: IS it required?
image

ppekrol: This is not needed and can be removed.

src/Raven.Studio/typescript/models/database/tasks/migrateDatabaseModel.ts
```diff
@@ -102,14 +103,17 @@ class migrateDatabaseModel {
             ApiKey: this.showApiKeyCredentialInputs() ? this.apiKey() : null, 
             EnableBasicAuthenticationOverUnsecuredHttp: this.apiKey() ? this.enableBasicAuthenticationOverUnsecuredHttp() : false, 
             BuildMajorVersion: this.serverMajorVersion(),
-            BuildVersion: this.buildVersion()
+            BuildVersion: this.buildVersion(),
+            ProductVersion: this.productVersion()
```
ppekrol: This is not needed and can be removed.

ppekrol (changes requested):

PR #7124, opened 2018-07-19
RavenDB-11572 add @id to CSV export when no fields are provided

talweiss1982:


ppekrol:
@talweiss1982 can you check failing tests?

talweiss1982:
sure

talweiss1982:
I have modified the test to reflect the fact that we export the id now

src/Raven.Server/Documents/Handlers/StreamCsvDocumentQueryResultWriter.cs
```diff
@@ -101,9 +102,13 @@ private void WriteCsvHeaderIfNeeded(Document res)
             _csvWriter.NextRecord();
         }
 
-        private IEnumerable<(string Property, string Path)> GetPropertiesRecursive((string ParentProperty, string ParentPath) propertyTuple, BlittableJsonReaderObject obj)
+        private IEnumerable<(string Property, string Path)> GetPropertiesRecursive((string ParentProperty, string ParentPath) propertyTuple, BlittableJsonReaderObject obj, bool addId = true)
         {
             var inMetadata = Constants.Documents.Metadata.Key.Equals(propertyTuple.ParentPath);
+            if (addId)
```
arekpalinski, line 108: How do we prevent from adding `@id` column twice when doing export csv in collection? (according to issue description `@id` was already exported then)

talweiss1982: This code is only invoked when not specifying the fields and it is taken from the first document. When specifying fields the fields you specified will be in the header.

arekpalinski: OK

test/SlowTests/Issues/RavenDB_9519.cs
```diff
@@ -17,16 +17,25 @@ public class RavenDB_9519 : RavenTestBase
...
                     session.SaveChanges();
+                    cv = session.Advanced.GetChangeVectorFor(_testCompany);
                 }
 
                 var client = new HttpClient();
                 var stream = await client.GetStreamAsync($"{store.Urls[0]}/databases/{store.Database}/streams/queries?query=From%20companies&format=csv");
+                
+                using (var session = store.OpenSession())
+                {
+                    session.Delete(id);
```
arekpalinski: Why to delete it? You added the assertion that CV has changed so we'll be sure the doc was overwritten

maximburyak: I guess that this code could be removed, because SaveChanges is not called anyway

arekpalinski: Right, I didn't notice that :)

PR #7127, opened 2018-07-20
RavenDB-7787 reading from StandardError and displaying that to the user

ppekrol:


ayende (approved):
I wonder if it wouldn't be better to merge the output / error to a single string.
That would be similar to what you actually see? 

On the other hand, if there is an error, you'll get just the error...

PR #7131, opened 2018-07-20
Fixing a race conditions when moving from Passive state to Follower s…

ayende:
…tate when adding new node to the cluster

We'll not start the timer clock until _after_ the leader pushed the snapshot to us if we are already passive.

### DIFF
--- a/src/Raven.Server/Rachis/Follower.cs
+++ b/src/Raven.Server/Rachis/Follower.cs
@@ -467,6 +467,8 @@ namespace Raven.Server.Rachis
 
                 context.Transaction.Commit();
             }
+            // we might have moved from passive node, so we need to start the timeout clock
+            _engine.Timeout.Start(_engine.SwitchToCandidateStateOnTimeout);
             _debugRecorder.Record("Snapshot installed");
             //Here we send the LastIncludedIndex as our matched index even for the case where our lastCommitIndex is greater
             //So we could validate that the entries sent by the leader are indeed the same as the ones we have.
@@ -756,7 +758,8 @@ namespace Raven.Server.Rachis
         
         public void AcceptConnection(LogLengthNegotiation negotiation)
         {
-            _engine.Timeout.Start(_engine.SwitchToCandidateStateOnTimeout);
+            if(_engine.CurrentState != RachisState.Passive)
+                _engine.Timeout.Start(_engine.SwitchToCandidateStateOnTimeout);
             
             // if leader / candidate, this remove them from play and revert to follower mode
             _engine.SetNewState(RachisState.Follower, this, _term,
--- a/src/Raven.Server/Rachis/Follower.cs
+++ b/src/Raven.Server/Rachis/Follower.cs
@@ -467,8 +467,33 @@ namespace Raven.Server.Rachis
 
                 context.Transaction.Commit();
             }
+
+            _engine.Timeout.Defer(_connection.Source);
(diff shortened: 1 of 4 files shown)


ayende:
All tests are passing

src/Raven.Server/Rachis/Follower.cs
```diff
@@ -467,6 +467,33 @@ private void NegotiateWithLeader(TransactionOperationContext context, LogLengthN
...
             }
+
+            _engine.Timeout.Defer(_connection.Source);
+            _debugRecorder.Record("Invoking StateMachine.SnapshotInstalled");
+
+            // notify the state machine, we do this in an async manner, and start
+            // the operator in a separate thread to avoid timeouts while this is
+            // going on
+
+            var task = Task.Run(() => _engine.SnapshotInstalledAsync(snapshot.LastIncludedIndex));
```
talweiss1982, line 478: Do we need to pass a cancelation token here, otherwise if we dispose of this Follower we can end up with a rogue task, no?

ayende: We can't really get that, because the code will wait until the task is completed.
But that might not be a bad idea, yes.

ayende: Just checked it, and there doesn't seem to be anything there that uses a cancellation token that we can pass, so don't think it will work

ppekrol:
@ayende can you check failing tests http://jenkins.hibernatingrhinos.com/job/PR_Tests_v4.0/1277/ ?

src/Raven.Server/Rachis/Follower.cs
```diff
@@ -467,6 +467,33 @@ private void NegotiateWithLeader(TransactionOperationContext context, LogLengthN
...
+            {
+                // this may take a while, so we let the other side know that
+                // we are still processing, and we reset our own timer while
+                // this is happening
+                MaybeNotifyLeaderThatWeAreStillAlive(context, sp);
+            }
+
+            _debugRecorder.Record("Done with StateMachine.SnapshotInstalled");
+
+            // we might have moved from passive node, so we need to start the timeout clock
```
arekpalinski, line 494: "might"? what if we haven't? is it safe to start the timeout clock if we are still passive?

talweiss1982: (writing from my phone)
Can you check what happens when an old v4.0 raft is getting this messages, I'm not sure if we should maybe bump the raft tcp protocol version for this or not.

arekpalinski: Is it related to my question in the above comment or separate one?

Anyway, `MaybeNotifyLeaderThatWeAreStillAlive` sends `InstallSnapshotResponse` message having `Done = false`. I've found the following piece of code which has some logic based on value of `Done` property:
https://github.com/ravendb/ravendb/blob/v4.0/src/Raven.Server/Rachis/FollowerAmbassador.cs#L518-L527
As it's not a new concept it doesn't seem to me to be a protocol change.

ayende: @arekpalinski  When we get here, we aren't passive, because we are now part of a cluster.  This happened in the `InstallSnapshot` section. So if we were passive, now we aren't and we want to register the timeout handler.

@talweiss1982 This is the same behavior as we had previous. We essentially have the same logic, notifying the other side that we are pending. Not protocol change here.

ayende:
Cannot get these tests to fail on my end.

PR #7134, opened 2018-07-22
RavenDB-11415 Raft and cluster transaction fixes

karmeli87:


src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -47,6 +47,20 @@ public DatabasesLandlord(ServerStore serverStore)
             _concurrentDatabaseLoadTimeout = _serverStore.Configuration.Databases.ConcurrentLoadTimeout.AsTimeSpan;
             _logger = LoggingSource.Instance.GetLogger<DatabasesLandlord>("Server");
             CatastrophicFailureHandler = new CatastrophicFailureHandler(this, _serverStore);
+            PoolOfThreads.GlobalRavenThreadPool.LongRunning(_ =>
```
ayende: I think it might bet better to do this as an async task. 
Taking up a thread just for this (relatively rare) seems expensive.

ayende: Also, note that you aren't waiting for this to complete on disposal

ayende: Why is this at the `DatabaseLandlord` level instead of the `DocumentDatabase` level?

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -165,71 +179,92 @@ private void HandleClusterDatabaseChanged(string databaseName, long index, strin
...
+
+                _newClusterTx.Reset();
+
+                var didWork = true;
+                while (true)
                 {
-                    _logger.Info($"Failed enqueue cluster transaction command for database: '{databaseName}'", e);
+                    if (didWork == false)
+                    {
+                        Thread.Sleep(500);
```
ayende: Why do you have this here?

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -165,71 +179,92 @@ private void HandleClusterDatabaseChanged(string databaseName, long index, strin
...
+                                _logger.Info($"Can't perform cluster transaction on database '{key}', failed to load.", e);
+                            }
+                        }
+                    }
                 }
             }
         }
 
-        public void ExecutePendingClusterTransactions(DocumentDatabase database, long index, Action<List<Task>> customWait = null)
+        public void ExectueClusterTransactionOnDatabase(DocumentDatabase database, Action<List<Task>> customWait = null)
```
ayende: This should be a method on `DocumentDatabase`

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -165,71 +179,92 @@ private void HandleClusterDatabaseChanged(string databaseName, long index, strin
...
-                // As a work around we pass 'fromCount: 0'
-                var commands = ClusterTransactionCommand.ReadCommandsBatch(serverContext, database.Name, fromCount: 0).ToList();
+                var currentCount = _clusterTxCount.GetOrAdd(database.Name, 0);
+                var commands = ClusterTransactionCommand.ReadCommandsBatch(serverContext, database.Name, fromCount: currentCount).ToList();
+                if (commands.Count == 0)
+                {
+                    return;
+                }
+
+                _clusterTxCount[database.Name] = commands.Last().PreviousCount;
```
ayende: Should only happen after successful command execution.

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -475,6 +475,11 @@ public override int Execute(DocumentsOperationContext context)
                                     try
                                     {
                                         var item = Database.DocumentsStorage.GetDocumentOrTombstone(context, cmd.Id);
+                                        if (item.Missing)
+                                        {
+                                            var missingMsg = $"This shouldn't happened and probably a bug! Missing id: {cmd.Id}, doc: {_count}, raft:{current} on {context.DocumentDatabase.ServerStore.NodeTag}";
```
ayende: why can't this happen?

src/Raven.Server/Rachis/Candidate.cs
```diff
@@ -94,17 +94,7 @@ private void Run()
                     {
                         if (_peersWaiting.WaitOne(_engine.Timeout.TimeoutPeriod) == false)
                         {
-                            ElectionTerm = _engine.CurrentTerm;
-
-                            // timeout? 
-                            if (IsForcedElection)
-                            {
-                                CastVoteForSelf(ElectionTerm + 1, "Timeout during forced elections", setStateChange: false);
```
ayende, line 102: Where did this move to?

src/Raven.Server/Rachis/Leader.cs
```diff
@@ -606,37 +606,45 @@ private void CheckPromotables()
...
+
+        public async Task<(long Index, object Result)> PutAsync(CommandBase command, TimeSpan timeout)
         {
 
-            Task<(long Index, object Result)> task;
-            long index;
+            _commandsQueue.Enqueue(command);
+
             using (_engine.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
             using (context.OpenWriteTransaction()) // this line prevents concurrency issues on the PutAsync
```
ayende: The idea here is to not have to wait too much for this. 
This way, even if we already executed, we are still convoying on the lock

ayende (changes requested):

src/Raven.Server/Documents/DatabasesLandlord.cs
karmeli87: I want to have a single task per server and not per database, otherwise not sure about the performance of this when we will have let say 50 databases..

src/Raven.Server/Documents/DatabasesLandlord.cs
karmeli87: Avoid a tight loop when we are waiting the database to load

src/Raven.Server/Documents/Handlers/BatchHandler.cs
karmeli87: It happened, but  I wasn't sure how this possible... still need to figure this out.

src/Raven.Server/Rachis/Candidate.cs
karmeli87: At the start of the candidate we already check if this is forced and casting the vote.

src/Raven.Server/Documents/DatabasesLandlord.cs
ayende: Why? With a single server wide task, you are causing convoys. 
Better to have one per database and have them handle it directly without one db interfering with another

src/Raven.Server/Documents/DatabasesLandlord.cs
ayende: why do you have a look to wait for the db in the first place?

src/Raven.Server/Documents/Handlers/BatchHandler.cs
ayende: What is this scenario?

src/Raven.Server/Rachis/Candidate.cs
ayende: This can happen on the _next_ term, though.

karmeli87: Not sure I'm following.. In force election we casting the vote and increment the term to give to this node an advantage for the coming elections. (it might happened that the other nodes are already during an election process and have higher terms..)

src/Raven.Server/Documents/Handlers/BatchHandler.cs
karmeli87: Ran `CanPreformSeveralClusterTransactions(5)` in a loop..

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -477,8 +477,8 @@ public override int Execute(DocumentsOperationContext context)
                                         var item = Database.DocumentsStorage.GetDocumentOrTombstone(context, cmd.Id);
                                         if (item.Missing)
                                         {
-                                            var missingMsg = $"This shouldn't happened and probably a bug! Missing id: {cmd.Id}, doc: {_count}, raft:{current} on {context.DocumentDatabase.ServerStore.NodeTag}";
-                                            throw new InvalidDataException(missingMsg);
+                                            // this item is missing and there is nothing we can do about it.
+                                            continue;
```
ayende: This is going to cause issues down the line. If the value is missing, you are skipping it, and in that case, the result to the client will have one item less, no?

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -165,122 +163,21 @@ private void HandleClusterDatabaseChanged(string databaseName, long index, strin
             }
         }
 
-        private void NotifyPendingClusterTransaction(string databaseName, Task<DocumentDatabase> task, long index)
+        public void NotifyPendingClusterTransaction(string name, Task<DocumentDatabase> task)
```
ayende, line 166: Given that we _know_ that the tx is completed here, why are getting as `Task` instead of the `DocumentDatabase` ?

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -285,6 +287,19 @@ public void Initialize(InitializeOptions options = InitializeOptions.None)
...
+
+                Task.Run(() =>
+                {
+                    try
+                    {
+                        ExectueClusterTransactionTask();
+                    }
+                    catch
+                    {
+                        //
```
ayende: ??

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -475,6 +475,11 @@ public override int Execute(DocumentsOperationContext context)
                                     try
                                     {
                                         var item = Database.DocumentsStorage.GetDocumentOrTombstone(context, cmd.Id);
+                                        if (item.Missing)
+                                        {
+                                            // this item is missing and there is nothing we can do about it.
```
ayende: This modify the expected results that are sent to the client, no?

src/Raven.Server/Rachis/Follower.cs
```diff
@@ -687,13 +698,31 @@ private void MaybeNotifyLeaderThatWeAreStillAlive(TransactionOperationContext co
...
                 _engine.Timeout.Defer(_connection.Source);
 
                 _engine.ValidateTerm(_term);
-                
+
+                if (midpointIndex == negotiation.PrevLogIndex && midpointTerm != negotiation.PrevLogTerm)
+                {
+                    // our appended entries has been diverged, same index with different terms.
+                    var msg = $"Our appended entries has been diverged, same index with different terms. " +
+                              $"My index/term {midpointIndex}/{midpointTerm}, while yours is {negotiation.PrevLogIndex}/{negotiation.PrevLogTerm}.";
```
ayende, line 711: Here, make sure that you explain that we will just get the snapshot and that's it.

However, why not go further back in time to find the first matching index/term pair?

src/Raven.Server/Documents/Handlers/BatchHandler.cs
karmeli87: Yes, the only valid scenario that I can think, is if the deleted document was purged, right? 
So, should we recreate the document in that case?

ayende (changes requested):

src/Raven.Server/Documents/DatabasesLandlord.cs
karmeli87: Simply to have same behavior as `NotifyDatabaseAboutStateChange` and `NotifyDatabaseAboutValueChange`.

src/Raven.Server/Rachis/Follower.cs
karmeli87: If this happened we need to go to try the leader's last log of the previous term, which is unknown on the follower side and we don't have an index on the term.

I see a lot of pitfalls in going further back, so maybe create a minor performance issue for 4.1 for this?

src/Raven.Server/Documents/DocumentDatabase.cs
karmeli87: Do we even need this `try/catch`? since the `ExectueClusterTransactionOnDatabase` function is already wrapped with `try/catch`.

src/Raven.Server/Documents/DatabasesLandlord.cs
ayende: They should be changed to avoid this as well, yes.

src/Raven.Server/Rachis/Follower.cs
ayende: Why? Isn't this just considering this as also something that I don't have a match and keep going back using binary search?

src/Raven.Server/Documents/DocumentDatabase.cs
ayende: Probably can remove that, yes.

ayende:
Comments will be handled in separate PR

PR #7144, opened 2018-07-23
[4.0] RavenDB-11415 better cluster commands merger

karmeli87:


ayende:
We don't want to make that big a change for 4.0

PR #7145, opened 2018-07-23
[4.1] RavenDB-11604

aviv86:
http://issues.hibernatingrhinos.com/issue/RavenDB-11604

### DIFF
--- a/src/Raven.Client/Documents/Linq/RavenQueryProviderProcessor.cs
+++ b/src/Raven.Client/Documents/Linq/RavenQueryProviderProcessor.cs
@@ -2096,14 +2096,7 @@ The recommended method is to use full text search (mark the field as Analyzed an
 
                         if (newExpression.Arguments[index] is MethodCallExpression mce && LinqPathProvider.IsCounterCall(mce))
                         {
-                            var counterInfo = LinqPathProvider.CreateCounterResult(mce);
-                            if (_fromAlias == null && counterInfo.Args[0] == lambdaExpression?.Parameters[0].Name)
-                            {
-                                AddFromAlias(counterInfo.Args[0]);
-                            }
-
-                            AddCallArgumentsToPath(counterInfo.Args, ref counterInfo.Path, out _);
-                            AddToFieldsToFetch(counterInfo.Path, GetSelectPath(newExpression.Members[index]));
+                            HandleSelectCounter(mce, lambdaExpression, newExpression.Members[index]);                            
                             continue;
                         }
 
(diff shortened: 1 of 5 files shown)


src/Raven.Client/Documents/Linq/RavenQueryProviderProcessor.cs
```diff
@@ -58,11 +58,15 @@ internal class RavenQueryProviderProcessor<T>
         private StringBuilder _declareBuilder;
         private DeclareToken _declareToken;
         private List<LoadToken> _loadTokens;
+        private HashSet<string> _loadAliases;
         private readonly HashSet<string> _loadAliasesMovedToOutputFuction;
         private int _insideLet = 0;
         private string _loadAlias;
         private bool _selectLoad;
         private const string DefaultLoadAlias = "__load";
+        private const string DefaultAlias = "__ravenDefaultAlias";
```
ppekrol: I think it would be better to make it shorter e.g. `__alias` ?

arekpalinski: Also the variable would be better off by having `DefaultAliasPrefix` name instead of `DefaultAlias` as we append incremented number at the end

src/Raven.Client/Documents/Linq/RavenQueryProviderProcessor.cs
```diff
@@ -2896,6 +2927,39 @@ private void AddToFieldsToFetch(string field, string alias)
...
+
+        private void AddFromAliasToFieldToFetch(ref string field, ref string alias, bool quote = false)
+        {
+            if (field == alias)
+            {
+                alias = null;
+            }
+
+            field = quote
+                ? $"{_fromAlias}.'{field}'"
```
arekpalinski, line 2968: I wonder if it works when we have a query like:

```
session.Query<User>()
                        .Select(user => new
                        {
                            user.Group.Order
                        });
```

I'm pretty sure it will work but let's add such test case for that.

aviv86: no actually that didn't work, good thinking.
fixed this + added test case

ppekrol (approved):

ppekrol:
@arekpalinski can you review? @aviv86 can you check failing tests? Nothing major, assertions are failing because of the default alias change, but I cannot fix it myself till tomorrow.

arekpalinski (approved):

PR #7146, opened 2018-07-23

iftahbe:


src/Raven.Server/RavenServer.cs
```diff
@@ -494,16 +493,35 @@ public bool RefreshClusterCertificate(object state)
...
-                        // we are already in the process of updating the certificate, so we need
-                        // to nudge all the nodes in the cluster to check the replacement state.
-                        // If a node confirmed but failed with the actual replacement (e.g. file permissions)
-                        // this will make sure it will try again in the next round (1 hour).
-                        await ServerStore.SendToLeaderAsync(new RecheckStatusOfServerCertificateCommand());
-                        return;
+                        if (certUpdate.TryGet("Confirmations", out int confirmations) == false)
+                            throw new InvalidOperationException("Expected to get confirmations count");
+
+                        if (certUpdate.TryGet("Replaced", out int replaced) == false)
```
ayende, line 499: Isn't this a new property? If so, what will happen in a cross versioned cluster?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -351,8 +359,63 @@ private void ConfirmReceiptServerCertificate(TransactionOperationContext context
...
+                        throw new ArgumentException("Thumbprint property didn't exist in 'server/cert' value");
+
+                    if (storedThumbprint != thumbprint)
+                        return; // confirmation for a different cert, ignoring
+
+                    certInstallation.TryGet("Replaced", out int replaced);
+
+                    certInstallation.Modifications = new DynamicJsonValue(certInstallation)
+                    {
+                        ["Replaced"] = replaced + 1
```
arekpalinski, line 393: Can we put `"Replaced"` as const and use it across the board

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -863,11 +848,90 @@ private void OnValueChanged(object sender, (long Index, string Type) t)
...
+                    }
+                    catch (Exception e)
+                    {
+                        if (Logger.IsOperationsEnabled)
+                            Logger.Operations($"Failed to process {t.Type}.", e);
+
+                        NotificationCenter.Add(AlertRaised.Create(
+                            null,
+                            "Server certificate",
+                            $"Failed to process {t.Type}.",
```
arekpalinski, line 868: Shouldn't we include in the alert message some info that it is related to certificate replacement?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
iftahbe: Yes we can

src/Raven.Server/ServerWide/ServerStore.cs
iftahbe: sure

src/Raven.Server/RavenServer.cs
iftahbe: Yes that's a problem, will fix.

iftahbe:
Closing, handling this in a different PR

PR #7147, opened 2018-07-24
[4.1] RavenDB-11415 Don't use SetResult on TaskCompletionSource

karmeli87:
It throws "System.InvalidOperationException: An attempt was made to transition a task to a final state when it had already completed."
When the SetResult called twice.

### DIFF
--- a/src/Raven.Server/Documents/ClusterTransactionWaiter.cs
+++ b/src/Raven.Server/Documents/ClusterTransactionWaiter.cs
@@ -34,7 +34,7 @@ namespace Raven.Server.Documents
             Database.RachisLogIndexNotifications.NotifyListenersAbout(index, null);
             if (_results.TryGetValue(id, out var task))
             {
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Documents/ClusterTransactionWaiter.cs
```diff
@@ -34,7 +34,7 @@ public void SetResult(string id, long index, object result)
             Database.RachisLogIndexNotifications.NotifyListenersAbout(index, null);
             if (_results.TryGetValue(id, out var task))
             {
-                task.SetResult(result);
+                task.TrySetResult(result);
```
ppekrol, line 37: Are we expecting this to be called multiple times? We might be hiding another issue here.

ayende: It should't be set multiple times

PR #7148, opened 2018-07-24
RavenDB-11040 Added more stats to reduce phase on indexing performance graph

arekpalinski:


ppekrol (approved):

src/Raven.Studio/typescript/viewmodels/database/indexes/indexPerformance.ts
```diff
@@ -1152,10 +1152,22 @@ class indexPerformance extends viewModelBase {
             }
             if (element.ReduceDetails) {
                 let reduceDetails: string;
-                reduceDetails = `<br/>*** Reduce details ***<br/>`;
-                reduceDetails += `Compressed leafs: ${element.ReduceDetails.NumberOfCompressedLeafs.toLocaleString()}<br/>`;
-                reduceDetails += `Modified branches: ${element.ReduceDetails.NumberOfModifiedBranches.toLocaleString()}<br/>`;
-                reduceDetails += `Modified leafs: ${element.ReduceDetails.NumberOfModifiedLeafs.toLocaleString()}`;
+
+                if (element.ReduceDetails.TreesReduceDetails) {
```
ppekrol, line 1156: Let's add it to @ml054 TODO list and merge it. Thoughts?

ml054:
@arekpalinski please make sure you don't hit this line:
https://github.com/ravendb/ravendb/blob/v4.1/src/Raven.Studio/typescript/viewmodels/database/indexes/indexPerformance.ts#L953

If you introduced new types we should also add new colors.

src/Raven.Client/Documents/Indexes/ReduceRunDetails.cs
```diff
@@ -1,6 +1,23 @@
 ﻿namespace Raven.Client.Documents.Indexes
 {
     public class ReduceRunDetails
+    {
```
ayende, line 4: How about adding the amount of allocated memory in this thread?

src/Raven.Studio/typescript/viewmodels/database/indexes/indexPerformance.ts
ayende: Won't this break the studio for the meantime?

src/Raven.Client/Documents/Indexes/ReduceRunDetails.cs
ayende: What kind of help would these details give us?

arekpalinski: 1) Allocated memory in this thread is already here, it's `CurrentlyAllocated` property
2) It's the same kind of help like in map phase - those are exactly the same values we have there, so we know memory stats and details about numbers of processed map entries

src/Raven.Studio/typescript/viewmodels/database/indexes/indexPerformance.ts
arekpalinski: The studio works correctly. The changes I made in this PR are:
- more details in the tooltips
- added additional timing of Linq / Jint processing (same color like in map so it doesn't break anything)

arekpalinski:
@ml054 No _new_ type was introduced here. Just added Linq / Jint timings so they will have the same colors like in map phase

ml054:
@arekpalinski that's great. I just skim though this and wasn't sure.

PR #7153, opened 2018-07-25
[4.1] RavenDB-11624 NRE in InstallSnapshot on 4.0 when 4.1 talks to 4.0

karmeli87:
- Need to take the version into account when sending the snapshot.
- Connecting a follower (or adding a new node) can affect the version.
- Change the base version from 400 to 400,000 in order to be more flexible.

### DIFF
--- a/src/Raven.Server/Rachis/FollowerAmbassador.cs
+++ b/src/Raven.Server/Rachis/FollowerAmbassador.cs
@@ -9,6 +9,7 @@ using System.Threading;
 using Raven.Client.Exceptions;
 using Raven.Client.Http;
 using Raven.Client.ServerWide;
+using Raven.Server.ServerWide;
 using Raven.Server.ServerWide.Context;
 using Raven.Server.Utils;
 using Sparrow;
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Rachis/FollowerAmbassador.cs
```diff
@@ -804,7 +807,15 @@ private long InitialNegotiationWithFollower()
                 } while (true);
             }
         }
-        
+
+        private int GetFollowerVersion(LogLengthNegotiationResponse llr)
+        {
+            var version = llr.CommandsVersion ?? 400_000;
```
ppekrol, line 813: I think that 40_000 would be enough

karmeli87: Let's be on the (very) safe side this time :)

talweiss1982: if we don't plan to support versioning of the same command then we don't really need so many versions because they correlate to the server version 1:1

ayende (approved):

src/Raven.Server/Rachis/FollowerAmbassador.cs
```diff
@@ -804,7 +807,15 @@ private long InitialNegotiationWithFollower()
...
+            if (version == 400)
```
arekpalinski, line 814: Maybe put some comment here that `400` was used in the past but we switched to higher range

PR #7159, opened 2018-07-26
[4.1] RavenDB-11636

aviv86:


src/Raven.Client/Util/JavascriptConversionExtensions.cs
```diff
@@ -1087,20 +1089,25 @@ public override void ConvertToJavascript(JavascriptConversionContext context)
...
                             writer.Write(".");
-
                         }
 
                         var name = member.Member.Name;
-                        if (ReservedWordsSupport.JsReservedWords.Contains(name))
+
+                        if (DoNotIgnore && name.StartsWith(TransparentIdentifier))
+                        {
+                            name = name.Substring(2);
```
arekpalinski: its' `<>` right? let's put a comment or better move it to const and use .Length here - it will make it easier to figure out what's going on here in the future

src/Raven.Client/Util/JavascriptConversionExtensions.cs
```diff
@@ -1113,7 +1120,7 @@ public override void ConvertToJavascript(JavascriptConversionContext context)
                         {
                             writer.Write(parameter.Name.Substring(2));
                             writer.Write(".");
-                            name = name.Replace("<>h__TransparentIdentifier", "h__TransparentIdentifier");
+                            name = name.Replace(TransparentIdentifier, "h__TransparentIdentifier");
```
arekpalinski: Better to put `"h__TransparentIdentifier"` also as a const next to `internal const string TransparentIdentifier`

src/Raven.Client/Util/JavascriptConversionExtensions.cs
```diff
@@ -18,6 +18,8 @@ namespace Raven.Client.Util
 {
     internal class JavascriptConversionExtensions
     {
+        internal const string TransparentIdentifier = "<>h__TransparentIdentifier";
```
arekpalinski, line 21: BTW we already have it defined in: 

https://github.com/ravendb/ravendb/blob/v4.0/src/Raven.Client/Documents/Linq/LinqPathProvider.cs#L33

Would be nice to have such const in single location

arekpalinski (approved):

src/Raven.Client/Util/JavascriptConversionExtensions.cs
aviv86: done

src/Raven.Client/Util/JavascriptConversionExtensions.cs
aviv86: done

src/Raven.Client/Util/JavascriptConversionExtensions.cs
aviv86: done

PR #7160, opened 2018-07-26
RavenDB-11630 Adding Graph query syntax to RQL - WIP

talweiss1982:
WIP - For review
I'm opening a PR because we need to decide on how flexible our queries will be.
I have taken an approach and would appreciate feedbacks at an early stage.
This PR only contains support for the 'With' clause and doesn't yet support 'Match' so don't try to parse a whole query.


src/Raven.Server/Documents/Queries/AST/FromClause.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Sparrow;
+
+namespace Raven.Server.Documents.Queries.AST
+{
+    public class FromClause
```
ayende: Can't we reuse our existing query AST?

src/Raven.Server/Documents/Queries/AST/Query.cs
```diff
@@ -8,7 +8,9 @@ public class Query
     {
         public bool IsDistinct;
         public QueryExpression Where;
-        public (FieldExpression From, StringSegment? Alias, QueryExpression Filter, bool Index) From;
+        public FromClause From;
+        public List<WithClause> WithClauses;
```
ayende: Dictionary?

src/Raven.Server/Documents/Queries/AST/Query.cs
```diff
@@ -8,7 +8,9 @@ public class Query
...
+        public List<MatchClause> MatchClauses { get; set; }
```
ayende: Are we going to have multiple of them?

src/Raven.Server/Documents/Queries/AST/FromClause.cs
ayende: In the Query it is a tuple, probably need to be modified there as well, but that that this is a struct there.

src/Raven.Server/Documents/Queries/Parser/QueryParser.cs
```diff
@@ -91,12 +106,27 @@ public Query Parse(QueryType queryType = QueryType.Select)
...
+            if (recursive == false && Scanner.AtEndOfInput() == false)
                 ThrowParseException("Expected end of query");
 
             return q;
         }
 
+        private WithClause With()
+        {
+            if (Scanner.TryScan('{') == false)
+                throw new InvalidQueryException("With clause contains invalid body", Scanner.Input, null);
```
ayende: Better errors here

src/Raven.Server/Documents/Queries/Parser/QueryParser.cs
```diff
@@ -355,7 +385,7 @@ private List<(QueryExpression, StringSegment?)> SelectClauseExpressions(string c
         }
 
 
-        private (FieldExpression From, StringSegment? Alias, QueryExpression Filter, bool Index) FromClause()
+        private FromClause FromClause()
```
ayende, line 466: Okay, this is great, the.

src/Raven.Server/Documents/Queries/AST/Query.cs
talweiss1982: We will but they can all reside under MatchClause and then have MatchExpression because we will need to support things like (foo)-[:Like]->(bar) and (cat)-[:Hates]->(dogs)

src/Raven.Server/Documents/Queries/AST/PatternMatchBinaryExpression.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+
+namespace Raven.Server.Documents.Queries.AST
+{
+    public class PatternMatchBinaryExpression : PatternMatchExpression
+    {
+        public enum Operator
+        {
+            // ReSharper disable once InconsistentNaming
+            and,
```
ayende, line 10: Please use `And` and `Or` here.

src/Raven.Server/Documents/Queries/AST/PatternMatchElementExpression.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace Raven.Server.Documents.Queries.AST
+{
+    public class PatternMatchElementExpression : PatternMatchExpression
+    {
+        public enum Direction
```
ayende, line 5: Why do we need direction, isn't that what the `From` / `To` for?

src/Raven.Server/Documents/Queries/AST/PatternMatchExpressEdge.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using Sparrow;
+
+namespace Raven.Server.Documents.Queries.AST
+{
+    public class PatternMatchExpressEdge : PatternMatchExpression
```
ayende, line 6: This should be removed. There is no such thing.
Instead, this should be a `with edge` expression, and the filter should be in the `where` clause

src/Raven.Server/Documents/Queries/AST/Query.cs
```diff
@@ -38,22 +37,18 @@ public override string ToString()
...
+            if (GraphQuery == null)
             {
-                WithClauses = new Dictionary<StringSegment, WithClause>();
+                GraphQuery = new GraphQuery();
+                GraphQuery.WithDocumentQueries = new Dictionary<StringSegment, Query>();
             }
 
-            var allias = with.Query.From.Alias;
+            if (GraphQuery.WithDocumentQueries.ContainsKey(withClause.Allias)  )
+                    return (false, $"Allias {withClause.Allias} is already in use on a diffrent 'With' clause");
```
ayende: `Alias`

src/Raven.Server/Documents/Queries/AST/StringGraphQueryVisitor.cs
```diff
@@ -0,0 +1,378 @@
...
+            {
+                Sb.Append($"WITH EDGES({withEdgesClause.Value.EdgeType}) {{ ");
+                    VisitWithEdgesExpression(withEdgesClause.Value);
+                Sb.AppendLine($" }} AS {withEdgesClause.Key}");
+            }
+        }
+
+        public override void VisitPatternMatchClause(PatternMatchExpression expression)
+        {
+            Sb.AppendLine($"MATCH {expression}");
```
ayende, line 41: You need to `Visti` the expression here.

src/Raven.Server/Documents/Queries/AST/WithEdgesExpression.cs
```diff
@@ -0,0 +1,50 @@
...
+            EdgeType = edgeType ?? throw new ArgumentNullException(nameof(edgeType));
+            Type = ExpressionType.WithEdge;
+        }
+
+        public override string ToString() => GetText();
+        public override string GetText(IndexQueryServerSide parent) => GetText();
+
+        private string GetText()
+        {
+            var sb = new StringBuilder();
```
ayende, line 34: Missing `WITH EDGES` here

ayende (changes requested):

src/Raven.Server/Documents/Queries/AST/PatternMatchElementExpression.cs
myarichuk: We don't, this is mostly for debugging -> I use it in query's ToString()

src/Raven.Server/Documents/Queries/AST/PatternMatchExpressEdge.cs
myarichuk: Will inline the data about the edge in PatternMatchElementExpression

src/Raven.Server/Documents/Queries/AST/PatternMatchElementExpression.cs
ayende: Let's remove that, we don't need this.

`(u1)-[:Rated]->(m)<-[:Rated]-(u2)` is the same as:  `(u1)-[:Rated]->(m) and (u2)-[:Rated]->(m)`

src/Raven.Server/Documents/Queries/AST/PatternMatchExpressEdge.cs
ayende: What do you mean, inline?

myarichuk: I will move the information about the edge (edge alias/edge type) to the PatternMatchElementExpression because it represents "to/from" pair on the pattern expression

PR #7161, opened 2018-07-26
RavenDB-11635 Handle Load / Include calls in static indexes etag gene…

ayende:
…ration as well

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -2092,7 +2092,7 @@ namespace Raven.Server.Documents.Indexes
                             }
                         }
 
-                        FillQueryResult(resultToFill, isStale, documentsContext, indexContext);
+                        FillQueryResult(resultToFill, isStale, query.Metadata, documentsContext, indexContext);
 
                         using (var reader = IndexPersistence.OpenIndexReader(indexTx.InnerTransaction))
                         {
@@ -2265,7 +2265,9 @@ namespace Raven.Server.Documents.Indexes
                                 continue;
                             }
 
-                            FillFacetedQueryResult(result, IsStale(documentsContext, indexContext), facetQuery.FacetsEtag, documentsContext, indexContext);
+                            FillFacetedQueryResult(result, IsStale(documentsContext, indexContext), 
+                                facetQuery.FacetsEtag, facetQuery.Query.Metadata,
+                                documentsContext, indexContext);
 
                             documentsContext.CloseTransaction();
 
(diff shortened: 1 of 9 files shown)


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -2610,21 +2612,37 @@ private static bool WillResultBeAcceptable(bool isStale, IndexQueryBase<Blittabl
...
+        {
+            if (q?.HasIncludeOrLoad == true)
+            {
+                Debug.Assert(length > sizeof(long) * 3);
+
+                long* buffer = (long*)indexEtagBytes;
+                var numberOfDocuments = documentsContext.DocumentDatabase.DocumentsStorage.GetNumberOfDocuments(documentsContext);
+                buffer[0] = DocumentsStorage.ReadLastDocumentEtag(documentsContext.Transaction.InnerTransaction);
+                buffer[1] = DocumentsStorage.ReadLastTombstoneEtag(documentsContext.Transaction.InnerTransaction);
+                buffer[2] = numberOfDocuments;
```
arekpalinski, line 2642: `buffer[2]` is 

https://github.com/ravendb/ravendb/blob/d0c1c05d8e45b6324d44d0889ae9f75369f9c0b7/src/Raven.Server/Documents/Indexes/Index.cs#L2585

https://github.com/ravendb/ravendb/blob/d0c1c05d8e45b6324d44d0889ae9f75369f9c0b7/src/Raven.Server/Documents/Indexes/Index.cs#L2578

It means you won't detect index changes if there is no updates to documents. That's wrong.

BTW why to override the values here instead of appending them at the end of `indexEtagBytes` buffer?

ayende: I'm setting this here because this is only valid if there is load / include.
The idea is that I'm using these three values to know if there was any change _to any document_, not just the ones covered by the index.

How can there _be_ index changes if the docs haven't been updated?

arekpalinski: Because indexing might be still in-progress. Let's say that you put a new index into db with load/include and you start querying it. There are no doc changes but indexing batches change
the last indexed etag

arekpalinski: @ayende Please make sure you push https://github.com/ravendb/ravendb/pull/7162/commits/77fbb4d7e88ea66d2e7176eccf9642b525ed4773 here as well so we won't merge this 4.1 PR without it

arekpalinski: I'll handle that

arekpalinski: https://github.com/ravendb/ravendb/pull/7166

ayende: Thanks, got distracted

PR #7162, opened 2018-07-26
RavenDB-11635 Handle Load / Include calls in static indexes etag gene…

ayende:
…ration as well

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -2566,13 +2566,13 @@ namespace Raven.Server.Documents.Indexes
         {
             if (q?.HasIncludeOrLoad == true)
             {
-                Debug.Assert(length > sizeof(long) * 3);
+                Debug.Assert(length > sizeof(long) * 4);
 
                 long* buffer = (long*)indexEtagBytes;
-                var numberOfDocuments = documentsContext.DocumentDatabase.DocumentsStorage.GetNumberOfDocuments(documentsContext);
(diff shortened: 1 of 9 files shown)


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -2543,21 +2545,37 @@ private static bool WillResultBeAcceptable(bool isStale, IndexQueryBase<Blittabl
...
+        protected virtual unsafe long CalculateIndexEtag(DocumentsOperationContext documentsContext,
+            TransactionOperationContext indexContext, QueryMetadata q, bool isStale)
         {
             var length = MinimumSizeForCalculateIndexEtagLength();
 
             var indexEtagBytes = stackalloc byte[length];
 
             CalculateIndexEtagInternal(indexEtagBytes, isStale, State, documentsContext, indexContext);
 
+            UseAllDocumentsEtag(documentsContext, q, length, indexEtagBytes);
```
maximburyak, line 2557: I'm not sure I follow: After we set all the values for the collections in the "indexEtagBytes", we then just override the first 16 bytes with some other values? I know that we are hashing it all at the end, but we do have higher chance for collision here, wouldn't we want to just have 16 more bytes for that usage?

ppekrol: There is no need. We are overriding the lastDocumentEtag/lastTombstoneEtag from the 'first' collection that index is working on with lastDocumentEtag/lastTombstoneEtag for all docs. At the end this gives us needed uniqueness without the need of extending the buffer length.

PR #7164, opened 2018-07-26
RavenDB-11624 Check cluster version before issuing a clean up command

karmeli87:


src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
```diff
@@ -342,6 +342,12 @@ class ClusterObserver : IDisposable
 
         private long? CleanUpDatabaseValues(string database, DatabaseRecord record, Dictionary<string, ClusterNodeStatusReport> stats)
         {
+            if (ClusterCommandsVersionManager.CurrentClusterMinimalVersion <
```
ppekrol, line 345: What if
- `CurrentClusterMinimalVersion` will be 41_000
- `ClusterCommandsVersionManager.ClusterCommandsVersions[nameof(CleanUpClusterStateCommand)])` will be 41_001 ?

Shouldn't we use a threshold value e.g. `ClusterCommandsVersionManager.CurrentClusterMinimalVersion` < 41_000 ?

karmeli87: if `ClusterCommandsVersionManager.ClusterCommandsVersions[nameof(XXX)])` will be greater than the `CurrentClusterMinimalVersion`, the command is not supported in the cluster..

ayende: If the command version is higher than the cluster min version, then we can't use that.

PR #7180, opened 2018-07-27
RavenDB-11515 Provide a way to filter deletions in ETL

arekpalinski:


src/Raven.Client/Documents/Operations/ETL/Transformation.cs
```diff
@@ -154,6 +159,50 @@ public virtual bool Validate(ref List<string> errors, EtlType type)
...
+                        CollectionToDeleteDocumentsBehaviorFunction = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+
+                        for (int i = 0; i < deleteBehaviors.Count; i++)
+                        {
+                            var deleteBehaviorFunction = deleteBehaviors[i];
+
+                            if (deleteBehaviorFunction.Groups.Count != 2)
+                            {
+                                errors.Add(
+                                    "Invalid load counters behavior function. It is expected to have the following signature: " +
```
ayende: the error seems be to wrong, using delete function but talking bout load counters

arekpalinski: fixed

ayende (approved):

PR #7185, opened 2018-07-29
RavenDB-11640 Bug fixed + test - lazy string escape characters

haludi:
BlittableMetadataModifier changed to add information about escape characters of id when there is none

### DIFF
--- a/src/Raven.Server/Documents/BlittableMetadataModifier.cs
+++ b/src/Raven.Server/Documents/BlittableMetadataModifier.cs
@@ -86,6 +86,11 @@ namespace Raven.Server.Documents
             {
                 lazyStringValueFromParserState.EscapePositions = state.EscapePositions.ToArray();
             }
+            else
+            {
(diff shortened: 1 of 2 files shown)


src/Raven.Server/Documents/BlittableMetadataModifier.cs
```diff
@@ -86,6 +86,11 @@ private unsafe LazyStringValue CreateLazyStringValueFromParserState(JsonParserSt
             {
                 lazyStringValueFromParserState.EscapePositions = state.EscapePositions.ToArray();
             }
+            else
+            {
+                lazyStringValueFromParserState.EscapePositions = Array.Empty<int>();
```
ppekrol, line 91: IIRC it can be set to `null`?

haludi: It is assume that if the EscapePositions is null so the information about the escape characters placed in the bytes after the string bytes but if no such information were written this bytes are just garbage

PR #7188, opened 2018-07-29
[4.0] RavenDB-11644 Fix TCP versioning negotiation

karmeli87:
(cherry-pick from 4.1)

- If the version wasn't exist we would drop the connection, instead of keep searching.
- Some code refactoring, make it all just sync.


talweiss1982:
Why don't we try to look for a matching version using our own TCP version and going back from that?
This means that old server could never communicate with new servers meaning we only have backward compatibility and not forward compatibility and I think we should support this.

PR #7190, opened 2018-07-30
[4.0] RavenDB-11643 Let's decompress the page that we couldn't find precomp…

arekpalinski:
…uted results for. If it's empty then skip throwing the exception.

### DIFF
--- a/src/Raven.Server/Documents/Indexes/MapReduce/ReduceMapResultsBase.cs
+++ b/src/Raven.Server/Documents/Indexes/MapReduce/ReduceMapResultsBase.cs
@@ -340,7 +340,7 @@ namespace Raven.Server.Documents.Indexes.MapReduce
 
                         var parentPage = tree.GetParentPageOf(page);
 
-                        using (var result = AggregateBranchPage(page, table, indexContext, branchesToAggregate, compressedEmptyLeafs, failedAggregatedLeafs, token))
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Documents/Indexes/MapReduce/ReduceMapResultsBase.cs
```diff
@@ -545,12 +547,37 @@ private void StoreAggregationResult(TreePage page, Table table, AggregationResul
...
+                using (var decompressed = tree.DecompressPage(relatedTreePage, skipCache: true))
+                {
+                    if (decompressed.NumberOfEntries == 0)
+                    {
+                        // it's empty so there is no related aggregation result, we can safely skip it
+
+                        return false;
+                    }
+
+                    decompressedDebug = decompressed.ToString();
```
talweiss1982, line 567: @arekpalinski is this a scenario that should never happen but seems to sometime happen?

arekpalinski: Yes, kind of. It happened on prod. Having a page that is empty after decompressing it is valid case. I believe we'll remove it eventually:

https://github.com/ravendb/ravendb/blob/10196851fc9fafb6206e5eb91ac9b9d62c58e9bd/src/Raven.Server/Documents/Indexes/MapReduce/ReduceMapResultsBase.cs#L396-L415

It's safe to add this check and skip it here. Although I'm not sure that was the case - couldn't reproduce it locally.

PR #7193, opened 2018-07-30
RavenDB-11476 Gracefully shutdown server

IdanHaim:
- Add new server start flag --non-console
- Add GracefullyExitTimeout to ServerOptions
- When dispose the server we will try to Gracefully exit
- Include Embedded to Tryouts and include internals


src/Raven.Embedded/ServerOptions.cs
```diff
@@ -19,6 +19,8 @@ public class ServerOptions
 
         public string ServerUrl { get; set; }
 
+        public int GracefullyExitTimeout { get; set; } = 3;
```
ppekrol, line 22: Let's change it to TimeSpan
Rename it to GracefulShutdownTimeout

ppekrol: Default value should be 30 sec

src/Raven.Server/Program.cs
```diff
@@ -277,7 +277,8 @@ private static bool RunInteractive(RavenServer server)
             //stop dumping logs
             LoggingSource.Instance.DisableConsoleLogging();
 
-            return new RavenCli().Start(server, Console.Out, Console.In, true);
+            var consoleColoring = !CommandLineSwitches.NonConsole;
+            return new RavenCli().Start(server, Console.Out, Console.In, consoleColoring);
```
ppekrol, line 281: What is the issue with coloring?

src/Raven.Server/Utils/Cli/CommandLineSwitches.cs
```diff
@@ -100,6 +104,11 @@ public static string[] Process(string[] args)
                 "-l | --log-to-console",
                 "Print logs to console (when run in non-interactive mode)",
                 CommandOptionType.NoValue);
+            _nonConsoleOption = _app.Option(
+                "--non-console",
```
ppekrol, line 108: --no-color ?

src/Raven.Embedded/EmbeddedServer.cs
```diff
@@ -81,6 +84,19 @@ public Task<IDocumentStore> GetDocumentStoreAsync(string database, CancellationT
             return GetDocumentStoreAsync(new DatabaseOptions(database), token);
         }
 
+        public void Quit()
```
ppekrol, line 87: Remove this method?

src/Raven.Embedded/EmbeddedServer.cs
```diff
@@ -144,10 +160,22 @@ private void KillSlavedServerProcess(Process process)
                 return;
 
             if (_logger.IsInfoEnabled)
```
ppekrol, line 162: Rename this method to 'ShutdownServerProcess'

src/Raven.Embedded/EmbeddedServer.cs
```diff
@@ -144,10 +160,22 @@ private void KillSlavedServerProcess(Process process)
...
-                _logger.Info($"Killing global server PID { process.Id }.");
+                _logger.Info($"Try shutdown server PID { process.Id } gracefully.");
+
+            using (var inputStream = process.StandardInput)
+            {
+                inputStream.WriteAsync($"q{Environment.NewLine}y{Environment.NewLine}");
```
ppekrol, line 167: You are not awaiting this write and disposing the stream afterwards

src/Raven.Embedded/EmbeddedServer.cs
```diff
@@ -144,10 +160,22 @@ private void KillSlavedServerProcess(Process process)
                 return;
 
             if (_logger.IsInfoEnabled)
-                _logger.Info($"Killing global server PID { process.Id }.");
+                _logger.Info($"Try shutdown server PID { process.Id } gracefully.");
+
+            using (var inputStream = process.StandardInput)
```
ppekrol, line 165: This entire block should be in try-catch

ppekrol (changes requested):

src/Raven.Server/Program.cs
IdanHaim: When we use the embedded we redirect the stdin output.
When coloring is true we do some checks when we try to write something and then we get an exception.

src/Raven.Server/Utils/Cli/CommandLineSwitches.cs
IdanHaim: The idea was to write this flag when we are using a program without any console to write too,
so non-console will be better

src/Raven.Embedded/EmbeddedServer.cs
IdanHaim: will do

src/Raven.Server/Utils/Cli/CommandLineSwitches.cs
ppekrol: Still, it would be better to go with `--no-color` because this is what it does actually. It can be used outside of embedded.

IdanHaim:
Very Confusing the with non-console user will understand why because he
will get an exception that tell him that he don't have console like I did.

ppekrol:
This is transparent for embedded user. And for a non-embedded user this option would be confusing.

IdanHaim:
And non-color will be better? We can change the description maybe. And then
we can use it in the future for no console application that have stdin
redirect

ppekrol:
`--console-no-color` would be best

PR #7194, opened 2018-07-31
[4.1] - RavenDB-11642 - Setup wizard on macOS issues

grisha-kotler:
http://issues.hibernatingrhinos.com/issue/RavenDB-11642

### DIFF
--- a/src/Raven.Server/Utils/Cli/BrowserHelper.cs
+++ b/src/Raven.Server/Utils/Cli/BrowserHelper.cs
@@ -6,7 +6,7 @@ namespace Raven.Server.Utils.Cli
 {
     public static class BrowserHelper
     {
-        public static bool OpenStudioInBrowser(string url, out string error)
+        public static bool OpenStudioInBrowser(string url, Action<object> onError = null)
         {
             try
             {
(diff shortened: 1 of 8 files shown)


src/Raven.Server/Utils/Cli/BrowserHelper.cs
```diff
@@ -6,22 +6,38 @@ namespace Raven.Server.Utils.Cli
...
-        public static void OpenStudioInBrowser(string url)
+        public static bool OpenStudioInBrowser(string url, Action<object> onError = null)
         {
             try
             {
                 if (PlatformDetails.RunningOnPosix == false)
                 {
                     Process.Start(new ProcessStartInfo("cmd", $"/c start {url}"));
                 }
+                else if (PlatformDetails.RunningOnMacOsx)
```
ppekrol, line 17: In 4.1 there is a similar method in EmbeddedServer. Can you apply this change there also?

ppekrol (changes requested):

src/Raven.Server/Utils/Cli/BrowserHelper.cs
grisha-kotler: done

PR #7210, opened 2018-08-01
RavenDB-11594 Add Use Experimental Features check-box in the setup wi…

ml054:
…zard


ml054:
I will change bg-color + add icon

PR #7217, opened 2018-08-01
RavenDB-11476 Gracefully shutdown server

IdanHaim:
- Add GracefullyExitTimeout to ServerOptions
- When dispose the server we will try to Gracefully exit
- Include Embedded to Tryouts and include internals


ayende (changes requested):
Exact same comments as the previous one

ayende:
Still missing handling of `WaitForExit` returning false

PR #7218, opened 2018-08-01
[4.0] RavenDB-11614

aviv86:


test/InterversionTests/ClusterTests.cs
```diff
@@ -0,0 +1,154 @@
...
+{
+    public class ClusterTests : MixedClusterTestBase
+    {
+        [Fact]
+        public async Task ReplicationInMixedCluster_41Leader()
+        {
+            var (leader, peers, local) = await CreateMixedCluster(new[]
+            {
+                "4.1.0-nightly-20180801-0430",
+                "4.1.0-nightly-20180801-0430"
```
arekpalinski, line 22: We'll be better off by using `[Theory]` and `[InlineData("4.1.0-nightly-20180801-0430", "4.1.0-nightly-20180801-0430"]` for that so we'll be able to add another versions in the future easily

PR #7222, opened 2018-08-01
[4.0] RavenDB-11632 handle delete and store for same document inside session

lahma:


test/SlowTests/Issues/RavenDB_11632.cs
```diff
@@ -0,0 +1,98 @@
...
+            {
+                using (var session = store.OpenSession())
+                {
+                    session.Store(new Document
+                    {
+                        Id = id,
+                        Text = "text-1"
+                    });
+                    
+                    var doc = session.Load<Document>(id);
```
ayende, line 69: This load it from the local session, mind

test/SlowTests/Issues/RavenDB_11632.cs
```diff
@@ -0,0 +1,98 @@
...
+
+                    session.Delete(doc);
+
+                    doc = new Document
+                    {
+                        Id = id,
+                        Text = "text-2"
+                    };
+                    
+                    session.Store(doc);
```
ayende: I'm really nervous about something like that. It opens up the option that you'll be mislead by the code.
In your issue, you mentioned passing the id explicitly, which I think is a good compromise here.

lahma: I'll see if I can make that work, thanks for the review.

test/SlowTests/Issues/RavenDB_11632.cs
lahma: That actually was the intent, to make sure local session returns the state that will be deleted, maybe a bit silly.

test/SlowTests/Issues/RavenDB_11632.cs
ppekrol: I do not think that this is the flow that we want to support.

Supporting following operations within a session is very dangerous
```
// load ID
// delete ID
// store ID
// saveChanges
```

What we should support is:
```
// load ID
// delete ID
// saveChanges
// store ID
// saveChanges
```

lahma: The case I'm experiencing is adapting Quartz.NET's job store logic to pass connection (UoW) around template methods that handle various tasks, some might store, some might delete, the end result is then stored as transaction (session.SaveChanges()).

So it might not be obvious if composition caused load-delete-store combination, but SaveChanges should be the transaction.

lahma: @ppekrol How you model this kind of templated process, where I need to support single transaction spanning composed methods? Composition methods are generic methods having generic type for passed connection/tx pair.

```
var uow = BeginTransactionalOperation(); // holder for ADO.NET connection+tx or Quartz.NET session
var job = LoadJob(trigger.JobId, uow);
var triggger = LoadJob(trigger.Id, uow);
MaybeDeleleteJobIfStale(job, uow);
MaybeAddJob(trigger, job, uow);
SomeOtherThing(trigger, job, uow);
uow.Commit();
```

lahma: I've updated the PR to to only allow such overwrite if explicit id is provided, otherwise the goold old exception. I'll leave it to you to decide whether it makes sense or not - feel free to close and create and alternative solution. I don't get paid enough from OSS to fight the persistence layer 😄

ppekrol: I will look tomorrow at this, but there is also a notion of deferred commands that needs to be taken into account. They are executed 'after' the session operations.

Regarding your example, you could always do:
```
var uow = BeginTransactionalOperation(); // holder for ADO.NET connection+tx or Quartz.NET session
var job = LoadJob(trigger.JobId, uow);
var trigger = LoadJob(trigger.Id, uow);
var delete = ShouldDeleleteJobIfStale(job, uow);
var added = MaybeAddJob(trigger, job, uow); // modifies 'job' entity
if (added == false && deleted)
{
    DeleteJob(job, uow);
}
SomeOtherThing(trigger, job, uow);
uow.Commit();
```

ppekrol:
@lahma At this moment we do not want to introduce such changes to the session behavior. Could you push your Quartz code to a branch so we can review it and fix it on Quartz-side?

PR #7230, opened 2018-08-01
Fix: 3yo bug on ByteString ``.ToString()`` method

redknightlois:
Avoid the debugger to crash the process when trying to convert a ByteString to string.

### DIFF
--- a/src/Sparrow/ByteString.cs
+++ b/src/Sparrow/ByteString.cs
@@ -291,7 +291,7 @@ namespace Sparrow
 
             EnsureIsNotBadPointer();
 
-            return new string((char*)_pointer->Ptr, 0, _pointer->Length);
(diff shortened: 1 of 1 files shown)


src/Sparrow/ByteString.cs
```diff
@@ -284,14 +284,14 @@ public void CopyTo(int from, byte[] dest, int offset, int count)
             }
         }
 
-        public override string ToString()
```
ayende: Huh? Why aren't you overriding tostring?

redknightlois: Because I copy and pasted, and made a mistake...

redknightlois:
Fixed

ppekrol:
Still there :)

redknightlois:
I am using a new git client (GitKraken) for some things is great, for others suck big time.

PR #7242, opened 2018-08-02
[4.1] RavenDB-11415 Rachis and cluster transaction fixes

karmeli87:


src/Raven.Client/ServerWide/DatabaseRecord.cs
```diff
@@ -70,7 +70,7 @@ public DatabaseRecord(string databaseName)
 
         public ClientConfiguration Client;
 
-        public long TruncatedClusterTransactionIndex;
+        public long TruncatedClusterTransactionCommandsCount;
```
ayende, line 73: This is serialized, right? If so, it is fine to make a change here, since 4.1 isn't released yet, but need to make sure to remember this in the future.

ayende: Note that our production system already has a 4.1 instance running tryouts, so need to check that this won't impact it there.

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -363,90 +366,106 @@ private void ExectueClusterTransactionTask()
...
+                        if (batch.Count == 0)
+                        {
+                            continue;
+                        }
+                        
+                        try
+                        {
+                            var mergedCommands = new BatchHandler.ClusterTransactionMergedCommand(this, batch);
+                            TxMerger.Enqueue(mergedCommands).Wait(DatabaseShutdown);
+                            foreach (var command in batch)
```
ayende: Only the `Wait` here (for the completion of the commands) need to be in the try catch for the one at a time run.

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -363,90 +366,106 @@ private void ExectueClusterTransactionTask()
...
+                var mergedCommand = new BatchHandler.ClusterTransactionMergedCommand(this, singleCommand);
+                try
                 {
-                    _logger.Info($"Can't perform cluster transaction on database '{Name}'.", e);
+                    TxMerger.Enqueue(mergedCommand).Wait(DatabaseShutdown);
+                    OnClusterTransactionCompletion(command.Index, mergedCommand);
+                    _nextClusterCommand = command.PreviousCount + command.Commands.Length;
+                }
+                catch (Exception e)
+                {
```
ayende, line 428: Need to send notification here, if the situation persists.

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -137,22 +138,40 @@ private async Task HandleClusterTransaction(DocumentsOperationContext context, M
                 throw new ConcurrencyException($"Failed to execute cluster transaction due to the following issues: {string.Join(Environment.NewLine, errors)}");
             }
 
+            // wait for the command to be applied on this node
+            await ServerStore.WaitForCommitIndexChange(RachisConsensus.CommitIndexModification.GreaterOrEqual, result.Index);
+
             var array = new DynamicJsonArray();
             if (clusterTransactionCommand.DatabaseCommandsCount > 0)
             {
+#if DEBUG
```
ayende: This needs to be removed, even in debug only.
It's fine if you have an error / assert, but console write line are not a good idea in our code except while trouble shooting.

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
```diff
@@ -305,7 +325,7 @@ private static unsafe SingleClusterDatabaseCommand ReadCommand(TransactionOperat
             var ptr = reader.Read((int)TransactionCommandsColumn.Commands, out var size);
             if (ptr == null)
                 return null;
-            var blittable = new BlittableJsonReaderObject(ptr, size, context);
+            var blittable = new BlittableJsonReaderObject(ptr, size, context).Clone(context);
```
ayende: Is the clone required now?

ayende (changes requested):

src/Raven.Client/ServerWide/DatabaseRecord.cs
karmeli87: Good point. Don't think that this property has propagated to the DatabaseRecord into production.
(haven't saw it on the DatabaseRecord)

PR #7243, opened 2018-08-02
[4.0] RavenDB-11415 fix rachis

karmeli87:
- Concurrecy exception due to term change should close the ambassador.
- Retry to put the command on NotLeadingException during PutAsync.
- Fix 'System.InvalidOperationException: Collection was modified; enumerat

### DIFF
--- a/src/Raven.Server/Rachis/Leader.cs
+++ b/src/Raven.Server/Rachis/Leader.cs
@@ -48,12 +48,12 @@ namespace Raven.Server.Rachis
         private readonly ManualResetEvent _noop = new ManualResetEvent(false);
         private long _lowestIndexInEntireCluster;
 
-        private readonly Dictionary<string, FollowerAmbassador> _voters =
-            new Dictionary<string, FollowerAmbassador>(StringComparer.OrdinalIgnoreCase);
-        private readonly Dictionary<string, FollowerAmbassador> _promotables =
-            new Dictionary<string, FollowerAmbassador>(StringComparer.OrdinalIgnoreCase);
-        private readonly Dictionary<string, FollowerAmbassador> _nonVoters =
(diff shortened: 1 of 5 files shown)


ayende (approved):

src/Raven.Server/Rachis/Leader.cs
```diff
@@ -184,15 +189,15 @@ private void RefreshAmbassadors(ClusterTopology clusterTopology, Dictionary<stri
                     if (old.TryGetValue(voter.Key, out FollowerAmbassador existingInstance))
                     {
                         existingInstance.UpdateLeaderWake(_voterResponded);
-                        _voters.Add(voter.Key, existingInstance);
+                        _voters[voter.Key] = existingInstance;
```
ppekrol, line 192: I think that you wanted to use `.AddOrUpdate`. The `this[key]` is an equivalent of using `.TryAdd`.

Check here: https://github.com/dotnet/corefx/blob/23915a606017197aed1d24960096b532d24d626e/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs#L888

ppekrol (changes requested):

src/Raven.Server/Rachis/Leader.cs
karmeli87: No, this should work as expected.

karmeli87: https://msdn.microsoft.com/en-us/library/dd267253(v=vs.110).aspx

ppekrol: Correct. I just noticed that they are passing additional flag when this method is used.

PR #7249, opened 2018-08-04
Graph API WIP

ayende:


src/Raven.Server/Documents/Queries/Graph/GraphQueryRunner.cs
```diff
@@ -0,0 +1,193 @@
...
+                var withClauseResults = new Dictionary<string, List<Match>>();
+                foreach (var withClause in Query.WithDocumentQueries)
+                {
+                    var withClauseQueryResult =
+                        await withQueryRunner.ExecuteQuery(new IndexQueryServerSide(withClause.Value.ToString()), ctx, null, OperationCancelToken.None);
+                    var queryMatches = new List<Match>();
+                    foreach (var result in withClauseQueryResult.Results)
+                    {
+                        var match = new Match();
+                        match.Set(withClause.Key, result.Data);
```
talweiss1982: I'm trying to follow the logic here, you add the key of the clause which I assume is the alias and the document data but the Match.Set will add this to a dictionary and this is done per Query result so i think Match.Set will throw if you add two documents to the same alias (I didn't run the code)

talweiss1982: Also, I wonder about adding those blittables to the Dictionary while the context is been dispose of very soon after, shouldn't we either clone the data or have a longer living context here?

src/Raven.Server/Documents/Queries/GraphQueryRunner.cs
```diff
@@ -0,0 +1,256 @@
+﻿using System;
```
talweiss1982, line 1: I'm not sure which GraphQueryRunner.cs is the right one but the other should probably be removed or renamed if it something else (this seems to be the working version)

src/Raven.Server/Documents/Queries/GraphQueryRunner.cs
```diff
@@ -0,0 +1,256 @@
...
+            foreach (var documentQuery in q.GraphQuery.WithDocumentQueries)
+            {
+                var queryMetadata = new QueryMetadata(documentQuery.Value, query.QueryParameters, 0);
+                var results = await Database.QueryRunner.ExecuteQuery(new IndexQueryServerSide(queryMetadata),
+                    documentsContext, existingResultEtag, token);
+
+                foreach (var result in results.Results)
+                {
+                    var match = new Match();
+                    match.Set(documentQuery.Key, result);
```
talweiss1982: I'm pretty sure this will throw if the query return more than one result

src/Raven.Server/Documents/Queries/GraphQueryRunner.cs
```diff
@@ -0,0 +1,256 @@
...
+                    match.Populate(ref ir);
+                }
+            }
+
```
talweiss1982, line 57: TODO: need to handle With Edges

src/Raven.Server/Documents/Queries/GraphQueryRunner.cs
```diff
@@ -0,0 +1,256 @@
...
+                });
+
+            }
+            final.TotalResults = final.Results.Count;
+            return final;
+        }
+
+        private List<Match> ExecutePatternMatch(DocumentsOperationContext documentsContext, Query q, IntermediateResults ir)
+        {
+            var results = new DocumentQueryResult();
```
talweiss1982: This is not used

src/Raven.Server/Documents/Queries/GraphQueryRunner.cs
```diff
@@ -0,0 +1,256 @@
...
+                aliasDic[instance.Id] = match;
+            }
+
+            public bool TryGetByAlias(string alias, out Dictionary<string,Match> value)
+            {
+                return MatchesByAlias.TryGetValue(alias, out value);
+            }
+        }
+
+        private class GraphExecuteVisitor : QueryVisitor
```
talweiss1982, line 144: I think that this class is entitled to its own file

PR #7251, opened 2018-08-05

iftahbe:


src/Raven.Server/Rachis/AppendEntries.cs
```diff
@@ -13,6 +13,7 @@ public class AppendEntries
         public bool ForceElections { get; set; }
         public long TimeAsLeader { get; set; }
         public int SendingThread { get; set; }
+        public int MinClusterCommandVersion { get; set; }
```
ayende: Why send it here and not during the negotiation process?

src/Raven.Server/Rachis/Follower.cs
```diff
@@ -83,7 +84,9 @@ private void FollowerSteadyState()
 
                         return;
                     }
-                    
+
+                    ClusterCommandsVersionManager.SetClusterVersion(appendEntries.MinClusterCommandVersion);
```
ayende: This doesn't change after election, I would expect this during negotiations, not every time.

src/Raven.Server/RavenServer.cs
```diff
@@ -487,28 +486,50 @@ private async Task DoActualCertificateRefresh(CertificateHolder currentCertifica
                 if (ServerStore.IsLeader() == false)
                     return;
 
+                if (ClusterCommandsVersionManager.CurrentClusterMinimalVersion < ClusterCommandsVersionManager.ClusterCommandsVersions[nameof(ConfirmServerCertificateReplacedCommand)])
+                    throw new ClusterNodesVersionMismatchException("It is not possible to refresh/replace the cluster certificate in the current cluster topology. Please make sure that all the cluster nodes run the same version.");
```
ayende: Here you need to give the actual version and expected version

src/Raven.Server/RavenServer.cs
```diff
@@ -487,28 +486,50 @@ private async Task DoActualCertificateRefresh(CertificateHolder currentCertifica
                 if (ServerStore.IsLeader() == false)
                     return;
 
+                if (ClusterCommandsVersionManager.CurrentClusterMinimalVersion < ClusterCommandsVersionManager.ClusterCommandsVersions[nameof(ConfirmServerCertificateReplacedCommand)])
```
ayende: Use `TryGetValue` instead of indexer, will throw otherwise.

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -392,16 +401,88 @@ private void InstallUpdatedServerCertificate(TransactionOperationContext context
...
+                    {
+                        [Constants.Certificates.Replaced] = replaced + 1
+                    };
+
+                    certInstallation = context.ReadObject(certInstallation, "server.cert.update");
+
+                    UpdateValue(index, items, key, key, certInstallation);
+
+                    if (_parent.Log.IsOperationsEnabled)
+                        _parent.Log.Operations("Confirming that certificate replacement has happened.");
```
ayende: Add what certificate and from what certificate.

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -865,16 +851,36 @@ private void OnValueChanged(object sender, (long Index, string Type) t)
...
                                 $"The server certificate was successfully replaced on node {NodeTag}.",
                                 AlertType.Certificates_ReplaceSuccess,
-                                NotificationSeverity.Success,
-                                "Cluster.Certificate.Replace.Success"));
+                                NotificationSeverity.Success));
 
                             if (Logger.IsOperationsEnabled)
                                 Logger.Operations($"The server certificate was successfully replaced on node {NodeTag}.");
+
+                            if (ClusterCommandsVersionManager.CurrentClusterMinimalVersion < ClusterCommandsVersionManager.ClusterCommandsVersions[nameof(ConfirmServerCertificateReplacedCommand)])
```
ayende: use `TryGetValue`

I also don't like `CurrentClusterMinimalVersion` in general. This is us setting a global static property. I would rather have it in the engine itself.

src/Raven.Server/Rachis/AppendEntries.cs
karmeli87: So we can change the cluster version on the fly. The simplest example is when we have a 4.0 cluster and we want to update it to 4.1.
So when updating the last node we need to change the whole cluster to 4.1 and let the follower know about it,

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
```diff
@@ -786,6 +788,67 @@ public Task ClusterDomains()
...
+                    writer.WriteBool(replaceImmediately);
+                    writer.WriteComma();
+                    writer.WritePropertyName(Constants.Certificates.Replaced);
+                    writer.WriteInteger(replaced);
+                    writer.WriteComma();
+                    writer.WriteEndObject();
+                }
+                else
+                {
+                    return NoContent();
```
ppekrol: Not Found

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
```diff
@@ -786,6 +788,67 @@ public Task ClusterDomains()
...
+                    writer.WriteComma();
+                    writer.WritePropertyName(Constants.Certificates.OldThumbprint);
+                    writer.WriteString(oldThumbprint);
+                    writer.WriteComma();
+                    writer.WritePropertyName(Constants.Certificates.ReplaceImmediately);
+                    writer.WriteBool(replaceImmediately);
+                    writer.WriteComma();
+                    writer.WritePropertyName(Constants.Certificates.Replaced);
+                    writer.WriteInteger(replaced);
+                    writer.WriteComma();
```
ppekrol: Please remove this comma. You are generating invalid JSON.

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -939,12 +943,94 @@ private void OnValueChanged(object sender, (long Index, string Type) t)
...
                             AlertType.Certificates_ReplaceError,
                             NotificationSeverity.Error,
-                            "Cluster.Certificate.Replace.Error",
-                            new ExceptionDetails(e)));
+                            details: new ExceptionDetails(e)));
+                    }
+                    break;
+                case nameof(RecheckStatusOfServerCertificateReplacementCommand):
+                case nameof(ConfirmServerCertificateReplacedCommand):
+                    try
```
ppekrol: Wrap this inside a method?

src/Raven.Client/Constants.cs
```diff
@@ -88,6 +88,13 @@ private Certificates()
             }
 
             public const string Prefix = "certificates/";
+            public const string Certificate = "Certificate";
+            public const string Thumbprint = "Thumbprint";
+            public const string OldThumbprint = "OldThumbprint";
+            public const string Confirmations = "Confirmations";
+            public const string Replaced = "Replaced";
+            public const string ReplaceImmediately = "ReplaceImmediately";
+            public const string CertReplaceAlertTitle = "Server Certificate Replacement";
```
ppekrol: Please remove this from Constants. This is not a place for 'property names' and 'titles'.

ppekrol (changes requested):

src/Raven.Client/ServerWide/Operations/Certificates/CertificateReplacement.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using Sparrow.Json.Parsing;
+
+namespace Raven.Client.ServerWide.Operations.Certificates
+{
+    public class CertificateReplacement
```
ppekrol: Please move it from the client. It is not used here. Only on server-side.

src/Raven.Server/Rachis/AppendEntries.cs
ppekrol: Rename to 'MinCommandVersion' maybe? Less to send over the wire and you know that this is 'cluster' version because of the context.

src/Raven.Client/ServerWide/Operations/Certificates/CertificateReplacement.cs
ppekrol: Also, for future reference, when you are inserting internally used code to Client, then it is safer to mark it as 'internal' to avoid backward compatibility issues.

src/Raven.Server/Rachis/AppendEntries.cs
ayende: You cannot change the cluster version on the fly. 
The cluster version is fixed on a per term basis.
Therefor, if we need to update the 4.0 to 4.1, we will have to wait until the next election.

src/Raven.Server/Rachis/FollowerAmbassador.cs
```diff
@@ -730,7 +729,9 @@ private long InitialNegotiationWithFollower()
                 var llr = _connection.Read<LogLengthNegotiationResponse>(context);
 
                 FollowerCommandsVersion = GetFollowerVersion(llr);
-                ClusterCommandsVersionManager.SetMinimalClusterVersion(FollowerCommandsVersion);
+                _engine.CurrentLeader.PeersVersion[_tag] = FollowerCommandsVersion;
```
ayende, line 732: Here, have the state on the Follower, and the leader will read it during the candidate's winning elections

src/Raven.Server/Rachis/AppendEntries.cs
iftahbe: I will address the command version related issues in the next PR because it is not really related to the certificate replacement command.
See: http://issues.hibernatingrhinos.com/issue/RavenDB-11694

src/Raven.Server/Rachis/FollowerAmbassador.cs
iftahbe: Will address this in http://issues.hibernatingrhinos.com/issue/RavenDB-11694

PR #7254, opened 2018-08-05
[4.1] RavenDB-11415 Raft command will throw timeout exception only when the…

karmeli87:
… command wasn't consumed.

### DIFF
--- a/src/Raven.Server/Rachis/Leader.cs
+++ b/src/Raven.Server/Rachis/Leader.cs
@@ -618,15 +618,27 @@ namespace Raven.Server.Rachis
             }
         }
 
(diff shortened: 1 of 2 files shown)


ayende (approved):

src/Raven.Server/Rachis/Leader.cs
```diff
@@ -682,10 +699,17 @@ private void EmptyQueue()
                 {
                     while (_commandsQueue.TryDequeue(out var cmd))
                     {
```
grisha-kotler, line 701: This will run forever if we have a lot of incoming commands (cluster transactions for example).
It should be limited by the number of transactions or the memory size.

PR #7265, opened 2018-08-06
RavenDB-11107 Give more clear error message when failing to run the server

IdanHaim:


src/Raven.Server/Program.cs
```diff
@@ -187,9 +187,17 @@ public static int Main(string[] args)
                         }
                         catch (Exception e)
                         {
+                            var message =
```
ppekrol: We need to detect that this was a port binding exception.

src/Raven.Server/Program.cs
```diff
@@ -187,9 +187,17 @@ public static int Main(string[] args)
...
+                                $"{e.Message}{Environment.NewLine}Port might be already in use.{Environment.NewLine}Try running with an unused port.{Environment.NewLine}" +
+                                $"You can change the port using one of the following options:{Environment.NewLine}" +
+                                $"1) Run the server from the command line with --ServerUrl option.{Environment.NewLine}" +
+                                $"2) Change the ServerUrl property in setting.json file.{Environment.NewLine}" +
```
ppekrol: This should be swapped with 1)

src/Raven.Server/Program.cs
```diff
@@ -187,9 +187,17 @@ public static int Main(string[] args)
...
+                                $"3) Add RAVEN_ServerUrl in linux or RAVEN.ServerUrl in windows to the Environment Variables.{Environment.NewLine}" +
```
ppekrol: linux -> Linux
windows -> Windows

using `RAVEN_ServerUrl` on both platforms is sufficient.

src/Raven.Server/Program.cs
```diff
@@ -187,9 +187,17 @@ public static int Main(string[] args)
...
+                                "For more information go to https://ravendb.net/docs/article-page/4.0/csharp/server/configuration/configuration-options";
```
ppekrol: Talk with Tom about this. He will generate a link for you.

ppekrol (changes requested):

src/Raven.Server/Program.cs
IdanHaim: Do we want to cover both ways?

src/Raven.Server/Program.cs
IdanHaim: Running a server from the command line will use the args from the command line and not the Setting.json file

src/Raven.Server/Program.cs
IdanHaim: Done

src/Raven.Server/Program.cs
IdanHaim: We can detect this by checking the InnerException e.InnerException will be AddressInUseException

src/Raven.Server/Program.cs
IdanHaim: Going to keep this

src/Raven.Server/Program.cs
ayende: On Linux as well?
What about not permission for low ports on Linux?

PR #7267, opened 2018-08-07
RavenDB-11684 don't build embedded on linux

gregolsky:


build.ps1
```diff
@@ -149,7 +149,8 @@ if (ShouldBuildStudio $STUDIO_OUT_DIR $DontRebuildStudio $DontBuildStudio) {
     write-host "Not building studio..."
 }
 
-if ($JustStudio -eq $False) {
+$IsPosix = $IsWindows -eq $False
```
gregolsky, line 152: Reason it's like this is `$IsWindows` is not always populated for all PS versions, but it's always `$False` for Posix.

PR #7272, opened 2018-08-07
RavenDB-11697 Add an option to specify the node tag when adding the n…

fitzchak:
…ode to the cluster

### DIFF
--- a/src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
@@ -464,7 +464,7 @@ namespace Raven.Server.Documents.Handlers.Admin
                         }
                     }
 
-                    await ServerStore.AddNodeToClusterAsync(nodeUrl, nodeTag, validateNotInTopology: false, asWatcher: watcher ?? false);
+                    await ServerStore.AddNodeToClusterAsync(nodeUrl, nodeTag, validateNotInTopology: true, asWatcher: watcher ?? false);
 
                     using (ctx.OpenReadTransaction())
                     {
--- a/src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
@@ -316,6 +316,7 @@ namespace Raven.Server.Documents.Handlers.Admin
             SetupCORSHeaders();
 
             var nodeUrl = GetQueryStringValueAndAssertIfSingleAndNotEmpty("url");
+            var tag = GetStringQueryString("tag", false);
             var watcher = GetBoolValueQueryString("watcher", false);
             var assignedCores = GetIntValueQueryString("assignedCores", false);
             if (assignedCores <= 0)
@@ -406,7 +407,7 @@ namespace Raven.Server.Documents.Handlers.Admin
                         }
                     }
 
-                    var nodeTag = nodeInfo.NodeTag == RachisConsensus.InitialTag ? null : nodeInfo.NodeTag;
+                    var nodeTag = nodeInfo.NodeTag == RachisConsensus.InitialTag ? tag : nodeInfo.NodeTag;
                     CertificateDefinition oldServerCert = null;
                     X509Certificate2 certificate = null;
 
--- a/src/Raven.Studio/wwwroot/App/views/manage/addClusterNode.html
+++ b/src/Raven.Studio/wwwroot/App/views/manage/addClusterNode.html
@@ -22,7 +22,7 @@
                             <label for="tag" class="control-label">Tag</label>
                             <div class="flex-grow">
                                 <div data-bind="validationElement: nodeTag">
-                                    <input autocomplete="off" type="text" class="form-control" data-bind="textInput: nodeTag" id="tag" placeholder="Manually specify the node tag (optional)">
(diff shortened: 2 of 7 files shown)


src/Raven.Studio/wwwroot/App/views/manage/addClusterNode.html
```diff
@@ -14,7 +14,15 @@
                             <label for="url" class="control-label">URL</label>
                             <div class="flex-grow">
                                 <div data-bind="validationElement: serverUrl">
-                                    <input autocomplete="off" type="text" class="form-control" data-bind="textInput: serverUrl" id="url" placeholder="Enter server url"> 
+                                    <input autocomplete="off" type="text" class="form-control" data-bind="textInput: serverUrl" id="url" placeholder="Enter server url">
+                                </div>
+                            </div>
+                        </div>
+                        <div class="form-group">
+                            <label for="url" class="control-label">TAG</label>
```
ml054: Can we replace TAG-> Tag + changelabel for url - since it is copy & paste issue.

ml054 (changes requested):
@fitzchak Please have a look at my comments.

src/Raven.Studio/wwwroot/App/views/manage/addClusterNode.html
ml054: Do we need validation here? Or tag can be any string?

src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
```diff
@@ -406,7 +407,7 @@ public async Task AddNode()
                         }
                     }
 
-                    var nodeTag = nodeInfo.NodeTag == RachisConsensus.InitialTag ? null : nodeInfo.NodeTag;
+                    var nodeTag = nodeInfo.NodeTag == RachisConsensus.InitialTag ? tag : nodeInfo.NodeTag;
```
ppekrol, line 410: Do we need to validate if node with given tag already exist in the cluster? @karmeli87 ?

src/Raven.Studio/wwwroot/App/views/manage/addClusterNode.html
```diff
@@ -14,7 +14,15 @@
...
+                            <div class="flex-grow">
+                                <div data-bind="validationElement: nodeTag">
+                                    <input autocomplete="off" type="text" class="form-control" data-bind="textInput: nodeTag" id="nodeTag" placeholder="Manually specify node tag, optional">
```
ppekrol: `Manually specify node tag, optional` -> `Manually specify the node tag (optional)` ?

src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
ppekrol: I would also suggest applying `.Trim()` on this

ppekrol (changes requested):

src/Raven.Studio/wwwroot/App/views/manage/addClusterNode.html
fitzchak: Changed

src/Raven.Studio/wwwroot/App/views/manage/addClusterNode.html
fitzchak: Fixed

src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
karmeli87: We do it here:
https://github.com/ravendb/ravendb/blob/cda4226f616ec7dcaac6ab173be79b75c9683134/src/Raven.Server/Rachis/Leader.cs#L905

karmeli87: But we need to validate that the generated node is not in the cluster... 
Because right now we rely on the last node tag, but this is wrong if we expose this to the user.
The user can add node B first then node A and then let the cluster to assign the tag... which will resulted in assigning node B again.
https://github.com/ravendb/ravendb/blob/cda4226f616ec7dcaac6ab173be79b75c9683134/src/Raven.Server/Rachis/Leader.cs#L997

src/Raven.Studio/wwwroot/App/views/manage/addClusterNode.html
fitzchak: I added server side validation.

ppekrol: Can we add a simple JS validation here for better UX?

src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
fitzchak: @karmeli87 I tested and there is no issue here, it remembering the bigger tag as the last node tag.

fitzchak: @ppekrol In case of adding a node on existing node tag, it reports that the node was added successfully, but the actual case was no-op. I'll fix it to throw an error.

src/Raven.Studio/wwwroot/App/views/manage/addClusterNode.html
fitzchak: Done

ppekrol (approved):

PR #7280, opened 2018-08-08
RavenDB-11700 Revert the usage of Lock Free when guideline doesn't apply

redknightlois:


redknightlois:
Good enough to take out the show-stopper tag into 4.1 and 4.0, need to measure carefully to understand the full impact of the changes, but I dont expect radical differences.

PR #7289, opened 2018-08-09
RavenDB-11611 fixing race condition in backup test

talweiss1982:


ayende (approved):

ppekrol:
1. Please check the compilation
2. This is a breaking change. It would be better to create an operation in SlowTests for this purpose only (no need to BackupDatabaseNowResult from Server to Client then)

talweiss1982:
1. Resharper didn't refactor properly
2. No, this is something that should have been there from the first place.

ppekrol:
2. This is a maintenance operation, so it's OK to do so. 
3. I think it would be better to put the `BackupDatabaseNowResult` inside the `StartBackupOperation` and just rename it to Result

PR #7297, opened 2018-08-09
RavenDB-10910 Disallow Server bind to an active TCP port address

IdanHaim:


src/Raven.Server/RavenServer.cs
```diff
@@ -1007,13 +1008,34 @@ public TcpListenerStatus StartTcpListener()
             return status;
         }
 
+        private bool PortInUse(int port)
+        {
+            var ipGlobalProperties = IPGlobalProperties.GetIPGlobalProperties();
+            return ipGlobalProperties.GetActiveTcpListeners().FirstOrDefault(endPoint => endPoint.Port == port) != null;
```
ayende, line 1014: Does this work on Linux?

src/Raven.Server/RavenServer.cs
```diff
@@ -1007,13 +1008,34 @@ public TcpListenerStatus StartTcpListener()
...
+
         private void StartListeners(string host, int port, TcpListenerStatus status)
         {
             try
             {
                 _tcpAuditLog = LoggingSource.AuditLog.IsInfoEnabled ? LoggingSource.AuditLog.GetLogger("TcpConnections", "Audit") : null;
                 bool successfullyBoundToAtLeastOne = false;
                 var errors = new List<Exception>();
+
+                if (PortInUse(port))
```
ayende, line 1025: Why do the check instead of letting the OS do that? Better to handle the error in this case.

IdanHaim: The OS will throw an error only if will try to connect the TCP

PR #7298, opened 2018-08-09
[4.1] RavenDB-11571

aviv86:


test/SlowTests/Issues/RavenDB-11571.cs
```diff
@@ -0,0 +1,178 @@
...
+
+                using (var session = store.OpenSession())
+                {
+                    var documentCounters = session.CountersFor("users/1");
+                    var val = documentCounters.Get("Downloads");
+                    Assert.Equal(long.MaxValue, val);
+
+                    documentCounters.Increment("Downloads");
+                    var e = Assert.Throws<RavenException>(() => session.SaveChanges());
+                    Assert.Contains("Arithmetic operation resulted in an overflow", e.Message);
```
ayende: We need to tell on what counter / document this happened.

ayende (changes requested):

test/SlowTests/Issues/RavenDB-11571.cs
ppekrol: @aviv86 Let's create a CounterOverflowException for that then. We can push there document id, counter name, old value and new value and OverflowException as inner.

ppekrol: E.g. `Could not increment counter '{name}' from document '{documentId}' with value '{value}' by '{increment}'. Arithmetic operation resulted in an overflow.`

aviv86: done

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -317,11 +318,23 @@ public string IncrementCounter(DocumentsOperationContext context, string documen
...
+                    {
+                        value = checked(prev + delta); //inc
+
+                    }
+                    catch (OverflowException e)
+                    {
+                        throw new CounterOverflowException(
+                            $"Could not increment counter '{name}' from document '{documentId}' " +
+                            $"with value '{prev}' by '{delta}'. " +
+                            "Arithmetic operation resulted in an overflow.", e);
```
ayende, line 336: The last sentence should be removed, that is the message of the overflow exception.

ppekrol: Fixed

PR #7301, opened 2018-08-09
RavenDB-11503 fix

myarichuk:


.gitignore
```diff
@@ -117,3 +117,4 @@ temp
 .vscode
 /test/Tryouts/Databases/*
 bench/Micro.Benchmark/BenchmarkDotNet.Artifacts/
+/src/Raven.Server/*.DotSettings
```
ppekrol, line 120: how you ended up with *.DotSettings file inside that dir? We have it in the root

test/SlowTests/Server/Replication/ReplicationBasicTestsSlow.cs
```diff
@@ -96,10 +97,15 @@ public async Task Master_master_replication_from_etag_zero_without_conflict_shou
         [Fact]
         public async Task DisableExternalReplication()
         {
-
-            using (var store1 = GetDocumentStore())
-            using (var store2 = GetDocumentStore())
+            using (var store1 = GetDocumentStore(new Options
+            {
+                ExternalReplicationEnabledForTests = true
```
ppekrol, line 102: We do not want to have this

ppekrol:
Please check failing tests.

test/Tests.Infrastructure/RavenTestBase.cs
```diff
@@ -95,6 +97,11 @@ protected DocumentStore GetDocumentStore(Options options = null, [CallerMemberNa
                     options = options ?? Options.Default;
                     var serverToUse = options.Server ?? Server;
 
+                    if (options.ExternalReplicationEnabledForTests)
```
ayende, line 100: Why do this instead of the previous method?

test/Tests.Infrastructure/RavenTestBase.cs
```diff
@@ -256,6 +263,29 @@ protected DocumentStore GetDocumentStore(Options options = null, [CallerMemberNa
             }
         }
 
+        private static void SetExternalReplicationToLicenseToTrue(RavenServer serverToUse)
```
ayende, line 266: No. We already have a way to setup license for tests.

.gitignore
myarichuk: Not sure. I think it was added by R# somehow.

PR #7304, opened 2018-08-09
[4.1] RavenDB-11703

aviv86:
http://issues.hibernatingrhinos.com/issue/RavenDB-11703

### DIFF
--- a/src/Raven.Server/Documents/CountersStorage.cs
+++ b/src/Raven.Server/Documents/CountersStorage.cs
@@ -558,11 +558,14 @@ namespace Raven.Server.Documents
             var table = GetCountersTable(context.Transaction.InnerTransaction, collectionName);
 
             long deletedEtag = -1;
+            string documentId = null;
+            string name = null;
             if (table.DeleteByPrimaryKeyPrefix(key, tvh =>
                 {
                     long etag = *(long*)tvh.Reader.Read((int)CountersTable.Etag, out var size);
                     deletedEtag = Math.Max(Bits.SwapBytes(etag), deletedEtag);
                     Debug.Assert(size == sizeof(long));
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -574,6 +577,14 @@ public string DeleteCounter(DocumentsOperationContext context, Slice key, string
 
             CreateTombstone(context, key, collection, deletedEtag, lastModifiedTicks, newEtag, newChangeVector);
 
+            context.Transaction.AddAfterCommitNotification(new DocumentChange
```
ayende, line 580: Let's create a dedicate notification for counters.
Should include doc id, counter name and the current (overall) counter value.
Subscribing to a document should also subscribe to its counteers

ayende (changes requested):

src/Raven.Server/Documents/CountersStorage.cs
ppekrol: 'Subscribing to a document should also subscribe to its counters' - how would that look from the Client API perspective? I'm not sure if we should go that way. User should either subscribe to a DocumentChange or to a CounterChange.

ppekrol: I'm merging this. Let's handle the notification change as a separate issue.

ppekrol: http://issues.hibernatingrhinos.com/issue/RavenDB-11727

ayende: The idea is that when you want to hear about a document, you do it once, for all its notifications.
On the client side, you can `Subscribe(IObserver<CounterChanges>)` or not.

PR #7305, opened 2018-08-09
[4.0] - RavenDB-11710 - Cannot create database on macOS

grisha-kotler:
http://issues.hibernatingrhinos.com/issue/RavenDB-11710

### DIFF
--- a/src/Raven.Server/Documents/DocumentDatabase.cs
+++ b/src/Raven.Server/Documents/DocumentDatabase.cs
@@ -308,6 +308,11 @@ namespace Raven.Server.Documents
                 _writeLockFile.SetLength(1);
                 _writeLockFile.Lock(0, 1);
             }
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -308,6 +308,11 @@ private void TryAcquireWriteLock()
                 _writeLockFile.SetLength(1);
                 _writeLockFile.Lock(0, 1);
             }
+            catch (PlatformNotSupportedException)
```
ppekrol, line 311: I'm wondering if we want to add `when (PlatformDetails.IsMacOs)` here?

ayende: No, any `PlatformNotSupported` would work here. 
I'm thinking about BSD in the future.

PR #7307, opened 2018-08-09
RavenDB-10910 minor

ppekrol:


src/Raven.Server/RavenServer.cs
```diff
@@ -1034,7 +1034,7 @@ private void StartListeners(string host, int port, TcpListenerStatus status)
                         $"You can change the TCP port using one of the following options:{Environment.NewLine}" +
                         $"1) Change the ServerUrl.Tcp property in setting.json file.{Environment.NewLine}" +
                         $"2) Run the server from the command line with --ServerUrl.Tcp option.{Environment.NewLine}" +
-                        $"3) Add RAVEN_ServerUrl.Tcp to the Environment Variables.{Environment.NewLine}" +
+                        $"3) Add RAVEN_ServerUrl_Tcp to the Environment Variables.{Environment.NewLine}" +
```
maximburyak, line 1037: Actually, I think that both are correct https://github.com/ravendb/ravendb/blob/v4.0/src/Raven.Server/Config/RavenConfiguration.cs#L424

ppekrol: Correct, but using `.` is not always easy (IIRC bash has issues with it). It is better to use `_` and stay on the safe side.

PR #7308, opened 2018-08-09
RavenDB-11712 attaching continuation only once

ppekrol:


ppekrol:
@karmeli87 please review

PR #7310, opened 2018-08-09
[DO-NOT-MERGE] For review only: Generalized Allocators

redknightlois:
This is the first working implementation of Generalized Allocators, includes stress testing facilities, crude leaks detection, ByteString implementation using the new allocators, Spanified Memory Access, Spanified Pointer Types...

Because this is going to be the foundation of memory management at Voron and in some fashion the POC for spanification of the core we must get this right from day 1. Please take your time to review, and anything that crosses your mind even if you think it could be a dumb comment, do it anyway :) ... 

All release FastTests work, Debug fails because of memory leaks detected by the allocators. Many things missing.

Copy for review anyone you think it should take a look into this. cc @ayende, @arekpalinski, @ppekrol.

PS: I left the history in case anyone wants to look into the design history. If you think it is better to collapse the history, let me know and I will squash it.


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1192,7 +1192,8 @@ private void ReduceMemoryUsage()
 
             DocumentDatabase.DocumentsStorage.ContextPool.Clean();
             _contextPool.Clean();
-            ByteStringMemoryCache.CleanForCurrentThread();
+            // TODO: Centralize the blocks fragmentation to ensure we can clean up the memory on-demand without having to go to each context to do so. 
+            //ByteStringMemoryCache.CleanForCurrentThread();
```
ayende, line 1199: This is taking advantage on the fact that we are aware of the access pattern of indexing code, so we can do better here than usually.

src/Sparrow/Allocators.Arena.cs
```diff
@@ -0,0 +1,336 @@
...
+            if (!allocator._currentBuffer.IsValid || allocator._used + size > allocator._currentBuffer.Size)
+            {
+                allocator.GrowArena(ref allocator, size);
+            }
+
+            Pointer ptr = new Pointer(allocator._ptrCurrent, size);
+            allocator._ptrCurrent += ptr.Size;
+            allocator._used += ptr.Size;
+
+            allocator.InUse += ptr.Size;
```
ayende, line 169: Why tracking both `_used` and `InUsed` ?

src/Sparrow/Allocators.Arena.cs
```diff
@@ -0,0 +1,336 @@
...
+            if (!allocator._currentBuffer.IsValid)
+            {
+                allocator._currentBuffer = allocator._nativeAllocator.Allocate(ref allocator._nativeAllocator, (int)allocator._preferredSize);
+                allocator.Allocated += allocator._currentBuffer.Size;
+            }
+
+            if (allocator._olderBuffers != null)
+            {
+                // Free old buffers not being used anymore. A single arena ought to be enough for anybody, you know ;) 
+                foreach (var unusedBuffer in allocator._olderBuffers)
```
ayende, line 294: This is already done in the `Reset`, why is it here?

src/Sparrow/Allocators.Fragment.cs
```diff
@@ -0,0 +1,252 @@
...
+        }
+
+        /// <summary>
+        /// Comparer for comparing two keys, handling equality as beeing greater
+        /// Use this Comparer e.g. with SortedLists or SortedDictionaries, that don't allow duplicate keys
+        /// </summary>
+        /// <typeparam name="TKey"></typeparam>
+        public class DuplicateKeyComparer<TKey> : IComparer<TKey> where TKey : IComparable
+        {
+            public int Compare(TKey x, TKey y)
```
ayende, line 152: Pretty sure that this violates the requirements, no? It need to be reversible

src/Sparrow/Allocators.Fragment.cs
```diff
@@ -0,0 +1,252 @@
...
+            if (bytesLeft > allocator._options.ReuseBlocksBiggerThan)
+            {
+                // The allocation is bigger than ReuseBlocksBiggerThan and therefore we can leave this one available to be used later.
+                allocator._internalReadyToUseMemorySegments.Add(bytesLeft, new Pointer(allocator._ptrCurrent, bytesLeft));
+            }
+
+            // TODO: Time the impact of using a heap structure instead of a SortedList.
+            // Get the biggest segment on the Heap and if size < the biggest one, we avoid the allocation.
+            Pointer segment;
+            if (allocator._internalReadyToUseMemorySegments.Count > 0 && allocator._internalReadyToUseMemorySegments.Values[0].Size >= size)
```
ayende: This actually gets the _smallest_ segment you have here, no? This is sorted by `bytesLeft`, and therefor from smallest to largest.

src/Sparrow/Allocators.Fragment.cs
```diff
@@ -0,0 +1,252 @@
...
+            allocator.InUse = 0;
+        }
+
+        public void OnAllocate(ref FragmentAllocator<TOptions> allocator, Pointer ptr) {}
+        public void OnRelease(ref FragmentAllocator<TOptions> allocator, Pointer ptr) {}
+
+        public void Dispose(ref FragmentAllocator<TOptions> allocator)
+        {
+            // We are not going to track this memory chunks anymore.
+            allocator._internalReadyToUseMemorySegments.Clear();
```
ayende: Aren't we going to release these first?

src/Sparrow/Allocators.Native.cs
```diff
@@ -0,0 +1,177 @@
...
+                memory = ElectricFencedMemory.Allocate(size);
+            else if (allocator._options.UseSecureMemory)
+                throw new NotImplementedException();
+            else
+                memory = NativeMemory.AllocateMemory(size);
+
+            if (allocator._options.Zeroed)
+                Memory.Set(memory, 0, size);
+
+            allocator.Allocated += size;
```
ayende: Aren't we concerned with multithreading here?

src/Sparrow/Allocators.PointerTypes.cs
```diff
@@ -0,0 +1,491 @@
...
+        public bool IsValid
+        {
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            get { return Ptr != null; }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public Span<byte> AsSpan()
+        {
+            return new Span<byte>(Ptr, (int)Size);
```
ayende: why the cast to int here?

src/Sparrow/Allocators.PointerTypes.cs
```diff
@@ -0,0 +1,491 @@
...
+
+            return new ReadOnlySpan<byte>(Ptr, length);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public ReadOnlySpan<byte> AsReadOnlySpan(int start, int length)
+        {
+#if VALIDATE
+            if (start + length > Size)
+                throw new ArgumentException($"{nameof(length)} cannot be bigger than block size.");
```
ayende, line 85: I'm not so sure that we should discard these changes in release.

src/Sparrow/Allocators.PointerTypes.cs
```diff
@@ -0,0 +1,491 @@
...
+    /// The BlockPointer is a type of pointer that carries with it 2 different sizes at the same time. The pointer value
+    /// with the proper memory size allocated to it AND the actual requested size when asked. A BlockPointer created from
+    /// a straight Pointer will have the same size for the allocation size and the actual size. 
+    /// </summary>
+    [StructLayout(LayoutKind.Sequential)]
+    public readonly unsafe struct BlockPointer : IPointerType, Meta.IDescribe
+    {
+        public readonly void* Ptr;
+
+        public readonly int Size;
```
ayende, line 270: Better names `Used`, `Total` ?

src/Sparrow/Allocators.Pool.FixedSize.cs
```diff
@@ -0,0 +1,245 @@
...
+                allocator.ReleaseMemoryPool(ref allocator);
+            }
+
+            allocator._internalAllocator.Dispose();
+        }
+
+        private void ResetMemoryPool(ref FixedSizePoolAllocator<TOptions> allocator)
+        {
+            // We dont own the memory pool, so we just reset the state and let the owner give us memory again on the next cycle.
+            // This is the typical mode of operation when the underlying allocator is able to reuse memory (ex. NativeAllocator).
```
ayende: This comment and the one on `ReleaseMemoryPool` both mention `NativeAllocator`, so this is confusing.
What is the use cases for the pool not owning their memory?

src/Sparrow/Allocators.Pool.FixedSizeThreadAffine.cs
```diff
@@ -0,0 +1,329 @@
...
+
+            public bool AcceptOnlyBlocks => false;
+
+            public ThreadAffineWorkload Workload => ThreadAffineWorkload.Default;
+
+            public bool HasOwnership => true;
+            public IAllocatorComposer<Pointer> CreateAllocator()
+            {
+                var allocator = new Allocator<NativeAllocator<PoolAllocator.Default>>();
+                allocator.Initialize(default(PoolAllocator.Default));
```
ayende: Shouldn't this be `DefaultAcceptArbitrarySize`?

src/Sparrow/Allocators.Pool.cs
```diff
@@ -0,0 +1,280 @@
...
+
+            public IAllocatorComposer<Pointer> CreateAllocator()
+            {
+                var allocator = new Allocator<NativeAllocator<Default>>();
+                allocator.Initialize(default(Default));
+                return allocator;
+            }
+        }
+
+        public struct Static : IPoolAllocatorOptions, INativeOptions, INativeGlobalAllocatorOptions
```
ayende: How is this different from the `Default`?

src/Sparrow/Allocators.cs
```diff
@@ -0,0 +1,524 @@
...
+                return ptr;
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public Pointer<TType> Allocate<TType>() where TType : struct
+        {
+            unsafe
+            {
+                var ptr = _allocator.Allocate(ref _allocator, this._blockSize * Unsafe.SizeOf<TType>());
```
ayende, line 579: How is multi threading handled in this case?

src/Sparrow/Meta.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Sparrow
+{
+    public static class Meta
+    {
+        public interface IDescribe
```
ayende, line 9: Why not just use `ToString`?

src/Raven.Server/Documents/Indexes/Index.cs
redknightlois: Yes, but I think we have a better facility now to do so using composition. Lets talk about the underlying issue (so I can have the big picture) and figure out what we need to do so.

src/Sparrow/Allocators.Arena.cs
redknightlois: Two different metrics. ``InUse`` is how many bytes the allocator has given out during the current cycle. ``_used`` is an internal pointer that tells us how much we have used of the current arena. Probably ``_used`` is obsolete right now, but I haven't started polishing that kind of stuff yet.

src/Sparrow/Allocators.Arena.cs
redknightlois: I changed the semantics of ``Reset`` and ``Renew`` to unify criteria. That was the second opinion I needed yesterday.

src/Sparrow/Allocators.Fragment.cs
redknightlois: You mean for the ``SortedList<T>``? I don't expect it to be the final solution though, I would prefer to have a different implementation there (HEAP)

src/Sparrow/Allocators.Fragment.cs
redknightlois: Will check, it should be largest.

src/Sparrow/Allocators.Pool.FixedSize.cs
redknightlois: Composition, the idea is that the pool does not allocate memory, it uses a real allocator to do so. However, given that you can have Self Healing and Leaky allocators, the ownership is about transferring the responsibility. Pools don't need to release explicitly if the allocator is self-healing.

src/Sparrow/Allocators.Pool.FixedSizeThreadAffine.cs
redknightlois: Yes, there is a bug in all 3.

src/Sparrow/Allocators.Pool.cs
redknightlois: It is expected to die when the process dies, no explicit deallocation, therefore no checking for leaks.

src/Sparrow/Allocators.cs
redknightlois: Underlying allocator must support it. The shell is Thread Safe.

src/Sparrow/Meta.cs
redknightlois: I didn't finish the ``IDescribe`` interface yet, for now, it is just a marker. Allocators are composable so there are multiple metrics along the way and each allocator in the chain paints a different picture. We need to be able to gather information along the way in a more structured way. Also ``.ToString()`` is vulnerable to out of band calls via the debugger and inspectors, which may complicate things. I would move away from the ``.ToString()`` mechanism for general inspection altogether and keep it bare essential for identification purposes instead.

src/Sparrow/Allocators.Native.cs
redknightlois: This is not a thread safety allocator, but probably we should pay the cost here anyways.

src/Sparrow/Allocators.Fragment.cs
redknightlois: Yes, bug.

src/Sparrow/Allocators.PointerTypes.cs
redknightlois: Because Span constructor is not able to handle chunks of memory beyond int.MaxInt

src/Sparrow/Allocators.PointerTypes.cs
redknightlois: We do, we shouldn't (potentially) kill the process because of a leak... we should degrade 'gracefully' instead. That is the reason why I set it to die horribly in debug instead.

redknightlois: Makes sense, better to check than touch memory that is potentially not ours... I originally though it was the Allocator Leaks detection. In this case, you are right.

src/Sparrow/Allocators.Fragment.cs
redknightlois: No, that's fine. For a moment I got confused. Those are actually partial blocks of memory that are big enough to be reused... say you request a 1Mb chuck of fragment memory, now you can either pool 1Mb (What the Pooling allocator does) or you can allow the allocator to slice it up to support smaller sizes (What the fragment allocator does). The difference is that Fragment allocator is tuned for very small, high reuse of memory... while the Pooling allocator may have keep lots of unused memory if still havent met the quota.

src/Sparrow/Allocators.Fragment.cs
redknightlois: That's right. Also reflected that there is no test yet that stresses that condition because there was another bug there too.

src/Sparrow/Allocators.Fragment.cs
redknightlois: Just checked, no invariant is violated; it just works... would be if you would expect a stable sort though (which we dont care).

src/Sparrow/Allocators.Pool.FixedSize.cs
redknightlois: There was an error in the comment. In the standard pool it was right.

src/Sparrow/Allocators.Arena.cs
ayende: What _are_ the semantics then, now? This need to be explicitly stated.

src/Sparrow/Allocators.PointerTypes.cs
ayende: `Size` is already int, no? `public readonly int Size;`?

src/Sparrow/Allocators.cs
ayende: If you are trying to allocate from one thread and release from another, bad things will happen right now.
This is also the case when you have chains of allocators where some are threaded and some are not.

src/Sparrow/Meta.cs
ayende: `.ToString()` is too convenient to give up on. If the call is self contain and doesn't mutate the internal state (and it shouldn't), that that is what we want to use.

src/Sparrow/Allocators.Fragment.cs
ayende: What is the scenario where you'll use this?

HibernatingRhinosServiceAccount:
Degrading gracefully when you have an out of band memory error means
pointer chasing and dump reading and a lot of hairloss.
I want to be able to get immediate and obvious error in production when
this happens.
Doesn't have to take down the process, but throwing an error up is required.

*Hibernating Rhinos Ltd  *

Oren Eini* l CEO l *Mobile: + 972-52-548-6969

Office: +972-4-622-7811 *l *Fax: +972-153-4-622-7811

src/Sparrow/Allocators.PointerTypes.cs
```diff
@@ -269,21 +269,17 @@ public Span<byte> AsSpan()
...
 
             return new Span<byte>(Ptr, length);
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public Span<byte> AsSpan(int start, int length)
         {
-#if VALIDATE
             if (start + length > Size)
                 throw new ArgumentException($"{nameof(length)} cannot be bigger than block size.");
```
ayende, line 314: This JIT is now smart enough to move the exception throwing code to a cold location without creating dedicated method?

src/Sparrow/Allocators.cs
redknightlois: Currently it is like that, I modeled it based on current operation. We can implement underlying thread safe and also thread sharing allocators which would fixed that if we want. The idea of composability is that you can implement behaviors based on the requirements and pay the cost only for those cases. Now the question is: do we need multithread safe allocators for pools and fragments? If that is so, we can implement those too

src/Sparrow/Meta.cs
redknightlois: I haven't killed ToString I don't want it to generate chained information from all allocators in the chain or use it to inspect data. The idea is that the describe will return structured information that can be translated to a json for inspection on higher levels

redknightlois:
Are we talking about leaks or invalid pointer access? I got lost in the details

src/Sparrow/Allocators.cs
ayende: I'm perfectly fine with stating that allocators being single threaded, but it needs to be stated clearly as such.

ayende: In particular, what allocators _are_ safe for multi threading and can be re-used.

src/Sparrow/Meta.cs
ayende: The proper way to handle this is either: Return an _object_, not a string, that can be used for the debug endpoint.
Or: Use `ToString` and call it a day

ayende:
I'm talking specifically here about the `AsSpan` calls with the `start`, `length` calls where we _have_ to do bound checking in prod as well

src/Sparrow/Allocators.Arena.cs
```diff
@@ -0,0 +1,332 @@
...
+                    allocator._nativeAllocator.Release(ref _nativeAllocator, ref ptr);
+                }
+            }
+
+            if (allocator._currentBuffer.IsValid)
+            {
+                allocator._nativeAllocator.Release(ref _nativeAllocator, ref _currentBuffer);
+            }
+
+            allocator._nativeAllocator.Dispose(ref allocator._nativeAllocator);
```
arekpalinski: Why do we pass the caller (`_nativeAllocator`) here and when calling `Release()` a few lines up?

src/Sparrow/Meta.cs
redknightlois: The idea is to return an object. ``string`` is a placeholder at the moment.

src/Sparrow/Allocators.Arena.cs
redknightlois: The semantics are:
``Renew()`` ensures that if possible and supported by the allocator, no memory is going to be released to the underlying allocator. Let's say a pool, a pool unless is faced with memory pressure, it is not going to release the memory to the Arena Allocator or Native Allocator behind it if called ``Renew()`` but it will if called ``Reset``.

``Reset()`` is a fresh start, let's say that the ArenaAllocator has been running for a long time and now has an arena size (``.Allocated``) of 400Mb. With a Renew it will just move the pointer to zero and go on with its business, with Reset it will go, check the new preferred size given the behavior of the last round and return the memory to the underlying memory system if different. 
 
The current behavior of the allocators is confusing, because some behave like this, an others behave differently. What I am trying to do is to have a clear sematic for both operations. Now there is no reason why you would call ``Renew`` and ``Reset`` at the same time as we do in ``JsonContext``

src/Sparrow/Allocators.Arena.cs
redknightlois: That is a design decision that may or may not have merits, but until I have something working really well I won't revisit. The idea of why allocators go always by reference is because C# does not support static methods on generic types. The problem is that you can create copies of the allocators by mistake, that prevents that and also helps the JIT to figure out some stuff when inlining this kind of code at the caller size ``Allocator<T>``. It's highly probable that we may get away without it, but because of how difficult this code is I preferred to err on the side of safety.

redknightlois:
@ayende Ohh yes, we are bound checking explicitly now on creation.

src/Sparrow/Allocators.cs
redknightlois: Only multi-threaded safe allocators now are:
- NativeAllocator
- ThreadAffinePoolAllocator

Because they are underlying allocators, they can be used as shared allocators to thread safe ones providing multithreaded shared pools backends to then thread-unsafe allocators built per context.

src/Sparrow/Allocators.PointerTypes.cs
redknightlois: We should yes, but I didnt want to change overnight what has been long since day 1.

src/Sparrow/Allocators.Arena.cs
ayende: The problem we have here is that there are two distinct stages to deal with.
A context is checked out, used and returned.
The problem is what you are expected to do on the return.
You might want to return the memory back, retain it, etc. 

We need to play with the lifecycle of the context usage.

Given `AllocateContext` and `ReturnContext` in their use in web requests, for example, what will be called on either?

redknightlois: The idea behind a clear cut definition of what each operation do is to avoid the current situation, where each allocator decides how it interprete each one. It is the job of ``AllocateContext`` and ``ReturnContext`` to assess which one they should call based on usage patterns.

redknightlois:
Dealing with memory is a pain but good exercise to review everything. Does anyone knows why ``UnmanagedWriteBuffer`` handles its own memory allocations / cache instead of just request size X when it needs it? To me it looks like it is implementing its own 'arena' style allocation policy.

ayende:
`UnmanagedWriteBuffer` just uses `GetManagedBuffer`, it shouldn't do any additional memory allocations. Where are you seeing that?
Note that it just need a temp buffer and is one of the targets for Pipes with the spanification

redknightlois:
What I mean is that ``UnmanagedWriteBuffer`` does memory pooling in what it looks like an `arena` style allocation. That means that every instance is acquiring and keeping memory just in case, while it could be solved elegantly just asking the memory and returning it to an arena, because it should be extremely fast and the use pattern looks like it can be even optimal in that case. 

Yes, I am in the process of looking up cases for Pipes and Spans, but right now the biggest issue is that Pipes are not supported in older netstandard and Sparrow must support older clients too. 

EDIT: I see, they are deploying it for all versions

ayende:
`UnmanagedWriteBuffer` is a short lived struct, it's only memory usage is with the `_buffer` it uses, which is owned by the context and released when the context is reset. I'm not sure where you see arena memory management?

redknightlois:
Most uses of ``UnmanagedWriteBuffer`` ends up being enclosed in ``using`` statements. That is why I am saying it is arena style. You acquire the memory you need (aka 'the buffer'), you can return it as soon as you are done to a high-performance allocator for reuse (we are not doing it).

ayende:
It just uses a buffer, doesn't matter from where.

ayende:
@redknightlois Please check `RavenDB_10502` test. It is one of several the cause memory corruption.
I used electric fence to force it to fail consistently in `PoolAllocator.Release`, where it is corruption some totally different memory, not quite sure how.

ayende:
Please rebase of `v4.2`

PR #7313, opened 2018-08-10
RavenDB-11727 added CounterChange

ppekrol:


src/Raven.Client/Documents/Changes/ChangeNotification.cs
```diff
@@ -57,9 +52,7 @@ public DynamicJsonValue ToJson()
                 [nameof(Type)] = Type.ToString(),
```
maximburyak, line 52: Could we pass here just the numeric value? this way we could save some perf on the ToJson and on the FromJson?

maximburyak: hmm.. no I see that it'll probably be a backward compatability break...

src/Raven.Server/Documents/DocumentsTransaction.cs
```diff
@@ -17,6 +17,8 @@ public class DocumentsTransaction : RavenTransaction
 
         private List<DocumentChange> _documentNotifications;
 
+        private List<CounterChange> _counterNotifications;
```
maximburyak, line 20: We might want to have a pool of _documentNotifications and _counterNotifications

src/Raven.Client/Documents/Changes/ChangeNotification.cs
ayende: And given the choice between have `Type=3` and `Type='Document'`, I ll take the readable option.

src/Raven.Client/Documents/Changes/DatabaseChanges.cs
```diff
@@ -22,6 +20,8 @@ namespace Raven.Client.Documents.Changes
 {
     public class DatabaseChanges : IDatabaseChanges
     {
+        internal const string Separator = ":||:";
```
ayende: Why do we need a separator?

src/Raven.Client/Documents/Changes/DatabaseChanges.cs
```diff
@@ -220,6 +229,61 @@ public IChangesObservable<DocumentChange> ForDocumentsInCollection<TEntity>()
...
+        }
+
+        public IChangesObservable<CounterChange> ForCounterOfDocument(string documentId, string counterName)
+        {
+            if (string.IsNullOrWhiteSpace(documentId))
+                throw new ArgumentException("Value cannot be null or whitespace.", nameof(documentId));
+            if (string.IsNullOrWhiteSpace(counterName))
+                throw new ArgumentException("Value cannot be null or whitespace.", nameof(counterName));
+
+            var counter = GetOrAddConnectionState($"counter/{counterName}/document/{documentId}", "watch-document-counter", "unwatch-document-counter", string.Join(Separator, documentId, counterName));
```
ayende: I would actually expect it to be document, then counter, since counter is inside doc.

src/Raven.Server/Documents/ChangesClientConnection.cs
```diff
@@ -506,15 +634,71 @@ public DynamicJsonValue GetDebugInfo()
...
                 ["WatchIndexes"] = _matchingIndexes.ToArray(),
-                ["WatchDocuments"] = _matchingDocuments.ToArray()
+                ["WatchDocuments"] = _matchingDocuments.ToArray(),
+                ["WatchCounters"] = _matchingCounters.ToArray(),
+                ["WatchCounterOfDocument"] = _matchingDocumentCounter.Select(x => x.ToJson()).ToArray(),
+                ["WatchCountersOfDocument"] = _matchingDocumentCounters.ToArray()
             };
         }
+
+        private static DocumentIdAndCounterNamePair GetDocumentCounterParameters(string parameter)
```
ayende: We not send this properly, as two arguments?

ppekrol: Fixed

src/Raven.Client/Documents/Changes/DatabaseChanges.cs
ppekrol: Not sure that I follow?

ayende: `counter/{counterName}/document/{documentId}` vs: 
`document/{documentId}/counter/{counterName}`

ppekrol: I wanted to be consistent with `ForCounterOfDocument` -> `counter/of/document`. This is used only internally in ConcurrentDict.

ayende: Okay, this is an in memory only value, anyway.

ppekrol: fixed

PR #7314, opened 2018-08-10
[4.0] RavenDB-11728 Invalid GetHashCode() calculation for nested blittable …

arekpalinski:
…resulting in invalid map-reduce results

### DIFF
--- a/src/Sparrow/Json/BlittableJsonReaderObject.cs
+++ b/src/Sparrow/Json/BlittableJsonReaderObject.cs
@@ -20,6 +20,7 @@ namespace Sparrow.Json
         private readonly int _propCount;
         private readonly long _currentOffsetSize;
         private readonly long _currentPropertyIdSize;
+        private readonly bool _isRoot;
         private byte* _objStart;
 
         public DynamicJsonValue Modifications;
@@ -49,6 +50,7 @@ namespace Sparrow.Json
             if (size == 0)
                 ThrowOnZeroSize(size);
 
+            _isRoot = true;
             _buffer = buffer;
             _mem = mem; // get beginning of memory pointer
             _size = size; // get document size
@@ -112,6 +114,7 @@ namespace Sparrow.Json
         public BlittableJsonReaderObject(int pos, BlittableJsonReaderObject parent, BlittableJsonToken type)
             : base(parent._context)
         {
+            _isRoot = false;
             _parent = parent;
             _mem = parent._mem;
             _size = parent._size;
@@ -1113,7 +1116,7 @@ NotFound:
 
         public override int GetHashCode()
         {
-            return _size ^ _propCount;
+            return _isRoot ? _size ^ _propCount : _propCount;
         }
 
         [Conditional("DEBUG")]
--- a/test/FastTests/Blittable/BlittableJsonEqualityTests.cs
+++ b/test/FastTests/Blittable/BlittableJsonEqualityTests.cs
@@ -1,4 +1,5 @@
-﻿using Sparrow.Json;
+﻿using System.Collections.Generic;
+using Sparrow.Json;
 using Sparrow.Json.Parsing;
 using Xunit;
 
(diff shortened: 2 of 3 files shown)


src/Sparrow/Json/BlittableJsonReaderObject.cs
```diff
@@ -1113,7 +1116,7 @@ protected bool Equals(BlittableJsonReaderObject other)
 
         public override int GetHashCode()
         {
-            return _size ^ _propCount;
+            return _isRoot ? _size ^ _propCount : _propCount;
```
maximburyak, line 1119: Maybe, instead of just "_propCount" we could calculate the hash code for the metadata sequence?

ppekrol: It might not be worth it to calculate the hash for the whole metadata sequence.

@redknightlois can we do any better here? The only place where this is actively used is the GroupBy statement in Reduce.

maximburyak: we could at least use the metadata sequence length, but it's still no good, especially for map reduce.. it will ultimately be using the same hash, because we expect all reduce keys to be with the same item count. In the bottom line we'll end up performing o(n^2) comparisons between blittables, which is more expensive (see equals implementation) then having a o(n) of hash code calculations and storing them in order to prevent repeated calculations. It's going to be a o(n) calculations, because we know that bjro are immutable, so it won't change after we've already calculated the hash.

redknightlois: @ppekrol Many of our data structures assume properly behaved ``.GetHashCode()`` implementations; I cannot overstate how badly behaved is this one, if this ends up in one of our dictionaries, we are going to be in deep problems. Use ``Hashing.Combine(x,y)`` from https://github.com/ravendb/ravendb/blob/v4.1/src/Sparrow/Hashing.cs#L737 if you really need to check ``_size`` and ``_propCount`` only. Having said so, the general guideline for hashes should be: "Implement the proper ``.GetHashCode()`` for the general object, override in cases they are too costly and you can get away with less" enclosing the object on an struct instead. The same if when you calculated it, and it's going to stay fixed afterwards. 

Does that respond the question?

ayende: Is there any reason we aren't using the actual binary data itself to generate the hash. That would be the _right_ thing to do, no?

arekpalinski: @ayende that would be actually "perfect" `GetHashCode()` implementation, so actually it would be `Equals()`. We cannot calculate it easily since:
1) we don't know the size of nested objects (`_size` property is the size of root BJRO) so cannot calculate hash in one call
2) it would require to go over properties in insertion order and hash each property {name, value} to properly implement that

I think we'll be better off by having "weak" `GetHashCode` and call `Equals` when needed

@redknightlois what is advantage of using https://github.com/ravendb/ravendb/blob/v4.1/src/Sparrow/Hashing.cs#L737 instead of single `^` operation?

ayende: we know what the start position is, can't we go to the last property by insertion order and then get it's position / size and compute the hash over that entire range?

arekpalinski: Just to be sure, are we _guaranteed_ that:

```
{
  "Foo": "Bar",
  "Baz": { "Blah":"Foo", "Bam": "Bla"}
}
```

```
{
 "Baz": { "Bam": "Bla", "Blah":"Foo"},
  "Foo": "Bar"
}
```

will result in _the same_ blittable?

ayende: We are guaranteed that it _won't_.

ayende: In this case, we need to compute hash recursively, by iterating over the properties in sorted orders and combining all the hashes of property names / values.

ppekrol: This is more or less what Equals is doing, but we stop the computation on first difference. I think it would be better to have a weak hash and do the Equals than perform that computation in GetHashCode.

redknightlois: @arekpalinski The problem arise from the distribution behavior of the hash function. Lets say that you have ``0x0001`` and ``0x0011``, when you XOR them numbers that are near, end up being near... Now, suppose that the distribution of your inputs is not so random (size is not :D)... now your actual output dimensionality is very small, leading to huge collision spaces and also huge spaces of unused hash potential. Think when you do that with memory addresses too, it is easy to see that the distribution of the output is clustered and the cluster is a high collision cluster zone. 

Even the ``HashCombiner.Combine()`` which is a rotation based combiner is not good if the inputs are not guaranteed to be good (as in well behaved) hash values. The Combine function ensures that the probability of a single bit to be on and off is as near as 50% as possible even if the 2 input values differ in a single bit.

redknightlois: @ppekrol @ayende I would go in and implement the proper hash for the Equals (which is the right thing to do) but for the purpose of indexing (which is the use case) enclose the object into a struct say: ``WeakHash<T>`` which will implement a weaker version for use in this problem... We can then take a look and see how weak we can get along with.

arekpalinski: http://issues.hibernatingrhinos.com/issue/RavenDB-11739

PR #7322, opened 2018-08-12
RavenDB-11503 fix (including core review related changes)

myarichuk:


test/Tests.Infrastructure/ClusterTestBase.cs
```diff
@@ -579,27 +579,41 @@ protected override Task<DocumentDatabase> GetDocumentDatabaseInstanceFor(IDocume
...
         public async Task<(long Index, List<RavenServer> Servers)> CreateDatabaseInCluster(DatabaseRecord record, int replicationFactor, string leadersUrl)
         {
+            var serverCount = Servers.Count(s => s.Disposed == false);
+            if(serverCount < replicationFactor)
+            {
+                throw new InvalidOperationException($"Cannot create database with replication factor = {replicationFactor} when there is only {serverCount} servers in the cluster.");
+            }
+
             DatabasePutResult databaseResult;
+            string[] urls;
```
ayende, line 589: what is this used for?

myarichuk: It is possible that there will be multiple server clusters in the same test.
Thus, the *Servers* collection can be comprised from instances of multiple clusters. 

The statement [here](https://github.com/ravendb/ravendb/blob/ff67ed8d606958c2c7ef795b9a22efc9082e9d18/test/Tests.Infrastructure/ClusterTestBase.cs#L600)  fetches urls of instances of current cluster and then used to filter the current servers [here](https://github.com/ravendb/ravendb/blob/ff67ed8d606958c2c7ef795b9a22efc9082e9d18/test/Tests.Infrastructure/ClusterTestBase.cs#L603).

This change is needed to make sure that the filtering of *currentServers* [here](https://github.com/ravendb/ravendb/blob/ff67ed8d606958c2c7ef795b9a22efc9082e9d18/test/Tests.Infrastructure/ClusterTestBase.cs#L603) is not affected by a race condition.

ppekrol:
Will be handled here https://github.com/ravendb/ravendb/pull/7370

PR #7326, opened 2018-08-12
RavenDB-8791 Use `eval` with function `this` to parse the parameters

fitzchak:


src/Raven.Studio/wwwroot/Content/ace/mode-rql.js
```diff
@@ -1037,3 +1037,7 @@ oop.inherits(Mode, TextMode);
 exports.Mode = Mode;
 
 });
+
+this.window.evalUnstrict = function (str) {
```
ml054: Can we put this in some *.ts file?

ml054 (changes requested):
Can you use new Function('here goes function body') instead? eval overrides global objects + it is not secure.

ml054:
Optionally you can use es6 proxies to intercept global variables creation (or to catching assigning to this) + 'with' keyword to sandbox execution.

src/Raven.Studio/wwwroot/Content/ace/mode-rql.js
fitzchak: This was removed. I had to put it there to be outside the 'use strict' mode, which isn't possible in .ts file.

PR #7327, opened 2018-08-12
RavenDB-11657 Add the source url to PutRaftCommand

IdanHaim:


src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -2008,6 +2010,9 @@ public override void OnResponseFailure(HttpResponseMessage response)
             public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, ServerNode node, out string url)
             {
                 url = $"{node.Url}/admin/rachis/send";
+                Console.WriteLine(_source);
```
ppekrol: I think we can remove this?

src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
```diff
@@ -70,6 +70,7 @@ public async Task ApplyCommand()
                 command.VerifyCanExecuteCommand(ServerStore, context, isClusterAdmin);
 
                 var (etag, result) = await ServerStore.Engine.PutAsync(command);
+                var source = GetStringQueryString("source", required: false);
```
ayende: We don't actually need it here.
We put the `source` here _just_ for debug.

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -1963,7 +1963,7 @@ private async Task<(long Index, object Result)> SendToNodeAsync(string engineLea
                 if (clusterTopology.Members.TryGetValue(engineLeaderTag, out string leaderUrl) == false)
                     throw new InvalidOperationException("Leader " + engineLeaderTag + " was not found in the topology members");
 
-                var command = new PutRaftCommand(cmdJson);
+                var command = new PutRaftCommand(cmdJson, _engine.Url);
```
ayende: Let's also send the type of the command.

ayende (changes requested):

PR #7332, opened 2018-08-13
[4.1] RavenDB-11415 Cluster transaction fixes

karmeli87:
- Do backoff on cluster transction failure.
- Delete on PUT document to avoid exception if we put new document in a different collection.

### DIFF
--- a/src/Raven.Server/Documents/DocumentDatabase.cs
+++ b/src/Raven.Server/Documents/DocumentDatabase.cs
@@ -372,7 +372,7 @@ namespace Raven.Server.Documents
                     return;
 
                 _hasClusterTransaction.Reset();
-               
+
                 using (ServerStore.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
                 using (context.OpenReadTransaction())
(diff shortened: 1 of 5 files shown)


src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -363,16 +363,16 @@ public void NotifyOnPendingClusterTransaction()
 
         private long? _nextClusterCommand;
      
-        private void ExectueClusterTransactionTask()
+        private async Task ExecuteClusterTransactionTask()
         {
             while (DatabaseShutdown.IsCancellationRequested == false)
             {
-                _hasClusterTransaction.Wait(DatabaseShutdown);
+                await _hasClusterTransaction.WaitAsync(DatabaseShutdown);
```
ayende, line 386: what happens if this throws?

karmeli87: Fixed.

ppekrol (approved):

PR #7336, opened 2018-08-14
[4.1] RavenDB-8351 One node, voting in progress when passive

karmeli87:


ayende:
What is the reasoning for this change?

karmeli87:
We need this so the studio will identify when the node is in passive/candidate state, since right now we rely on the node tag.

See http://issues.hibernatingrhinos.com/issue/RavenDB-8351

ayende (approved):

PR #7342, opened 2018-08-14
[4.1] RavenDB-11473 failed test - CanImportIndexesAndTransformers

efratshenhar:


src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -565,6 +565,7 @@ private void RemoveNodeFromCluster(TransactionOperationContext context, Blittabl
                         if (record.Topology.Count == 0)
                         {
                             DeleteDatabaseRecord(context, index, items, lowerKey, record.DatabaseName);
+                            OnTransactionDispose(context, index);
```
ayende, line 568: Why do we need this here?

ayende (approved):

PR #7351, opened 2018-08-14
RavenDB-11746 Failed to properly close RavenServer (Delegate to an in…

ayende:
…stance method cannot have null 'this'.) when server configuration is broken

### DIFF
--- a/src/Raven.Server/Program.cs
+++ b/src/Raven.Server/Program.cs
@@ -190,10 +190,10 @@ namespace Raven.Server
                         }
                         catch (Exception e)
                         {
-                            var message = e.Message;
+                            string message = null;
                             if (e.InnerException is AddressInUseException)
                             {
-                                message +=
+                                message =
                                     $"{Environment.NewLine}Port might be already in use.{Environment.NewLine}Try running with an unused port.{Environment.NewLine}" +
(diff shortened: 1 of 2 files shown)


src/Raven.Server/Program.cs
```diff
@@ -202,15 +202,23 @@ public static int Main(string[] args)
...
+                            {
                                 Logger.Operations("Failed to initialize the server", e);
+                                Logger.Operations(message);
+                            }
+
                             Console.WriteLine(message);
 
+                            Console.WriteLine();
+
+                            Console.WriteLine(e);
```
ppekrol, line 220: This line reverts assumption from http://issues.hibernatingrhinos.com/issue/RavenDB-11107

ppekrol: maybe wrap this in `#if DEBUG` ?

ayende: We need this to get a clear error for _us_ in prod.
When a user runs into a problem, we need to have the full details.
You can see that the message is still shown first, along with the full error stack that we can use.

ayende: Right now, see what happens when you configure a bad certificate and start the server.

PR #7352, opened 2018-08-14
RavenDB-11608 Better security error message on admin only ops

ayende:


ayende:
Test failure doesn't seem to be related to the code changes

PR #7355, opened 2018-08-15
RavenDB-11748 Using Single() or SingleOrDefault() in a projection fails

ayende:


src/Raven.Client/Util/JavascriptConversionExtensions.cs
```diff
@@ -385,6 +385,8 @@ public override void ConvertToJavascript(JavascriptConversionContext context)
                             }
                             return;
                         }
+                    case "Single":
+                    case "SingleOrDefault":
```
ppekrol: Maybe it would be better to throw for these two methods? It will be confusing for some users that Single/SingleOrDefault will behave same as First/FirstOrDefault.

The exception could be: `Methods 'Single' and 'SingleOrDefault' are not supported in projections. Use 'First' or 'FirstOrDefault' instead`

ayende: I'm not sure that I like that, but I get what you mean. 
The question is whatever the user actually meant to assert `Single` or not?
For `First`, we don't error if there are no values (we translate this a `FirstOrDefault`).
The query is whatever this is likely to be meaningful from the user POV. What do you think?

ppekrol: I think by throwing an exception here we will make our self clear about the situation. We do not support that and the alternative is to use the `First/FirstOrDefault`. If one day we will want to support this then we can add a proper implementation without breaking backward compatibility.

We should not assume that user by executing `Single` wanted to do `First`. 'Single' checks if there is one element and throws otherwise. One user will assume that, another will not. This might result in another YT issue in the future when someone will expect 'the other' behavior.

I suggest staying on the safe side and stating as is.

Bishbulb: I would probably agree that if Single/SingleOrDefault isn't supported, I'd rather see that thrown as an exception. I would probably extend that to First as well, since you're saying that under the covers it really behaves as FirstOrDefault.

ayende: Okay, in this case, then.
However, we can't change `First` behavior now.

Bishbulb: @ayende Could you implement this by throwing an exception in the default case of the switch statement? As a user, I'd prefer to get a runtime error if I'm trying to use an unsupported LINQ expression (be it Single/SingleOrDefault, or something else).

ayende: Yes, I made the modification

ppekrol:
Thx!

PR #7368, opened 2018-08-16
[4.1] RavenDB-11757 Fix RachisTests.SubscriptionsFailover.DistributedRevisi…

karmeli87:
…onsSubscriptio

### DIFF
--- a/src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
+++ b/src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
@@ -144,7 +144,8 @@ namespace Raven.Server.Documents.TcpHandlers
 
                     if (ShouldSendDocumentWithRevisions(_subscription, run, _patch, docsContext, item, revisionTuple, out var transformResult, out var exception) == false)
                     {
-                        includesCmd.AddRange(run.Includes);
+                        if(run != null)
+                            includesCmd.AddRange(run.Includes);
                         if (exception != null)
                         {
                             yield return (item, exception);
--- a/src/Sparrow/AsyncManualResetEvent.cs
+++ b/src/Sparrow/AsyncManualResetEvent.cs
@@ -91,11 +91,21 @@ namespace Sparrow
                 var waitAsync = _tcs.Task;
                 _parent._token.ThrowIfCancellationRequested();
                 var result = await Task.WhenAny(waitAsync, TimeoutManager.WaitFor(timeout, _parent._token)).ConfigureAwait(false);
+
+                if (result.IsFaulted)
+                    throw result.Exception;
+
                 if (_parent._token != CancellationToken.None)
                     return result == waitAsync && !_parent._token.IsCancellationRequested;
 
                 return result == waitAsync;
             }
+        }
+
+        public void SetException(Exception e)
+        {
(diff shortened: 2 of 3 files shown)


src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
```diff
@@ -595,6 +595,9 @@ private bool ShouldTryToReconnect(Exception ex)
         {
             switch (ex)
             {
+                case InvalidOperationException _:
```
ayende: why?

src/Sparrow/AsyncManualResetEvent.cs
```diff
@@ -91,13 +91,23 @@ public async Task<bool> WaitAsync(TimeSpan timeout)
...
                     return result == waitAsync && !_parent._token.IsCancellationRequested;
 
                 return result == waitAsync;
             }
         }
 
+        public void SetException(Exception e)
+        {
+            _token.ThrowIfCancellationRequested();
+            _tcs.SetException(e);
```
ppekrol: We should use here `TrySetException`

ppekrol:
@maximburyak can you review this?

src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
ppekrol: When can we get this?

src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
```diff
@@ -144,7 +144,8 @@ public class SubscriptionDocumentsFetcher
 
                     if (ShouldSendDocumentWithRevisions(_subscription, run, _patch, docsContext, item, revisionTuple, out var transformResult, out var exception) == false)
                     {
-                        includesCmd.AddRange(run.Includes);
+                        if(run != null)
```
maximburyak, line 147: @karmeli87 did you encounter a situation when run was null?

ayende: I did, yes. This can happen if there is no script to run `from Users`

src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
karmeli87: This is what is thrown when an unexpected error is sent.
https://github.com/ravendb/ravendb/blob/a73056b86822a71a6401dd5475ea904cdabed732/src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs#L455

But probably need to set `MaxErroneousPeriod` instead.

src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
karmeli87: e.g. `from Users (Revisions = true) as doc`

ppekrol (approved):

src/Sparrow/AsyncManualResetEvent.cs
maximburyak: Maybe we should add another method for that (TrySetException)

PR #7370, opened 2018-08-16
fix for RavenDB-11754 [v4.1]

myarichuk:


src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -1743,6 +1743,13 @@ public Task<(long Index, object Result)> SendToLeaderAsync(CommandBase cmd)
             return SendToLeaderAsyncInternal(cmd);
         }
 
+        //this is needed for cases where Result or any of its fields are blittable json.
+        //(for example, this is needed for use with AddOrUpdateCompareExchangeCommand, since it returns BlittableJsonReaderObject as result)
+        public Task<(long Index, object Result)> SendToLeaderAsync(TransactionOperationContext context, CommandBase cmd)
```
ayende: I would rather _always_ use this. Is there a reason not to?

myarichuk: The method with *public Task<(long Index, object Result)> SendToLeaderAsync(CommandBase cmd)* signature is used in gazillion of places.
Haven't checked ALL of them, but in some there are duplicate contexts used, and I guess SendToLeaderAsync() with context parameter CAN be used.

Since I am working on tests, I preferred to make minimum necessary changes to code. Do you think I should refactor its usage?

ppekrol: We can do it now or in 4.2. After 4.1 RTM this change would not be a good option.

Maybe it would be better for the method that does not have an external context to do (in debug):
```
#if DEBUG
if (result.Result is BlittableJsonReaderBase)
    throw new InvalidOperationException();
#endif
```

ayende: The problem isn't with direct dependency. It is if it returns _anything_ that holds BJRO

ayende: And the problem with it being used everywhere it what scared me. Doing a deep check is fine with me on debug.

ppekrol: Great, wanted to propose a deep check.

myarichuk: In case of those commands, it returns a tuple where one of tuple's values is blittable

myarichuk: http://issues.hibernatingrhinos.com/issue/RavenDB-11761

myarichuk:
Added recursive code to check for blittables in command results.

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -1738,9 +1739,22 @@ public bool IsPassive()
...
 
-        public Task<(long Index, object Result)> SendToLeaderAsync(CommandBase cmd)
+        public async Task<(long Index, object Result)> SendToLeaderAsync(CommandBase cmd)
         {
-            return SendToLeaderAsyncInternal(cmd);
+            var response = await SendToLeaderAsyncInternal(cmd);
+
+            #if DEBUG
+            
+            //this is a temporary measure until RavenDB-11761 is done
```
ayende: Will probably keep this in place to make sure that we aren't making this mistake, though.

src/Raven.Server/Utils/TypeUtils.cs
```diff
@@ -0,0 +1,98 @@
...
+using Sparrow.Json;
+
+namespace Raven.Server.Utils
+{
+    public static class TypeUtils
+    {
+        //detect if an object is a blittable or has any blittable objects somewhere in its object hierarchy
+        public static bool ContainsBlittableObject(this object obj)
+        {
+            return obj.ContainsBlittableObject(new List<object>());
```
ayende: Use `HashSet` with reference equality comparer, instead

src/Raven.Server/Utils/TypeUtils.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            var type = obj.GetType();
+
+            if (type.IsPointer || type.IsEnum || type.IsCOMObject) //obviously not what we are looking for
+                return false;
+
+            if (obj is BlittableJsonReaderObject || obj is BlittableJsonReaderArray)
+                return true;
+
+            if (typeof(IEnumerable).IsAssignableFrom(type))
```
ayende: You can avoid this line in favor of the one below

ayende (changes requested):

PR #7374, opened 2018-08-16
[2.5] RavenDB-9469 - RavenDB process suddenly starts to consume 100%

grisha-kotler:
http://issues.hibernatingrhinos.com/issue/RavenDB-9469

### DIFF
--- a/Raven.Database/Commercial/ValidateLicense.cs
+++ b/Raven.Database/Commercial/ValidateLicense.cs
@@ -99,10 +99,20 @@ namespace Raven.Database.Commercial
 				},firstTime:firstTime);
 
 				var attributes = new Dictionary<string, string>(alwaysOnAttributes, StringComparer.OrdinalIgnoreCase);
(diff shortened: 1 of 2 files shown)


Raven.Database/Commercial/ValidateLicense.cs
```diff
@@ -99,10 +99,20 @@ private void ExecuteInternal(InMemoryRavenConfiguration config, bool firstTime =
 				},firstTime:firstTime);
 
 				var attributes = new Dictionary<string, string>(alwaysOnAttributes, StringComparer.OrdinalIgnoreCase);
-				foreach (var licenseAttribute in licenseValidator.LicenseAttributes)
-				{
-					attributes[licenseAttribute.Key] = licenseAttribute.Value;
-				}
+
+			    Monitor.Enter(licenseValidator.LicenseAttributesLock);
```
ayende, line 103: Did you get all the places that read it as well? Otherwise we didn't do much.
Can we create a new instance, isntead?

PR #7389, opened 2018-08-19
Raven db 9479

talweiss1982:
This PR is for reviewing the changes while i continue to debug NewtonSoft Behavior under ravendb

### DIFF
--- a/src/Sparrow/Json/BlittableJsonTextWriter.cs
+++ b/src/Sparrow/Json/BlittableJsonTextWriter.cs
@@ -247,10 +247,9 @@ namespace Sparrow.Json
 
             var size = str.Size;
 
-            if (size == 1 && str.IsControlCodeCharacter(out var b))
+            if (size == 1 && char.IsControl((char)str[0]))
             {
-                WriteString($@"\u{b:X4}", skipEscaping: true);
-                return;
+                
             }
 
             var strBuffer = str.Buffer;
--- a/src/Sparrow/Json/Parsing/JsonParserState.cs
+++ b/src/Sparrow/Json/Parsing/JsonParserState.cs
@@ -6,6 +6,7 @@ namespace Sparrow.Json.Parsing
     public unsafe class JsonParserState
     {
         public const int EscapePositionItemSize = 5;
+        public const int ControlCharacterItemSize = 5;
         public byte* StringBuffer;
         public int StringSize;
         public int? CompressedSize;
@@ -43,7 +44,7 @@ namespace Sparrow.Json.Parsing
             return count;
         }
 
-        public static (int MaxEscapes, int ExactControls) FindEscapePositionsMaxSize(string str)
+        public static int FindEscapePositionsMaxSize(string str)
         {
             var count = 0;
             var controlCount = 0;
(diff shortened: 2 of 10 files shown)


src/Sparrow/Json/BlittableJsonTextWriter.cs
```diff
@@ -247,10 +247,9 @@ public void WriteString(LazyStringValue str, bool skipEscaping = false)
 
             var size = str.Size;
 
-            if (size == 1 && str.IsControlCodeCharacter(out var b))
+            if (size == 1 && char.IsControl((char)str[0]))
```
ayende, line 250: This code needs to be removed

src/Sparrow/Json/Parsing/JsonParserState.cs
```diff
@@ -104,10 +116,33 @@ public static void FindEscapePositionsIn(FastList<int> buffer, byte* str, int le
...
+                if (value < 32 || value >= 127 && value <= 159)
+                {
+                    // move rest of buffer 
+                    // write \u0000
+                    // update size
+                    var from = str + i + 1;
+                    var to = str + i + 1 + SizeOfUtf8AsStringInBytesWithoutEscaping;
+                    var sizeToCopy = len - i;
+                    //here we only shifting by 5 bytes since we are going to override the byte at the current position.
+                    Memory.Copy(to, from, sizeToCopy ); //TODO: we must verify that we have enough space in our buffer otherwise we must request a bigger one
```
ayende, line 132: We shouldn't request more, but should validate that we aren't over the max size

src/Sparrow/Json/Parsing/JsonParserState.cs
```diff
@@ -131,5 +166,37 @@ public void Reset()
             Continuation = JsonParserTokenContinuation.None;
             EscapePositions.Clear();
         }
+
+        //This table was programaticly generated 
+        public static readonly byte[][] EscapeCharactersStringsAsBytes = new byte[160][]
```
ayende, line 171: Generate this at runtime, instead. This is not going to be readable long term.

src/Sparrow/Json/Parsing/JsonParserState.cs
maximburyak: @talweiss1982 that todo remark is very important! I think that we may want to perform that modification before that function, moreover, we should first count that amount of "control charachters", according to that, to allocate or reallocate the right memory chunk and only then to perform that code

maximburyak: So after talking with @talweiss1982 , it seems that I've overlooked the modifications to FindEscapePositionsMaxSize, It wtill seems a bit awkward to modify data inside FindEscapePositionsIn, but as an intermidiate solution, it does the work

src/Sparrow/Json/Parsing/JsonParserState.cs
```diff
@@ -104,10 +116,33 @@ public static void FindEscapePositionsIn(FastList<int> buffer, byte* str, int le
...
+                    // move rest of buffer 
+                    // write \u0000
+                    // update size
+                    var from = str + i + 1;
+                    var to = str + i + 1 + SizeOfUtf8AsStringInBytesWithoutEscaping;
+                    var sizeToCopy = len - i;
+                    //here we only shifting by 5 bytes since we are going to override the byte at the current position.
+                    Memory.Copy(to, from, sizeToCopy ); //TODO: we must verify that we have enough space in our buffer otherwise we must request a bigger one
+                    //TODO: find a way to generate the table as pinned
+                    fixed (byte* controlString = EscapeCharactersStringsAsBytes[value])
```
maximburyak, line 134: it's GCHandle.Allocate

PR #7391, opened 2018-08-19
RavenDB-11770 Allow to get a revision of a document before a specific date

ayende:


src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -809,6 +814,33 @@ private static long CountOfRevisions(DocumentsOperationContext context, Slice pr
...
+                // This is because we want to handle out of order revisions from multiple nodes so the local etag
+                // order is different than the last modified order
+                Document result = null;
+                var table = new Table(RevisionsSchema, context.Transaction.InnerTransaction);
+                foreach (var tvr in table.SeekBackwardFrom(RevisionsSchema.Indexes[IdAndEtagSlice], prefixSlice, lastKey, 0))
+                {
+                    var document = TableValueToRevision(context, ref tvr.Result.Reader);
+                    if (document.LastModified > max)
+                        continue;
+                    if(result == null ||
```
maximburyak, line 833: Can we assume here that the revisions are ordered by data here?
Because if we have hundreds of revisions, but the most relevant are the first 50, then we iterate over 150 unneeded revisions..
Also, maybe we could avoid caliing TableValueToRevision for each revision, and instead call to a function that will return the raw dateTime value

ayende: We can get the raw date time yes, but I don't think we can avoid iterating them.
The problem is that the revisions may happen on different nodes, and be notified at different times.
This is a edge case, but we need to be aware of this.

src/Raven.Client/Documents/Session/DocumentSessionRevisions.cs
```diff
@@ -19,6 +21,15 @@ public DocumentSessionRevisions(InMemoryDocumentSessionOperations session) : bas
         {
         }
 
+        public T GetBefore<T>(string id, DateTime before)
```
ppekrol, line 24: @ayende maybe it would be better to add paging and give user more flexibility here?

```
public List<T> GetBefore<T>(string id, DateTime before, start = 0, int pageSize = 25)
```

ayende: I specifically wanted to give a single revision here, because multiple revisions are complex.
They are complex because we cannot assume that revisions are sorted by last modified, and I want to avoid creating a new storage index for this. So we do a single search for the latest revisions before the max date and return just that.

PR #7392, opened 2018-08-19
[4.1] RavenDB-11757 RachisTests.SubscriptionsFailover.DistributedRevisionsS…

karmeli87:
…ubscription


test/Tests.Infrastructure/ClusterTestBase.cs
```diff
@@ -438,6 +438,7 @@ protected async Task<RavenServer> CreateRaftClusterAndGetLeader(int numberOfNode
                 {
                     [RavenConfiguration.GetKey(x=>x.Cluster.MoveToRehabGraceTime)] = "1",
                     [RavenConfiguration.GetKey(x=>x.Cluster.ElectionTimeout)] = _electionTimeoutInMs.ToString(),
+                    [RavenConfiguration.GetKey(x=>x.Cluster.StabilizationTime)] = "1",
```
ayende, line 441: what was the issue here?

karmeli87: On leader change we wait for 5 seconds before the observer starts to analyze the stats.
So in this case we wouldn't change the database topology in time for the node to realize that it need to move the subscription responsibility.

PR #7393, opened 2018-08-19
[4.1] RavenDB-11772 use defaults (80 & 443) when port is null in request 

iftahbe:


src/Raven.Server/Extensions/HttpExtensions.cs
```diff
@@ -35,7 +35,7 @@ public static string ExtractNodeUrlFromRequest(HttpRequest request)
                 var uriBuilder = new UriBuilder(
                     request.Scheme,
                     request.Host.Host,
-                    request.Host.Port ?? 8080,
+                    request.Host.Port ?? (request.IsHttps ? 443 : 80),
```
ayende, line 38: why does this matter, these are the defaults anyway

iftahbe: The port in the request was 80 (implicit) so request.Host.Port was null. This caused us to append 8080 to the node url and that wrong url was written to the topology.

iftahbe: Also, now the test 'AddingRemovedNodeShouldWork' fails. Not sure if it's related to this change yet.

PR #7395, opened 2018-08-19
[4.1] RavenDB-11771 AlertRaised/DatabaseTopologyWarning Write Transaction e…

karmeli87:
…very 0.5 second

Open write transaction only if needed.

### DIFF
--- a/src/Raven.Server/NotificationCenter/NotificationsStorage.cs
+++ b/src/Raven.Server/NotificationCenter/NotificationsStorage.cs
@@ -80,30 +80,32 @@ namespace Raven.Server.NotificationCenter
 
         public bool Store(Notification notification, DateTime? postponeUntil = null)
         {
-            if (Logger.IsInfoEnabled)
-                Logger.Info($"Saving notification '{notification.Id}'.");
-
             using (_contextPool.AllocateOperationContext(out TransactionOperationContext context))
-            using (var tx = context.OpenWriteTransaction())
             {
-                // if previous notification had postponed until value pass this value to newly saved notification
-                var existing = Get(notification.Id, context, tx);
-
-                if (postponeUntil == null)
+                using (var tx = context.OpenReadTransaction())
                 {
-                    if (existing?.PostponedUntil == DateTime.MaxValue) // postponed until forever
-                        return false;
+                    // if previous notification had postponed until value pass this value to newly saved notification
+                    var existing = Get(notification.Id, context, tx);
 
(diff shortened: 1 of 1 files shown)


src/Raven.Server/NotificationCenter/NotificationsStorage.cs
```diff
@@ -80,30 +80,32 @@ public void Initialize(StorageEnvironment environment, TransactionContextPool co
...
+                    }
                 }
 
+                if (Logger.IsInfoEnabled)
+                    Logger.Info($"Saving notification '{notification.Id}'.");
+
                 using (var json = context.ReadObject(notification.ToJson(), "notification", BlittableJsonDocumentBuilder.UsageMode.ToDisk))
+                using (var tx = context.OpenWriteTransaction())
                 {
                     Store(context.GetLazyString(notification.Id), notification.CreatedAt, postponeUntil, json, tx);
```
ayende, line 106: Need to do the check here again, you might have had another tx come in the meantime.

karmeli87: In any case someone will be overwritten here.. We identify the notification by the Id, which should be unique to the case.
But when 2 notifications with the same Id are concurrently stored, I'm not sure if we should care which one will be written.

ayende: OK

src/Raven.Server/NotificationCenter/NotificationsStorage.cs
```diff
@@ -80,30 +80,32 @@ public void Initialize(StorageEnvironment environment, TransactionContextPool co
...
+                            return false;
 
-                    if (existing?.PostponedUntil != null && existing.PostponedUntil.Value > SystemTime.UtcNow)
-                        postponeUntil = existing.PostponedUntil;
+                        if (existing?.PostponedUntil != null && existing.PostponedUntil.Value > SystemTime.UtcNow)
+                            postponeUntil = existing.PostponedUntil;
+                    }
                 }
 
+                if (Logger.IsInfoEnabled)
```
aviviadi, line 100: what happens if something fails above? (should we at least try to log it)

karmeli87: See here: https://github.com/ravendb/ravendb/blob/cbcffe3c89b8b37b57f34ec6510854079f895dee/src/Raven.Server/NotificationCenter/NotificationCenter.cs#L68

PR #7396, opened 2018-08-20
RavenDB-11697 fixing tests

ppekrol:


test/RachisTests/EmergencyOperations.cs
```diff
@@ -24,7 +24,7 @@ public async Task LeaderCanCecedeFromClusterAndNewLeaderWillBeElected()
             old = GetServerTopology(leader);
             new AdminJsConsole(leader,null).ApplyScript(new AdminJsScript
             (
-               @"server.ServerStore.Engine.HardResetToNewCluster();"
+               @"server.ServerStore.Engine.HardResetToNewCluster('A');"
```
ayende, line 27: Why is this now required? We have docs that don't specify this.

PR #7397, opened 2018-08-20
RavenDB-11373 Recording and replaying a sequence of MergedTransaction…

ayende:
…Commands in relation to the transaction in which they were written to the disk including end points & client commands and operations + test


src/Raven.Client/Json/BlittableJsonWriter.cs
```diff
@@ -550,6 +550,25 @@ public override void WriteValue(Uri value)
 
         public override void WriteValue(object value)
         {
+            switch (value)
+            {
+                case BlittableJsonReaderObject readerObject:
+                    if (false == readerObject.HasParent)
```
ayende, line 556: You did too much C :-)

src/Raven.Client/Json/Converters/BlittableJsonConverter.cs
```diff
@@ -0,0 +1,53 @@
...
+                    blittableReader.SkipBlittableInside();
+                }
+
+                return
+                    blittableValue.BelongsToContext(blittableReader.Context) &&
+                    blittableValue.HasParent == false
+                        ? blittableValue
+                        : blittableValue.Clone(blittableReader.Context);
+            }
+            throw new SerializationException($"Try to read {nameof(BlittableJsonReaderObject)} from non {nameof(BlittableJsonReaderObject)} value");
```
ayende: What _is_ the value that you go? We need to provide as many details as possible to make it easy for the user to understand.

src/Raven.Client/Json/Converters/BlittableJsonConverter.cs
```diff
@@ -0,0 +1,53 @@
...
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            writer.WriteValue(value);
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            if (!(reader is BlittableJsonReader blittableReader))
+            {
+                throw new SerializationException($"{nameof(BlittableJsonReader)} must to be used for convert to {nameof(BlittableJsonReaderObject)}");
```
ayende: Please provide good error here on what value _was_ passed here.

src/Raven.Client/Json/Converters/LazyStringValueJsonConverter.cs
```diff
@@ -0,0 +1,42 @@
...
+        public static readonly LazyStringValueJsonConverter Instance = new LazyStringValueJsonConverter();
+
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            writer.WriteValue(value);
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            if (!(reader is BlittableJsonReader blittableReader))
```
ayende: Better error on what was passed here.

src/Raven.Client/Json/Converters/LazyStringValueJsonConverter.cs
```diff
@@ -0,0 +1,42 @@
...
+            {
+                throw new SerializationException($"{nameof(BlittableJsonReader)} must to be used for convert to {nameof(BlittableJsonReaderObject)}");
+            }
+
+            if (blittableReader.Value == null)
+            {
+                return null;
+            }
+
+            //Todo It will be better to change the reader to set the value as LazyStringValue
```
ayende: What if the value is _already_ `LazyBlittableString` ?

src/Raven.Client/Json/Converters/StreamConverter.cs
```diff
@@ -0,0 +1,59 @@
...
+{
+    internal sealed class StreamConverter : RavenJsonConverter
+    {
+        public static readonly StreamConverter Instance = new StreamConverter();
+
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            if (!(writer is BlittableJsonWriter blittableJsonWriter))
+            {
+                throw new SerializationException($"{nameof(BlittableJsonWriter)} must to be used for serialize {nameof(Stream)}");
```
ayende: Better error specify what was the value we did get here

src/Raven.Client/Json/Converters/StreamConverter.cs
```diff
@@ -0,0 +1,59 @@
...
+            }
+
+            if (!(value is Stream stream))
+            {
+                throw new SerializationException($"Try to write non {nameof(Stream)} value with {nameof(StreamConverter)}");
```
ayende: What value did we get here?

src/Raven.Client/Json/Converters/StreamConverter.cs
```diff
@@ -0,0 +1,59 @@
...
+            }
+
+            //Todo To consider if Seek need to be here or should it come ready
```
ayende: You do need it here, and you need to check that you `CanSeek` and error appropriately.
Because you consume the full data, the serialization mutates the state of the object. We want to return it to `0` after the call.

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1767,8 +1767,7 @@ public static LazyStringValue TableValueToId(JsonOperationContext context, int i
         {
             // See format of the lazy string ID in the GetLowerIdSliceAndStorageKey method
```
ayende: Comment needs to move to `GetLazyStringValue` method

src/Raven.Server/Documents/Expiration/ExpiredDocumentsCleaner.cs
```diff
@@ -123,10 +124,30 @@ public DeleteExpiredDocumentsCommand(Dictionary<Slice, List<(Slice LowerId, Lazy
...
+                };
+            }
+        }
+    }
+
+    internal class DeleteExpiredDocumentsCommandDto : TransactionOperationsMerger.IReplayableCommandDto<ExpiredDocumentsCleaner.DeleteExpiredDocumentsCommand>
+    {
+        public ExpiredDocumentsCleaner.DeleteExpiredDocumentsCommand ToCommand(DocumentsOperationContext context, DocumentDatabase database)
+        {
+            Dictionary<Slice, List<(Slice LowerId, LazyStringValue Id)>> expired = Expired.ToDictionary(kv => kv.Key, kv => kv.Value);
```
ayende: Why not clone it?

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -948,4 +966,47 @@ private class WaitForIndexItem
...
+    public class MergedBatchCommandDto : TransactionOperationsMerger.IReplayableCommandDto<BatchHandler.MergedBatchCommand>
+    {
+        public BatchRequestParser.CommandData[] ParsedCommands { get; set; }
+        public Queue<BatchHandler.MergedBatchCommand.AttachmentStream> AttachmentStreams;
+
+        public BatchHandler.MergedBatchCommand ToCommand(DocumentsOperationContext context, DocumentDatabase database)
+        {
+            if (ParsedCommands == null || ParsedCommands.Any() == false)
+            {
+                //Todo To check if empty array can append & if it is make sense to check it here
```
ayende: You don't need the check, we should never get an empty array here.
If we do get it, however, just handle it normally.

src/Raven.Server/Documents/Handlers/BulkInsertHandler.cs
```diff
@@ -206,6 +210,38 @@ public override int Execute(DocumentsOperationContext context)
...
+
+    public class MergedInsertBulkCommandDto : TransactionOperationsMerger.IReplayableCommandDto<BulkInsertHandler.MergedInsertBulkCommand>
+    {
+        public BatchRequestParser.CommandData[] Commands { get; set; }
+
+        public BulkInsertHandler.MergedInsertBulkCommand ToCommand(DocumentsOperationContext context, DocumentDatabase database)
+        {
+            if (Commands == null || Commands.Any() == false) 
+            {
+                //Todo To check if empty array can append & if it is make sense to check it here
```
ayende: Don't need to check here.

src/Raven.Server/Documents/Handlers/TransactionsRecordingHandler.cs
```diff
@@ -0,0 +1,173 @@
...
+            {
+                return await Task.Run(() =>
+                {
+                    long commandsProgress = 0;
+                    var stopwatch = Stopwatch.StartNew();
+                    stopwatch.Start();
+                    foreach (var replayProgress in Database.TxMerger.Replay(replayStream))
+                    {
+                        commandsProgress = replayProgress.CommandsProgress;
+                        //TODO Maybe should be relative to time or size
```
ayende: We typically use the number of operations to push this.
This give the user predictable behavior in most cases, this is fine

src/Raven.Server/Documents/Handlers/TransactionsRecordingHandler.cs
```diff
@@ -0,0 +1,173 @@
...
+                    return new ReplayTrxOperationResult
+                    {
+                        ExecutedCommandsAmount = commandsProgress,
+                        PassedTime = stopwatch.Elapsed
+                    };
+                }, token);
+            }
+            catch (Exception e)
+            {
+                HttpContext.Response.Headers["Connection"] = "close";
```
ayende: Add a comment on why this is required.

src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
```diff
@@ -25,8 +29,9 @@ public class PatchDocumentCommand : TransactionOperationsMerger.MergedTransactio
         private readonly ScriptRunner.SingleRun _runIfMissing;
         private ScriptRunner.ReturnRun _returnRun;
         private ScriptRunner.ReturnRun _returnRunIfMissing;
-        private readonly BlittableJsonReaderObject _patchIfMissingArgs;
-        private readonly BlittableJsonReaderObject _patchArgs;
+
+        private readonly (PatchRequest run, BlittableJsonReaderObject args) _patchIfMissing;
```
ayende: We use `PascalCase` for tuple values.

src/Raven.Server/Documents/Queries/AbstractQueryRunner.cs
```diff
@@ -169,7 +169,8 @@ private static IndexQueryServerSide ConvertToOperationQuery(IndexQueryServerSide
             };
         }
 
-        private class BulkOperationCommand<T> : TransactionOperationsMerger.MergedTransactionCommand where T : TransactionOperationsMerger.MergedTransactionCommand
+        //Todo To check if it make sense to changed the accessibility of this type for adding it to the test list and if not to find alternative
```
ayende: This is fine

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -1317,6 +1348,167 @@ public void Dispose()
                     disposable?.Dispose();
                 }
             }
+
+            public override TransactionOperationsMerger.IReplayableCommandDto<TransactionOperationsMerger.MergedTransactionCommand> ToDto(JsonOperationContext context)
+            {
+                var buffer = new byte[_totalSize];
```
ayende, line 1355: This is a short lived buffer that may be large. Let's use `ArrayPool` here

src/Raven.Client/Documents/Operations/TransactionsRecording/ReplayTrxOperationResult.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using Sparrow.Json.Parsing;
+
+namespace Raven.Client.Documents.Operations.TransactionsRecording
+{
+    public class ReplayTrxOperationResult : IOperationResult
```
ppekrol: We do not use 'Trx'. Change to 'Tx'

src/Raven.Client/Documents/Operations/TransactionsRecording/ReplayTrxProgress.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using Sparrow.Json.Parsing;
+
+namespace Raven.Client.Documents.Operations.TransactionsRecording
+{
+    public class ReplayTrxProgress : IOperationProgress
```
ppekrol: We do not use 'Trx'. Change to 'Tx'

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -1317,6 +1348,167 @@ public void Dispose()
...
+                Marshal.Copy((IntPtr)_buffer, buffer, 0, _totalSize);
+                var strBuffer = Convert.ToBase64String(buffer);
```
ayende, line 1357: This may be big, we need to investigate if we can do this in unmanaged memory more easily. 
`Sparrow.Utils.Base64` give you the option to go from `byte*` of buffer to `byte*` of base 64 without allocating managed memory.

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -1317,6 +1348,167 @@ public void Dispose()
...
+                SourceDatabaseId = SourceDatabaseId
+            };
+            var replicationItems = ReplicatedItemDtos.Select(d => d.ToItem(context)).ToList();
+
+            var replicatedAttachmentStreams = ReplicatedAttachmentStreams.Select(i => CreateReplicationAttachmentStream(context, i)).ToArray();
+            var replicationHandler = new IncomingReplicationHandler(database, replicationItems, connectionInfo, database.ReplicationLoader, replicatedAttachmentStreams);
+
+            unsafe
+            {
+                var buffer = Convert.FromBase64String(Buffer);
```
ayende, line 1438: Same here, let's avoid these big allocations.

src/Raven.Server/Documents/TransactionOperationsMerger.cs
```diff
@@ -65,9 +91,35 @@ public void Start()
...
-        public abstract class MergedTransactionCommand
+        public interface IRecordableCommand
         {
-            public abstract int Execute(DocumentsOperationContext context);
+            IReplayableCommandDto<MergedTransactionCommand> ToDto(JsonOperationContext context);
+        }
+
+        public interface IReplayableCommandDto<out T> where T : MergedTransactionCommand
+        {
+            T ToCommand(DocumentsOperationContext context, DocumentDatabase database); // receive some parameters here?
```
ayende: Can remove the comment, I think

src/Raven.Server/Documents/TransactionOperationsMerger.cs
```diff
@@ -110,6 +162,178 @@ private static void ThrowTxMergerWasDisposed()
...
+                private bool _isDisposed = false;
+
+                public EnabledRecordingState(TransactionOperationsMerger txOpMerger)
+                {
+                    _txOpMerger = txOpMerger;
+                }
+
+                public void Init(string filePath)
+                {
+                    var recordingFileStream = new FileStream(filePath, FileMode.Create);
```
ayende: If we provide a bad file path (`f:\foo`, no permissions, file already exists, etc). What will happen?

src/Raven.Server/Documents/TransactionOperationsMerger.cs
```diff
@@ -848,5 +1125,283 @@ public void Dispose()
...
+                            try
+                            {
+                                var cmd = DeserializeCommand(strType, readersItr.Current, context, peepingTomStream);
+                                cmd.ExecuteDirectly(txCtx);
+                                commandsProgress++;
+                                UpdateGlobalReplicationInfoBeforeCommit(txCtx);
+                            }
+                            catch (Exception)
+                            {
+                                //TODO To accept exceptions that was thrown while recording
```
ayende: Reminder on this

src/Raven.Server/Documents/TransactionOperationsMerger.cs
```diff
@@ -848,5 +1125,283 @@ public void Dispose()
...
+            {
+                reader.Init(commandReader);
+                var dto = DeserializeCommandDto(type, jsonSerializer, reader, peepingTomStream);
+                return dto.ToCommand(context, _parent);
+            }
+        }
+
+        private IReplayableCommandDto<MergedTransactionCommand> DeserializeCommandDto(string type, JsonSerializer jsonSerializer, BlittableJsonReader reader, PeepingTomStream peepingTomStream)
+        {
+            //Todo Maybe should be simplified by convention & reflection
```
ayende: This is fine

src/Raven.Server/Documents/TransactionOperationsMerger.cs
```diff
@@ -848,5 +1125,283 @@ public void Dispose()
...
+                {
+                    Context = Encodings.Utf8.GetString(peepingTomStream.PeepInReadStream());
+                }
+                catch (Exception e)
+                {
+                    Context = "Failed to generated peepedWindow: " + e;
+                }
+            }
+
+            public override string Message => base.Message + ";\rContext:\r" + Context;
```
ayende: Use `\r\n` or `Environment.NewLine` here.

src/Raven.Server/Json/Converters/SliceJsonConverter.cs
```diff
@@ -0,0 +1,72 @@
...
+{
+    internal sealed class SliceJsonConverter : RavenJsonConverter
+    {
+        public static readonly SliceJsonConverter Instance = new SliceJsonConverter();
+
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            if (!(writer is BlittableJsonWriter blittableJsonWriter))
+            {
+                throw new SerializationException($"{nameof(BlittableJsonWriter)} must to be used for write {nameof(LazyStringValue)}");
```
ayende: Better error here, what did we get?

src/Raven.Server/Json/Converters/SliceJsonConverter.cs
```diff
@@ -0,0 +1,72 @@
...
+            }
+
+            if (!(value is Slice slice))
+            {
+                throw new SerializationException($"Try to write non {nameof(Slice)} value with {nameof(SliceJsonConverter)}");
```
ayende: What did we get?

src/Raven.Server/Json/Converters/SliceJsonConverter.cs
```diff
@@ -0,0 +1,72 @@
...
+            }
+
+            var buffer = new byte[slice.Size];
```
ayende: Base we use `Base64` and avoid the managed allocations?

src/Raven.Server/Json/Converters/SliceJsonConverter.cs
```diff
@@ -0,0 +1,72 @@
...
+            var strValue = Convert.ToBase64String(buffer);
+
+            blittableJsonWriter.WriteValue(strValue);
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            if (!(reader is BlittableJsonReader blittableReader))
+            {
+                throw new SerializationException($"{nameof(BlittableJsonReader)} must to be used for convert to {nameof(BlittableJsonReaderObject)}");
```
ayende: Better error

src/Raven.Server/Json/Converters/SliceJsonConverter.cs
```diff
@@ -0,0 +1,72 @@
...
+                return null;
+            }
+
+            if (!(blittableReader.Value is string strValue))
+            {
+                throw new SerializationException($"Try to read {nameof(Slice)} from non string value");
+            }
+
+            //TODO To create slice by JsonOperationContext
+            if (!(blittableReader.Context is DocumentsOperationContext context))
```
ayende: Why do you need `DocumentsOperationContext` and not just `JsonOperationContext`?

src/Sparrow/Json/BlittableJsonReaderObject.cs
```diff
@@ -782,6 +782,13 @@ private static void InvalidAttemptToCopyNestedObject()
                 "Attempted to copy a nested object. This will actually copy the whole object, which is probably not what you wanted.");
         }
 
+        //There is an issue when write blittable & at the same time clone another blittable on the same context
+        //Todo To check if should remove this method after issue is fixed
+        public BlittableJsonReaderObject CloneOnTheSameContext()
```
ayende: What is the status of this?

src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -526,6 +526,13 @@ private unsafe LazyStringValue GetLazyString(StringSegment field, bool longLived
             }
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public unsafe LazyStringValue GetLazyStringValue(byte* ptr)
+        {
+            var size = BlittableJsonReaderBase.ReadVariableSizeInt(ptr, 0, out var offset);
```
ayende: Bring the comment on the structure here.

src/Raven.Server/Documents/Handlers/TransactionsRecordingHandler.cs
```diff
@@ -0,0 +1,173 @@
...
+using Sparrow.Json;
+
+namespace Raven.Server.Documents.Handlers
+{
+    public class TransactionsRecordingHandler : DatabaseRequestHandler
+    {
+        [RavenAction("/databases/*/transactions/replay", "POST", AuthorizationStatus.ValidUser)]
+        public async Task ReplayRecording()
+        {
+            var operationId = GetLongQueryString("operationId");
```
ppekrol, line 21: The operation should not be passed from the outside (unless used via Studio).

This endpoint should return the OperationId only (check other similar methods e.g. in SmugglerHandler)

The ReplayOperation in client should return OperationIdResult

src/Raven.Server/Documents/Handlers/TransactionsRecordingHandler.cs
```diff
@@ -0,0 +1,173 @@
...
+
+                using (ContextPool.AllocateOperationContext(out DocumentsOperationContext context))
+                using (var writer = new BlittableJsonTextWriter(context, ResponseBodyStream()))
+                {
+                    context.Write(writer, result.ToJson());
+                }
+            }
+        }
+
+        private async Task<IOperationResult> DoReplay(
```
ppekrol: This method does not have to be async?

src/Raven.Server/Documents/Handlers/TransactionsRecordingHandler.cs
```diff
@@ -0,0 +1,173 @@
...
+            using (ContextPool.AllocateOperationContext(out DocumentsOperationContext context))
+            {
+                var json = await context.ReadForMemoryAsync(RequestBodyStream(), null);
+                var parameters = JsonDeserializationServer.StartTransactionsRecordingOperationParameters(json);
+                if (parameters.File == null)
+                {
+                    ThrowRequiredPropertyNameInRequest(nameof(parameters.File));
+                }
+
+                var command = new TransactionsRecordingCommand(
```
ppekrol, line 100: isn't this an overkill? can we do `Database.TxMerger.Start(parameters.File)` ?

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -1317,6 +1348,167 @@ public void Dispose()
...
+        public IncomingReplicationHandler.MergedDocumentReplicationCommand ToCommand(DocumentsOperationContext context, DocumentDatabase database)
+        {
+            var connectionInfo = new IncomingConnectionInfo
+            {
+                SourceDatabaseId = SourceDatabaseId
+            };
+            var replicationItems = ReplicatedItemDtos.Select(d => d.ToItem(context)).ToList();
+
+            var replicatedAttachmentStreams = ReplicatedAttachmentStreams.Select(i => CreateReplicationAttachmentStream(context, i)).ToArray();
+            var replicationHandler = new IncomingReplicationHandler(database, replicationItems, connectionInfo, database.ReplicationLoader, replicatedAttachmentStreams);
```
ppekrol, line 1434: Can we avoid doing that?

src/Raven.Server/Documents/Replication/ResolveConflictOnReplicationConfigurationChange.cs
```diff
@@ -161,6 +163,21 @@ public override int Execute(DocumentsOperationContext context)
...
                 return count;
             }
+
+            public override TransactionOperationsMerger.IReplayableCommandDto<TransactionOperationsMerger.MergedTransactionCommand> ToDto(JsonOperationContext context)
+            {
+                // The LowerId created as in memory LazyStringValue
+                // so EscapePositions set to empty to avoid reference to escape bytes (after string bytes) while serializing
+                foreach (var conflict in _resolvedConflicts)
+                {
+                    conflict.ResolvedConflict.LowerId.EscapePositions = Array.Empty<int>();
```
ppekrol: Don't we need to make a copy of them then?

src/Raven.Server/Documents/TransactionOperationsMerger.cs
```diff
@@ -848,5 +1125,283 @@ public void Dispose()
...
+
+        private class StartRecordingDetails : RecordingDetails
+        {
+            private const string DetailsType = "StartRecording";
+            public string Version { get; }
+
+            public StartRecordingDetails()
+            : base(DetailsType)
+            {
+                Version = ServerVersion.FullVersion;
```
ppekrol: We typically use RavenVersionAttribute.Instance.Build only

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
```diff
@@ -705,6 +705,9 @@ public JsonSerializer CreateSerializer()
             jsonSerializer.Converters.Add(ParametersConverter.Instance);
             jsonSerializer.Converters.Add(JsonLinqEnumerableConverter.Instance);
             jsonSerializer.Converters.Add(JsonIMetadataDictionaryConverter.Instance);
+            jsonSerializer.Converters.Add(BlittableJsonConverter.Instance);
+            jsonSerializer.Converters.Add(LazyStringValueJsonConverter.Instance);
+            jsonSerializer.Converters.Add(StreamConverter.Instance);
```
ppekrol: Not sure if we want to have these as default in client? Those are used only on server-side?

src/Raven.Client/Documents/Operations/TransactionsRecording/ReplayTransactionsRecordingOperation.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.IO;
+using System.Net.Http;
+using Raven.Client.Documents.Conventions;
+using Raven.Client.Http;
+using Raven.Client.Json.Converters;
+using Sparrow.Json;
+
+namespace Raven.Client.Documents.Operations.TransactionsRecording
+{
+    public class ReplayTransactionsRecordingOperation : IMaintenanceOperation<ReplayTrxOperationResult>
```
ppekrol: As written below. This needs to return the OperationIdResult

src/Raven.Server/Documents/TransactionOperationsMerger.cs
```diff
@@ -848,5 +1125,283 @@ public void Dispose()
...
+            var recordingState = _recordingState;
+            if (recordingState != null)
+            {
+                recordingState.Shutdown();
+                _waitHandle.Set();
+            }
+        }
+
+        public IEnumerable<ReplayProgress> Replay(Stream replayStream)
+        {
```
ppekrol: Can we move the implementation to some dedicated helper class?

ppekrol (changes requested):

src/Raven.Server/Documents/Handlers/TransactionsRecordingHandler.cs
ayende: We have to do it this way to avoid starting recording in the middle of a transaction.
This way, we always start from a brand new tx.

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
ayende: Yes, it is better to avoid them there by default.

src/Raven.Server/Documents/Handlers/TransactionsRecordingHandler.cs
ppekrol: Makes sense.

src/Raven.Client/Json/Converters/LazyStringValueJsonConverter.cs
haludi: https://github.com/haludi/ravendb/blob/f13bbd5c65c488aa7c227f6de329e735570a3d44/src/Raven.Client/Json/BlittableJsonReader.cs#L205

The reader convert all the serialized string types to string before it set the value

src/Raven.Server/Documents/Expiration/ExpiredDocumentsCleaner.cs
haludi: https://www.newtonsoft.com/json/help/html/SerializationGuide.htm#Dictionarys

At the time I did not look into it in depth, but using the dictionary rose a problem
Now I've found that there is special regard to the key of the dictionary
that cause it not going through the SliceJsonConverter

src/Raven.Server/Documents/Handlers/TransactionsRecordingHandler.cs
haludi: So how to decide how many operations should it be?

src/Raven.Server/Documents/Expiration/ExpiredDocumentsCleaner.cs
ayende: Use `new Dictionary<...>(Expires)`

`ToDictionary` is invoking two delegates per instance and can throw if there are duplicates (which isn't an issue when reading from a dictionary, but in that case, why bother with `ToDictionary` ?

src/Raven.Server/Documents/Handlers/TransactionsRecordingHandler.cs
ayende: 1024

src/Raven.Server/Json/Converters/SliceJsonConverter.cs
haludi: [`Slice.From(context.Allocator, buffer, ByteStringType.Immutable, out var slice);`](https://github.com/ravendb/ravendb/pull/7397/files/1ab8eb87a068abb6e93c176b114e45468a6c1b1b#diff-d9907d27cc6f398f4ccad4f9d7d9823bR62)

As I understand it, only the TransactionOperationContext hold the Allocator that I need to create Slice. 
I can create it by myself but in this case, I will need to dispose it.
For that, I need to bring the allocator from the outside, the reader in this case,
and release the memory after the replay is done (or after the command execute).
It seems to me very messy to hold two objects (context & allocator) than handle memory in the reader.
The other option I can think about is to hold an allocator in the JsonOperationContext but this is a huge change to the system. Maybe lazy allocator?
Or there is something I didn't think about...

ayende: No, that is fine.

src/Raven.Server/Documents/Handlers/TransactionsRecordingHandler.cs
haludi: As I understand it, the request is executed while the file stream is uploaded.
So, if the OperationId will be passed at the end of the request the client will not be able to monitor the operation that can take a lot of time.

src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
haludi: You talk about the tuple properties, right?

ayende: `(PatchRequest Run, BlittableJsonReaderObject Args)`
Yes

src/Raven.Server/Documents/Replication/ResolveConflictOnReplicationConfigurationChange.cs
haludi: In memory LazyStringValue doesn't have escape characters.
Maybe I will change the comment?

> The LowerId created as in memory LazyStringValue that <b> doesn't have escape characters </b>
so EscapePositions set to empty to avoid reference to escape bytes (after string bytes) while serializing

ayende: change the comment, yes

src/Raven.Client/Json/Converters/BlittableJsonConverter.cs
haludi: done

src/Raven.Client/Json/Converters/BlittableJsonConverter.cs
haludi: done

src/Raven.Client/Json/Converters/LazyStringValueJsonConverter.cs
haludi: done

src/Raven.Client/Json/Converters/StreamConverter.cs
haludi: done

src/Raven.Client/Json/Converters/StreamConverter.cs
haludi: done

src/Raven.Client/Json/Converters/StreamConverter.cs
haludi: done

src/Raven.Server/Documents/DocumentsStorage.cs
haludi: done

src/Raven.Server/Documents/Expiration/ExpiredDocumentsCleaner.cs
haludi: done

src/Raven.Server/Documents/Handlers/BatchHandler.cs
haludi: done

src/Raven.Server/Documents/Handlers/BulkInsertHandler.cs
haludi: done

src/Raven.Server/Documents/Handlers/TransactionsRecordingHandler.cs
haludi: done

src/Raven.Server/Documents/Handlers/TransactionsRecordingHandler.cs
haludi: done

src/Raven.Server/Documents/Queries/AbstractQueryRunner.cs
haludi: done

src/Raven.Server/Documents/TransactionOperationsMerger.cs
haludi: done

src/Raven.Server/Documents/TransactionOperationsMerger.cs
haludi: done

src/Raven.Server/Documents/TransactionOperationsMerger.cs
haludi: done

src/Raven.Server/Json/Converters/SliceJsonConverter.cs
haludi: done

src/Raven.Server/Json/Converters/SliceJsonConverter.cs
haludi: done

src/Raven.Server/Json/Converters/SliceJsonConverter.cs
haludi: done

src/Sparrow/Json/JsonOperationContext.cs
haludi: done

src/Raven.Client/Documents/Operations/TransactionsRecording/ReplayTrxOperationResult.cs
haludi: done

src/Raven.Client/Documents/Operations/TransactionsRecording/ReplayTrxProgress.cs
haludi: done

src/Raven.Server/Documents/Handlers/TransactionsRecordingHandler.cs
haludi: done

src/Raven.Server/Documents/TransactionOperationsMerger.cs
haludi: done

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
haludi: done

src/Raven.Client/Json/Converters/LazyStringValueJsonConverter.cs
```diff
@@ -18,7 +18,8 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
         {
             if (!(reader is BlittableJsonReader blittableReader))
             {
-                throw new SerializationException($"{nameof(BlittableJsonReader)} must to be used for convert to {nameof(BlittableJsonReaderObject)}");
+                throw new SerializationException(
+                    $"Try to read {nameof(LazyStringValue)} property/field by {reader.GetType()} witch is unsuitable reader. Should use {nameof(BlittableJsonReader)}");
```
ayende: witch

src/Raven.Client/Json/Converters/StreamConverter.cs
```diff
@@ -16,17 +16,23 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
         {
             if (!(writer is BlittableJsonWriter blittableJsonWriter))
             {
-                throw new SerializationException($"{nameof(BlittableJsonWriter)} must to be used for serialize {nameof(Stream)}");
+                throw new SerializationException(
+                    $"Try to write {nameof(LazyStringValue)} property/field by {writer.GetType()} witch is unsuitable reader. Should use {nameof(BlittableJsonWriter)}");
```
ayende: witch

src/Raven.Client/Json/Converters/StreamConverter.cs
```diff
@@ -16,17 +16,23 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
...
             }
 
             if (!(value is Stream stream))
             {
-                throw new SerializationException($"Try to write non {nameof(Stream)} value with {nameof(StreamConverter)}");
+                throw new SerializationException($"Try to write {value.GetType()} witch is not {nameof(Stream)} value with {nameof(StreamConverter)}");
```
ayende: witch

src/Raven.Client/Json/Converters/StreamConverter.cs
```diff
@@ -35,7 +41,8 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
         {
             if (!(reader is BlittableJsonReader blittableReader))
             {
-                throw new SerializationException($"{nameof(BlittableJsonReader)} must to be used for deserialize {nameof(Stream)}");
+                throw new SerializationException(
+                    $"Try to read {nameof(Stream)} property/field by {reader.GetType()} witch is unsuitable reader. Should use {nameof(BlittableJsonReader)}");
```
ayende: witch

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -606,9 +606,14 @@ private CollectionName GetFirstConflictCollection(DocumentsOperationContext cont
                     return null;
                 return Database.DocumentsStorage.ExtractCollectionName(context, conflicts[0].Collection);
             }
+
+            public override TransactionOperationsMerger.IReplayableCommandDto<TransactionOperationsMerger.MergedTransactionCommand> ToDto(JsonOperationContext context)
+            {
+                throw new NotImplementedException();
```
ayende: Why?

src/Raven.Server/Documents/Handlers/CountersHandler.cs
```diff
@@ -199,6 +199,11 @@ void LoadDocument()
                 return CountersDetail.Counters.Count;
             }
 
+            public override TransactionOperationsMerger.IReplayableCommandDto<TransactionOperationsMerger.MergedTransactionCommand> ToDto(JsonOperationContext context)
+            {
+                throw new NotImplementedException();
```
ayende: Why?

src/Raven.Server/Documents/TransactionOperationsMerger.cs
haludi: I choose to move the ownership of the record stream to the TxMerger and create the file while the start record request so the error will be arrived from there.
As I implemented it now the record will overwrite the existing file.
I can think about three ways to handle it:
1. To overwrite
2. To return error
3. To add an extension (maybe current time) that differentiate it from the existing

src/Raven.Server/Documents/Replication/ResolveConflictOnReplicationConfigurationChange.cs
haludi: done

src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
haludi: done

src/Raven.Server/Documents/Handlers/BatchHandler.cs
haludi: Just for compiling

src/Raven.Server/Documents/Handlers/CountersHandler.cs
haludi: Just for compiling

ayende: We need to actually implement these, no?

src/Raven.Server/Documents/TransactionOperationsMerger.cs
haludi: done

src/Raven.Server/Documents/Handlers/BatchHandler.cs
haludi: For compiling only

src/Raven.Server/Documents/Handlers/CountersHandler.cs
haludi: For compiling only

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
haludi: Avoiding what? Linq?

src/Raven.Server/Documents/Handlers/CountersHandler.cs
haludi: done

ayende: Okay, so that was the first pass, then you are going to impl these?

src/Raven.Client/Json/Converters/BlittableJsonConverter.cs
```diff
@@ -0,0 +1,55 @@
...
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            writer.WriteValue(value);
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            if (!(reader is BlittableJsonReader blittableReader))
+            {
+                throw new SerializationException(
```
maximburyak: It's better to move the throws to seperate functions

src/Raven.Client/Json/Converters/StreamConverter.cs
```diff
@@ -0,0 +1,66 @@
...
+
+            if (false == stream.CanSeek)
+            {
+                throw new SerializationException("Try to serialize stream that can't seek");
+            }
+
+            stream.Seek(0, SeekOrigin.Begin);
+            var buffer = stream.ReadData();
+            stream.Seek(0, SeekOrigin.Begin);
+            var strBuffer = Convert.ToBase64String(buffer);
```
maximburyak: Maybe we could consider reducing allocations here, by using LazyStringValue, that is being fed by Convert.ToBase64CharArray

src/Raven.Server/Documents/Expiration/ExpiredDocumentsCleaner.cs
```diff
@@ -123,10 +124,37 @@ public DeleteExpiredDocumentsCommand(Dictionary<Slice, List<(Slice LowerId, Lazy
...
 
-            public override int Execute(DocumentsOperationContext context)
+            protected override int ExecuteCmd(DocumentsOperationContext context)
             {
                 return DeletionCount = _database.DocumentsStorage.ExpirationStorage.DeleteExpiredDocuments(context, _expired);
             }
+
+            public override TransactionOperationsMerger.IReplayableCommandDto<TransactionOperationsMerger.MergedTransactionCommand> ToDto(JsonOperationContext context)
+            {
+                var keyValuePairs = new List<KeyValuePair<Slice, List<(Slice LowerId, LazyStringValue Id)>>>();
```
maximburyak: We could use an Array here and call Array.Copy, rather than using list

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
maximburyak: @haludi I think @ppekrol was referencing the creationg of IncomingReplicationHandler, which is pretty heavy object, that does a lot of stuff under the hood. I think that at this stage it will be a bit hard to get rid of it, because of the none trivial code in MergedDocumentReplicationCommand.Execute. But yes, it looks like a big TODO

src/Raven.Client/Documents/Operations/TransactionsRecording/ReplayTransactionsRecordingOperation.cs
haludi: https://github.com/ravendb/ravendb/pull/7397#discussion_r212181012

src/Raven.Client/Json/Converters/LazyStringValueJsonConverter.cs
haludi: done

src/Raven.Client/Json/Converters/StreamConverter.cs
haludi: done

src/Raven.Client/Json/Converters/StreamConverter.cs
haludi: done

src/Raven.Client/Json/Converters/StreamConverter.cs
haludi: done

src/Raven.Server/Documents/Handlers/BatchHandler.cs
haludi: done

src/Raven.Server/Documents/Handlers/CountersHandler.cs
haludi: done

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
haludi: Issue created
[RavenDB-11852](http://issues.hibernatingrhinos.com/issue/RavenDB-11852)

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
haludi: Issue created
[RavenDB-11852](http://issues.hibernatingrhinos.com/issue/RavenDB-11852)

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
haludi: Issue created
[RavenDB-11852](http://issues.hibernatingrhinos.com/issue/RavenDB-11852)

src/Raven.Server/Documents/TransactionOperationsMerger.cs
haludi: [RavenDB-11751](http://issues.hibernatingrhinos.com/issue/RavenDB-11751)

src/Raven.Server/Json/Converters/SliceJsonConverter.cs
haludi: Issue created
[RavenDB-11852](http://issues.hibernatingrhinos.com/issue/RavenDB-11852)

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
haludi: An issue created
[RavenDB-11857](http://issues.hibernatingrhinos.com/issue/RavenDB-11857)

src/Raven.Client/Json/Converters/BlittableJsonConverter.cs
haludi: I move the check to a base class and I think now it's better.
What do you think?

src/Raven.Client/Json/Converters/StreamConverter.cs
haludi: An issue created
[RavenDB-11852](http://issues.hibernatingrhinos.com/issue/RavenDB-11852)

src/Raven.Server/Documents/Expiration/ExpiredDocumentsCleaner.cs
haludi: done

PR #7399, opened 2018-08-20
[4.1] RavenDB-11771 AlertRaised/DatabaseTopologyWarning opens write transac…

karmeli87:
…tion

- Better error when the request executor can't find the host.
- The max value of notification postpone should be DateTime.MaxValue.

### DIFF
--- a/src/Raven.Client/Http/RequestExecutor.cs
+++ b/src/Raven.Client/Http/RequestExecutor.cs
@@ -1216,9 +1216,9 @@ namespace Raven.Client.Http
             {
                 Url = request.RequestUri.ToString(),
                 Message = e.Message,
-                Error = e.ToString(),
+                Error = $"An exception occurred while contacting {request.RequestUri}.{Environment.NewLine}{e}.",
                 Type = e.GetType().FullName
-            }, HttpStatusCode.InternalServerError));
+            }, HttpStatusCode.ServiceUnavailable));
         }
 
         protected Task _firstTopologyUpdate;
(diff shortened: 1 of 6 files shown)


src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -735,18 +736,24 @@ protected virtual bool TryLoadFromCache(JsonOperationContext context)
                 }
                 catch (HttpRequestException e) // server down, network down
                 {
+                    Exception ex = e;
+                    if (e.InnerException is SocketException se && se.ErrorCode == 11001)
```
ayende: Is that going to be the same in Linux? Is that the same in Windows?

ayende: I would rather add the URL at all times always.

ayende (changes requested):

src/Raven.Client/Http/RequestExecutor.cs
ppekrol: It might also be different when SocketsHttpHandler is not used?

src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
```diff
@@ -117,6 +114,33 @@ public async Task Run(CancellationToken token)
             }
         }
 
+        private readonly Dictionary<string, long> _lastLogs = new Dictionary<string, long>();
+
+        private void LogMessage(string message, Exception e = null, bool info = true)
+        {
+            if (_iteration % 10_000 == 0)
+                _lastLogs.Clear();
```
ppekrol, line 122: Why do we want to keep those logs here? We are not doing anything with them?

Debug only code?

karmeli87: I don't want to spam the log file, so I keep track of the logs and write the _same_ message to the file only if enough (10 is enough?) iterations have passed. But that dictionary has to be cleaned.. let's say.. once in a 10,000 iterations.

PR #7408, opened 2018-08-21
[4.1] RavenDB-11781 - Failed to deploy a Map-Reduce index

grisha-kotler:
http://issues.hibernatingrhinos.com/issue/RavenDB-11781

### DIFF
--- a/src/Raven.Server/Documents/Indexes/IndexStore.cs
+++ b/src/Raven.Server/Documents/Indexes/IndexStore.cs
@@ -379,8 +379,14 @@ namespace Raven.Server.Documents.Indexes
 
             var instance = IndexCompilationCache.GetIndexInstance(definition, _documentDatabase.Configuration); // pre-compile it and validate
 
+            if (NeedToUpdateIndex(definition, out Index currentIndex, out var validateMapReduceCollectionName) == false)
+            {
+                Debug.Assert(currentIndex != null);
+                return currentIndex;
+            }
+
             if (definition.Type == IndexType.MapReduce)
-                MapReduceIndex.ValidateReduceResultsCollectionName(definition, instance, _documentDatabase);
+                MapReduceIndex.ValidateReduceResultsCollectionName(definition, instance, _documentDatabase, validateMapReduceCollectionName);
 
             var command = new PutIndexCommand(definition, _documentDatabase.Name);
 
@@ -396,6 +402,39 @@ namespace Raven.Server.Documents.Indexes
                 throw new IndexCreationException($"Failed to create static index: {definition.Name}, the cluster is probably down. " +
                                                  $"Node {_serverStore.NodeTag} state is {_serverStore.LastStateChangeReason()}", toe);
             }
+        }
+
+        private bool NeedToUpdateIndex(IndexDefinition definition, out Index currentIndex, out bool validateMapReduceCollectionName)
+        {
+            currentIndex = GetIndex(definition.Name);
+            if (currentIndex == null)
+            {
+                // new index
+                validateMapReduceCollectionName = true;
+                return true;
+            }
+
+            var creationOptions = GetIndexCreationOptions(definition, currentIndex, out var _);
+            if (creationOptions != IndexCreationOptions.Noop)
+            {
+                // update/create an index definition
+                validateMapReduceCollectionName = creationOptions == IndexCreationOptions.UpdateWithoutUpdatingCompiledIndex;
+                return true;
+            }
+
+            var replacementIndexName = Constants.Documents.Indexing.SideBySideIndexNamePrefix + definition.Name;
+            if (GetIndex(replacementIndexName) != null)
+            {
+                // original index and side by side index exist
+                // we restored the original index definition
+                // need to delete the side by side index
+                // this is done on each server separately so we need to send the create index command
+                validateMapReduceCollectionName = false;
+                return true;
+            }
+
+            validateMapReduceCollectionName = false;
+            return false;
         }
 
         public async Task<Index> CreateIndex(IndexDefinitionBase definition)
--- a/src/Raven.Server/Documents/Indexes/MapReduce/Static/MapReduceIndex.cs
+++ b/src/Raven.Server/Documents/Indexes/MapReduce/Static/MapReduceIndex.cs
@@ -70,7 +70,8 @@ namespace Raven.Server.Documents.Indexes.MapReduce.Static
         public static MapReduceIndex CreateNew(IndexDefinition definition, DocumentDatabase documentDatabase, bool isIndexReset = false)
         {
             var instance = CreateIndexInstance(definition, documentDatabase.Configuration);
-            ValidateReduceResultsCollectionName(definition, instance._compiled, documentDatabase, isIndexReset);
+            ValidateReduceResultsCollectionName(definition, instance._compiled, documentDatabase,
+                validateMapReduceCollectionName: isIndexReset == false);
 
             instance.Initialize(documentDatabase,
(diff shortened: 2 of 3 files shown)


src/Raven.Server/Documents/Indexes/MapReduce/Static/MapReduceIndex.cs
```diff
@@ -79,7 +80,7 @@ public static MapReduceIndex CreateNew(IndexDefinition definition, DocumentDatab
             return instance;
         }
 
-        public static void ValidateReduceResultsCollectionName(IndexDefinition definition, StaticIndexBase index, DocumentDatabase database, bool isIndexReset = false)
+        public static void ValidateReduceResultsCollectionName(IndexDefinition definition, StaticIndexBase index, DocumentDatabase database, bool validateMapReduceCollectionName)
```
arekpalinski, line 83: I don't really like the name of the renamed parameter because it suggests that it'll skip the validation completely (`ValidateReduceResultsCollectionName()` vs `validateMapReduceCollectionName:false`). What we actually want to check or not check is to verify that the collection is empty so maybe `checkIfCollectionEmpty` of something like that.

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -379,8 +379,14 @@ public async Task<Index> CreateIndex(IndexDefinition definition)
 
             var instance = IndexCompilationCache.GetIndexInstance(definition, _documentDatabase.Configuration); // pre-compile it and validate
 
+            if (NeedToUpdateIndex(definition, out Index currentIndex, out var validateMapReduceCollectionName) == false)
```
arekpalinski, line 382: Why do we need to handle that here? Based on the included test I understand the issue was that we didn't recognize the index definition has _not_ been changed meanwhile, right?

Why we didn't filter it out here?:
https://github.com/ravendb/ravendb/blob/v4.1/src/Raven.Client/ServerWide/DatabaseRecord.cs#L93

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -398,6 +404,39 @@ public async Task<Index> CreateIndex(IndexDefinition definition)
...
+                // original index and side by side index exist
+                // we restored the original index definition
+                // need to delete the side by side index
+                // this is done on each server separately so we need to send the create index command
+                validateMapReduceCollectionName = false;
+                return true;
+            }
+
+            validateMapReduceCollectionName = false;
+            return false;
```
arekpalinski, line 437: This code is seems to have very similar logic as `IndexStore.HandleStaticIndexChange` why do we need it here?

grisha-kotler: There is no need to send a RAFT command if the index didn't change.
HandleStaticIndexChange is after applying the RAFT command.

src/Raven.Server/Documents/Indexes/IndexStore.cs
grisha-kotler: This is before we try to add it to the database record

src/Raven.Server/Documents/Indexes/IndexStore.cs
ayende: Why should it matter? This is a rare thing, so not much point in optimizing it.

grisha-kotler: Calling IndexCreation.CreateIndexes() on app restart with 100 indexes for example

ayende: Which would be rare, and would usually be batched anyway, so one call only

ayende: Talking with @grisha-kotler, we have an issue because we want to make this validation _before_ we accept the command into the cluster.

arekpalinski: I wonder if we really need to have this validation _before_ sending the command. What if we let it send it to the cluster and do the validation then (on the actual attempt to create an index - unless we already detected that such index def already exists and filtered it out)

That validation is always done here:

https://github.com/ravendb/ravendb/blob/v4.1/src/Raven.Server/Documents/Indexes/MapReduce/Static/MapReduceIndex.cs#L73

grisha-kotler: This is too late. The caller for creating the index won't see this error and the current index will be errored (if we changed the definition).

arekpalinski: I've commented this validation:

https://github.com/ravendb/ravendb/blob/v4.1/src/Raven.Server/Documents/Indexes/IndexStore.cs#L383

and run `OutputReduceToCollectionTests` - all passed

grisha-kotler: The diffrenece now is that an errored index will be created

arekpalinski: Oh I see. That is because:

https://github.com/ravendb/ravendb/blob/v4.1/src/Raven.Server/Documents/Indexes/IndexStore.cs#L240

What if we _won't_ create faulty index def if the caught exception is `IndexInvalidException`. In order words if we failed due to validation we just error and won't create errored index - it seems to be reasonable

ayende: what about the previous index def at this point?

arekpalinski: just make a quick test and everything seems to be ok. I got the validation error and previous index def remained unchanged

grisha-kotler: same here. (the replacement index failed to be created which is what we need)

arekpalinski: hmm.. my approach doesn't seem to be okay. if we skip the above validation we'll end up with DatabaseRecord containing invalid index def :/

src/Raven.Server/Documents/Indexes/MapReduce/Static/MapReduceIndex.cs
```diff
@@ -150,7 +151,7 @@ public static void ValidateReduceResultsCollectionName(IndexDefinition definitio
                 }
             }
 
-            if (isIndexReset == false)
+            if (validateMapReduceCollectionName)
             {
```
ppekrol, line 155: If we will check here that we have index with same name and same output collection then we can skip that 'if'? Wouldn't that be the case?

If we previously allowed a creation of index 'MyIndex' that Outputs to collection 'Companies' then if we have still that index and we are overriding it then the assumption that it was valid configuration still takes place?

ppekrol: https://github.com/ppekrol/ravendb/commit/22e90f9c337c58a3ca6e42513d2128b729656497 seems to work?

grisha-kotler: You are still creating an errored index.
(I don't think that there is a test for that)
1) deploy a map-reduce index with output to collection with some documents
2) modify the index definition and deploy.
There will be an errored index (side by side one in this case) while it shouldn't been even deployed.

ppekrol: Correct.

We need a test for this case.

PR #7422, opened 2018-08-22
RavenDB-11787 Optimistic concurrency should thrown not supported exce…

karmeli87:
…ption when used in cluster transactions

### DIFF
--- a/src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
+++ b/src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
@@ -826,6 +826,9 @@ more responsive application.
                 {
                     case CommandType.PUT:
                     case CommandType.DELETE:
+                        if (command.ChangeVector != null)
+                            throw new NotSupportedException($"Optimistic concurrency for '{command.Id}' is not supported when using a cluster transaction.");
+                        break;
                     case CommandType.CompareExchangeDELETE:
                     case CommandType.CompareExchangePUT:
                         break;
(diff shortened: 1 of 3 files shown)


src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -788,6 +788,10 @@ internal SaveChangesData PrepareForSaveChanges()
             DeferredCommands.Clear();
             DeferredCommandsDictionary.Clear();
 
+            if (UseOptimisticConcurrency && TransactionMode == TransactionMode.ClusterWide)
+                throw new NotSupportedException(
```
ppekrol: Can we move this throwing to separate method?

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -816,6 +816,10 @@ internal void ValidateClusterTransaction(SaveChangesData result)
             if (TransactionMode != TransactionMode.ClusterWide)
                 return;
 
+            if (UseOptimisticConcurrency)
```
ayende, line 819: You also need to do this check for every individual check that might have been configured.
We can set only a specific document to be optimistic concurrency checked.

ayende (changes requested):

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
karmeli87: Done.

ayende (approved):

PR #7431, opened 2018-08-23
Ravendb 11795

aviv86:


test/SlowTests/Issues/RavenDB-11795.cs
```diff
@@ -0,0 +1,85 @@
...
+                {
+                    var query = session.Query<BookingIndex.Result, BookingIndex>()
+                        .Where(x => x.FullName == "Alex Me")
+                        .Select(x => new {
+                            FullName = x.FullName,
+                            StartDate = x.Start.ToString("dd.MM.yyyy")
+                        });
+
+                    Assert.Equal("from index 'BookingIndex' as x where x.FullName = $p0 " +
+                                 "select { FullName : x.FullName, StartDate : toDateString(new Date(Date.parse(x.Start)), \"dd.MM.yyyy\") }"
```
arekpalinski: Will `toDateString()` support _any_ custom format that you can use in c#? (https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings)

aviv86: it should, it is implemented in c# in `ScriptRunner` :
https://github.com/ravendb/ravendb/blob/be4c5d0d81754bb0747a0f966302b56202673a9d/src/Raven.Server/Documents/Patch/ScriptRunner.cs#L653-L656
should I add more tests with other formats?

src/Raven.Client/Util/JavascriptConversionExtensions.cs
```diff
@@ -1166,6 +1166,24 @@ private DateTimeSupport()
 
             public override void ConvertToJavascript(JavascriptConversionContext context)
             {
+                if (context.Node is MethodCallExpression mce &&
+                    mce.Method.DeclaringType == typeof(DateTime) &&
```
ayende: Let's change things around.
Instead of specializing only for DateTime, let's support any `ToString`. I want to be able to use this for `this.Age.ToString('000')`, for example.
We also want to support passing the culture.

ayende (changes requested):

src/Raven.Client/Util/JavascriptConversionExtensions.cs
maximburyak: That's not something that is supported in javascript "out of the box", it will require writing some custom jint functions that will do that from c#. That's a .net functionality, that I'm not sure we want to "port" to RQL at all, and if we do, it's not a trivial move and we better think it through..

src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -630,6 +633,97 @@ private static JsValue ConvertJsTimeToTimeSpanString(JsValue self, JsValue[] arg
...
+                if (split.Length < 2 ||
+                    split[0] != nameof(CultureInfo))
+                {
+                    format = arg;
+                }
+                else
+                {
+                    switch (split[1])
+                    {
+                        case "CurrentCulture":
```
ppekrol: I think that we want to support actual culture over the wire with InvariantCulture being `null` or empty string `''`

e.g.
`x.Start.ToString(CultureInfo.InvariantCulture)` => `toStringWithFormat(new Date(Date.parse(x.Start)))`
`x.Start.ToString(CultureInfo.InvariantCulture)` => `toStringWithFormat(new Date(Date.parse(x.Start)), '')`
`x.Start.ToString(CultureInfo.InvariantCulture)` => `toStringWithFormat(new Date(Date.parse(x.Start)), null)`

and when we have a different culture (let say `en-US`)
`x.Start.ToString(CultureInfo.CurrentCulture)` => `toStringWithFormat(new Date(Date.parse(x.Start)), 'en-US')`

ppekrol (changes requested):

src/Raven.Server/Documents/Patch/ScriptRunner.cs
aviv86: done

test/SlowTests/Issues/RavenDB-11795.cs
```diff
@@ -190,7 +229,45 @@ public void NumberToStringWithCulture()
                         });
 
                     Assert.Equal("from Bookings as x where x.FirstName = $p0 " +
-                                 "select { Number : toStringWithFormat(x.Number, \"CultureInfo.InvariantCulture\") }"
+                                 "select { Number : toStringWithFormat(x.Number, \"\") }"
```
ppekrol: `toStringWithFormat(x.Number)` will work also?

src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -113,6 +113,8 @@ public class SingleRun
             public bool ReadOnly;
             private readonly ConcurrentLruRegexCache _regexCache = new ConcurrentLruRegexCache(1024);
 
+            private static Dictionary<string, CultureInfo> _cultureInfoDict;
```
ppekrol: Move this to `CultureHelper` and do this:

```
        private static readonly Lazy<Dictionary<string, CultureInfo>> CulturesLazy = new Lazy<Dictionary<string, CultureInfo>>(() => CultureInfo.GetCultures(CultureTypes.AllCultures).ToDictionary(x => x.Name, x => x, StringComparer.OrdinalIgnoreCase));

        internal static Dictionary<string, CultureInfo> Cultures => CulturesLazy.Value;
```

aviv86: done

PR #7432, opened 2018-08-23
RavenDB-11799 Fix Initialize server with the right url when start with port 0

IdanHaim:


ayende (approved):

ppekrol:
@IdanHaim please check test failures

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -536,7 +536,7 @@ public void Initialize()
 
             _engine = new RachisConsensus<ClusterStateMachine>(this);
             _engine.BeforeAppendToRaftLog += BeforeAppendToRaftLog;
-            var myUrl = Configuration.Core.PublicServerUrl.HasValue ? Configuration.Core.PublicServerUrl.Value.UriValue : Configuration.Core.ServerUrls[0];
+            var myUrl = Configuration.Core.PublicServerUrl.HasValue ? Configuration.Core.PublicServerUrl.Value.UriValue : GetNodeHttpServerUrl();
```
gregolsky: I think when getting url using `GetNodeHttpServerUrl();` public server url takes priority over everything else, so you don't need `Configuration.Core.PublicServerUrl.HasValue` since it's done inside it AFAIR

https://github.com/ravendb/ravendb/blob/v4.1/src/Raven.Server/Config/Categories/CoreConfiguration.cs#L84

IdanHaim: I think that you are right and we can remove it.

PR #7439, opened 2018-08-24
Make sure the transactions are open when deserializing blittables (RavenDB-11793)

ppekrol:


ayende (approved):

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
```diff
@@ -460,21 +460,21 @@ public Task GetCertificates()
                             using (context.OpenReadTransaction())
                                 localCertKeys = ServerStore.Cluster.GetCertificateKeysFromLocalState(context).ToList();
 
-                            foreach (var localCertKey in localCertKeys)
+                            using (context.OpenReadTransaction())
```
maximburyak, line 463: Can we shorten the transaction time by performing the deserialization during the transaction and only then closing it?

ayende: Doesn't matter that much, the overall duration is roughly the same

PR #7453, opened 2018-08-26
RavenDB-11816 - Replace legacy properties when migrating indexes from a v3.x dump

grisha-kotler:
http://issues.hibernatingrhinos.com/issue/RavenDB-11816

### DIFF
--- a/src/Raven.Server/Smuggler/Documents/Processors/IndexProcessor.cs
+++ b/src/Raven.Server/Smuggler/Documents/Processors/IndexProcessor.cs
@@ -236,7 +236,7 @@ namespace Raven.Server.Smuggler.Documents.Processors
             if (string.IsNullOrWhiteSpace(str))
                 return str;
 
-            str = str.Replace("new Raven.Abstractions.Linq.DynamicList", string.Empty);
+            str = str.Replace("new Raven.Abstractions.Linq.DynamicList", "DynamicArray");
             var regex = new Regex(@"([\w_.]+)\.__document_id");
             return regex.Replace(str, "Id($1)");
         }
--- a/src/Raven.Server/Smuggler/Documents/Processors/IndexProcessor.cs
+++ b/src/Raven.Server/Smuggler/Documents/Processors/IndexProcessor.cs
@@ -236,7 +236,7 @@ namespace Raven.Server.Smuggler.Documents.Processors
             if (string.IsNullOrWhiteSpace(str))
                 return str;
 
-            str = str.Replace("new Raven.Abstractions.Linq.DynamicList", "DynamicArray");
+            str = str.Replace("new Raven.Abstractions.Linq.DynamicList", "new DynamicArray");
             var regex = new Regex(@"([\w_.]+)\.__document_id");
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Smuggler/Documents/Processors/IndexProcessor.cs
```diff
@@ -225,6 +231,16 @@ private static IndexDefinition ReadLegacyIndexDefinition(BlittableJsonReaderObje
             return indexDefinition;
         }
 
+        private static string ReplaceLegacyProperties(string str)
+        {
+            if (string.IsNullOrWhiteSpace(str))
+                return str;
+
+            str = str.Replace("new Raven.Abstractions.Linq.DynamicList", string.Empty);
```
ppekrol: I think it should be changed to 'DynamicArray' ?
Check here:
https://github.com/ravendb/ravendb/blob/v4.1/src/Raven.Client/Documents/Indexes/ExpressionStringBuilder.cs#L1464

ppekrol: in v3.x we had: https://github.com/ravendb/ravendb/blob/v3.5/Raven.Client.Lightweight/Indexes/ExpressionStringBuilder.cs#L1588

src/Raven.Server/Smuggler/Documents/Processors/IndexProcessor.cs
```diff
@@ -142,6 +144,8 @@ private static IndexDefinition ReadLegacyIndexDefinition(BlittableJsonReaderObje
 
             if (indexDefinition.Maps != null)
             {
+                indexDefinition.Maps = indexDefinition.Maps.Select(ReplaceLegacyProperties).ToHashSet();
```
ayende, line 147: Shouldn't use _hash_ set. The user will want to get them in the same order, and if they have duplicates (rare, but possible) we want to maintain them.

ppekrol: Talked with @ayende . HashSet is valid at this point. No changes needed here.

src/Raven.Server/Smuggler/Documents/Processors/IndexProcessor.cs
grisha-kotler: done

src/Raven.Server/Smuggler/Documents/Processors/IndexProcessor.cs
```diff
@@ -225,6 +231,16 @@ private static IndexDefinition ReadLegacyIndexDefinition(BlittableJsonReaderObje
             return indexDefinition;
         }
 
+        private static string ReplaceLegacyProperties(string str)
+        {
+            if (string.IsNullOrWhiteSpace(str))
+                return str;
+
+            str = str.Replace("new Raven.Abstractions.Linq.DynamicList", "DynamicArray");
```
ppekrol: `DynamicArray` -> `new DynamicArray` ?

PR #7456, opened 2018-08-27
RavenDB-11810 Add icon-os

kwiato:


ppekrol:
This will be handled in a separate PR

PR #7465, opened 2018-08-27
RavenDB-11825 Allow customization of topology files creation location

ppekrol:


src/Raven.Client/Documents/Conventions/DocumentConventions.cs
```diff
@@ -532,6 +535,27 @@ public OperationStatusFetchMode OperationStatusFetchMode
...
+            set
+            {
+                AssertNotFrozen();
+
+                if (string.IsNullOrWhiteSpace(value))
+                    throw new ArgumentNullException(nameof(value));
+
+                var directory = new DirectoryInfo(value);
+                if (directory.Exists == false)
+                    throw new InvalidOperationException("Topology cache location directory does not exist. Please create the directory first.");
```
arekpalinski, line 553: Wouldn't we create it if it doesn't exist? See here:

https://github.com/ravendb/ravendb/blob/b6d16d116e6fce7c22c401a4c8f939aa95b55eee/src/Raven.Client/Http/ClusterTopologyLocalCache.cs#L71

We still need to check if we have write permissions to this directory

ayende: Agree, and we need to handle failure of the creation of the directory as well

PR #7483, opened 2018-08-29
RavenDB-11196 JsBlittableBridge.Translate should not mutate the Blitt…

talweiss1982:
…ableObjectInstance

### DIFF
--- a/src/Raven.Server/Documents/Patch/JsBlittableBridge.cs
+++ b/src/Raven.Server/Documents/Patch/JsBlittableBridge.cs
@@ -455,7 +455,7 @@ namespace Raven.Server.Documents.Patch
             foreach (var modificationKvp in obj.OwnValues)
             {
                 //We already iterated through those properties while iterating the original properties set.
-                if(modifiedProperties.Contains(modificationKvp.Key))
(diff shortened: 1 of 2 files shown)


src/Raven.Server/Documents/Patch/JsBlittableBridge.cs
```diff
@@ -443,6 +454,10 @@ private void WriteBlittableInstance(BlittableObjectInstance obj, bool isRoot, bo
 
             foreach (var modificationKvp in obj.OwnValues)
             {
+                //We already iterated through those properties while iterating the original properties set.
+                if(modifiedProperties.Contains(modificationKvp.Key))
```
ayende: Can't this be null here? (`OwnValue that doesn't `existInObject` ?)

ayende (changes requested):

src/Raven.Server/Documents/Patch/JsBlittableBridge.cs
talweiss1982: yes it can

ayende (approved):

PR #7485, opened 2018-08-29
RavenDB-11773 Forbid specifying the same port that server should be u…

ml054:
…sing as current setup wizard is

### DIFF
--- a/src/Raven.Studio/typescript/models/wizard/unsecureSetup.ts
+++ b/src/Raven.Studio/typescript/models/wizard/unsecureSetup.ts
@@ -28,12 +28,28 @@ class unsecureSetup {
     }
     
     private initValidation() {
+        const currentHttpPort = window.location.port || "80";
+        
         this.port.extend({
-            number: true
+            number: true,
+            notEqual: {
+                params: currentHttpPort,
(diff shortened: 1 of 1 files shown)


src/Raven.Studio/typescript/models/wizard/unsecureSetup.ts
```diff
@@ -28,12 +28,28 @@ class unsecureSetup {
     }
     
     private initValidation() {
+        const currentHttpPort = window.location.port || "80";
```
ayende, line 31: Technically, they can run the setup while using HTTPS if they re-run by editing the `settings.json` to say `SetupMode=Initial`. Can we add the check for 443 here?

ppekrol: Yes

ppekrol:
```
const currentHttpPort = window.location.port || (window.location.protocol === "https:" ? "443" : "80")
```

ayende: Yes

PR #7486, opened 2018-08-29
 RavenDB-11837 - Patching with counters fails to update the correct document

grisha-kotler:
http://issues.hibernatingrhinos.com/issue/RavenDB-11837

### DIFF
--- a/src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
+++ b/src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
@@ -165,9 +165,26 @@ namespace Raven.Server.Documents.Patch
                     return 1;
                 }
 
-                if (_run.UpdateDocumentCounters)
+                if (_run.UpdatedDocumentCounterIds != null)
                 {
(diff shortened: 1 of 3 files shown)


grisha-kotler:
@aviv86 Can you review this?

aviv86 (commented):
LGTM

src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
```diff
@@ -165,9 +165,26 @@ public override int Execute(DocumentsOperationContext context)
                     return 1;
                 }
 
-                if (_run.UpdateDocumentCounters)
+                if (_run.UpdatedDocumentCounterIds != null)
                 {
-                    modifiedDocument = UpdateCountersInMetadata(context, modifiedDocument, originalDocument);
+                    foreach (var docId in _run.UpdatedDocumentCounterIds)
```
maximburyak, line 170: Maybe we could sort those first

grisha-kotler: Why does it matter?

PR #7489, opened 2018-08-29
RavenDB-11842 added 'maxTasks' query string parameter

ppekrol:


src/Raven.Server/Smuggler/Documents/Handlers/SmugglerHandler.cs
```diff
@@ -220,8 +220,9 @@ public async Task PostImportDirectory()
 
         private async Task BulkImport(BlockingCollection<Func<Task<Stream>>> files, string directory)
         {
+            var maxTasks = GetIntValueQueryString("maxTasks", required: false) ?? ProcessorInfo.ProcessorCount / 2;
             var results = new ConcurrentQueue<SmugglerResult>();
-            var tasks = new Task[Math.Max(1, ProcessorInfo.ProcessorCount / 2)];
+            var tasks = new Task[Math.Max(1, maxTasks)];
```
maximburyak, line 225: Maybe we should limit this one to max of ProcessorInfo.ProcessorCount, so max task will allow to be less than that value, but not more than this one

PR #7493, opened 2018-08-30
RavenDB-11469 Map-reduce:  Handling reduce key hash collisions when O…

arekpalinski:
…utputReduceToCollection is used. Additionally:

- using index context instead of ShortTermSingleUse
- fixed missing disposal of OutputReduceToCollectionCommand

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -1422,7 +1422,7 @@ namespace Raven.Server.Documents.Indexes
                 using (CurrentIndexingScope.Current =
                     new CurrentIndexingScope(DocumentDatabase.DocumentsStorage, databaseContext, Definition, indexContext, GetOrAddSpatialField))
                 {
-                    var writeOperation = new Lazy<IndexWriteOperation>(() => IndexPersistence.OpenIndexWriter(indexContext.Transaction.InnerTransaction));
+                    var writeOperation = new Lazy<IndexWriteOperation>(() => IndexPersistence.OpenIndexWriter(indexContext.Transaction.InnerTransaction, indexContext));
 
                     try
                     {
--- a/src/Raven.Server/Documents/Indexes/Persistence/Lucene/OutputReduceIndexWriteOperation.cs
+++ b/src/Raven.Server/Documents/Indexes/Persistence/Lucene/OutputReduceIndexWriteOperation.cs
@@ -19,12 +19,13 @@ namespace Raven.Server.Documents.Indexes.Persistence.Lucene
     {
         private readonly OutputReduceToCollectionCommand _outputReduceToCollectionCommand;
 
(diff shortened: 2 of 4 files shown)


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1422,7 +1422,7 @@ public bool DoIndexingWork(IndexingStatsScope stats, CancellationToken cancellat
                 using (CurrentIndexingScope.Current =
                     new CurrentIndexingScope(DocumentDatabase.DocumentsStorage, databaseContext, Definition, indexContext, GetOrAddSpatialField))
                 {
-                    var writeOperation = new Lazy<IndexWriteOperation>(() => IndexPersistence.OpenIndexWriter(indexContext.Transaction.InnerTransaction));
+                    var writeOperation = new Lazy<IndexWriteOperation>(() => IndexPersistence.OpenIndexWriter(indexContext.Transaction.InnerTransaction, indexContext));
```
ayende, line 1425: Isn't the indexContext going to be disposed soon? Can the lazy be evaluated _after_ the context has been disposed?

src/Raven.Server/Documents/Indexes/Persistence/Lucene/OutputReduceIndexWriteOperation.cs
```diff
@@ -65,55 +64,78 @@ public override void DeleteReduceResult(LazyStringValue reduceKeyHash, IndexingS
...
             }
 
             public override int Execute(DocumentsOperationContext context)
             {
-                foreach (var reduceDocument in _reduceDocuments)
+                foreach (var reduceKeyHash in _reduceKeyHashesToDelete)
                 {
-                    var key = reduceDocument.Key;
-                    if (reduceDocument.IsDelete)
+                    var deleteResult = _database.DocumentsStorage.Delete(context, GetOutputDocumentKey(reduceKeyHash), null);
```
ayende, line 99: What happens if we have a hash collision from _another_ batch on unrelated doc?

src/Raven.Server/Documents/Indexes/Index.cs
arekpalinski: No, `indexContext` will be disposed at the end of the batch:
https://github.com/ravendb/ravendb/blob/83a41c8341566392a6ff02063018bb0589129f37/src/Raven.Server/Documents/Indexes/Index.cs#L1416

The sequence of actions is: 
- index writer commit
- transaction commit
- dispose contex

We're sure that lazy is evaluated always _before_ the context disposal

src/Raven.Server/Documents/Indexes/Persistence/Lucene/OutputReduceIndexWriteOperation.cs
arekpalinski: If we have hash collision then we're guaranteed that map results will be put into _the same_ reduce tree. Here we output docs per reduce tree so we're good.

Another batch of docs with collision would end up into putting them to the same reduce tree - we always output _all_ documents from a given reduce tree

ayende (approved):

PR #7530, opened 2018-09-03
RavenDB-11747 => make sure ByteStringMemoryCache properly uses NativeMemoryCleaner

myarichuk:
Because the ByteStringMemoryCache was a struct, it passed the wrong reference to RegisterLowMemoryHandler(), thus not executing the low memory event when needed.

### DIFF
--- a/src/Raven.Server/Program.cs
+++ b/src/Raven.Server/Program.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
@@ -16,6 +17,7 @@ using Raven.Server.Utils;
 using Raven.Server.Utils.Cli;
 using Sparrow.Logging;
 using Sparrow.Platform;
+using Sparrow.Utils;
 
 namespace Raven.Server
 {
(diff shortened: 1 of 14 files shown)


src/Raven.Server/Program.cs
```diff
@@ -317,6 +319,82 @@ private static bool RunInteractive(RavenServer server)
...
             Console.WriteLine("Showing stats, press any key to close...");
+            ContinuouslyWriteServerStats(server);
+
+            Console.ReadKey(true);
+            Console.WriteLine();
+            Console.WriteLine($"Stats halted.");
+        }
+
+        private static readonly Random Random = new Random();
+        public static  (
```
ayende, line 330: This is ugly, let's introduce an actual struct/class for this

src/Sparrow/ByteString.cs
```diff
@@ -467,7 +467,7 @@ public interface IByteStringAllocator
     /// <summary>
     /// This class implements a direct allocator, mostly used for testing.   
     /// </summary>
-    public struct ByteStringDirectAllocator : IByteStringAllocator
+    public class ByteStringDirectAllocator : IByteStringAllocator
```
ayende, line 470: This is a _big_ change!
@redknightlois Can you also comment on this?

src/Sparrow/ByteString.cs
```diff
@@ -679,7 +681,7 @@ public int SizeLeft
                 get { return (int)(End - Current); }
             }
         }
-
+        private static readonly TAllocator Allocator = new TAllocator();
```
ayende, line 684: I'm pretty sure that this isn't thread safe.

src/Sparrow/ByteString.cs
```diff
@@ -494,8 +494,12 @@ static ByteStringMemoryCache()
             Cleaner = new NativeMemoryCleaner<SegmentStack, UnmanagedGlobalSegment>(SegmentsPool, LowMemoryFlag, TimeSpan.FromMinutes(1), TimeSpan.FromMinutes(1));
 
             ThreadLocalCleanup.ReleaseThreadLocalState += CleanForCurrentThread;
+        }
 
-            LowMemoryNotification.Instance.RegisterLowMemoryHandler(ByteStringContext.Allocator);
+        public ByteStringMemoryCache()
+        {
+            //each instance of this should handle low memory...
+            LowMemoryNotification.Instance.RegisterLowMemoryHandler(this);
```
ayende, line 502: Note that the state here is all completely static. Better to create a dedicated class that would do the cleanup than change these to structs, I think

ayende (changes requested):

src/Sparrow/ByteString.cs
redknightlois: This will kill the performance of the allocator.

redknightlois:
You need to register the context itself to propagate the low memory signal. That is how it is being done on the generalized allocators.

PR #7538, opened 2018-09-04
RavenDB-11892 Memory leak in ByteStringContext

karmeli87:
Resetting the context doesn't free the memory as long as we had only 2 segments.

So with many small allocation, this caused to the hot segments to grow to their maximal value of 16 MB.
And in presence of many contexts we ended up in a high memory usage.

Since ByteStringContext is mostly used for converting from string to slice document ids and different keys for voron tress/indexes, we do not expect to have large segments.

Therefor we changed the default size of minimum block from 64k to 4k and the maximal segment size to 2MB.

### DIFF
--- a/src/Sparrow/ByteString.cs
+++ b/src/Sparrow/ByteString.cs
@@ -747,8 +747,17 @@ namespace Sparrow
             _externalFastPoolCount = 0;
             _externalCurrentLeft = (int)(_externalCurrent.End - _externalCurrent.Start) / _externalAlignedSize;
 
+            Debug.Assert(_wholeSegments.Count >= 2);
+            // We need to make ensure that the _internalCurrent is linked to an unmanaged segment
+            var index = _wholeSegments.Count - 1;
+            if (_wholeSegments[index] == _externalCurrent)
+            {
+                index = _wholeSegments.Count - 2;
+            }
+            _internalCurrent = _wholeSegments[index];
+
             _internalCurrent.Current = _internalCurrent.Start;
-            _externalCurrent.Current = _externalCurrent.Start;
+            _externalCurrent.Current = _externalCurrent.Start; // no need to reset it, always whole section
 
             _currentlyAllocated = 0;
 
@@ -757,10 +766,11 @@ namespace Sparrow
 
             for (int i = 0; i < _wholeSegments.Count; i++)
             {
-                if (_wholeSegments[i] == _internalCurrent || _wholeSegments[i] == _externalCurrent)
+                var segment = _wholeSegments[i];
+                if (segment == _internalCurrent || segment == _externalCurrent)
                     continue;
 
-                ReleaseSegment(_wholeSegments[i]);
+                ReleaseSegment(segment);
                 _wholeSegments.RemoveAt(i);
                 i--;
             }
--- a/src/Sparrow/ByteString.cs
+++ b/src/Sparrow/ByteString.cs
@@ -626,7 +626,7 @@ namespace Sparrow
 
     public sealed class ByteStringContext : ByteStringContext<ByteStringMemoryCache>
     {
-        public const int MinBlockSizeInBytes = 64 * 1024; // If this is changed, we need to change also LogMinBlockSize.
+        public const int MinBlockSizeInBytes = 4 * 1024; // If this is changed, we need to change also LogMinBlockSize.
         public const int MaxAllocationBlockSizeInBytes = 256 * MinBlockSizeInBytes;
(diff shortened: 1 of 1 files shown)


karmeli87:
@redknightlois can you please review this.

src/Sparrow/ByteString.cs
```diff
@@ -626,7 +626,7 @@ public class SegmentStack : StackHeader<UnmanagedGlobalSegment>
 
     public sealed class ByteStringContext : ByteStringContext<ByteStringMemoryCache>
     {
-        public const int MinBlockSizeInBytes = 64 * 1024; // If this is changed, we need to change also LogMinBlockSize.
+        public const int MinBlockSizeInBytes = 4 * 1024; // If this is changed, we need to change also LogMinBlockSize.
```
ayende, line 629: @redknightlois Can you suggest what change is required in `LogMinBlockSize`?

redknightlois: The calculation is Log(2, 4*1024) ... those are the actual potential buckets you could have given the minimum block size in bytes we are defining.

src/Sparrow/ByteString.cs
```diff
@@ -809,7 +809,7 @@ private ByteString AllocateExternal(byte* valuePtr, int size, ByteStringType typ
             {
                 if (_externalCurrentLeft == 0)
                 {
-                    var tmp = Math.Min(16 * Constants.Size.Megabyte, _allocationBlockSize * 2);
+                    var tmp = Math.Min(2 * Constants.Size.Megabyte, _allocationBlockSize * 2);
                     AllocateExternalSegment(tmp);
```
redknightlois, line 824: No biggy here, we are reserving less (which is essentially what I am doing on the new allocators).

src/Sparrow/ByteString.cs
```diff
@@ -735,9 +735,6 @@ public void Reset()
             if (_disposed)
                 ThrowObjectDisposed();
 
-            if (_wholeSegments.Count == 2)
-                return; // nothing to do
-
             Array.Clear(_internalReusableStringPoolCount, 0, _internalReusableStringPoolCount.Length);
```
redknightlois, line 739: This is the actual issue, we are not resetting the pointers. I wouldnt perform both changes simultaneously (fix + size change) because you can be impacting performance hard... Fix the issue where you reset the actual pointers, check if the problem continue to be repro and if that is the case then perform the size change.

src/Sparrow/ByteString.cs
```diff
@@ -1013,6 +1013,7 @@ private ByteString AllocateWholeSegment(int length, ByteStringType type)
 
             var byteString = Create(segment.Current, length, segment.Size, type);
             segment.Current += byteString._pointer->Size;
+            _currentlyAllocated += size;
```
redknightlois: `_currentlyAllocated` only track whole segments (which are the only actual memory allocated from the OS). The only place where `_currentlyAllocated` should count is there. To me this is not the case.

src/Sparrow/ByteString.cs
myarichuk: We cannot reset pointers since we have places where we want to keep pointers until a call to Reset().
For example [here](https://github.com/ravendb/ravendb/blob/v4.1/src/Voron/Data/Tables/TableSchema.cs#L366) - we want to keep the Slice alive until the transaction commits (and the context is reset when it is returned to the pool)

redknightlois (commented):
EDIT: Deleted

src/Sparrow/ByteString.cs
redknightlois: But this is under `Reset`, therefore after that the memory is not guaranteed to exist.

src/Sparrow/ByteString.cs
ayende: Okay, so we should change it to 12

src/Sparrow/ByteString.cs
ayende: Yes, but the problem is that callers have allocated memory and did _not_ give it back.
We have to reset the state of the allocator back to the original

src/Sparrow/ByteString.cs
karmeli87: As I see it we are using this value only for debugging.. But even so `AllocateWholeSegment` calls `AllocateSegment(size)` so it is an actual allocation.

redknightlois: What I mean is that the only place where `_currentlyAllocated` must be updated is right after we allocate on `AllocateSegment(size)` when it hits the actual allocator...

PR #7539, opened 2018-09-04
RavenDB-11893

aviv86:
http://issues.hibernatingrhinos.com/issue/RavenDB-11893

### DIFF
--- a/src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
+++ b/src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
@@ -186,6 +186,8 @@ namespace Raven.Server.Documents.Patch
                             }
                         }
                     }
+
+                    _run.UpdatedDocumentCounterIds.Clear();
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
```diff
@@ -186,6 +186,8 @@ public override int Execute(DocumentsOperationContext context)
                             }
                         }
                     }
+
+                    _run.UpdatedDocumentCounterIds.Clear();
```
ayende: Did this mean that we were updating _all_ the counters for all the previous documents on each operation?

ppekrol: This should go to `SingleRun.Reset` probably?

aviv86: Yes

PR #7540, opened 2018-09-04
RavenDB-11894 Embedded should not inherit ASPNETCORE environment variables from parent process

ppekrol:


ayende (commented):
Why does this matter?

ppekrol:
https://github.com/ravendb/ravendb/blob/9e47eadebc18abf734295e21c13c2d5ed62221ce/src/Raven.Server/Web/System/StudioHandler.cs#L290

They are changing `env.ContentRootPath` (and probably more) and this breaks Studio in Embedded when used in ASP.NET Core project.

ayende (approved):

PR #7544, opened 2018-09-05
RavenDB-11890 Let's skip getting process memory usage if reduce phase…

arekpalinski:
… took <= 1 sec so it won't affect very small reduce batches


ppekrol:
cc @ayende @maximburyak

PR #7548, opened 2018-09-05
RavenDB-11888 Replace all the certificate instances in the server whe…

iftahbe:
…n replacing the cluster certificate

### DIFF
--- a/src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtl.cs
+++ b/src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtl.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Threading;
 using Raven.Client.Documents.Commands.Batches;
 using Raven.Client.Documents.Conventions;
 using Raven.Client.Documents.Operations.Counters;
@@ -19,7 +20,7 @@ namespace Raven.Server.Documents.ETL.Providers.Raven
     {
         public const string RavenEtlTag = "Raven ETL";
 
-        private readonly RequestExecutor _requestExecutor;
+        private RequestExecutor _requestExecutor;
         private string _recentUrl;
         public string Url => _recentUrl;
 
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Smuggler/Migration/Migrator.cs
```diff
@@ -242,7 +242,6 @@ public long StartMigratingSingleDatabase(DatabaseMigrationSettings databaseMigra
                     result.AddInfo(message);
 
                     using (cancelToken)
-                    using (_httpClient)
```
ppekrol: Can you explain?

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -1989,6 +1989,20 @@ private async Task<(long Index, object Result)> SendToNodeAsync(TransactionOpera
                 _clusterRequestExecutor?.Dispose();
                 _clusterRequestExecutor = ClusterRequestExecutor.CreateForSingleNode(leaderUrl, Server.Certificate.Certificate);
                 _clusterRequestExecutor.DefaultTimeout = Engine.OperationTimeout;
+
+                Server.ServerCertificateChanged += (sender, args) =>
+                {
+                    // When the server certificate changes, we need to start using the new one.
+                    // Since the request executor has the old certificate, we will re-create it and it will pick up the new certificate.
+                    var newClusterRequestExecutor = ClusterRequestExecutor.CreateForSingleNode(leaderUrl, Server.Certificate.Certificate);
+                    newClusterRequestExecutor.DefaultTimeout = Engine.OperationTimeout;
```
ppekrol: wrap those two lines into a method and use this in both places. You will avoid a situation where someone will change the initial creation of the executor and forget about the event.

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtl.cs
```diff
@@ -32,7 +33,19 @@ public RavenEtl(Transformation transformation, RavenEtlConfiguration configurati
...
                 _requestExecutor = RequestExecutor.Create(configuration.Connection.TopologyDiscoveryUrls, configuration.Connection.Database, serverStore.Server.Certificate.Certificate, DocumentConventions.Default);
             
-            _script = new RavenEtlDocumentTransformer.ScriptInput(transformation);            
+            _script = new RavenEtlDocumentTransformer.ScriptInput(transformation);         
+			
+			serverStore.Server.ServerCertificateChanged += (sender, args) =>
+            {
+                // When the server certificate changes, we need to start using the new one.
+                // Since the request executor has the old certificate, we will re-create it and it will pick up the new certificate.
+                var newRequestExecutor = RequestExecutor.Create(configuration.Connection.TopologyDiscoveryUrls, configuration.Connection.Database, serverStore.Server.Certificate.Certificate, DocumentConventions.Default);
```
ppekrol: wrap this into a method and use this in both places. You will avoid a situation where someone will change the initial creation of the executor and forget about the event.

src/Raven.Server/Smuggler/Migration/Migrator.cs
iftahbe: https://aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/

src/Raven.Server/ServerWide/ServerStore.cs
iftahbe: Yes ok

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtl.cs
iftahbe: OK

src/Raven.Server/Smuggler/Migration/Migrator.cs
ppekrol: I do not think this is the case, since this client is ONLY used here inside Migrator. It is not shared with other operations.

PR #7550, opened 2018-09-05
RavenDB-11819 Counters metadata Issues

haludi:
http://issues.hibernatingrhinos.com/issue/RavenDB-11819

### DIFF
--- a/src/Raven.Server/Documents/CountersStorage.cs
+++ b/src/Raven.Server/Documents/CountersStorage.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Text;
 using Raven.Client;
 using Raven.Server.Documents.Replication;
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -660,98 +661,118 @@ public long GetNumberOfCounterEntries(DocumentsOperationContext context)
...
-                            break;
-                        default:
-                            throw new ArgumentOutOfRangeException($"Unknown value {operation.Type}");
-                    }
-                }
+                initCounterListSize += metadataCounters.Length;
             }
-            else
+
+            //Todo Maybe should not read from the table if there is metadata on document
```
haludi: Maybe should not read from the table if there is metadata on the document?
The problem has been rise because there was no metadata in the document.
But if there is, we can avoid reading from the counter table and rely on the metadata on the document like before

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -660,101 +661,130 @@ public long GetNumberOfCounterEntries(DocumentsOperationContext context)
...
+            BlittableJsonReaderArray metadataCounters = null;
+            var initCounterListSize = countersOperations.Count;
+            if (doc.TryGet(Constants.Documents.Metadata.Key, out BlittableJsonReaderObject metadata) &&
+                metadata.TryGet(Constants.Documents.Metadata.Counters, out metadataCounters))
             {
-                foreach (var operation in countersOperations)
+                initCounterListSize += metadataCounters.Length;
+            }
+
+            //Todo Maybe should not read from the table if there is metadata on document
```
ayende: We can't assume that the metadata value is correct, so we have to read from the table anyway

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -660,101 +661,130 @@ public long GetNumberOfCounterEntries(DocumentsOperationContext context)
...
-                            }
-                            break;
-                        case CounterOperationType.None:
-                        case CounterOperationType.Get:
-                            break;
-                        default:
-                            throw new ArgumentOutOfRangeException($"Unknown value {operation.Type}");
-                    }
+            //Todo To consider if can count on that this already in order
+            counters?.Sort(StringComparer.OrdinalIgnoreCase);
```
ayende: When reading from the table, yes, you can rely on it being sorted based on the index you are scanning

src/Raven.Server/Documents/CountersStorage.cs
haludi: https://github.com/ravendb/ravendb/pull/7550#discussion_r215441220

ppekrol (approved):

ayende (approved):

PR #7560, opened 2018-09-07
RavenDB-10516 Studio omits hostname in Traffic Watch for lazy queries

ppekrol:


src/Raven.Server/Documents/Handlers/MultiGetHandler.cs
```diff
@@ -87,6 +89,8 @@ public async Task PostMultiGet()
                         httpContext.Response.StatusCode = 0;
                         httpContext.Request.Headers.Clear();
                         httpContext.Response.Headers.Clear();
+                        httpContext.Request.Host = host;
+                        httpContext.Request.Scheme = scheme;
```
ayende, line 93: Need to also use `Port` here, if not the default one

ppekrol: The 'Host' in HttpContext.Request already contains that.

PR #7563, opened 2018-09-08
[v4.1]IssueWithNoneNeededSQLReplicationDeletes: We want to prevent from del…

maximburyak:
…eting from tables that we didn't change

### DIFF
--- a/src/Raven.Server/Documents/ETL/Providers/SQL/SqlDocumentTransformer.cs
+++ b/src/Raven.Server/Documents/ETL/Providers/SQL/SqlDocumentTransformer.cs
@@ -155,11 +155,16 @@ namespace Raven.Server.Documents.ETL.Providers.SQL
 
                 DocumentScript.Run(Context, Context, "execute", new object[] { Current.Document }).Dispose();
             }
-
+            
+         
             // ReSharper disable once ForCanBeConvertedToForeach
             for (int i = 0; i < _config.SqlTables.Count; i++)
             {
                 // delete all the rows that might already exist there
+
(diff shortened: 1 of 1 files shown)


maximburyak:
@arekpalinski please review

src/Raven.Server/Documents/ETL/Providers/SQL/SqlDocumentTransformer.cs
```diff
@@ -155,12 +155,17 @@ public override void Transform(ToSqlItem item)
...
-
+            
+         
             // ReSharper disable once ForCanBeConvertedToForeach
             for (int i = 0; i < _config.SqlTables.Count; i++)
             {
                 // delete all the rows that might already exist there
 
+                // if we haven't processed this table, we should not try deleting from it
+                if (_tables.ContainsKey(_config.SqlTables[i].TableName) == false)
```
arekpalinski, line 166: At the time of processing the deletion we don't know where a document was inserted. It could be inserted into one table, it could be split and inserted into multiple tables (e.g. order and order lines) etc. So we delete if from all tables.

Note that `_tables` is modified only here and during `loadTo<Table>()` call. So if you process only deletion this is collection is empty and we always filter it out - that's why tests failed for this PR.

The issue is that we shouldn't iterate over `_config.SqlTables` (that are global for all scripts) but _only_ over tables that are specific per script.

arekpalinski:
I've created http://issues.hibernatingrhinos.com/issue/RavenDB-11910 for this

PR #7572, opened 2018-09-10
RavenDB-11914 DeleteByQueryOperation<TEntity> throws when DocumentStore does not specify default database

ppekrol:


src/Raven.Client/Documents/Operations/DeleteByQueryOperation.cs
```diff
@@ -40,7 +40,7 @@ public override RavenCommand<OperationIdResult> GetCommand(IDocumentStore store,
         {
             if (_queryToDelete == DummyQuery)
             {
-                using (var session = store.OpenSession())
+                using (var session = store.OpenSession(store.Database ?? "DummyDatabase"))
```
maximburyak, line 43: I think we better just throw here a more significant exception

ppekrol: We do not want to throw here

ppekrol: Session is only used to generate IndexQuery

PR #7577, opened 2018-09-11
[4.1] - RavenDB-11915 - added counters metrics

grisha-kotler:
http://issues.hibernatingrhinos.com/issue/RavenDB-11915

### DIFF
--- a/src/Raven.Server/Documents/Handlers/StatsHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/StatsHandler.cs
@@ -173,7 +173,7 @@ namespace Raven.Server.Documents.Handlers
                     },
                     [nameof(Database.Metrics.Attachments)] = new DynamicJsonValue
                     {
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Documents/Handlers/StatsHandler.cs
```diff
@@ -145,7 +145,11 @@ public Task PutsMetrics()
                     },
                     [nameof(Database.Metrics.Attachments)] = new DynamicJsonValue
                     {
-                        [nameof(Database.Metrics.Attachments.PutsPerSec)] = Database.Metrics.Docs.PutsPerSec.CreateMeterData(true, empty)
+                        [nameof(Database.Metrics.Attachments.PutsPerSec)] = Database.Metrics.Attachments.PutsPerSec.CreateMeterData(true, empty)
```
ppekrol, line 148: Same fix needs to be applied to `/databases/*/metrics/bytes`

grisha-kotler: done

PR #7587, opened 2018-09-12
RavenDB-10554: Add requests content to custom info in traffic watch

garayx:
WIP do not merge

### DIFF
--- a/src/Raven.Client/Documents/Changes/ChangeNotification.cs
+++ b/src/Raven.Client/Documents/Changes/ChangeNotification.cs
@@ -246,7 +246,6 @@ namespace Raven.Client.Documents.Changes
         public object QueryTimings { get; set; }
     }
 
-    [Flags]
     public enum TrafficWatchChangeType
     {
         None,
--- a/src/Raven.Client/Documents/Changes/ChangeNotification.cs
+++ b/src/Raven.Client/Documents/Changes/ChangeNotification.cs
@@ -241,7 +241,23 @@ namespace Raven.Client.Documents.Changes
         public string AbsoluteUri { get; set; }
         public string DatabaseName { get; set; }
         public string CustomInfo { get; set; }
+        public TrafficWatchChangeType Type { get; set; }
         public int InnerRequestsCount { get; set; }
         public object QueryTimings { get; set; }
+    }
+
+    public enum TrafficWatchChangeType
+    {
+        None,
+        Queries,
+        Operations,
+        MultiGet,
+        Patch,
+        Index,
+        Counters,
+        Hilo,
+        Subscriptions,
+        Streams,
+        Documents
     }
 }
--- a/src/Raven.Client/Documents/Changes/ChangeNotification.cs
+++ b/src/Raven.Client/Documents/Changes/ChangeNotification.cs
@@ -252,7 +252,7 @@ namespace Raven.Client.Documents.Changes
         Queries,
         Operations,
         MultiGet,
-        Patch,
+        BulkDocs,
         Index,
         Counters,
         Hilo,
--- a/src/Raven.Client/Documents/Changes/ChangeNotification.cs
+++ b/src/Raven.Client/Documents/Changes/ChangeNotification.cs
@@ -246,6 +246,7 @@ namespace Raven.Client.Documents.Changes
         public object QueryTimings { get; set; }
     }
 
+    [Flags]
     public enum TrafficWatchChangeType
     {
         None,
--- a/src/Raven.Server/Documents/DatabaseRequestHandler.cs
+++ b/src/Raven.Server/Documents/DatabaseRequestHandler.cs
@@ -6,6 +6,7 @@ using Raven.Server.ServerWide;
 using Raven.Server.ServerWide.Context;
 using Raven.Server.Web;
 using Raven.Client;
+using Raven.Client.Documents.Changes;
 using Raven.Client.Exceptions;
 using Raven.Client.ServerWide;
 using Raven.Server.NotificationCenter.Notifications.Details;
@@ -93,12 +94,15 @@ namespace Raven.Server.Documents
             }
         }
         /// <summary>
-        /// puts given string in TrafficWatch property of HttpContext.Items
+        /// puts the given string in TrafficWatch property of HttpContext.Items
+        /// puts the given type in TrafficWatchChangeType property of HttpContext.Items
         /// </summary>
         /// <param name="str"></param>
-        public void AddStringToHttpContext(string str)
+        /// <param name="type"></param>
+        public void AddStringToHttpContext(string str, TrafficWatchChangeType type)
         {
             HttpContext.Items["TrafficWatch"] = str;
+            HttpContext.Items["TrafficWatchChangeType"] = type;
         }
 
         protected OperationCancelToken CreateTimeLimitedOperationToken()
--- a/src/Raven.Server/Documents/DatabaseRequestHandler.cs
+++ b/src/Raven.Server/Documents/DatabaseRequestHandler.cs
@@ -101,8 +101,7 @@ namespace Raven.Server.Documents
         /// <param name="type"></param>
         public void AddStringToHttpContext(string str, TrafficWatchChangeType type)
         {
-            HttpContext.Items["TrafficWatch"] = str;
-            HttpContext.Items["TrafficWatchChangeType"] = type;
+            HttpContext.Items["TrafficWatch"] = (str, type);
         }
 
         protected OperationCancelToken CreateTimeLimitedOperationToken()
--- a/src/Raven.Server/Documents/DatabaseRequestHandler.cs
+++ b/src/Raven.Server/Documents/DatabaseRequestHandler.cs
@@ -92,6 +92,14 @@ namespace Raven.Server.Documents
                 }
             }
         }
+        /// <summary>
+        /// puts given string in TrafficWatch property of HttpContext.Items
+        /// </summary>
+        /// <param name="str"></param>
+        public void AddStringToHttpContext(string str)
+        {
+            HttpContext.Items["TrafficWatch"] = str;
+        }
 
         protected OperationCancelToken CreateTimeLimitedOperationToken()
         {
--- a/src/Raven.Server/Documents/Handlers/BatchHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/BatchHandler.cs
@@ -20,6 +20,7 @@ using Sparrow;
 using Sparrow.Json;
 using Sparrow.Json.Parsing;
 using System.Runtime.ExceptionServices;
+using Raven.Client.Documents.Changes;
 using Raven.Client.Documents.Operations;
 using Raven.Client.Json;
 using Raven.Server.Json;
@@ -66,25 +67,7 @@ namespace Raven.Server.Documents.Handlers
                     ThrowNotSupportedType(contentType);
                 if (TrafficWatchManager.HasRegisteredClients)
                 {
-                    var sb = new StringBuilder();
-                    for (var i = 0; i < command.ParsedCommands.Count; i++)
-                    {
-                        // watching only patch commands
-                        if (command.ParsedCommands.Array[i].Patch != null)
-                        {
-                            // reconstruct in json format
-                            sb.Append("{\"").Append(command.ParsedCommands.Array[i].Type).Append("\":{");
-                            sb.Append("\"Script\":\"").Append(command.ParsedCommands.Array[i].Patch.Script).Append("\"");
-                            if (command.ParsedCommands.Array[i].PatchArgs != null)
-                            {
-                                sb.Append(",\"Values\":").Append(command.ParsedCommands.Array[i].PatchArgs);
-                            }
-                            sb.Append("}}").AppendLine();
-                        }
-                    }
-                    // add sb to httpContext only if we had a patch type item in array
-                    if (sb.Length > 0)
-                        AddStringToHttpContext(sb.ToString());
+                    BatchTrafficWatch(command.ParsedCommands);
                 }
 
                 var waitForIndexesTimeout = GetTimeSpanQueryString("waitForIndexesTimeout", required: false);
@@ -157,6 +140,27 @@ namespace Raven.Server.Documents.Handlers
             public Task IndexTask;
             public DynamicJsonArray Array;
         }
+
+
+        private void BatchTrafficWatch(ArraySegment<BatchRequestParser.CommandData> parsedCommands)
+        {
+            var sb = new StringBuilder();
+            for (var i = parsedCommands.Offset; i < (parsedCommands.Offset + parsedCommands.Count); i++)
+            {
+                // watching only patch commands
+                if (parsedCommands.Array[i].Patch != null)
+                {
+                    sb.Append("Type: ").Append(parsedCommands.Array[i].Type).AppendLine();
+                    sb.Append("Id: ").Append(parsedCommands.Array[i].Id).AppendLine();
+                    sb.Append("Script: ").Append(parsedCommands.Array[i].Patch.Script).AppendLine();
+                    sb.Append("Values: ").Append(parsedCommands.Array[i].PatchArgs).AppendLine();
+                }
+            }
+            // add sb to httpContext only if we had a patch type item in array
+            if (sb.Length > 0)
+                AddStringToHttpContext(sb.ToString(), TrafficWatchChangeType.Patch);
+        }
+
 
         private async Task HandleClusterTransaction(DocumentsOperationContext context, MergedBatchCommand command, ClusterTransactionCommand.ClusterTransactionOptions options)
         {
--- a/src/Raven.Server/Documents/Handlers/BatchHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/BatchHandler.cs
@@ -141,24 +141,27 @@ namespace Raven.Server.Documents.Handlers
             public DynamicJsonArray Array;
         }
 
-
         private void BatchTrafficWatch(ArraySegment<BatchRequestParser.CommandData> parsedCommands)
         {
             var sb = new StringBuilder();
             for (var i = parsedCommands.Offset; i < (parsedCommands.Offset + parsedCommands.Count); i++)
             {
-                // watching only patch commands
-                if (parsedCommands.Array[i].Patch != null)
+                // log script and args if type is patch
+                if (parsedCommands.Array[i].Type == CommandType.PATCH)
                 {
-                    sb.Append("Type: ").Append(parsedCommands.Array[i].Type).AppendLine();
-                    sb.Append("Id: ").Append(parsedCommands.Array[i].Id).AppendLine();
-                    sb.Append("Script: ").Append(parsedCommands.Array[i].Patch.Script).AppendLine();
-                    sb.Append("Values: ").Append(parsedCommands.Array[i].PatchArgs).AppendLine();
+                    sb.Append(parsedCommands.Array[i].Type).Append("    ")
+                        .Append(parsedCommands.Array[i].Id).Append("    ")
+                        .Append(parsedCommands.Array[i].Patch.Script).Append("    ")
+                        .Append(parsedCommands.Array[i].PatchArgs).AppendLine();
+                }
+                else 
+                {
+                    sb.Append(parsedCommands.Array[i].Type).Append("    ")
+                        .Append(parsedCommands.Array[i].Id).AppendLine();
                 }
             }
-            // add sb to httpContext only if we had a patch type item in array
-            if (sb.Length > 0)
-                AddStringToHttpContext(sb.ToString(), TrafficWatchChangeType.Patch);
+            // add sb to httpContext
+            AddStringToHttpContext(sb.ToString(), TrafficWatchChangeType.BulkDocs);
         }
 
 
--- a/src/Raven.Server/Documents/Handlers/BatchHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/BatchHandler.cs
@@ -64,6 +64,28 @@ namespace Raven.Server.Documents.Handlers
                 }
                 else
                     ThrowNotSupportedType(contentType);
+                if (TrafficWatchManager.HasRegisteredClients)
+                {
+                    var sb = new StringBuilder();
+                    for (var i = 0; i < command.ParsedCommands.Count; i++)
+                    {
+                        // watching only patch commands
+                        if (command.ParsedCommands.Array[i].Patch != null)
+                        {
+                            // reconstruct in json format
+                            sb.Append("{\"").Append(command.ParsedCommands.Array[i].Type).Append("\":{");
+                            sb.Append("\"Script\":\"").Append(command.ParsedCommands.Array[i].Patch.Script).Append("\"");
+                            if (command.ParsedCommands.Array[i].PatchArgs != null)
+                            {
+                                sb.Append(",\"Values\":").Append(command.ParsedCommands.Array[i].PatchArgs);
+                            }
+                            sb.Append("}}").AppendLine();
+                        }
+                    }
+                    // add sb to httpContext only if we had a patch type item in array
+                    if (sb.Length > 0)
+                        AddStringToHttpContext(sb.ToString());
+                }
 
                 var waitForIndexesTimeout = GetTimeSpanQueryString("waitForIndexesTimeout", required: false);
                 var waitForIndexThrow = GetBoolValueQueryString("waitForIndexThrow", required: false) ?? true;
--- a/src/Raven.Server/Documents/Handlers/BatchHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/BatchHandler.cs
@@ -34,6 +34,7 @@ using Raven.Server.Config.Categories;
 using Raven.Server.Exceptions;
 using Raven.Server.Rachis;
 using Raven.Server.ServerWide.Commands;
+using Raven.Server.TrafficWatch;
 using Raven.Server.Utils;
 using Voron;
 using Raven.Server.Documents.Indexes.Persistence.Lucene.Documents.Fields;
@@ -125,6 +126,7 @@ namespace Raven.Server.Documents.Handlers
                         [nameof(BatchCommandResult.Results)] = command.Reply
                     });
                 }
+
             }
         }
 
--- a/src/Raven.Server/Documents/Handlers/CountersHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/CountersHandler.cs
@@ -7,6 +7,7 @@
 using System;
 using System.Collections.Generic;
 using System.Net;
+using System.Text;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Primitives;
 using Raven.Client.Documents.Operations.Counters;
@@ -17,6 +18,7 @@ using Raven.Server.Config.Categories;
 using Raven.Server.Exceptions;
 using Raven.Server.Routing;
 using Raven.Server.ServerWide.Context;
+using Raven.Server.TrafficWatch;
 using Sparrow.Json;
 
 namespace Raven.Server.Documents.Handlers
@@ -279,6 +281,13 @@ namespace Raven.Server.Documents.Handlers
                 var countersBlittable = await context.ReadForMemoryAsync(RequestBodyStream(), "counters");
 
                 var counterBatch = JsonDeserializationClient.CounterBatch(countersBlittable);
+
+                if (TrafficWatchManager.HasRegisteredClients)
+                {
+                    var sb = new StringBuilder();
+                    sb.Append(/*"Counter:\n"+ */countersBlittable);
+                    HttpContext.Items["TrafficWatch"] = sb.ToString();
+                }
 
                 var cmd = new ExecuteCounterBatchCommand(Database, counterBatch);
 
--- a/src/Raven.Server/Documents/Handlers/CountersHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/CountersHandler.cs
@@ -283,12 +283,7 @@ namespace Raven.Server.Documents.Handlers
                 var counterBatch = JsonDeserializationClient.CounterBatch(countersBlittable);
 
                 if (TrafficWatchManager.HasRegisteredClients)
-                {
-                    var sb = new StringBuilder();
-                    sb.Append(/*"Counter:\n"+ */countersBlittable);
-                    HttpContext.Items["TrafficWatch"] = sb.ToString();
-                }
-
+                    AddStringToHttpContext(countersBlittable.ToString());
                 var cmd = new ExecuteCounterBatchCommand(Database, counterBatch);
 
                 if (cmd.HasWrites)
--- a/src/Raven.Server/Documents/Handlers/CountersHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/CountersHandler.cs
@@ -10,6 +10,7 @@ using System.Net;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Primitives;
+using Raven.Client.Documents.Changes;
 using Raven.Client.Documents.Operations.Counters;
 using Raven.Client.Exceptions.Documents;
 using Raven.Client.Exceptions.Documents.Counters;
@@ -283,7 +284,7 @@ namespace Raven.Server.Documents.Handlers
                 var counterBatch = JsonDeserializationClient.CounterBatch(countersBlittable);
 
                 if (TrafficWatchManager.HasRegisteredClients)
-                    AddStringToHttpContext(countersBlittable.ToString());
+                    AddStringToHttpContext(countersBlittable.ToString(), TrafficWatchChangeType.Counters);
                 var cmd = new ExecuteCounterBatchCommand(Database, counterBatch);
 
                 if (cmd.HasWrites)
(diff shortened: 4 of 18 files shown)


src/Raven.Server/RavenServerStartup.cs
```diff
@@ -157,23 +159,47 @@ private async Task RequestHandler(HttpContext context)
                 context.Response.StatusCode = (int)HttpStatusCode.OK;
                 context.Response.Headers["Content-Type"] = "application/json; charset=utf-8";
                 context.Response.Headers[Constants.Headers.ServerVersion] = RavenVersionAttribute.Instance.AssemblyVersion;
+                // Check context url AND TW clients
+                var isQueryBool = (IsQuery(context) && TrafficWatchManager.HasRegisteredClients);
```
ppekrol: Not sure if this is possible, but maybe it would be better to introduce some kind dedicated context that would wrap the original HttpContext and this new context (if traffic watch is enabled) would 'record' the original body for the traffic watch. Thoughts?

ayende: It is better to do this at the `_router` level. Where we'll have an indication on the action itself if we want to record its contents.

ayende: Actually, would be better if we could just add the data to log directly to the context directly.

src/Raven.Server/RavenServerStartup.cs
```diff
@@ -218,11 +244,69 @@ private async Task RequestHandler(HttpContext context)
...
-        private void LogTrafficWatch(HttpContext context, long elapsedMilliseconds, string database)
+        private void LogTrafficWatch(HttpContext context, long elapsedMilliseconds, string database, bool queryOperation)
         {
             var requestId = Interlocked.Increment(ref _requestId);
+            
+            string customInfoString;
+            if (context.Request.Method == "GET" && queryOperation)                          // GET method and URL has query <=> query from Raven.Studio
+            {
+                customInfoString = context.Request.QueryString.Value;
+                customInfoString = Uri.UnescapeDataString(customInfoString.Substring(7, context.Request.QueryString.Value.Length - 65));
```
ayende: What is this `7` here?

ayende: What is this 65?

src/Raven.Server/RavenServerStartup.cs
```diff
@@ -218,11 +244,69 @@ private async Task RequestHandler(HttpContext context)
...
+            }
+            else if (context.Request.Method == "POST" && queryOperation)                  // POST method AND URL has query <=> Query from Raven.Client
```
ayende: We want to do this from the handler, not here. 
The handler gets to decide what goes on in here.

ayende (changes requested):
Instead of doing this externally, define a property on the `Request.Items` that you can use to pass data from the controller.

src/Raven.Server/RavenServerStartup.cs
garayx: done

src/Raven.Server/RavenServerStartup.cs
garayx: done

src/Raven.Server/RavenServerStartup.cs
```diff
@@ -218,11 +287,31 @@ private async Task RequestHandler(HttpContext context)
...
         }
+        /// <summary>
+        /// IsQuery method checks if watched URL contains "?query" OR "multi_get" OR "index-fields"
+        /// </summary>
+        /// <param name="context"></param>
+        /// <returns>boolean</returns>
+        private static bool IsQuery(HttpContext context)
+        {
+            return (
+                    context.Request.QueryString.Value.StartsWith("?query")
```
ppekrol: We need to watch everything. Not only those three.

src/Raven.Server/RavenServerStartup.cs
```diff
@@ -157,23 +159,90 @@ private async Task RequestHandler(HttpContext context)
                 context.Response.StatusCode = (int)HttpStatusCode.OK;
                 context.Response.Headers["Content-Type"] = "application/json; charset=utf-8";
                 context.Response.Headers[Constants.Headers.ServerVersion] = RavenVersionAttribute.Instance.AssemblyVersion;
+                // Check context url AND TW clients
+                bool isQueryBool = (TrafficWatchManager.HasRegisteredClients && IsQuery(context));
+
+                if (isQueryBool)
+                {
```
ppekrol: try to wrap this inside dedicated httpcontext that will handle all of this additional stuff

src/Raven.Server/RavenServerStartup.cs
```diff
@@ -157,23 +159,90 @@ private async Task RequestHandler(HttpContext context)
...
+                    // init new MS to save context.request.body
+                    MemoryStream memoryStreamForRequest = new MemoryStream();
```
ppekrol: need to register this for disposal
httpContext.Response.RegisterForDisposal

src/Raven.Server/RavenServerStartup.cs
```diff
@@ -157,23 +159,90 @@ private async Task RequestHandler(HttpContext context)
...
+                            // if there is a ContentType
+                            if (context.Request.ContentType == "application/json; charset=UTF-8")
+                            {
+                                // get the memory stream from context
+                                MemoryStream memStream = context.Items.TryGetValue("requestMemoryStream", out val) ? (MemoryStream)val : null;
+                                if (memStream != null)
+                                {
+                                    // read the memory stream
+                                    memStream.Position = 0;
+                                    var reader = new StreamReader(memStream);
```
ppekrol: needs to be disposed

src/Raven.Server/RavenServerStartup.cs
```diff
@@ -157,23 +159,90 @@ private async Task RequestHandler(HttpContext context)
...
+                                    customInfoString = reader.ReadToEnd();
+                                }
+                            }
+                            else if (context.Request.ContentType == null)
+                            {
+                                MemoryStream memStream = context.Items.TryGetValue("requestMemoryStream", out val) ? (MemoryStream)val : null;
+                                if (memStream != null)
+                                {
+                                    memStream.Position = 0;
+                                    var reader = new StreamReader(new GZipStream(memStream, CompressionMode.Decompress, true));
```
ppekrol: needs to be disposed

src/Raven.Server/RavenServerStartup.cs
```diff
@@ -157,23 +159,90 @@ private async Task RequestHandler(HttpContext context)
...
+                        customInfoString = context.Request.Query.TryGetValue("query", out value) ? value.ToString() : "N/A";
+                    }
+                    else if (context.Request.Method == "POST")
+                    {
+                        try
+                        {
+                            // object for the memory stream stored at the context
+                            object val;
+                            // if there is a ContentType
+                            if (context.Request.ContentType == "application/json; charset=UTF-8")
```
ppekrol: why we have those two cases?

I think you should use: RequestBodyStream() method in both cases?

src/Raven.Server/RavenServerStartup.cs
garayx: all urls are currently watched, those three actions currently are returning content to the customInfo field

ppekrol: Yes, but we should not limit ourselfs. If there is a body and we have a traffic-watch turned on, then we should track the body and show it in the UI

ayende: @ppekrol I'm not sure that we want to just get the body of all requests in this manner.
In particular, `bulk_docs` and `BulkInsert` are two scenarios that we don't want to capture in this way.
Too many details that would overwhelm the user.

ayende: Body can be very large, and we don't usually want to show the raw data.
For queries / patch is pretty much the only case.
bulk_docs isn't, we don't want the full docs there.

src/Raven.Server/RavenServerStartup.cs
```diff
@@ -157,23 +159,90 @@ private async Task RequestHandler(HttpContext context)
...
                 context.Response.Headers[Constants.Headers.ServerVersion] = RavenVersionAttribute.Instance.AssemblyVersion;
+                // Check context url AND TW clients
+                bool isQueryBool = (TrafficWatchManager.HasRegisteredClients && IsQuery(context));
+
+                if (isQueryBool)
+                {
+                    // init new MS to save context.request.body
+                    MemoryStream memoryStreamForRequest = new MemoryStream();
+                    // copy stream
+                    await context.Request.Body.CopyToAsync(memoryStreamForRequest);
```
ayende: I don't like this approach. Better to pass a flag to the actions and have them construct their own custom data, instead.

ayende (changes requested):

src/Raven.Server/RavenServerStartup.cs
ppekrol: For bulk insert I agree, was thinking that the responsibility should be inverted, we could list the endpoints that we do not want to process in this way.

Anyway, I think that allowing endpoints to inject the customData is a better approach.

src/Raven.Server/Documents/Handlers/CountersHandler.cs
```diff
@@ -261,6 +263,13 @@ public async Task Batch()
 
                 var counterBatch = JsonDeserializationClient.CounterBatch(countersBlittable);
 
+                if (TrafficWatchManager.HasRegisteredClients)
+                {
+                    var sb = new StringBuilder();
+                    sb.Append(/*"Counter:\n"+ */countersBlittable);
```
ayende: What is this ? `/*"Counter:\n"+ */`

ayende: Can just call `countersBlittable.ToString()` here, no?

src/Raven.Server/Documents/Handlers/CountersHandler.cs
```diff
@@ -261,6 +263,13 @@ public async Task Batch()
...
+                    HttpContext.Items["TrafficWatch"] = sb.ToString();
```
ayende: Please create a method that would accept a string only and set this internally. Avoid having the string constant repeated

src/Raven.Server/Documents/Handlers/HiLoHandler.cs
```diff
@@ -80,6 +82,12 @@ public async Task GetNextHiLo()
                         ["ServerTag"] = ServerStore.NodeTag,
                         ["LastRangeAt"] = DateTime.UtcNow.ToString(DefaultFormat.DateTimeOffsetFormatsToWrite)
                     });
+                    if (TrafficWatchManager.HasRegisteredClients)
+                    {
+                        var sb = new StringBuilder();
+                        sb.Append(/*"HiLo:\n"+ */writer);
+                        HttpContext.Items["TrafficWatch"] = sb.ToString();
```
ayende: Same

src/Raven.Server/Documents/Handlers/MultiGetHandler.cs
```diff
@@ -184,16 +185,17 @@ public async Task PostMultiGet()
                             }
                         }
                         writer.WriteEndObject();
-
                         writer.WriteEndObject();
+                        // have to be at the end of the for
+                        if (TrafficWatchManager.HasRegisteredClients)
+                        {
+                            HttpContext.Items["TrafficWatch"] = (string)HttpContext.Items["TrafficWatch"] + content.ToString() + "\n";
```
ayende: Use `StringBuilder` here and set it once

src/Raven.Server/Documents/Handlers/QueriesHandler.cs
```diff
@@ -333,6 +339,21 @@ public Task Patch()
...
+        /// </summary>
+        /// <param name="indexQuery"></param>
+        private void TrafficWatchQuery(IndexQueryServerSide indexQuery)
+        {
+            var sb = new StringBuilder();
+            // append stringBuilder with the query
+            sb.Append(/*"Query:\n"+ */indexQuery.Query);
+            // if query got parameters append with parameters
+            if (indexQuery.QueryParameters != null && indexQuery.QueryParameters.Count > 0)
+                sb.Append("\n" + indexQuery.QueryParameters);
```
ayende: This will _not_ generate good output.

garayx:
Please write additional endpoints, which data I have to log.

ayende (changes requested):

src/Raven.Server/Documents/Handlers/QueriesHandler.cs
garayx: This is the output generated

image

src/Raven.Server/Documents/Handlers/MultiGetHandler.cs
garayx: done

src/Raven.Server/Documents/Handlers/QueriesHandler.cs
ayende: How, `QueryParameters` should just return the type name

src/Raven.Server/Documents/Handlers/CountersHandler.cs
garayx: I can create a method which accepts a string as argument, in each class that we are taking the customInfo from.
Or I can create a method inside the trafficWatchManager and pass the httpContext with string to it.

ayende: All the handlers inherit from a common base, put that there.

src/Raven.Server/Documents/Handlers/CountersHandler.cs
garayx: I didnt know if raw data is needed, or formatted with a title, like this:
Query:
.....
Parameters:
.....

So i kept it commented.

There are few places that I can use HttpContext.Items["TrafficWatch"] = String / .toString()
directly.

src/Raven.Server/Documents/Handlers/CountersHandler.cs
garayx: done

src/Raven.Server/Documents/Handlers/MultiGetHandler.cs
```diff
@@ -48,7 +49,7 @@ public async Task PostMultiGet()
                     var httpContext = new DefaultHttpContext(features);
                     var host = HttpContext.Request.Host;
                     var scheme = HttpContext.Request.Scheme;
-
+                    var sb = new StringBuilder();
```
ppekrol: Do not initialize SB if `HasRegisteredClients` is false. This will avoid doing unnecessary work.

src/Raven.Server/Documents/Handlers/QueriesHandler.cs
```diff
@@ -333,6 +339,20 @@ public Task Patch()
...
+        /// </summary>
+        /// <param name="indexQuery"></param>
+        private void TrafficWatchQuery(IndexQueryServerSide indexQuery)
+        {
+            var sb = new StringBuilder();
+            // append stringBuilder with the query
+            sb.Append(indexQuery.Query);
+            // if query got parameters append with parameters
+            if (indexQuery.QueryParameters != null && indexQuery.QueryParameters.Count > 0)
+                sb.Append("\n" + indexQuery.QueryParameters);
```
ppekrol: `\n` -> `Environment.NewLine` ?

src/Raven.Server/Documents/Handlers/MultiGetHandler.cs
```diff
@@ -186,12 +187,11 @@ public async Task PostMultiGet()
                         }
                         writer.WriteEndObject();
                         writer.WriteEndObject();
-                        // have to be at the end of the for
                         if (TrafficWatchManager.HasRegisteredClients)
-                        {
-                            HttpContext.Items["TrafficWatch"] = (string)HttpContext.Items["TrafficWatch"] + content.ToString() + "\n";
-                        }
+                            sb.Append(content.ToString() + "\n");
```
ayende: Use `AppendLine` here.

src/Raven.Server/Documents/Handlers/MultiGetHandler.cs
garayx: done

src/Raven.Server/Documents/Handlers/QueriesHandler.cs
garayx: done, I used an AppendLine()

src/Raven.Server/Documents/Handlers/MultiGetHandler.cs
garayx: done

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -56,6 +56,28 @@ public async Task BulkDocs()
                 }
                 else
                     ThrowNotSupportedType(contentType);
+                if (TrafficWatchManager.HasRegisteredClients)
+                {
+                    var sb = new StringBuilder();
+                    for (var i = 0; i < command.ParsedCommands.Count; i++)
+                    {
+                        // watching only patch commands
+                        if (command.ParsedCommands.Array[i].Patch != null)
```
ayende: Probably need to handle `Offset` here, just to be safe

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -56,6 +56,28 @@ public async Task BulkDocs()
                 }
                 else
                     ThrowNotSupportedType(contentType);
+                if (TrafficWatchManager.HasRegisteredClients)
+                {
```
ayende, line 69: Move this to a separate method

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -56,6 +56,28 @@ public async Task BulkDocs()
...
+                    var sb = new StringBuilder();
+                    for (var i = 0; i < command.ParsedCommands.Count; i++)
+                    {
+                        // watching only patch commands
+                        if (command.ParsedCommands.Array[i].Patch != null)
+                        {
+                            // reconstruct in json format
+                            sb.Append("{\"").Append(command.ParsedCommands.Array[i].Type).Append("\":{");
+                            sb.Append("\"Script\":\"").Append(command.ParsedCommands.Array[i].Patch.Script).Append("\"");
```
ayende: Let's NOT use json as the format here. You are not escaping the output.
Just write this in human readable form.

src/Raven.Server/Documents/Handlers/MultiGetHandler.cs
```diff
@@ -188,7 +190,7 @@ public async Task PostMultiGet()
                         writer.WriteEndObject();
                         writer.WriteEndObject();
                         if (TrafficWatchManager.HasRegisteredClients)
```
ayende: Check `sb != null` instead, to avoid concurrency issues.

ayende (changes requested):

src/Raven.Server/Documents/Handlers/BatchHandler.cs
garayx: done

src/Raven.Server/Documents/Handlers/BatchHandler.cs
garayx: done

src/Raven.Server/Documents/Handlers/MultiGetHandler.cs
garayx: done

src/Raven.Server/Documents/DatabaseRequestHandler.cs
```diff
@@ -93,12 +94,15 @@ public override void Init(RequestHandlerContext context)
...
+        /// puts the given string in TrafficWatch property of HttpContext.Items
+        /// puts the given type in TrafficWatchChangeType property of HttpContext.Items
         /// </summary>
         /// <param name="str"></param>
-        public void AddStringToHttpContext(string str)
+        /// <param name="type"></param>
+        public void AddStringToHttpContext(string str, TrafficWatchChangeType type)
         {
             HttpContext.Items["TrafficWatch"] = str;
+            HttpContext.Items["TrafficWatchChangeType"] = type;
```
ayende: Let's have a single property here, and use a tuple, instead.

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -150,6 +133,27 @@ public class ClusterTransactionCompletionResult
...
         }
 
+
+        private void BatchTrafficWatch(ArraySegment<BatchRequestParser.CommandData> parsedCommands)
+        {
+            var sb = new StringBuilder();
+            for (var i = parsedCommands.Offset; i < (parsedCommands.Offset + parsedCommands.Count); i++)
+            {
+                // watching only patch commands
+                if (parsedCommands.Array[i].Patch != null)
```
ayende: Why are you watching only patches?

garayx: I am already iterating through this array so I could log the documents that are added via this function (Task BulkDocs()),

ayende: Yes, the idea is that you would write `PUT users/1`, `DEL users/2`, etc for each of the commands in this case

garayx: I made it in this pattern:

```
PUT    employees/1972-A
DELETE    employees/1963-A
PATCH    employees/15-A    this.FirstName = args.val_14;    {"val_14":"Robert33"}
```

ayende: @garayx This looks fine now, yes.
Just make sure that this aligns nicely in the studio

src/Raven.Studio/wwwroot/App/views/manage/trafficWatch.html
```diff
@@ -41,6 +41,37 @@
     </div>
     <div class="flex-horizontal flex-noshrink">
         <div class="flex-separator"></div>
+    <div class="btn-group dropdown-right">
+        <button class="btn btn-default btn-sm dropdown-toggle" type="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true"
+                title="Choose visible columns">
```
garayx: oops, changed this already.

src/Raven.Studio/typescript/viewmodels/manage/trafficWatch.ts
```diff
@@ -72,13 +90,42 @@ class trafficWatch extends viewModelBase {
...
+        return uri.includes(filterLowered) && this.matchFilteredTypeData(item);
     }
-    
+
+    private matchFilteredTypeData(item: Raven.Client.Documents.Changes.TrafficWatchChange) {
+        return this.filteredTypeData.find(x => x.propertyName === item.Type && x.isChecked());
+    }
+
+    filterOnClick(data: typeData):  void {
+        if (data.isChecked() === false) {
```
ml054: can we reverse logic to avoid if (something === false) { doFirst() } else { doSecond} ?)
Instead we can do: if (something) { doSecond() } else { doFirst}

src/Raven.Studio/typescript/viewmodels/manage/trafficWatch.ts
```diff
@@ -72,13 +90,42 @@ class trafficWatch extends viewModelBase {
...
-        return uri.includes(filterLowered);
+        return uri.includes(filterLowered) && this.matchFilteredTypeData(item);
     }
-    
+
+    private matchFilteredTypeData(item: Raven.Client.Documents.Changes.TrafficWatchChange) {
+        return this.filteredTypeData.find(x => x.propertyName === item.Type && x.isChecked());
+    }
+
+    filterOnClick(data: typeData):  void {
```
ml054: remove double spaces before void

src/Raven.Studio/typescript/viewmodels/manage/trafficWatch.ts
```diff
@@ -111,15 +158,26 @@ class trafficWatch extends viewModelBase {
...
-            
-            const avg = countForAvg ? generalUtils.formatNumberToStringFixed(sum * 1.0 / countForAvg, 2) + " ms" : "n/a"; 
+
+            const avg = countForAvg ? this.formatMs((sum * 1.0 / countForAvg)) : "n/a"; 
             this.stats.avg(avg);
         }
     }
-    
+
+    private formatMs(num: number): string {
```
ml054: I think we already have formatters for this - no need to introduce new one

src/Raven.Studio/wwwroot/App/views/manage/trafficWatch.html
```diff
@@ -41,6 +41,37 @@
...
+                <div class="flex-header">
+                    <div class="checkbox">
+                        <input type="checkbox" id="select_all" class="styled" data-bind="checked: isSelectAllChecked, event: { change: toggleSelectAll }">
+                        <label for="select_all">
+                            Select All
+                        </label>
+                    </div>
+                </div>
+                <div class="flex-grow">
+                    <ul class="column-list scroll" data-bind="foreach: { data: filteredTypeData, as: 'item' }">
```
ml054: we usually don't alias items in foreach loops - please use foreach: filteredTypeData

src/Raven.Studio/wwwroot/App/views/manage/trafficWatch.html
```diff
@@ -41,6 +41,37 @@
...
+                        <li>
+                            <div class="column-name">
+                                <div class="checkbox">
+                                    <input class="styled" type="checkbox" data-bind="checked: isChecked(), attr: { id: 'id' + item.propertyName }, event: { change: $parent.filterOnClick.bind($parent, $data) }">
```
ml054: don't bind on click here - subscribe to observable instead.

src/Raven.Studio/wwwroot/App/views/manage/trafficWatch.html
```diff
@@ -41,6 +41,37 @@
...
+                                    <label data-bind="text: propertyName, attr: { for: 'id' + item.propertyName }"></label>
```
ml054: please use prefix which is more unique

src/Raven.Studio/wwwroot/App/views/manage/trafficWatch.html
```diff
@@ -41,6 +41,37 @@
...
     <div class="flex-horizontal flex-noshrink">
         <div class="flex-separator"></div>
+    <div class="btn-group dropdown-right">
+        <button class="btn btn-default btn-sm dropdown-toggle" type="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true"
+                title="Choose visible columns">
+            <i class="icon-table"></i><span>Display</span>
+            <span class="caret"></span>
+        </button>
+        <ul class="dropdown-menu colselect-dropdown slidein-style" role="menu" data-bind="dropdownPanel: true">
+            <div class="columns-list-container flex-vertical">
```
ml054: it isn't columns list container for sure - you should use markup for standard dropdown ?

ml054 (changes requested):

src/Raven.Client/Documents/Changes/ChangeNotification.cs
```diff
@@ -246,6 +246,7 @@ public class TrafficWatchChange : DatabaseChange
         public object QueryTimings { get; set; }
     }
 
+    [Flags]
```
ayende: This doesn't do what I expect you think it does.
It also not a good idea, because these are all mutually exclusive.

src/Raven.Studio/typescript/viewmodels/manage/trafficWatch.ts
garayx: done

src/Raven.Studio/typescript/viewmodels/manage/trafficWatch.ts
garayx: done

src/Raven.Studio/typescript/viewmodels/manage/trafficWatch.ts
garayx: done

src/Raven.Studio/wwwroot/App/views/manage/trafficWatch.html
garayx: done

arekpalinski:
@ml054 Please review again

src/Raven.Studio/typescript/viewmodels/manage/trafficWatch.ts
```diff
@@ -74,62 +80,31 @@ class trafficWatch extends viewModelBase {
...
-        if (!filter) {
-            return this.matchFilteredTypeData(item);
-        }
-        const filterLowered = filter.toLocaleLowerCase();
-        const uri = item.RequestUri.toLocaleLowerCase();
-        return uri.includes(filterLowered) && this.matchFilteredTypeData(item);
+
+    attached() {
+        super.attached();
+        awesomeMultiselect.build($("#visibleIndexesSelector"), opts => {
```
ml054: I think it isn't indexes selector ;-)

src/Raven.Studio/typescript/viewmodels/manage/trafficWatch.ts
```diff
@@ -163,21 +138,21 @@ class trafficWatch extends viewModelBase {
                 countForAvg++;
                 sum += item.ElapsedMilliseconds;
             }
-            this.stats.min(generalUtils.formatTimeSpan(min, false));
-            this.stats.max(generalUtils.formatTimeSpan(max, false));
+
+            // if number == 0 formatTimeSpan return empty string, have to set it to 0ms manually
+            min === 0 ? this.stats.min("0 ms") : this.stats.min(generalUtils.formatTimeSpan(min, false));
```
ml054: why this logic is here? maybe if format time span gets 0 as value it should return string "0 ms"?

src/Raven.Studio/typescript/viewmodels/manage/trafficWatch.ts
```diff
@@ -163,21 +138,21 @@ class trafficWatch extends viewModelBase {
...
+            max === 0 ? this.stats.max("0 ms") : this.stats.max(generalUtils.formatTimeSpan(max, false));
             this.stats.count(this.filteredData.length.toLocaleString());
 
-            const avg = countForAvg ? generalUtils.formatTimeSpan((sum * 1.0 / countForAvg), false) : "n/a"; 
-            this.stats.avg(avg);
+            const avg = countForAvg ? generalUtils.formatTimeSpan((sum * 1.0 / countForAvg), false) : "n/a";
+            avg === "" ? this.stats.avg("0 ms") : this.stats.avg(avg);
```
ml054: same here

garayx: I could do that, just didn't want to touch anything outdidey scope.

ml054 (approved):

src/Raven.Server/Documents/DatabaseRequestHandler.cs
ayende: Reminder, TBD

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -134,6 +141,30 @@ public class ClusterTransactionCompletionResult
...
+                if (parsedCommands.Array[i].Type == CommandType.PATCH)
+                {
+                    sb.Append(parsedCommands.Array[i].Type).Append("    ")
+                        .Append(parsedCommands.Array[i].Id).Append("    ")
+                        .Append(parsedCommands.Array[i].Patch.Script).Append("    ")
+                        .Append(parsedCommands.Array[i].PatchArgs).AppendLine();
+                }
+                else 
+                {
+                    sb.Append(parsedCommands.Array[i].Type).Append("    ")
```
ayende, line 159: This is great

ayende (approved):

src/Raven.Server/Documents/Handlers/BatchHandler.cs
garayx: done

src/Raven.Server/Documents/Handlers/MultiGetHandler.cs
```diff
@@ -129,6 +129,9 @@ public async Task PostMultiGet()
                                 httpContext.Request.Body = requestBody;
                                 httpContext.Request.Body.Position = 0;
                             }
+                        } else if (method == HttpMethod.Get.Method)
+                        {
+                            content = request.ToString();
```
arekpalinski, line 134: Shouldn't we do it only  `if (TrafficWatchManager.HasRegisteredClients)` ?

ayende: Yes, good point.

garayx: i think it should be if(method == HttpMethod.Get.Method && sb != null)
because sb is initiated only if TrafficWatchManager.HasRegisteredClients

arekpalinski: This will work too. Minor improvement could be to rename `sb` variable to something more meaningful

garayx: done

PR #7590, opened 2018-09-12
RavenDB-11928 Auto assign backup task name when missing during update

ml054:


src/Raven.Server/ServerWide/Commands/PeriodicBackup/UpdatePeriodicBackupCommand.cs
```diff
@@ -26,13 +26,14 @@ public override string UpdateDatabaseRecord(DatabaseRecord record, long etag)
...
                 Configuration.TaskId = etag;
-                if (string.IsNullOrEmpty(Configuration.Name))
-                {
-                    Configuration.Name = record.EnsureUniqueTaskName(Configuration.GetDefaultTaskName());
-                }
-
-                record.EnsureTaskNameIsNotUsed(Configuration.Name);
             }
+            
+            if (string.IsNullOrEmpty(Configuration.Name))
```
ppekrol: @ayende if task had a name (`TaskId != 0`), do we want to regenerate it if on update `null` will be send or throw?

ayende: The question here is, why not use the _old_ name?

ppekrol: We can do that if 'name' is null. @ml054 can you apply that?

ml054: I've applied that same logic we had here as in UpdateExternalReplicationCommand

PR #7591, opened 2018-09-12
RavenDB-11930

grisha-kotler:


src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -271,6 +271,8 @@ private void PutCounterImpl(DocumentsOperationContext context, string documentId
                     }
 
                     _documentDatabase.Metrics.Counters.PutsPerSec.MarkSingleThreaded(1);
+                    // counter key + name + etag + value + change vector + collection
+                    _documentDatabase.Metrics.Counters.BytesPutsPerSec.MarkSingleThreaded(counterKey.Size + nameSlice.Size + 8 + 8 + changeVector.Length + collectionName.Name.Length);
```
ppekrol: Let's extract this to a method. Write comments describing what each argument is.

8? sizeof(long) ?

grisha-kotler: done

PR #7598, opened 2018-09-13
RavenDB-11906

ppekrol:


src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexOperationBase.cs
```diff
@@ -37,39 +35,43 @@ protected IndexOperationBase(Index index, Logger logger)
             _logger = logger;
         }
 
-        protected static RavenPerFieldAnalyzerWrapper CreateAnalyzer(Func<Analyzer> createDefaultAnalyzer, IndexDefinitionBase indexDefinition, bool forQuerying = false)
+        protected static RavenPerFieldAnalyzerWrapper CreateAnalyzer(Func<LowerCaseKeywordAnalyzer> createDefaultAnalyzer, IndexDefinitionBase indexDefinition, bool forQuerying = false)
```
ayende: Why this change?

src/Raven.Server/Documents/Indexes/Static/StaticIndexBase.cs
```diff
@@ -108,37 +108,57 @@ public dynamic LoadDocument(object keyOrEnumerable, string collectionName)
...
 
+        protected IEnumerable<AbstractField> CreateField(string name, object value, CreateFieldOptions options)
+        {
+            // IMPORTANT: Do not delete this method, it is used by the indexes code when using CreateField
+
+            options = options ?? CreateFieldOptions.Default;
+
+            IndexFieldOptions allFields = null;
+            var scope = CurrentIndexingScope.Current;
+            if (scope.IndexDefinition is MapIndexDefinition mapIndexDefinition)
```
ayende, line 119: what about map /reduce indexes?

ayende: JS indexes?

src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexOperationBase.cs
ppekrol: Just to be clear what default analyzer actually is. All places that call `CreateAnalyzer` are creating instance of `LowerCaseKeywordAnalyzer`

src/Raven.Server/Documents/Indexes/Static/StaticIndexBase.cs
ppekrol: MapReduceIndexDefinition inherits from MapIndexDefinition so it is fine.

JS indexes do not support 'CreateField' at all

src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexOperationBase.cs
ayende: But we may want to change that in the future, and I don't see what benefit the lowered type gives us here?

src/Raven.Server/Documents/Indexes/Static/StaticIndexBase.cs
ayende: You can certainly do dynamic fields in JS, see:

```
 @"map('Users', function (u){
                                    return {
                                        Name: u.Name,
                                        _: {$value: u.Name, $name:'AnalyzedName', $options:{index: true, store: true}}
                                    };
                                })",
```

src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexOperationBase.cs
ppekrol: fixed

src/Raven.Server/Documents/Indexes/Static/StaticIndexBase.cs
ppekrol: I will check that tomorrow

ppekrol: Done. Changed syntax a bit to cover more options.

ppekrol: 'storage' supports both:
- true/false
- 'Yes' / 'No'

test/FastTests/Client/Indexing/JavaScriptIndexTests.cs
```diff
@@ -665,7 +665,7 @@ public UsersByNameAndAnalyzedName()
                             @"map('Users', function (u){
                                     return {
                                         Name: u.Name,
-                                        _: {$value: u.Name, $name:'AnalyzedName', $options:{index: true, store: true}}
+                                        _: {$value: u.Name, $name:'AnalyzedName', $options:{indexing: 'Search', storage: true}}
```
ayende, line 668: isn't this a breaking change?
This is experimental feature, though.

ppekrol: This was my thinking. We can support both if needed but I do not think that this is a good idea in this case. Can you make a call here?

ayende: It's fine to make this change, yes

arekpalinski (approved):

PR #7612, opened 2018-09-16
Test Fixes

ayende:


ppekrol:
@ayende I think that this needs to go to v4.0 also?

ayende:
Yes

PR #7617, opened 2018-09-17
RavenDB-11896: rename CalculatedAvailableMemory -> AvailableWithoutTo…

aviviadi:
…talCleanMemory

### DIFF
--- a/src/Raven.Server/Dashboard/MachineResourcesNotificationSender.cs
+++ b/src/Raven.Server/Dashboard/MachineResourcesNotificationSender.cs
@@ -3,6 +3,7 @@ using System.Buffers;
 using System.Diagnostics;
 using System.Threading;
 using System.Threading.Tasks;
+using Raven.Client.Extensions;
 using Raven.Server.Background;
 using Raven.Server.NotificationCenter;
 using Raven.Server.Utils;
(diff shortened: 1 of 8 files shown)


src/Raven.Server/Dashboard/MachineResourcesNotificationSender.cs
```diff
@@ -83,7 +84,7 @@ public static MachineResources GetMachineResources(SmapsReader smapsReader)
                 var memInfo = MemoryInformation.GetMemoryInfo(smapsReader);
                 var isLowMemory = LowMemoryNotification.Instance.IsLowMemory(memInfo);
                 var workingSet = PlatformDetails.RunningOnLinux
-                    ? MemoryInformation.GetRssMemoryUsage(currentProcess.Id) - memInfo.SharedCleanMemory.GetValue(SizeUnit.Bytes)
+                    ? MemoryInformation.GetRssMemoryUsage(currentProcess.Id) - memInfo.AvailableWithoutTotalCleanMemory.GetValue(SizeUnit.Bytes)
```
ppekrol, line 87: @aviviadi I want to double check this change. Can you confirm?

aviviadi: Let us put it this way:
1. "workingSet" in windows and "RSS" in linux are different things
2. Giving a single "workingSet" value which is roughly equal across the different platforms demands calculations which will give us "something" close to the real value (we often assume Raven is the only process running for those calculations)
3. And the big question: DO WE NEED to read and show "Working Set" or "Rss" ?? (my answer is : defiantly no! We don't do with this value nothing and shouldn't, and also it is not a neccesarry debugging value)

ayende: Agree, this is too complex for an informative only value, let's drop it.

PR #7619, opened 2018-09-17
RavenDB-11954: Retry on EINVAL pwrite fail if cifs/nfs mount

aviviadi:


src/Sparrow/Platform/Posix/Syscall.cs
```diff
@@ -275,12 +276,34 @@ public static int AllocateFileSpace(int fd, long size, string file, out bool usi
                         // fallocate is not supported, we'll use lseek instead
                         usingWrite = true;
                         byte b = 0;
-                        if (pwrite(fd, &b, 1, size - 1) != 1L)
+                        int cifsRetry = 3;
+                        while (true)
```
ayende: You can't do a `while(true)`

src/Sparrow/Platform/Posix/Syscall.cs
```diff
@@ -275,12 +276,34 @@ public static int AllocateFileSpace(int fd, long size, string file, out bool usi
...
+                                    // lets give it few retries with short pauses between them - RavenDB-11954
+                                    if (CheckSyncDirectoryAllowed(file) == false) // cifs/nfs
+                                    {
+                                        if (cifsRetry-- <= 0)
+                                        {
+                                            ThrowLastError(err,
+                                                "Failed to pwrite WITH retries (on cifs/nfs mount) in order to fallocate where fallocate is not supported for " + file);
+                                        }
+
+                                        Thread.Sleep(200);
```
ayende: Can we do less than that? Maybe compensate with increasing the number of retries

ayende (changes requested):

src/Sparrow/Platform/Posix/Syscall.cs
aviviadi: I'll write `static int RetriesOrThrow()` for both cifs retries and EINTR w/wo timeouts

src/Sparrow/Platform/Posix/Syscall.cs
aviviadi: Thread.Sleep(10 * (cifsRetry ^ 3)); // wait upto 3 time in intervals of : 10mSec, 80mSec, 270mSec

src/Sparrow/Platform/Posix/Syscall.cs
ayende: ok

PR #7625, opened 2018-09-19
[RavenDB-11866] Improve the prefetching behavior to diminish IO requests

redknightlois:


src/Voron/Data/BTrees/TreePage.cs
```diff
@@ -722,16 +722,17 @@ public void EnsureHasSpaceFor(LowLevelTransaction tx, Slice key, int len)
                 throw new InvalidOperationException("Could not ensure that we have enough space, this is probably a bug");
         }
 
-        public List<long> GetAllOverflowPages()
+        public long[] GetPagesReferencedBy()
         {
-            var results = new List<long>(NumberOfEntries);
+            var results = new long[NumberOfEntries];
```
ayende: Won't this cause a LOT of allocations now? Can we avoid the allocations?

src/Voron/Platform/Win32/WindowsMemoryMapPager.cs
```diff
@@ -132,10 +152,49 @@ public struct SplitValue
...
 
             SetPagerState(CreatePagerState());
         }
 
+        public override byte* AcquirePagePointer(IPagerLevelTransactionState tx, long pageNumber, PagerState pagerState = null)
+        {
+            // We need to decide what pager we are going to use right now or risk inconsistencies when performing prefetches from disk.
+            var state = pagerState ?? _pagerState;
+
+            if (PlatformDetails.CanPrefetch)
```
ayende, line 142: Isn't this really eager operation?

src/Voron/Platform/Win32/WindowsMemoryMapPager.cs
```diff
@@ -410,26 +475,124 @@ public override void ReleaseAllocationInfo(byte* baseAddress, long size)
...
+            this._prefetchTable[segment / 2] = value;
+        }
+
+        private void ResetPrefetchTable()
+        {
+            this._refreshCounter = 0;
+
+            // We will zero out the whole table to reset the prefetching behavior. 
+            for (int i = 0; i < this._prefetchTable.Length; i++)
+                this._prefetchTable[i] = 0;
```
ayende: `Array.Clear` ?

src/Voron/Platform/Win32/WindowsMemoryMapPager.cs
```diff
@@ -410,26 +475,124 @@ public override void ReleaseAllocationInfo(byte* baseAddress, long size)
...
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void MaybePrefetchSegment(byte* baseAddress, long segment)
+        {
+            Win32MemoryMapNativeMethods.WIN32_MEMORY_RANGE_ENTRY entry;
+            entry.NumberOfBytes = (IntPtr)PrefetchSegmentSize;
+            entry.VirtualAddress = baseAddress + segment * PrefetchSegmentSize;
+
+            Win32MemoryMapNativeMethods.PrefetchVirtualMemory(Win32Helper.CurrentProcess, (UIntPtr)1, &entry, 0);
```
ayende: Why do it one item at a time?

src/Voron/Data/BTrees/TreePage.cs
redknightlois: Nope, this happens only once. And also now those arrays are not going to be bigger than 1K elements or so.

src/Voron/Platform/Win32/WindowsMemoryMapPager.cs
redknightlois: From the call chain, there is really no reason why to delay that. At that point the call-stack is already in control of the pager, so noone is gonna change that either way.

src/Voron/Platform/Win32/WindowsMemoryMapPager.cs
redknightlois: Yes. Cause it is untyped I forgot about it. I was writing that thinking, there is no zero memory for arrays?

src/Voron/Platform/Win32/WindowsMemoryMapPager.cs
redknightlois: Because that is only used for the case of acquiring a pointer, so there is no 'context' beyond someone else has asked for this one before too.

src/Voron/Data/BTrees/TreePage.cs
redknightlois: If you ask me, I am more worried about the `MaybePrefetchPagesReferencedBy` allocations than this ones.

ayende: Why once? This happens every time you `Seek`, which we do a lot.

redknightlois: You are right. That is exactly the one I was worried about. I was under the impression you were talking about the initialization one. I was toying with the idea of using a struct iterator over a generic interface. It will avoid the allocation, completely thread safe and also generate the most efficient code when used. What do you think?

ayende: Wouldn't a simple buffer pool be enough here?

redknightlois:
Yes. A shared one would work too. Those are small arrays now

Obtener Outlook para Android<https://aka.ms/ghei36>

________________________________

redknightlois:
Iterator based to avoid allocations (even shared) altogether, and POSIX support. I may need someone like @aviviadi to take a look into it to ensure I didn't break anything... I have no way to test it without installing a whole linux environment on my end.

arekpalinski:
@redknightlois Please check failing tests: http://jenkins.hibernatingrhinos.com/job/PR_Tests_v4.1/1651/

redknightlois:
There is something weird on `SlowTests.Client.Attachments.AttachmentsSessionAsync.PutLotOfAttachments(1000)` and I would need a second pair of eyes. 

The story goes like this:
- The prefetch table is based on the size of the actual pager file. 
- The test is creating a memmap and the data suggest that there are many pages (1000+) of data, which makes sense given the name of the test.
- When the prefetcher gets page `1148` at https://github.com/ravendb/ravendb/pull/7625/files#diff-ce617e4ca933be8605b0ed4c5b91de7dR558 it calculates the proper segment for it which is way beyond the reported size of the memmap (65K)...

The underlying issue is that page `1148` should be on position `Base + 1148 * 8Kb` which is waaaaay beyond the 65K reported as the memmap total size. Any idea?

redknightlois:
Interesting new issue:
```
Error occurred during flushing journals to the data file
Voron.Exceptions.VoronUnrecoverableErrorException: Error occurred during flushing journals to the data file ---> System.OverflowException: Array dimensions exceeded supported range.
   at Voron.Platform.Win32.WindowsMemoryMapPager.AllocateMorePages(Int64 newLength) in S:\Source\ravendb-40.git\src\Voron\Platform\Win32\WindowsMemoryMapPager.cs:line 282
   at Voron.Impl.Paging.Simple4KbBatchWrites.Write(Int64 posBy4Kbs, Int32 numberOf4Kbs, Byte* source) in S:\Source\ravendb-40.git\src\Voron\Impl\Paging\AbstractPager.cs:line 520
   at Voron.Impl.Paging.AbstractPager.CopyPageImpl(I4KbBatchWrites destwI4KbBatchWrites, Int64 p, PagerState pagerState) in S:\Source\ravendb-40.git\src\Voron\Impl\Paging\AbstractPager.cs:line 471
   at Voron.Impl.Journal.WriteAheadJournal.JournalApplicator.ApplyPagesToDataFileFromScratch(Dictionary`2 pagesToWrite) in S:\Source\ravendb-40.git\src\Voron\Impl\Journal\WriteAheadJournal.cs:line 1073
   at Voron.Impl.Journal.WriteAheadJournal.JournalApplicator.ApplyLogsToDataFile(CancellationToken token, TimeSpan timeToWait) in S:\Source\ravendb-40.git\src\Voron\Impl\Journal\WriteAheadJournal.cs:line 569
   at Voron.StorageEnvironment.BackgroundFlushWritesToDataFile() in S:\Source\ravendb-40.git\src\Voron\StorageEnvironment.cs:line 1005
   --- End of inner exception stack trace ---
   at Voron.Exceptions.VoronUnrecoverableErrorException.Raise(StorageEnvironment env, String message, Exception inner) in S:\Source\ravendb-40.git\src\Voron\Exceptions\VoronUnrecoverableErrorException.cs:line 44
   at Voron.StorageEnvironment.BackgroundFlushWritesToDataFile() in S:\Source\ravendb-40.git\src\Voron\StorageEnvironment.cs:line 1025
   at Voron.GlobalFlushingBehavior.<FlushEnvironments>b__19_0(Object env) in S:\Source\ravendb-40.git\src\Voron\GlobalFlushingBehavior.cs:line 267
--- End of stack trace from previous location where exception was thrown ---
   at Voron.Impl.LowLevelTransaction..ctor(StorageEnvironment env, Int64 id, TransactionPersistentContext transactionPersistentContext, TransactionFlags flags, IFreeSpaceHandling freeSpaceHandling, ByteStringContext context) in S:\Source\ravendb-40.git\src\Voron\Impl\LowLevelTransaction.cs:line 230
   at Voron.StorageEnvironment.NewLowLevelTransaction(TransactionPersistentContext transactionPersistentContext, TransactionFlags flags, ByteStringContext context, Nullable`1 timeout) in S:\Source\ravendb-40.git\src\Voron\StorageEnvironment.cs:line 657
   at Raven.Server.ServerWide.Context.DocumentsOperationContext.CreateWriteTransaction(Nullable`1 timeout) in S:\Source\ravendb-40.git\src\Raven.Server\ServerWide\Context\DocumentsOperationContext.cs:line 47
   at Raven.Server.ServerWide.Context.TransactionOperationContext`1.OpenWriteTransaction(Nullable`1 timeout) in S:\Source\ravendb-40.git\src\Raven.Server\ServerWide\Context\TransactionOperationContext.cs:line 93
   at Raven.Server.Documents.TransactionOperationsMerger.MergeTransactionsOnce() in S:\Source\ravendb-40.git\src\Raven.Server\Documents\TransactionOperationsMerger.cs:line 471
--- End of stack trace from previous location where exception was thrown ---
   at Raven.Server.Documents.TransactionOperationsMerger.Enqueue(MergedTransactionCommand cmd) in S:\Source\ravendb-40.git\src\Raven.Server\Documents\TransactionOperationsMerger.cs:line 123
   at Raven.Client.Util.AsyncHelpers.RunSync(Func`1 task) in S:\Source\ravendb-40.git\src\Raven.Client\Util\AsyncHelpers.cs:line 79
   at Raven.Server.Smuggler.Documents.DatabaseDestination.DatabaseDocumentActions.FinishBatchOfDocuments() in S:\Source\ravendb-40.git\src\Raven.Server\Smuggler\Documents\DatabaseDestination.cs:line 243
   at Raven.Server.Smuggler.Documents.DatabaseSmuggler.ProcessDocuments(SmugglerResult result, BuildVersionType buildType) in S:\Source\ravendb-40.git\src\Raven.Server\Smuggler\Documents\DatabaseSmuggler.cs:line 553
   at Raven.Server.Smuggler.Documents.DatabaseSmuggler.ProcessType(DatabaseItemType type, SmugglerResult result, BuildVersionType buildType, Boolean ensureStepsProcessed) in S:\Source\ravendb-40.git\src\Raven.Server\Smuggler\Documents\DatabaseSmuggler.cs:line 140
   at Raven.Server.Smuggler.Documents.DatabaseSmuggler.Execute(Boolean ensureStepsProcessed) in S:\Source\ravendb-40.git\src\Raven.Server\Smuggler\Documents\DatabaseSmuggler.cs:line 70
   at Raven.Server.Documents.PeriodicBackup.Restore.RestoreBackupTask.ImportSingleBackupFile(DocumentDatabase database, Action`1 onProgress, RestoreResult restoreResult, String filePath, DocumentsOperationContext context, DatabaseDestination destination, DatabaseSmugglerOptionsServerSide options, Action`1 onIndexAction, Action`1 onDatabaseRecordAction) in S:\Source\ravendb-40.git\src\Raven.Server\Documents\PeriodicBackup\Restore\RestoreBackupTask.cs:line 518
   at Raven.Server.Documents.PeriodicBackup.Restore.RestoreBackupTask.SmugglerRestore(String backupDirectory, DocumentDatabase database, DocumentsOperationContext context, DatabaseRecord databaseRecord, Action`1 onProgress, RestoreResult result) in S:\Source\ravendb-40.git\src\Raven.Server\Documents\PeriodicBackup\Restore\RestoreBackupTask.cs:line 448
   at Raven.Server.Documents.PeriodicBackup.Restore.RestoreBackupTask.Execute(Action`1 onProgress) in S:\Source\ravendb-40.git\src\Raven.Server\Documents\PeriodicBackup\Restore\RestoreBackupTask.cs:line 183
   at Raven.Server.Documents.PeriodicBackup.Restore.RestoreBackupTask.Execute(Action`1 onProgress) in S:\Source\ravendb-40.git\src\Raven.Server\Documents\PeriodicBackup\Restore\RestoreBackupTask.cs:line 243
   at Raven.Server.Web.System.AdminDatabasesHandler.<>c__DisplayClass12_1.<<RestoreDatabase>b__1>d.MoveNext() in S:\Source\ravendb-40.git\src\Raven.Server\Web\System\AdminDatabasesHandler.cs:line 604
```

redknightlois:
@ayende Now I am pretty sure something doesnt add up. I made the change so that `PagerState` keeps the state of the table size. 

When I run the following code:
```csharp
using (var test = new AttachmentsSession())
{
   test.PutLotOfAttachments(71);
}
```
The pager sais that the amount of data available is 65K, though I am getting requested a page number 1148 which is way outside the bounds. Namely that page is around 9Mb or so. Any idea what it may be happening?

redknightlois:
Found it, the problem is that when trying to prefetch ahead and hitting the scratchpad the page may not exist. So what I did is disable the prefetching ahead when we hit the scratchpad.

src/Raven.Server/Config/Categories/StorageConfiguration.cs
```diff
@@ -50,5 +51,18 @@ public class StorageConfiguration : ConfigurationCategory
         [SizeUnit(SizeUnit.Megabytes)]
         [ConfigurationEntry("Storage.MaxScratchBufferSizeInMb", ConfigurationEntryScope.ServerWideOrPerDatabase)]
         public Size? MaxScratchBufferSize { get; set; }
+
+        [Description("Size of the batch that will be requested to the OS from disk when prefetching (value in powers of 2). Some OSs may not honor certain values. Experts only.")]
+        [DefaultValue(1024)]
+        [SizeUnit(SizeUnit.Kilobytes)]
+        [ConfigurationEntry("Storage.IOPrefetchBatchSize", ConfigurationEntryScope.ServerWideOrPerDatabase)]
```
ppekrol: Can you change `Storage.IOPrefetchBatchSize` to `Storage.IOPrefetchBatchSizeInKb` ? The property name should stay as is, but the configuration entry needs to include the units.

src/Raven.Server/Config/Categories/StorageConfiguration.cs
```diff
@@ -50,5 +51,18 @@ public class StorageConfiguration : ConfigurationCategory
...
+        public Size IOPrefetchBatchSize { get; set; }
+
+        [Description("How many gigabytes of memory should be prefetched before restarting the prefetch tracker table. Experts only.")]
+        [DefaultValue(8)]
+        [SizeUnit(SizeUnit.Gigabytes)]
+        [ConfigurationEntry("Storage.IOPrefetchResetThreshold", ConfigurationEntryScope.ServerWideOrPerDatabase)]
```
ppekrol: Can you change `Storage.IOPrefetchResetThreshold` to `Storage.IOPrefetchResetThresholdInGb` ? The property name should stay as is, but the configuration entry needs to include the units.

ppekrol (commented):
Can you resolve conflicts?

src/Voron/Platform/Win32/WindowsMemoryMapPager.cs
```diff
@@ -66,6 +61,8 @@ public struct SplitValue
             FileName = file;
             _logger = LoggingSource.Instance.GetLogger<StorageEnvironment>($"Pager-{file}");
 
+            _canPrefetchAhead = !fileAttributes.HasFlag(Win32NativeFileAttributes.Temporary);
```
ayende: Why can't we prefetch on temporary files?

redknightlois: I was a bit confused myself for a while. So let me try to explain. It is not that we cannot prefetch on temporary files, we cannot prefetchAhead which is an entirely different thing; the standard usage prefetch works on temporary files too. 

The underlying problem is that we don't have a unified view of the 'page translation' at all moments. 

So whenever we have a page number it may happen that:
- It is in the proper location on the data file (we can prefetch ahead those)
- It has been modified and written to the journal. 
- It is in the scratchpad (that is hot anyway)

Prefetching ahead assumes that the page number corresponds directly on a memory position, and that is why you prefetch, therefore no translation has to be done. If we would like to prefetch ahead, we need to know exactly where the pages are stored (which probably is not contiguously anyways). To make it work we must do the page translation but that information is not available at the time we want to do the prefetch. 

So, in essence, to prefetch ahead from a Journal is not such a good idea in general because it assumes that the data layout is already optimized. So instead of trying to page translate every entry in the tree, we just let the dynamic prefetching to take care of it and ignore any prefetch ahead request happening to end into a temporary map file.

redknightlois:
@ppekrol Done.

src/Voron/Platform/Win32/WindowsMemoryMapPager.cs
redknightlois: Moreover, the usage pattern of the prefetch ahead is especially useful when we are just starting. Which also has the side effect to disabling itself by design when the database has moved into a high frequency write mode.

ayende: Okay, make sense.

src/Voron/Platform/Posix/PosixAbstractPager.cs
```diff
@@ -63,43 +71,62 @@ public override unsafe void TryPrefetchingWholeFile()
...
+            while (pagesToPrefetch.MoveNext());
 
-                range.Add(new Win32MemoryMapNativeMethods.WIN32_MEMORY_RANGE_ENTRY
-                {
-                    VirtualAddress = baseAddress,
-                    NumberOfBytes = new IntPtr(size)
-                });
+            if (prefetchIdx != 0)
+            {
+                // We dont have enough space, so we send the batch to the kernel
```
ayende: Bad comment here.

src/Voron/Platform/Win32/WindowsMemoryMapPager.cs
```diff
@@ -190,94 +211,19 @@ protected internal override PagerState AllocateMorePages(long newLength)
...
-                PagerState.DebugVerify(newLengthAfterAdjustment);
-            }
-#endif
 
             _totalAllocationSize += allocationSize;
             NumberOfAllocatedPages = _totalAllocationSize / Constants.Storage.PageSize;
-            return newPagerState;
-        }
-
-        private bool TryAllocateMoreContinuousPages(long allocationSize)
```
ayende, line 220: This greatly simplify things.

src/Voron/Platform/Win32/WindowsMemoryMapPager.cs
```diff
@@ -410,26 +348,45 @@ public override void ReleaseAllocationInfo(byte* baseAddress, long size)
...
+                }                
             }
+            while (pagesToPrefetch.MoveNext());
 
-            fixed (Win32MemoryMapNativeMethods.WIN32_MEMORY_RANGE_ENTRY* entriesPtr = entries)
+            if (prefetchIdx != 0)
             {
-                Win32MemoryMapNativeMethods.PrefetchVirtualMemory(Win32Helper.CurrentProcess,
-                    (UIntPtr)PagerState.AllocationInfos.Length, entriesPtr, 0);
+                // We dont have enough space, so we send the batch to the kernel
```
ayende: Bad comment

src/Voron/Impl/PagerState.cs
```diff
@@ -141,5 +182,93 @@ public void DebugVerify(long size)
...
+        {
+            long segmentNumber = (pageNumber * Constants.Storage.PageSize) >> this._segmentShift;
+
+            int segmentState = GetSegmentState(segmentNumber);
+            if (segmentState < AlreadyPrefetch)
+            {
+                // We update the current segment counter
+                segmentState++;
+
+                int previousSegmentState = GetSegmentState(segmentNumber - 1);
```
ayende, line 246: Need to also go over the  next segment as well, to make sure that we are not going backward.

src/Voron/Impl/PagerState.cs
```diff
@@ -141,5 +182,93 @@ public void DebugVerify(long size)
...
+            }
+
+            sizeInBytes = 0;
+            virtualAddress = null;
+            return false;
+        }
+
+        public void CheckResetPrefetchTable()
+        {
+            if (_refreshCounter > this._prefetchResetThreshold)
```
ayende, line 280: This isn't thread safe.

src/Voron/Impl/PagerState.cs
redknightlois: We are not, forgot about that one... in a sec

PR #7629, opened 2018-09-20
RavenDB-11923 Smuggler.LegacySmugglerTests.CanImportIndexesAndTransfo…

karmeli87:
…rmers

When running in-memory all of the indexes were created in the same temp folder, which in this case genreated over 1,500 files in a single folder.
This crippled the database deletion, since it had to enumerate so many files (`Directory.GetFiles(tempPath.FullPath, "*.buffers")`).

So the fix is to create each index in a seperate folder.


ayende:
This looks like it broke some of the tests. `WhoIsLocking` seems to have failed and masked the actual error

PR #7641, opened 2018-09-26
[v4.1]Raven db 11985

maximburyak:
http://issues.hibernatingrhinos.com/issue/RavenDB-11985

### DIFF
--- a/src/Sparrow/Json/JsonOperationContext.cs
+++ b/src/Sparrow/Json/JsonOperationContext.cs
@@ -572,7 +572,7 @@ namespace Sparrow.Json
         public unsafe BlittableJsonReaderObject ReadForMemory(string jsonString, string documentId)
         {
             // todo: maybe use ManagedPinnedBuffer here
-            var maxByteSize = Encodings.Utf8.GetByteCount(jsonString);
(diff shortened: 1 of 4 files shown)


src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -569,6 +569,32 @@ public BlittableJsonReaderObject ReadForMemory(Stream stream, string documentId)
             return ParseToMemory(stream, documentId, BlittableJsonDocumentBuilder.UsageMode.None);
         }
 
+        public unsafe BlittableJsonReaderObject ReadForMemory(string jsonString, string documentId)
+        {
+            // todo: maybe use ManagedPinnedBuffer here
```
ayende, line 574: No need, but can use MaxByteCount, which is cheaper and avoid traversing the whole string.
We don't need an exact count here.

arekpalinski:
LGTM

src/Sparrow/Json/JsonOperationContext.cs
maximburyak: done

PR #7651, opened 2018-09-28
RavenDB-11923 Reverting recent change which caused that temp files of…

arekpalinski:
… indexes were put in Indexes\Temp\{index-name} instead of Indexes\{index-name}\Temp


arekpalinski:
CC @karmeli87

PR #7657, opened 2018-09-29
[4.0] RavenDB-11908 - We want to report CPU percentage that are based on the number of cores we actually use

grisha-kotler:
http://issues.hibernatingrhinos.com/issue/RavenDB-11908

### DIFF
--- a/src/Raven.Server/Utils/CpuUsage.cs
+++ b/src/Raven.Server/Utils/CpuUsage.cs
@@ -77,6 +77,15 @@ namespace Raven.Server.Utils
                 return _lastCpuInfo?.ProcessCpuUsage ?? 0;
             }
 
+            if (currentInfo.ActiveCores == 0)
+            {
+                // shouldn't happen
+                if (Logger.IsInfoEnabled)
+                    Logger.Info($"ActiveCores == 0, OS: {RuntimeInformation.OSDescription}");
+
+                return _lastCpuInfo?.ProcessCpuUsage ?? 0;
+            }
+
             var processCpuUsage = (processorTimeDiff * 100.0) / timeDiff / currentInfo.ActiveCores;
             if (currentInfo.ActiveCores == ProcessorInfo.ProcessorCount)
             {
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Utils/CpuUsage.cs
```diff
@@ -329,62 +267,154 @@ private static unsafe MacInfo GetMacInfo()
...
+                return Bits.NumberOfSetBits(process.ProcessorAffinity.ToInt64());
+            }
+            catch (NotSupportedException)
+            {
+                return ProcessorInfo.ProcessorCount;
+            }
+            catch (Exception e)
+            {
+                if (Logger.IsInfoEnabled)
+                    Logger.Info($"Failure to get the number of active cores, error: {e.Message}", e);
```
ayende: There is not need to include the `Message` here, it is already in the exception

src/Raven.Server/Utils/CpuUsage.cs
```diff
@@ -46,24 +48,43 @@ public static (double MachineCpuUsage, double ProcessCpuUsage) Calculate()
...
+        {
+            var processorTimeDiff = currentInfo.TotalProcessorTimeTicks - previousInfo.TotalProcessorTimeTicks;
+            var timeDiff = currentInfo.TimeTicks - previousInfo.TimeTicks;
+            if (timeDiff <= 0)
+            {
+                //overflow
+                return _lastCpuInfo?.ProcessCpuUsage ?? 0;
+            }
+
+            var processCpuUsage = (processorTimeDiff * 100.0) / timeDiff / currentInfo.ActiveCores;
```
ayende, line 89: Need to check that `ActiveCores` isn't zero either

grisha-kotler: done

ppekrol:
@grisha-kotler can you resolve conflicts?

grisha-kotler:
done

PR #7664, opened 2018-09-30
[4.1] RavenDB-12016 - No upload progress when importing a database

grisha-kotler:
http://issues.hibernatingrhinos.com/issue/RavenDB-12016

### DIFF
--- a/src/Raven.Studio/typescript/commands/commandBase.ts
+++ b/src/Raven.Studio/typescript/commands/commandBase.ts
@@ -105,13 +105,19 @@ class commandBase {
         const url = baseUrl ? baseUrl + appUrl.forDatabaseQuery(db) + relativeUrl : appUrl.forDatabaseQuery(db) + relativeUrl;
 
         const xhrConfiguration = (xhr: XMLHttpRequest) => {
+            const isUploadingFile = url.includes("smuggler/import");
             xhr.upload.addEventListener("progress", (evt: ProgressEvent) => {
-                if (evt.lengthComputable) {
-                    const percentComplete = (evt.loaded / evt.total) * 100;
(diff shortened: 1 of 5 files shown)


src/Raven.Studio/typescript/commands/commandBase.ts
```diff
@@ -105,13 +105,19 @@ class commandBase {
         const url = baseUrl ? baseUrl + appUrl.forDatabaseQuery(db) + relativeUrl : appUrl.forDatabaseQuery(db) + relativeUrl;
 
         const xhrConfiguration = (xhr: XMLHttpRequest) => {
+            const isUploadingFile = url.includes("smuggler/import");
```
ml054: Can we avoid putting if in common class? It is better to customize this in target class instead of common class. I think we should put this logic in importDatabaseFromFile + importDatabaseFromCsv. See https://github.com/ravendb/ravendb/blob/0b46405e609328bcff6629479a4a1e580e806f46/src/Raven.Studio/typescript/commands/commandBase.ts#L135 and https://github.com/ravendb/ravendb/blob/4c56afcd420d952ea8d87398cb26e9ba96c84528/src/Raven.Studio/typescript/commands/database/documents/attachments/uploadAttachmentCommand.ts#L30 

Please also make sure that attachment progess is still working fine. 

After changes "UploadProgress" event should not be needed.

ml054 (changes requested):

src/Raven.Studio/typescript/commands/commandBase.ts
grisha-kotler: done

ml054 (approved):

PR #7675, opened 2018-10-02
RavenDB-7836: Added system / db directory validation before using offline-operations

garayx:
WIP do not merge

### DIFF
--- a/tools/rvn/CommandLineApp.cs
+++ b/tools/rvn/CommandLineApp.cs
@@ -362,28 +362,29 @@ namespace rvn
             return cmd.Option("--server-dir", "RavenDB Server directory", CommandOptionType.SingleValue);
         }
 
-        private static void ValidateRavenSystemDir(CommandArgument systemDirArg)
+        private static void ValidateRavenDirectory(CommandArgument argument)
         {
-            var fullPath = systemDirArg.Value.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar)
-                           + Path.DirectorySeparatorChar;
-            if (string.IsNullOrEmpty(systemDirArg.Value))
+            if (string.IsNullOrWhiteSpace(argument.Value))
             {
                 throw new InvalidOperationException("RavenDB system directory argument is mandatory.");
             }
 
-            if (Directory.Exists(systemDirArg.Value) == false)
+            var trimmedFullPath = argument.Value.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+            var directory = new DirectoryInfo(trimmedFullPath);
+
+            if (directory.Exists == false)
             {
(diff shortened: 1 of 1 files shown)


tools/rvn/CommandLineApp.cs
```diff
@@ -364,6 +364,8 @@ private static CommandOption ConfigureServerDirOption(CommandLineApplication cmd
 
         private static void ValidateRavenSystemDir(CommandArgument systemDirArg)
         {
+            var fullPath = systemDirArg.Value.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar)
```
ppekrol: use DirectoryInfo instead

tools/rvn/CommandLineApp.cs
```diff
@@ -373,6 +375,17 @@ private static void ValidateRavenSystemDir(CommandArgument systemDirArg)
...
             }
+
+            if (fullPath.EndsWith("System\\"))
+            {
+                if (File.Exists(fullPath + "Raven.voron") == false)
+                    throw new InvalidOperationException("Please provide a valid System directory.");
+            }
+            else
+            {
+                if (Directory.Exists(fullPath + "Journals") == false)
```
ppekrol: use Path.Combine

tools/rvn/CommandLineApp.cs
```diff
@@ -373,6 +375,17 @@ private static void ValidateRavenSystemDir(CommandArgument systemDirArg)
             {
                 throw new InvalidOperationException($"Directory does not exist: { systemDirArg.Value }.");
             }
+
+            if (fullPath.EndsWith("System\\"))
```
ppekrol: If you will use DirectoryInfo then DirectoryInfo.Name should be what you want IIRC

tools/rvn/CommandLineApp.cs
garayx: In case of systemDirArg = "C:\Raven_DB\Server\RavenData\System" 
DirectoryInfo returns "C:\Raven_DB\Server\RavenData"
in case of systemDirArg = "C:\Raven_DB\Server\RavenData\System\" 
DirectoryInfo returns "C:\Raven_DB\Server\RavenData\System"

I need System folder in both cases (with/without backslash in the end of systemDirArg).

ppekrol: I've meant that after creating a full path you can do:
1. `var targetDirectory = new DirectoryInfo(fullPath)`
2. `if (targetDirectory.Name == "System")` -> you cannot assume that `\` is a valid separator on every platform
3. `if (Path.Combine(targetDirectory.FullName, Voron.Constants.DatabaseFilename))`
4. `if (Directory.Exists(targetDirectory.FullName, "Journals"))`

ppekrol (approved):

PR #7683, opened 2018-10-03
RavenDB-12025 High CPU usage in idle state

karmeli87:
This was usually the case when we had many indexes and/or databases (several hundreds in total).

- Reduce network traffic by sending an empty report to the supervisor if no changes were made to any of the database's envs.
- Reduce openning of read transactions and disk access when no change was made to a specific index or database.
- Reduce DatabaseRecord deserialization costs when the database is disabled by accessing the blittable directly.
- Bump the maintenance TCP protocol version and rename `Heartbeats` to `Maintenance`.

### DIFF
--- a/src/Raven.Server/ServerWide/Maintenance/ClusterMaintenanceSupervisor.cs
+++ b/src/Raven.Server/ServerWide/Maintenance/ClusterMaintenanceSupervisor.cs
@@ -251,9 +251,9 @@ namespace Raven.Server.ServerWide.Maintenance
 
                                 var nodeReport = BuildReport(rawReport);
                                 timeoutEvent.Defer(_parent._leaderClusterTag);
-
-                                unchangedReports.Clear();
+                                
                                 UpdateNodeReportIfNeeded(nodeReport, unchangedReports);
+                                unchangedReports.Clear();
 
                                 ReceivedReport = _lastSuccessfulReceivedReport = nodeReport;
                             }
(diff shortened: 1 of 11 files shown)


karmeli87:
If we want to port this change to 4.0 version also, then it would be preferable to change the TCP version to something like 40_700.

src/Raven.Server/ServerWide/Maintenance/ClusterMaintenanceSupervisor.cs
```diff
@@ -248,7 +249,16 @@ private void ListenToMaintenanceWorker()
 
                                 var report = BuildReport(rawReport);
                                 timeoutEvent.Defer(_parent._leaderClusterTag);
-
+                                foreach (var name in report.Report.Keys.ToList())
+                                {
+                                    var dbReport = report.Report[name];
+                                    if (dbReport.Status == DatabaseStatus.NoChange)
+                                    {
+                                        report.Report[name] = _lastSuccessfulReceivedReport.Report[name];
```
ayende: Is this correct? It looks like we'll miss out on the things that are updated, no? Like the replication stats?

src/Raven.Server/ServerWide/Maintenance/ClusterMaintenanceSupervisor.cs
```diff
@@ -248,7 +249,16 @@ private void ListenToMaintenanceWorker()
 
                                 var report = BuildReport(rawReport);
                                 timeoutEvent.Defer(_parent._leaderClusterTag);
-
+                                foreach (var name in report.Report.Keys.ToList())
```
ayende: Better to avoid this allocation (should be false most of the time when we are active) and only allocate if we indeed get a no chance notification.
For that matter, since we are single threaded, we can just have a separate list of the modified values that we'll then set after the `foreach`

ayende (changes requested):

src/Raven.Server/ServerWide/Maintenance/ClusterMaintenanceSupervisor.cs
karmeli87: Fixed.

src/Raven.Server/ServerWide/Maintenance/ClusterMaintenanceSupervisor.cs
karmeli87: Fixed.

src/Raven.Server/ServerWide/Maintenance/ClusterMaintenanceSupervisor.cs
```diff
@@ -269,6 +275,35 @@ private void ListenToMaintenanceWorker()
...
+                    }
+                }
+
+                if (unchangedReports.Count == 0)
+                    return;
+
+                foreach (var dbReport in unchangedReports)
+                {
+                    var dbName = dbReport.Name;
+                    if(_lastSuccessfulReceivedReport.TryGetValue(name, out var previous) == false)
```
ayende: @karmeli87 Need to see if the caller code can handle the case where we get the unchanged directly.

ayende:
Please check the failing tests

karmeli87:
I think they fail regardless to this PR
http://issues.hibernatingrhinos.com/issue/RavenDB-11917
http://issues.hibernatingrhinos.com/issue/RavenDB-11968

PR #7686, opened 2018-10-04
RavenDB-11815 support for grouping by nested field values

ppekrol:


arekpalinski:
@ppekrol please remember about testing multi map-reduce index which I'm pretty sure will fail

PR #7690, opened 2018-10-04
RavenDB-11643 Fixing unhandled case in the reduce phase

arekpalinski:


src/Raven.Server/Documents/Indexes/MapReduce/MapReduceIndexingContext.cs
```diff
@@ -20,7 +20,16 @@ public class MapReduceIndexingContext : IDisposable
...
         public Dictionary<string, long> ProcessedTombstoneEtags = new Dictionary<string, long>();
-        
+
+        public event Action<long> PageModifiedInReduceTree;
+
+        public void OnPageModifiedInReduceTree(long page)
+        {
+            var pageModified = PageModifiedInReduceTree;
+
+            pageModified?.Invoke(page);
```
ayende, line 30: This can be simply `PageModifiedInReduceTree?.Invoke(page)` with the same exact behavior

arekpalinski: Right. Initially I did it the old way, then applied r# suggestion but didn't remove the local var. I'll change it.

PR #7691, opened 2018-10-04
RavenDB-11968 Fix CanPreformSeveralClusterTransactions test that thro…

karmeli87:
…ws NotLeadingException

The node could have been the leader while the rachis request arrived, but lost it afterwards.
In that case we need to throw an appropriate exception to the client (the executor that issued the request).
Moreover, this fix will cause for the raft command to be retried as long as we have any valid leader.

### DIFF
--- a/src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
@@ -45,51 +45,55 @@ namespace Raven.Server.Documents.Handlers.Admin
                 HttpContext.Response.Headers["Reached-Leader"] = "true";
 
                 var commandJson = await context.ReadForMemoryAsync(RequestBodyStream(), "external/rachis/command");
-                CommandBase command;
                 try
                 {
-                    command = CommandBase.CreateFrom(commandJson);
+                    var command = CommandBase.CreateFrom(commandJson);
+                    switch (command)
+                    {
+                        case AddOrUpdateCompareExchangeBatchCommand batchCmpExchange:
+                            batchCmpExchange.ContextToWriteResult = context;
+                            break;
+                        case CompareExchangeCommandBase cmpExchange:
+                            cmpExchange.ContextToWriteResult = context;
+                            break;
+                    }
+
+                    var isClusterAdmin = IsClusterAdmin();
+                    command.VerifyCanExecuteCommand(ServerStore, context, isClusterAdmin);
+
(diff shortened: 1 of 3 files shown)


ppekrol (approved):

ppekrol:
@ayende and feedback here?

src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
```diff
@@ -45,51 +45,55 @@ public async Task ApplyCommand()
...
-                switch (command)
-                {
-                    case AddOrUpdateCompareExchangeBatchCommand batchCmpExchange:
-                        batchCmpExchange.ContextToWriteResult = context;
-                        break;
-                    case CompareExchangeCommandBase cmpExchange:
-                        cmpExchange.ContextToWriteResult = context;
-                        break;
-                }
+                    var command = CommandBase.CreateFrom(commandJson);
```
ayende, line 50: What happened to the `AssertClientVersion` ?

src/Raven.Server/Rachis/Leader.cs
```diff
@@ -732,6 +739,10 @@ private void EmptyQueue()
             }
             catch (Exception e)
             {
+                if (_running.IsRaised() == false)
+                {
+                    e = lostLeadershipException;
```
ayende: Probably a bad idea. We need to create a new instance with the inner exception set.

ayende (commented):
I'm not seeing where the client will retry on no leading?

src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
karmeli87: In the 'catch' part

karmeli87:
`PutAsync` is basically called from 2 places. 
First, from `SendToLeaderAsyncInternal`, if we are the leader we will try to put the command directly and on catching `NotLeadingException` we will retry to put it directly again in case we are still the leader or to send this command to the new leader.

in case we are not the leader, we will send the command on the wire to `/admin/rachis/send` endpoint where we also call `PutAsync`, in case of `NotLeadingException` the request will return an error with `Reached-Leader` set to `false`, which will cause the sender to retry.

src/Raven.Server/Rachis/Leader.cs
karmeli87: done.

ayende (approved):

PR #7692, opened 2018-10-04
RavenDB-11919 Fix DistributedRevisionsSubscription test

karmeli87:
We have to wait for the subscription to be created on the responsible node.

### DIFF
--- a/src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
@@ -155,14 +155,14 @@ namespace Raven.Server.Documents.Handlers
 
                 var name = options.Name ?? subscriptionId.ToString();
 
-                // need to wait on the relevant remote node
-                var node = Database.SubscriptionStorage.GetResponsibleNode(name);
-                if (node != null && node != ServerStore.NodeTag)
+                using (ServerStore.ContextPool.AllocateOperationContext(out TransactionOperationContext serverContext))
+                using (serverContext.OpenReadTransaction())
                 {
-                    using (ServerStore.ContextPool.AllocateOperationContext(out TransactionOperationContext serverCtx))
-                    using (serverCtx.OpenReadTransaction())
+                    // need to wait on the relevant remote node
+                    var node = Database.SubscriptionStorage.GetResponsibleNode(serverContext, name);
+                    if (node != null && node != ServerStore.NodeTag)
                     {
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
```diff
@@ -152,13 +152,27 @@ public async Task Create()
                 var disabled = GetBoolValueQueryString("disabled", required: false);
                 var mentor = options.MentorNode;
                 var subscriptionId = await Database.SubscriptionStorage.PutSubscription(options, id, disabled, mentor: mentor);
+
+                var name = options.Name ?? subscriptionId.ToString();
+
+                // need to wait on the relevant remote node
+                var node = Database.SubscriptionStorage.GetResponsibleNode(name);
```
ppekrol: can we avoid opening 2 server-store transactions by adding ctx argument to this method and moving it next to `WaitForExecutionOnSpecificNode` ?

ppekrol: Done

src/Raven.Server/Web/RequestHandler.cs
```diff
@@ -132,6 +138,63 @@ private static bool HeadersAllowGzip(IDictionary<string, StringValues> headers,
             return false;
         }
 
+        protected async Task WaitForExecutionOnSpecificNode(TransactionOperationContext context, ClusterTopology clusterTopology, string node, long index)
+        {
+            using (var requester = ClusterRequestExecutor.CreateForSingleNode(clusterTopology.GetUrlFromTag(node), ServerStore.Server.Certificate.Certificate))
```
ayende, line 143: Why is this an instance method that also create a new _instance_ for this?
What does this give us?

src/Raven.Server/Web/RequestHandler.cs
```diff
@@ -132,6 +138,63 @@ private static bool HeadersAllowGzip(IDictionary<string, StringValues> headers,
...
+            };
+            var cts = CancellationTokenSource.CreateLinkedTokenSource(ServerStore.ServerShutdown);
+            try
+            {
+                foreach (var member in members)
+                {
+                    var url = member == ServerStore.NodeTag ?
+                        Server.WebUrl :
+                        clusterTopology.GetUrlFromTag(member);
+                    var requester = ClusterRequestExecutor.CreateForSingleNode(url, ServerStore.Server.Certificate.Certificate);
```
ayende, line 168: Same thing here, actually. We are creating new instances from an instance method, which seems strange.

PR #7696, opened 2018-10-05
RavenDB-12033 Fixed NRE. We no longer needs files to be collection, w…

arekpalinski:
…e always use single value


redknightlois:
Looks good

PR #7701, opened 2018-10-05
RavenDB-11692 sibling propagation timer should fire in interval, not once

ppekrol:


ayende (approved):
😢

ppekrol:
I've also switched to `TimerManager` now

PR #7708, opened 2018-10-08
RavenDB-12023 - threads usage in the console

grisha-kotler:
http://issues.hibernatingrhinos.com/issue/RavenDB-12023

### DIFF
--- a/src/Raven.Server/Dashboard/ThreadsInfoNotificationSender.cs
+++ b/src/Raven.Server/Dashboard/ThreadsInfoNotificationSender.cs
@@ -12,7 +12,7 @@ namespace Raven.Server.Dashboard
     {
         private readonly ConcurrentSet<ConnectedWatcher> _watchers;
         private readonly TimeSpan _notificationsThrottle;
-
+        private readonly ThreadsUsage _threadsUsage;
         private DateTime _lastSentNotification = DateTime.MinValue;
 
         public ThreadsInfoNotificationSender(string resourceName,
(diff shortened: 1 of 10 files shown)


src/Raven.Server/Utils/ThreadsUsage.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        public ThreadsUsage()
+        {
+            using (var process = Process.GetCurrentProcess())
+            {
+                _processTimes = CpuUsage.GetProcessTimes(process);
+            }
+
+            // calculate the initial threads usage
+            Thread.Sleep(10);
```
ayende: why?

src/Raven.Server/Dashboard/ThreadsInfoNotificationSender.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Raven.Server.Background;
+using Raven.Server.NotificationCenter;
+using Raven.Server.Utils;
+using Sparrow.Collections;
+
+namespace Raven.Server.Dashboard
+{
+    public class ThreadsInfoNotificationSender : BackgroundWorkBase
```
ppekrol, line 11: This is not used anywhere?

grisha-kotler: This will be used when the studio part will be implemented

src/Raven.Server/Utils/ThreadsUsage.cs
grisha-kotler: need initial values for _threadTimesInfo

ayende: Isn't there a better way to do that rather than force a sleep?

grisha-kotler: done

ayende (approved):

PR #7711, opened 2018-10-08
RavenDB-12025 Reduce the deserialization costs in the cluster observer

karmeli87:
- Most of the time we need only the database topology and no need to deserialize the whole database record (which can be big when there are many indexes).
- Run the cluster observer on a dedicated thread.

### DIFF
--- a/src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
+++ b/src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
@@ -14,21 +14,22 @@ using Raven.Client.Util;
 using Raven.Server.Config;
 using Raven.Server.Config.Settings;
 using Raven.Server.Documents.Indexes;
+using Raven.Server.Json;
 using Raven.Server.NotificationCenter.Notifications;
 using Raven.Server.Rachis;
 using Raven.Server.ServerWide.Commands;
 using Raven.Server.ServerWide.Commands.Indexes;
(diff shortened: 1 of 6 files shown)


src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
```diff
@@ -62,12 +64,24 @@ class ClusterObserver : IDisposable
...
+
+            _observe = PoolOfThreads.GlobalRavenThreadPool.LongRunning(_ =>
+            {
+                try
+                {
+                    Run(_cts.Token);
+                }
+                catch
+                {
+                   // nothing we can do here
```
ayende, line 82: You have to at least log this.

karmeli87: Any exception within the observer loop will be caught and logged inside the loop itself.
If we catch anything here, it probably means logging error or some race on the cancellation token.

ppekrol (approved):

src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
ayende: This is the case right now, but it might not be there in the future. We don't try to protect ourselves from logging errors, note.

ppekrol:
@karmeli87 I'm merging. Please send a separate comment with the adjustments.

PR #7712, opened 2018-10-08
Fix for RavenDB-12041

myarichuk:
Make sure not to create during Smuggler "between" operations, Bulk Insert connection for the export side - so it won't time out and crash Smuggler process

### DIFF
--- a/Raven.Database/Smuggler/SmugglerDatabaseApi.cs
+++ b/Raven.Database/Smuggler/SmugglerDatabaseApi.cs
@@ -37,10 +37,16 @@ namespace Raven.Smuggler
             SetDatabaseNameIfEmpty(betweenOptions.To);
 
             using (var exportStore = CreateStore(betweenOptions.From))
-            using (var exportBulkOperation = CreateBulkInsertOperation(exportStore))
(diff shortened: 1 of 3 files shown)


Raven.Database/Smuggler/SmugglerDatabaseApi.cs
```diff
@@ -61,7 +67,7 @@ public override async Task Between(SmugglerBetweenOptions<RavenConnectionStringO
                         From = exportOperations,
                         To = importOperations,
                         IncrementalKey = betweenOptions.IncrementalKey
-                    }, Options, exportBulkOperation)
+                    }, Options, null)
```
ppekrol: It looks like SmugglerBetweenOperations does not require the ExportBulkInsert in any of the cases now. It can be removed from arguments?

myarichuk: No reason for it to be there I think.

myarichuk: Will change that before the actual PR

PR #7713, opened 2018-10-09
RavenDB-12040  Limit the number of slow write records that we keep in…

arekpalinski:
… storage

### DIFF
--- a/test/SlowTests/Issues/RavenDB_12040.cs
+++ b/test/SlowTests/Issues/RavenDB_12040.cs
@@ -14,15 +14,18 @@ namespace SlowTests.Issues
         {
             using (var store = GetDocumentStore())
             {
+                const int limitExceededCount = 10;
+
                 var database = await GetDatabase(store.Database);
 
-                for (int i = 0; i < SlowWritesDetails.MaxNumberOfWrites + 10; i++)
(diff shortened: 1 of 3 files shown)


ayende (approved):

test/SlowTests/Issues/RavenDB_12040.cs
```diff
@@ -0,0 +1,42 @@
...
+            {
+                var database = await GetDatabase(store.Database);
+
+                for (int i = 0; i < SlowWritesDetails.MaxNumberOfWrites + 10; i++)
+                {
+                    database.NotificationCenter
+                        .SlowWrites
+                        .Add($"C:\\Raven\\Indexes\\MyIndex\\{i}.journal", 1, 10);
+
+                    Thread.Sleep(1);
```
ayende: That said, can we avoid the `Thread.Sleep` (will take 0.51 seconds here) in favor of passing the date directly?

arekpalinski: I don't want to expose a parameter allowing to pass the date. Reduced the extra time to 10 ms: https://github.com/ravendb/ravendb/pull/7713/commits/c1a442ffc97591b6012e964dd1edc915281f97ac

ayende: That's fine, let's get this merged.

PR #7715, opened 2018-10-09
RavenDB-12025 Fix NRE and InvalidCast

karmeli87:


src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
```diff
@@ -1177,7 +1177,10 @@ public bool TryGetAutoIndex(string name,out AutoIndexDefinition definition)
                 var deletionInProgress = new Dictionary<string, DeletionInProgressStatus>();
                 foreach (var prop in obj.GetPropertyNames())
                 {
-                    deletionInProgress[prop] = (DeletionInProgressStatus)obj[prop];
+                    if (Enum.TryParse(obj[prop].ToString(), out DeletionInProgressStatus result))
```
ayende, line 1180: Is this always going to be a string here?

karmeli87: Something should go very wrong if this isn't the case.

ppekrol: Yes, this is from:
```
public Dictionary<string, DeletionInProgressStatus> DeletionInProgress;
```

ppekrol: @karmeli87 

```
obj[prop]?.ToString()
```

Maybe just in case change to that?

PR #7720, opened 2018-10-10
RavenDB-12025 Reduce dashboard resource utilization in idle work

karmeli87:
- Disk space checks are expensive, so we check for it every 30 seconds on idle database
- Cache the database info for reuse on idle databases
- Use the blittable database record for setting offline info

### DIFF
--- a/src/Raven.Server/Dashboard/DatabasesInfoNotificationSender.cs
+++ b/src/Raven.Server/Dashboard/DatabasesInfoNotificationSender.cs
@@ -9,6 +9,7 @@ using System.Threading.Tasks;
 using Raven.Client;
 using Raven.Client.ServerWide;
 using Raven.Client.ServerWide.Operations;
+using Raven.Client.Util;
 using Raven.Server.Background;
 using Raven.Server.Documents;
 using Raven.Server.Json;
@@ -30,7 +31,7 @@ namespace Raven.Server.Dashboard
         private DateTime _lastSentNotification = DateTime.MinValue;
 
         public DatabasesInfoNotificationSender(string resourceName, ServerStore serverStore,
-            ConcurrentSet<ConnectedWatcher> watchers, TimeSpan notificationsThrottle, CancellationToken shutdown) 
+            ConcurrentSet<ConnectedWatcher> watchers, TimeSpan notificationsThrottle, CancellationToken shutdown)
             : base(resourceName, shutdown)
         {
             _serverStore = serverStore;
@@ -40,7 +41,7 @@ namespace Raven.Server.Dashboard
 
         protected override async Task DoWork()
         {
-            var now = DateTime.UtcNow;
+            var now = SystemTime.UtcNow;
             var timeSpan = now - _lastSentNotification;
             if (timeSpan < _notificationsThrottle)
             {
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Dashboard/DatabasesInfoNotificationSender.cs
```diff
@@ -170,6 +204,20 @@ public static IEnumerable<AbstractDashboardNotification> FetchDatabasesInfo(Serv
...
+            foreach (var mountPointUsage in database.GetMountPointsUsage())
+            {
+                if (cts.IsCancellationRequested)
+                    return;
+
+                UpdateMountPoint(mountPointUsage, databaseName, drivesUsage);
+                CachedDatabaseInfo[databaseName].MountPoints.Add(mountPointUsage);
+            }
+
+            CachedDatabaseInfo[databaseName].NextDiskSpaceCheck = DateTime.UtcNow.AddSeconds(30);
```
ppekrol: SystemTime.UtcNow

src/Raven.Server/Dashboard/DatabasesInfoNotificationSender.cs
```diff
@@ -148,12 +146,48 @@ public static IEnumerable<AbstractDashboardNotification> FetchDatabasesInfo(Serv
...
-                        foreach (var mountPointUsage in database.GetMountPointsUsage())
+                        var currentEnvironmentsHash = database.GetEnvironmentsHash();
+                        if (CachedDatabaseInfo.TryGetValue(databaseName, out var cached) && cached.Hash == currentEnvironmentsHash)
                         {
-                            if (cts.IsCancellationRequested)
-                                yield break;
+                            databasesInfo.Items.Add(cached.Item);
 
-                            UpdateMountPoint(mountPointUsage, databaseName, drivesUsage);
+                            if (cached.NextDiskSpaceCheck < DateTime.UtcNow)
```
ppekrol: SystemTime.UtcNow

ppekrol (approved):

src/Raven.Server/Dashboard/DatabasesInfoNotificationSender.cs
```diff
@@ -148,12 +146,48 @@ public static IEnumerable<AbstractDashboardNotification> FetchDatabasesInfo(Serv
...
+                            {
+                                CachedDatabaseInfo[databaseName].MountPoints = new List<Client.ServerWide.Operations.MountPointUsage>();
```
ppekrol: Actually, you need to use cached.MountPoints here and pass it to the DiskUsageCheck to avoid concurrency issues

src/Raven.Server/Dashboard/DatabasesInfoNotificationSender.cs
```diff
@@ -170,6 +204,20 @@ public static IEnumerable<AbstractDashboardNotification> FetchDatabasesInfo(Serv
...
 
+        private static void DiskUsageCheck(CancellationTokenSource cts, DocumentDatabase database, string databaseName, DrivesUsage drivesUsage)
+        {
+            foreach (var mountPointUsage in database.GetMountPointsUsage())
+            {
+                if (cts.IsCancellationRequested)
+                    return;
+
+                UpdateMountPoint(mountPointUsage, databaseName, drivesUsage);
+                CachedDatabaseInfo[databaseName].MountPoints.Add(mountPointUsage);
```
ppekrol: See my comment with concurrency

src/Raven.Server/Dashboard/DatabasesInfoNotificationSender.cs
ppekrol: see my comment with concurrency

ppekrol (changes requested):

src/Raven.Server/Dashboard/DatabasesInfoNotificationSender.cs
ppekrol: Passing the DatabaseInfoCache instance should suffice or your case

ppekrol:
@karmeli87 can you review?

karmeli87:
Looks good, but we wanted to do things slightly more efficient in the `database.GetMountPointsUsage()`, where we have many redundant calls and calculations, also we can avoid the concurrent dictionary, but both issues could be done in the next PR :-)

PR #7731, opened 2018-10-11
RavenDB-12067 Fix ThrowTransactionAlreadyOpened if there is few number of retries in executeQuery

garayx:
If there is are few NumberOfRetries to execute a query (happens on 32bit) we are getting a TransactionAlreadyOpened exception and if we are using CmpXchangeInternal we are getting fail to read a disposed object exception..

### DIFF
--- a/src/Raven.Server/Documents/Patch/ScriptRunner.cs
+++ b/src/Raven.Server/Documents/Patch/ScriptRunner.cs
@@ -1034,18 +1034,17 @@ namespace Raven.Server.Documents.Patch
             {
                 if (string.IsNullOrEmpty(key))
                     return JsValue.Undefined;
-                BlittableJsonReaderObject value = null;
+
                 using (_database.ServerStore.ContextPool.AllocateOperationContext(out TransactionOperationContext ctx))
                 using (ctx.OpenReadTransaction())
                 {
-                    value = _database.ServerStore.Cluster.GetCompareExchangeValue(ctx, key).Value;
+                    var value = _database.ServerStore.Cluster.GetCompareExchangeValue(ctx, key).Value;
+                    if (value == null)
+                        return null;
+
+                    var jsValue = TranslateToJs(ScriptEngine, _jsonCtx, value.Clone(_jsonCtx));
+                    return jsValue.AsObject().Get("Object");
                 }
-
-                if (value == null)
-                    return null;
-
-                var jsValue = TranslateToJs(ScriptEngine, _jsonCtx, value);
-                return jsValue.AsObject().Get("Object");
             }
 
             private JsValue LoadDocumentInternal(string id)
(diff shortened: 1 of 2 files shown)


src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -1034,18 +1034,17 @@ private JsValue CmpXchangeInternal(string key)
...
-                    return null;
+                    var value = _database.ServerStore.Cluster.GetCompareExchangeValue(ctx, key).Value;
+                    if (value == null)
+                        return null;
 
-                var jsValue = TranslateToJs(ScriptEngine, _jsonCtx, value);
-                return jsValue.AsObject().Get("Object");
+                    var jsValue = TranslateToJs(ScriptEngine, _jsonCtx, value.Clone(_jsonCtx));
+                    return jsValue.AsObject().Get("Object");
+                }
```
ppekrol, line 1047: Why do we have to extend this to 'TranslateToJs'. The `documentsContext.CloseTransaction();` is not enough in `ExecuteQuery` ?

src/Raven.Server/Documents/Queries/QueryRunner.cs
```diff
@@ -58,6 +58,8 @@ public override async Task<DocumentQueryResult> ExecuteQuery(IndexQueryServerSid
             {
                 try
                 {
+                    documentsContext.CloseTransaction();
```
ppekrol, line 61: Do we need to apply this to other retried queries? Check where we are using `CreateRetriesFailedException`

src/Raven.Server/Documents/Patch/ScriptRunner.cs
garayx: Those are two different transaction: one for the compEx access and the other for the document access. We accessed here a blittable who'se context was disposed (the .CloseTransaction and the TranlateToJS changes are two different fixes to two different problems, that end up "meeting" in that scenario)

src/Raven.Server/Documents/Queries/QueryRunner.cs
garayx: done

src/Raven.Server/Documents/Patch/ScriptRunner.cs
ppekrol: Right. Thanks!

ppekrol:
Please check failing tests.

ayende (approved):

src/Raven.Server/Documents/Queries/QueryRunner.cs
```diff
@@ -148,7 +148,6 @@ public TermsQueryResultServerSide ExecuteGetTermsQuery(string indexName, string
 
                 try
                 {
-                    context.CloseTransaction();
```
garayx: ExecuteGetTermsQuery is executed by using(context.OpenReadTransaction()){} so on i = 0 the tx is already open therefore it will be used on the first iteration and we cannot close it.

PR #7732, opened 2018-10-11
RavenDB-12064

aviv86:
http://issues.hibernatingrhinos.com/issue/RavenDB-12064

### DIFF
--- a/src/Raven.Client/Documents/Linq/RavenQueryProviderProcessor.cs
+++ b/src/Raven.Client/Documents/Linq/RavenQueryProviderProcessor.cs
@@ -66,6 +66,7 @@ namespace Raven.Client.Documents.Linq
         private const string DefaultLoadAlias = "__load";
         private const string DefaultAliasPrefix = "__alias";
         private bool _addedDefaultAlias;
+        private Type _ofType;
 
         private readonly HashSet<string> _aliasKeywords = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
         {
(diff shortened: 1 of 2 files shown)


src/Raven.Client/Documents/Linq/RavenQueryProviderProcessor.cs
```diff
@@ -1467,6 +1468,12 @@ private void VisitQueryableMethodCall(MethodCallExpression expression)
                         var type = expression.Arguments[0].Type.GetGenericArguments()[0];
                         _documentQuery.AddRootType(type);
                     }
+
+                    if (expression.Type.GetTypeInfo().IsGenericType)
+                    {
+                        _ofType = expression.Type.GetGenericArguments()[0];
```
ppekrol, line 1476: What happens when `OfType` is used after the Select?

arekpalinski: I think this implementation works because `AddToFieldsToFetch` is called when we are inside `Select` processing. Although let's add a test for that (to ensure `OfType` used after the `Select` doesn't break anything). Also it might be good idea to skip setting `_ofType` field if  `Select` was already processed

aviv86: a query with `OfType` after `Select`, like this one:
```
var query = session.Query<Result, UsersIndex>()
	.Select(u => u.Friend)  //Friend type is 'User'
	.OfType<Result>();
```
will still be processing `OfType` _before_ processing the `Select`.
so the `_ofType` variable would be set before we reach `AddToFieldToFetch`, which seems a bit wrong.. 
but it will only matter if the `Select` is selecting Id, and I'm not sure that this is a valid use-case. thoughts?

arekpalinski: Do we have any way to figure out if `OfType` was before or after `Select`? We need to use the type specified by `OfType` only if it precedes the projection

aviv86: yes we can figure it out

PR #7741, opened 2018-10-14
RavenDB-12023 - fix top threads usage display in linux

karmeli87:


src/Raven.Server/Program.cs
```diff
@@ -357,7 +357,7 @@ public static void WriteServerStatsAndWaitForEsc(RavenServer server)
                     Thread.Sleep(100);
                 }
             }
-
+            char x = '\x1b';
```
grisha-kotler: this is not used

src/Raven.Server/Program.cs
```diff
@@ -377,12 +377,22 @@ public static void WriteThreadsInfoAndWaitForEsc(RavenServer server, int maxTopT
             var maxNameLength = 0;
             while (Console.KeyAvailable == false)
             {
-                Console.SetCursorPosition(cursorLeft, cursorTop);
+                if (PlatformDetails.RunningOnPosix)
+                {
+                    Console.Out.Write('\x1b');
```
ayende: What is this for?

karmeli87: This tells to the terminal `set cursor position` 
http://www.tldp.org/HOWTO/Bash-Prompt-HOWTO/x361.html

ppekrol (approved):

PR #7742, opened 2018-10-15
RavenDB-12070 NullReferenceException access to notifications before t…

karmeli87:
…hey are properly initialized.

### DIFF
--- a/src/Raven.Server/ServerWide/ServerStore.cs
+++ b/src/Raven.Server/ServerWide/ServerStore.cs
@@ -95,6 +95,7 @@ namespace Raven.Server.ServerWide
         public readonly LicenseManager LicenseManager;
         public readonly FeedbackSender FeedbackSender;
         public readonly SecretProtection Secrets;
+        public readonly AsyncManualResetEvent InitializationCompleted;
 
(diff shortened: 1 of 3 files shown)


src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -95,6 +95,7 @@ public class ServerStore : IDisposable
         public readonly LicenseManager LicenseManager;
         public readonly FeedbackSender FeedbackSender;
         public readonly SecretProtection Secrets;
+        public readonly AsyncManualResetEvent InitializationCompleted;
```
ppekrol, line 98: Probably we should use regular 'bool' instead?

ppekrol: Maybe we should initialize ServerStore before starting the Kestrel?

karmeli87: We can't, because if we bind to port 0, we don't know the real port.

ppekrol (approved):

ayende (approved):

PR #7744, opened 2018-10-15
Fix memory leak (RavenDB-12058)

fabsenet:
This is a fix for the memory leak I discovered in http://issues.hibernatingrhinos.com/issue/RavenDB-12058

The buffer pool try to gives larger data than requested to exchange outer fragmentation with inner fragmentation. The index into the array of _freesegments needs to be updated accordingly. I have done this with a ref to minimize the change I had to done to your code.

The remaining text is in the youtrack issue.

### DIFF
--- a/src/Sparrow/Json/UnmanagedBuffersPool.cs
+++ b/src/Sparrow/Json/UnmanagedBuffersPool.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Concurrent;
 using Sparrow.Binary;
 using Sparrow.Logging;
@@ -133,11 +133,12 @@ namespace Sparrow.Json
                 };
             }
 
+            actualSize = GetIndexSize(ref index, actualSize); // when we request 7 bytes, we want to get 16 bytes
+
             if (_freeSegments[index].TryPop(out AllocatedMemoryData list))
             {
                 return list;
             }
-            actualSize = GetIndexSize(index, actualSize); // when we request 7 bytes, we want to get 16 bytes
             return new AllocatedMemoryData
             {
                 SizeInBytes = actualSize,
@@ -148,7 +149,7 @@ namespace Sparrow.Json
         }
 
 
-        private static int GetIndexSize(int index, int powerBy2Size)
+        private static int GetIndexSize(ref int index, int powerBy2Size)
         {
             switch (index)
             {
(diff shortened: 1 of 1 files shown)


fabsenet:
I would love to add proper unit tests for this if any of you guys point me to the correct spot?

code would be
```cs
var pool = new UnmanagedBuffersPool("HUB");

for(int i=0; i < 1000; i++)
{
    var data = pool.Allocate(2048); //or 1
    pool.Return(data);
}
```

the Assert would be `Assert.IsTrue(pool._freeSegments.All(fs => fs.Count <= 1));` but _freeSegments is private. How do you handle this normally? make it internal?

ppekrol:
@fabsenet Thanks for finding and fixing this.

@redknightlois can you take a look and help @fabsenet with the test? The fix looks valid, because we are modifying the size there, but index is not recalculated. I'm wondering if we should not make the code more universal to avoid issues with future changes inside 'GetIndexFromSize' that will affect the hardcoded numbers 'GetIndexSize' in e.g.
```
var finalSize = GetIndexSize(index, actualSize);
if (finalSize != actualSize)
    index = GetIndexFromSize(finalSize);
```

redknightlois:
@fabsenet excellent writeup and investigation on the issue. I concur this LGTM. I am having issues to test the internals without exposing them too, so it doesn't have an easy solution. The new allocators allow for better inspection through composition. 

@ppekrol the universalization of that code is there: http://issues.hibernatingrhinos.com/issue/RavenDB-8951 :) ... those allocators were completely rewritten and now we distinguish between BlockPointers and Pointers. Where BP have 3 values, the pointer, the size and the actual size.

redknightlois:
@fabsenet As soon as I finish porting the `UnmanagedBuffersPool` to the new allocators I will add that test in there with the new composable allocators to see how difficult it could be.

fabsenet:
this is merged, but is it released? I currently use RavenDB.Client Version 4.1.2 and I am not sure, this is already in it? Can I see the actual SHA for the code used to build 4.1.2?

ayende:
This is in the nightlies, yes.

ppekrol:
@fabsenet All the SHAs are here: https://github.com/ravendb/ravendb/releases

Your PR is included in `4.1.3-patch-41005`: https://github.com/ravendb/ravendb/commit/4938cb86477939be7ff9fe767d399dab8f2db8bb

PR #7758, opened 2018-10-16
RavenDB-9784 & RavenDB-11777

ppekrol:


test/SlowTests/Utils/Analyzers/NGramAnalyzer.cs
```diff
@@ -4,25 +4,43 @@
...
 using Lucene.Net.Analysis.Standard;
-using Lucene.Net.Util;
 using Raven.Server.Documents.Indexes;
+using Raven.Server.Documents.Indexes.Static;
+using Version = Lucene.Net.Util.Version;
 
 namespace SlowTests.Utils.Analyzers
 {
     [NotForQuerying]
     public class NGramAnalyzer : Analyzer
```
ayende: Am I missing something, or is this a test only analyzer?
Shouldn't we be moving this to the server?

ayende (changes requested):

ayende:
Can you also create a full example on how you would create an ngram with specific sizes just for this index (using the new `Index.Configuration` )?

test/SlowTests/Utils/Analyzers/NGramAnalyzer.cs
ppekrol: Yes, forgot to move it!

ppekrol:
Yes, but I will wait for AbstractIndexCreationTask.Configuration PR to be merged, it will be easier then

ppekrol:
@ayende moved. You can also check `RavenDB_11777` I'm overriding default NGram's there via Index configuration

PR #7760, opened 2018-10-16
RavenDB-10106 support for exact in startsWith and endsWith

ppekrol:


src/Raven.Client/Documents/Session/IAbstractDocumentQuery.cs
```diff
@@ -149,16 +149,12 @@ public interface IAbstractDocumentQuery<T>
         /// <summary>
         ///   Matches fields which starts with the specified value.
         /// </summary>
-        /// <param name = "fieldName">Name of the field.</param>
-        /// <param name = "value">The value.</param>
-        void WhereStartsWith(string fieldName, object value);
+        void WhereStartsWith(string fieldName, object value, bool exact = false);
```
ayende, line 152: This is a breaking change. Better to create an overload here.

ayende (changes requested):

src/Raven.Client/Documents/Session/IAbstractDocumentQuery.cs
ppekrol: We guarantee source-level compatibility between minors, so it should be OK?

PR #7764, opened 2018-10-16
RavenDB-12021 Fix a race between disposing a database and creating a …

karmeli87:
…new index on it

It might happened that when the two commands: `PutIndex` and `DeleteDatabase` are accpeted closly on to another,
that the `DeleteDatabase` will complete it's execution first and skip the disposing of the later index.

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -571,6 +571,12 @@ namespace Raven.Server.Documents.Indexes
             if (_initialized == false)
                 throw new InvalidOperationException($"Index '{Name}' was not initialized.");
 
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -571,6 +571,12 @@ public virtual void Start()
             if (_initialized == false)
                 throw new InvalidOperationException($"Index '{Name}' was not initialized.");
 
+            if (DocumentDatabase.IndexStore.IsDisposed.IsRaised())
```
ayende, line 574: What happens if the database disposed _after_ this line?

karmeli87: Then it should be OK, since this index was added to the `CollectionOfIndexes` and will be disposed as usual.

PR #7767, opened 2018-10-16
Fix for RavenDB-12088, additional tests for  RavenDB-12028

myarichuk:
* Properly handle edge cases for SELECT : throw if one or more aliases in SELECT are not defined, properly return result shape the same as document queries (RavenDB-12088)
* Tests for RavenDB-12088
* Additional tests for RavenDB-12028


ayende:
Already merged to https://github.com/ravendb/ravendb/pull/7771

PR #7780, opened 2018-10-17
RavenDB-11804

aviv86:
http://issues.hibernatingrhinos.com/issue/RavenDB-11804
this solution handles wrong alias detection for queries only, 
doesn't handle patches

### DIFF
--- a/src/Raven.Server/Documents/Queries/QueryMetadata.cs
+++ b/src/Raven.Server/Documents/Queries/QueryMetadata.cs
@@ -479,7 +479,15 @@ namespace Raven.Server.Documents.Queries
             if (RootAliasPaths.Count == 0)
                 ThrowMissingAliasOnSelectFunctionBody(parameters);
 
-            VerifyKnownAliasesInScript(parameters);
+            // validate that this is valid JS code
+            try
+            {
+                Query.SelectFunctionBody.Program = ValidateScript(parameters);
+            }
+            catch (Exception e)
+            {
+                throw new InvalidQueryException("Select clause contains invalid script", QueryText, parameters, e);
+            }
 
             var name = "__selectOutput";
             if (Query.DeclaredFunctions != null &&
@@ -1994,12 +2002,12 @@ namespace Raven.Server.Documents.Queries
             "Math", "Number", "Object", "Date"
         };
 
-        private void VerifyKnownAliasesInScript(BlittableJsonReaderObject parameters)
+        private Esprima.Ast.Program ValidateScript(BlittableJsonReaderObject parameters)
         {
             HashSet<string> maybeUnknowns = null;
             Identifier currentProp = null;
 
-            void VerifyKnownAlias(INode node)
+            void VerifyKnownAliases(INode node)
             {
                 switch (node)
                 {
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/Queries/QueryMetadata.cs
```diff
@@ -165,6 +166,14 @@ private void AddExistField(QueryFieldName fieldName, BlittableJsonReaderObject p
             string methodName = null
             )
         {
+            var split = fieldName.Value.Split(".");
+            if (split.Length > 1 && 
+                RootAliasPaths.Count > 0 &&
+                RootAliasPaths.TryGetValue(split[0], out _) == false)
```
arekpalinski: Maybe let's extract:
```
RootAliasPaths.Count > 0 &&
RootAliasPaths.TryGetValue(split[0], out _) == false
```
to a method and use it in all places that repeat this code

src/Raven.Server/Documents/Queries/QueryMetadata.cs
```diff
@@ -1964,6 +1995,68 @@ public string GetUpdateBody(BlittableJsonReaderObject parameters)
...
+                        maybeUnknowns = maybeUnknowns ?? new HashSet<string>();
+                        maybeUnknowns.Add(id.Name);
+                        break;
+                }
+            }
+
+            return new JavaScriptParser("return " + Query.SelectFunctionBody.FunctionText, new ParserOptions(), VerifyKnownAliases).ParseProgram();
+        }
+
+        private bool UnknownIdentifier(string identifier, BlittableJsonReaderObject parameters)
```
arekpalinski: Does it need to be method of `QueryMetadata` class, it's called only from local `VerifyKnownAliases` function? Can it be local function as well?

src/Raven.Server/Documents/Queries/QueryMetadata.cs
```diff
@@ -1964,6 +1995,68 @@ public string GetUpdateBody(BlittableJsonReaderObject parameters)
...
+        {
+            return RootAliasPaths.TryGetValue(identifier, out _) == false &&
+                   _jsBaseObjects.Contains(identifier) == false &&
+                   (parameters == null || 
+                    identifier.StartsWith("$p") == false || 
+                    parameters.TryGet(identifier.Substring(1), out object _) == false);
+        }
+
+        private static void RemoveFromUnknowns(HashSet<string> maybeUnknowns, List<INode> parameters)
```
arekpalinski: Does it need to be method of `QueryMetadata` class, it's called only from local `VerifyKnownAliases` function? Can it be local function as well?

src/Raven.Server/Documents/Queries/QueryMetadata.cs
```diff
@@ -1964,6 +1995,68 @@ public string GetUpdateBody(BlittableJsonReaderObject parameters)
...
+
+            return new JavaScriptParser("return " + Query.SelectFunctionBody.FunctionText, new ParserOptions(), VerifyKnownAliases).ParseProgram();
+        }
+
+        private bool UnknownIdentifier(string identifier, BlittableJsonReaderObject parameters)
+        {
+            return RootAliasPaths.TryGetValue(identifier, out _) == false &&
+                   _jsBaseObjects.Contains(identifier) == false &&
+                   (parameters == null || 
+                    identifier.StartsWith("$p") == false ||
```
arekpalinski: Can we change `"$p"` to `"$"` ? We shouldn't assume that parameters will be always defined using `p` letter e.g. this is a valid query that you can issue from the studio:

```
$company = 'companies/85-A'

from index 'Orders/Totals' where Company  = $company    
```

PR #7799, opened 2018-10-18
RavenDB-11873 New database from v3.x usability issues

ml054:


src/Raven.Client/ServerWide/Operations/Migration/OfflineMigrationConfiguration.cs
```diff
@@ -8,6 +8,10 @@ namespace Raven.Client.ServerWide.Operations.Migration
 {
     public class OfflineMigrationConfiguration
     {
+        private const string StorageExporterExecutable = "Raven.StorageExporter.exe";
+        private const string EsentDataFile = "Data.jfm";
```
ayende: The Esent filename is supposed to be just `Data`, no?

ml054: image

I can see data and data.jfm. I used data.jfm as it is has better selectivity then simply *data*. Anyway I can change that if you wish.

ml054: I just saw conversation on youtrack. I will change this to Data or Data.ravenfs

ml054: Fixed

src/Raven.Server/Web/Studio/StudioTasksHandler.cs
```diff
@@ -56,6 +57,55 @@ public Task FullDataDirectory()
             return Task.CompletedTask;
         }
 
+        [RavenAction("/studio-tasks/offline-migration-test", "GET", AuthorizationStatus.Operator)]
+        public Task OfflineMigrationTest()
+        {
```
ppekrol, line 62: I think we should throw here when server is on non-Windows?

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -1036,13 +1036,10 @@ public async Task MigrateDatabaseOffline()
             }
 
             var dataDir = configuration.DataDirectory;
-            if (Directory.Exists(dataDir) == false)
```
ppekrol, line 1039: This method should also throw for non-Windows server. UI in Studio should be blocked

ppekrol (changes requested):

ml054:
I've blocked this on non-Windows

ml054:
fixed url conventions

PR #7800, opened 2018-10-18
RavenDB-12109 - Traffic watch doesn't display errors as errors in the studio

grisha-kotler:
http://issues.hibernatingrhinos.com/issue/RavenDB-12109

### DIFF
--- a/src/Raven.Server/Routing/RequestRouter.cs
+++ b/src/Raven.Server/Routing/RequestRouter.cs
@@ -51,28 +51,24 @@ namespace Raven.Server.Routing
             return tryMatch.Value;
         }
 
-        public async Task HandlePath(HttpContext context, string method, string path, Reference<string> database)
+        public async ValueTask HandlePath(RequestHandlerContext reqCtx)
         {
-            var tryMatch = _trie.TryMatch(method, path);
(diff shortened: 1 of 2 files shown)


src/Raven.Server/Routing/RequestRouter.cs
```diff
@@ -51,7 +51,7 @@ public RouteInformation GetRoute(string method, string path, out RouteMatch matc
             return tryMatch.Value;
         }
 
-        public async ValueTask<string> HandlePath(HttpContext context, string method, string path)
+        public async Task HandlePath(HttpContext context, string method, string path, Reference<string> database)
```
ayende: Why did you make this change? This is `ValueTask` for a reason.

ayende: We don't need the `database`, pass the `RequestHandlerContext` into this method and set the value as appropriate. That would save an allocation

ayende (changes requested):

src/Raven.Server/Routing/RequestRouter.cs
grisha-kotler: done

ppekrol (approved):

PR #7805, opened 2018-10-18
RavenDB-11739 Optimize Equals of BlittableJsonReaderObject by first d…

arekpalinski:
…oing memcmp and then doing more expensive comparison if it failed only

### DIFF
--- a/src/Sparrow/Json/BlittableJsonReaderObject.cs
+++ b/src/Sparrow/Json/BlittableJsonReaderObject.cs
@@ -1173,12 +1173,16 @@ NotFound:
 
                 if (token != BlittableJsonToken.StartObject && token != BlittableJsonToken.EmbeddedBlittable)
                     continue;
-                
-                var current = (BlittableJsonReaderObject)this[propName];
-                var otherCurrent = (BlittableJsonReaderObject)other[propName];
 
-                if (current.HasSamePropertyNames(otherCurrent) == false)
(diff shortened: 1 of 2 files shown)


src/Sparrow/Json/BlittableJsonReaderObject.cs
```diff
@@ -1099,11 +1099,47 @@ protected bool Equals(BlittableJsonReaderObject other)
...
+                        // data of property values is the same, we also need to check names of properties
+                        // they as are defined in root so it requires separate check
+
+                        return HasSamePropertyNames(other);
+                    }
+                }
+            }
+            else if (_isRoot && other._isRoot)
+            {
+                if (_size != other._size)
```
arekpalinski: No sure if this is always correct assumption. Depending what we decide in http://issues.hibernatingrhinos.com/issue/RavenDB-12122

arekpalinski: According to tests added there we might have the same blittables but with different sizes. I think the issue is only when creating blittables with the usage of `ManualBlittableJsonDocumentBuilder`

arekpalinski: We got some test failures because of that: http://jenkins.hibernatingrhinos.com/job/PR_Tests_v4.1/1830/

I dropped this condition in https://github.com/ravendb/ravendb/pull/7805/commits/e5c6a27eda19b1a3b0b15378538c6947f4ca646d

ayende (approved):

PR #7808, opened 2018-10-19
RavenDB-10851 deploy indexes in one batch

ppekrol:


ppekrol:
@karmeli87 Can you review?

arekpalinski (approved):

ayende (approved):

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -1439,6 +1466,87 @@ public IndexMergeResults ProposeIndexMergeSuggestions()
...
+            }
+
+            public async Task SaveAsync()
+            {
+                if (_command == null || _command.Static.Count == 0 && _command.Auto.Count == 0)
+                    return;
+
+                try
+                {
+                    var (index, _) = await _store._serverStore.SendToLeaderAsync(_command);
```
karmeli87: If an exception occurred during `SendToLeaderAsync` it is probably due to network issues / cluster down etc.
Meaning that the command is not committed to the raft log and wouldn't be executed.

Any later exception will mean that the command is committed but it's execution has failed or wasn't completed in time.

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -1439,6 +1466,87 @@ public IndexMergeResults ProposeIndexMergeSuggestions()
...
+
+                    var indexCount = _command.Static.Count + _command.Auto.Count;
+                    var operationTimeout = _store._serverStore.Engine.OperationTimeout;
+                    var timeout = TimeSpan.FromSeconds((indexCount / 50.0) * operationTimeout.TotalSeconds);
+                    if (operationTimeout > timeout)
+                        timeout = operationTimeout;
+
+                    await _store._documentDatabase.RachisLogIndexNotifications.WaitForIndexNotification(index, timeout);
```
karmeli87: Note that this will ensure the execution only on this specific node and not on the entire database group.

src/Raven.Server/Documents/Indexes/IndexStore.cs
ppekrol: That's fine. We are only catching TimeoutExeception here. Same flow as it was/is in `CreateIndex`

src/Raven.Server/Documents/Indexes/IndexStore.cs
ppekrol: That is enough for us here IMO. Same as in `CreateIndex`.

src/Raven.Server/Documents/Indexes/IndexStore.cs
karmeli87: Okay, but if we passed this line then the explanation for `TimeoutExeception` is incorrect, because the cluster is not down :)

ppekrol: Oh... that is what you meant. I will fix the exeception. Give me few minutes.

ppekrol: Done

PR #7809, opened 2018-10-19
RavenDB-12125

ppekrol:
- during database dispose we should prevent handling of any new state changed notifications and wait for a short period of time for current notification to finish processing
- early exiting IndexStore RecordChanged process when shutdown is requested

### DIFF
--- a/src/Raven.Server/Documents/DocumentDatabase.cs
+++ b/src/Raven.Server/Documents/DocumentDatabase.cs
@@ -57,13 +57,13 @@ namespace Raven.Server.Documents
         private readonly ServerStore _serverStore;
         private readonly Action<string> _addToInitLog;
         private readonly Logger _logger;
-        private DisposeOnce<SingleAttempt> _disposeOnce;
+        private readonly DisposeOnce<SingleAttempt> _disposeOnce;
 
         private readonly CancellationTokenSource _databaseShutdown = new CancellationTokenSource();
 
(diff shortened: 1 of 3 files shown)


ppekrol:
cc @karmeli87

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -416,6 +417,11 @@ private unsafe void DisposeInternal()
 
             var exceptionAggregator = new ExceptionAggregator(_logger, $"Could not dispose {nameof(DocumentDatabase)} {Name}");
 
+            var lockTaken = false;
+            Monitor.TryEnter(_locker, TimeSpan.FromSeconds(5), ref lockTaken);
+            if (lockTaken == false && _logger.IsOperationsEnabled)
```
ayende: Let's wrap this in a `try/finally`, to be safe.

ayende: I know that we are using the aggregator here, but it is possible that other things could throw that aren't expected and it bugs me that the lock may be leaked.

ayende (changes requested):

src/Raven.Server/Documents/DocumentDatabase.cs
ppekrol: Done

PR #7818, opened 2018-10-21
RavenDB-11679 expose ServerSideQuery

talweiss1982:


ppekrol:
I think that this should be handled like other debug options (index entries, explanations, etc):
https://github.com/ravendb/ravendb/blob/5431af3022fd3e7b55b763bc3431a050ce918d6c/src/Raven.Server/Documents/Handlers/QueriesHandler.cs#L52

talweiss1982:
@ayende requested that it will be returned with the query itself.

ppekrol:
Then I think that it would be better to retrieve whole `QueryMetadata` that was used instead?

https://github.com/ravendb/ravendb/blob/v4.1/src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs#L123

talweiss1982:
I don't think that it is what we want since it contains things related to the Jint generated functions

ppekrol:
I do not think we are exposing such things in QueriesDebugHandler and the output could be the same.

src/Raven.Server/Documents/Queries/IndexQueryServerSide.cs
```diff
@@ -45,6 +45,18 @@ public IndexQueryServerSide(QueryMetadata metadata)
...
+        public string ServerSideQuery 
+        {
+            get
+            {
+                if (Metadata.IsGraph)
+                {
+                    return Metadata.Query.GraphQuery.ToString();
+                }
+
+                return Query; //For none graph queries the Query string is identical to what the server will execute
```
ayende: Here I want to get the results of the query visitor, on purpose.
That give us consistent output regardless.

ayende:
The reason I want it in the query is that I want to make it easier for the studio to request it and allow us to demonstrate the query processing on the server side when doing demos.

src/Raven.Server/Documents/Queries/IndexQueryServerSide.cs
talweiss1982: i'm not sure which object you refer to.
Is this the output you want to see:
{DECLARE function __selectOutputfunction __selectOutput(__alias1, o, p, rvnQueryArgs) { 
    return {
    OrderId: id(o),
    Product: p.Name
};
}


WITH {
    FROM Orders WHERE id() = 'orders/828-A'
} AS o
WITH {
    FROM Products
} AS p
WITH...

?

ayende: Yes, I want to strip it of as much magic as ossible.

PR #7824, opened 2018-10-21
RavenDB-12132 - CompareExchange missing the Id property

grisha-kotler:
http://issues.hibernatingrhinos.com/issue/RavenDB-12132

### DIFF
--- a/src/Raven.Client/Documents/Operations/CompareExchange/PutCompareExchangeValueOperation.cs
+++ b/src/Raven.Client/Documents/Operations/CompareExchange/PutCompareExchangeValueOperation.cs
@@ -47,7 +47,7 @@ namespace Raven.Client.Documents.Operations.CompareExchange
                 _key = key;
                 _value = value;
                 _index = index;
-                _conventions = conventions ?? DocumentConventions.Default;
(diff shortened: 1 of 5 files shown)


ayende (approved):

src/Raven.Client/Documents/Operations/CompareExchange/PutCompareExchangeValueOperation.cs
```diff
@@ -47,7 +47,7 @@ public PutCompareExchangeValueCommand(string key, T value, long index, DocumentC
                 _key = key;
                 _value = value;
                 _index = index;
-                _conventions = conventions ?? DocumentConventions.Default;
+                _conventions = DocumentConventions.DefaultConventionsForCompareExchange;
```
ppekrol: I think that this will not take into account the casing (if changed from default) in serializer?

ayende: And since this is a user's default, we need it to.
We can copy the serializer customizations, I think?

grisha-kotler: done

PR #7828, opened 2018-10-21
RavenDB-12024: v4.1 ported to v4.0.

garayx:
Also ported this issue RavenDB-12067.


ayende:
There are the 32 bits stabilization changes

PR #7834, opened 2018-10-22
RavenDB-11680 (WIP) refactoring IntermediateResults so it won't run queries…

talweiss1982:
… until the result are actually requested.

We need this so if we decide to change the actual query during the MATCH phase we won't already pay the price at the WITH phase making any optimization futile.

### DIFF
--- a/src/Raven.Server/Documents/Queries/GraphQueryRunner.IntermediateResults.cs
+++ b/src/Raven.Server/Documents/Queries/GraphQueryRunner.IntermediateResults.cs
@@ -1,14 +1,44 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Raven.Client.Util;
+using Raven.Server.Documents.Queries.AST;
+using Raven.Server.ServerWide;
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/Queries/GraphQueryRunner.IntermediateResults.cs
```diff
@@ -21,19 +51,32 @@ public void Add(Match match)
...
+                    if (_queryPerAlias.TryGetValue(alias, out var query) == false)
+                        return false;
 
-            public void EnsureExists(string alias)
-            {
-                if (MatchesByAlias.TryGetValue(alias, out _) == false)
-                    MatchesByAlias[alias] =  new Dictionary<string, Match>();
+                    _matchesByAlias[alias] = new Dictionary<string, Match>();
+                    //TODO: should we move everything to be async?
+                    var results = AsyncHelpers.RunSync( ()=>_database.QueryRunner.ExecuteQuery(query, _documentsContext, _existingResultEtag, _token));
```
ayende, line 66: This should be an async method, , not `RunSync`

PR #7839, opened 2018-10-22
RavenDB-12133 Allow to add node to a cluster only if both nodes can e…

karmeli87:
…stablish connection to each other

Also the test connection button for the cluster will check both directions.

### DIFF
--- a/src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
@@ -198,6 +198,7 @@ namespace Raven.Server.Documents.Handlers.Admin
                 context.Write(writer, json);
                 writer.Flush();
             }
+
             return Task.CompletedTask;
         }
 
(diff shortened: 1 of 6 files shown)


src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
```diff
@@ -335,6 +336,16 @@ public async Task AddNode()
...
                 requestExecutor.DefaultTimeout = ServerStore.Engine.OperationTimeout;
 
+                // test connection to remote.
+                var result = await ServerStore.TestConnectionToRemote(nodeUrl, database: null);
+                if (result.Success)
+                {
+                    // test connection from remote to destination
+                    result = await ServerStore.TestConnectionFromRemote(requestExecutor, ctx, nodeUrl);
+                    if(result.Success == false)
+                        throw new InvalidOperationException(result.Error);
```
ayende: Here need to indicate that we were able to connect to the remote, and it wasn't able to connect to us

src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
```diff
@@ -335,6 +336,16 @@ public async Task AddNode()
             {
                 requestExecutor.DefaultTimeout = ServerStore.Engine.OperationTimeout;
 
+                // test connection to remote.
+                var result = await ServerStore.TestConnectionToRemote(nodeUrl, database: null);
+                if (result.Success)
```
ayende: where is the error handling on the `Success == false`?

src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
karmeli87: The `result.Error` already contains this info.

src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
karmeli87: fixed

karmeli87:
And this is the exception if the remote was unable to connect the source

```
You are able to reach 'http://127.0.0.1:50633', but he failed to reach you back on http://fake.url:8080.
Please validate the correctness of your 'PublicServerUrl'/'PublicServerUrl.Tcp' configuration and the firewall rules.
Please visit https://ravendb.net/docs/article-page/4.1/csharp/start/installation/setup-wizard#installation--setup-wizard-walkthrough for the RavenDB setup instructions.
```

ayende:
` but he failed` -> `but the *remote node* failed`

PR #7840, opened 2018-10-22
RavenDB-7725

ppekrol:


src/Raven.Server/Documents/Handlers/Debugging/ServerWideDebugInfoPackageHandler.cs
```diff
@@ -208,8 +220,88 @@ public async Task GetInfoPackage()
...
-        private async Task<Stream> GetDebugInfoFromNodeAsync(JsonOperationContext jsonOperationContext,
-            string url, IEnumerable<string> databaseNames, X509Certificate2 certificate)
+        private void DumpStacktraces(ZipArchive archive, string prefix)
+        {
+            var stacktraces = archive.CreateEntry($"{prefix}/stacktraces.json", CompressionLevel.Optimal);
+
+            var tempPath = ServerStore.Configuration.Storage.TempPath?.FullPath ?? Path.GetTempPath();
+
+            var ravenDebugExec = Path.Combine(AppContext.BaseDirectory, "Raven.Debug.exe");
+            var ravenDebugOutput = Path.Combine(tempPath, "stacktraces.json");
```
ayende: Any reason we aren't just pumping this directly to the console as the output of `Raven.Debug`?

tools/Raven.Debug/CommandLineApp.cs
```diff
@@ -0,0 +1,229 @@
...
+        [DllImport("kernel32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
+        private static extern IntPtr LoadLibraryEx(string lpFileName, IntPtr hReservedNull, int dwFlags);
+
+        private static void EnsureProperDebugDllsAreLoaded()
+        {
+            var systemDirectory = Environment.GetFolderPath(Environment.SpecialFolder.System);
+
+            var res = LoadLibraryEx(Path.Combine(systemDirectory, "dbghelp.dll"), IntPtr.Zero, LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR);
+            if (res == IntPtr.Zero)
+                throw new Win32Exception(Marshal.GetLastWin32Error());
```
ayende, line 32: Need to explain here that we couldn't find `dbghelp.dll` and how to get ti

tools/Raven.Debug/CommandLineApp.cs
```diff
@@ -0,0 +1,229 @@
...
+
+            res = LoadLibraryEx(Path.Combine(systemDirectory, "dbgeng.dll"), IntPtr.Zero, LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR);
+            if (res == IntPtr.Zero)
```
ayende, line 35: Same

ayende (changes requested):

src/Raven.Server/Documents/Handlers/Debugging/ServerWideDebugInfoPackageHandler.cs
ppekrol: Was thinking that there was, because it was done like that in 3.5, but... it works fine. Writing the output to the zip entry now.

tools/Raven.Debug/CommandLineApp.cs
ppekrol: It was fixed by us (you) in 3.5. We are taking the DLL from the system dir, and the files comes with the OS so it needs to be there.

ppekrol:
Raven.Debug writes to console after it detaches, so we are good.

PR #7841, opened 2018-10-22
RavenDB-12077

aviv86:
http://issues.hibernatingrhinos.com/issue/RavenDB-12077

### DIFF
--- a/src/Raven.Client/Documents/Session/Operations/QueryOperation.cs
+++ b/src/Raven.Client/Documents/Session/Operations/QueryOperation.cs
@@ -11,6 +11,7 @@ using Raven.Client.Documents.Queries;
 using Raven.Client.Documents.Session.Tokens;
 using Raven.Client.Exceptions.Documents.Indexes;
 using Raven.Client.Extensions;
+using Raven.Client.Util;
 using Sparrow.Json;
 using Sparrow.Json.Parsing;
 using Sparrow.Logging;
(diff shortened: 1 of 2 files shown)


ayende (approved):

src/Raven.Client/Documents/Session/Operations/QueryOperation.cs
```diff
@@ -182,6 +193,32 @@ internal static T Deserialize<T>(string id, BlittableJsonReaderObject document,
...
 
+        private static ConcurrentDictionary<Type, Type> _wrapperTypes;
+        private const string DummyPropertyName = "Result";
+
+        private static T DeserializeInnerArray<T>(BlittableJsonReaderObject document, string fieldToFetch, InMemoryDocumentSessionOperations session,
+            BlittableJsonReaderArray blittableArray)
+        {
+            document.Modifications = new DynamicJsonValue(document)
+            {
+                [DummyPropertyName] = blittableArray
```
arekpalinski, line 204: Can it happen that `document`  will have `"Result"` (`DummyPropertyName`) property already defined?

arekpalinski: I mean, will it work:

```
select { o.Lines, Result = o.Company } 
```

and won't cause property name clash ?

aviv86: yes, it's possible if `fieldToFetch` is `"Result"`. in this case modifications are redundant, but it will still work. we know that the query projected a single field so `document` shouldn't have other properties besides `fieldToFetch` (and `@metadata`)

aviv86: `select { o.Lines, Result = o.Company } ` won't reach this code, it will go to https://github.com/ravendb/ravendb/blob/ad558af8c2a8f8dbd15ea10bcbdc94610c997b96/src/Raven.Client/Documents/Session/Operations/QueryOperation.cs#L182

arekpalinski: OK, thanks for clarification. It might be good idea to add `Debug.Assert` to make sure that's always the case

ppekrol (approved):

PR #7842, opened 2018-10-22
RavenDB-12110: Failed queries data are now logged in requestLatencyDetails and TrafficWatch.

garayx:


src/Raven.Server/Documents/Queries/IndexQueryServerSide.cs
```diff
@@ -37,36 +49,69 @@ public IndexQueryServerSide(string query, BlittableJsonReaderObject queryParamet
...
             if (result.PageSize == 0 && json.TryGet(nameof(PageSize), out int _) == false)
                 result.PageSize = int.MaxValue;
 
             if (string.IsNullOrWhiteSpace(result.Query))
-                throw new InvalidOperationException($"Index query does not contain '{nameof(Query)}' field.");
+            {
+                var errorMessage = $"Index query does not contain '{nameof(Query)}' field.";
+
+                if (TrafficWatchManager.HasRegisteredClients)
+                    AddStringToHttpContext(httpContext, errorMessage, TrafficWatchChangeType.Queries);
```
ayende: Why not simply wrap the whole Query operation in a `try/catch` ?

ayende (changes requested):

src/Raven.Server/Documents/Queries/IndexQueryServerSide.cs
garayx: done

ppekrol: Can you push it?

src/Raven.Server/Documents/Queries/IndexQueryServerSide.cs
```diff
@@ -37,85 +49,132 @@ public IndexQueryServerSide(string query, BlittableJsonReaderObject queryParamet
...
+                    tracker.Query = result.Query;
 
-            if (cache.TryGetMetadata(result, out var metadataHash, out var metadata))
-            {
-                result.Metadata = metadata;
                 return result;
             }
+            catch
+            {
+                errorMessage = errorMessage ?? (result == null ? $"Failed to parse index query : {json}" : $"Failed to parse query: {result.Query}");
```
ppekrol: why can't we take the message from the catched exception? We do not need the 'errorMessage' variable here.

garayx: I want to display the Failed query to tracker and trafficwatch, not the exception it throws, also I treat few scenarios here when and log them to trafficwatch and tracker.

ppekrol: Please ping me on Slack

ppekrol (approved):

PR #7844, opened 2018-10-23
RavenDB-12024-2_v4.1: Added single allocation size limit when running on 32 bits

garayx:


src/Sparrow/Json/ArenaMemoryAllocator.cs
```diff
@@ -198,12 +189,23 @@ private void GrowArena(int requestedSize)
...
-            return Math.Max(Bits.NextPowerOf2(requestedSize) * 3, _initialSize);
+            return ApplyLimit(Math.Max(Bits.NextPowerOf2(requestedSize) * 3, _initialSize));
+
+            int ApplyLimit(int size)
+            {
+                if (SingleAllocationSizeLimit == null)
+                    return size;
+
+                if (SingleAllocationSizeLimit.Value > size)
+                    return Math.Max(Bits.NextPowerOf2(requestedSize), SingleAllocationSizeLimit.Value);
```
ppekrol, line 205: we want to return SingleAllocationSizeLimit.Value here not `Math.Max` probably?

ayende: No, this is by design.
The point here is that if the `requestedSize` is over the single allocation limit, we have to return at least `requestedSize`.
However, we need to user power of 2 to avoid fragmenetation

src/Sparrow/Json/ArenaMemoryAllocator.cs
```diff
@@ -198,12 +189,23 @@ private void GrowArena(int requestedSize)
         private int GetPreferredSize(int requestedSize)
         {
             if (AvoidOverAllocation)
-                return Bits.NextPowerOf2(requestedSize);
+                return ApplyLimit(Bits.NextPowerOf2(requestedSize));
```
ppekrol, line 192: We can remove ApplyLimit from here I think?

src/Sparrow/Json/ArenaMemoryAllocator.cs
ppekrol: Makes sense

src/Sparrow/Json/ArenaMemoryAllocator.cs
ayende: Don't do any harm, and avoid needing to think why it isn't here.
Okay either way

src/Sparrow/Json/ArenaMemoryAllocator.cs
ppekrol: I'm wondering... maybe it is worth setting 'AvoidOverAllocation' to true on 32-bit?

ayende: This isn't a global flag, it is used in specific circumstances, where we know we are done with writing and need the full operation

ayende: full buffer, once.

PR #7847, opened 2018-10-23
RavenDB-11900

aviv86:
http://issues.hibernatingrhinos.com/issue/RavenDB-11900

### DIFF
--- a/src/Raven.Client/Documents/Linq/RavenQueryProviderProcessor.cs
+++ b/src/Raven.Client/Documents/Linq/RavenQueryProviderProcessor.cs
@@ -579,11 +579,22 @@ namespace Raven.Client.Documents.Linq
                 selectPath += ".Length";
             }
 
-            var propertyName = IndexName == null && _collectionName != null
-                ? QueryGenerator.Conventions.FindPropertyNameForDynamicIndex(typeof(T), IndexName, CurrentPath,
-                    selectPath)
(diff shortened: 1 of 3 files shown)


src/Raven.Client/Documents/Linq/RavenQueryProviderProcessor.cs
```diff
@@ -579,11 +579,22 @@ private string GetPropertyName(string selectPath, ExpressionType type)
...
-                : QueryGenerator.Conventions.FindPropertyNameForIndex(typeof(T), IndexName, CurrentPath,
+            string propertyName;
+            if (IndexName == null && _collectionName != null)
+            {
+                propertyName = QueryGenerator.Conventions.FindPropertyNameForDynamicIndex(typeof(T), IndexName, CurrentPath, 
+                    selectPath);
+            }
+            else if (_insideSelect > 0 && QueryGenerator.Conventions.FindProjectedPropertyNameForIndex != null)
+            {
+                propertyName = QueryGenerator.Conventions.FindProjectedPropertyNameForIndex(typeof(T), IndexName, CurrentPath,
```
ppekrol: We need to support returning null from 'FindProjectedPropertyNameForIndex' and then fallback to 'FindPropertyNameForIndex' also

aviv86: fixed

PR #7849, opened 2018-10-23
Raven db 7018

ml054:


src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -885,7 +886,11 @@ protected override int ExecuteCmd(DocumentsOperationContext context)
                             });
                             break;
                         case CommandType.AttachmentCOPY:
-                            var attachmentCopyResult = Database.DocumentsStorage.AttachmentsStorage.CopyAttachment(context, cmd.Id, cmd.Name, cmd.DestinationId, cmd.DestinationName, cmd.ChangeVector);
+                            if (cmd.AttachmentType == 0)
+                            {
```
ppekrol, line 890: add a note here that when 'AttachmentType' is not send, we need to fallback to default which is 'AttachmentType.Document'

ppekrol (approved):

PR #7850, opened 2018-10-23
RavenDB-10225

arekpalinski:


src/Raven.Server/Storage/StorageSpaceMonitor.cs
```diff
@@ -0,0 +1,221 @@
...
+
+        private Dictionary<LowDiskSpace, List<StorageEnvironment>> GetLowSpaceDisks()
+        {
+            var result = new Dictionary<LowDiskSpace, List<StorageEnvironment>>();
+
+            var drives = DriveInfo.GetDrives();
+            
+            foreach (var item in _database.GetAllStoragesEnvironment())
+            {
+                var dataDrive = DiskSpaceChecker.GetFreeDiskSpace(item.Environment.Options.BasePath.FullPath, drives);
```
ayende: Remember that user may also use  junction points, need to verify this for the `Journals` and `Temp` directory as well inside here.

ayende (changes requested):

src/Raven.Server/Storage/StorageSpaceMonitor.cs
arekpalinski: `Temp` is checked below. But `Journals` needs to be checked separately. I'll fix it

ayende: `Temp` is checked only if it is specially declared, but you can also setup junction point there too

arekpalinski: Just to be sure, does it mean that we need to call `GetDiskFreeSpaceEx`:
https://github.com/ravendb/ravendb/blob/v4.1/src/Raven.Server/Utils/DiskSpaceChecker.cs#L14
for `Journals` and `Temp` directories for each environment we have just in case it's a link?

grisha-kotler: GetDiskFreeSpaceEx doesn't provide the free disk space if it's a link.

http://issues.hibernatingrhinos.com/issue/RavenDB-12010

ayende: The docs says it does: https://docs.microsoft.com/en-us/windows/desktop/fileio/symbolic-link-effects-on-file-systems-functions#getdiskfreespaceex

ayende: @arekpalinski And yes, I think we need to check if each of them is a junction.
@grisha-kotler Note that we also need this on Linux

arekpalinski: `GetDiskFreeSpaceEx` seems to be working as expected - http://issues.hibernatingrhinos.com/issue/RavenDB-12010#focus=streamItem-67-31686-0-0

arekpalinski: Always checking temp and journal dirs as well added in: https://github.com/ravendb/ravendb/pull/7850/commits/a6dcded55dee85d7ca721022304352348ffe3a32

Handling of symbolic links and proper drive detection then will be done by @grisha-kotler (RavenDB-12010)

src/Raven.Server/Storage/StorageSpaceMonitor.cs
```diff
@@ -0,0 +1,249 @@
...
+                    }
+                }
+
+                var lowSpaceDisks = GetLowSpaceDisksAndRelevantEnvironments().Disks;
+
+                if (lowSpaceDisks.Count == 0)
+                    return;
+
+                var message =
+                    $"The following disks are running out of space:{Environment.NewLine} {string.Join(Environment.NewLine, lowSpaceDisks)}";
```
ppekrol, line 90: On Linux it would be better to write 'The following mount points...'

arekpalinski: Fixed in https://github.com/ravendb/ravendb/pull/7879

PR #7852, opened 2018-10-23
Merging 4.1 => 4.2

ayende:


ppekrol:
The `SlowTests.Issues.RavenDB_12077.CanProjectSingleCollectionProperty` is failing. Did we forgot to port something or merged incorrectly?

ppekrol:
This is probably caused by https://github.com/ravendb/ravendb/commit/8cd41e0989109fe065cd3ed251489fdd0edd5992
and https://github.com/ravendb/ravendb/blob/v4.1/src/Raven.Client/Documents/Session/Operations/QueryOperation.cs#L212-L217

Didn't have time to look more deeply there.

ppekrol:
@ayende I've pushed more commits here + https://github.com/ravendb/ravendb/pull/7852/commits/d6a4eac9ca8f02e20f70e213c02b902f0168bc80 fixes the deserialization issue, but I'm wondering if this is correct? During deserialization we are not taking into account the modifications?

ayende:
I'm checking into this, but some of the changes I did in `v4.2` is to the way we handle modifications.
We are no longer wiping them completely on each `ReadObject`, that might have had an impact, checking

ayende:
Yes, it looks like that is the issue. Your fix doesn't address the actual issue (which is that `BlittableJosnReader` doesn't handle modifications properly at all).
I fixed the problem and will merge @myarichuk's changes as well and let the build server have a go at this

PR #7854, opened 2018-10-23
RavenDB-12148 Prevent from killing the process if we get any exceptio…

arekpalinski:
…n in indexing thread. Let's ignore disk full and timeout exceptions when cleaning the scratch buffers pool.

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -1138,6 +1138,11 @@ namespace Raven.Server.Documents.Indexes
                         }
                     }
                 }
+                catch (Exception e)
+                {
+                    if (_logger.IsOperationsEnabled)
+                        _logger.Operations($"Unexpected error in '{Name}' index. This should never happen.", e);
+                }
                 finally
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1138,6 +1138,11 @@ protected void ExecuteIndexing()
                         }
                     }
                 }
+                catch (Exception e)
+                {
+                    if (_logger.IsOperationsEnabled)
+                        _logger.Operations($"Unexpected error in '{Name}' index. This should never happen.", e);
```
ayende, line 1144: Given that this kills the indexing thread, this should also be an alert so user can see it.
Better to also (in memory only) mark the index as errored, so queries will fail

arekpalinski: Fixed https://github.com/ravendb/ravendb/pull/7854/commits/d64ed16ec215411f09c57eb88cc2ed0bf4298d3c

ppekrol (approved):

PR #7855, opened 2018-10-23
RavenDB-11841 run user supplied OnCreateDirectory executable

iftahbe:


src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -503,6 +503,12 @@ public void Initialize()
                 options.MaxScratchBufferSize = Configuration.Storage.MaxScratchBufferSize.Value.GetValue(SizeUnit.Bytes);
             options.PrefetchSegmentSize = Configuration.Storage.PrefetchBatchSize.GetValue(SizeUnit.Bytes);
             options.PrefetchResetThreshold = Configuration.Storage.PrefetchResetThreshold.GetValue(SizeUnit.Bytes);
+            options.OnCreateDirectoryExec = Configuration.Storage.OnCreateDirectoryExec;
+            options.OnCreateDirectoryExecArguments = Configuration.Storage.OnCreateDirectoryExecArguments;
+            options.OnCreateDirectoryExecTimeoutInSec = Configuration.Storage.OnCreateDirectoryExecTimeoutInSec.AsTimeSpan;
+            options.DatabaseName = EnvironmentType.System.ToString();
```
ppekrol: use `nameof`

src/Raven.Server/Config/Categories/StorageConfiguration.cs
```diff
@@ -64,5 +64,20 @@ public class StorageConfiguration : ConfigurationCategory
         [ConfigurationEntry("Storage.PrefetchResetThresholdInGb", ConfigurationEntryScope.ServerWideOrPerDatabase)]
         public Size PrefetchResetThreshold { get; set; }
 
+        [Description("A command or executable to run when creating a new directory (storage environment). RavenDB will execute: command [user-arg-1] ... [user-arg-n] <environment-type> <database-name> <path-to-dir>")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Storage.OnCreateDirectory.Exec", ConfigurationEntryScope.ServerWideOnly)]
```
ppekrol: Maybe we want it to be configurable per database also? I do not see a reason why not?

src/Raven.Server/Config/Categories/StorageConfiguration.cs
```diff
@@ -64,5 +64,20 @@ public class StorageConfiguration : ConfigurationCategory
...
+        [Description("The user arguments for the 'Storage.OnCreateDirectory.Exec' command or executable.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Storage.OnCreateDirectory.Exec.Arguments", ConfigurationEntryScope.ServerWideOnly)]
+        public string OnCreateDirectoryExecArguments { get; set; }
+
+        [Description("The number of seconds to wait for the OnCreateDirectory executable to exit. Default: 30 seconds")]
+        [DefaultValue(30)]
+        [TimeUnit(TimeUnit.Seconds)]
+        [ConfigurationEntry("Storage.OnCreateDirectory.Exec.TimeoutInSec", ConfigurationEntryScope.ServerWideOnly)]
+        public TimeSetting OnCreateDirectoryExecTimeoutInSec { get; set; }
```
ppekrol: Drop 'InSec' from the name of the property

src/Voron/StorageEnvironment.cs
```diff
@@ -163,6 +167,83 @@ public StorageEnvironment(StorageEnvironmentOptions options)
...
+            {
+                StartInfo = new ProcessStartInfo
+                {
+                    FileName = _options.OnCreateDirectoryExec,
+                    Arguments = args,
+                    UseShellExecute = false,
+                    RedirectStandardOutput = false,
+                    RedirectStandardError = true,
+                    CreateNoWindow = true,
+                    Verb = "runas"
```
ppekrol: This looks like a Windows-specific? Do we need this?

src/Voron/StorageEnvironment.cs
```diff
@@ -163,6 +167,83 @@ public StorageEnvironment(StorageEnvironmentOptions options)
             }
         }
 
+        private void OnCreateDirectory()
+        {
+
+            var userArgs = _options.OnCreateDirectoryExecArguments ?? string.Empty;
+            var args = $"{userArgs} {_options.Type} {_options.DatabaseName} {_options.BasePath} ";
```
ppekrol: Do we want to pass also other paths here? Not only the base path?

src/Voron/StorageEnvironment.cs
```diff
@@ -163,6 +167,83 @@ public StorageEnvironment(StorageEnvironmentOptions options)
...
+                    $"Unable to execute '{_options.OnCreateDirectoryExec} {args}', waited for {(int)_options.OnCreateDirectoryExecTimeoutInSec.TotalMilliseconds} ms but the process didn't exit. Stderr: {GetStdError()}",
+                    e);
+
+            }
+
+            if (_log.IsOperationsEnabled)
+            {
+                var errors = GetStdError();
+                _log.Operations(string.Format($"Executing '{_options.OnCreateDirectoryExec} {args}' took {sw.ElapsedMilliseconds:#,#;;0} ms"));
+                if (!string.IsNullOrWhiteSpace(errors))
```
ppekrol: I think that this should be moved to next 'if' statement?

src/Voron/StorageEnvironment.cs
```diff
@@ -163,6 +167,83 @@ public StorageEnvironment(StorageEnvironmentOptions options)
...
+            try
+            {
+                process.Start();
+            }
+            catch (Exception e)
+            {
+                throw new InvalidOperationException($"Unable to execute '{_options.OnCreateDirectoryExec} {args}'. Failed to start process.", e);
+            }
+
+            var readErrors = process.StandardError.ReadToEndAsync();
```
ppekrol: Maybe using `process.ErrorDataReceived` would be better (needs to be attached before `process.Start()`) ?

ppekrol (changes requested):

src/Voron/StorageEnvironment.cs
ayende: `_options.BasePath` at least, needs to be quoted

src/Voron/StorageEnvironment.cs
ayende: Also, we want to have a _single_ log entry, not multiple of them

src/Voron/StorageEnvironmentOptions.cs
```diff
@@ -1199,4 +1205,13 @@ public void SetEnvironmentId(Guid environmentId)
             _environmentId = environmentId;
         }
     }
+
+    public enum EnvironmentType
```
ayende: This shouldn't be in Voron, this is a RavenDB concept, not a Voron concept

ayende (changes requested):

src/Voron/StorageEnvironment.cs
ppekrol: For escaping use: https://github.com/ravendb/ravendb/blob/v4.1/src/Sparrow/Utils/CommandLineArgumentEscaper.cs

src/Voron/StorageEnvironmentOptions.cs
ppekrol: Hmm, maybe it would be better, instead of adding all of those fields to Voron, add 'OnBeforeDirectoryCreated' event and attach to it in RavenDB?

ayende: Yes, absolutely.

src/Voron/StorageEnvironment.cs
iftahbe: The problem with this approach is that each error line invokes a new callback asynchronously, and even though the process finishes and my program continues, not all the callbacks have been invoked. So to use this feature I'd have to add synchronization and continue my program only after I get the last error (empty line).
The current way is simpler and I can be sure that when I read the entire stream, all errors will be there for sure, because I waited for the process to exit.

src/Raven.Server/Utils/DirectoryExecUtils.cs
```diff
@@ -0,0 +1,152 @@
...
+            options.OnCreateDirectory += OnDirectory;
+        }
+
+        public static void OnCreateDirectory(StorageEnvironmentOptions options, DirectoryParameters parameters, Logger log)
+        {
+            Process process = null;
+            try
+            {
+                var journalPath = string.Empty;
+                if (options is StorageEnvironmentOptions.DirectoryStorageEnvironmentOptions dirOptions)
```
ayende, line 42: Should we have any special handling for in memory ? We still use directories there.

src/Raven.Server/Utils/DirectoryExecUtils.cs
```diff
@@ -0,0 +1,152 @@
...
+                    journalPath = dirOptions.JournalPath.FullPath;
+
+                var userArgs = parameters.OnCreateDirectoryExecArguments ?? string.Empty;
+                var args = $"{userArgs} {parameters.Type} {parameters.DatabaseName} " +
```
ayende, line 46: Does this handle null paths properly, so you'll get the right order of args?

src/Raven.Server/Utils/DirectoryExecUtils.cs
ppekrol: I think that this is handled in options 'basePath' and 'tempPath' few lines below?

But I'm wondering now if we should not pass the information that this is 'inMemory' as an additional argument?

ppekrol: Actually 'basePath' will have ':memory:' value so probably there is no need to do that.

src/Raven.Server/Utils/DirectoryExecUtils.cs
```diff
@@ -0,0 +1,152 @@
...
+                           $"{CommandLineArgumentEscaper.EscapeSingleArg(options.BasePath.ToString())} " +
+                           $"{CommandLineArgumentEscaper.EscapeSingleArg(options.TempPath.ToString())} " +
+                           $"{CommandLineArgumentEscaper.EscapeSingleArg(journalPath)}";
+
+                process = new Process
+                {
+                    StartInfo = new ProcessStartInfo
+                    {
+                        FileName = parameters.OnCreateDirectoryExec,
```
ppekrol, line 55: We should check if this file exists

ayende: Why? That would just give us a failure to start the process, which is already handled.

src/Raven.Server/Utils/DirectoryExecUtils.cs
ppekrol: 'basePath' and 'tempPath' cannot be null as I recall, so the order should be fine as long as we do not put a new argument after 'journalPath' ?

ayende: Can quoting an empty string work here?

ppekrol: It looks like we can't (at least not in C#):

```
dotnet .\Tryouts.dll aaa "" bbb
Number of arguments: 2
Argument 0: aaa
Argument 1: bbb
```

src/Raven.Server/Utils/DirectoryExecUtils.cs
```diff
@@ -0,0 +1,152 @@
...
+
+                string GetStdError()
+                {
+                    try
+                    {
+                        return readErrors.Result;
+                    }
+                    catch
+                    {
+                        return "Unable to get stderr";
```
arekpalinski, line 86: No need to include the actual exception? Same below

arekpalinski (commented):
Can we have any test for that? Something that would run this code and check we don't have any failure. The script could be  just `echo 'a'`

src/Raven.Server/Utils/DirectoryExecUtils.cs
iftahbe: Yes, Igal will finish his extension point soon and we'll write tests for all four invocations of processes (directory, metrics, master key, certificate)

src/Voron/StorageEnvironment.cs
```diff
@@ -140,6 +140,9 @@ public StorageEnvironment(StorageEnvironmentOptions options)
                 _validPages[_validPages.Length - 1] |= unchecked(((long)ulong.MaxValue << (int)remainingBits));
 
                 _decompressionBuffers = new DecompressionBuffersPool(options);
+
+                options.InvokeOnCreateDirectory();
```
arekpalinski, line 144: Shouldn't we call it _only_ if we create a new database?

https://github.com/ravendb/ravendb/blob/412d9c18257789e0012b0673e50635c345c3056d/src/Voron/StorageEnvironment.cs#L155

iftahbe: The requirement was to call it whenever we create/open a storage environment.

iftahbe: We can also send an extra argument indicating whether it's new or not.

arekpalinski: Yes, argument would be very helpful. I think we really need to allow users to distinguish if _create_ or _open_.

Given when it's called I wonder if `Storage.OnCreateDirectory...` isn't a bit confusing name. Maybe just `Storage.OnInitialize...` or `Storage.OnDirectoryInitialize...`. What do you think?

ayende: Changing the name is better.
And the script can _check_ if the directories already exists, we don't need to provide that.

iftahbe: OK, Storage.OnInitialize sounds good

PR #7867, opened 2018-10-24
RaveDB-12074 - final batch of tests

myarichuk:
also make sure that map/reduce query cannot be the destination in pattern match
(for the pattern (vertexAliasA)-[:edge]->(vertexAliasB) ==> make sure vertexAliasB is not a result of map/reduce query)


src/Voron/Util/UnsafeUtils.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Voron.Util
+{
+    public static unsafe class UnsafeUtils
+    {
+        //TODO : play around with it to make sure it works as I suspect it is
+        public static byte* AddressOf<T>(this T val) where T : struct
```
ayende, line 6: This shouldn't be here

PR #7873, opened 2018-10-25
RavenDB-12010 - Wrong free disk space for symbolic links in the dashboard

grisha-kotler:
http://issues.hibernatingrhinos.com/issue/RavenDB-12010


ppekrol:
@grisha-kotler it looks that your changes are killing the runtime. Can you take a look?

grisha-kotler:
done

ppekrol:
Still runtime is failing. Running tests again.

grisha-kotler:
The issue was that passing a StringBuilder with initial capacity and max capacity to the system call didn't work as expected. The system call used the underlying buffer and when it reached the initial capacity limit, we had AccessViolation/StackOverflow.

src/Sparrow/Utils/DiskSpaceChecker.cs
```diff
@@ -0,0 +1,281 @@
...
+        public static DiskSpaceResult GetDiskSpaceInfo(string pathToCheck, DriveInfoBase driveInfoBase = null)
+        {
+            if (string.IsNullOrEmpty(pathToCheck))
+                return null;
+
+            if (PlatformDetails.RunningOnPosix)
+            {
+                var statvfs = default(Statvfs);
+                if (Syscall.statvfs(pathToCheck, ref statvfs) != 0)
+                    return null;
```
ayende, line 35: Should we log here as well?

src/Sparrow/Utils/DiskSpaceChecker.cs
```diff
@@ -0,0 +1,281 @@
...
+                }
+
+                return path;
+            }
+
+            try
+            {
+                bool GetPath(uint bufferSize, out string outputPath)
+                {
+                    var charArray = new char[bufferSize];
```
ayende, line 185: Can we use `ArrayPool.Rent` ? This is called quite frequently, as I recall

grisha-kotler: It's called only once for each storage environment.
The drive/mount point won't change while it's open.

ppekrol (approved):

src/Sparrow/Utils/DiskSpaceChecker.cs
grisha-kotler: done

ppekrol:
Can you apply this to v4.1?

PR #7876, opened 2018-10-25
RavenDB-12172 add import option to skip server certificate validation

lahma:
With these changes I was able to import from 3.5 server that was using certificate that was not trusted.


ayende (approved):

ayende:
@ml054 Can you check the studio changes?

ml054 (commented):
LGTM

PR #7880, opened 2018-10-26
Recursive graph queries

ayende:
A whole bunch of work around supporting queries like:

```
match (s:Dogs (id()='dogs/arava'))-recursive as friends (1,3) { [:Likes]->(dog:Dogs) }
select s.Name, friends.dog.Name as friend
```


ayende:
This is on me, I used the wrong email when I edited @myarichuk 's commit

PR #7881, opened 2018-10-26
RavenDB-12151

arekpalinski:


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -2940,6 +2985,24 @@ private int MinBatchSize
                 }
             }
 
+            if (ScratchSpaceLimitInBytes != null && txAllocations > ScratchSpaceLimitInBytes)
+            {
+                stats.RecordMapCompletedReason($"Reached scratch space limit ({ScratchSpaceLimitInBytes / 1024:#,#0} kb). Allocated {txAllocations / 1024:#,#0} kb of scratch space in current transaction");
```
grisha-kotler: Can you use size here?

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -2940,6 +2985,24 @@ private int MinBatchSize
...
+                return false;
+            }
+
+            var globalIndexingScratchSpaceUsage = DocumentDatabase.ServerStore.GlobalIndexingScratchSpaceMonitor;
+
+            if (globalIndexingScratchSpaceUsage?.IsLimitExceeded == true && count > MinBatchSize)
+            {
+                _globalIndexingScratchSpaceLimitExceeded = true;
+
+                stats.RecordMapCompletedReason(
```
grisha-kotler, line 3017: Same here

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -2940,6 +2985,24 @@ private int MinBatchSize
...
+                    $"Reached global scratch space limit for indexing ({globalIndexingScratchSpaceUsage.LimitInBytes / 1024:#,#0} kb). Current scratch space is {globalIndexingScratchSpaceUsage.ScratchSpaceInBytes / 1024:#,#0} kb");
```
grisha-kotler: Same here

src/Raven.Server/Documents/Indexes/Index.cs
arekpalinski: Sure. I did it this way for consistency with existing messages. Will fix it since I saw you changed it it another PR

ppekrol (approved):

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
```diff
@@ -123,6 +124,27 @@ public virtual bool RunInMemory
         [ConfigurationEntry("Indexing.CleanupIntervalInMin", ConfigurationEntryScope.ServerWideOnly)]
         public TimeSetting CleanupInterval { get; set; }
 
+        [Description("Maximum amount of scratch space that we allow to use for an indexing batch")]
+        [DefaultValue(null)]
+        [SizeUnit(SizeUnit.Megabytes)]
+        [IndexUpdateType(IndexUpdateType.Refresh)]
+        [ConfigurationEntry("Indexing.ScratchSpaceLimitInMb", ConfigurationEntryScope.ServerWideOrPerDatabase)]
```
ayende, line 138: can this be defined per index?

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
```diff
@@ -123,6 +124,27 @@ public virtual bool RunInMemory
...
+
+        [Description("Maximum amount of scratch space that we allow to use for entire indexing work per server")]
+        [DefaultValue(null)]
+        [SizeUnit(SizeUnit.Megabytes)]
+        [IndexUpdateType(IndexUpdateType.Refresh)]
+        [ConfigurationEntry("Indexing.GlobalScratchSpaceLimitInMb", ConfigurationEntryScope.ServerWideOnly)]
+        public Size? GlobalScratchSpaceLimit { get; protected set; }
+
+        [Description("Max time to wait when forcing the environment flush and sync after exceeding global scratch buffer limit")]
+        [DefaultValue(5)]
```
ayende: This should probably be a higher value by default.
We are in a bad place if this is hit, so we want to give the entire system more time to process things one at a time

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -2940,6 +2985,24 @@ private int MinBatchSize
                 }
             }
 
+            if (ScratchSpaceLimit != null && txAllocations > ScratchSpaceLimit.Value.GetValue(SizeUnit.Bytes))
+            {
+                stats.RecordMapCompletedReason($"Reached scratch space limit ({ScratchSpaceLimit.Value}). Allocated {new Size(txAllocations, SizeUnit.Bytes)} of scratch space in current transaction");
+                return false;
+            }
+
```
ayende, line 3010: I think we are also missing another limit here. The total amount of scratch page used in this index.
The `txAllocations` is just for the current batch, if we didn't flush yet from the scratch files, we may be gathering a lot of them, not on a single batch basis, but across many of them.
We need a limit on the amount of scratch files we have

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
arekpalinski: Yes. It can be defined per index as setting any other index specific configuration using `IndexConfiguration` class in client API via `IndexDefinition.Configuration`. It's also exposed in the studio.

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
arekpalinski: 15 seconds would be enough as the default?

src/Raven.Server/Documents/Indexes/Index.cs
arekpalinski: Good point. I'll rename this config to something that will state it's about max transaction size per _batch_. 

Then current `Indexing.ScratchSpaceLimitInMb` will track the scratch space using the introduced `IScratchSpaceMonitor` but on index environment level.

 If we hit this limit then I assume we should behave the same like hitting `Indexing.GlobalScratchSpaceLimitInMb` - flush and force sync, right?

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
ayende: Lets go with 30 or so, just to be safer

src/Raven.Server/Documents/Indexes/Index.cs
ayende: Yes

arekpalinski: Fixed in https://github.com/ravendb/ravendb/pull/7881/commits/c017182f974b2fe0f6eff87f7549b7af16caf381

ppekrol:
LGTM. @ayende ?

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1120,6 +1125,21 @@ protected void ExecuteIndexing()
                                     new Size(NativeMemory.CurrentThreadStats.TotalAllocated, SizeUnit.Bytes));
                             }
 
+                            if (_scratchSpaceLimitExceeded)
+                            {
+                                if (_logger.IsInfoEnabled)
+                                    _logger.Info($"Indexing exceeded global limit for scratch space usage. Going to flush environment of '{Name}' index and forcing sync of data file");
+
+                                GlobalFlushingBehavior.GlobalFlusher.Value.MaybeFlushEnvironment(storageEnvironment);
+                                GlobalFlushingBehavior.GlobalFlusher.Value.ForceSyncEnvironment(storageEnvironment);
```
ayende, line 1134: Don't we need to wait until the flushing is done before doing the sync?

arekpalinski: `MaybeFlushEnvironment` and `ForceSyncEnvironment` just enqueue environments for relevant operations The global flusher processes the flush operations first, next it do the syncs. So we're guaranteed that sync will be processed after the flush.

ayende: Okay, then LGTM

PR #7883, opened 2018-10-26
RavenDB-8355 support for custom sorters

ppekrol:


test/SlowTests/Issues/RavenDB_8355.cs
```diff
@@ -0,0 +1,392 @@
...
+using Raven.Server.Documents.Queries;
+using Sparrow.Json;
+using Xunit;
+
+namespace SlowTests.Issues
+{
+    public class RavenDB_8355 : RavenTestBase
+    {
+        private const string SorterCode = @"
+using System;
```
ayende: Can probably put these in files and then use them as embedded resources.
Would probably be more readable

ppekrol: Done

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -74,11 +75,26 @@ public void HandleDatabaseRecordChange(DatabaseRecord record, long index)
             if (record == null)
                 return;
 
+            HandleSorters(record, index);
```
maximburyak, line 78: I'm not sure it should be "here", it probably should be treated in the database level.

ppekrol: Wasn't sure either. In general this is related to Indexes/Indexing/Querying, so I thought that it could be handled in one place.

PR #7885, opened 2018-10-26
RavenDB-12194 Fixing JsonNetBlittableEntitySerializer - _customize fi…

arekpalinski:
…eld needs to be per thread in which the serializer was created so we'll properly detect that we need to customize it

### DIFF
--- a/src/Raven.Client/Documents/Conventions/JsonNetBlittableEntitySerializer.cs
+++ b/src/Raven.Client/Documents/Conventions/JsonNetBlittableEntitySerializer.cs
@@ -15,8 +15,9 @@ namespace Raven.Client.Documents.Conventions
         private static BlittableJsonReader _reader;
         [ThreadStatic]
         private static JsonSerializer _serializer;
+        [ThreadStatic]
+        private static Action<JsonSerializer> _customize;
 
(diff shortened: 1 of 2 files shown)


ppekrol (approved):

src/Raven.Client/Documents/Conventions/JsonNetBlittableEntitySerializer.cs
```diff
@@ -15,8 +15,9 @@ internal class JsonNetBlittableEntitySerializer
         private static BlittableJsonReader _reader;
         [ThreadStatic]
         private static JsonSerializer _serializer;
+        [ThreadStatic]
+        private static Action<JsonSerializer> _customize;
```
ayende, line 19: This is wrong, we can't use `ThreadStatic` here, we are better off using `ThreadLocal`, because we may have mutiple different serializers

arekpalinski: Fixed in https://github.com/ravendb/ravendb/pull/7888 / https://github.com/ravendb/ravendb/pull/7889

PR #7890, opened 2018-10-27
RavenDB-12143 The recyclable journals need to be counted as temporary…

arekpalinski:
… buffers on the actual drive where journals are stored

### DIFF
--- a/src/Voron/Debugging/StorageReportGenerator.cs
+++ b/src/Voron/Debugging/StorageReportGenerator.cs
@@ -148,7 +148,8 @@ namespace Voron.Debugging
                 return new TempBufferReport
                 {
                     Name = file.Name,
-                    AllocatedSpaceInBytes = file.Length
+                    AllocatedSpaceInBytes = file.Length,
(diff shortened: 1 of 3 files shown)


src/Voron/Debugging/StorageReportGenerator.cs
```diff
@@ -161,7 +162,8 @@ private List<TempBufferReport> GenerateTempBuffersReport(VoronPathSetting tempPa
                     return new TempBufferReport
                     {
                         Name = file.Name,
-                        AllocatedSpaceInBytes = file.Length
+                        AllocatedSpaceInBytes = file.Length,
+                        Type = TempBufferType.RecyclableJournal
```
ppekrol, line 166: Do we want to display this in Storage Report in Studio differently maybe?

arekpalinski: We could mark it's in `Journals` dir. I'll reopen the ticket for studio adjustments

PR #7893, opened 2018-10-27
Merging 4.1 changes

ayende:


ppekrol:
@ayende it looks like this merge is done against old v4.2. We should close it and do it again using latest ravendb/v4.2 code

PR #7894, opened 2018-10-28

aviviadi:

### DIFF
--- a/src/Voron/Platform/Posix/PosixAbstractPager.cs
+++ b/src/Voron/Platform/Posix/PosixAbstractPager.cs
@@ -53,18 +53,16 @@ namespace Voron.Platform.Posix
 
         protected PosixAbstractPager(StorageEnvironmentOptions options, bool canPrefetchAhead, bool usePageProtection = false) : base(options, canPrefetchAhead, usePageProtection: usePageProtection)
         {
+            _supportUnmapping = true;
         }
 
-        protected unsafe void ReleaseAllocationInfoWithoutUnmapping(byte* baseAddress, long size)
-        {
-            // should be called from Posix32BitsMemoryMapPager in order to bypass unmapping
-            base.ReleaseAllocationInfo(baseAddress, size);
-        }
-        
         public override unsafe void ReleaseAllocationInfo(byte* baseAddress, long size)
         {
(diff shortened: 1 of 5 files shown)


src/Voron/Platform/Posix/PosixAbstractPager.cs
```diff
@@ -47,5 +49,24 @@ protected internal override unsafe void PrefetchRanges(Win32MemoryMapNativeMetho
...
+        
+        protected override void DisposeInternal()
+        {
+            if (_fd != -1)
+            {
+                // note that the orders of operations is important here, we first unlink the file
+                // we are supposed to be the only one using it, so Linux would be ready to delete it
+                // and hopefully when we close it, won't waste any time trying to sync the memory state
+                // to disk just to discard it
+                if (DeleteOnClose)
```
ayende, line 94: Need `madvise` here

aviviadi: madvise should be used in the unmapping events, and maybe we can just fadvise instead : fadvise(fd, DONTNEED) ?

aviviadi: okey, irrelevant. fadvise is not good here. 
We will:
if (DeleteOnClose == true)
  madvise(addr, DONTNEED);
munmap(addr);

src/Voron/Platform/Posix/PosixAbstractPager.cs
```diff
@@ -47,5 +54,53 @@ protected internal override unsafe void PrefetchRanges(Win32MemoryMapNativeMetho
...
+        {
+            // 32 bits should override this method and call AbstractPager::ReleaseAllocationInfo
+            base.ReleaseAllocationInfo(baseAddress, size);
+            var ptr = new IntPtr(baseAddress);
+
+            if (DeleteOnClose)
+            {
+                if (Syscall.madvise(ptr, new UIntPtr((ulong)size), MAdvFlags.MADV_DONTNEED) != 0)
+                {
+                    _log.Info($"Failed to madvise MDV_DONTNEED for {FileName?.FullPath}");
```
grisha-kotler: missing IsInfoEnabled

PR #7895, opened 2018-10-28

aviviadi:


aviviadi:
Conflicts issues... easier to create a new PR for it

PR #7897, opened 2018-10-28
RavenDB-11680 WIP Graph Query Plan for review

talweiss1982:


src/Raven.Server/Documents/Queries/Graph/EdgeQueryStep.cs
```diff
@@ -0,0 +1,143 @@
...
+        public EdgeQueryStep(IQueryStep left, IQueryStep right, WithEdgesExpression edgesExpression, MatchPath edgePath, BlittableJsonReaderObject queryParameters)
+        {
+            _left = left;
+            _right = right;
+            _edgePath = edgePath;
+            _queryParameters = queryParameters;
+            _edgesExpression = edgesExpression;
+        }
+
+        public async ValueTask<IEnumerable<Match>> Execute(Dictionary<IQueryStep, IEnumerable<Match>> matches)
```
ayende:
```suggestion
        public ValueTask<IEnumerable<Match>> Execute(Dictionary<IQueryStep, IEnumerable<Match>> matches)
```

You'll typically want _either_ `async` or `ValueTask`, they don't generally mix.

src/Raven.Server/Documents/Queries/Graph/EdgeQueryStep.cs
```diff
@@ -0,0 +1,143 @@
...
+        {
+            var leftResult = await _left.Execute(matches);
+            var rightResult = await _right.Execute(matches);
```
ayende, line 27: Here you need to check if they are completed or wait.

src/Raven.Server/Documents/Queries/Graph/EdgeQueryStep.cs
```diff
@@ -0,0 +1,143 @@
...
+        }
+
+        private IQueryStep _left;
+        private IQueryStep _right;
+        private MatchPath _edgePath;
+        private readonly BlittableJsonReaderObject _queryParameters;
+        private WithEdgesExpression _edgesExpression;
+        private Dictionary<string, BlittableJsonReaderObject> _included = new Dictionary<string, BlittableJsonReaderObject>();
+
+        public IEnumerable<IQueryStep> Dependencies => GetDependencies();
```
ayende: Why do you care about this?

src/Raven.Server/Documents/Queries/Graph/GraphQueryPlan.cs
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        private IQueryStep BuildQueryPlanForMatchVertex(MatchPath vertex)
+        {            
+            Sparrow.StringSegment alias = vertex.Alias;
+            if (GraphQuery.WithDocumentQueries.TryGetValue(alias, out var query) == false)
+            {
+                throw new InvalidOperationException($"BuildQueryPlanForMatchVertex was invoked for allias='{alias}' which suppose to be a vertex but no corresponding WITH clause was found.");
+            }
+            //TODO: we can tell at this point if it is a collection query or not, we might want to build a diffrent step for collection queries in the future.
```
ayende, line 99: We want to have:

* Index step
* Collection source step
* Collection dest step
* Collection dest step + filter

src/Raven.Server/Documents/Queries/Graph/GraphQueryPlan.cs
```diff
@@ -0,0 +1,98 @@
...
+            return new QueryQueryStep(_database.QueryRunner, alias, query, queryMetadata, _context, _resultEtag, _token);
+        }
+
+        public void OptimizeQueryPlan()
+        {
+            //TODO: identify pattrens for optimization
+            return;
+        }
+
+        public async ValueTask<IEnumerable<Match>> ExecuteQueryPlan()
```
ayende, line 110: async & value task

src/Raven.Server/Documents/Queries/Graph/IQueryPlan.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+using static Raven.Server.Documents.Queries.GraphQueryRunner;
+
+namespace Raven.Server.Documents.Queries.Graph
+{
+    public interface IQueryPlan
```
ayende: Why do we need the `QueryPlan` to be an interface?

src/Raven.Server/Documents/Queries/Graph/EdgeQueryStep.cs
talweiss1982: At first, the query plan was executing the steps but I figured it makes no sense.
I left it thinking that the optimization might care about it but it is probably redundant and will be removed.

src/Raven.Server/Documents/Queries/Graph/GraphQueryPlan.cs
talweiss1982: The question is would we want to go complete OOP and have those as subclasses or would an enum be enough since they will all require pretty much the same parameters and logic to run.

src/Raven.Server/Documents/Queries/Graph/IQueryPlan.cs
talweiss1982: Nope

src/Raven.Server/Documents/Queries/Graph/EdgeQueryStep.cs
talweiss1982: just to be clear i got you right, you want to remove the async method and replace it with.Wait()?

talweiss1982: the task themselves will never be completed at this point since the execution is reccursive in a post order manner.

ppekrol:
@talweiss1982 can you fix compilation?

talweiss1982:
@ppekrol This was only sent as a PR so Oren could review and pull it.
This is already obsolete, I'm going to close it.

PR #7901, opened 2018-10-29

aviviadi:

### DIFF
--- a/src/Voron/Platform/Posix/PosixAbstractPager.cs
+++ b/src/Voron/Platform/Posix/PosixAbstractPager.cs
@@ -113,18 +113,16 @@ namespace Voron.Platform.Posix
 
         protected PosixAbstractPager(StorageEnvironmentOptions options, bool usePageProtection = false) : base(options, usePageProtection: usePageProtection)
         {
+            _supportUnmapping = true;
         }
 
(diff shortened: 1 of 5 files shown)


src/Voron/Platform/Posix/PosixAbstractPager.cs
```diff
@@ -105,5 +112,51 @@ public override unsafe void MaybePrefetchMemory(List<long> pagesToPrefetch)
         protected PosixAbstractPager(StorageEnvironmentOptions options, bool usePageProtection = false) : base(options, usePageProtection: usePageProtection)
         {
         }
+
+        protected unsafe void ReleaseAllocationInfoWithoutUnmapping(byte* baseAddress, long size)
```
ppekrol: I think it would be much better to introduce field 'SupportsUnmapping' and set it in 32-bit ctor to false and use it in 'ReleaseAllocationInfo'.

ayende: I like it, yes

aviviadi: will be done.  please add your comments at https://github.com/ravendb/ravendb/pull/7894 as this v4.0 branch will be overridden by v4.1 cherry-picks

PR #7928, opened 2018-10-30
Moving graph queries to use a query plan

ayende:


ayende:
*Wow* that broke ton of stuff. Looking into it

ppekrol:
@ayende this can be merged?

PR #7930, opened 2018-10-31
RavenDB-12222 Pass cancellation token to the index compaction operati…

arekpalinski:
…on so we can abort the database compaction


ppekrol:
It looks like the 2nd commits is invalid, because we are unloading the database in compaction?

arekpalinski:
I've amended last commit

PR #7934, opened 2018-10-31
RavenDB-11679 Added ServerSideQuery as a query debug parameter so we …

talweiss1982:
…could extract the ServerSideQuery without the query

### DIFF
--- a/src/Raven.Server/Documents/Handlers/QueriesHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/QueriesHandler.cs
@@ -216,6 +216,21 @@ namespace Raven.Server.Documents.Handlers
             }
         }
 
+        private async Task ServerSideQuery(DocumentsOperationContext context, RequestTimeTracker tracker, HttpMethod method)
+        {
+            var indexQuery = await GetIndexQuery(context, method, tracker);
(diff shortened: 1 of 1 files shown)


ayende (approved):

src/Raven.Server/Documents/Handlers/QueriesHandler.cs
```diff
@@ -216,6 +216,21 @@ private async Task Explain(DocumentsOperationContext context, RequestTimeTracker
             }
         }
 
+        private async Task ServerSideQuery(DocumentsOperationContext context, RequestTimeTracker tracker, HttpMethod method)
+        {
+            var indexQuery = await GetIndexQuery(context, method, tracker);
+            using (var writer = new AsyncBlittableJsonTextWriter(context, ResponseBodyStream(), Database.DatabaseShutdown))
+            {
+                writer.WriteStartObject();
+                writer.WriteComma();
```
ppekrol: This will generate invalid JSON. Can you remove the comma?

talweiss1982: tested this on fiddler didn't notice i accidentally copied the comma line

PR #7941, opened 2018-11-01
Graph endpoints and better implicit aliases

ayende:


ppekrol:
Rebased here: https://github.com/ravendb/ravendb/pull/7967

PR #7943, opened 2018-11-01
RavenDB-12232 - retry opening the connection in case of a failure

grisha-kotler:


src/Raven.Server/Documents/ETL/Providers/SQL/RelationalWriters/RelationalDatabaseWriter.cs
```diff
@@ -47,37 +47,59 @@ public RelationalDatabaseWriter(SqlEtl etl, DocumentDatabase database)
...
+                        continue;
+                    }
 
-            _stringParserList = GenerateStringParsers();
+                    using (_connection)
+                    {
+                        database.NotificationCenter.Add(AlertRaised.Create(
+                            database.Name,
+                            SqlEtl.SqlEtlTag,
+                            $"SQL ETL could not open connection to {_connection.ConnectionString}",
```
arekpalinski, line 85: Maybe let's put here just the name of the connection string as it was defined in RavenDB instead of the actual connection string containing login / pass

arekpalinski: Changed in https://github.com/ravendb/ravendb/pull/7947

PR #7945, opened 2018-11-02
RavenDB-12237 - Memory Calculations fixes for indexes

grisha-kotler:
http://issues.hibernatingrhinos.com/issue/RavenDB-12237

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -1498,9 +1498,9 @@ namespace Raven.Server.Documents.Indexes
 
                     try
                     {
+                        var disableTxCompression = false;
                         using (InitializeIndexingWork(indexContext))
                         {
-
                             foreach (var work in _indexWorkers)
                             {
                                 using (var scope = stats.For(work.Name))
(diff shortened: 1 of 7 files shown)


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1565,6 +1574,23 @@ public bool DoIndexingWork(IndexingStatsScope stats, CancellationToken cancellat
...
+                _environment.Options.CompressTxAboveSizeInBytes = long.MaxValue;
+            }
+
+            return new DisposableAction(() =>
+            {
+                if (disableCompression)
+                {
+                    _environment.Options.CompressTxAboveSizeInBytes = prevValue;
+                }
+            });
```
arekpalinski: I wonder if wouldn't we be better off by controlling this at transaction level (or even `Commit` method) instead of dealing with storage environment options

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1515,16 +1515,25 @@ public bool DoIndexingWork(IndexingStatsScope stats, CancellationToken cancellat
...
                                 using (var indexWriteOperation = writeOperation.Value)
                                 {
                                     indexWriteOperation.Commit(stats);
                                 }
+
+                                var modifiedAfterCommit = GetNumberOfModifiedPages(indexContext);
+                                var modifiedByIndexMerge = modifiedAfterCommit - modifiedBeforeCommit;
+                                // compress the tx only if the more than 80% of modifications were made
+                                // before the index commit. we want to avoid compressing lucenes merged segments.
+                                disableTxCompression = (modifiedByIndexMerge / (double)modifiedAfterCommit) >= 0.2;
```
arekpalinski: How did you determine that 80% threshold detects that case? I wonder if we could provide own merge scheduler based on `SerialMergeScheduler` here:

https://github.com/ravendb/ravendb/blob/942f8803f848fe68d0f29b61995cd44c72b134bc/src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexWriter.cs#L114

and be aware if the merge was actually made (maybe there is no need for that)

src/Raven.Server/Documents/Indexes/Index.cs
grisha-kotler: Removed

ayende:
Am I missing something, I don't see any behavior related to the actual tx compression handling?

grisha-kotler:
Renamed the PR

ayende:
LGTM

PR #7954, opened 2018-11-02
RavenDB-12240 map-reduce does not map documents properly when document ID contains culture-specific characters

ppekrol:


src/Raven.Server/Documents/Indexes/MapReduce/MapReduceIndexBase.cs
```diff
@@ -53,7 +53,7 @@ public override IDisposable InitializeIndexingWork(TransactionOperationContext i
         public override unsafe void HandleDelete(Tombstone tombstone, string collection, IndexWriteOperation writer,
             TransactionOperationContext indexContext, IndexingStatsScope stats)
         {
-            using (Slice.External(indexContext.Allocator, tombstone.LowerId.Buffer, tombstone.LowerId.Length, out Slice docKeyAsSlice))
+            using (Slice.External(indexContext.Allocator, tombstone.LowerId, out Slice docKeyAsSlice))
```
ayende, line 56: Why did this change fix anything?

ppekrol: Instead of `LazyStringValue.Length` we should use `LazyStringValue.Size` if not, we were creating Slice with incorrect size, so for similar Ids last characters were lost which resulted in overriding results in Map-Reduce.

ayende: Okay, I see. The difference between bytes & chars

PR #7960, opened 2018-11-04
 RavenDB-12161 Extension point to override OS metrics

haludi:
http://issues.hibernatingrhinos.com/issue/RavenDB-12161


src/Raven.Server/Config/Categories/MonitoringConfiguration.cs
```diff
@@ -1,12 +1,23 @@
...
 namespace Raven.Server.Config.Categories
 {
     public class MonitoringConfiguration : ConfigurationCategory
     {
+        [Description("A command or executable to run which will provide machine cpu usage and total machine cores to standard output. If specified, RavenDB will use this information for monitoring cpu usage.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Monitoring.Cpu.Exec", ConfigurationEntryScope.ServerWideOnly)]
+        public string CpuUsageMonitorExec { get; set; }
+
+        [Description("The command line arguments for the 'Security.MasterKey.Exec' command or executable.")]
```
iftahbe: forgot to change 'Security.MasterKey.Exec'

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
```diff
@@ -0,0 +1,263 @@
...
+                Dispose();
+            }
+
+            var _ = ReadProcess(); // explicitly starting async task without waiting for it
+            _ = ReadErrors();
+        }
+
+        private async Task ReadErrors()
+        {
+            var errors = await _process.StandardError.ReadLineAsync();
```
iftahbe: This will read only one line of the error.

src/Raven.Server/Utils/CpuUsage.cs
```diff
@@ -77,7 +94,7 @@ private static double CalculateProcessCpuUsage(ProcessInfo currentInfo, ProcessI
                 return _lastCpuInfo?.ProcessCpuUsage ?? 0;
             }
 
-            if (currentInfo.ActiveCores == 0)
+            if (Math.Abs(currentInfo.ActiveCores) < 0.01)
```
grisha-kotler: You need to skip those checks when using the CpuUsageExtensionPoint

src/Raven.Server/Utils/CpuUsage.cs
```diff
@@ -87,7 +104,7 @@ private static double CalculateProcessCpuUsage(ProcessInfo currentInfo, ProcessI
             }
 
             var processCpuUsage = (processorTimeDiff * 100.0) / timeDiff / currentInfo.ActiveCores;
-            if (currentInfo.ActiveCores == ProcessorInfo.ProcessorCount)
+            if ((int)currentInfo.ActiveCores == ProcessorInfo.ProcessorCount)
```
grisha-kotler: same here

src/Raven.Server/Utils/CpuUsage.cs
```diff
@@ -18,11 +18,24 @@ public static class CpuUsage
...
         private static readonly (double MachineCpuUsage, double ProcessCpuUsage) EmptyCpuUsage = (0, 0);
+        private static CpuUsageExtensionPoint _cpuUsageExtensionPoint;
+
+
+        public static CpuUsageExtensionPoint CpuUsageExtensionPoint
+        {
+            get => _cpuUsageExtensionPoint;
+            set
+            {
+                _previousExtensionPointInfo = new ExtensionPointInfo(value.Data);
```
maximburyak: What do we do with the previous value, if we have one, and if we don't, then it should be just a Lazy value?

src/Raven.Server/Utils/CpuUsage.cs
```diff
@@ -96,6 +113,24 @@ private static double CalculateProcessCpuUsage(ProcessInfo currentInfo, ProcessI
...
+            var extensionPoint = new ExtensionPointInfo(CpuUsageExtensionPoint.Data);
+            var machineCpuUsage = extensionPoint.MachineCpuUsage;
+            if (machineCpuUsage < 0)
+            {
+                if (Logger.IsOperationsEnabled)
+                {
+                    Logger.Operations($"Received bad data for cpu usage from extension point : {machineCpuUsage}");
+                }
+
+                machineCpuUsage = 0;
```
maximburyak: Not sure that in this case we should use 0, maybe we should fall back to the "normal" calculation

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
```diff
@@ -0,0 +1,263 @@
...
+                _process.Start();
+            }
+            catch (Exception e)
+            {
+                NotifyWarning("Could not start cpu usage extension point process", e);
+
+                Dispose();
+            }
+
+            var _ = ReadProcess(); // explicitly starting async task without waiting for it
```
maximburyak: What if we failed starting the process?

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
```diff
@@ -0,0 +1,263 @@
...
+                _process = null;
+            }
+        }
+
+
+        private async Task ReadProcess()
+        {
+            while (IsDisposed == false)
+            {
+                var nextLine = _process.StandardOutput.ReadLineAsync();
```
maximburyak: We should Try-Catch here, we don't want an unhandled failed task, we could for an instance have a temporary out of memory problem here, that will just kill the task, without us knowing/monitoring it.

src/Raven.Server/RavenServer.cs
```diff
@@ -1628,6 +1644,7 @@ public void Dispose()
                 ea.Execute(() => _clusterMaintenanceWorker?.Dispose());
 
                 ea.ThrowIfNeeded();
+                CpuUsage.CpuUsageExtensionPoint?.Dispose();
```
ayende: This should be inside the `ea.Execute`

src/Raven.Server/Dashboard/ThreadsInfo.cs
```diff
@@ -14,7 +14,7 @@ public class ThreadsInfo : AbstractDashboardNotification
 
         public double CpuUsage { get; set; }
 
-        public long ActiveCores { get; set; }
+        public double ActiveCores { get; set; }
```
ayende: There is no meaning from our perspective to `0.35` of a core. The extension script needs to do the transaction, not us

src/Raven.Server/Utils/CpuUsage.cs
```diff
@@ -96,6 +113,24 @@ private static double CalculateProcessCpuUsage(ProcessInfo currentInfo, ProcessI
             return Math.Min(100, processCpuUsage);
         }
 
+        private static (double MachineCpuUsage, double ProcessCpuUsage) CalculateCpuFromExtensionPoint()
+        {
+            var extensionPoint = new ExtensionPointInfo(CpuUsageExtensionPoint.Data);
+            var machineCpuUsage = extensionPoint.MachineCpuUsage;
+            if (machineCpuUsage < 0)
```
ayende: In what case this can happen? And why shouldn't we abort the extension at this point?

src/Raven.Server/Utils/CpuUsage.cs
```diff
@@ -276,7 +311,25 @@ private static unsafe MacInfo GetMacInfo()
...
+                return GetNumberOfActiveCoresInternal(process);
+            }
+            if (extensionPointRawData.ActiveCores >= 0)
+            {
+                return extensionPointRawData.ActiveCores;
+            }
+
+            if (Logger.IsOperationsEnabled)
+            {
+                Logger.Operations($"Received bad data for active cores from extension point : {extensionPointRawData.ActiveCores}");
```
ayende: This like the admin, what is `from extension point` which one, etc?
Need to be very explicit about this.

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
ayende: Also need error handling here

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
```diff
@@ -0,0 +1,263 @@
...
+            if (data == null)
+            {
+                HandleError("The output stream of the process has closed.");
+                return false;
+            }
+            using (_contextPool.AllocateOperationContext(out var context))
+            {
+                try
+                {
+                    var blittable = context.ReadForMemory(data, string.Empty);
```
ayende: Please use a descriptive name for the tag, not `string.Empty`

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
```diff
@@ -0,0 +1,263 @@
...
+            catch (Exception)
+            {
+                // When the process terminating, killed or exited
+            }
+            finally
+            {
+                if (_didRestart == false)
+                {
+                    _didRestart = true;
+                    NotifyWarning($"{msg} \nTherefore the process will restart.", e);
```
ayende: use `Environment.NewLine`

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
```diff
@@ -0,0 +1,263 @@
...
+            }
+            catch (Exception)
+            {
+                // When the process terminating, killed or exited
+            }
+            finally
+            {
+                if (_didRestart == false)
+                {
+                    _didRestart = true;
```
ayende: what resets this value?

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
```diff
@@ -0,0 +1,263 @@
...
+                    NotifyWarning($"{msg} \nTherefore the process will restart.", e);
+
+                    InitNewProcessHandle();
```
ayende: what if errors are thrown here?

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
```diff
@@ -0,0 +1,263 @@
...
+                if (_process.HasExited == false)
+                {
+                    _process.Kill();
+                }
+            }
+            catch (Exception)
+            {
+                // When the process terminating, killed or exited
+            }
+            finally
```
ayende: This shouldn't be in the finally

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
```diff
@@ -0,0 +1,263 @@
...
+                    NotificationSeverity.Warning);
+                _notificationCenter.Add(alert);
+            }
+            catch
+            {
+                // nothing to do if we can't report it
+            }
+        }
+
+        private void InitNewProcessHandle()
```
ayende: Why does this need to be separate from the `Start`?

src/Raven.Server/Dashboard/ThreadsInfo.cs
ppekrol: Can you explain what is the case when this is not int/long ?

ayende (changes requested):

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
```diff
@@ -0,0 +1,263 @@
...
+                {
+                    _didRestart = true;
+                    NotifyWarning($"{msg} \nTherefore the process will restart.", e);
+
+                    InitNewProcessHandle();
+                    Start();
+                }
+                else
+                {
+                    NotifyWarning($"{msg} \nTherefore the process will terminate.", e);
```
ppekrol: Need to use `Environment.NewLine` instead of `\n`

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
ppekrol: Need to use `Environment.NewLine` instead of `\n`

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
```diff
@@ -0,0 +1,263 @@
...
+            };
+
+            InitNewProcessHandle();
+        }
+
+        public void Start()
+        {
+            try
+            {
+                _process.Start();
```
ppekrol: Wouldn't it be more convenient for the user if the script would be started every X seconds and we would only read one line from it?

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
ppekrol: Is JSON a way to go when it comes to scripts? It is not a commonly used format.

I would expect output like:
```
80.06 3.0
```

src/Raven.Server/RavenServer.cs
```diff
@@ -296,6 +294,24 @@ void ConfigureKestrel(KestrelServerOptions options)
...
+        private void RunCpuUsageExtensionPoint()
+        {
+            if (string.IsNullOrEmpty(Configuration.Monitoring.CpuUsageMonitorExec))
+            {
+                return;
+            }
+
+            var inspector = new CpuUsageExtensionPoint(
+                _tcpContextPool,
+                Configuration.Monitoring.CpuUsageMonitorExec,
```
ppekrol: We probably need to escape this?

src/Raven.Server/RavenServer.cs
```diff
@@ -296,6 +294,24 @@ void ConfigureKestrel(KestrelServerOptions options)
...
+                Configuration.Monitoring.CpuUsageMonitorExecArguments,
```
ppekrol: We probably need to escape this?

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
```diff
@@ -0,0 +1,263 @@
...
+            }
+            catch (Exception e)
+            {
+                NotifyWarning("Could not start cpu usage extension point process", e);
+
+                Dispose();
```
ppekrol: Need to rethrow here?

ppekrol (changes requested):

src/Raven.Server/Dashboard/ThreadsInfo.cs
ayende: The problem is when running in k8s, when you can assign 0.5 cores

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
ayende: No, we want this to be long lived, otherwise there are a lot of extra costs.
In particular, the common scenario we see for this is making requests to a k8s controller, which means that it will need to auth, which means that we want to have connection pooling, etc.

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
ayende: We expect this to be a full fledged program, not just a script.
And we want to make sure that we are future proof here.

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
haludi: By "error handling" do you mean exception from `StreamReader` method?

src/Raven.Server/Utils/CpuUsage.cs
haludi: @grisha-kotler Why? If I will get 0 cores I want the same behavior. To return the last  Cpu usage calculated or 0

src/Raven.Server/Utils/CpuUsage.cs
haludi: @maximburyak The `CpuUsageExtensionPoint` should be set once in the server initialization. Maybe if I will pass it to a function it will make more sense?

src/Raven.Server/Utils/CpuUsage.cs
haludi: @maximburyak But the situation in the first place is the "normal" calculation is wrong.
In my humble opinion, it is better to display 0 that show you something is wrong

src/Raven.Server/Utils/CpuUsage.cs
haludi: @ayende This can happen when: 
* There were issues in the extension point process and there was one try to restart
*  If the process returns a negative value.
What do you mean by abort? Just to kill the process? In the first scenario, it is already the situation. In the second scenario if I check for bottom edge maybe to check the upper side?

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
haludi: @ayende Nothing, the implementation now is if the process has an issue the system trying to restart it once and no more

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
haludi: @ayende The `InitNewProcessHandle` function just set the new process object properties and `Start` function handles itself

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
haludi: @ayende I should not catch the exception?

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
haludi: @ppekrol As I understand it the error handling should be soft so I just notify there is a problem. What do you think? how the error should be handled?

ppekrol: Then you probably want to return here, not 'ReadProcess' ? (check @maximburyak comment)

src/Raven.Server/Dashboard/ThreadsInfo.cs
ayende: This still needs to be changed. The translation from partial units to integer number of cores should be done in the script, not by RavenDB

src/Raven.Server/RavenServer.cs
ayende: Yes, we do

src/Raven.Server/RavenServer.cs
ayende: Yes

src/Raven.Server/Utils/CpuUsage.cs
ayende: Yes, on bad input from the script, kill it and retry.

src/Raven.Server/Utils/CpuUsage.cs
ayende: 0 is _not_ wrong.
0 is a valid value

haludi: So should I return a negative value?

src/Raven.Server/Dashboard/ThreadsInfo.cs
haludi: In a situation where the number of cores is a fraction, the process cpu calculation will be wrong. 
To calculate the process cpu usage we compare the license limit vs the actual number of cores and use the min.

haludi: https://github.com/ravendb/ravendb/pull/7960#discussion_r232563223

src/Raven.Server/Utils/CpuUsage.cs
haludi: https://github.com/ravendb/ravendb/pull/7960#discussion_r232561460
And to return a negative value or keep it 0?

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
haludi: I want to restart or dispose the object while notifying the problem any way

src/Raven.Server/Utils/CpuUsage.cs
ayende: Yes, and make sure that this is user visible

src/Raven.Server/Dashboard/ThreadsInfo.cs
ayende: That is not our responsibility. That is what the script is for.
It would be doing that translation

ayende: To give more context, this is important because "core" isn't well defined.
A Raspberry Pi has 4 cores, or example, but my laptop can probably beat all of them together with just one core.

The script defines what a core means in this context, not the system. 
We do ensure that we aren't going over the number of licensed core, but that is it.

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
haludi: In this case, the origin methods to calculate CPU usage give wrong values so if the process didn't start or if it failed while running I still want to read the data from the disposed object to get error value from it. Another option I can think of is to switch the `CpuUsageExtensionPoint` to "Failed object" that will do this work and use it is I can't start. Is it make sense?

src/Raven.Server/Dashboard/ThreadsInfo.cs
iftahbe: We calculate machine usage and process usage separately. If you want to delegate all that responsibility to the script, it will have to give us RavenDB's process usage as well. So the script will have to monitor the RavenDB process. It makes it much harder than just calling the K8s API metrics-server.

src/Raven.Server/Utils/CpuUsage.cs
maximburyak: Maybe the CpuUsage class shouldn't be a "static" class, but a singleton? And this way you'll initialize all the values you need inside the constructor? It looks to me that it lacks structure and state this way..

src/Raven.Server/Dashboard/ThreadsInfo.cs
ayende: I pretty sure that we have to. 
And we can do that in the same way we do now, just in the script

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
ppekrol: If process died, then you want to `ReadErrors`, not the normal output I think?

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
```diff
@@ -0,0 +1,284 @@
...
+                    if (_process.HasExited == false)
+                    {
+                        _process.Kill();
+                    }
+                }
+                catch (Exception)
+                {
+                    // When the process terminating, killed or exited
+                }
+                errors = errors + await _process.StandardError.ReadToEndAsync();
```
ppekrol: Can't we read to end at the beginning?

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
```diff
@@ -0,0 +1,284 @@
...
+            if (data == null)
+            {
+                HandleError("The output stream of the process has closed.");
+                return false;
+            }
+            using (_contextPool.AllocateOperationContext(out var context))
+            {
+                try
+                {
+                    var blittable = context.ReadForMemory(data, "cpuUsageExtensionPointData");
```
ppekrol: We need to dispose this

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
haludi: As I understand it the `ReadToEndAsync` would not return until the stream is closed.
So if the process send error without exiting it will continue to wait for the continuation

src/Raven.Server/Utils/CpuUsageExtensionPoint.cs
haludi: There is a task that responsible to read errors, notify them and kill the process if needed. But after that or if the process killed/existed by another reason when for instance the Dashboard requests for CPU usage it will get negative values to indicate the problem

ppekrol:
Don't mind the random `System.Net.Sockets.SocketException : The I/O operation has been aborted because of either a thread exit or an application request` it is related to https://github.com/ravendb/ravendb/pull/8028

src/Raven.Server/Dashboard/ThreadsInfo.cs
ayende: This is still double, and should be int

src/Raven.Server/Utils/Cpu/CpuUsage.cs
```diff
@@ -0,0 +1,185 @@
...
+using Sparrow.Json;
+using Sparrow.Logging;
+using Sparrow.Platform;
+using Sparrow.Platform.Posix;
+using Sparrow.Platform.Posix.macOS;
+using Sparrow.Utils;
+
+namespace Raven.Server.Utils.Cpu
+{
+    public static class CpuUsage
```
ayende: This should be an instance class, not a static class. 
The problem is that in the tests, we have multiple raven servers, and you are going to impact all of them in this way

src/Raven.Server/RavenServer.cs
```diff
@@ -1654,6 +1656,7 @@ public void Dispose()
                 ea.Execute(() => ServerMaintenanceTimer?.Dispose());
                 ea.Execute(() => AfterDisposal?.Invoke());
                 ea.Execute(() => _clusterMaintenanceWorker?.Dispose());
+                ea.Execute(CpuUsage.Dispose);
```
ayende: Only dispose the local instance, not a static one

src/Raven.Server/Utils/Cpu/CpuUsage.cs
```diff
@@ -0,0 +1,185 @@
...
+        [StructLayout(LayoutKind.Sequential)]
+        internal struct FileTime
+        {
+            public int dwLowDateTime;
+            public int dwHighDateTime;
+        }
+
+        [return: MarshalAs(UnmanagedType.Bool)]
+        [DllImport("kernel32.dll", SetLastError = true)]
+        internal static extern bool GetSystemTimes(
```
ayende: Why is this here and not in the Windows one?

src/Raven.Server/Utils/Cpu/CpuUsageCalculator.cs
```diff
@@ -0,0 +1,265 @@
...
+    internal interface ICpuUsageCalculator : IDisposable
+    {
+        (double MachineCpuUsage, double ProcessCpuUsage) Calculate();
+        void Init();
+    }
+
+    [SuppressMessage("ReSharper", "StaticMemberInGenericType")]
+    internal abstract class CpuUsageCalculator<T> : ICpuUsageCalculator where T : ProcessInfo
+    {
+        protected static readonly Logger Logger = LoggingSource.Instance.GetLogger<MachineResources>("Server");
```
ayende: Why _do_ you have multiple static members here for each type?

src/Raven.Server/Utils/Cpu/CpuUsageCalculator.cs
```diff
@@ -0,0 +1,265 @@
...
+            if (CpuUsage.GetSystemTimes(ref systemIdleTime, ref systemKernelTime, ref systemUserTime) == false)
+            {
+                if (Logger.IsInfoEnabled)
+                    Logger.Info("Failure when trying to get GetSystemTimes from Windows, error code was: " + Marshal.GetLastWin32Error());
+                return null;
+            }
+
+            return new WindowsInfo
+            {
+                SystemIdleTime = CpuUsage.GetTime(systemIdleTime),
```
ayende: `CpuUsage.GetTime` move to this class

src/Raven.Server/Utils/Cpu/CpuUsageExtensionPoint.cs
```diff
@@ -0,0 +1,277 @@
...
+            MachineCpuUsage = -1
+        };
+
+        public ExtensionPointData Data
+        {
+            get
+            {
+                if (IsDisposed && _logger.IsOperationsEnabled)
+                {
+                    _logger.Operations($"Try to get data from {nameof(CpuUsageExtensionPoint)} which was disposed");
```
ayende: This will be called all the time, which will flood the log

src/Raven.Server/Utils/Cpu/CpuUsageExtensionPoint.cs
```diff
@@ -0,0 +1,277 @@
...
+                    if (_process.HasExited == false)
+                    {
+                        _process.Kill();
+                    }
+                }
+                catch (Exception)
+                {
+                    // When the process terminating, killed or exited
+                }
+                errors = errors + await _process.StandardError.ReadToEndAsync();
```
ayende: need new line here

src/Raven.Server/Utils/Cpu/CpuUsageExtensionPoint.cs
```diff
@@ -0,0 +1,277 @@
...
+                {
+                    using (var blittable = context.ReadForMemory(data, "cpuUsageExtensionPointData"))
+                    {
+                        if (TryGetValue(blittable, nameof(ExtensionPointData.MachineCpuUsage), out var machineCpuUsage)
+                            && TryGetValue(blittable, nameof(ExtensionPointData.ProcessCpuUsage), out var processCpuUsage))
+                        {
+                            _data.MachineCpuUsage = machineCpuUsage;
+                            _data.ProcessCpuUsage = processCpuUsage;
+                            return true;
+                        }
```
ayende, line 177: comment here that `HandleError` is called from inside the `TryGetValue`

src/Raven.Server/Utils/Cpu/CpuUsageExtensionPoint.cs
```diff
@@ -0,0 +1,277 @@
...
+                catch (Exception e)
+                {
+                    HandleError($"Unable to parse \"{data}\" to json", e);
+                }
+            }
+
+            return false;
+        }
+
+        private bool TryGetValue(BlittableJsonReaderObject blittable, string machineCpuUsageName, out double cpuUsage)
```
ayende: `TryGetValue` is a generic name, and you are also doing error handling here, call it as such

src/Raven.Server/Utils/Cpu/CpuUsageExtensionPoint.cs
```diff
@@ -0,0 +1,277 @@
...
+            }
+            catch (Exception)
+            {
+                // When the process terminating, killed or exited
+            }
+            finally
+            {
+                if (_didRestart == false)
+                {
+                    _didRestart = true;
```
ayende: `_didRestart` change this to a `DateTime` and allow it if `DateTime.UtcNow - _lastRestart` is greater than 15 minutes

src/Raven.Server/Utils/ThreadsUsage.cs
```diff
@@ -40,7 +41,7 @@ public ThreadsInfo Calculate()
                 var activeCores = CpuUsage.GetNumberOfActiveCores(process);
                 threadsInfo.ActiveCores = activeCores;
 
-                if (timeDiff == 0 || activeCores == 0)
+                if (timeDiff == 0 || Math.Abs(activeCores) < 0.01)
```
ayende: activeCores should not be a double

ayende (changes requested):

ppekrol (approved):

PR #7961, opened 2018-11-04
RavenDB-12024: 32bits failing tests

garayx:


ayende (approved):

ppekrol:
@garayx Don't we want to apply tests fixes to 4.0 also?

garayx:
> @garayx Don't we want to apply tests fixes to 4.0 also?

The indexing tests  are only on v4.1, the two others I want to see if they will fail with the ArenaMemoryAllocator.cs fix if so I will port the code.

PR #7966, opened 2018-11-05
RavenDB-12184: Check if there is a ConnectionString before adding rav…

garayx:
…en/sql ETL command

### DIFF
--- a/src/Raven.Server/ServerWide/ClusterStateMachine.cs
+++ b/src/Raven.Server/ServerWide/ClusterStateMachine.cs
@@ -1029,23 +1029,6 @@ namespace Raven.Server.ServerWide
                         }
 
                         databaseRecord = JsonDeserializationCluster.DatabaseRecord(databaseRecordJson);
-                        
-                        if (type == "AddRavenEtlCommand" || type == "AddSqlEtlCommand")
-                        {
-                            if (cmd.TryGet("Configuration", out BlittableJsonReaderObject obj))
-                            {
-                                var connectionStringName = obj["ConnectionStringName"].ToString();
-                                switch (type)
-                                {
-                                    case "AddSqlEtlCommand" when databaseRecord.SqlConnectionStrings.TryGetValue(connectionStringName, out var sqlConnection) == false:
-                                        ThrowUnknownConnectionStringName(connectionStringName);
-                                        break;
-                                    case "AddRavenEtlCommand" when databaseRecord.RavenConnectionStrings.TryGetValue(connectionStringName, out var sqlConnection) == false:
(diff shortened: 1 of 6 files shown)


src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1029,6 +1029,23 @@ private void UpdateDatabase(TransactionOperationContext context, string type, Bl
                         }
 
                         databaseRecord = JsonDeserializationCluster.DatabaseRecord(databaseRecordJson);
+                        
+                        if (type == "AddRavenEtlCommand" || type == "AddSqlEtlCommand")
```
ppekrol: This should be handled when you are putting new ETL. In `OutgoingTasksHandler.AddEtl`

ppekrol (changes requested):

src/Raven.Server/ServerWide/ClusterStateMachine.cs
ppekrol: Also I think that when we want to remove connection string in `OutgoingTasksHandler.RemoveConnectionString` then we need to check if any ETLs are using it or not and throw exception with list of them when we find any.

garayx: done

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -1324,13 +1324,19 @@ public async Task<(long, object)> ModifyPeriodicBackup(TransactionOperationConte
...
             UpdateDatabaseCommand command;
+            var databaseRecord = LoadDatabaseRecord(databaseName, out var _);
+            var connectionStringName = etlConfiguration["ConnectionStringName"].ToString();
 
             switch (EtlConfiguration<ConnectionString>.GetEtlType(etlConfiguration))
             {
                 case EtlType.Raven:
+                    if (databaseRecord.RavenConnectionStrings.TryGetValue(connectionStringName, out var rvnConnection) == false)
+                        ThrowUnknownConnectionStringName(connectionStringName);
                     command = new AddRavenEtlCommand(JsonDeserializationCluster.RavenEtlConfiguration(etlConfiguration), databaseName);
```
ppekrol: We are deserializing the configuration here (`JsonDeserializationCluster.RavenEtlConfiguration(etlConfiguration)`). Use this instance and:
1. Fire `.Validate()` method and throw if there are any errors
2. Add new method in `abstract EtlConfiguration` called `.ValidateConnectionString(DatabaseRecord)`

ppekrol (approved):

PR #7972, opened 2018-11-06
RavenDB-12110: Failing before getting to actual query will be tracked now.

garayx:


src/Raven.Server/Documents/Handlers/QueriesHandler.cs
```diff
@@ -43,17 +44,41 @@ public Task Get()
...
+                        {
+                            await Debug(context, debug, token, tracker, httpMethod);
+                            return;
+                        }
+
+                        await Query(context, token, tracker, httpMethod);
+                    }
+                }
+                catch (Exception e)
+                {
```
ppekrol, line 64: I think that this will not work the way you want.
E.g. 'Query' method catches the exceptions and sets the tracker query already (e.g. in IndexQueryServerSide.Create) and after that it rethrows the exception, which you will catch again here and override the tracker value.

grisha-kotler: if (tracker.Query == null)

garayx: I overwrite only if tracker.Query is null otherwise I just throw.

ppekrol: I'm getting old :)

Another question:
```csharp
    AddStringToHttpContext(httpContext, errorMessage, TrafficWatchChangeType.Queries);
```

do we need sth like that here?

src/Raven.Server/Documents/Handlers/QueriesHandler.cs
```diff
@@ -43,17 +44,41 @@ public Task Get()
...
+                        {
+                            errorMessage = "Failed: " +
+                                           HttpContext.Request.Path.Value +
+                                           e.ToString();
+                        }
+                        tracker.Query = errorMessage;
+                    }
+                    throw;
                 }
-
```
ayende, line 55: What happens if the `tracker.Query` _isn't_ null? How does the user get the error?

garayx: Its handled inside _await Query(context, token, tracker, httpMethod);_ and the query is passed to the tracker.
There is a problem with case that if we didnt succeed to _parse the query to memory_ OR didnt succeed to _AllocateOperationContext_ then the _tracker.Query_ remains null and this aims to change that.

src/Raven.Server/Documents/Handlers/QueriesHandler.cs
garayx: done

PR #7973, opened 2018-11-06
RavenDB-12151 - add the size of the lucene files to the tx allocations

grisha-kotler:


src/Voron/GlobalFlushingBehavior.cs
```diff
@@ -242,7 +242,7 @@ private void FlushEnvironments(HashSet<StorageEnvironment> avoidDuplicates)
                     // we haven't reached the point where we have to flush, but we might want to, if we have enough 
                     // resources available, if we have more than half the flushing capacity, we can do it now, otherwise, we'll wait
                     // until it is actually required.
-                    if (_concurrentFlushes.CurrentCount < StorageEnvironment.MaxConcurrentFlushes / 2)
+                    if (_concurrentFlushes.CurrentCount > StorageEnvironment.MaxConcurrentFlushes / 2)
```
ppekrol, line 245: Nice

src/Raven.Server/Indexing/VoronIndexOutput.cs
```diff
@@ -37,6 +44,7 @@ public VoronIndexOutput(StorageEnvironmentOptions options, string name, Transact
         public override void FlushBuffer(byte[] b, int offset, int len)
         {
             _file.Write(b, offset, len);
+            TotalWritten += len;
```
ppekrol: I'm wondering if we can do better here... not sure if this is enough to measure the allocations, but I'm not sure what else could we do?

src/Raven.Server/Indexing/LuceneVoronDirectory.cs
```diff
@@ -140,7 +142,10 @@ public override IndexOutput CreateOutput(string name, IState s)
             if (state == null)
                 throw new ArgumentNullException(nameof(s));
 
-            return new VoronIndexOutput(_environment.Options, name, state.Transaction, _name);
+            var voronIndexOutput = new VoronIndexOutput(_environment.Options, name, state.Transaction, _name, 
+                beforeFileDispose: () => _voronIndexFiles.Remove(name, out _));
```
ppekrol: Maybe instead of creating an action here per file (Lucene tends to create dozen of those). We can create a class `ReleaseFile` and pass one instance of it?

```
public class ReleaseFile
{
    ctor(_voronIndexFiles) { }

   void ReleaseFile(name)
   {
        _voronIndexFiles.Remove(name);
   }
}
```

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1820,73 +1821,82 @@ public virtual IndexProgress GetProgress(DocumentsOperationContext documentsCont
...
+                    var progress = new IndexProgress
                     {
                         Name = Name,
-                        Type = Type
+                        Type = Type,
+                        IndexRunningStatus = Status,
+                        Collections = new Dictionary<string, IndexProgress.CollectionStats>(StringComparer.OrdinalIgnoreCase)
                     };
+
+                    UpdateIndexProgress(documentsContext, progress, null);
```
ppekrol: Is this safe? We can have a `DatabaseShutdown` or `_disposeOne.Disposed` here.

src/Raven.Server/Indexing/VoronIndexOutput.cs
grisha-kotler: The total value is the same as _file.Length (which is more expensive)

src/Raven.Server/Documents/Indexes/Index.cs
grisha-kotler: We check that before that:
https://github.com/ravendb/ravendb/pull/7973/files#diff-a98877097408ea1a038c4101886f9a9bR1827

src/Raven.Server/Indexing/LuceneVoronDirectory.cs
```diff
@@ -160,6 +165,19 @@ public IDisposable SetTransaction(Transaction tx, out IState state)
 
         protected override void Dispose(bool disposing)
         {
+            // should be empty since we remove it from the dictionary on file dispose
+            Debug.Assert(_voronIndexFiles.Count == 0);
```
ppekrol: Let's change this to a real exception maybe?

src/Raven.Server/Documents/Indexes/Index.cs
ppekrol: Yes, and then you are executing this method here, even if those are true?

src/Raven.Server/Indexing/LuceneVoronDirectory.cs
grisha-kotler: done

src/Raven.Server/Documents/Indexes/Index.cs
grisha-kotler: Correct! fixed

src/Raven.Server/Indexing/LuceneVoronDirectory.cs
grisha-kotler: Wouldn't it cause index dispose to throw?

ppekrol: Lucene in general is copying back and forth the outputs, so you might be right, but if you are then Debug.Assert will kill the runtime when tests are executed in debug.

ppekrol:
Fixed compilation

src/Raven.Server/Indexing/LuceneVoronDirectory.cs
grisha-kotler: done

ppekrol (approved):

ppekrol:
@ayende can you review this? I'm merging, we will do additional fixes later on (if needed).

redknightlois:
I have been testing this on the production database LGTM

PR #7975, opened 2018-11-07
RavenDB-12204: Document metadata won't have hasCounter/hasRevisions flags when migrating without counters/revisions. 

garayx:
Also metadata won't have counters array when migrating without counters.

### DIFF
--- a/src/Raven.Server/Documents/Document.cs
+++ b/src/Raven.Server/Documents/Document.cs
@@ -70,6 +70,14 @@ namespace Raven.Server.Documents
             if (IndexScore.HasValue)
                 mutatedMetadata[Constants.Documents.Metadata.IndexScore] = IndexScore;
 
+            if (NonPersistentFlags.HasFlag(NonPersistentDocumentFlags.FromSmuggler) &&
+                Flags.HasFlag(DocumentFlags.HasCounters) == false &&
+                metadata != null &&
+                metadata.TryGet(Constants.Documents.Metadata.Counters, out BlittableJsonReaderArray _))
+            {
+                mutatedMetadata.Remove(Constants.Documents.Metadata.Counters);
+            }
+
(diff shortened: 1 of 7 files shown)


src/Raven.Server/Documents/Document.cs
```diff
@@ -70,6 +70,14 @@ public void EnsureMetadata()
             if (IndexScore.HasValue)
                 mutatedMetadata[Constants.Documents.Metadata.IndexScore] = IndexScore;
 
+            if (NonPersistentFlags.HasFlag(NonPersistentDocumentFlags.FromSmuggler) &&
```
ppekrol: This should be handled in StreamSource.GetDocuments.
We have a BlittableMetadataModifier there and you set a mode there that will strip the `@counters` property from the metadata when the document is being read.

ppekrol (changes requested):

src/Raven.Server/Documents/Document.cs
garayx: done

test/SlowTests/Issues/RavenDB-12204.cs
```diff
@@ -0,0 +1,127 @@
...
+using Xunit;
+
+namespace SlowTests.Issues
+{
+    public class RavenDB_12204 : RavenTestBase
+    {
+        [Fact]
+        public async Task CanMigrateFromRavenDb()
+        {
+            var file = Path.GetTempFileName();
```
ppekrol: use `NewDataPath` for this It will delete the directories/files when test will end.

garayx: done

ppekrol (approved):

ayende (approved):

PR #7977, opened 2018-11-07
RavenDB-7070 cleanup

ppekrol:


src/Raven.Server/Documents/Queries/Graph/RecursionQueryStep.cs
```diff
@@ -124,10 +123,10 @@ private void CompleteInitialization()
             }
         }
 
-        private async Task CompleteInitializationForStepAsyc(int position, ValueTask stepTask)
+        private async Task CompleteInitializationForStepAsync(int position, ValueTask stepTask)
         {
             await stepTask;
-            CompleteInitializationAfterLeft(position + 1);
+            await CompleteInitializationAfterLeft(position + 1);
```
ppekrol, line 129: I've added missing `await` here. @ayende is this correct?

PR #7982, opened 2018-11-08
RavenDB-11352 Pull Replication

karmeli87:
In pull replication there are two roles: central and edge.

In the central cluster we define on the database record a pull replication task with all the relevant settings like mentor, delay and certificates.
Once defined the central will wait for a connection to be initiated by an edge.

On the edge node we define an external replication with pull replication settings.
This means that a node in an edge cluster will request the tcp info for some node in the central and initiate the TCP connection so the central will send its documents via this connection.

Security:
The definition on the central will contain thumbprints for allowed certificates.
The edge need to put a certificate (with the private key) that corresponds to the thumbprint of the central definition. This certificate is store with different prefix so we wouldn't trust it by mistake.
Initiating a pull replication will use the edge certificate which the central should allow using the stored thumbprint in the corresponded definition.


src/Raven.Client/Documents/Operations/Replication/FeatureTaskDefinition.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System.Collections.Generic;
+using Raven.Client.Exceptions.Security;
+using Sparrow.Json;
+using Sparrow.Json.Parsing;
+
+namespace Raven.Client.Documents.Operations.Replication
+{
+    public abstract class FeatureTaskDefinition : IDynamicJsonValueConvertible
+    {
+        public List<string> Certificates; // thumbprint
```
ayende: We want to hold the full cert here, so we can do things like show the expiration

src/Raven.Client/Documents/Operations/Replication/ExternalReplication.cs
```diff
@@ -105,4 +108,31 @@ public string GetTaskName()
             return Name;
         }
     }
+
+    public class PullReplicationAsEdgeSettings
+    {
+        public bool PullReplicationEnabled = true;
+        public string CertificateThumbprint;
```
ayende: This is the definition on the _edge_, right? How do you decide which certificate to use here?

src/Raven.Client/ServerWide/Operations/Certificates/PutFeatureCertificateOperation.cs
```diff
@@ -0,0 +1,73 @@
...
+                    Content = new BlittableJsonContent(stream =>
+                    {
+                        using (var writer = new BlittableJsonTextWriter(ctx, stream))
+                        {
+                            writer.WriteStartObject();
+                            writer.WritePropertyName(nameof(CertificateDefinition.Name));
+                            writer.WriteString(_name.ToString());
+                            writer.WriteComma();
+                            writer.WritePropertyName(nameof(CertificateDefinition.Certificate));
+                            writer.WriteString(Convert.ToBase64String(_certificate.Export(X509ContentType.Pkcs12)));
```
ayende: Unless I miss something, this stores the _private_ key inside the server, that is _not_ something that we usually want to do.
Why is this done here?

ayende: Especially when you are talking about the global cluster level?

ayende:
```suggestion
                            writer.WriteString(Convert.ToBase64String(_certificate.Export(X509ContentType.Cert)));
```

src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
```diff
@@ -18,6 +19,28 @@ public enum OperationTypes
...
+        {
+            public enum AuthorizeMethod
+            {
+                Server,
+                PullReplication
+            }
+
+            public AuthorizeMethod AuthorizeAs;
+
+            public string RemoteConnectionInfo;
```
ayende: What is this string?

src/Raven.Server/Commercial/LicenseManager.cs
```diff
@@ -1320,6 +1320,19 @@ public void AssertCanAddExternalReplication()
             throw GenerateLicenseLimit(LimitType.ExternalReplication, details);
         }
 
+        public void AssertCanAddPullReplication()
+        {
+            //TODO: add this feature to the license
+//            if (IsValid(out var licenseLimit) == false)
```
ayende, line 1326: @grisha-kotler What is required here?

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
```diff
@@ -75,9 +77,14 @@ public class OutgoingReplicationHandler : IDisposable
...
         private readonly ConcurrentQueue<OutgoingReplicationStatsAggregator> _lastReplicationStats = new ConcurrentQueue<OutgoingReplicationStatsAggregator>();
         private OutgoingReplicationStatsAggregator _lastStats;
-        private readonly TcpConnectionInfo _connectionInfo;
+        private readonly TcpConnectionInfo[] _connectionInfo;
+        private TcpConnectionInfo _currentConnectionInfo;
 
-        public OutgoingReplicationHandler(ReplicationLoader parent, DocumentDatabase database, ReplicationNode node, bool external, TcpConnectionInfo connectionInfo)
+        // In case this is an outgoing pull replication from the central
+        // we need to associate this instance to the replication definition. 
+        public string PullReplicationDefinitionName;
```
ayende, line 83: What happen if we delete / disable the pull replication / change its settings when the connection is already alive?

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
```diff
@@ -118,237 +134,156 @@ public string GetNode()
             return node?.NodeTag;
         }
 
-        private void ReplicateToDestination()
+        private void PullReplication()
         {
-            try
+            NativeMemory.EnsureRegistered();
```
ayende, line 146: Pretty sure that this is already handled on `PoolOfThreads.GlobalRavenThreadPool.LongRunning`

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
```diff
@@ -118,237 +134,156 @@ public string GetNode()
...
-                            // the cluster that it has this db.
-                            if (young == false)
-                                AddAlertOnFailureToReachOtherSide(msg, e);
-
-                            throw;
-                        }
-                        catch (OperationCanceledException e)
+            if (connected == false)
+            {
+                throw new InvalidOperationException("Failed to connect to any of the supplied TCP connections.");
```
ayende: Bad idea here, We need to keep track of all the failures that we had an use them in an AggregateException as the inner exception here.

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
```diff
@@ -118,237 +134,156 @@ public string GetNode()
...
+            }
+
+            using (Interlocked.Exchange(ref _tcpClient, task.Result))
```
ayende, line 181: Why interlocked here?

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
```diff
@@ -118,237 +134,156 @@ public string GetNode()
...
-                            {
-                                if (_parent.DebugWaitAndRunReplicationOnce != null)
-                                {
-                                    _parent.DebugWaitAndRunReplicationOnce.Wait(_cts.Token);
-                                    _parent.DebugWaitAndRunReplicationOnce.Reset();
-                                }
+            var request = new DynamicJsonValue
+            {
+                ["Type"] = nameof(ReplicationInitialRequest),
+                [nameof(ReplicationInitialRequest.PullReplication)] = isPullReplication
```
ayende: Why do we need this is we have all the fields that are there just for pull replication?

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -160,7 +164,108 @@ public ReplicationLoader(DocumentDatabase database, ServerStore server)
...
+                    if (initialRequest.PullReplication)
+                    {
+                        PullReplicationDefinition pullReplicationDefinition;
+                        using (_server.Server.ServerStore.ContextPool.AllocateOperationContext(out TransactionOperationContext ctx))
+                        using (ctx.OpenReadTransaction())
+                        {
+                            pullReplicationDefinition = _server.Cluster.ReadPullReplicationDefinition(Database.Name, initialRequest.PullReplicationDefinition, ctx);
+                        }
+
+                        var taskId = (int)Hashing.XXHash64.Calculate(initialRequest.DatabaseGroupId, Encodings.Utf8);
```
ayende: Don't use `(int)` to get an integer here, use `Hashing.Mix` to make sure we include all the bits in the random effect

ayende: Also, why do you need the `taskId` here anyway?

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -160,7 +164,108 @@ public ReplicationLoader(DocumentDatabase database, ServerStore server)
...
+                }
+            }
+
+            CreateIncomingInstance(tcpConnectionOptions, buffer);
+        }
+
+        public void RunPullReplicationAsEdge(TcpConnectionOptions tcpConnectionOptions, JsonOperationContext.ManagedPinnedBuffer buffer, ReplicationNode destination)
+        {
+            var newIncoming = CreateIncomingReplicationHandler(tcpConnectionOptions, buffer);
+            //TODO: Incoming need to add, known new value
```
ayende: I don't understand this comment?

ayende: You mean `_incoming.GetOrAdd` ? It is _not_ a known new instance, you might already have this instance live already

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -715,11 +803,26 @@ private TcpConnectionInfo GetConnectionInfo(ReplicationNode node, bool external)
...
+                        {
+                            var cmd = new GetTcpInfoCommand("external-replication", database);
+                            requestExecutor.Execute(cmd, ctx);
+                            node.Database = database;
+                            node.Url = requestExecutor.Url;
+                            return new []{ cmd.Result };
+                        }
+                    }
+                   
+                    using (var requestExecutor = RequestExecutor.CreateForFixedTopology(exNode.ConnectionString.TopologyDiscoveryUrls, exNode.ConnectionString.Database, certificate, DocumentConventions.Default))
```
ayende: Can save some code by unifying the `requestExecutor ` and `ctx` usages across the `if`s

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -747,6 +853,38 @@ private TcpConnectionInfo GetConnectionInfo(ReplicationNode node, bool external)
...
         }
 
+        public X509Certificate2 GetCertificateForReplication(ReplicationNode node, out TcpConnectionHeaderMessage.AuthorizationInfo authorizationInfo)
+        {
+            authorizationInfo = null;
+            var exNode = node as ExternalReplication;
+            if (exNode == null)
+                return _server.Server.Certificate.Certificate;
+
+            if (exNode.PullReplicationAsEdgeSettings?.CertificateThumbprint != null)
```
ayende: No, we don't want this.
Instead, we want to have a `Certificate` property on the `ExternalReplication` directly that would hold the BASE64 certificate (including private key)

ayende: Actually, better would be to hold the actual certificate elsewhere (because the certificate private key is... private, but we need:
* To upload the actual pull certificate when creating the pull replication
* To prevent it from being downloaded

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -747,6 +853,38 @@ private TcpConnectionInfo GetConnectionInfo(ReplicationNode node, bool external)
...
+            {
+                using (_server.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
+                using (context.OpenReadTransaction())
+                {
+                    var key = Constants.Certificates.FeaturePrefix + exNode.PullReplicationAsEdgeSettings.CertificateThumbprint;
```
ayende: You have no way here to limit this for a single db. In particular, just knowing the certificate thumbprint allows you to use it from any db on the server, which is probably not ideal

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -747,6 +853,38 @@ private TcpConnectionInfo GetConnectionInfo(ReplicationNode node, bool external)
...
+                    if (certificate == null)
+                    {
+                        return _server.Server.Certificate.Certificate;
+                    }
+
+                    var definition = JsonDeserializationServer.CertificateDefinition(certificate);
+                    authorizationInfo = new TcpConnectionHeaderMessage.AuthorizationInfo
+                    {
+                        AuthorizeAs = TcpConnectionHeaderMessage.AuthorizationInfo.AuthorizeMethod.PullReplication,
+                        RemoteConnectionInfo = exNode.PullReplicationAsEdgeSettings.RemoteName
```
ayende: What is this remote info?

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
```diff
@@ -261,12 +269,18 @@ public async Task Put()
...
+            byte[] certBytes, string password, 
+            ServerStore serverStore,
+            TransactionOperationContext ctx, 
+            bool includePrivateKey = false)
         {
             var collection = new X509Certificate2Collection();
 
             if (string.IsNullOrEmpty(password))
-                collection.Import(certBytes, (string)null, X509KeyStorageFlags.MachineKeySet);
+                collection.Import(certBytes, (string)null, X509KeyStorageFlags.MachineKeySet | X509KeyStorageFlags.Exportable);
```
ayende: why this change?

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
```diff
@@ -296,12 +310,16 @@ public static async Task PutCertificateCollectionInCluster(CertificateDefinition
...
                     Password = certDef.Password,
-
                 };
 
-                if (x509Certificate.HasPrivateKey)
+                if (includePrivateKey)
                 {
-                    // avoid storing the private key
+                    // here we need to keep the private key, since in pull replication we are the initiator of the communication.
+                    currentCertDef.Certificate = Convert.ToBase64String(x509Certificate.Export(X509ContentType.Pfx));
```
ayende: This shouldn't be using the same mechanism, as certificates that we store that aren't private information.
This should be _encrypted_, talk with @iftahbe on how we store sensitive information like that.
I would also like it to be completely separated so you can't re-use existing value just by knowing its thumbprint.

src/Raven.Server/Web/System/TcpConnectionInfoHandler.cs
```diff
@@ -19,5 +27,116 @@ public Task Get()
 
             return Task.CompletedTask;
         }
+
+        [RavenAction("/info/tcp/pull", "GET", AuthorizationStatus.UnauthenticatedClients)]
```
ayende: Let's use a better name here: `info/remote-task/topology`, for example
This won't be the only usage of the endpoint

src/Raven.Server/Web/System/TcpConnectionInfoHandler.cs
```diff
@@ -19,5 +27,116 @@ public Task Get()
...
+        public async Task GetPullReplicationInfo()
+        {
+            var databaseName = GetStringQueryString("databaseName");
+            var databaseGroupId = GetStringQueryString("databaseGroupId");
+            var definitionName = GetStringQueryString("definitionName");
+
+            var nodes = GetResponsibleNode(databaseName, databaseGroupId, definitionName);
+
+            // TODO: cache the tcp info for each node?
```
ayende: No, there is no need for that. The number of calls you are likely to get is minuscule.

src/Raven.Server/Web/System/TcpConnectionInfoHandler.cs
```diff
@@ -19,5 +27,116 @@ public Task Get()
 
             return Task.CompletedTask;
         }
+
+        [RavenAction("/info/tcp/pull", "GET", AuthorizationStatus.UnauthenticatedClients)]
+        public async Task GetPullReplicationInfo()
+        {
```
ayende, line 37: Security check need to be right here, front and center so it would be obvious that is is happening

src/Raven.Server/Web/System/TcpConnectionInfoHandler.cs
```diff
@@ -19,5 +27,116 @@ public Task Get()
...
+                }
+                else
+                {
+                    var nodeUrl = clusterTopology.GetUrlFromTag(node);
+                    tasks.Add(GetNodeTcpInfo(nodeUrl, databaseName));
+                }
+            }
+
+            await Task.WhenAll(tasks);
+            foreach (var task in tasks)
```
ayende: Shouldn't this be in the proper order? Right now the current node tag is always first.

src/Raven.Server/Web/System/TcpConnectionInfoHandler.cs
```diff
@@ -19,5 +27,116 @@ public Task Get()
...
+            using (var writer = new BlittableJsonTextWriter(context, ResponseBodyStream()))
+            {
+                context.Write(writer, new DynamicJsonValue
+                {
+                    ["Results"] = output
+                });
+            }
+        }
+
+        private async Task<TcpConnectionInfo> GetNodeTcpInfo(string nodeUrl, string databaseName)
```
ayende: Why are you already getting the `TcpConnectionInfo` here? Let the client actually do that.
The response from here should be just the node urls as `http/s`

ayende: The reason this concerns me is that you are using the internal channels, and the remote server might be using a different network.

src/Raven.Server/Web/System/TcpConnectionInfoHandler.cs
```diff
@@ -19,5 +27,116 @@ public Task Get()
...
+
+        private List<string> GetResponsibleNode(string databaseName, string databaseGroupId, string definitionName)
+        {
+            var list = new List<string>();
+            using (ServerStore.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
+            using (context.OpenReadTransaction())
+            {
+                var pullReplication = ServerStore.Cluster.ReadPullReplicationDefinition(databaseName, definitionName, context);
+                var cert = HttpContext.Connection.ClientCertificate;
+                pullReplication.Validate(cert?.Thumbprint);
```
ayende: This is wrong, because it presume too many states.
You need to account for: 
* http node
* https with a known certificate that doesn't have access to this db
* https with a known certificate that _has_ access to this db
* https with a known admin certificate
* https with an unknown certificate that is explicitly allowed, etc
* using a database that doesn't exists.

This gives you a good way to check (with no auth) if a db exists.

ayende (changes requested):

src/Raven.Client/Documents/Operations/Replication/ExternalReplication.cs
karmeli87: The user should define it.

src/Raven.Client/Documents/Operations/Replication/FeatureTaskDefinition.cs
karmeli87: Done.

src/Raven.Client/ServerWide/Operations/Certificates/PutFeatureCertificateOperation.cs
karmeli87: This is a client code to upload the certificate along with it private key to the edge.
Storing the certificate is done at the endpoint.

src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
karmeli87: This is remote pull replication so, renamed to `PullReplicationDefinitionName`.

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
karmeli87: This is converted to `int` anyway in the `WhosTaskIsIt`.
We need this since this is a legit external replication and we keep its state across the central cluster. 
https://github.com/ravendb/ravendb/blob/ab8b050739d768eab847dfc3e60937bee07cff46/src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs#L260

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
karmeli87: Okay, so when putting a certificate on the edge we have to specify permissions.

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
karmeli87: renamed to `PullReplicationDefinition`. (note: this is the definition name on the central)

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
karmeli87: Need to extract the private key later. (see comment below)

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
karmeli87: Okay, so the encryption will be locally on each node, but I need to propagate this certificate across the cluster so I need to include the private key to be transferred on the wire.

src/Raven.Server/Web/System/TcpConnectionInfoHandler.cs
karmeli87: Done.

src/Raven.Server/Web/System/TcpConnectionInfoHandler.cs
karmeli87: The `GetResponsibleNode` take care of the proper order by executing `WhoseTaskIsIt` in a loop and removed the selected node every iteration.

src/Raven.Server/Web/System/TcpConnectionInfoHandler.cs
karmeli87: The reason here is that `/info/tcp` require `ValidUser` which the edge don't have.

src/Raven.Client/Documents/Operations/Replication/ExternalReplication.cs
ayende: Sort of. If the user specified a certificate to use here, you'll use that, otherwise you'll use the usual server certificate as normal.

src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
ayende: Which we'll need to change when we have pull ETL?

src/Raven.Server/Web/System/TcpConnectionInfoHandler.cs
ayende: Yes, _except_ that after that has been called, you call to `output.Add` directly, so it isn't in the right order

src/Raven.Server/Web/System/TcpConnectionInfoHandler.cs
ayende: So have another endpoint that would do the validation properly.
The problem of getting internal ip is something we want to avoid

src/Raven.Client/Documents/Commands/GetPullReplicationInfoCommand.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System.Net.Http;
+using Raven.Client.Http;
+using Raven.Client.Json.Converters;
+using Raven.Client.ServerWide.Commands;
+using Sparrow.Json;
+
+namespace Raven.Client.Documents.Commands
+{
+    public class GetPullReplicationInfoCommand : RavenCommand<TcpConnectionInfo[]>
```
ppekrol: internal

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
ayende: Why do you need to keep its state on the central cluster?  You may have many of them, so you don't really have _state_ here.
The problem isn't the int, the problem is the cast throwing away 50% of the bits without mixing them.

src/Raven.Client/Documents/Commands/GetPullReplicationInfoCommand.cs
```diff
@@ -0,0 +1,55 @@
...
+    {
+        private readonly string _remoteDatabase;
+        private readonly string _databaseGroupId;
+        private readonly string _name;
+
+        public GetPullReplicationInfoCommand(string remoteDatabase, string databaseGroupId, string name)
```
ppekrol: if all of them are required, check for nulls

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
ayende: No, you don't.
When you update a pull replication task you can provide it with a certificate, that will be stored internally (encrypted) and used.
No reuse between pull replications

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
ayende: This is something that we'll want to use for pull ETL as well

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
ayende: Hm.... you are correct. I didn't take into account the fact that we need this to be cross cluster.
In this case, I think we can just put the cert (with private key) in the db record and be done with it.
We already have similarly sensitive stuff with connection strings, and they go to the same location.

src/Raven.Client/Documents/Operations/Replication/UpdatePullReplicationDefinitionOperation.cs
```diff
@@ -0,0 +1,75 @@
...
+using Raven.Client.Documents.Conventions;
+using Raven.Client.Documents.Operations.OngoingTasks;
+using Raven.Client.Http;
+using Raven.Client.Json;
+using Raven.Client.Json.Converters;
+using Sparrow.Json;
+
+namespace Raven.Client.Documents.Operations.Replication
+{
+    public class UpdatePullReplicationDefinitionOperation : IMaintenanceOperation<ModifyOngoingTaskResult>
```
ppekrol: PutPullReplicationOperation?

src/Raven.Client/Documents/Replication/Messages/ReplicationLatestEtagRequest.cs
```diff
@@ -12,4 +14,17 @@ public class ReplicationLatestEtagRequest
...
         public string SourceMachineName { get; set; }
     }
-}
\ No newline at end of file
+
+    public class ReplicationInitialRequest
+    {
+        public bool PullReplication { get; set; }
+
+        public string PullReplicationDefinition { get; set; }
```
ppekrol: string?

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -224,6 +224,28 @@ public static RequestExecutor CreateForSingleNodeWithoutConfigurationUpdates(str
             return executor;
         }
 
+        public static RequestExecutor CreateForFixedTopology(string[] initialUrls, string databaseName, X509Certificate2 certificate, DocumentConventions conventions)
```
ppekrol: internal

src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
```diff
@@ -180,6 +206,7 @@ public class ReplicationFeatures
                 public bool MissingAttachments;
                 public bool Counters;
                 public bool ClusterTransaction;
+                public bool SupportPullReplication;
```
ppekrol: PullReplication

src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
```diff
@@ -37,12 +62,13 @@ public enum OperationTypes
         public static readonly int ReplicationBaseLine = 31;
         public static readonly int ReplicationAttachmentMissing = 40_300;
         public static readonly int ReplicationAttachmentMissingVersion41 = 41_300;
+        public static readonly int ReplicationWithPullOption = 42_000;
```
ppekrol: 42_300

src/Raven.Server/Documents/Handlers/PullReplicationHandler.cs
```diff
@@ -0,0 +1,36 @@
...
+        [RavenAction("/databases/*/admin/tasks/pull-replication", "PUT", AuthorizationStatus.Operator)]
+        public async Task DefinePullReplication()
+        {
+            if (ResourceNameValidator.IsValidResourceName(Database.Name, ServerStore.Configuration.Core.DataDirectory.FullPath, out string errorMessage) == false)
+                throw new BadRequestException(errorMessage);
+
+            ServerStore.LicenseManager.AssertCanAddPullReplication();
+            await DatabaseConfigurations((_, databaseName, blittableJson) =>
+                {
+                    var pullReplication = JsonDeserializationClient.PullReplicationDefinition(blittableJson);
```
ppekrol, line 25: Maybe do some validation?

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -747,6 +853,38 @@ private TcpConnectionInfo GetConnectionInfo(ReplicationNode node, bool external)
...
+            if (exNode.PullReplicationAsEdgeSettings?.CertificateThumbprint != null)
+            {
+                using (_server.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
+                using (context.OpenReadTransaction())
+                {
+                    var key = Constants.Certificates.FeaturePrefix + exNode.PullReplicationAsEdgeSettings.CertificateThumbprint;
+                    var certificate = _server.Cluster.Read(context, key);
+                    if (certificate == null)
+                    {
+                        return _server.Server.Certificate.Certificate;
```
ppekrol: Is this correct? When specified key does not exist then we want to use the server certificate? If yes, then probably we should check users intention with `string.IsNullOfWhiteSpace(CertificateThumbprint)` ?

src/Raven.Server/ServerWide/ClusterCommandsVersionManager.cs
```diff
@@ -95,8 +95,9 @@ public static class ClusterCommandsVersionManager
 
             [nameof(PutIndexesCommand)] = Base42CommandsVersion,
             [nameof(PutSortersCommand)] = Base42CommandsVersion,
-            [nameof(DeleteSorterCommand)] = Base42CommandsVersion
-        };
+            [nameof(DeleteSorterCommand)] = Base42CommandsVersion,
+	        [nameof(UpdatePullReplicationCommand)] = Base42CommandsVersion
+ };
```
ppekrol: :)

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1431,6 +1433,39 @@ public BlittableJsonReaderObject ReadRawDatabase<T>(TransactionOperationContext<
...
+                using (topology)
+                {
+                    return JsonDeserializationCluster.DatabaseTopology(topology);
+                }
+            }
+        }
+
+        public PullReplicationDefinition ReadPullReplicationDefinition(string database, string definitionName, TransactionOperationContext context)
+        {
+            var databaseRecord = ReadRawDatabase(context, database, out _);
```
ppekrol: put this in 'using' ?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1431,6 +1433,39 @@ public BlittableJsonReaderObject ReadRawDatabase<T>(TransactionOperationContext<
             return Read(context, "db/" + name.ToLowerInvariant(), out etag);
         }
 
+        public DatabaseTopology ReadDatabaseTopology(TransactionOperationContext context, string name)
+        {
+            using (var rawDatabaseRecord = ReadRawDatabase(context, name, out _))
+            {
+                rawDatabaseRecord.TryGet(nameof(DatabaseRecord.Topology), out BlittableJsonReaderObject topology);
+                using (topology)
```
ppekrol: No need if you are disposing the parent (rawDatabaseRecord) ?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1431,6 +1433,39 @@ public BlittableJsonReaderObject ReadRawDatabase<T>(TransactionOperationContext<
             return Read(context, "db/" + name.ToLowerInvariant(), out etag);
         }
 
+        public DatabaseTopology ReadDatabaseTopology(TransactionOperationContext context, string name)
+        {
+            using (var rawDatabaseRecord = ReadRawDatabase(context, name, out _))
+            {
+                rawDatabaseRecord.TryGet(nameof(DatabaseRecord.Topology), out BlittableJsonReaderObject topology);
```
ppekrol: Can this return false? Exception maybe in that case?

src/Raven.Server/ServerWide/Commands/UpdatePullReplicationCommand.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using Raven.Client.Documents.Operations.Replication;
+using Raven.Client.ServerWide;
+using Sparrow.Json.Parsing;
+
+namespace Raven.Server.ServerWide.Commands
+{
+    public class UpdatePullReplicationCommand : UpdateDatabaseCommand
+    {
+        public PullReplicationDefinition PullReplicationDefinition;
```
ppekrol: rename to Definition

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
```diff
@@ -990,7 +1019,7 @@ private static void ValidatePermissions(CertificateDefinition certificate, Serve
             foreach (var kvp in certificate.Permissions)
             {
                 if (string.IsNullOrWhiteSpace(kvp.Key))
-                    throw new ArgumentException("Error in permissions in the certificate definition, database name is empty");
+                    throw new ArgumentException("Error in permissions in the certificate centralDefinition, database name is empty");
```
ppekrol: centralDefinition ?

test/FastTests/Client/BulkInserts.cs
```diff
@@ -17,6 +17,7 @@ public class BulkInserts : RavenTestBase
     {
         [Theory]
         [InlineData(false)]
+        [InlineData(true)]
```
ppekrol: SSL one was moved to SlowTests. This can be removed.

src/Raven.Client/Documents/Operations/Replication/ExternalReplication.cs
```diff
@@ -16,6 +16,8 @@ public class ExternalReplication : ReplicationNode, IDatabaseTask, IDynamicJsonV
         public string MentorNode;
         public TimeSpan DelayReplicationFor;
 
+        public PullReplicationAsEdgeSettings PullReplicationAsEdgeSettings;
```
ppekrol: can we rename to Options? In most cases we are using Options.

src/Raven.Client/Documents/Operations/Replication/ExternalReplication.cs
```diff
@@ -105,4 +108,31 @@ public string GetTaskName()
             return Name;
         }
     }
+
+    public class PullReplicationAsEdgeSettings
+    {
+        public bool PullReplicationEnabled = true;
```
ppekrol: Just Enabled?

src/Raven.Client/ServerWide/DatabaseRecord.cs
```diff
@@ -61,7 +61,9 @@ public DatabaseRecord(string databaseName)
 
         public List<PeriodicBackupConfiguration> PeriodicBackups;
 
-        public List<ExternalReplication> ExternalReplications = new List<ExternalReplication>(); // Watcher only receives (slave)
+        public List<ExternalReplication> ExternalReplications = new List<ExternalReplication>();
+
+        public Dictionary<string, PullReplicationDefinition> PullReplicationDefinition = new Dictionary<string, PullReplicationDefinition>();
```
ppekrol: PullReplications

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
```diff
@@ -118,237 +134,156 @@ public string GetNode()
...
-                task.Wait(CancellationToken);
-                using (Interlocked.Exchange(ref _tcpClient, task.Result))
-                {
-                    var wrapSsl = TcpUtils.WrapStreamWithSslAsync(_tcpClient, _connectionInfo, _parent._server.Server.Certificate.Certificate,
-                        _parent._server.Engine.TcpConnectionTimeout);
-                    wrapSsl.Wait(CancellationToken);
+            using (_parent._server.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
+            using (context.OpenReadTransaction())
+            {
+                var record = _parent.LoadDatabaseRecord();
```
ppekrol, line 170: If this happens often, then we can use ReadRawRecord?

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
ayende: That is a really good point. If the user specified a value, it _must_ exists or generate an error

src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
karmeli87: I'm bad with names... how does `FeatureDefinitionName` sounds.

ayende: `AuthorizationFor` is probably better.

src/Raven.Client/Documents/Operations/Replication/FeatureTaskDefinition.cs
```diff
@@ -23,6 +24,10 @@ public virtual DynamicJsonValue ToJson()
             DynamicJsonArray certs = null;
             if (Certificates != null)
             {
+                foreach (var certificate in Certificates)
```
ayende: You can assume that this is valid here.

src/Raven.Client/Documents/Operations/Replication/FeatureTaskDefinition.cs
```diff
@@ -1,13 +1,14 @@
...
+using System.Security.Cryptography.X509Certificates;
 using Sparrow.Json;
 using Sparrow.Json.Parsing;
 
 namespace Raven.Client.Documents.Operations.Replication
 {
     public abstract class FeatureTaskDefinition : IDynamicJsonValueConvertible
     {
-        public List<string> Certificates; // thumbprint
+        public List<string> Certificates; // base64
```
ayende: Change to Dictionary<thubmprint, base64 cert>

src/Raven.Client/Documents/Operations/Replication/FeatureTaskDefinition.cs
```diff
@@ -33,21 +38,35 @@ public virtual DynamicJsonValue ToJson()
...
+                err = $"Certificate is needed for pull replication '{Name}'";
+                return false;
             }
 
-            if (Certificates.Contains(thumbprint) == false)
+            foreach (var certificate in Certificates)
             {
-                throw new AuthorizationException($"Certificate with the thumbprint {thumbprint} was not found for pull replication '{Name}'");
+                var cert = new X509Certificate2(Convert.FromBase64String(certificate));
+                if (cert.Thumbprint == thumbprint)
```
ayende: Now that you have a match, you need to note that you _don't_ need to validate the certificate here.
Why? Because it was already validated (expiration, etc) by the calling code.

src/Raven.Client/Documents/Operations/Replication/PullReplicationDefinition.cs
```diff
@@ -34,5 +34,13 @@ public override DynamicJsonValue ToJson()
             djv[nameof(DelayReplicationFor)] = DelayReplicationFor;
             return djv;
         }
+
+        public static string GetPrefix(string databaseName)
```
ayende: remove

src/Raven.Client/ServerWide/Operations/Certificates/PutFeatureCertificateOperation.cs
```diff
@@ -2,13 +2,14 @@
...
 using Raven.Client.Documents.Conventions;
+using Raven.Client.Documents.Operations;
 using Raven.Client.Http;
 using Raven.Client.Json;
 using Sparrow.Json;
 
 namespace Raven.Client.ServerWide.Operations.Certificates
 {
-    public class PutFeatureCertificateOperation : IServerOperation
+    public class PutFeatureCertificateOperation : IMaintenanceOperation
```
ayende: remove

src/Raven.Client/Documents/Operations/Replication/FeatureTaskDefinition.cs
karmeli87: Remove this, we don't need to validate this

src/Raven.Client/Documents/Operations/Replication/FeatureTaskDefinition.cs
karmeli87: Should be a dictionary<thumb, base 64 cert>

src/Raven.Client/Documents/Operations/Replication/FeatureTaskDefinition.cs
```diff
@@ -33,21 +38,35 @@ public virtual DynamicJsonValue ToJson()
...
-                throw new AuthorizationException($"Certificate is needed for pull replication '{Name}'");
+                err = $"Certificate is needed for pull replication '{Name}'";
+                return false;
             }
 
-            if (Certificates.Contains(thumbprint) == false)
+            foreach (var certificate in Certificates)
             {
-                throw new AuthorizationException($"Certificate with the thumbprint {thumbprint} was not found for pull replication '{Name}'");
+                var cert = new X509Certificate2(Convert.FromBase64String(certificate));
```
karmeli87: Just validating the thumbprint is fine, need to note here that we rely on calling code to validate the actual cert (expiration, etc)

src/Raven.Client/Documents/Operations/Replication/PullReplicationDefinition.cs
karmeli87: Remove

src/Raven.Client/ServerWide/Operations/Certificates/PutFeatureCertificateOperation.cs
karmeli87: Remove

src/Raven.Server/Documents/Handlers/PullReplicationHandler.cs
```diff
@@ -32,5 +42,62 @@ public async Task DefinePullReplication()
                     json[nameof(OngoingTask.TaskId)] = index;
                 }, statusCode: HttpStatusCode.Created);
         }
+
+        [RavenAction("/databases/*/admin/certificates/feature", "PUT", AuthorizationStatus.Operator)]
```
karmeli87: Remove

src/Raven.Server/Documents/Handlers/PullReplicationHandler.cs
```diff
@@ -32,5 +42,62 @@ public async Task DefinePullReplication()
...
+        public Task PutFeatureCertificateOnEdge()
+        {
+            return PutCertificate(PullReplicationDefinition.GetPrefix(Database.Name), includePrivateKey: true);
+        }
+
+        [RavenAction("/databases/*/info/remote-task/tcp", "GET", AuthorizationStatus.RestrictedAccess)]
+        public Task GetRemoteTaskTcp()
```
karmeli87: You cannot do this. This leaks the existence of a database by the fact that you are using `DatabaseRequestHandler`
You need to use a `RequestHandler` and accept the database name externally.

src/Raven.Server/Documents/Handlers/PullReplicationHandler.cs
```diff
@@ -32,5 +42,62 @@ public async Task DefinePullReplication()
...
+                context.Write(writer, output);
+            }
+
+            return Task.CompletedTask;
+        }
+
+        public static void Authenticate(HttpContext httpContext, ServerStore serverStore, string database, string remoteTask)
+        {
+            // TODO: think of some use cases like:
+            // 1. HTTP, but we defined cert.
```
karmeli87: Should allow, this is already handled externally and not here.
If we don't have HTTPS, we always allow.

src/Raven.Server/Documents/Handlers/PullReplicationHandler.cs
```diff
@@ -32,5 +42,62 @@ public async Task DefinePullReplication()
...
+            // 2. HTTPs, but no cert is defined.
```
karmeli87: This is invalid, and even if we have restricted here, we should abort it early

src/Raven.Server/Documents/Handlers/PullReplicationHandler.cs
```diff
@@ -32,5 +42,62 @@ public async Task DefinePullReplication()
...
+            // 3. HTTPs with some cert, but got the server cert.
```
karmeli87: If the cert has access, we get access.

src/Raven.Server/Documents/Handlers/PullReplicationHandler.cs
```diff
@@ -32,5 +42,62 @@ public async Task DefinePullReplication()
...
+            var feature = httpContext.Features.Get<IHttpAuthenticationFeature>() as RavenServer.AuthenticateConnection;
+            switch (feature?.Status)
+            {
+                case RavenServer.AuthenticationStatus.Allowed:
+                case RavenServer.AuthenticationStatus.Operator:
+                case RavenServer.AuthenticationStatus.ClusterAdmin:
+                    // we can trust this certificate
+                    return;
+
+                case null:
```
karmeli87: This is invalid

src/Raven.Server/Routing/RequestRouter.cs
```diff
@@ -239,6 +239,12 @@ private bool TryAuthorize(RouteInformation route, HttpContext context, DocumentD
                         case RavenServer.AuthenticationStatus.NotYetValid:
                         case RavenServer.AuthenticationStatus.None:
                         case RavenServer.AuthenticationStatus.UnfamiliarCertificate:
```
karmeli87: create another case and if this fails, `goto null` will handle this

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
```diff
@@ -198,175 +196,169 @@ public static void WriteCertificateAsPem(string name, byte[] rawBytes, string ex
...
-        }
-
+        
         [RavenAction("/admin/certificates", "PUT", AuthorizationStatus.Operator)]
         public Task Put()
         {
             return PutCertificate(Constants.Certificates.Prefix, includePrivateKey: false);
         }
 
-        private async Task PutCertificate(string prefix, bool includePrivateKey)
```
karmeli87: nope

src/Raven.Server/Web/RequestHandler.cs
```diff
@@ -537,5 +541,198 @@ protected void SetupCORSHeaders()
         {
             SetupCORSHeaders(HttpContext);
         }
+
+        protected async Task PutCertificate(string prefix, bool includePrivateKey)
```
karmeli87: nope

src/Raven.Server/Web/RequestHandler.cs
```diff
@@ -537,5 +541,198 @@ protected void SetupCORSHeaders()
...
+                {
+                    throw new InvalidOperationException($"Failed to put certificate {certificate.Name} in the cluster.", e);
+                }
+
+                NoContentStatus();
+                HttpContext.Response.StatusCode = (int)HttpStatusCode.Created;
+            }
+        }
+
+        public static async Task PutCertificateCollectionInCluster(
```
karmeli87: nope

src/Raven.Client/Documents/Operations/Replication/FeatureTaskDefinition.cs
```diff
@@ -0,0 +1,69 @@
...
+            };
+        }
+
+        public bool CanAccess(string thumbprint, out string err)
+        {
+            err = null;
+            if (string.IsNullOrEmpty(thumbprint))
+            {
+                if (Certificates == null || Certificates.Count == 0)
+                    return true;
```
ayende: This is wrong.  There no scenario where we have an "anonymous" pull replication, which is this is what this looks like.
If `Certificates` is null / empty, we return a negative result and do not allow access.

Note that this is different from the behavior when we use a known client certificate that already have access to this node / database.

src/Raven.Server/Commercial/LicenseManager.cs
ayende: @grisha-kotler Please check this. http://issues.hibernatingrhinos.com/issue/RavenDB-12415

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
ayende: Stilling missing handling for this

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
ayende: remove this line

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
```diff
@@ -118,237 +132,130 @@ public string GetNode()
...
 
-                using (_parent._server.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
-                using (context.OpenReadTransaction())
-                {
-                    var record = _parent.LoadDatabaseRecord();
-                    if (record == null)
-                        throw new InvalidOperationException($"The database record for {_parent.Database.Name} does not exist?!");
+        private void Replication()
+        {
+            NativeMemory.EnsureRegistered();
```
ayende, line 163: remove this line

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
ayende: Probably a good idea

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
```diff
@@ -425,6 +332,225 @@ void HandleException(Exception e)
...
+                    {
+                        using (var scope = stats.CreateScope())
+                        {
+                            try
+                            {
+                                if (Destination is InternalReplication dest)
+                                {
+                                    _parent.EnsureNotDeleted(dest.NodeTag);
+                                }
+
```
ayende, line 375: Probably need to validate that this is still valid here. 
Need to revalidate if the pull replication def changed (especially if just a cert was removed)

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -160,7 +164,107 @@ public ReplicationLoader(DocumentDatabase database, ServerStore server)
...
+            IncomingReplicationAdded?.Invoke(newIncoming);
+            newIncoming.Failed += RetryPullReplication;
+
+            // Update current thread name
+            newIncoming.DoIncomingReplication();
+
+            void RetryPullReplication(IncomingReplicationHandler incomingReplicationHandler, Exception exception)
+            {
+                // if the stream closed, it is our duty to reconnect
+                incomingReplicationHandler.Failed -= RetryPullReplication;
```
ayende: what about retries, backoff strategy, etc?

src/Raven.Server/RavenServer.cs
```diff
@@ -1563,6 +1583,22 @@ private bool TryAuthorize(RavenConfiguration configuration, Stream stream, TcpCo
...
                     }
+                case AuthenticationStatus.UnfamiliarCertificate:
+                    var info = header.AuthorizeInfo;
+                    if (info != null && info.AuthorizeAs == TcpConnectionHeaderMessage.AuthorizationInfo.AuthorizeMethod.PullReplication)
+                    {
+                        using (ServerStore.ContextPool.AllocateOperationContext(out TransactionOperationContext ctx))
+                        using (ctx.OpenReadTransaction())
+                        {
+                            var pullReplicationDefinition = ServerStore.Cluster.ReadPullReplicationDefinition(header.DatabaseName, info.AuthorizationFor, ctx);
+                            if (pullReplicationDefinition.CanAccess(certificate.Thumbprint, out var err))
```
ayende: What if the pull replication doesn't exists?
Note that in this case we must provide the same error as if the cert doesn't exists.
What error do we get from missing db in this case, for that matter?

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -1329,58 +1328,20 @@ public async Task<(long, object)> ModifyPeriodicBackup(TransactionOperationConte
...
-                        rvnEtlErr.Add($"Could not find connection string named '{rvnEtl.ConnectionStringName}'. Please supply an existing connection string.");
-
-                    ThrowInvalidConfigurationIfNecessary(rvnEtlErr);
-
-                    command = new AddRavenEtlCommand(rvnEtl, databaseName);
+                    command = new AddRavenEtlCommand(JsonDeserializationCluster.RavenEtlConfiguration(etlConfiguration), databaseName);
                     break;
                 case EtlType.Sql:
-                    var sqlEtl = JsonDeserializationCluster.SqlEtlConfiguration(etlConfiguration);
-                    sqlEtl.Validate(out var sqlEtlErr, validateName: false, validateConnection: false);
```
ayende, line 1348: Where is the validation now?

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -1329,58 +1328,20 @@ public async Task<(long, object)> ModifyPeriodicBackup(TransactionOperationConte
             string databaseName, BlittableJsonReaderObject etlConfiguration)
         {
             UpdateDatabaseCommand command;
-            var databaseRecord = LoadDatabaseRecord(databaseName, out _);
 
             switch (EtlConfiguration<ConnectionString>.GetEtlType(etlConfiguration))
             {
                 case EtlType.Raven:
-                    var rvnEtl = JsonDeserializationCluster.RavenEtlConfiguration(etlConfiguration);
-                    rvnEtl.Validate(out var rvnEtlErr, validateName: false, validateConnection: false);
```
ayende, line 1338: Where is the validation now?

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
```diff
@@ -999,5 +997,14 @@ private static void ValidatePermissions(CertificateDefinition certificate, Serve
                     throw new ArgumentException($"Error in permissions in the certificate definition, invalid access {kvp.Value} for database {kvp.Key}");
             }
         }
+
```
ayende, line 1000: Do you actually need to do any of the changes to this file now?

src/Raven.Server/Web/System/TcpConnectionInfoHandler.cs
```diff
@@ -19,5 +30,139 @@ public Task Get()
...
+                    // we can trust this certificate
+                    return;
+
+                case RavenServer.AuthenticationStatus.UnfamiliarCertificate:
+                    using (serverStore.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
+                    using (context.OpenReadTransaction())
+                    {
+                        var pullReplication = serverStore.Cluster.ReadPullReplicationDefinition(database, remoteTask, context);
+                        var cert = httpContext.Connection.ClientCertificate;
+                        if (pullReplication.CanAccess(cert?.Thumbprint, out var err) == false)
```
ayende: what happens if the db doesn't exists? What happens if the pull replication doesn't exists?

src/Raven.Server/Web/System/TcpConnectionInfoHandler.cs
```diff
@@ -19,5 +30,139 @@ public Task Get()
...
+        public static void Authenticate(HttpContext httpContext, ServerStore serverStore, string database, string remoteTask)
+        {
+            var feature = httpContext.Features.Get<IHttpAuthenticationFeature>() as RavenServer.AuthenticateConnection;
+
+            if (feature == null) // we are not using HTTPS 
+                return;
+
+            switch (feature.Status)
+            {
+                case RavenServer.AuthenticationStatus.Allowed:
```
ayende: This isn't true. The certificate is allowed to _some_ database, not necessarily this one.

ayende (changes requested):

src/Raven.Server/Web/System/TcpConnectionInfoHandler.cs
karmeli87: If the db doesn't exists `throw new DatabaseDoesNotExistException($"The database '{database}' doesn't exists.")`
If the pull replication doesn't exists `throw new InvalidOperationException($"Pull replication with the name '{definitionName}' isn't defined for the database '{database}'.");`

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
karmeli87: Don't think so. But since it was changed in different commits I'm not sure I can simply revert the file in this PR.

src/Raven.Server/Web/System/TcpConnectionInfoHandler.cs
ayende: Both of which leak information that you may not be allowed to see. 
We need to make sure that regardless of the method you use, if you don't have access to the db / pull replication, we reply with the same message:

"Cannot connect to '{pullReplicationName}' on '{database}'. The database or task may not exists or you don't have the credentials to access it"`

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
ayende: yes

src/Raven.Server/Web/System/TcpConnectionInfoHandler.cs
```diff
@@ -45,6 +46,9 @@ public Task GetRemoteTaskTopology()
             using (context.OpenReadTransaction())
             {
                 var pullReplication = ServerStore.Cluster.ReadPullReplicationDefinition(database, remoteTask, context);
+                if (pullReplication.Disabled)
```
ayende, line 49: You are only checking this _when the connection is started_, you need to also check it whenever the configuration changes and abort any existing connections

karmeli87: This is done under `HandleCentralPullReplication`.

karmeli87:
Closing this PR. 

Will create a new one with the new terms `Hub` and `Sink`

PR #7983, opened 2018-11-08
RavenDB-12269 Fixing handling of index replacement

arekpalinski:


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -554,6 +551,20 @@ internal ExitWriteLock DrainRunningQueries()
             }
         }
 
+        private void InitializeComponentsUsingEnvironment(DocumentDatabase documentDatabase, StorageEnvironment environment)
+        {
+            _contextPool?.Dispose();
+            _contextPool = new TransactionContextPool(environment);
+            
+            _indexStorage = new IndexStorage(this, _contextPool, documentDatabase);
```
ayende, line 559: You aren't calling `_indexStorage ?.Dispose()` intentionally?

test/SlowTests/Issues/RavenDB_12269.cs
```diff
@@ -0,0 +1,84 @@
...
+
+                db.IndexStore.StartIndexing();
+
+                WaitForIndexing(store);
+
+                // let's try to force calling storageEnvironment.Cleanup() inside ExecuteIndexing method
+                replacementIndex.LowMemory();
+                replacementIndex._indexStorage.Environment().LogsApplied();
+
+                Thread.Sleep(500);
```
ayende: This is recipe for test failures. Can you add an event, instead? Or do active wiating?

ayende (changes requested):

src/Raven.Server/Documents/Indexes/Index.cs
ppekrol: It is not disposable, it just wraps _environment

test/SlowTests/Issues/RavenDB_12269.cs
arekpalinski: Actually we don't need that. The below query will assert that that index state isn't Error. Fixed in https://github.com/ravendb/ravendb/pull/7983/commits/3094df574da4e8f74005637ab4ebae1211f8c631

ppekrol (approved):

PR #7988, opened 2018-11-08
RavenDB-12268 - cleanup the recycled scratch buffers only when there is no open read transaction that is still using it

grisha-kotler:
http://issues.hibernatingrhinos.com/issue/RavenDB-12268

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -1197,6 +1197,11 @@ namespace Raven.Server.Documents.Indexes
             }
         }
 
+        public void Cleanup()
+        {
+            _environment?.Cleanup();
+        }
+
         protected virtual bool ShouldReplace()
         {
             return false;
@@ -1807,7 +1812,7 @@ namespace Raven.Server.Documents.Indexes
 
                 Stop(disableIndex: true);
                 SetState(IndexState.Disabled);
-                _environment?.Cleanup();
+                Cleanup();
             }
         }
 
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -1131,11 +1131,13 @@ namespace Raven.Server.Documents.Indexes
                                     _logger.Info($"Indexing exceeded global limit for scratch space usage. Going to flush environment of '{Name}' index and forcing sync of data file");
 
                                 GlobalFlushingBehavior.GlobalFlusher.Value.MaybeFlushEnvironment(storageEnvironment);
-                                GlobalFlushingBehavior.GlobalFlusher.Value.ForceSyncEnvironment(storageEnvironment);
 
                                 if (_logsAppliedEvent.Wait(Configuration.MaxTimeToWaitAfterFlushAndSyncWhenExceedingScratchSpaceLimit.AsTimeSpan))
                                 {
-                                    // we've just flushed let's try to cleanup scratch space immediately
+                                    // we've just flushed let's start the sync process
+                                    GlobalFlushingBehavior.GlobalFlusher.Value.ForceSyncEnvironment(storageEnvironment);
+
+                                    // and cleanup scratch space immediately
                                     storageEnvironment.Cleanup();
                                 }
                             }
--- a/src/Voron/Impl/Scratch/ScratchBufferPool.cs
+++ b/src/Voron/Impl/Scratch/ScratchBufferPool.cs
@@ -384,12 +384,12 @@ namespace Voron.Impl.Scratch
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private void AddScratchBufferFile(ScratchBufferItem scratch)
         {
-            RemoveInactiveScratches(scratch);
+            RemoveInactiveScratches(scratch, cleanupRecycled: false);
 
             _scratchBuffers.AddOrUpdate(scratch.Number, scratch, (_, __) => scratch);
         }
 
-        private void RemoveInactiveScratches(ScratchBufferItem except)
+        private void RemoveInactiveScratches(ScratchBufferItem except, bool cleanupRecycled)
         {
             foreach (var item in _scratchBuffers)
             {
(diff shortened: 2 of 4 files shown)


test/SlowTests/Voron/Issues/RavenDB-12268.cs
```diff
@@ -0,0 +1,63 @@
...
+
+namespace SlowTests.Voron.Issues
+{
+    public class RavenDB_12268 : StorageTest
+    {
+        [Fact]
+        public void UsedScratchBuffersArentDeleted()
+        {
+            // prevent the work of the global flushing behavior
+            Env.Options.MaxNumberOfPagesInJournalBeforeFlush = long.MaxValue;
```
ayende: You can just set flushing to false, which is much better.
You are also creating the `Env` and then setting the options, that isn't supported.
`override` the customize options method

src/Voron/Impl/Scratch/ScratchBufferPool.cs
```diff
@@ -403,8 +403,12 @@ private void RemoveInactiveScratches(ScratchBufferItem except)
                 if (_scratchBuffers.TryRemove(scratchBufferItem.Number, out _) == false)
                     ThrowUnableToRemoveScratch(scratchBufferItem);
 
-                if (_recycleArea.Contains(scratchBufferItem) == false)
+                var isInRecycledArea = _recycleArea.Contains(scratchBufferItem);
+                if (cleanupRecycled || isInRecycledArea == false)
```
ayende: I'm not following how this helps us, can you expand on this?

ayende (changes requested):

test/SlowTests/Voron/Issues/RavenDB-12268.cs
grisha-kotler: done

src/Voron/Impl/Scratch/ScratchBufferPool.cs
grisha-kotler: I tried to use the existing function RemoveInactiveScratches. This solves the issue but doesn't clean the all the recycled scratches.
Fixed.

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1131,11 +1131,13 @@ protected void ExecuteIndexing()
                                     _logger.Info($"Indexing exceeded global limit for scratch space usage. Going to flush environment of '{Name}' index and forcing sync of data file");
 
                                 GlobalFlushingBehavior.GlobalFlusher.Value.MaybeFlushEnvironment(storageEnvironment);
-                                GlobalFlushingBehavior.GlobalFlusher.Value.ForceSyncEnvironment(storageEnvironment);
 
                                 if (_logsAppliedEvent.Wait(Configuration.MaxTimeToWaitAfterFlushAndSyncWhenExceedingScratchSpaceLimit.AsTimeSpan))
                                 {
-                                    // we've just flushed let's try to cleanup scratch space immediately
+                                    // we've just flushed let's start the sync process
+                                    GlobalFlushingBehavior.GlobalFlusher.Value.ForceSyncEnvironment(storageEnvironment);
```
arekpalinski, line 1138: Why? We _always_ want to schedule the sync. Even if we don't manage to do the flush in specified time. BTW this is unrelated to RavenDB-12268 please

grisha-kotler: - When you schedule a flush it will a schedule a sync afterwards anyway.
https://github.com/ravendb/ravendb/blob/v4.1/src/Voron/GlobalFlushingBehavior.cs#L229

- There is no point in starting a sync process before the flush process has finished.
And calling MaybeFlushEnvironment will only schedule it.

(After finishing the flush process we can delete the scratch files).

arekpalinski: - No, it won't. The line you put only filters out the duplicates. Maybe you meant different line?
The sync is run based on `_syncIsRequired` queue:

https://github.com/ravendb/ravendb/blob/v4.1/src/Voron/GlobalFlushingBehavior.cs#L161

- `ForceSyncEnvironment` doesn't start the process, it enqueues the sync:
https://github.com/ravendb/ravendb/blob/v4.1/src/Voron/GlobalFlushingBehavior.cs#L308

So both calls `MaybeFlushEnvironment` and `ForceSyncEnvironment` schedule the operations but we are guaranteed that order of operations will be _flush_ and then _sync_ because:

https://github.com/ravendb/ravendb/blob/v4.1/src/Voron/GlobalFlushingBehavior.cs#L89-L91

arekpalinski: But you're right that sync isn't required to release scratch space, only the flush.

grisha-kotler: I meant this:
https://github.com/ravendb/ravendb/blob/v4.1/src/Voron/GlobalFlushingBehavior.cs#L259

It schedules the flush and afterwards schedules the sync for the same environment.

grisha-kotler: I think that the after finishing the flush process, we should schedule a sync for that environment.

arekpalinski: After talking to Grisha I agree that `ForceSyncEnvironment` should be called as he changed it. The issue is that this guys:

https://github.com/ravendb/ravendb/blob/87367b6cc801708485f101b688a17797dcc7a6c1/src/Voron/GlobalFlushingBehavior.cs#L89-L91

just call `ThreadPool.QueueUserWorkItem`. So we don't have guarantee that for a given environment the sync will be called _after_ the flush.

src/Voron/Impl/Scratch/ScratchBufferPool.cs
```diff
@@ -500,6 +487,30 @@ public void Cleanup()
...
+            if (_recycleArea.Count == 0)
+                return;
+
+            var scratchNode = _recycleArea.First;
+            while (scratchNode != null)
+            {
+                var next = scratchNode.Next;
+
+                var recycledScratch = scratchNode.Value;
+                if (recycledScratch.File.HasActivelyUsedBytes(_env.PossibleOldestReadTransaction(null)) == false)
```
arekpalinski, line 501: Here is the fix - `HasActivelyUsedBytes` was missing

PR #7989, opened 2018-11-08
RavenDB-12286 - Early out of memory after flush

grisha-kotler:


src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -688,20 +689,26 @@ private void WaitForJournalStateToBeUpdated(CancellationToken token, Transaction
...
                             break;
                         }
+                        catch (EarlyOutOfMemoryException)
+                        {
+                            // we already finished the flush process, the next step is to update
+                            // the journal state and free up the scratch buffer pool used pages.
+                            // after this process is done we'll be able to delete the scratch buffers
+                            // that we flushed and start the sync process
+
+                            if (CanContinueWaiting(token))
```
ayende: I'm not following how this is helpful for us.
We attempted to allocated some native memory (why, I can't think of a code path here that needs it) and failed.
why are we waiting instead of just failing upward?
Is this because it is potentially expected?

grisha-kotler: We try to allocate memory when opening a write tx.
See the stacktrace here: http://issues.hibernatingrhinos.com/issue/RavenDB-12286
In this case we aren't going to do any work using this write transaction (we hold it to prevent others to open a write tx). Can we disable this low memory check here?
Failing here will result in VoronUnrecoverableErrorException to the storage environment.

ayende: If we don't need the write tx, what about simply providing API to get the write tx lock, instead?
I like it better than starting to mess with this here, and IIRC, there are other places where we start a write tx just to be able to get the lock

grisha-kotler: done

src/Voron/StorageEnvironment.cs
```diff
@@ -600,6 +601,17 @@ public Transaction WriteTransaction(ByteStringContext context = null, TimeSpan?
             return writeTransaction;
         }
 
+        public LowLevelTransaction SafeNewLowLevelTransaction(
+            TransactionPersistentContext transactionPersistentContext,
+            TransactionFlags transactionFlags,
+            TimeSpan? timeout = null)
+        {
+            using (MemoryInformation.DisableOutOfMemoryChecks())
```
arekpalinski, line 609: This way we disable the check but still do the allocation - isn't it dangerous?

arekpalinski: Providing the API as suggested in https://github.com/ravendb/ravendb/pull/7989#discussion_r232018490 is more complex but would guarantee we won't do the allocation

src/Sparrow/LowMemory/MemoryInformation.cs
```diff
@@ -80,8 +81,21 @@ public struct MemoryStatusEx
...
         public static bool EnableEarlyOutOfMemoryChecks = false; // we don't want this to run on the clients
 
+        [ThreadStatic]
+        private static bool _disableOutOfMemoryChecks;
+
+        public static IDisposable DisableOutOfMemoryChecks()
+        {
+            _disableOutOfMemoryChecks = true;
+
+            return new DisposableAction(() => _disableOutOfMemoryChecks = false);
```
arekpalinski, line 91: In Voron we typically create dedicated struct which is disposable, no need the use DisposableAction and move to to Sparrow e.g. `public struct ExitWriteLock : IDisposable`

src/Voron/StorageEnvironment.cs
grisha-kotler: We do allocate 4KB for this and we don't do any significant work (write do write a single page or just hold the write lock).

arekpalinski: My concern is: can this 4KB allocation kill? We have `EarlyOutOfMemoryException` to prevent that, if we disable it then we don't have this protection

arekpalinski: I meant *kill us

arekpalinski: Where is this allocation - `LowLevelTransaction.InitTransactionHeader` ?

arekpalinski: I wonder if we could mark that this write tx isn't going to be committed (and throw on attempt to commit) and then don't do this allocation

arekpalinski: The problem is that we have two cases where we create a write tx mostly to keep the write tx lock but we also do the commit:

1) WriteAheadJournal.WaitForJournalStateToBeUpdated

https://github.com/ravendb/ravendb/blob/54cc370bfeecf2bf1bb8bca27718a2d7d945525e/src/Voron/Impl/Journal/WriteAheadJournal.cs#L709-L713

2) ScratchBufferPool.Cleanup
https://github.com/ravendb/ravendb/blob/54cc370bfeecf2bf1bb8bca27718a2d7d945525e/src/Voron/Impl/Scratch/ScratchBufferPool.cs#L460-L463

Although in first case I think we have the commit because of possible lazy tx mode usage?
https://github.com/ravendb/ravendb/blob/54cc370bfeecf2bf1bb8bca27718a2d7d945525e/src/Voron/Impl/Journal/WriteAheadJournal.cs#L764-L767

arekpalinski: Maybe since the allocation potentially causing problems in low memory is made in `InitTransactionHeader` and we don't always need it, we use it only if we write to journal:

https://github.com/ravendb/ravendb/blob/6c7905bdcd1f5c0b84c2be0371adc68a88523a77/src/Voron/Impl/Journal/WriteAheadJournal.cs#L1467

then maybe we could do this allocation on demand (lazily) instead of calling it in the ctor of LowLevelTransaction

arekpalinski:
We'll take different approach. We'll be opening the write transaction with the usage of dedicated method which will _not_ do any allocation so we won't fail with `EarlyOutOfMemory` (actually we can return in the first line of LowLevelTransaction when we already acquired the lock) but we'll assure that attempt to do an action on this tx, in particulate the commit, will fail.

PR #7991, opened 2018-11-09
RavenDB-12290 Check why Smuggler import and export is not returning progress

ppekrol:


src/Raven.Client/Documents/Operations/Operation.cs
```diff
@@ -153,23 +153,34 @@ protected async Task FetchOperationStatus()
         {
             await _lock.WaitAsync().ConfigureAwait(false);
 
-            OperationState state;
+            OperationState state = null;
             try
             {
-                if (_result.Task.IsCompleted)
-                    return;
+                for (var i = 0; i < 10; i++)
```
ayende, line 159: Why do we have 10 retries here?

ppekrol: GetOperationStateCommand can return null when operation does not exist, and it can happen because of a race between sending an operation e.g. ImportCommand and actually registering it on the server-side as an operation. Not sure how it can be solved better.

ayende: Okay, can you add a comment there explaining this?

PR #8000, opened 2018-11-11
RavenDB-12094

aviv86:


ayende (approved):

src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -173,7 +175,7 @@ public SingleRun(DocumentDatabase database, RavenConfiguration configuration, Sc
                 ScriptEngine.SetValue("Raven_Min", new ClrFunctionInstance(ScriptEngine, Raven_Min));
                 ScriptEngine.SetValue("Raven_Max", new ClrFunctionInstance(ScriptEngine, Raven_Max));
 
-                ScriptEngine.SetValue("convertJsTimeToTimeSpanString", new ClrFunctionInstance(ScriptEngine, ConvertJsTimeToTimeSpanString));
```
ppekrol: This means that old client libraries will stop working due to the lack of this method?

src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -926,6 +976,17 @@ private static JsValue ToStringWithFormat(JsValue self, JsValue[] args)
                         num.ToString(cultureInfo);
                 }
 
+                if (args[0].IsString())
+                {
+                    if (DateTime.TryParseExact(args[0].AsString(), "o", CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind, out var date) == false)
```
ppekrol: I think that instead of 'o' we should use 'DefaultFormat.DateTimeFormatsToRead' ?

ppekrol: Maybe we should use LazyStringParser instead?

ppekrol (changes requested):

src/Raven.Server/Documents/Patch/ScriptRunner.cs
ayende: I think so, I believe we need to keep it around.

src/Raven.Server/Documents/Patch/ScriptRunner.cs
ppekrol: @aviv86 can we use LazyStringParser here?

ppekrol (approved):

PR #8001, opened 2018-11-12
RavenDB-12292: Added the ability to filter out Attachments when doing smuggler import.

garayx:


grisha-kotler:
What about import database?

ppekrol:
I've talked with @garayx we will add it there too. Also we will check if this (along with Counters) can be added to export.

garayx:
> What about import database?

done

src/Raven.Studio/typescript/models/database/tasks/exportDatabaseModel.ts
```diff
@@ -69,8 +74,14 @@ class exportDatabaseModel {
 
     private initValidation() {
         this.exportDefinitionHasIncludes = ko.pureComputed(() => {
-            return this.includeDatabaseRecord() || this.includeDocuments() || (this.includeRevisionDocuments() && this.revisionsAreConfigured()) || this.includeConflicts() ||
-                this.includeIndexes() || this.includeIdentities() || this.includeCompareExchange() || this.includeCounters();
+            if (this.includeAttachments()) {
+                this.validationMessage("Note: Attachments can only be included with documents.");
```
ml054, line 78: Can we automatically check documents when user checks attachments + avoid having generic field: validationMessage in this model?

ml054 (changes requested):
@garayx Please have a look at my comment

ppekrol (approved):

PR #8002, opened 2018-11-12
fix for RavenDB-12276

myarichuk:
Make graph queries throw when trying to fetch vertices from non-existing collection + relevant test

### DIFF
--- a/src/Raven.Server/Documents/Queries/AbstractQueryRunner.cs
+++ b/src/Raven.Server/Documents/Queries/AbstractQueryRunner.cs
@@ -33,16 +33,17 @@ namespace Raven.Server.Documents.Queries
             Database = database;
         }
 
-        public Index GetIndex(string indexName)
+        public Index GetIndex(string indexName, bool throwIfDoesNotExist = false)
         {
             var index = Database.IndexStore.GetIndex(indexName);
-            if (index == null)
+            if (index == null && !throwIfDoesNotExist)
                 IndexDoesNotExistException.ThrowFor(indexName);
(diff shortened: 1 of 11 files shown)


src/Raven.Server/Documents/Queries/Graph/QueryQueryStep.cs
```diff
@@ -54,7 +54,7 @@ public ValueTask Initialize()
                 return default;
 
             var results = _queryRunner.ExecuteQuery(new IndexQueryServerSide(_queryMetadata),
-                  _context, _resultEtag, _token);
+                  _context, _resultEtag, _token, true);
```
ayende, line 57: Use named arg here to make this clearer

ayende (changes requested):

src/Raven.Server/Documents/Queries/AbstractQueryRunner.cs
```diff
@@ -33,16 +33,17 @@ protected AbstractQueryRunner(DocumentDatabase database)
             Database = database;
         }
 
-        public Index GetIndex(string indexName)
+        public Index GetIndex(string indexName, bool throwIfDoesNotExist = false)
         {
             var index = Database.IndexStore.GetIndex(indexName);
-            if (index == null)
+            if (index == null && !throwIfDoesNotExist)
```
ppekrol, line 39: This should be the other way around and all values by default should be 'true'

myarichuk:
Will open another PR for this issue

PR #8009, opened 2018-11-13
RavenDB-12309 High CPU on single node cluster when the leader disposed

karmeli87:


src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -1770,6 +1770,8 @@ public string GetLeaderTag(bool safe = false)
                     return safe ? Volatile.Read(ref _leaderTag) : _leaderTag;
                 case RachisState.LeaderElect:
                 case RachisState.Leader:
+                    if (CurrentLeader?.Running != true)
```
ayende, line 1773: Shouldn't we have some way to recover from this?

PR #8017, opened 2018-11-14
RavenDB-12094

aviv86:


src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -175,7 +174,7 @@ public SingleRun(DocumentDatabase database, RavenConfiguration configuration, Sc
                 ScriptEngine.SetValue("Raven_Min", new ClrFunctionInstance(ScriptEngine, Raven_Min));
                 ScriptEngine.SetValue("Raven_Max", new ClrFunctionInstance(ScriptEngine, Raven_Max));
 
-                ScriptEngine.SetValue("convertJsTimeToTimeSpanString", new ClrFunctionInstance(ScriptEngine, ConvertJsTimeToTimeSpanString));
```
ayende: What happened when a 4.1 client talks to us?

aviv86: then it will fail.. i thought it was fine to drop it in 4.2 but you're right, we need to keep it

PR #8019, opened 2018-11-14
RavenDB-11970 Lets Encrypt client : retry sending request once

iftahbe:
Not relevant for 4.0.x, the LE client was replaced


ayende (commented):
What is the error that we are trying to recover from here?

iftahbe:
http://issues.hibernatingrhinos.com/issue/RavenDB-11970#focus=streamItem-67-32117-0-0

ayende:
That is a network error when the remote server refuses us.
Any idea what can be causing this?

iftahbe:
It might be because we test against the staging servers but I have no idea. That's why I was thinking to retry once when it happens.

iftahbe:
I reproduced it again with the fix, and the retry worked well.

PR #8020, opened 2018-11-14
RavenDB-12295 Unexpected ConcurrencyException when using optimistic c…

karmeli87:
…oncurrency in a cluster

Normally, when we receive a document/tombstone from replication we store it with the original change vector, so it wouldn't replicate back.

The bug occurred when we received a tombstone without having a document first and stored it with a completely new change vector.


ayende (approved):

ppekrol:
Can you check failing tests?

karmeli87:
Done :)

PR #8035, opened 2018-11-15
RavenDB-12316 support for graph query streaming

talweiss1982:


ppekrol:
Can we add a test for streaming the graph query results?

PR #8039, opened 2018-11-15
RavenDB-12227 -  Added support for compaction of fixed size trees tha…

arekpalinski:
…t aren't stored in $Root but inside regular trees. Actually we have only one such usage in RavenDB code, in map-reduce indexes (ResultsStoreTypes inside ReducePhaseTree tree), but map-reduce indexes aren't compacted anyway.

- Fixed behavior of LargeIterator - we shouldn't throw exception after iterating over all entries, that's how we behave in other iterators

### DIFF
--- a/src/Voron/Data/Fixed/FixedSizeIterators.cs
+++ b/src/Voron/Data/Fixed/FixedSizeIterators.cs
@@ -237,9 +237,6 @@ namespace Voron.Data.Fixed
             {
                 AssertNoChanges();
 
-                if (_currentPage == null)
-                    throw new InvalidOperationException("No current page was set");
-
                 while (_currentPage != null)
                 {
                     _currentPage.LastSearchPosition++;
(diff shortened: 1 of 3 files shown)


ppekrol (approved):

src/Voron/Data/Fixed/FixedSizeIterators.cs
```diff
@@ -237,9 +237,6 @@ public bool MoveNext()
             {
                 AssertNoChanges();
 
-                if (_currentPage == null)
```
ayende, line 240: why remove this?

arekpalinski: In order to _not_ throw on `MoveNext()` if we already iterated over all values. See the following test:
https://github.com/ravendb/ravendb/pull/8039/files#diff-e373029fe165f4fccd9ea5c3a9ea4b94R134

Now the behavior is consistent with our other iterators and .net framework enumerators.

It could be thrown during compaction, note `MoveNext()` calls inside `while` loops:
https://github.com/ravendb/ravendb/blob/791dec72ff727ca72ce82c6001d1774fe6448664/src/Voron/Impl/Compaction/StorageCompaction.cs#L228-L240

When first `MoveNext` call returned `false` then the second one thew this exception

PR #8045, opened 2018-11-15
[WIP] RavenDB-12115 (do not merge)

redknightlois:


src/Sparrow/Memory.cs
```diff
@@ -242,5 +246,22 @@ public static void SetInline(byte* dest, byte value, long n)
             Finish:
             ;
         }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static void Discard(void* baseAddress, long size)
+        {
+            if (PlatformDetails.CanDiscardMemory)
+            {
+                // We explicitely ignore the return codes because even if it fails, from all uses and purposes we dont care.
```
ayende, line 255: I think that, for debug only, we should error if we fail here.
That is important if we pass bad information, otherwise we might miss this entirely

src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -1606,7 +1606,12 @@ public void ReduceSizeOfCompressionBufferIfNeeded(bool forceReduce = false)
         {
             var initialSize = _env.Options.InitialFileSize ?? _env.Options.InitialLogFileSize;
             if (ShouldReduceSizeOfCompressionPager(initialSize, forceReduce) == false)
+            {
+                //PERF: Compression buffer will be reused, it is safe to discard the content to clear the modified bit.
+                _compressionPager.DiscardWholeFile();
```
ayende, line 1611: I think we _always_ want to do that, even if we are about to close the file, no?

ayende: Also, we can do this _immediately_ after writing to the journal, why wait for the next round?

src/Voron/Impl/Paging/CryptoPager.cs
```diff
@@ -80,6 +81,16 @@ public override void ReleaseAllocationInfo(byte* baseAddress, long size)
             Inner.ReleaseAllocationInfo(baseAddress, size);
         }
 
+        public override void DiscardPages(long pageNumber, int numberOfPages)
+        {
+            // REVIEW: Nothing to do here?.
```
ayende, line 86: Need to pass to inner

src/Voron/Impl/Paging/CryptoPager.cs
```diff
@@ -80,6 +81,16 @@ public override void ReleaseAllocationInfo(byte* baseAddress, long size)
...
+        }
+
+        public override void DiscardWholeFile()
+        {
+            // REVIEW: Nothing to do here?.
```
ayende, line 91: Need to pass to inner

src/Voron/Impl/Paging/Windows32BitsMemoryMapPager.cs
```diff
@@ -269,6 +269,16 @@ public override I4KbBatchWrites BatchWriter()
             return new Windows32Bit4KbBatchWrites(this);
         }
 
+        public override void DiscardPages(long pageNumber, int numberOfPages)
+        {
+            // REVIEW: Nothing to do here?.
```
ayende, line 274: We unmap things pretty fast, so that shouldn't matter, yes.

src/Voron/Impl/Paging/Windows32BitsMemoryMapPager.cs
```diff
@@ -269,6 +269,16 @@ public override I4KbBatchWrites BatchWriter()
...
+        }
+
+        public override void DiscardWholeFile()
+        {
+            // REVIEW: Nothing to do here?.
```
ayende, line 279: We unmap things pretty fast, so that shouldn't matter, yes.

src/Voron/Impl/Paging/Windows32BitsMemoryMapPager.cs
```diff
@@ -269,6 +269,16 @@ public override I4KbBatchWrites BatchWriter()
             return new Windows32Bit4KbBatchWrites(this);
         }
 
+        public override void DiscardPages(long pageNumber, int numberOfPages)
```
ayende, line 272: However, _because_ we unmap, we may want to discard before unmapping on temp data.
Not sure if we can always do that, though. We rely on map/unmap/map show the same data, so we might need to map, discard and then map.

src/Voron/Impl/Paging/AbstractPager.cs
```diff
@@ -517,6 +517,47 @@ public static bool IsKeySizeValid(int keySize)
             this._pagerState.CheckResetPrefetchTable();
         }
 
+        public virtual void DiscardPages(long pageNumber, int numberOfPages)
```
ayende, line 520: Need to override it for both 32 bits posix / windows pagers

src/Voron/Platform/Posix/Posix32BitsMemoryMapPager.cs
```diff
@@ -191,6 +191,16 @@ public override I4KbBatchWrites BatchWriter()
             return new Posix32Bit4KbBatchWrites(this);
         }
 
+        public override void DiscardPages(long pageNumber, int numberOfPages)
```
ayende, line 194: Same as the 32 bits windows pagers

src/Voron/Platform/Win32/WindowsMemoryMapPager.cs
```diff
@@ -131,7 +131,11 @@ public struct SplitValue
 
             NumberOfAllocatedPages = _totalAllocationSize / Constants.Storage.PageSize;
 
-            SetPagerState(CreatePagerState());
+            var pager = CreatePagerState();
+            if (fileAttributes.HasFlag(Win32NativeFileAttributes.Temporary))
```
ayende, line 135: Add comment here.

ayende (changes requested):

src/Sparrow/Memory.cs
arekpalinski: Maybe create an alert / performance hint then as we do if we detect "non durable file system":

https://github.com/ravendb/ravendb/blob/791dec72ff727ca72ce82c6001d1774fe6448664/src/Voron/StorageEnvironmentOptions.cs#L47

https://github.com/ravendb/ravendb/blob/791dec72ff727ca72ce82c6001d1774fe6448664/src/Raven.Server/Documents/DocumentDatabase.cs#L1353-L1363

src/Sparrow/Memory.cs
```diff
@@ -242,5 +246,22 @@ public static void SetInline(byte* dest, byte value, long n)
...
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static void Discard(void* baseAddress, long size)
+        {
+            if (PlatformDetails.CanDiscardMemory)
+            {
+                // We explicitely ignore the return codes because even if it fails, from all uses and purposes we dont care. 
+                if (PlatformDetails.RunningOnPosix)
+                {
+                    Syscall.madvise(new IntPtr(baseAddress), new UIntPtr((ulong)size), MAdvFlags.MADV_DONTNEED);
```
arekpalinski, line 258: We already have usage of `madvise` with `MADV_DONTNEED`

https://github.com/ravendb/ravendb/blob/791dec72ff727ca72ce82c6001d1774fe6448664/src/Voron/Platform/Posix/PosixAbstractPager.cs#L66-L73

Might be good to call `Memory.Discard` there then

src/Voron/Impl/Journal/WriteAheadJournal.cs
arekpalinski: @ayende If we discard this after each commit then next time we do the commit the OS will need to map the file again, won't it? Wouldn't it affect the overall commit performance?

ayende:
Discarding after every commit shouldn't remap, it just put the memory in a position for the OS to reclaim it.
We need to test on Linux to see if this page the memory again, but I hope not.
On Windows, it doesn't page the memory again, even though I'm pretty sure it _should_.

src/Sparrow/Memory.cs
redknightlois: What you mean by non durable file system?

src/Voron/Impl/Journal/WriteAheadJournal.cs
redknightlois: @arekpalinski I am still investigating the issue, but at least on Window `.DiscardVirtualMemory()` behavior suggests that is not the case.

src/Voron/Impl/Paging/Windows32BitsMemoryMapPager.cs
redknightlois: Not worth it. Per page discard seems to be too costly.

src/Sparrow/Memory.cs
redknightlois: Given latest blog on Oren's blog the behavior on Linux may not be appropriate and we sholdnt do discard there. Other Posix based system do implement it properly though.

redknightlois:
I have been running this on Windows live environments and it appears to be behaving properly. We might want to do a few other tests there, but the approach seems to be sound.

src/Sparrow/Memory.cs
ayende: Certain file systems do not support o_direct, but we still want to support then.
For example cifs in docker

src/Voron/Impl/Journal/WriteAheadJournal.cs
arekpalinski: ok

ayende:
@redknightlois Can you fix the conflicts and the comments on this PR and re-submit it?

redknightlois:
Still running this on production to figure out the kinks. I am still not satisfied

PR #8047, opened 2018-11-16
RavenDB-12268  Cleanup the recycled scratch buffers only when there i…

arekpalinski:
…s no open read transaction that is still using it

### DIFF
--- a/src/Voron/Impl/Scratch/ScratchBufferPool.cs
+++ b/src/Voron/Impl/Scratch/ScratchBufferPool.cs
@@ -450,23 +450,31 @@ namespace Voron.Impl.Scratch
                 {
                     RemoveInactiveScratches(_current);
 
-                    if (_recycleArea.Count == 0)
-                        return;
-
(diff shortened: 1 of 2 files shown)


src/Voron/Impl/Scratch/ScratchBufferPool.cs
```diff
@@ -470,6 +459,25 @@ public void Cleanup()
...
+        private void RemoveInactiveRecycledScratches()
+        {
+            if (_recycleArea.Count == 0)
+                return;
+            var scratchNode = _recycleArea.First;
+            while (scratchNode != null)
+            {
+                var next = scratchNode.Next;
+                var recycledScratch = scratchNode.Value;
+                if (recycledScratch.File.HasActivelyUsedBytes(_env.PossibleOldestReadTransaction(null)) == false)
```
arekpalinski, line 471: Here is the fix - `HasActivelyUsedBytes` was missing

PR #8051, opened 2018-11-18
RavenDB-12321: Dedicated error on schema errors added.

garayx:
Also added schema tests


ayende (approved):

src/Raven.Server/RavenServerStartup.cs
```diff
@@ -320,7 +320,8 @@ private static void MaybeSetExceptionStatusCode(HttpContext httpContext, Excepti
                 exception is DatabaseLoadTimeoutException ||
                 exception is DatabaseConcurrentLoadTimeoutException ||
                 exception is NodeIsPassiveException ||
-                exception is ClientVersionMismatchException)
+                exception is ClientVersionMismatchException ||
+                exception is SchemaErrorException)
```
ppekrol: should be DatabaseSchemaErrorException

ayende: I think this is an issue about references. Voron doesn't reference Raven.Client

ppekrol: This is in Raven.Server so it shouldn't be a problem?

PR #8065, opened 2018-11-19
RavenDB-7070 unskipping tests

ppekrol:


ppekrol:
@iftahbe can you take a look?

PR #8068, opened 2018-11-20
fix for RavenDB-12285

myarichuk:
Make sure that in graph queries after a node comes an edge, so the query (foo)-[bar]->[barfoo] should throw an appropriate error

### DIFF
--- a/test/FastTests/Graph/GraphPermissionTests.cs
+++ b/test/FastTests/Graph/GraphPermissionTests.cs
@@ -83,6 +83,7 @@ select  i.Name as Issue,
             using (var store = GetDocumentStore())
             {
                 CreateData(store);
+                WaitForUserToContinueTheTest(store);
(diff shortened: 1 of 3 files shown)


test/FastTests/Graph/GraphPermissionTests.cs
```diff
@@ -83,6 +83,7 @@ public void MultiplePathsWillAllow(string uid, string name)
             using (var store = GetDocumentStore())
             {
                 CreateData(store);
+                WaitForUserToContinueTheTest(store);
```
ayende: Please create a test for this and remove this line

ayende (changes requested):

test/FastTests/Graph/GraphPermissionTests.cs
myarichuk: *facepalm*
will do.

PR #8072, opened 2018-11-21
 RavenDB-12348 Linux - remove user related properties from ProcessStartInfo

iftahbe:


ayende (approved):

iftahbe:
Got some Windows failures now, please don't merge

iftahbe:
@ppekrol when we set bInheritHandles to false we also turned off the inheritance of the standard handles. I think we need to turn it back on and set the flags to include only STARTF_USESTDHANDLES.

See: https://docs.microsoft.com/en-us/windows/desktop/ProcThread/inheritance

ppekrol:
@iftahbe This is exactly that MS is setting when input/output/error is redirected: https://github.com/dotnet/corefx/blob/53d1d60e3dca6c9cc61cfac628b393de27fff649/src/System.Diagnostics.Process/src/System/Diagnostics/Process.Windows.cs#L501

But in order for it to work, we need to set 'inheritHandles' to 'true' which we don't want to do (we need to avoid socket inheritance).

@iftahbe @ayende Can we assume that we have two opposite requirements and we cannot do much here, except reverting to built-in Process and waiting for MS to fix the handle issue?

ayende:
Yes, I don't think that we have any way to handle this that would actually work for us, in this case.
Let's mark this as waiting for MS and close this.

ayende:
Is there any code that was already merged that we need to revert?

ppekrol:
Yes, I will do the reverts.

PR #8075, opened 2018-11-21
RavenDB-12362 Create new context every async commit

karmeli87:


karmeli87:
Note: haven't did a proper performance test yet.

karmeli87:
From my tests there is no performance hit.
Both server and client are local
Single bulk insert is ~20k doc/s
5 Parallel bulk insert tasks ~80k docs/s

ayende (approved):

PR #8078, opened 2018-11-22
RavenDB-12354 Allow to filter out Studio notifications

iftahbe:


src/Raven.Server/Config/Categories/StudioConfiguration.cs
```diff
@@ -1,13 +1,78 @@
...
     {
-        [Description("The directory into which RavenDB will search the studio files, defaults to the base directory")]
+        [Description("The directory in which RavenDB will search the studio files, defaults to the base directory")]
         [DefaultValue(null)]
         [ConfigurationEntry("Studio.Path", ConfigurationEntryScope.ServerWideOnly)]
         public string Path { get; set; }
+        
+        [Description("Semicolon seperated list of notification names which will not be shown in the Studio. If not specified, all notifications are allowed. Example: \"SlowIO;Server_NewVersionAvailable;ClusterTopologyWarning\".")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Studio.Notifications.FilterOut", ConfigurationEntryScope.ServerWideOnly)]
```
ayende: That isn't a Studio decisions, it belongs to notifications

src/Raven.Server/Config/Categories/StudioConfiguration.cs
```diff
@@ -1,13 +1,78 @@
...
+        {
+            base.Initialize(settings, serverWideSettings, type, resourceName);
+
+            if (type != ResourceType.Server)
+                return;
+
+            ValidateFilteredOutNotificationNames();
+        }
+
+        private void ValidateFilteredOutNotificationNames()
```
ayende: Let's not do that, we might change this and we don't want to break things if the user has a configuration for new version, frex.

src/Raven.Server/Config/Categories/StudioConfiguration.cs
```diff
@@ -1,13 +1,78 @@
...
-        [Description("The directory into which RavenDB will search the studio files, defaults to the base directory")]
+        [Description("The directory in which RavenDB will search the studio files, defaults to the base directory")]
         [DefaultValue(null)]
         [ConfigurationEntry("Studio.Path", ConfigurationEntryScope.ServerWideOnly)]
         public string Path { get; set; }
+        
+        [Description("Semicolon seperated list of notification names which will not be shown in the Studio. If not specified, all notifications are allowed. Example: \"SlowIO;Server_NewVersionAvailable;ClusterTopologyWarning\".")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Studio.Notifications.FilterOut", ConfigurationEntryScope.ServerWideOnly)]
+        public string[] FilteredOutNotifications { get; set; }
```
ayende: You are always comparing on this, easier if you will use `HashSet<string>` (case insensitive)

ayende (changes requested):

ppekrol (changes requested):
All of this should be moved to:
https://github.com/ravendb/ravendb/blob/v4.1/src/Raven.Client/ServerWide/Operations/Configuration/ServerWideStudioConfiguration.cs
or
https://github.com/ravendb/ravendb/blob/v4.1/src/Raven.Client/Documents/Operations/Configuration/StudioConfiguration.cs

it depends if we want to have it configurable server-wide only or database and server-wide.

After that we can add it to Studio.

arekpalinski:
Please create a test

iftahbe:
@ppekrol I don't follow why you want this. 
It is all handled server side, and the Studio is not aware of the filtered out notifications.

About the question whether it should also be per-database... I assumed that people will filter out mostly performance hints which should be server wide but if you think there are use-cases where people need this per database I can change that.

ppekrol:
I've meant that we should move `public HashSet<string> FilteredOutNotifications { get; set; }` to `StudioConfiguration.cs` and let @ml054 add this to the UI, so users can configure it via Studio when Server is running.

Filtering should be done server-side as you do it now.

ayende:
@ppekrol One of the reasons to do this is to allow the admin to shut down notifications.
This shouldn't be something that any user can edit.

I'm fine with having _additional_ items that are filtered, but the base line is fixed

ppekrol:
@iftahbe add what I've wrote about to 4.2. For 4.1 this is enough.

PR #8079, opened 2018-11-22
RavenDB-12365: Replace unmanaged memory pInvokes with Unsafe calls 

aviviadi:


src/Sparrow/Memory.cs
```diff
@@ -1,19 +1,22 @@
...
-        public const int CompareInlineVsCallThreshold = 256;
+        private const int CompareInlineVsCallThreshold = 256;
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static int Compare(byte* p1, byte* p2, int size)
         {
             return CompareInline(p1, p2, size);
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
ayende: This is explicitly meant to not be inlined

src/Sparrow/Memory.cs
```diff
@@ -1,19 +1,22 @@
...
+using Sparrow.Platform.Posix;
 
 namespace Sparrow
 {
     public static unsafe class Memory
     {
-        public const int CompareInlineVsCallThreshold = 256;
+        private const int CompareInlineVsCallThreshold = 256;
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
ayende: This is explicitly meant to not be inlined

src/Sparrow/Memory.cs
```diff
@@ -210,37 +203,58 @@ public static void Set(byte* dest, byte value, uint n)
...
         public static void Set(byte* dest, byte value, long n)
         {
-            SetInline(dest, value, n);
+            if (n < uint.MaxValue) // Common code-path
+            {
+                Set(dest, value, (uint)n);
+            }
+            else
+            {
+                for (long i = 0; i < n; i += uint.MaxValue)
```
ayende: Move to separate method

src/Sparrow/Memory.cs
```diff
@@ -167,40 +172,28 @@ public static int CompareInline(void* p1, void* p2, int size, out int position)
...
-            if (n < uint.MaxValue)
+            if (n < uint.MaxValue) // Common code-path
             {
-                Unsafe.CopyBlock(dest, src, (uint)n); // Common code-path
+                Copy(dest, src, (uint)n);
                 return;
             }
 
-            BulkCopy(dest, src, n);
+            for (long i = 0; i < n; i += uint.MaxValue)
```
ayende, line 193: Move to separate method

src/Sparrow/Memory.cs
```diff
@@ -210,37 +203,58 @@ public static void Set(byte* dest, byte value, uint n)
...
-            if (n == 0)
-                goto Finish;
-
-            if (n < int.MaxValue)
-            {
-                Unsafe.InitBlock(dest, value, (uint)n);
-            }
-            else
+            // if dest and src overlaps, we need to call specifically to memmove pinvoke supporting overlapping
+            var d = new IntPtr(dest).ToInt64();
```
ayende: Do the comparisons on the `byte*` directly, do not turn them to `int64`

src/Sparrow/Platform/Posix/Syscall.cs
```diff
@@ -260,7 +252,18 @@ public static long lseek64(int fd, long offset, WhenceFlags whence)
         public static extern int statvfs(string path, ref Statvfs buf);
 
         [DllImport(LIBC_6, SetLastError = true)]
-        public static extern int mprotect(IntPtr start, ulong size, ProtFlag protFlag);
+        public static extern int mprotect_64bit(IntPtr start, ulong size, ProtFlag protFlag);
+        
+        [DllImport(LIBC_6, SetLastError = true)]
+        public static extern int mprotect_32bit(IntPtr start, IntPtr size, ProtFlag protFlag);
```
ayende: No, this is _always_ `IntPtr`, no need for both of them. And you didn't set the name, so this would always fail

ayende (changes requested):

ayende:
This doesn't compile

ayende (approved):

PR #8080, opened 2018-11-22
RavenDB-12365: Replace unmanaged memory pInvokes with Unsafe calls

aviviadi:


ppekrol:
@aviviadi Can you fix compilation?

src/Sparrow/Platform/Win32/Win32ElectricFencedMemory.cs
```diff
@@ -31,7 +31,7 @@ internal unsafe class Win32ElectricFencedMemory
 
             var firstWritablePage = virtualAlloc + 4096;
 
-            Win32UnmanagedMemory.Set(firstWritablePage, 0xED, 4096 * sizeInPages); // don't assume zero'ed mem
```
arekpalinski, line 34: Can we remove `Win32UnmanagedMemory.cs` file now?

aviviadi: No, we still use : memcmp, memove, which doesn't have Unsafe methods and we probably do not want to implement them by ourselfs (or may be we will..?)

ayende (approved):

src/Sparrow/Platform/Win32/Win32ElectricFencedMemory.cs
arekpalinski: ok, no need to implement them on our own, we're better of by using those syscals I think

aviviadi: in 4.2 we replace memove with Buffer.MemoryCopy

PR #8082, opened 2018-11-22
Raven db 12145

talweiss1982:
WIP for review

### DIFF
--- a/src/Raven.Server/Documents/Queries/GraphQueryOrderByFieldSorter.cs
+++ b/src/Raven.Server/Documents/Queries/GraphQueryOrderByFieldSorter.cs
@@ -1,0 +1,143 @@
+﻿using System;
+using System.Collections.Generic;
+using Raven.Client.Json;
+using Sparrow.Json;
+
+namespace Raven.Server.Documents.Queries
+{
+    internal class GraphQueryOrderByFieldSorter: IComparer<GraphQueryRunner.Match>
+    {
+        private OrderByField _field;
+        private string _alias;
+        private BlittablePath _path;
+
+        public GraphQueryOrderByFieldSorter(OrderByField field)
+        {
+            var fieldName = field.Name.Value;
+            var indexOfDot = fieldName.IndexOf('.');
+            if (indexOfDot < 0)
+                throw new NotSupportedException($"{GetType().Name} got an order by field: {fieldName} that isn't in the expected format of alias.fieldName");
+            _alias = fieldName.Substring(0, indexOfDot);
+            _path = new BlittablePath(fieldName.Substring(indexOfDot + 1, fieldName.Length - indexOfDot - 1));
+            _field = field;
+        }
+
+
+
+        public int Compare(GraphQueryRunner.Match x, GraphQueryRunner.Match y)
+        {
+            int order = _field.Ascending ? 1 : -1;
+            object xObject = null;
+            object yObject = null;
+            var xResult = x.GetResult(_alias);
+            var yResult = y.GetResult(_alias);
+            switch (xResult)
+            {
+                case Document xDocument:
+                    if (yResult is Document yDocument)
+                    {
+                        xDocument.EnsureMetadata();
+                        yDocument.EnsureMetadata();
+                        xObject = _path.Evaluate(xDocument.Data, true);
+                        yObject = _path.Evaluate(yDocument.Data, true);
+                        break;
+                    }
+                    ThrowMissmatchTypes(xResult, yResult);                   
+                    break;
+                case BlittableJsonReaderObject bjroX:
+                    if (yResult is BlittableJsonReaderObject bjroY)
+                    {
+                        xObject = _path.Evaluate(bjroX, true);
+                        yObject = _path.Evaluate(bjroY, true);
+                        break;
+                    }
+                    ThrowMissmatchTypes(xResult, yResult);
+                    break;
+                case LazyStringValue xLazyStringValue:
+                    return xLazyStringValue.CompareTo(yResult) * order;
+                case string xString:
+                    if (yResult is string yString)
+                    {
+                        xObject = xString;
+                        yObject = yString;
+                    }
+                    ThrowMissmatchTypes(xResult, yResult);
+                    break;
+                case LazyNumberValue xLazyNumberValue:
+                    return xLazyNumberValue.CompareTo(yResult) * order;                
+                case int xInt:
+                    if (yResult is int yInt)
+                    {
+                        xObject = xInt;
+                        yObject = yInt;
+                    }
+                    ThrowMissmatchTypes(xResult, yResult);
+                    break;
+                case long xLong:
+                    if (yResult is long yLong)
+                    {
+                        xObject = xLong;
+                        yObject = yLong;
+                    }
+                    ThrowMissmatchTypes(xResult, yResult);
+                    break;
+                case double xDouble:
+                    if (yResult is double yDouble)
+                    {
+                        xObject = xDouble;
+                        yObject = yDouble;
+                    }
+                    ThrowMissmatchTypes(xResult, yResult);
+                    break;
+                default:
+                    throw new NotSupportedException($"Got unexpected types for compare ${xResult} and ${yResult}");
+            }
+
+            switch (xObject)
+            {
+                case LazyStringValue xLazyStringValue:
+                    return xLazyStringValue.CompareTo(yObject) * order;
+                case string xString:
+                    if (yObject is string yString)
+                    {
+                        return string.CompareOrdinal(xString, yString) * order;
+                    }
+                    ThrowMissmatchFieldsTypes(xObject, yObject);
+                    break;
+                case LazyNumberValue xLazyNumberValue:
+                    return xLazyNumberValue.CompareTo(yObject) * order;
+                case int xInt:
+                    if (yObject is int yInt)
+                    {
+                        return xInt.CompareTo(yInt) * order;
+                    }
+                    ThrowMissmatchFieldsTypes(xObject, yObject);
+                    break;
+                case long xLong:
+                    if (yObject is long yLong)
+                    {
+                        return xLong.CompareTo(yLong) * order;
+                    }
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Documents/Queries/GraphQueryOrderByFieldSorter.cs
```diff
@@ -0,0 +1,143 @@
...
+                    if (yResult is BlittableJsonReaderObject bjroY)
+                    {
+                        xObject = _path.Evaluate(bjroX, true);
+                        yObject = _path.Evaluate(bjroY, true);
+                        break;
+                    }
+                    ThrowMissmatchTypes(xResult, yResult);
+                    break;
+                case LazyStringValue xLazyStringValue:
+                    return xLazyStringValue.CompareTo(yResult) * order;
```
ayende, line 57: This is wrong, you don't know that the types match here.

src/Raven.Server/Documents/Queries/GraphQueryOrderByFieldSorter.cs
```diff
@@ -0,0 +1,143 @@
...
+                    ThrowMissmatchTypes(xResult, yResult);                   
+                    break;
+                case BlittableJsonReaderObject bjroX:
+                    if (yResult is BlittableJsonReaderObject bjroY)
+                    {
+                        xObject = _path.Evaluate(bjroX, true);
+                        yObject = _path.Evaluate(bjroY, true);
+                        break;
+                    }
+                    ThrowMissmatchTypes(xResult, yResult);
```
ayende, line 54: I wonder if we should fail here or just have a default? @ppekrol, @arekpalinski thoughts?

src/Raven.Server/Documents/Queries/GraphQueryOrderByFieldSorter.cs
```diff
@@ -0,0 +1,143 @@
...
+                case string xString:
+                    if (yResult is string yString)
+                    {
+                        xObject = xString;
+                        yObject = yString;
+                    }
+                    ThrowMissmatchTypes(xResult, yResult);
+                    break;
+                case LazyNumberValue xLazyNumberValue:
+                    return xLazyNumberValue.CompareTo(yResult) * order;
```
ayende, line 67: Need to handle type mismatch

src/Raven.Server/Documents/Queries/GraphQueryOrderByFieldSorter.cs
```diff
@@ -0,0 +1,143 @@
...
+                    ThrowMissmatchTypes(xResult, yResult);
+                    break;
+                default:
+                    throw new NotSupportedException($"Got unexpected types for compare ${xResult} and ${yResult}");
+            }
+
+            switch (xObject)
+            {
+                case LazyStringValue xLazyStringValue:
+                    return xLazyStringValue.CompareTo(yObject) * order;
```
ayende, line 99: Type mismatch

src/Raven.Server/Documents/Queries/GraphQueryOrderByFieldSorter.cs
```diff
@@ -0,0 +1,143 @@
...
+                case string xString:
+                    if (yObject is string yString)
+                    {
+                        return string.CompareOrdinal(xString, yString) * order;
+                    }
+                    ThrowMissmatchFieldsTypes(xObject, yObject);
+                    break;
+                case LazyNumberValue xLazyNumberValue:
+                    return xLazyNumberValue.CompareTo(yObject) * order;
```
ayende, line 108: type mismatch

src/Raven.Server/Documents/Queries/GraphQueryOrderByFieldSorter.cs
```diff
@@ -0,0 +1,143 @@
...
+                    break;
+                case long xLong:
+                    if (yObject is long yLong)
+                    {
+                        return xLong.CompareTo(yLong) * order;
+                    }
+                    ThrowMissmatchFieldsTypes(xObject, yObject);
+                    break;
+                case IComparable xComparable:
+                    return xComparable.CompareTo(yObject) * order;
```
ayende, line 124: Type mismatch? Not really sure how this would work, though.

src/Raven.Server/Documents/Queries/GraphQueryOrderByFieldSorter.cs
```diff
@@ -0,0 +1,143 @@
...
+                    throw new NotSupportedException($"Can't compare two object ${xObject} and ${yObject}");
+                    break;
+
+            }
+            return 1; //TODO: do actual compare here...
+        }
+
+        private void ThrowMissmatchFieldsTypes(object xObject, object yObject)
+        {
+            throw new NotSupportedException($"Can't compare two fields ${xObject} and ${yObject} of diffrent types");
```
ayende, line 135: Need to explicitly mention the types here. 
For example, double and long would _look_ the same, but be different.
And we actually do want to allow comparisions between long and double, for example.

src/Raven.Server/Documents/Queries/GraphQueryRunner.cs
```diff
@@ -134,13 +138,25 @@ public void AddNode(string key, object val)
 
                         final.AddResult(result);
                     }
+
+                    //include clause
                 }
```
ayende, line 144: Shouldn't the include clause just always be here?

src/Raven.Server/Documents/Queries/GraphQueryRunner.cs
```diff
@@ -134,13 +138,25 @@ public void AddNode(string key, object val)
...
                 return final;
             }
         }
 
+        private void Sort((List<Match> Matches, GraphQueryPlan QueryPlan) qr, OrderByField[] orderBy)
+        {
+            foreach (var field in orderBy)
+            {
+                var orderByFieldSorter = new GraphQueryOrderByFieldSorter(field);
+                qr.Matches.Sort(orderByFieldSorter);
```
ayende, line 155: Pretty sure that this won't do the right thing when you have multiple fields.

src/Raven.Server/Documents/Queries/GraphQueryRunner.cs
```diff
@@ -164,9 +180,8 @@ private async Task<(List<Match> Matches, GraphQueryPlan QueryPlan)> GetQueryResu
                             matchResults[i] = default;
                     }
                 }
-            }
-
-            return (matchResults, qp);
+            }            
+            return (matchResults.Skip(query.Start).Take(query.PageSize).ToList(), qp);
```
ayende, line 184: Let's check if we can avoid the allocation on `ToList`

src/Raven.Server/Documents/Queries/GraphQueryOrderByFieldSorter.cs
talweiss1982: i expect LazyNumber Value to throw in this case, i was trying to avoid yet another switch because the other side could be LazyNumber int long double i guess there is no avoiding it

src/Raven.Server/Documents/Queries/GraphQueryRunner.cs
talweiss1982: I think you're right I need to concatenate the compares on each item separately, good catch.

src/Raven.Server/Documents/Queries/GraphQueryRunner.cs
talweiss1982: The output is a list so I need to generate the result, do you want me to change the result to be IEnumerable (this will have a ripple affect on the code)? 
I could save the two IEnumerable calls using a while loop.

src/Raven.Server/Documents/Queries/GraphQueryOrderByFieldSorter.cs
ayende: I believe we must, yes

src/Raven.Server/Documents/Queries/GraphQueryRunner.cs
ayende: If `query.Start` is 0, then you don't need to allocate anything.
Simply set remove anything from the end of the list until it hit `PageSize`

ayende: `.ToList()` always allocated, but you usually won't need to

talweiss1982:
Updated

src/Raven.Server/Documents/Queries/GraphQueryOrderByFieldSorter.cs
arekpalinski: What would be the default? How can we deal with that if types are different? Implicitly return `false`?

I'm for throwing an exception and be explicit about the reason if failed.

ayende: The problem is that we don't have types, and a single bad document can cause the query to _fail_, which can take prod down.
We need to define the sort order and just let it go in that manner

talweiss1982: @ayende, in that case, we should maybe return -1 meaning the bad cases will always show up last I think we do the same with NULL_VALUE in v3.x
@arekpalinski there is no false here since it is a compare so it could be big small equal or throw
We could even return those values last regardless ordering so they won't show up when doing ascending or descending order.

talweiss1982:
I'm pretty sure i pushed a commit yesterday and its not showing

src/Raven.Server/Documents/Queries/GraphQueryOrderByFieldSorter.cs
ayende: I'm fine with that

talweiss1982: i'm going to open a new PR it seems my commit from yesterday didn't get in

PR #8085, opened 2018-11-23
[DO NOT MERGE} RavenDB-12373 Let's wrap array created in index definition into Dynam…

arekpalinski:
…icArray so we won't fail when executing indexing function due to casts to dynamics we add (e.g. (Func<dynamic, bool>))

PR created to run all tests on CI


ayende (commented):
This should be done by defining a `LinqOnDynamic` method that would pick this  up. 
I'm not sure why it didn't, to be honest.

arekpalinski:
Note that `let array = new string[] {"one", "two"}` isn't the collection of dynamics. The problem is `string[]` which becomes `System.Array` that doesn't have `Where()`. Please check:

http://issues.hibernatingrhinos.com/issue/RavenDB-12373#focus=streamItem-67-32349-0-0

PR #8099, opened 2018-11-25
RavenDB-12023 - Add ClrStack info

grisha-kotler:


ppekrol:
We already have this in 4.2, we are missing only graph there. Please make any additions there.

In general, you cannot attach to yourself and passive mode on live-system is returning incomplete results. Check how we are handling that there.

grisha-kotler:
You are correct. It doesn't provide the stack trace for running threads.
I also added it to the rvn tool. I'll create another PR.

ppekrol:
Ok, remember that we want same format as in Raven.Debug.

PR #8100, opened 2018-11-25
RavenDB-12346 support moving from RavenQueryInspector<T> to DocumentQuery

ppekrol:


test/SlowTests/Issues/RavenDB_12346.cs
```diff
@@ -0,0 +1,69 @@
...
+                {
+                    var docQuery = session.Advanced.DocumentQuery<User>();
+                    var q = docQuery.ToQueryable();
+                    q = q.Where(x => x.Name == "123");
+                    q = q.Search(x => x.Age, "123");
+
+                    docQuery = q.ToDocumentQuery();
+
+                    Assert.Equal(q.ToString(), docQuery.ToString());
+                    Assert.Equal("from Users where Name = $p0 and search(Age, $p1) select id() as Id, Name, LastName, AddressId, Count, Age", docQuery.ToString());
```
ppekrol, line 27: @ayende something that I have noticed. It looks like `ToQueryable` creates and artifical projection? Is this by design?

https://github.com/ravendb/ravendb/blob/eef63a81ffff8c5a496a65100111c8bda72d7c14/src/Raven.Client/Documents/Session/AsyncDocumentQuery.cs#L970

ayende: No, I wouldn't expect it.

ppekrol: I will fix it in 4.1 and merge it to 4.2 afterwards

PR #8102, opened 2018-11-26
RavenDB-12145 PR Comments + Tests + fixes

talweiss1982:


src/Raven.Server/Documents/Queries/GraphQueryOrderByFieldComparer.cs
```diff
@@ -0,0 +1,235 @@
...
+                        yObject = yLazyStringValue2;
+                        break;
+                    }
+                    return LogMissmatchTypesReturnOrder(xResult, yResult);
+                    break;
+                case LazyNumberValue xLazyNumberValue:
+                    xObject = xLazyNumberValue;
+                    switch (yResult)
+                    {
+                        case int i:
```
ayende:
```
case int _:
case long _:
  yObject = yResult;
  break;
```
Is easier

src/Raven.Server/Documents/Queries/GraphQueryOrderByFieldComparer.cs
```diff
@@ -0,0 +1,235 @@
...
+                    }
+                    return LogMissmatchTypesReturnOrder(xResult, yResult);
+                case long xLong:
+                    if (yResult is long yLong)
+                    {
+                        xObject = xLong;
+                        yObject = yLong;
+                        break;
+                    }
+                    return LogMissmatchTypesReturnOrder(xResult, yResult);
```
ayende: And if it is double, we should still support it.

src/Raven.Server/Documents/Queries/GraphQueryOrderByFieldComparer.cs
```diff
@@ -0,0 +1,235 @@
...
+                    }                                        
+                    break;
+                case int xInt:
+                    if (yResult is int yInt)
+                    {
+                        xObject = xInt;
+                        yObject = yInt;
+                        break;
+                    }
+                    return LogMissmatchTypesReturnOrder(xResult, yResult);
```
ayende: and if it is double, long?

src/Raven.Server/Documents/Queries/GraphQueryOrderByFieldComparer.cs
```diff
@@ -0,0 +1,235 @@
...
+                            return xLazyNumberValue.CompareTo(lnv) * _order;
+                        default:
+                            return LogMissmatchTypesReturnOrder(xObject, yObject);
+                    }
+                case long xLong:
+                    if (yObject is long yLong)
+                    {
+                        return xLong.CompareTo(yLong) * _order;
+                    }
+                    return LogMissmatchTypesReturnOrder(xObject, yObject);
```
ayende: double ?

src/Raven.Server/Documents/Queries/GraphQueryOrderByFieldComparer.cs
```diff
@@ -0,0 +1,235 @@
...
+                case IComparable xComparable:
+                    if(yObject.GetType() == xObject.GetType())
+                        return xComparable.CompareTo(yObject) * _order;
+                    return LogMissmatchTypesReturnOrder(xObject, yObject);
```
ayende: long -> double, etc

src/Raven.Server/Documents/Queries/GraphQueryOrderByFieldComparer.cs
```diff
@@ -0,0 +1,235 @@
...
+                case BlittableJsonReaderArray xBlittableJsonReaderArray:
+                    if (yObject is BlittableJsonReaderArray yBlittableJsonReaderArray)
+                    {
+                        return string.CompareOrdinal(xBlittableJsonReaderArray.ToString(), yBlittableJsonReaderArray.ToString());
```
ayende: You are allocating what may be very large value.
Need to recurse here.

src/Raven.Server/Documents/Queries/GraphQueryOrderByFieldComparer.cs
```diff
@@ -0,0 +1,235 @@
...
+                    }
+                    return LogMissmatchTypesReturnOrder(xObject, yObject);
+                case BlittableJsonReaderObject xBlittableJsonReaderObject:
+                    if (yObject is BlittableJsonReaderObject yBlittableJsonReaderObject)
+                    {
+                        return string.CompareOrdinal(xBlittableJsonReaderObject.ToString(), yBlittableJsonReaderObject.ToString());
```
ayende: There is no meaningful way to compare them, and I certainly don't want to use strings.
Use the `Base` as a way to get just a stable sort order.

ayende (changes requested):

src/Raven.Server/Documents/Queries/GraphQueryOrderByFieldComparer.cs
talweiss1982: its actually does the conversion

talweiss1982:
updated

ayende:
Also, check the failing tests

ayende (approved):

ayende:
There is a failing test that needs to be handled.

ayende:
There are also conflicts from my recent merge, please fix them and re-sumbit the PR

talweiss1982:
I have resolved conflicts in my PR are there new modification you made?
 Which test is failing i have passed fast test and on slow test only SQL test failed from what i recall.

arekpalinski:
Probably `SlowTests.Graph.SortGraphQueries.SortOnObjectShouldWork` test - http://jenkins.hibernatingrhinos.com/job/PR_Tests_v4.2/184/

talweiss1982:
@arekpalinski 
I was afraid this test might be flaky.
Since now we don't really do a real comparison on blittable might as well delete the test since we can't know the order of objects.

ayende:
Yes, we have a stable sort on the _existing_ system, but never between runs, so that might certainly be it.
Can you remove the test and reissue the PR?

talweiss1982:
@ayende I'm just running the test and will submit a fixed PR once they pass

talweiss1982:
acouple of tests are failing i might have done something wrong in the merge, checking.

PR #8107, opened 2018-11-27
RavenDB-12379 removed experimental from JS indexes, counters and cluster transactions

ppekrol:


ml054 (approved):
studio part LGTM

src/Raven.Studio/wwwroot/App/views/wizard/welcome.html
```diff
@@ -101,9 +101,7 @@ <h2 class="text-center margin-bottom margin-lg text-warning">UNSECURE</h2>
         </div>
         
         <ul class="margin-left">
-            <li>JavaScript Indexes</li>
-            <li>Counters</li>
-            <li>Cluster-wide Transactions</li>
+            <li>Graph API</li>
```
ayende, line 104: We should add actual checks on graph queries to the code. @talweiss1982

PR #8110, opened 2018-11-27
RavenDB-12364 Support Raven.Client in .net framework that uses aspnetcor

arekpalinski:
http://issues.hibernatingrhinos.com/issue/RavenDB-12364

### DIFF
--- a/Raven.Client.Lightweight/Connection/Async/AsyncServerClient.cs
+++ b/Raven.Client.Lightweight/Connection/Async/AsyncServerClient.cs
@@ -140,7 +140,7 @@ namespace Raven.Client.Connection.Async
                         () => Tuple.Create(DateTime.UtcNow, UpdateTopologyAsync())
                         ));
                 }
-                if (val.Value.Item1.Add(convention.TimeToWaitBetweenReplicationTopologyUpdates) <= DateTime.UtcNow)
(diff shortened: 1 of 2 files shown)


talweiss1982:
@arekpalinski are you sure that all requests are going through the ExecuteWithReplication method (i don't remember)?
The issue is that we might have a wrong request executor between the constructor of the AsyncServerClient and the first execute with replication since the executor is selected according to the topology.
The original issue was that we always started with replication executor and if we found out in an async manner that we are actually running as a cluster node we would switch executors midway leaving the AsyncServerClient in a faulty state where leader node was null and we expected it to be set and such errors.

arekpalinski:
I'm not 100% sure this is a valid change. I'm waiting for tests.

All of the public methods go through `ExecuteWithReplication` and this is the only place where `RequestExecuter.ExecuteOperationAsync` is called.

arekpalinski:
@talweiss1982 
Tests seems to be fine (I've just pushed a commit fixing one test).

As I mentioned `RequestExecuter.ExecuteOperationAsync` is called only in `ExecuteWithReplication` so I think it's enough to ensure that `UpdateTopologyAsync` is completed before calling `RequestExecuter.ExecuteOperationAsync`.

Although, please pull my changes and review the code. You have much wider view on the clustering stuff and how this change could affect it.

talweiss1982:
@arekpalinski I have pulled your code yesterday and it looks legit.

arekpalinski:
OK. Thank you. I'm merging it.

PR #8111, opened 2018-11-27
RavenDB-12355 limit database paths

iftahbe:


ppekrol (approved):

src/Voron/Util/Settings/PathSettingBase.cs
```diff
@@ -94,5 +94,25 @@ public static string ToFullPath(string inputPath, string baseDataDirFullPath)
...
+        {
+            var rootDirInfo = new DirectoryInfo(rootPath);
+            var userDirInfo = new DirectoryInfo(userPath);
+
+            if (userDirInfo.FullName == rootDirInfo.FullName)
+                return true;
+
+            while (userDirInfo.Parent != null)
+            {
+                if (userDirInfo.Parent.FullName == rootDirInfo.FullName)
```
ayende: I wonder if we need to make this a case insensitive comparison (on Windows, at least?)

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -227,6 +230,17 @@ public async Task Put()
                     throw new ArgumentException("DatabaseName property has invalid value (null, empty or whitespace only)");
                 databaseRecord.DatabaseName = databaseRecord.DatabaseName.Trim();
 
+                if (ServerStore.Configuration.Core.EnforceDataDirectoryPath 
+                    && databaseRecord.Settings.TryGetValue("DataDir", out var dir))
```
ppekrol: Use `RavenConfiguration.GetKey(x => x.Core.DataDir)`

ayende (approved):

ayende:
The test failure is a check to show that a all admin routes have `admin` in them. Need to handle that somehow.
I don't think we have any backward compatibility concerns here, though.

ppekrol:
This is only used by studio so it can be changed to `/admin/studio-tasks/full-data-directory` without a problem.

PR #8116, opened 2018-11-28
RavenDB-11650

ayende:
- If user specifies limit / offset in the query itself and in the query string (common in studio):
    - the limit will be the minimum between the querystring variable and the limit in the query.
    - the offset will be the offset from the query PLUS the offset in the query string. This is to allow paging in the studio of queries with offset in them.
- Properly apply limit offset clause

### DIFF
--- a/src/Raven.Client/Documents/Queries/QueryResult.cs
+++ b/src/Raven.Client/Documents/Queries/QueryResult.cs
@@ -20,6 +20,12 @@ namespace Raven.Client.Documents.Queries
         /// </summary>
         public int TotalResults { get; set; }
 
+       /// <summary>
+       /// The total results for the query, taking into account the 
+       /// offset / limit clauses for this query
+       /// </summary>
+        public int TotalResultsWithOffsetAndLimit { get; set; }
+
         /// <summary>
         /// Gets or sets the skipped results
         /// </summary>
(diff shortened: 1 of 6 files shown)


ml054:
I can do studio part, once you add extra property which tells about total results with regards to RQL limit, offset.

talweiss1982:
@ayende I have made those changes myself but @ppekrol didn't agree with this logic

arekpalinski:
@talweiss1982 Can you link your PR or a commit? What were Pawel's comments there?

The logic here looks legit although Marcin noticed some problem - http://issues.hibernatingrhinos.com/issue/RavenDB-11650#focus=streamItem-67-32466-0-0 We'll have a look at this shortly

arekpalinski:
After discussing it with Tal we agreed those changes are fine. Marcin converted `TotalResultsWithOffsetAndLimit` to be nullable and writing it to the response only if `!= null` so he can distinguish properly in the studio. His PR: https://github.com/ravendb/ravendb/pull/8131 Although he's currently checking how it works with fanout indexes.

ayende:
@talweiss1982 The reason for this is that the impl we started with had a lot of issues for usage in the stuido

arekpalinski:
Closing this. It will be merged in https://github.com/ravendb/ravendb/pull/8131

PR #8119, opened 2018-11-28
In graph queries, throw error if the collection doesn't exist (RavenD…

ayende:
…B-12276)

Also, fix affected tests

### DIFF
--- a/src/Raven.Server/Documents/Queries/Graph/QueryQueryStep.cs
+++ b/src/Raven.Server/Documents/Queries/Graph/QueryQueryStep.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Primitives;
+using Raven.Client.Exceptions;
 using Raven.Server.Documents.Queries.AST;
 using Raven.Server.ServerWide;
 using Raven.Server.ServerWide.Context;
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Documents/Queries/Graph/QueryQueryStep.cs
```diff
@@ -40,6 +41,18 @@ public class QueryQueryStep : IGraphQueryStep
             _context = documentsContext;
             _resultEtag = existingResultEtag;
             _token = token;
+
+            if (!string.IsNullOrEmpty(queryMetadata.CollectionName)) //not a '_' collection
```
arekpalinski, line 45: I'm probably missing something but how this can filter out '_' collection as stated in the comment?

test/FastTests/Graph/BasicGraphQueries.cs
```diff
@@ -444,6 +444,18 @@ public void CanHandleFiltersInMultiHopQueryWithEndNode()
...
         }
 
+        [Fact]
+        public void Query_with_non_existing_collection_should_fail()
+        {
+            using (var store = GetDocumentStore())
+            {
+                using (var session = store.OpenSession())
+                {
+                    Assert.Throws<RavenException>(() => session.Advanced.RawQuery<JObject>(@"match (FooBar)").ToList());
```
arekpalinski: Shouldn we check exception message to ensure we got the exception that we really expected?

arekpalinski: BTW this assertion fails :)
`Expected: typeof(Raven.Client.Exceptions.RavenException)\r\nActual:   typeof(Raven.Client.Exceptions.InvalidQueryException): `

src/Raven.Server/Documents/Queries/Graph/QueryQueryStep.cs
ayende: That is handled by the query parser, if we see `_`, it is sent as `null`

arekpalinski: OK, although it's a bit tricky. Maybe usage of `_` deserves a dedicated property in QueryMetadata.

ayende: That would result in "index / all / collection" options that we'll always need to check.
I don't really like that

PR #8121, opened 2018-11-28
RavenDB-12378

aviv86:
http://issues.hibernatingrhinos.com/issue/RavenDB-12378

### DIFF
--- a/src/Raven.Server/Documents/Patch/ScriptRunner.cs
+++ b/src/Raven.Server/Documents/Patch/ScriptRunner.cs
@@ -23,6 +23,7 @@ using Raven.Client.Extensions;
 using Raven.Server.Config;
 using Raven.Server.Config.Categories;
 using Raven.Server.Documents.Indexes;
+using Raven.Server.Documents.Queries;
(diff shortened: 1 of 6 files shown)


src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -1122,6 +1120,22 @@ public ScriptRunnerResult Run(JsonOperationContext jsonCtx, DocumentsOperationCo
...
             }
 
+            private void SetArgs(JsonOperationContext jsonCtx, string method, object[] args)
+            {
+                if (_args.Length != args.Length)
+                    _args = new JsValue[args.Length];
+                for (var i = 0; i < args.Length; i++)
+                    _args[i] = TranslateToJs(ScriptEngine, jsonCtx, args[i]);
+
+                if (method != "__selectOutput" &&
```
arekpalinski: Don't we have already a constant for that? If not, can we introduce it?

aviv86: done

PR #8127, opened 2018-11-28
RavenDB-12393 Fixing memory leak in BlittableMetadataModifier usage

talweiss1982:
* Moved responsibility of clearing allocation of the modifier to the context

### DIFF
--- a/src/Raven.Server/Documents/BlittableMetadataModifier.cs
+++ b/src/Raven.Server/Documents/BlittableMetadataModifier.cs
@@ -24,6 +24,15 @@ namespace Raven.Server.Documents
             _ctx = context;
         }
 
+        public BlittableMetadataModifier(JsonOperationContext context, bool legacyImport, bool readLegacyEtag, DatabaseItemType operateOnTypes)
+        {
+            _ctx = context;
+            ReadFirstEtagOfLegacyRevision = legacyImport;
+            ReadLegacyEtag = readLegacyEtag;
+            OperateOnTypes = operateOnTypes;
+
+
+        }
         public LazyStringValue Id;
         public string ChangeVector;
         public DocumentFlags Flags;
--- a/src/Raven.Server/Smuggler/Documents/StreamSource.cs
+++ b/src/Raven.Server/Smuggler/Documents/StreamSource.cs
@@ -441,7 +441,7 @@ namespace Raven.Server.Smuggler.Documents
             if (_state.CurrentTokenType != JsonParserToken.StartObject)
                 UnmanagedJsonParserHelper.ThrowInvalidJson("Expected start object, got " + _state.CurrentTokenType, _peepingTomStream, _parser);
 
-            using (var builder = CreateBuilder(_context, null))
+            using (var builder = CreateBuilder(_context))
             {
                 _context.CachedProperties.NewDocument();
                 ReadObject(builder);
@@ -530,7 +530,7 @@ namespace Raven.Server.Smuggler.Documents
             if (_state.CurrentTokenType != JsonParserToken.StartArray)
                 UnmanagedJsonParserHelper.ThrowInvalidJson("Expected start array, got " + _state.CurrentTokenType, _peepingTomStream, _parser);
 
-            var builder = CreateBuilder(_context, null);
+            var builder = CreateBuilder(_context);
             try
             {
                 while (true)
@@ -547,7 +547,7 @@ namespace Raven.Server.Smuggler.Documents
                         if (_context != oldContext)
                         {
                             builder.Dispose();
-                            builder = CreateBuilder(context, null);
+                            builder = CreateBuilder(context);
                         }
                     }
                     builder.Renew("import/object", BlittableJsonDocumentBuilder.UsageMode.ToDisk);
@@ -597,8 +597,7 @@ namespace Raven.Server.Smuggler.Documents
                 UnmanagedJsonParserHelper.ThrowInvalidJson("Expected start array, but got " + _state.CurrentTokenType, _peepingTomStream, _parser);
 
             var context = _context;
-            var modifier = new BlittableMetadataModifier(context);
-            var builder = CreateBuilder(context, modifier);
+            var builder = CreateBuilder(context, new BlittableMetadataModifier(context));
             try
             {
                 while (true)
@@ -616,7 +615,7 @@ namespace Raven.Server.Smuggler.Documents
                         if (oldContext != context)
                         {
                             builder.Dispose();
-                            builder = CreateBuilder(context, modifier);
+                            builder = CreateBuilder(context, new BlittableMetadataModifier(context));
                         }
                     }
                     builder.Renew("import/object", BlittableJsonDocumentBuilder.UsageMode.ToDisk);
@@ -708,12 +707,7 @@ namespace Raven.Server.Smuggler.Documents
 
             var context = _context;
             var legacyImport = _buildVersionType == BuildVersionType.V3;
-            var modifier = new BlittableMetadataModifier(context)
-            {
-                ReadFirstEtagOfLegacyRevision = legacyImport,
-                ReadLegacyEtag = _readLegacyEtag,
-                OperateOnTypes = _operateOnTypes
-            };
+            var modifier = new BlittableMetadataModifier(context, legacyImport, _readLegacyEtag, _operateOnTypes);
             var builder = CreateBuilder(context, modifier);
             try
             {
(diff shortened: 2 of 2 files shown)


src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -1356,6 +1364,7 @@ public void Dispose()
         }
 
         private Dictionary<Type, (Action<Array> Releaser, List<Array> Array)> _pooledArrays = null;
+        private IBlittableDocumentModifier _modifier;
```
arekpalinski: Shouldn't this be a list? What if someone would call `context.CreateBlittableMetadataModifier` twice on the same `context`? Then we won't clear allocations of the first one.

arekpalinski (commented):
I think we need to apply this fix to 4.0 and 4.1. Then we'll do the merge to 4.2 from 4.1

src/Sparrow/Json/JsonOperationContext.cs
talweiss1982: @arekpalinski there are plenty of places in the code where this happens, let me explain the reasoning for not clearing it.
At first i thought i should throw if the merger is not null or dispose of the old one but context might be reset in two cases once when it is returned to the pull and second when somebody is manualy invoking reset (and in this case it might not be correct to replace the merger) i left the responsibility to whom ever writes the code. 
there are only 3 places in our code using a modifier there was a lot of consideration not to shake things up to strongly.

talweiss1982:
@arekpalinski regaring this fixed needing to be back port into v4.0 and v4.1, i started working on v4.2 since the issue was found on v4.2 and i felt it is more safe to introduce changes to v4.2 and if everything works backport to v4.0 and v4.1, do you agree or should i have worked on v4.0 from the start?

src/Raven.Server/Documents/Handlers/BatchRequestParser.cs
```diff
@@ -665,7 +665,7 @@ private static void ThrowMissingNameProperty()
             BlittableJsonReaderObject reader;
             using (var builder = new BlittableJsonDocumentBuilder(ctx,
                 BlittableJsonDocumentBuilder.UsageMode.ToDisk,
-                id, parser, state, modifier: new BlittableMetadataModifier(ctx)))
+                id, parser, state, modifier: ctx.CreateBlittableMetadataModifier(c=>new BlittableMetadataModifier(c))))
```
ayende: This is odd, why have a delegate here?

src/Raven.Server/Smuggler/Documents/StreamSource.cs
```diff
@@ -708,12 +709,13 @@ private IEnumerable<DocumentItem> ReadDocuments(INewDocumentActions actions = nu
 
             var context = _context;
             var legacyImport = _buildVersionType == BuildVersionType.V3;
-            var modifier = new BlittableMetadataModifier(context)
+            var modifier = (BlittableMetadataModifier)context.CreateBlittableMetadataModifier(c => new BlittableMetadataModifier(context)
```
ayende: If you need the options, better to pass them explicitly, rather than use the lambda.
The issue is that we are usually using this in hot spots.

ayende: Better yet, after you get the modifier from the context, just set these properties directly.

talweiss1982: There is a huge mess regarding modifiers and builders and in order i didn't want to go too deep into cleaning the codebase since the risk of creating a new bug is big. I'll change the code to invoke a member function of the modifier to reduce the cost of a lambda call.
I'll open an issue for the future to clean things up, builders should be created from the context and the context should clean everything related to builders.

talweiss1982: I'll replace this with register for cleanup

src/Raven.Server/Smuggler/Documents/StreamSource.cs
```diff
@@ -708,12 +709,13 @@ private IEnumerable<DocumentItem> ReadDocuments(INewDocumentActions actions = nu
 
             var context = _context;
             var legacyImport = _buildVersionType == BuildVersionType.V3;
-            var modifier = new BlittableMetadataModifier(context)
+            var modifier = (BlittableMetadataModifier)context.RegisterModifierForCleanup(new BlittableMetadataModifier(context)
```
ayende: Why do we have the `IBlittableMetadataModifier`, then? If there is just one impl?

src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -958,6 +964,13 @@ protected internal virtual void Renew()
 
         protected internal virtual unsafe void Reset(bool forceReleaseLongLivedAllocator = false)
         {
+            //TODO: Maybe not clearing allocation since we are going to reset the arena anyway and this is redundent and costly?
```
ayende: I think that this is a good idea, yes

ayende (changes requested):

src/Raven.Server/Smuggler/Documents/StreamSource.cs
talweiss1982: This is an excellent question

src/Sparrow/Json/JsonOperationContext.cs
talweiss1982: If this is the case then we don't really need to do anything but generate a new modifier per builder.
Maybe we should just remove the interface generate a new modifier and that's all.

ayende: Fine by me. Can you check if we can make this a struct, to avoid the allocation per doc?

src/Raven.Server/Smuggler/Documents/StreamSource.cs
talweiss1982: okay the reason that this is an interface is because the builder reside in sparrow and the modifier is in the server so this is just for decoupling purpeses

src/Sparrow/Json/JsonOperationContext.cs
talweiss1982: modifying it to a struct breaks the code since it is passed as a parameter

talweiss1982:
I think that there is a need for bigger refactoring of how we work with builders but i think this should be out of the scope of this issue, I'll open a new issue for future refactoring.

src/Raven.Server/Documents/BlittableMetadataModifier.cs
```diff
@@ -24,6 +24,15 @@ public BlittableMetadataModifier(JsonOperationContext context)
             _ctx = context;
         }
 
+        public BlittableMetadataModifier(JsonOperationContext context, bool legacyImport, bool readLegacyEtag, DatabaseItemType operateOnTypes)
```
arekpalinski, line 27: Any particular reason to create a dedicated constructor for that? Why not to leave it as it was - create an instance of the modifier and set those options then

src/Raven.Server/Smuggler/Documents/StreamSource.cs
```diff
@@ -1084,6 +1079,12 @@ private BlittableJsonDocumentBuilder CreateBuilder(JsonOperationContext context,
                 modifier: modifier);
         }
 
+        private BlittableJsonDocumentBuilder CreateBuilder(JsonOperationContext context)
+        {
+            return new BlittableJsonDocumentBuilder(context,
```
arekpalinski, line 1084: No need to create a dedicated method. Just change the signature of the already existing method:
`private BlittableJsonDocumentBuilder CreateBuilder(JsonOperationContext context, BlittableMetadataModifier modifier)`

so `modifier` could be optional - `modifier = null`

src/Raven.Server/Smuggler/Documents/StreamSource.cs
redknightlois: I always wanted to kill `IBlittableMetadataModifier` but I was reluctant to do so because I didnt fully grasp why it was there in the first place. Probably a good idea to look into that because that causes slowdowns down the chain.

src/Sparrow/Json/JsonOperationContext.cs
redknightlois: A technique similar to what I do on the Allocators branch could probably works.

src/Raven.Server/Documents/BlittableMetadataModifier.cs
maximburyak: let's leave it to the greater refactoring issue

src/Raven.Server/Smuggler/Documents/StreamSource.cs
maximburyak: same

arekpalinski:
We need that for 4.0 / 4.1

PR #8128, opened 2018-11-28
RavenDB-12337: Export CSV exports the stored fields now.

garayx:


src/Raven.Studio/typescript/viewmodels/database/query/query.ts
```diff
@@ -1044,11 +1044,16 @@ class query extends viewModelBase {
         const args = {
             format: "csv",
         };
-
-        const payload = {
-            Query: this.criteria().queryText()
-        };
-
+        let payload: any;
```
grisha-kotler: let payload: { Query: string }

ml054: const won't work ;-)

```
const queryText = this.criteria().queryText();
const query = this.criteria().showFields() ? queryUtil.replaceSelectAndIncludeWithFetchAllStoredFields(queryText) : queryText;

const payload = {
   Query: query
}
```

ml054 (approved):

PR #8129, opened 2018-11-28
RavenDB-12406 - Fix socket bind error on Linux

grisha-kotler:
http://issues.hibernatingrhinos.com/issue/RavenDB-12406

### DIFF
--- a/src/Raven.Server/Program.cs
+++ b/src/Raven.Server/Program.cs
@@ -200,11 +200,13 @@ namespace Raven.Server
                                     $"2) Run the server from the command line with --ServerUrl option.{Environment.NewLine}" +
                                     $"3) Add RAVEN_ServerUrl to the Environment Variables.{Environment.NewLine}" +
                                     "For more information go to https://ravendb.net/l/EJS81M/4.1";
-                            }else if (e is SocketException && PlatformDetails.RunningOnPosix)
+                            }
+                            else if (e is SocketException && PlatformDetails.RunningOnPosix)
                             {
                                 message =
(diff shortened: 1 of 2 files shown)


arekpalinski (approved):

src/Raven.Server/Program.cs
```diff
@@ -200,11 +200,13 @@ public static int Main(string[] args)
...
-                            }else if (e is SocketException && PlatformDetails.RunningOnPosix)
+                            }
+                            else if (e is SocketException && PlatformDetails.RunningOnPosix)
                             {
                                 message =
-                                    $"{Environment.NewLine}In Linux low-level port (below 1024) will need a special permission, if this is your case please run{Environment.NewLine}" +
-                                    $"sudo setcap CAP_NET_BIND_SERVICE=+eip {typeof(RavenServer).Assembly.Location}";
+                                    $"{Environment.NewLine}In Linux low-level port (below 1024) will need a special permission, " +
+                                    $"if this is your case please run{Environment.NewLine}" +
+                                    $"sudo setcap CAP_NET_BIND_SERVICE=+eip Raven.Server";
```
ppekrol, line 209: @grisha-kotler Is this correct? The `{typeof(RavenServer).Assembly.Location}` returns full path and I understand that `.dll` should be removed, but the rest of the path should still be there? When installing on our servers, we are passing full path there.

grisha-kotler: fixed in https://github.com/ravendb/ravendb/pull/8254

PR #8130, opened 2018-11-28
RavenDB-12396: Report will now include a histogram of tree balance depth

redknightlois:


src/Voron/Debugging/StorageReportGenerator.cs
```diff
@@ -210,19 +211,20 @@ public static TreeReport GetReport(FixedSizeTree fst, bool calculateExactSizes)
...
+        public static TreeReport GetReport(Tree tree, bool includeDetails)
         {
             List<double> pageDensities = null;
-
-            if (calculateExactSizes)
+            Dictionary<int, int> pageBalance = null;
+            if (includeDetails)
             {
                 pageDensities = GetPageDensities(tree);
+                pageBalance = GatherBalanceDistribution(tree);
```
arekpalinski, line 227: I wonder if we could calculate both page densities and balance distribution at once and this way iterate over all pages of a tree one time

redknightlois: No difference, the execution time is dominated hard by hard page faults. On codealike database it takes a minute the first run. Second is under a second. Also I was under the impression that the details enabled is not usually call.

arekpalinski (approved):

ayende (approved):

redknightlois:
Side question, I could optimize the process because I understand the underlying behavior. Would it make sense to do so as an additional ticket?

ayende:
What do you mean? How often do you think we'll be calling something like this?

redknightlois:
Not at all, I use it for debugging... but I dont know if it has any other use.

ayende:
no point in optimizing it if it is fast enough that on your scenario it takes only a minute

PR #8132, opened 2018-11-29
RavenDB-12023 - allow get a specific thread stack trace

grisha-kotler:


src/Raven.Server/Documents/Handlers/Debugging/ThreadsHandler.cs
```diff
@@ -13,6 +15,23 @@ namespace Raven.Server.Documents.Handlers.Debugging
 {
     public class ThreadsHandler : RequestHandler
     {
+        [RavenAction("/admin/debug/threads/stack-trace", "GET", AuthorizationStatus.Operator, IsDebugInformationEndpoint = true)]
```
arekpalinski: `IsDebugInformationEndpoint = true` will cause that it will be _always_ included in the debug package. In 3.x we had it as an option because it can be time consuming and pause the server process.

src/Raven.Server/Documents/Handlers/Debugging/ThreadsHandler.cs
```diff
@@ -94,5 +113,57 @@ internal class ThreadInfo
...
+                        RedirectStandardError = true,
+                        RedirectStandardOutput = true,
+                        UseShellExecute = false
+                    },
+                    EnableRaisingEvents = true
+                };
+
+                sb.Clear();
+                process.OutputDataReceived += (sender, args) => sw.Write(args.Data);
+                process.ErrorDataReceived += (sender, args) => sb.Append(args.Data);
```
arekpalinski, line 158: `sb.Append` on error? Did you mean `sw.Write`?

src/Raven.Server/Documents/Handlers/Debugging/ThreadsHandler.cs
```diff
@@ -94,5 +113,57 @@ internal class ThreadInfo
...
+
+            using (var currentProcess = Process.GetCurrentProcess())
+            {
+                var sb = new StringBuilder($"stack-traces --pid {currentProcess.Id}");
+
+                if (threadIds != null && threadIds.Count > 0)
+                {
+                    foreach (var threadId in threadIds)
+                    {
+                        sb.Append($" --tid {threadId}");
```
arekpalinski, line 134: Below you parse the values to ensure they are integers. We should have the same here I think

tools/Raven.Debug/CommandLineApp.cs
```diff
@@ -76,9 +93,11 @@ public static int Run(string[] args)
                     if (outputOption.HasValue())
                         output = outputOption.Value();
 
+                    var includeStackObjects = includeStackObjectsOption.Values.FirstOrDefault() == "on";
```
arekpalinski, line 96: It's case sensitive, on purpose?

src/Raven.Server/Documents/Handlers/Debugging/ThreadsHandler.cs
ayende: We shouldn't include it by default.
If we 'll stop the server to capture it, this can hurt production

src/Raven.Server/Documents/Handlers/Debugging/ThreadsHandler.cs
ayende: Do we need to worry about concurrency here?

src/Raven.Server/Documents/Handlers/Debugging/ThreadsHandler.cs
grisha-kotler: done

tools/Raven.Debug/CommandLineApp.cs
grisha-kotler: this is the result when you pass a command line option with no value:
Raven.Migrator --includeStackObjects

src/Raven.Server/Documents/Handlers/Debugging/ThreadsHandler.cs
grisha-kotler: done

src/Raven.Server/Documents/Handlers/Debugging/ThreadsHandler.cs
grisha-kotler: I meant to use the sb to throw the error message. Fixed.

src/Raven.Server/Documents/Handlers/Debugging/ThreadsHandler.cs
```diff
@@ -15,7 +15,7 @@ namespace Raven.Server.Documents.Handlers.Debugging
 {
     public class ThreadsHandler : RequestHandler
     {
-        [RavenAction("/admin/debug/threads/stack-trace", "GET", AuthorizationStatus.Operator, IsDebugInformationEndpoint = true)]
```
arekpalinski: Now we will never include it. The debug package endpoint should provide an option whether we want this one or not, like in 3.x

grisha-kotler:
The debug package is getting the stack traces from a different code path.
This endpoint is for getting a specific thread stack trace and will be used for the studio.

src/Raven.Server/Documents/Handlers/Debugging/ThreadsHandler.cs
arekpalinski: WRT https://github.com/ravendb/ravendb/pull/8132#issuecomment-443143720

Okay, that's fine then. Can you point where do we create stacktaces for the debug package - I wasn't aware we already have it.

PR #8133, opened 2018-11-29
RavenDB-12404 Improve Counters Batch performance

karmeli87:
- Try to aggrigate counters by doc id.
- Import documents removes @counter in order to be consistent on counters import.
- On counter inc/put don't iterate over all the existing counters on disk, instead rely on the document's metadata.
- Keep the counters sorted by name.

### DIFF
--- a/src/Raven.Server/Documents/CountersStorage.cs
+++ b/src/Raven.Server/Documents/CountersStorage.cs
@@ -409,12 +409,12 @@ namespace Raven.Server.Documents
 
             using (GetCounterPartialKey(context, docId, out var key))
             {
-                var prev = string.Empty;
+                LazyStringValue prev = null;
                 foreach (var result in table.SeekByPrimaryKeyPrefix(key, Slices.Empty, 0))
                 {
                     var current = ExtractCounterName(context, result.Value.Reader);
 
-                    if (prev.Equals(current))
+                    if (prev?.Equals(current) == true)
                     {
                         // already seen this one, skip it 
                         continue;
@@ -422,6 +422,7 @@ namespace Raven.Server.Documents
 
                     yield return current;
 
+                    prev?.Dispose();
                     prev = current;
                 }
             }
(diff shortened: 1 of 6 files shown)


src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -676,61 +677,72 @@ public long GetNumberOfCounterEntries(DocumentsOperationContext context)
...
+            if (metadata == null)
             {
-                if (metadataCounters == null)
+                doc.Modifications[Constants.Documents.Metadata.Key] = new DynamicJsonValue
                 {
-                    return;
-                }
-
-                metadata.Modifications = new DynamicJsonValue(metadata);
-                metadata.Modifications.Remove(Constants.Documents.Metadata.Counters);
```
arekpalinski, line 698: I think you missed `Remove(Constants.Documents.Metadata.Counters)` if `counters.Count == 0 `

src/Raven.Server/Documents/Handlers/CountersHandler.cs
```diff
@@ -83,30 +83,33 @@ public ExecuteCounterBatchCommand(DocumentDatabase database)
...
+                existing.Add(op);
             }
 
             protected override int ExecuteCmd(DocumentsOperationContext context)
             {
                 if (_database.ServerStore.Server.Configuration.Core.FeaturesAvailability == FeaturesAvailability.Stable)
                     FeaturesAvailabilityException.Throw("Counters");
 
+                var ops = 0;
+                var countersToAdd = new SortedSet<string>(StringComparer.OrdinalIgnoreCase);
```
arekpalinski, line 105: Why do we want to have them sorted?

src/Raven.Server/Documents/CountersStorage.cs
karmeli87: Right. Done.

src/Raven.Server/Documents/Handlers/CountersHandler.cs
karmeli87: This is a requirement.

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
```diff
@@ -696,8 +696,9 @@ private void AddToBatch(CounterDetail counter)
                     ChangeVector = counter.ChangeVector
                 };
 
-                _cmd.Add(counter.DocumentId, counterOp);
-                _batchSize++;
+                _cmd.Add(counter.DocumentId, counterOp, out var isNew);
+                if (isNew)
+                    _batchSize++;
```
aviv86: does it mean that a batch will contain counters of 1024 different documents, with an unlimited amount of counters per doc ?

ayende: No, each 1024 counters. But I think that we can safely increase the size back to 10,000, since this is now much cheaper, no?

src/Raven.Server/Documents/Handlers/CountersHandler.cs
ayende: did you push the code? I'm not seeing this removed?

src/Raven.Server/Documents/CountersStorage.cs
arekpalinski: Reminder - no fix was pushed

src/Raven.Server/Documents/Handlers/CountersHandler.cs
grisha-kotler: The list of counters was sorted in the previous implementation. I think that we use that when patching a document with a counter.

ayende: IIRC, we need that for searching if a counter exists in the document as well

karmeli87:
With this PR `CanMigrateFromRavenDb` test should fail and will be address in http://issues.hibernatingrhinos.com/issue/RavenDB-12414

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
aviv86: Are you sure? it looks like `isNew` is set to `true` only when the document id is new to this batch. Or am I missing something?

karmeli87: To improve performance we are trying to aggregate as many counters for the same document as possible. And it is indeed unbounded.. should I limit it, or simply revert this change completely?

ayende: Let's limit this to a high number, such as 32*1024 or something like this.

ayende (commented):
Note, still missing fixes

arekpalinski:
@karmeli87 Could you explain why those changes made `CanMigrateFromRavenDb` test fail? Where is the change which changed the import behavior?

karmeli87:
On documents import we remove the `@counter` metadata.
So when we start to import the counters it will generate this metadata property and therefore create a new revision.

arekpalinski (approved):

grisha-kotler (approved):

PR #8139, opened 2018-12-01
RavenDB-12364 If we are running in transaction scope and topology upd…

arekpalinski:
…ate task has not been completed yet then let's wait for it synchronously to ensure that we won't change the thread. This is needed because by default Transaction.Current is ThreadStatic (unless TransactionScopeAsyncFlowOption was used when creating TransactionScope) and the issue was that we could miss Raven-Transaction-Information header when creating a request.


ayende:
Test failed for `AppDomainUnload` reason, don't think it is related, re-running

ayende:
Tests passed in re-run

PR #8148, opened 2018-12-04
RavenDB-12351 Task to gather information to start sync that depended on flush lock  is passing to current lock owner

haludi:
http://issues.hibernatingrhinos.com/issue/RavenDB-12351
If while trying to gather information to start sync there is a thread that already taken the flush lock which required for the task the task is passing to the lock owner


src/Raven.Server/Config/Categories/StorageConfiguration.cs
```diff
@@ -64,6 +64,12 @@ public class StorageConfiguration : ConfigurationCategory
         [ConfigurationEntry("Storage.PrefetchResetThresholdInGb", ConfigurationEntryScope.ServerWideOrPerDatabase)]
         public Size PrefetchResetThreshold { get; set; }
 
+        [Description("Size of journal files in gigabytes for each storage, before forcing a sync and removing the journal files.")]
+        [DefaultValue(1)]
+        [SizeUnit(SizeUnit.Gigabytes)]
```
arekpalinski: Why not to use megabytes? Let's think about embedded solutions e.g. running on raspberry pi when users could set it much lower value

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1168,7 +1169,7 @@ protected void ExecuteIndexing()
                                 if (_logsAppliedEvent.Wait(Configuration.MaxTimeToWaitAfterFlushAndSyncWhenExceedingScratchSpaceLimit.AsTimeSpan))
                                 {
                                     // we've just flushed let's start the sync process
-                                    GlobalFlushingBehavior.GlobalFlusher.Value.ForceSyncEnvironment(storageEnvironment);
+                                //    GlobalFlushingBehavior.GlobalFlusher.Value.ForceSyncEnvironment(storageEnvironment);
```
arekpalinski: why is it commented?

src/Voron/GlobalFlushingBehavior.cs
```diff
@@ -242,7 +209,7 @@ private void FlushEnvironments(HashSet<StorageEnvironment> avoidDuplicates)
                     // we haven't reached the point where we have to flush, but we might want to, if we have enough 
                     // resources available, if we have more than half the flushing capacity, we can do it now, otherwise, we'll wait
                     // until it is actually required.
-                    if (_concurrentFlushes.CurrentCount > StorageEnvironment.MaxConcurrentFlushes / 2)
+                    if (_concurrentFlushes.CurrentCount < StorageEnvironment.MaxConcurrentFlushes / 2)
```
arekpalinski: Let's say that:

- CurrentCount: 2
- StorageEnvironment.MaxConcurrentFlushes: 20

2 < 10 - true, so we have enough resources available but we won't do the flush after this change

src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -426,8 +425,8 @@ public sealed class JournalApplicator : IDisposable
             private readonly object _flushingLock = new object();
             private readonly SemaphoreSlim _fsyncLock = new SemaphoreSlim(1);
             private readonly WriteAheadJournal _waj;
-            private MultipleUseFlag _forceDataSync = new MultipleUseFlag();
             private readonly ManualResetEventSlim _waitForJournalStateUpdateUnderTx = new ManualResetEventSlim();
+            private readonly LockTaskResponsible _flashLockTaskResponsible;
```
arekpalinski: flash -> flush

src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -579,33 +572,11 @@ public void ApplyLogsToDataFile(CancellationToken token, TimeSpan timeToWait)
 
                     var unusedJournals = GetUnusedJournalFiles(jrnls, lastProcessedJournal, lastFlushedTransactionId);
 
-                    var needImmediateFsync =
-                        _pendingSync.IsCompleted &&
-                        (_forceDataSync.Lower() ||
-                         Interlocked.Read(ref _totalWrittenButUnsyncedBytes) > 32 * Constants.Size.Megabyte ||
-                         _timeSinceLastImmediateSync.ElapsedMilliseconds > 3 * 60 * 1000);
```
arekpalinski, line 586: Where do we have this logic now? If those aren't relevant could you explain how it's supposed to work?

test/SlowTests/Voron/Bugs/FlushingToDataFile.cs
```diff
@@ -27,72 +27,6 @@ protected override void Configure(StorageEnvironmentOptions options)
...
-        public void SkipUpdateDatabaseStateAfterSyncIfNewerSyncCalled()
-        {
-            UpdateDatabaseStateAfterSyncIfNewerSyncCalled(false);
-        }
-
-        [Fact]
-        public void FailOnRaceConditionAtUpdateDatabaseStateAfterSyncIfNewerSyncCalled()
-        {
-            Assert.Throws<InvalidOperationException>(() => UpdateDatabaseStateAfterSyncIfNewerSyncCalled(true));
-        }
```
arekpalinski, line 41: Do we really need to remove them? Shouldn't we adjust to the current behavior? Maybe create new ones to verify how it currently works. I'm really reluctant to remove any  Voron tests unless there is a really good reason.

Please check the original issue: http://issues.hibernatingrhinos.com/issue/RavenDB-8515

src/Voron/GlobalFlushingBehavior.cs
haludi: This is current flushes available

arekpalinski: Ouch. You're absolutely right. Very misleading name which made that we changed it quite recently CC @grisha-kotler

src/Voron/Impl/Journal/WriteAheadJournal.cs
haludi: The sync operation run on [idle](https://github.com/ravendb/ravendb/pull/8148/files#diff-87366e0cab8e4af884a9411d055bbc61R72)
 and when it gets to the [configured threshold](https://github.com/ravendb/ravendb/pull/8148/files#diff-87366e0cab8e4af884a9411d055bbc61R126)

arekpalinski: But the previous threshold was `32 * Constants.Size.Megabyte` while now we have 1GB by default, right?

We also dropped the concept of forcing the sync after flush if we haven't done it since 3 minutes regardless the threshold and idleness? Please check http://issues.hibernatingrhinos.com/issue/RavenDB-10238

src/Raven.Server/Config/Categories/StorageConfiguration.cs
```diff
@@ -64,6 +64,12 @@ public class StorageConfiguration : ConfigurationCategory
         [ConfigurationEntry("Storage.PrefetchResetThresholdInGb", ConfigurationEntryScope.ServerWideOrPerDatabase)]
         public Size PrefetchResetThreshold { get; set; }
 
+        [Description("Size of journal files in gigabytes for each storage, before forcing a sync and removing the journal files.")]
```
ayende: Fix the comment size
In general, the entry name should indicate the unit, too. 
`Storage.JournalSizeThresholdInMB`

src/Raven.Server/Documents/Indexes/Index.cs
ayende: We don't want to force it to sync now, just flushing is fine, and then letting the OS deal with background flushing.
The comment should probably be removed, though

src/Voron/GlobalFlushingBehavior.cs
```diff
@@ -109,32 +108,32 @@ private void VoronEnvironmentFlushing()
...
                     continue;
 
-                if (avoidDuplicates.Add(storageEnvironment) == false)
-                    continue; // already seen
+                if (force == false && envSyncReq.Value.IsRequired == false &&
+                   env.Journal.Applicator.TotalWrittenButUnsyncedBytes < env.Options.JournalsSizeThreshold)
+                    continue;
 
-                if (storageEnvironment.Disposed)
+                var exchange = Interlocked.Exchange(ref envSyncReq.Value.IsSyncRun, 1);
```
ayende: Should use `CompareExchange` here. 
But note that we should verify that it is not hanging there with `InSync` all the time if there is an error

src/Voron/GlobalFlushingBehavior.cs
ayende: Yes, I think we should rename this to be clearer

src/Voron/Impl/Journal/WriteAheadJournal.cs
ayende: Sort of, the idea is that the trigger for a sync is the size of the journals, not anything else.
Flush is because of scratch, sync because of journals.

test/SlowTests/Voron/Bugs/FlushingToDataFile.cs
ayende: Yes, we want to keep them, in spirit, at least.

ayende (changes requested):

src/Voron/Impl/Journal/WriteAheadJournal.cs
arekpalinski: We don't have a test for http://issues.hibernatingrhinos.com/issue/RavenDB-10238 but from some reason we didn't hit 32MB threshold then so we might not hit it now.

CC @aviviadi

test/SlowTests/Voron/Bugs/FlushingToDataFile.cs
haludi: I have looked at this for a long time and correct me if I'm wrong.
[These tests are check for the possibility of one sync get to `UpdateDatabaseStateAfterSync`
while another one started a new sync](https://github.com/ravendb/ravendb/pull/8148/files#diff-33af42ec0439152b32cc47c6739f3211L910).
but there is `_fsyncLock` that need to be obtained( [before](https://github.com/ravendb/ravendb/pull/8148/files#diff-33af42ec0439152b32cc47c6739f3211L980)/ [after](https://github.com/ravendb/ravendb/pull/8148/files#diff-33af42ec0439152b32cc47c6739f3211R829) ). There is [code](https://github.com/ravendb/ravendb/pull/8148/files#diff-33af42ec0439152b32cc47c6739f3211L870) in the program that dedicated just for the [test](https://github.com/ravendb/ravendb/pull/8148/files#diff-d783b9bf1a95ac3207ff8ef49e16f12fL70) and if I remove the code the test will not compile.

src/Voron/GlobalFlushingBehavior.cs
```diff
@@ -66,29 +68,26 @@ private void VoronEnvironmentFlushing()
                 while (true)
                 {
                     avoidDuplicates.Clear();
-                    var maybeNeedSync = _maybeNeedToSync.Count;
-                    var millisecondsTimeout = 15000 - maybeNeedSync;
-                    if (millisecondsTimeout <= 0 ||
-                        _flushWriterEvent.Wait(millisecondsTimeout) == false)
+                    
+                    if (_flushWriterEvent.Wait(5000) == false)
```
haludi, line 72: In my checks, I ran a patch on 28.66 KB documents that each iteration multiple the document by 100
There were transactions that took for up to 7-8 seconds and lead to `SyncEnvironments(force: true);` even the system was in pressure and didn't get to journals threshold.
Maybe we should raise the time before `SyncEnvironments(force: true);`  ?

test/SlowTests/Voron/Bugs/FlushingToDataFile.cs
arekpalinski: We might not to have _exactly_ this tests but modify it or write something similar that will assure we won't suffer from the issue that made we added this test.

Note that `UpdateDatabaseStateAfterSync` calls `UpdateFileHeaderAfterDataFileSync(` and it's important to ensure we won't call it with outdated values (`SyncOperation._lastSyncedJournal` etc)

How do we prevent from happening it now?

@aviviadi could you take a look here?

src/Voron/GlobalFlushingBehavior.cs
arekpalinski: Probably we should have better detection of "no activity". Maybe we could use `StorageEnvironment.LastWorkTime`, also check if there is no active write transaction right now

ayende: We don't care about active tx, we care that there hasn't been a flush (on any storage) for long enough that we can can doing other stuff

ayende (approved):

src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -946,85 +887,119 @@ private void CallPagerSync()
...
+                }
 
-                        foreach (var kvp in _journalsToDelete)
+                public bool WaitForTaskToBeDone(Action task)
+                {
+                    var isLockTaken = false;
+                    try
+                    {
+                        _waitForTaskToBeDone.Reset();
+                        _task += task;
```
arekpalinski, line 960: We do `+=` but never `-=` and we have just one instance of `LockTaskResponsible` so we only append tasks and always execute all of them.

test/SlowTests/Voron/Bugs/FlushingToDataFile.cs
arekpalinski: @haludi Please check above comment, in particular "How do we prevent from happening it now?". I think it's handled by `LockTaskResponsible` now but please confirm

src/Voron/Impl/Journal/WriteAheadJournal.cs
haludi: In `RunTaskIfNotAlreadyRan` method the `_task` field assigned to `null` [<sup>ref</sup>](RunTaskIfNotAlreadyRan)

arekpalinski: I didn't notice that. Makes sense

arekpalinski:
Merging it. Tests might be added in separate PR.

PR #8150, opened 2018-12-04
RavenDB-12151 - take into account the work that needs to be done for a map-reduce index (during the reduce)

grisha-kotler:


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -206,6 +206,10 @@ protected Index(IndexType type, IndexDefinitionBase definition)
                 GetRegexFactory = GetOrAddRegex
             };
 
+            _txAllocationsRatio = type == IndexType.MapReduce ||
+                           type == IndexType.AutoMapReduce ||
+                           type == IndexType.JavaScriptMapReduce ? 2.5 : 2;
```
arekpalinski: You can use `type.IsMapReduce()`

arekpalinski (approved):

ayende (approved):

PR #8159, opened 2018-12-04
RavenDB-12382: Most cases of MemoryStream now use the Recyclable one.

redknightlois:
Dont push it yet, we should do a proper insert benchmark test to ensure the change does not cause a big regression on raw performance.


ayende (approved):

arekpalinski:
We got a lot of test failures. I've re-run the test suit

redknightlois:
@arekpalinski Yes, not all `MemoryStream` are created equal. There are a few that would require we look deeper into how to tackle them, luckily they are not such a big resource hog, so we can skip them... The whole pipes and `Span<T>` effort would make such work useless anyways.

arekpalinski:
All tests passed after your changes. Waiting for your confirmation it doesn't cause any perf issues so we can merge it.

PR #8160, opened 2018-12-04
[DO-NOT-MERGE] For performance testing purposes

redknightlois:


src/Voron/Impl/Paging/CryptoPager.cs
```diff
@@ -80,6 +81,16 @@ public override void ReleaseAllocationInfo(byte* baseAddress, long size)
             Inner.ReleaseAllocationInfo(baseAddress, size);
         }
 
+        public override void DiscardPages(long pageNumber, int numberOfPages)
+        {
+            // REVIEW: Nothing to do here?.
```
ayende: Need to call inner for this

src/Voron/Impl/Paging/CryptoPager.cs
```diff
@@ -80,6 +81,16 @@ public override void ReleaseAllocationInfo(byte* baseAddress, long size)
...
+        }
+
+        public override void DiscardWholeFile()
+        {
+            // REVIEW: Nothing to do here?.
```
ayende: Need to call inner for this

src/Voron/Impl/Paging/Windows32BitsMemoryMapPager.cs
```diff
@@ -269,6 +269,16 @@ public override I4KbBatchWrites BatchWriter()
             return new Windows32Bit4KbBatchWrites(this);
         }
 
+        public override void DiscardPages(long pageNumber, int numberOfPages)
+        {
+            // REVIEW: Nothing to do here?.
```
ayende: Probably nothing, we unmap it from memory entirely, so I would guess not.
However, for temp files, we might want to explicitly discard before unmap, to avoid writing to disk

src/Voron/Platform/Posix/Posix32BitsMemoryMapPager.cs
```diff
@@ -191,6 +191,16 @@ public override I4KbBatchWrites BatchWriter()
             return new Posix32Bit4KbBatchWrites(this);
         }
 
+        public override void DiscardPages(long pageNumber, int numberOfPages)
+        {
+            // REVIEW: Nothing to do here?.
```
ayende, line 196: Same here, in temp files, discard before unmap

src/Voron/Platform/Win32/WindowsMemoryMapPager.cs
```diff
@@ -131,7 +131,11 @@ public struct SplitValue
 
             NumberOfAllocatedPages = _totalAllocationSize / Constants.Storage.PageSize;
 
-            SetPagerState(CreatePagerState());
+            var pager = CreatePagerState();
+            if (fileAttributes.HasFlag(Win32NativeFileAttributes.Temporary))
+                pager.DiscardDataOnDisk();
```
ayende, line 136: It looks like this is called, and then we even though we _have_ ` if (PlatformDetails.CanDiscardMemory)`, it isn't good enough?

ayende (changes requested):

arekpalinski:
Almost all tests have failed. Exception:
```
Raven.Client.Exceptions.Server.ServerLoadFailureException : Failed to load system storage \r\nAt :memory:\r\n---- System.AggregateException : One or more errors occurred. 
(Unable to find an entry point named 'DiscardVirtualMemory' in DLL 'kernel32.dll'.)\r\n-------- System.EntryPointNotFoundException : Unable to find an entry point named 'DiscardVirtualMemory' in DLL 'kernel32.dll'.
```

grisha-kotler:
What is the operating system that we use?

arekpalinski:
Tests were run in scratch1 machine - windows server 2012 r2 standard

ayende:
I would expect this to actually have this API

src/Sparrow/Platform/PlatformDetails.cs
```diff
@@ -19,6 +19,7 @@ public static class PlatformDetails
         public static readonly bool RunningOnLinux = RuntimeInformation.IsOSPlatform(OSPlatform.Linux);
 
         public static readonly bool CanPrefetch = IsWindows8OrNewer() || RunningOnPosix;
+        public static readonly bool CanDiscardMemory = IsWindows8OrNewer() || RunningOnPosix;
```
grisha-kotler: DiscardVirtualMemory can be used in Windows 8.1 and Server 2012 R2 or newer.

You need version 6.3 or higher.

src/Sparrow/Memory.cs
```diff
@@ -261,5 +263,22 @@ public static T Read<T>(byte* ptr)
...
+            if (PlatformDetails.CanDiscardMemory)
+            {
+                // We explicitely ignore the return codes because even if it fails, from all uses and purposes we dont care. 
+                if (PlatformDetails.RunningOnPosix)
+                {
+                    Syscall.madvise(new IntPtr(baseAddress), new UIntPtr((ulong)size), MAdvFlags.MADV_DONTNEED);
+                }
+                else
+                {
+                    Win32MemoryProtectMethods.DiscardVirtualMemory(baseAddress, new UIntPtr((ulong)size));
```
grisha-kotler, line 279: Can we log this if we failed?
(Debug level?)

grisha-kotler (changes requested):

src/Sparrow/Memory.cs
```diff
@@ -261,5 +263,22 @@ public static T Read<T>(byte* ptr)
         {
             return Unsafe.Read<T>(ptr);
         }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static void Discard(void* baseAddress, long size)
```
grisha-kotler, line 268: We should ignore all exceptions here (log this and "disable" the discard for all future calls)

topalach:
The OS version is `Microsoft Windows [Version 6.3.9600]`

topalach:
I started another test run on scratch5 (Windows Server 2016 Standard `Version 10.0.14393`):
http://jenkins.hibernatingrhinos.com/job/PR_Tests_v4.1/2161/

src/Sparrow/Memory.cs
redknightlois: We could. Just in debug mode though, not that it makes any difference to know though. It can fail and we will still working properly.

src/Sparrow/Memory.cs
redknightlois: We should figure out which OSs do not support it correctly instead. Logging and disabling is not a good idea, because it hides from us important platform assumptions that are broken.

redknightlois:
That explains it... `DiscardVirtualMemory` is supported after `Windows Server 2012 R2 **Update** ` . Now check how to figure out whats the moniker for that one 🔮

ayende:
*Serious* There is a difference between R2 and R2 Update ?!
That is just seriously bad naming.

src/Sparrow/Memory.cs
arekpalinski: Maybe we could try it once on server startup and if we fail then log and create an alert?

Not sure if this call might be expensive but maybe if we fail a few times consecutively then let's skip this call completely.

arekpalinski: WRT my 2nd comment - already answered in https://github.com/ravendb/ravendb/pull/8160#discussion_r239216380

grisha-kotler:
Windows 8.1 (Update 1): 6.3.9600
Windows 8.1: 6.3.9200

Our build: 6.3.9600 and it is still failing

src/Voron/Platform/Win32/WindowsMemoryMapPager.cs
redknightlois: No, the problem is when we discard data which is not temporary :)

arekpalinski:
@redknightlois It seems tests failed due to 
`RavenDB-12382: Most cases of MemoryStream now use the Recyclable one.` commit which I guess should not be part of this PR

redknightlois:
No. It should not.

ayende:
Moved to v4.2 at https://github.com/ravendb/ravendb/pull/8323

PR #8172, opened 2018-12-06
RavenDB-12403: 'raw index entries' can be exported in csv format.

garayx:


src/Raven.Server/Documents/Handlers/StreamCsvBlittableQueryResultWriter.cs
```diff
@@ -0,0 +1,29 @@
...
+    public class StreamCsvBlittableQueryResultWriter : StreamCsvResultWriter<BlittableJsonReaderObject>
+    {
+        public override void AddResult(BlittableJsonReaderObject res)
+        {
+            WriteCsvHeaderIfNeeded(res, false);
+
+            foreach ((var property, var path) in GetProperties())
+            {
+                var o = new BlittablePath(path).Evaluate(res, false);
+                GetCscWriter().WriteField(o?.ToString());
```
arekpalinski: GetCscWriter -> GetCsvWriter ?

src/Raven.Server/Documents/Handlers/StreamCsvResultWriter.cs
```diff
@@ -0,0 +1,150 @@
...
+        private readonly HttpResponse _response;
+        private readonly DocumentsOperationContext _context;
+        private readonly StreamWriter _writer;
+        private CsvWriter _csvWriter;
+        private (string, string)[] _properties;
+        private bool writeHeader = true;
+
+        protected StreamCsvResultWriter(HttpResponse response, Stream stream, DocumentsOperationContext context, string[] properties = null, string csvFileName = "export")
+        {
+            csvFileName = $"{csvFileName}_{SystemTime.UtcNow.ToString("yyyyMMdd_HHmm", CultureInfo.InvariantCulture)}.csv";
```
arekpalinski: Why do we have default value for `csvFileName = "export"` while we are always overwriting it here? So the default it's just a prefix. Let's reflect that in parameter name: `csvFileName` -> `csvFilePrefixName`

src/Raven.Server/Documents/Handlers/StreamingHandler.cs
```diff
@@ -134,29 +134,56 @@ public async Task StreamQueryPost()
...
-                        await Database.QueryRunner.ExecuteStreamQuery(query, context, HttpContext.Response, writer, token).ConfigureAwait(false);
+                        try
+                        {
+                            await Database.QueryRunner.ExecuteStreamIndexEntriesQuery(query, context, HttpContext.Response, writer, token).ConfigureAwait(false);
+                        }
+                        catch (IndexDoesNotExistException)
+                        {
+                            HttpContext.Response.StatusCode = (int)HttpStatusCode.NotFound;
+                            writer.WriteError($"Index {query.Metadata.IndexName}  does not exists");
+                        }
```
arekpalinski, line 188: We never check for the value of `debug` variable. For now we have only `"entries"`. But in the future we might have more options. Let's throw if we get unexpected value here.

src/Raven.Server/Documents/Handlers/StreamingHandler.cs
```diff
@@ -134,29 +134,56 @@ public async Task StreamQueryPost()
...
-                        HttpContext.Response.StatusCode = (int)HttpStatusCode.NotFound;
-                        writer.WriteError("Index " + query.Metadata.IndexName + " does not exists");
+                        try
+                        {
+                            await Database.QueryRunner.ExecuteStreamQuery(query, context, HttpContext.Response, writer, token).ConfigureAwait(false);
+                        }
+                        catch (IndexDoesNotExistException)
+                        {
+                            HttpContext.Response.StatusCode = (int)HttpStatusCode.NotFound;
+                            writer.WriteError($"Index {query.Metadata.IndexName} does not exists");
```
arekpalinski: exists -> exist

src/Raven.Server/Documents/Handlers/StreamingHandler.cs
```diff
@@ -134,29 +134,56 @@ public async Task StreamQueryPost()
...
                     }
                 }
             }
         }
 
-        private IStreamDocumentQueryResultWriter GetQueryResultWriter(string format, HttpResponse response, DocumentsOperationContext context, Stream responseBodyStream,
+        private StreamCsvBlittableQueryResultWriter GetIndexEntriesQueryResultWriter(string format, HttpResponse response, DocumentsOperationContext context, Stream responseBodyStream,
+            string[] propertiesArray, string fileName = null)
+        {
+            if (string.IsNullOrEmpty(format) || format.StartsWith("csv") == false)
```
arekpalinski: `StartsWith("csv")` ? Shouldn't be exact match (case insensitive)?

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -2326,6 +2337,101 @@ public IndexDefinition GetIndexDefinition()
             }
         }
 
+        private async Task IndexEntriesQueryInternal<TQueryResult>(TQueryResult resultToFill, IndexQueryServerSide query,
```
arekpalinski, line 2325: Let's refactor `IndexEntries()` method so it will just call `IndexEntriesQueryInternal()`. See how `Query` is implemented

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -2326,6 +2337,101 @@ public IndexDefinition GetIndexDefinition()
...
+                {
+                    SetState(IndexState.Normal);
+                }
+                catch (Exception e)
+                {
+                    if (_logger.IsOperationsEnabled)
+                        _logger.Operations($"Failed to change state of '{Name}' index from {IndexState.Idle} to {IndexState.Normal}. Proceeding with running the query.",
+                            e);
+                }
+            }
```
arekpalinski: We already have the same piece of code in `QueryInternal` let's extract it to a private method

src/Raven.Server/Documents/Queries/StreamDocumentIndexEntriesQueryResult.cs
```diff
@@ -0,0 +1,114 @@
...
+using Raven.Client.Documents.Operations.Counters;
+using Raven.Server.Documents.Handlers;
+using Raven.Server.Documents.Includes;
+using Raven.Server.Documents.Queries.Explanation;
+using Raven.Server.ServerWide;
+using Sparrow.Json;
+
+namespace Raven.Server.Documents.Queries
+{
+    public class StreamDocumentIndexEntriesQueryResult : QueryResultServerSide<BlittableJsonReaderObject>
```
arekpalinski: It appears this class is the same as `StreamDocumentQueryResult`. Can we exact most of the code to a base class or have a generic `StreamDocumentQueryResult<T>` ?

arekpalinski: Seems like only `void AddResult` needs to be overridden in both implementations

garayx: done

src/Raven.Server/Documents/Handlers/StreamCsvBlittableQueryResultWriter.cs
garayx: done

src/Raven.Server/Documents/Indexes/Index.cs
garayx: done

src/Raven.Server/Documents/Indexes/Index.cs
garayx: done

src/Raven.Server/Documents/Handlers/StreamingHandler.cs
garayx: its sent as {format: "csv"} from query.ts, therefore it cant be anything aside "csv".

src/Raven.Server/Documents/Handlers/StreamingHandler.cs
garayx: done

src/Raven.Server/Documents/Handlers/StreamingHandler.cs
arekpalinski: If you send a request directly from a browser:
- streams/queries?debug=entries&format=CSV - will fail (it's case sensitive currently)
- streams/queries?debug=entries&format=csveeeee - won't fail but it should because we don't support `csveeee` format

garayx: done, but currently going directly to the URL:
- streams/queries?format=csv&debug=entries

throws System.InvalidOperationException

src/Raven.Server/Documents/Handlers/StreamingHandler.cs
garayx: done

src/Raven.Server/Documents/Handlers/StreamingHandler.cs
arekpalinski: Because the actual query is missing, right? I just wanted to give an example, not fully working URL

src/Raven.Server/Documents/Handlers/StreamCsvResultWriter.cs
garayx: done

src/Raven.Server/Documents/Handlers/StreamingHandler.cs
```diff
@@ -138,12 +138,12 @@ public async Task StreamQueryPost()
...
-                // set the exported file name
-                var fileName = query.Metadata.IsCollectionQuery ? query.Metadata.CollectionName + "_collection" : "query_result";
-                fileName = $"{Database.Name}_{fileName}";
+                // set the exported file name prefix
+                var fileNamePrefix = query.Metadata.IsCollectionQuery ? query.Metadata.CollectionName + "_collection" : "query_result";
+                fileNamePrefix = $"{Database.Name}_{fileNamePrefix}";
                 if (string.IsNullOrWhiteSpace(debug) == false)
                 {
-                    using (var writer = GetIndexEntriesQueryResultWriter(format, HttpContext.Response, context, ResponseBodyStream(), propertiesArray, fileName))
+                    using (var writer = GetIndexEntriesQueryResultWriter(format, debug, HttpContext.Response, context, ResponseBodyStream(), propertiesArray, fileNamePrefix))
```
arekpalinski: The value of `debug` should to be checked here, not inside `GetIndexEntriesQueryResultWriter`. If we add another debug option we'll call a dedicated method to handle it here, not from `GetIndexEntriesQueryResultWriter`

src/Raven.Server/Documents/Queries/StreamQueryResult.cs
```diff
@@ -0,0 +1,118 @@
...
+        {
+            return _writer;
+        }
+
+        public OperationCancelToken GetToken()
+        {
+            return _token;
+        }
+
+        public bool GetAnyWrites()
```
arekpalinski: GetAnyWrites -> HasAnyWrites

src/Raven.Server/Documents/Queries/StreamQueryResult.cs
```diff
@@ -0,0 +1,118 @@
...
+        {
+            return _anyWrites;
+        }
```
arekpalinski, line 42: BTW the above can be just properties e.g. `protected IStreamQueryResultWriter<T> Writer => _writer`

arekpalinski:
There is a conflict on `src/Raven.Studio/typescript/viewmodels/database/query/query.ts`. Please rebase from 4.1 and resolve it

src/Raven.Studio/typescript/viewmodels/database/query/query.ts
```diff
@@ -1041,9 +1041,17 @@ class query extends viewModelBase {
     exportCsv() {
         eventsCollector.default.reportEvent("query", "export-csv");
 
-        const args = {
-            format: "csv",
-        };
+        let args: any;
```
ml054: please avoid using any as a type.

src/Raven.Server/Documents/Queries/StreamQueryResult.cs
garayx: done

src/Raven.Server/Documents/Handlers/StreamingHandler.cs
garayx: done

src/Raven.Studio/typescript/viewmodels/database/query/query.ts
garayx: done

PR #8173, opened 2018-12-06
RavenDB-12401 Add features to query page when query setting are used

ml054:


maximburyak:
@ml054 I think that this is still very small notification to the fact that we "trim" the query past the select statement, consider the fact that a user could have a specific projeciton, and we don't "add" the "stored fields", instead we substitute the projection with the request to receive all stored fields.

ml054:
For me that fact that we trim query is implementation details. How would you indicate that we are showing all fields?

PR #8174, opened 2018-12-06
RavenDB-12444 Add top threads endpoint

karmeli87:


src/Raven.Server/Documents/Handlers/Debugging/ThreadsHandler.cs
```diff
@@ -13,6 +14,32 @@ namespace Raven.Server.Documents.Handlers.Debugging
 {
     public class ThreadsHandler : RequestHandler
     {
+        [RavenAction("/admin/debug/threads/top", "GET", AuthorizationStatus.Operator, IsDebugInformationEndpoint = true)]
```
ayende: Isn't this just runaway?

ayende: There we are using `thread.TotalProcessorTime.TotalMilliseconds`

grisha-kotler: It doesn't provide the actual CPU usage. @karmeli87 maybe we can combine the 2 endpoints?

karmeli87: I think those are 2 different measures. `runaway` measuring from the start-up, while `top` is only for the last second.

But we can merge them if you like, the question is how we will order them?

arekpalinski: Reminder - @grisha-kotler ?

ayende:
Reminder @grisha-kotler

src/Raven.Server/Documents/Handlers/Debugging/ThreadsHandler.cs
grisha-kotler: Merged both endpoints

ppekrol (approved):

ppekrol:
@grisha-kotler I'm merging this. Please create a separate PR with adjustments.

PR #8193, opened 2018-12-09
RavenDB-12441: Remove LINQ from statements from ToBlittableSupportedType

garayx:


src/Raven.Server/Utils/TypeConverter.cs
```diff
@@ -226,7 +226,7 @@ private static object ToBlittableSupportedType(object root, object value, bool f
 
                 return @object;
             }
-
+            // looks like best option
             if (value is IEnumerable<char> charEnumerable)
                 return new string(charEnumerable.ToArray());
```
ayende, line 247: That will almost always be either `string` of `char[]`, we can check for them explicitly

src/Raven.Server/Utils/TypeConverter.cs
```diff
@@ -237,14 +237,10 @@ private static object ToBlittableSupportedType(object root, object value, bool f
...
-
-                    if (value is IEnumerable<object> objectEnumerable)
-                        items = Enumerable.Select(objectEnumerable, x => ToBlittableSupportedType(root, x, flattenArrays, recursiveLevel + 1, engine, context));
-                    else
-                        items = Enumerable.Select(enumerable.Cast<object>(), x => ToBlittableSupportedType(root, x, flattenArrays, recursiveLevel + 1, engine, context));
-
-                    return new DynamicJsonArray(flattenArrays ? Flatten(items) : items);
+                    return value is IEnumerable<object> objectEnumerable
+                        ? EnumerableToJsonArray(flattenArrays ? Flatten(objectEnumerable) : objectEnumerable, root, flattenArrays, recursiveLevel, engine, context)
+                        : EnumerableToJsonArray(flattenArrays ? Flatten(enumerable.Cast<object>()) : enumerable.Cast<object>(), root, flattenArrays, recursiveLevel,
```
ayende: Can we avoid `.Cast<object>()` ?

src/Raven.Server/Utils/TypeConverter.cs
```diff
@@ -226,7 +226,7 @@ private static object ToBlittableSupportedType(object root, object value, bool f
 
                 return @object;
             }
-
+            // looks like best option
```
arekpalinski: best option considering performance / allocations right? please update the comment to reflect that

src/Raven.Server/Utils/TypeConverter.cs
garayx: done, i am checking for
if (value is char[] chars) AND if (value is string)
instead

src/Raven.Server/Utils/TypeConverter.cs
garayx: done, sorry thats a note to myself that I forgot to delete.

src/Raven.Server/Utils/TypeConverter.cs
garayx: should I cast each member of enumerable to object in a loop?

src/Raven.Server/Utils/TypeConverter.cs
```diff
@@ -292,6 +319,9 @@ private static void NestingLevelTooDeep(object value)
 
         private static IEnumerable<object> Flatten(IEnumerable items)
         {
+            if (items == null)
+                EnumerableNullException();
```
arekpalinski: I think that passing `DynamicNullObject` would make it will throw but it shouldn't. It's a bit dangerous, I would drop that check to avoid possible regression bug in indexing.

arekpalinski: Please check tests: http://jenkins.hibernatingrhinos.com/job/PR_Tests_v4.1/2195/testReport/

I'm pretty sure they failed due to this check

garayx: done

arekpalinski (approved):

src/Raven.Client/Json/TypeConverter.cs
```diff
@@ -63,13 +62,17 @@ public static object ToBlittableSupportedType(object value, DocumentConventions
                 return @object;
             }
 
-            if (value is IEnumerable enumerable)
+            if (value is IEnumerable)
```
lahma: there might be some perf benefit checking for IList **if** it's common input, now enumerator is always allocated vs direct indexing using old-school for-loop

garayx: I am not sure it will work, 
the typeof value received here is object, so in both cases (IList or IEnuramble) we will have to cast it. Because we want to just enumerate on collection (we dont need the indexes that are present in lists) I dont see a point to make an if-case for IList type.

lahma: Fair enough, I merely wanted to point out there can also be a performance difference which way you do it. [Here's some numbers](https://blog.lahma.net/2018/11/optimizing-jint-part-4-interfaces-are.html) for iterating the same List with different casts. So not entirely related whether you need the indexes used when iterating.

But as said, highly usage scenario dependent (if most are not lists, no point of having extra complexity). With IList there would still be interface dispatch slowing things down, but "a lot" faster than iterating IEnumerable.

arekpalinski: What do we get here depends on an index definition - so can't assume it will be IList mostly

PR #8198, opened 2018-12-10
RavenDB-12471 - DiskFullException doesn't result in client failover

grisha-kotler:
http://issues.hibernatingrhinos.com/issue/RavenDB-12471


ayende (approved):
Should also look at the catastrophic errors from voron, though, no?
@aviviadi

arekpalinski:
We already check if `exception is VoronUnrecoverableErrorException` there. Do you mean to also add `VoronErrorException` (thrown only in `Table.cs` at this point) ?

ayende:
Basically, any error that can kill just a single node

PR #8201, opened 2018-12-10
RavenDB-12474 update supported browser detection

ml054:


ayende (commented):
We need this on 4.0 as well, no?

ml054:
Yes, thanks. Will create PR in few seconds.

ayende:
Also, what about chormenium?

ml054:
What's chormenium ?

arekpalinski:
https://www.chromium.org/Home ?

ml054:
check returns true for chromium, but since chrome is based on chromium, both browsers are supported by us.

PR #8212, opened 2018-12-11
fix for RavenDB-12263

myarichuk:
Make sure to include ALL path permutations in recursive graph queries when using "all" strategy.
(includes some very minor refactorings and comments for better code clarity)

### DIFF
--- a/src/Raven.Server/Documents/Queries/Graph/RecursionQueryStep.cs
+++ b/src/Raven.Server/Documents/Queries/Graph/RecursionQueryStep.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Primitives;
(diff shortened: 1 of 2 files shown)


src/Raven.Server/Documents/Queries/Graph/RecursionQueryStep.cs
```diff
@@ -304,29 +306,42 @@ private void ProcessSingleResultRecursive(Match currentMatch, List<Match> matche
             _path.Push(new RecursionState { Src = startingPoint.Data, Match = currentMatch });
 
             int aliasBaseIndex = 0;
-
             Document cur = startingPoint;
+
+            //this function is needed for "documentation" purposes
```
ayende, line 311: What do you mean in this comment?

src/Raven.Server/Documents/Queries/Graph/RecursionQueryStep.cs
```diff
@@ -304,29 +306,42 @@ private void ProcessSingleResultRecursive(Match currentMatch, List<Match> matche
...
+            bool AddMatchToResultsAndCheckIfNeedToStop(Document current)
+            {
+                if (AddMatchToResults(current))
+                    return true;
+
+                _path.Pop();
```
ayende, line 317: Don't we need to clear the `_visited` here as well?

myarichuk: Don't think so, because we are backtracking + clearing ```_visited``` only in the block that comes after ```if (top.Matches == null || top.Matches.Count == 0)```
(https://github.com/myarichuk/ravendb/blob/RavenDB-12263/src/Raven.Server/Documents/Queries/Graph/RecursionQueryStep.cs#L368)

PR #8213, opened 2018-12-11
fix for RavenDB-12459

myarichuk:
…ression with multiple values (for example: Order -> OrderLines.Product) (RavenDB-12459)

### DIFF
--- a/src/Raven.Server/Documents/Queries/Graph/SingleEdgeMatcher.cs
+++ b/src/Raven.Server/Documents/Queries/Graph/SingleEdgeMatcher.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using Microsoft.Extensions.Primitives;
+using Raven.Client.Exceptions;
 using Raven.Server.Documents.Queries.AST;
 using Raven.Server.Json;
 using Raven.Server.Utils;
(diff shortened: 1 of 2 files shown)


arekpalinski (approved):

src/Raven.Server/Documents/Queries/Graph/SingleEdgeMatcher.cs
```diff
@@ -46,12 +47,17 @@ public void SingleMatch(Match left, string alias)
...
                         foreach (var item in array)
                         {
                             if (item is BlittableJsonReaderObject json &&
                                 Edge.Where?.IsMatchedBy(json, QueryParameters) != false)
                             {
-                                AddEdgeAfterFiltering(left, json, Edge.Project.FieldValue);
+                                AddEdgeAfterFiltering(left, json, Edge.Project?.FieldValue);
                             }
                         }
                         break;
```
ayende, line 63: Same issue below, line 67

ayende (changes requested):

src/Raven.Server/Documents/Queries/Graph/SingleEdgeMatcher.cs
```diff
@@ -43,6 +44,11 @@ public void SingleMatch(Match left, string alias)
                 if (BlittableJsonTraverser.Default.TryRead(leftDoc, Edge.Path.FieldValue, out var value, out _) == false)
                     return;
 
+                if (Edge.Project == null)
```
ayende, line 47: I'm trying to think if there is any case where we want a `where` without a `select`
`[Friend where startsWith(Friend, $prefix)]` looks like the only thing, but that seems... odd

PR #8215, opened 2018-12-12
RavenDB-12193: UnobservedTaskExceptions should be handled now.

garayx:
RavenDB-12481

### DIFF
--- a/test/SlowTests/Issues/RavenDB_12193.cs
+++ b/test/SlowTests/Issues/RavenDB_12193.cs
@@ -1,0 +1,41 @@
+using System;
+using Xunit;
+using System.Threading;
+using System.Threading.Tasks;
+using FastTests;
+using Raven.Client.Documents.Changes;
+using Raven.Client.Extensions;
+using Raven.Tests.Core.Utils.Entities;
+
+namespace SlowTests.Issues
+{
+    public class RavenDB_12193 : RavenTestBase
+    {
+        // RavenDB-12481
+        [Fact]
+        public async Task Should_Throw_On_UnobservedTaskException()
+        {
+            var count = 0;
+
+            EventHandler<UnobservedTaskExceptionEventArgs> task = (sender, args) =>
+            {
+                Interlocked.Increment(ref count);
+            };
+
+            try
+            {
+                TaskScheduler.UnobservedTaskException += task;
+                using (GetDocumentStore().Changes().ForAllDocuments().Subscribe(change => {{}})){}
+
+                GC.Collect(GC.MaxGeneration, GCCollectionMode.Forced);
+
+                await Task.Delay(1000);
+                Assert.Equal(0, count);
+            }
+            finally
+            {
+                TaskScheduler.UnobservedTaskException -= task;
+            }
+        }
+    }
+}
--- a/test/SlowTests/Issues/RavenDB_12193.cs
+++ b/test/SlowTests/Issues/RavenDB_12193.cs
@@ -28,8 +28,8 @@ namespace SlowTests.Issues
                 using (GetDocumentStore().Changes().ForAllDocuments().Subscribe(change => {{}})){}
 
                 GC.Collect(GC.MaxGeneration, GCCollectionMode.Forced);
+                GC.WaitForPendingFinalizers();
 
-                await Task.Delay(1000);
                 Assert.Equal(0, count);
             }
             finally
--- a/src/Raven.Client/Documents/Changes/DatabaseConnectionState.cs
+++ b/src/Raven.Client/Documents/Changes/DatabaseConnectionState.cs
@@ -2,6 +2,7 @@ using System;
 using System.Threading;
 using System.Threading.Tasks;
 using Raven.Client.Documents.Operations;
+using Raven.Client.Extensions;
 
 namespace Raven.Client.Documents.Changes
 {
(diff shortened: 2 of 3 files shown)


test/SlowTests/Issues/RavenDB_12193.cs
```diff
@@ -0,0 +1,41 @@
...
+                Interlocked.Increment(ref count);
+            };
+
+            try
+            {
+                TaskScheduler.UnobservedTaskException += task;
+                using (GetDocumentStore().Changes().ForAllDocuments().Subscribe(change => {{}})){}
+
+                GC.Collect(GC.MaxGeneration, GCCollectionMode.Forced);
+
```
ayende, line 32: `GC.WaitForPendingFinalizers` here
And remove the delay

garayx: done

test/SlowTests/Issues/RavenDB_12193.cs
```diff
@@ -0,0 +1,41 @@
...
+            var count = 0;
+
+            EventHandler<UnobservedTaskExceptionEventArgs> task = (sender, args) =>
+            {
+                Interlocked.Increment(ref count);
+            };
+
+            try
+            {
+                TaskScheduler.UnobservedTaskException += task;
```
arekpalinski, line 27: This line should be just before `try`. If we failed here from any reason then running the finally clause could throw another exception (not in this case probably, but in general)

test/SlowTests/Issues/RavenDB_12193.cs
```diff
@@ -0,0 +1,41 @@
...
+using FastTests;
+using Raven.Client.Documents.Changes;
+using Raven.Client.Extensions;
+using Raven.Tests.Core.Utils.Entities;
+
+namespace SlowTests.Issues
+{
+    public class RavenDB_12193 : RavenTestBase
+    {
+        // RavenDB-12481
```
arekpalinski, line 14: RavenDB-12481 ?

test/SlowTests/Issues/RavenDB_12193.cs
```diff
@@ -0,0 +1,41 @@
...
+        [Fact]
+        public async Task Should_Throw_On_UnobservedTaskException()
+        {
+            var count = 0;
+
+            EventHandler<UnobservedTaskExceptionEventArgs> task = (sender, args) =>
+            {
+                Interlocked.Increment(ref count);
```
arekpalinski, line 22: I wonder if we could somehow check here that `sender` is the guy that was causing the issue. I'm a bit worried that we can have failures of this tests due to other unknown unobserved tasks (which might be actually fine - something in server size)

test/SlowTests/Issues/RavenDB_12193.cs
garayx: just added it to mention that this test also checks the UnobservedTask from RavenDB-12481,

test/SlowTests/Issues/RavenDB_12193.cs
garayx: for now it checks all UnobservedTaskException, 
 I can check if args or sender is not null if it is check if its equal to exceptions that I want to track.

test/SlowTests/Issues/RavenDB_12193.cs
arekpalinski: ok

test/SlowTests/Issues/RavenDB_12193.cs
arekpalinski: I'm pretty sure `sender` and `args` will never be null. I don't really know how to assure that the test can fail only due to the task spawn here. Maybe that won't be a problem.

ayende: Do we get the task itself? We might be able to get better logs on what it is.
If it fails, we can handle that as well

arekpalinski: Okay, it seems we have first failure of this test:

http://jenkins.hibernatingrhinos.com/job/PR_Tests_v4.1/2223/testReport/junit/SlowTests.Issues/RavenDB_12193/Should_Throw_On_UnobservedTaskException/

It's from testing https://github.com/ravendb/ravendb/pull/8159 PR

arekpalinski: BTW I've just found we have: https://github.com/ravendb/ravendb/blob/436d63cd0b9b42473d7dc0945b8f4ebd2be6b2a3/test/Tests.Infrastructure/TestBase.cs#L70-L85

Regarding the above failure, personally I'm fine with skipping this test (but leave it with comment so one can run it manually if needed) if we cannot make it behave consistently. This wasn't a but that made some functionality wasn't working.

redknightlois: @arekpalinski does this PR include #8159 or we still have that error there? This thing of having to run the slow tests to catch the errors get me sidetracked every day. Have like 3 PRs open that I cannot end up closing.

arekpalinski: @redknightlois No this PR doesn't include your https://github.com/ravendb/ravendb/pull/8159. Although your PR failed because of the test added here. We saw a few failures when testing PR due to this test so I'll skip it - https://github.com/ravendb/ravendb/pull/8235

PR #8216, opened 2018-12-12
RavenDB-12469 fix

myarichuk:
Don't fail when there are duplicate implicit aliases in a graph query. In such case, remove them from results, because when we have duplicate implicit aliases, we don't know *which* are which when we retrieve them in a query result

### DIFF
--- a/src/Raven.Server/Documents/Queries/Parser/QueryParser.cs
+++ b/src/Raven.Server/Documents/Queries/Parser/QueryParser.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Text;
 using Esprima;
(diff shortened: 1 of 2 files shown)


src/Raven.Server/Documents/Queries/Parser/QueryParser.cs
```diff
@@ -324,9 +324,11 @@ private bool GraphAlias(GraphQuery gq,bool isEdge, StringSegment implicitPrefix,
                         return true;
 
                     if (_synteticWithQueries?.TryGetValue(collection.FieldValue, out prev) == true)
-                    {
+                    {               
                         if(prev.ImplicitAlias)
```
ayende: We can just rename the 2nd alias to be an anonymous missing one, no, using `_`?

ayende: And update the previous one as well

ayende (approved):

PR #8217, opened 2018-12-12
RavenDB-12096 support for 'exact' in edge where clause

talweiss1982:


src/Raven.Server/Documents/Queries/Graph/SingleEdgeMatcher.cs
```diff
@@ -69,6 +65,40 @@ public void SingleMatch(Match left, string alias)
...
+        {
+            if (string.Equals(methodExpression.Name.Value, "exact", StringComparison.OrdinalIgnoreCase) == false)
+                return false;
+            var arg = methodExpression.Arguments[0] as BinaryExpression;
+            var left = arg.Left as FieldExpression;
+            var right = arg.Right as ValueExpression;
+            if (left == null || right == null)
+                return false;
+            var value = BlittableJsonTraverser.Default.Read(blittableJsonReaderObject, left.FieldValue);
+            if (value is LazyStringValue lsv && lsv.CompareTo(right.Token.Value) == 0)
```
ayende: Also, `LazyCompressedStringValue`

src/Raven.Server/Documents/Queries/Graph/SingleEdgeMatcher.cs
```diff
@@ -69,6 +65,40 @@ public void SingleMatch(Match left, string alias)
...
+            if (Edge.Where.IsMatchedBy(blittableJsonReaderObject, QueryParameters) != false)
+            {
+                AddEdgeAfterFiltering(match, blittableJsonReaderObject, Edge.Project.FieldValue);
+            }
+        }
+
+        private bool EdgeMatchExact(BlittableJsonReaderObject blittableJsonReaderObject, MethodExpression methodExpression)
+        {
+            if (string.Equals(methodExpression.Name.Value, "exact", StringComparison.OrdinalIgnoreCase) == false)
+                return false;
```
ayende: I'm pretty sure that we don't want this. What if we have another method call here? `search`, for example, which isn't supported.

src/Raven.Server/Documents/Queries/Graph/SingleEdgeMatcher.cs
```diff
@@ -69,6 +65,40 @@ public void SingleMatch(Match left, string alias)
...
+            var arg = methodExpression.Arguments[0] as BinaryExpression;
+            var left = arg.Left as FieldExpression;
+            var right = arg.Right as ValueExpression;
+            if (left == null || right == null)
```
ayende: `exact(Field, $argThatIsNull)` is valid.

ayende (changes requested):

src/Raven.Server/Documents/Queries/Graph/SingleEdgeMatcher.cs
talweiss1982: You want me to throw? I was asked not to throw from queries before.

ayende: Throwing when the query itself is invalid - okay.
In this case, you are using a method that isn't supported.

Throwing when the query is valid and some of the docs do not match the expected shape - not okay.

It is about consistency.
If we have a query that would fail for everything, great. Users won't issue that.
IF we have a query that can fail if the data is bad, users might run into that in production, which is _bad_

src/Raven.Server/Documents/Queries/Graph/SingleEdgeMatcher.cs
```diff
@@ -86,16 +86,35 @@ private void EdgeMatch(Match match, BlittableJsonReaderObject blittableJsonReade
...
+                    break;
+                case ValueTokenType.Null:
+                    rightStringValue = null;
+                    break;
+                default:
+                    throw new NotSupportedException($"Where clause with method expression {methodExpression} has parameter {right.Token.Value} which is not of string type.");
+            }
             var value = BlittableJsonTraverser.Default.Read(blittableJsonReaderObject, left.FieldValue);
-            if (value is LazyStringValue lsv && lsv.CompareTo(right.Token.Value) == 0)
+            if (value is LazyStringValue lsv && lsv.CompareTo(rightStringValue) == 0)
```
ayende: Am I'm missing something? You are still not checking that `null` is valid.
Consider the case where the field itself is null, are you are getting `null` as a arg.
You aren't handling this, but you should.
Also, what about things like `Name != $name` ?
You have handled just the equality case, which isn't right.

Also, `Name = $foo AND Age > 1` is valid.
Also, `exact(Age > 2)` is also valid (in this case, the `exact` is doing nothing, obviously

talweiss1982: this is not null value this is a null because of the casting is null

talweiss1982: also exact(Age > 2 ) is strange i wouldn't expect it to work

ayende: It works for regular queries, should work as well
`value is LazyStringValue lsv` returns false for null values.

So a null field value to a null arg will return false but should return true

talweiss1982: I have added a test showing null args are been matched correctly

arekpalinski (approved):

ayende (approved):

ayende:
Please fix the conflict

PR #8219, opened 2018-12-12
RavenDB-12492: Fixed Weird multiplicative factor on MapReduce

redknightlois:


redknightlois:
@arekpalinski need some review here... I think it is right, but I am not that acquainted with the inner workings of this part of the code.

src/Raven.Server/Documents/Indexes/MapReduce/MapReduceResultsStore.cs
```diff
@@ -52,34 +55,38 @@ public MapReduceResultsStore(ulong reduceKeyHash, MapResultsStorageType type, Tr
...
-            ModifiedPages = new HashSet<long>();
-            FreedPages = new HashSet<long>();
+            if (!_alreadyInitializedTrees.Contains(treeName))
+            {
+                Tree.PageModified += (page, flags) =>
+                {
+                    if ((flags & PageFlags.Overflow) == PageFlags.Overflow)
+                        return;
 
-            _mapReduceContext.PageModifiedInReduceTree += page => FreedPages.Remove(page);
```
arekpalinski, line 64: Why did you remove `PageModifiedInReduceTree`? 
It made that  `RavenDB_11643.PageModificationInAnyTreeMustRemoveItFromListOfFreedPagesInAllStores` failed (http://jenkins.hibernatingrhinos.com/job/PR_Tests_v4.1/2220/)

Please check https://github.com/ravendb/ravendb/pull/7690/ it was added there to handle edge case

src/Raven.Server/Documents/Indexes/MapReduce/MapReduceResultsStore.cs
```diff
@@ -52,34 +55,38 @@ public MapReduceResultsStore(ulong reduceKeyHash, MapResultsStorageType type, Tr
...
+
         private void InitializeTree(bool create)
         {
             var treeName = ReduceTreePrefix + _reduceKeyHash;
             var options = _tx.LowLevelTransaction.Environment.Options.RunningOn32Bits ? TreeFlags.None : TreeFlags.LeafsCompressed;
             Tree = create ? _tx.CreateTree(treeName, flags: options) : _tx.ReadTree(treeName);
 
-            ModifiedPages = new HashSet<long>();
-            FreedPages = new HashSet<long>();
+            if (!_alreadyInitializedTrees.Contains(treeName))
```
arekpalinski, line 66: How this can be called multiple times for the same tree instance? `InitializeTree` is supposed to be called just once

ayende (approved):

ayende:
This looks good, but note the failing test

src/Raven.Server/Documents/Indexes/MapReduce/MapReduceResultsStore.cs
grisha-kotler: Maybe by IterateTrees?

src/Raven.Server/Documents/Indexes/MapReduce/MapReduceResultsStore.cs
```diff
@@ -52,34 +55,38 @@ public MapReduceResultsStore(ulong reduceKeyHash, MapResultsStorageType type, Tr
             }
         }
 
+        private readonly HashSet<string> _alreadyInitializedTrees = new HashSet<string>();
```
grisha-kotler, line 58: StringComparer.OrdinalIgnoreCase?

src/Raven.Server/Documents/Indexes/MapReduce/MapReduceResultsStore.cs
arekpalinski: It's used in debug endpoint only for Map-reduce visualizer purposes, so not very likely I think. Those are different instances of trees anyway so that shoudn't matter

redknightlois: It gets called once per every tree. So in essence it will create a linear list of trees per batch, causing a huge slowdown and lots of wasted effort. Just to give you an idea a 24000 batch will generate 500M calls to `.Remove()`

src/Raven.Server/Documents/Indexes/MapReduce/MapReduceResultsStore.cs
redknightlois: Back to the drawing board then. I need to understand what is the rationale of that edge case. Can you elaborate?

arekpalinski: Please check the commit message, and comment in test of https://github.com/ravendb/ravendb/pull/7690 PR. It would be probably the easiest to debug this test to see what is actually going on.

The main problem was that a page (_p1_) marked as _free_ when processing one reduce tree (_t1_), can be allocated by another reduce tree (_t2_). In the reduce phase, we go over the trees and remove the aggregation results for a pages that were marked as free:

https://github.com/ravendb/ravendb/blob/436d63cd0b9b42473d7dc0945b8f4ebd2be6b2a3/src/Raven.Server/Documents/Indexes/MapReduce/ReduceMapResultsBase.cs#L332-L336

In result, processing _t2_ tree added an entry to PageNumberToReduceResult table for _p1_ page but next processing of _t1_ removed it from that table.

arekpalinski: So yes, my fix was: if a page is modified (in any tree), let's remove it from _all_ `FreedPages` collections (for every `MapReduceResultsStore`)

src/Raven.Server/Documents/Indexes/MapReduce/MapReduceResultsStore.cs
arekpalinski: Okay, I understand. You're right it might affect the performance - I haven't thought about it. But still don't understand  how `_alreadyInitializedTrees.Contains(treeName)` can return `true` as `InitializeTree ` is called once

arekpalinski: Edited the above comment: can return false -> can return true

redknightlois: Because now there is a single HashMap per each per `MapReduceStore`. Having said that will see if that condition ever triggers, I started trying to figure out where was the multiplicative factor be generated, and could be useless.

arekpalinski: ok

redknightlois: Is not reusing pages at all a viable strategy?

src/Raven.Server/Documents/Indexes/MapReduce/MapReduceResultsStore.cs
redknightlois: Is not reusing pages at all a viable strategy? What about having a shared FreedPages among `MapReduceResultsStore`?

arekpalinski: Shared `FreedPages` makes sense to me. Since this code:

https://github.com/ravendb/ravendb/blob/436d63cd0b9b42473d7dc0945b8f4ebd2be6b2a3/src/Raven.Server/Documents/Indexes/MapReduce/ReduceMapResultsBase.cs#L332-L336

Doesn't need to be run in "contex" of any particular tree, we can run it before starting processing any tree (`HandleTreeReduction`) just once. That should be fine I think (tests will validate that for sure).

arekpalinski: One potential problem, `HandleTreeReduction` might be called recursively

https://github.com/ravendb/ravendb/blob/436d63cd0b9b42473d7dc0945b8f4ebd2be6b2a3/src/Raven.Server/Documents/Indexes/MapReduce/ReduceMapResultsBase.cs#L406-L427

So maybe we'll be better of by calling `table.DeleteByKey(pageNumberSlice); ` for shared `FreedPages` _after_ processing all trees

redknightlois:
@arekpalinski I implemented the shared free pages. It currently passes all fast tests and the indexing ones, but I am not entirely sure I am not missing something, because I haven't had to modify anything. Either the code is already handling the most complex case, so now it works as is OR I am missing something.

arekpalinski:
I have created https://github.com/ravendb/ravendb/pull/8230 and applied some changes on top of yours. The most important one is moving `table.DeleteByKey(pageNumberSlice);` outside iterating over the reduce trees. Please review

ayende:
I mreged https://github.com/ravendb/ravendb/pull/8230

Do we have something here that we still need?

PR #8221, opened 2018-12-13
RavenDB-12473 Let's ignore write errors when updating index state to …

arekpalinski:
…Error in methods which handle exceptions. This way we won't end up with "Unexpected error in '{Name}' index. This should never happen."


ayende (approved):

arekpalinski:
One test has failed although I couldn't make it fail when running in a loop. Re-running tests on jenkins

PR #8230, opened 2018-12-14
RavenDB-12492

arekpalinski:


src/Raven.Server/Documents/Indexes/MapReduce/ReduceMapResultsBase.cs
```diff
@@ -123,6 +123,19 @@ static ReduceMapResultsBase()
                             throw new ArgumentOutOfRangeException(modifiedStore.Type.ToString());
                     }
                 }
+
+                if (_mapReduceContext.FreedPages.Count > 0)
+                {
+                    long tmp = 0;
```
arekpalinski, line 129: TODO arek - we should measure the execution of the below deletions and put it on the indexing graph

ayende (approved):

PR #8236, opened 2018-12-16
RavenDB-12508 More detail about journals added to `StorageHandler`

haludi:
http://issues.hibernatingrhinos.com/issue/RavenDB-12508


src/Voron/Debugging/StorageReportGenerator.cs
```diff
@@ -103,7 +106,14 @@ public DetailedStorageReport Generate(DetailedReportInput input)
                 tables.Add(tableReport);
             }
 
-            var journals = GenerateJournalsReport(input.Journals);
+            var journals = new JournalsReport
```
haludi, line 109: Maybe the collecting data should be under flush lock?

ayende:
@haludi The code failed to compile, please fix this

src/Voron/Debugging/StorageReportGenerator.cs
ayende: Doesn't actually matter, and might actually impact operations, so let's not do that

arekpalinski:
@ml054 Probably we need to adjust Storage Report in the studio

ppekrol (approved):

PR #8237, opened 2018-12-16
RavenDB-11004 Revert revisions

karmeli87:


src/Raven.Server/Documents/Handlers/RevisionsHandler.cs
```diff
@@ -81,6 +82,31 @@ public Task GetRevisionsFor()
...
+                    Database,
+                    $"Revert database '{Database.Name}' to {time}.",
+                    Operations.Operations.OperationType.DatabaseRevert,
+                    onProgress => Task.Run(() => Database.DocumentsStorage.RevisionsStorage.RevertRevisions(time.Value, window, onProgress), token.Token), 
+                    operationId, 
+                    token: token);
+            }
+            catch (Exception)
+            {
+                HttpContext.Abort();
```
ayende: What about sending the error to the user? Here you just disconnect the connection, which is probably not a good idea.

src/Raven.Server/Documents/Handlers/RevisionsHandler.cs
```diff
@@ -81,6 +82,31 @@ public Task GetRevisionsFor()
...
+        {
+            var time = GetDateTimeQueryString("point-in-time");
+            var window = GetTimeSpanQueryString("window", required: false) ?? TimeSpan.FromHours(48);
+
+            var operationId = GetLongQueryString("operationId", required: false) ?? Database.Operations.GetNextOperationId();
+            var token = CreateOperationToken();
+
+            try
+            {
+                await Database.Operations.AddOperation(
```
ayende: The process may take a *long* time, we want to return the operation id and let the user poll the op state.

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -845,18 +847,260 @@ public Document GetRevisionBefore(DocumentsOperationContext context, string id,
...
+                        result.LastModified < lastModified)
+                    {
+                        result = TableValueToRevision(context, ref tvr.Result.Reader);
+                    }
+                }
+                return result;
+            }
+        }
+
+        public Document GetRevisionBefore(DocumentsOperationContext context, string id, DateTime max, out bool foundAfter, out bool wasConflicted)
```
ayende: You are missing the window here

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -845,18 +847,260 @@ public Document GetRevisionBefore(DocumentsOperationContext context, string id,
...
+                    result.Progress.Warnings.Clear();
+                }
+            }
+
+            return result;
+        }
+
+        private void WriteRevertedRevisions(List<Document> list, DocumentsOperationContext context)
+        {
+            if (list.Count > 0)
```
ayende: Reverse this

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -845,18 +847,260 @@ public Document GetRevisionBefore(DocumentsOperationContext context, string id,
...
+            {
+                using (var tx = context.OpenWriteTransaction())
```
ayende: We don't open write transactions directly. 
_everything_ must go through the transaction merger.
Otherwise, features such as replay won't work.

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -845,18 +847,260 @@ public Document GetRevisionBefore(DocumentsOperationContext context, string id,
...
+                {
+                    foreach (var document in list)
```
ayende: This can be a _very_ big list.
Need to portion it off to several independent operations so as not to block the transaction merger completely.

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -845,18 +847,260 @@ public Document GetRevisionBefore(DocumentsOperationContext context, string id,
...
+                    {
+                        if (document.Data != null)
```
ayende: I'm missing concurrency handling here.

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -845,18 +847,260 @@ public Document GetRevisionBefore(DocumentsOperationContext context, string id,
...
+
+            if (wasConflicted)
+            {
+                result.Progress.Warn(id, $"Skip revert, since the document was conflicted during '{before}'.");
+                return;
+            }
+
+            var collection = CollectionName.GetCollectionName(revision.Data);
+            var configuration = GetRevisionsConfiguration(collection);
+            if (configuration == _emptyConfiguration)
```
ayende: I don't understand the logic here? 
Why do you care about the revisions configuration here?

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -845,18 +847,260 @@ public Document GetRevisionBefore(DocumentsOperationContext context, string id,
...
+            {
+                result.Progress.Warn(id, $"Revision is defined for '{collection}' collection.");
+                return;
+            }
+
+            if (configuration.Disabled)
```
ayende: This shouldn't matter for the revert.
We revert _all_ found revisions.

ayende (changes requested):

src/Raven.Server/Documents/Handlers/RevisionsHandler.cs
```diff
@@ -81,6 +82,24 @@ public Task GetRevisionsFor()
             }
         }
 
+        [RavenAction("/databases/*/revisions/revert", "GET", AuthorizationStatus.ValidUser)]
+        public async Task Revert()
+        {
+            var operationId = GetLongQueryString("operationId");
+            var time = GetDateTimeQueryString("point-in-time");
```
ppekrol: Maybe just `time` ?
We are not using `-` anywhere IIRC

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -845,18 +849,317 @@ public Document GetRevisionBefore(DocumentsOperationContext context, string id,
...
+                if (result == null)
+                {
+                    progressResult.Progress.Warn(id, $"The oldest revision is after the '{parameters.Before}'.");
+                }
+
+                return result;
+            }
+        }
+
+        public class RevertProgress : IOperationProgress
```
ppekrol: Need to be moved to Raven.Client

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -845,18 +849,317 @@ public Document GetRevisionBefore(DocumentsOperationContext context, string id,
...
+                {
+                    [nameof(ScannedRevisions)] = ScannedRevisions,
+                    [nameof(ScannedDocuments)] = ScannedDocuments,
+                    [nameof(RevertedDocuments)] = RevertedDocuments,
+                    [nameof(Warnings)] = DynamicJsonValue.Convert(Warnings)
+                };
+            }
+        }
+
+        public class RevertResult : IOperationResult
```
ppekrol: Need to be moved to Raven.Client

src/Raven.Server/Documents/Handlers/RevisionsHandler.cs
```diff
@@ -81,6 +82,24 @@ public Task GetRevisionsFor()
...
+            var window = GetTimeSpanQueryString("window", required: false) ?? TimeSpan.FromHours(48);
```
ppekrol, line 89: If we consider this a 'time machine' feature, then the only value that would make that happen (keep the consistency) would be 'infinite' ?

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -845,18 +849,317 @@ public Document GetRevisionBefore(DocumentsOperationContext context, string id,
...
+
+                    if (parameters.ScannedIds.Add(id) == false)
+                        continue;
+
+                    result.Progress.ScannedDocuments++;
+
+                    if (etag > parameters.EtagBarrier)
+                    {
+                        result.Progress.Warn(id, "This document wouldn't be reverted, because it changed after the revert progress started.");
+                        continue;
```
ppekrol, line 1027: Shouldn't this be a `break` ?

karmeli87: Think of the following timeline:
1. We request the revert at local etag of 100
2. Some inserted several documents and increase the local etag to 110. 
3. We start iteration over the revisions from the top. (from etag 110)

So we want to skip those document that was changed after etag 100, but we want to revert the rest.

src/Raven.Server/Documents/Handlers/RevisionsHandler.cs
karmeli87: You mean the default value? 
The revisions aren't sorted by date so theoretically we need to iterate over everything. 
So we are using the `window` for performance sake.

ppekrol: Yes I understand that, but using any other value than infinite do not guarantee you consistency? That is my concern.

ppekrol (approved):

ayende (approved):

PR #8240, opened 2018-12-16
RavenDB-12015: Unable to start database because 'db.lock' file it is being used by another process

garayx:


src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -615,11 +616,12 @@ void AddToInitLog(string txt)
                 // if we have a very long init process, make sure that we reset the last idle time for this db.
                 LastRecentlyUsed.AddOrUpdate(databaseName, SystemTime.UtcNow, (_, time) => SystemTime.UtcNow);
 
-
                 return documentDatabase;
             }
             catch (Exception e)
             {
+                documentDatabase?.Dispose();
```
grisha-kotler, line 623: Need to make sure that we don't dispose twice.
We do dispose in `Initialize` (if we there is an exception)

grisha-kotler (changes requested):

src/Raven.Server/Documents/DatabasesLandlord.cs
ayende: Our code is fine with double disposal

grisha-kotler (approved):

PR #8242, opened 2018-12-16
RavenDB-12376: librvnpal

aviviadi:
currently implements WriteFileHeader in linux-x64 (tested)
TODO: other plats&archs plus implement publish copy of libs/librvnpal/* (csproj)


src/Voron/Platform/Pal.cs
```diff
@@ -0,0 +1,62 @@
...
+// ReSharper disable InconsistentNaming
+
+namespace Voron.Platform
+{
+    public static unsafe class Pal
+    {
+        static Pal()
+        {
+            var toFilename = LIBRVNPAL;
+            var fromFilename = $"{toFilename}.win.x64.dll";
```
ayende: why is this initialized here?

src/Voron/Platform/Pal.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                fromFilename = Environment.Is64BitProcess ? $"{toFilename}.mac.x64.dylib" : $"{toFilename}.mac.x86.dylib";
+                toFilename += ".dylib";
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                fromFilename = Environment.Is64BitProcess ? $"{toFilename}.win.x64.dll" : $"{toFilename}.win.x86.dll";
+                toFilename += ".dll";
+            }
+            else
```
ayende, line 43: ARM missing

src/Voron/Platform/PalFlags.cs
```diff
@@ -0,0 +1,189 @@
...
+// ReSharper disable InconsistentNaming
+// ReSharper disable UnusedMember.Local
+// ReSharper disable IdentifierTypo
+
+namespace Voron.Platform
+{
+    public static class PalFlags
+    {
+        [Flags]
+        public enum Errno
```
ayende: Let's remove this.
I would rather have a dedicated function in PAL that I can hand an error code and get a string back that would hold the error string.

src/Voron/Platform/PalHelper.cs
```diff
@@ -0,0 +1,25 @@
+using System;
+using System.IO;
+using static Voron.Platform.PalFlags;
+// ReSharper disable StringLiteralTypo
+
+namespace Voron.Platform
+{
+    public static class PalHelper
+    {
+        public static void ThrowLastError(PalFlags.Errno lastError, string msg)
```
ayende: Call to PAL, which will allocate a string, which we'll copy and then free.

src/Voron/StorageEnvironmentOptions.cs
```diff
@@ -666,10 +667,10 @@ public override unsafe bool ReadHeader(string filename, FileHeader* header)
             public override unsafe void WriteHeader(string filename, FileHeader* header)
             {
                 var path = _basePath.Combine(filename);
-                if (RunningOnPosix)
-                    PosixHelper.WriteFileHeader(header, path);
-                else
-                    Win32Helper.WriteFileHeader(header, path);
+                var errorCode = PalFlags.Errno.NONE;
+                var rc = Pal.rvn_write_header(path.FullPath, header, (uint)sizeof(FileHeader), ref errorCode);
```
ayende: You can define this as `out`, not `ref`

src/librvnpal/inc/rvn.h
```diff
@@ -0,0 +1,23 @@
+#ifndef RVN_H
+#define RVN_H
+
+#include <stdint.h>
+
+#if _WIN32
+_declspec(dllexport)
+#endif
+int32_t rvn_memcmp(const void *s1, const void *s2, int32_t n);
```
ayende: Remove this

src/librvnpal/inc/rvn.h
```diff
@@ -0,0 +1,23 @@
...
+
+#if _WIN32
+_declspec(dllexport)
+#endif
+uint64_t rvn_get_current_thread_id(void);
+
+#if _WIN32
+_declspec(dllexport)
+#endif
+int32_t rvn_write_header(const char *path, void *header, int32_t size, uint32_t *detailed_error_code);
```
ayende: `size` is defined as `int32_t` here, but `uint` in the pinvoke sig

src/librvnpal/src/posix/writefileheader.c
```diff
@@ -0,0 +1,248 @@
...
+
+#ifndef CIFS_MAGIC_NUMBER
+	#define CIFS_MAGIC_NUMBER     0xff534d42
+#endif
+
+typedef int bool;
+#define true 1
+#define false 0
+
+#define SUCCESS 			0
```
ayende: Shouldn't this be moved to header file?

src/librvnpal/src/posix/writefileheader.c
```diff
@@ -0,0 +1,248 @@
...
+	int rc;
+	int fd = open(dir_path, 0, 0);
+    if (fd == -1)
+    {
+		rc = FAIL_OPEN_FILE;
+		goto error_cleanup;
+	}
+
+	rc = sync_directory_allowed(fd);
+	if(rc == SYNC_DIR_ALLOWED ||
```
ayende: If checking if synced failed, we always sync ?

src/librvnpal/src/posix/writefileheader.c
```diff
@@ -0,0 +1,248 @@
...
+
+error_cleanup:
+	*detailed_error_code = errno;
+	if(fd != -1)
+		close(fd);
+	
+	return rc;
+}
+
+static int32_t sync_directory_maybe_symblink(char *dir_path, int depth, uint32_t* detailed_error_code)
```
ayende: The whole code need re-formatting

ayende (changes requested):

src/Voron/Platform/Pal.cs
aviviadi: it works (because executed before lib load) and makes logic to put it here in terms of context

ayende: Why is it init to `$"{toFilename}.win.x64.dll";` ?

PR #8245, opened 2018-12-17
This is for code review only - NOT FOR MERGE (Related to RavenDB-12503, RavenDB-12206 and RavenDB-12187)

myarichuk:
1) Make sure that filters & select in edge will work when there is string array as edge property
2) Add unit test for RavenDB-12187 - skipped for now, will be "unskipped" as the issue is fixed
3) Unskip & fix several unit tests that were skipped before (they were skipped because of unimplemented issues)

### DIFF
--- a/src/Raven.Server/Documents/Queries/Graph/SingleEdgeMatcher.cs
+++ b/src/Raven.Server/Documents/Queries/Graph/SingleEdgeMatcher.cs
@@ -7,6 +7,7 @@ using Raven.Server.Json;
 using Raven.Server.Utils;
 using Sparrow;
 using Sparrow.Json;
+using Sparrow.Json.Parsing;
 using static Raven.Server.Documents.Queries.GraphQueryRunner;
 
 namespace Raven.Server.Documents.Queries.Graph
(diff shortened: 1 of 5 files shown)


src/Raven.Server/Documents/Queries/Graph/SingleEdgeMatcher.cs
```diff
@@ -44,27 +45,59 @@ public void SingleMatch(Match left, string alias)
...
+                            switch (item)
                             {
-                                AddEdgeAfterFiltering(left, json, Edge.Project?.FieldValue);
+                                case BlittableJsonReaderObject json when Edge.Where?.IsMatchedBy(json, QueryParameters) != false:
+                                    if (projectFieldValue == null)
+                                        ThrowMissingEdgeProjection();
+                                    AddEdgeAfterFiltering(left, json, projectFieldValue);
+                                    break;
+                                case LazyStringValue lazyString:
+                                    var jsonStringForWhere = new DynamicJsonValue
```
ayende: Don't we need to check that `projectFieldValue` is null here?

myarichuk: Probably (will need to test if it can become null and how), and also we need to test for arrays of primitives that are not strings.
I did this PR because I am not 100% sure that the code so far is good enough - and I mean the code around IsMatchedBy() that I added

myarichuk:
added another commit, better handling of use-cases for RavenDB-12206
(planning to submit separate PR for RavenDB-12206, as soon as I synchronize with Tal's code changes)

myarichuk:
rewrote the commit, make sure all tests are passing

myarichuk:
will submit stuff as per-issue PRs

PR #8246, opened 2018-12-17
RavenDB-12506 PageLocator should throw on negative page numbers. In p…

arekpalinski:
…articular the issue was that it returned invalid page for -1 without any error as we initialize by -1 page numbers of pages in cache. That resulted in AVE occasionally.

### DIFF
--- a/src/Voron/PageLocator.cs
+++ b/src/Voron/PageLocator.cs
@@ -4,6 +4,7 @@ using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using Sparrow;
 using Sparrow.Binary;
+using Voron.Exceptions;
 using Voron.Impl;
 
 namespace Voron
@@ -72,6 +73,9 @@ namespace Voron
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public bool TryGetReadOnlyPage(long pageNumber, out Page page)
         {
+            if (pageNumber < 0)
+                ThrowNegativePageNumber(pageNumber);
+
             var position = pageNumber & _andMask;
 
             PageData* node = &_cache[position];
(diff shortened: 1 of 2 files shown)


arekpalinski:
@redknightlois  Please check http://issues.hibernatingrhinos.com/issue/RavenDB-12506 first for details.

The issue was for a page with number `-1` although I've added a check for any number < 0 to throw early on invalid page number.

redknightlois:
Page Locator is very performance sensitive. I would avoid to add the check if possible. Moreover this issue happens because EOOM as I have been saying all along. We have to get rid of it. @ayende

arekpalinski:
@redknightlois Can you suggest where we could move that check? What I'm trying to achieve is to avoid the page locator to return me an invalid page if I pass `-1` as the requested page number? I can move it to:

https://github.com/ravendb/ravendb/blob/077d74a1a0c0f412444a89d3cf383a662da3f3d2/src/Voron/Impl/LowLevelTransaction.cs#L440-L449

but that would be the same thing effectively, woundn't it?

Indeed, I forced EOOM to cause that but we could arrive here if we get any exception during db creation  (`StorageEnvironment.CreateNewDatabase`), it could also be due to end of disk error I think.

redknightlois:
The proper location for it would be `.GetPageInternal()`, because `PageLocator` is a cache, you should fail only when you are going to get those.

redknightlois:
@ppekrol Revert this one. Let @arekpalinski put the check in the proper place.

ppekrol:
I've wrote to him, he will address this tomorrow morning. It is not like this check is invalid. It just could be placed better.

arekpalinski:
@redknightlois that won't work because `_pageLocator.TryGetReadOnlyPage` is called _before_ `GetPageInternal` and for -1 it will return true and serve invalid page from the cache- that's the issue

redknightlois:
How is that even possible? If it is serving -1 is because someone put it in there.

arekpalinski:
That's because we initialize the cache items by -1:
https://github.com/ravendb/ravendb/blob/6bc40582e401244fa09c67c40406f3bd16290f72/src/Voron/PageLocator.cs#L63

redknightlois:
OK. I just noted that is a TryXXX method... don't throw on TRY... return `false`. 

EDIT: Throwing there have ripple effects on the inlining and code block layout. Returning 'false' does not impact anything.

arekpalinski:
Good point. I'll change it to return false if the page that is going to be returned is invalid (PageNumber == Invalid)

arekpalinski:
I've created: https://github.com/ravendb/ravendb/pull/8255

PR #8249, opened 2018-12-17
RavenDB-12502 support staleness in graph queries

talweiss1982:
* no test yet, will be added later.

### DIFF
--- a/src/Raven.Client/Documents/Queries/IndexQuery.cs
+++ b/src/Raven.Client/Documents/Queries/IndexQuery.cs
@@ -5,6 +5,7 @@
 //-----------------------------------------------------------------------
 
 using System;
+using System.Diagnostics;
 using Raven.Client.Documents.Queries.Timings;
 using Sparrow.Json;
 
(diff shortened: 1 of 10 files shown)


src/Raven.Client/Documents/Queries/IndexQuery.cs
```diff
@@ -151,6 +152,10 @@ public int PageSize
 
         public TimeSpan? WaitForNonStaleResultsTimeout { get; set; }
 
+        public (long? DocEtag, long? RefEtag)? CutoffEtag { get; set; }
+
+        public Stopwatch QueryDuration { get; set; }
+
```
ayende, line 158: I'm not sure if these should be on the client side. 
Can this be moved to `IndexQueryServerSide` ?

talweiss1982: I haven't noticed it's in the client, I navigates to the base class, yes this can be moved up to the derived class.

PR #8251, opened 2018-12-17
RavenDB-12015 - fixing AVE

grisha-kotler:
when we get an exception in the dispose of MapReduceIndexingContext we don't clear the internal structures and try to reuse it in the next transaction

### DIFF
--- a/src/Raven.Server/Documents/Indexes/MapReduce/MapReduceIndexingContext.cs
+++ b/src/Raven.Server/Documents/Indexes/MapReduce/MapReduceIndexingContext.cs
@@ -37,16 +37,21 @@ namespace Raven.Server.Documents.Indexes.MapReduce
 
         public void Dispose()
         {
-            DocumentMapEntries?.Dispose();
-            DocumentMapEntries = null;
-            MapPhaseTree = null;
-            ReducePhaseTree = null;
-            ProcessedDocEtags.Clear();
-            ProcessedTombstoneEtags.Clear();
-            StoreByReduceKeyHash.Clear();
-            FreedPages.Clear();
-
-            StoreNextMapResultId();
+            try
+            {
+                StoreNextMapResultId();
+            }
+            finally
+            {
+                DocumentMapEntries?.Dispose();
+                DocumentMapEntries = null;
+                MapPhaseTree = null;
+                ReducePhaseTree = null;
+                ProcessedDocEtags.Clear();
+                ProcessedTombstoneEtags.Clear();
+                StoreByReduceKeyHash.Clear();
+                FreedPages.Clear();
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Documents/Indexes/MapReduce/MapReduceIndexingContext.cs
```diff
@@ -43,15 +45,27 @@ public void Dispose()
             ProcessedTombstoneEtags.Clear();
             StoreByReduceKeyHash.Clear();
             FreedPages.Clear();
+
+            StoreNextMapResultId();
```
ppekrol: This will throw NRE on MapPhaseTree

grisha-kotler: done

ppekrol (approved):

ayende (approved):

ayende:
Note that this PR is fine, but the 4.0 one looks broken to me

src/Raven.Server/Documents/Indexes/MapReduce/MapReduceIndexingContext.cs
```diff
@@ -34,24 +37,40 @@ static MapReduceIndexingContext()
...
-            DocumentMapEntries = null;
-            MapPhaseTree = null;
-            ReducePhaseTree = null;
-            ProcessedDocEtags.Clear();
-            ProcessedTombstoneEtags.Clear();
-            StoreByReduceKeyHash.Clear();
-            FreedPages.Clear();
+            try
+            {
+                StoreNextMapResultId();
```
arekpalinski, line 42: I think we talked about moving it to the end of the reduce phase. Why not to call it here:

https://github.com/ravendb/ravendb/blob/5951c3593455d7a42024f538bf30183bfc0c2900/src/Raven.Server/Documents/Indexes/MapReduce/ReduceMapResultsBase.cs#L149-L151

?

arekpalinski: The problem with calling `StoreNextMapResultId();` here, in `Dispose()`, is that throwing here might hide the previous exception if it was already thrown. In general it's bad idea to have `Dispose` method that can throw I think.

If we moved it next to `WriteEtags` it would be called only if we complete the batch, without any unexpected errors, and that's when we actually want to store `NextMapResultId` (otherwise transaction won't be committed anyway)

grisha-kotler: Done

PR #8252, opened 2018-12-17
RavenDB-12015 - fixing AVE

grisha-kotler:
when we get an exception in the dispose of MapReduceIndexingContext we don't clear the internal structures and try to reuse it in the next transaction

### DIFF
--- a/src/Raven.Server/Documents/Indexes/MapReduce/MapReduceIndexingContext.cs
+++ b/src/Raven.Server/Documents/Indexes/MapReduce/MapReduceIndexingContext.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using Sparrow;
+using Sparrow.Logging;
 using Voron;
 using Voron.Data.BTrees;
 using Voron.Data.Fixed;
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Documents/Indexes/MapReduce/MapReduceIndexingContext.cs
```diff
@@ -40,24 +43,41 @@ static MapReduceIndexingContext()
...
-            ProcessedDocEtags.Clear();
-            ProcessedTombstoneEtags.Clear();
-            StoreByReduceKeyHash.Clear();
+            try
+            {
+                StoreNextMapResultId();
+            }
+            finally
+            {
+                StoreNextMapResultId();
```
ayende: Why are you calling `StoreNextMapResultId` twice? Once in the `try` and once in the `finally` ?

ayende (changes requested):

src/Raven.Server/Documents/Indexes/MapReduce/MapReduceIndexingContext.cs
grisha-kotler: This wasn't ported correctly from v4.1. Fixed.

PR #8258, opened 2018-12-18
Proper fix for RavenDB-12206

myarichuk:
1) Make sure that filters & select in edge will work when there is string array as edge property
2) Make sure to properly handle non-string arrays as edge properties.


src/Raven.Server/Documents/Queries/Graph/SingleEdgeMatcher.cs
```diff
@@ -41,47 +41,95 @@ public void SingleMatch(Match left, string alias)
 
             if (Edge.Where != null || Edge.Project != null)
             {
-                if (BlittableJsonTraverser.Default.TryRead(leftDoc, Edge.Path.FieldValue, out var value, out _) == false)
+                if (BlittableJsonTraverser.Default.TryRead(leftDoc, Edge.Path.FieldValue, out var value, out _) == false || value == null)
                     return;
 
-                if (Edge.Project == null)
+                //allow array of primitives
+                if (Edge.Project == null && !(value is BlittableJsonReaderArray))
```
ayende: You can probably remove this and do the check in the `switch`

src/Raven.Server/Documents/Queries/Graph/SingleEdgeMatcher.cs
```diff
@@ -41,47 +41,95 @@ public void SingleMatch(Match left, string alias)
...
-            {
+            {              
                 AddEdgeAfterFiltering(left, leftDoc, Edge.Path.FieldValue);
             }
         }
 
+        private void HandleStringAsArrayItem(Match left, LazyStringValue lazyStringValue)
+        {
+            var jsonStringForWhere = new DynamicJsonValue
+                {[Edge.Path.FieldValue] = lazyStringValue};
```
ayende: Reformat this

PR #8263, opened 2018-12-18
RavenDB-12536: add error state to LowLevelTransaction and handle it.

garayx:


ayende (approved):

ayende:
I'm still missing the allocator poisoning the transaction if there is an allocation failure.

ayende:
Please check the failing test

arekpalinski:
I wonder if we should mark the transaction as `Errored` not only when we get an exception on `Allocate()` and `Free()` but also when `VoronUnrecoverableErrorException` is raised (by passing the transaction to `VoronUnrecoverableErrorException.Raise` where possible and setting the error state then)

ayende:
Yes, I think that would be wise.
Basically, we must ensure that if we got any error in the transaction, the only thing it is valid to do is to rollback / dispose

src/Voron/Exceptions/VoronUnrecoverableErrorException.cs
```diff
@@ -6,11 +6,25 @@
...
 namespace Voron.Exceptions
 {
     public class VoronUnrecoverableErrorException : Exception
     {
+        public static void Raise(LowLevelTransaction tx, string message)
+        {
+            try
+            {
+                tx.MarkTransactionAsFailed();
+                throw new VoronUnrecoverableErrorException(message);
```
arekpalinski: Before we mark the transaction as failed let's add the current state of the transaction to the exception message. I'd like to be able to distinguish if the unrecoverable exception was thrown on a transaction in valid state or already marked as errored (just in case).

garayx: done

src/Voron/Impl/LowLevelTransaction.cs
```diff
@@ -1112,6 +1112,21 @@ public void RetrieveCommitStats(out CommitStats stats)
             _requestedCommitStats = stats = new CommitStats();
         }
 
+        public string GetTxState()
+        {
+            switch (_disposed)
```
ayende: Just return the enum value and let the `.ToString()` handle this

ayende (changes requested):

src/Voron/Exceptions/VoronUnrecoverableErrorException.cs
```diff
@@ -6,11 +6,26 @@
...
 {
     public class VoronUnrecoverableErrorException : Exception
     {
+        public static void Raise(LowLevelTransaction tx, string message)
+        {
+            try
+            {
+                var lastTxState = tx.GetTxState();
+                tx.MarkTransactionAsFailed();
+                throw new VoronUnrecoverableErrorException($"txState: {lastTxState}, message: {message}");
```
arekpalinski: The tx state will be mostly for debugging / troubleshooting so let's put it _after_ the actual message (`$"{message}. Current tx state: {lastTxState}"`

src/Voron/Impl/LowLevelTransaction.cs
garayx: done

src/Voron/Exceptions/VoronUnrecoverableErrorException.cs
garayx: done

ppekrol:
@garayx Can you fix the conflicts?

garayx:
> @garayx Can you fix the conflicts?

done

PR #8265, opened 2018-12-18
RavenDB-12539: Increasing the read buffer to improve copy performance.

redknightlois:


src/Voron/Data/BTrees/Tree.Stream.cs
```diff
@@ -133,6 +133,8 @@ public void Write(Stream stream)
 
                     _parent._tx.LowLevelTransaction.ShrinkOverflowPage(_currentPage.PageNumber, chunkSize + infoSize, _parent.State); 
                 }
+
+                ArrayPool<byte>.Shared.Return(_localBuffer);
```
grisha-kotler, line 137: This should be in try, finally.
We can have errors during copy, out of disk space, out of memory...

grisha-kotler (changes requested):

src/Voron/Data/BTrees/Tree.Stream.cs
redknightlois: Yeah, I looked into that, but decided not to do it. Worse that can happen is your 512Kb array gets collected. Meanwhile, if you add a Try-Finally you are paying for the cost of exception handling prolog and epilog every time.

PR #8266, opened 2018-12-18
RavenDB-12524 - VoronUnrecoverableErrorException when the disk is full

grisha-kotler:
http://issues.hibernatingrhinos.com/issue/RavenDB-12524

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -87,11 +87,15 @@ namespace Raven.Server.Documents.Indexes
 
         private long _analyzerErrors;
 
+        private long _diskFullErrors;
+
         private const long WriteErrorsLimit = 10;
 
         private const long UnexpectedErrorsLimit = 3;
 
         private const long AnalyzerErrorLimit = 0;
+
+        private const long DiskFullErrorLimit = 10;
 
         protected Logger _logger;
 
@@ -1053,6 +1057,10 @@ namespace Raven.Server.Documents.Indexes
                                 {
                                     HandleExcessiveNumberOfReduceErrors(scope, enre);
                                 }
+                                catch (DiskFullException dfe)
+                                {
+                                    HandleDiskFullErrors(scope, dfe);
+                                }
                                 catch (OperationCanceledException)
                                 {
                                     // We are here only in the case of indexing process cancellation.
@@ -1301,6 +1309,7 @@ namespace Raven.Server.Documents.Indexes
             Interlocked.Exchange(ref _writeErrors, 0);
             Interlocked.Exchange(ref _unexpectedErrors, 0);
             Interlocked.Exchange(ref _analyzerErrors, 0);
+            Interlocked.Exchange(ref _diskFullErrors, 0);
         }
 
         internal void HandleAnalyzerErrors(IndexingStatsScope stats, IndexAnalyzerException iae)
@@ -1377,6 +1386,22 @@ namespace Raven.Server.Documents.Indexes
                 return;
 
             _errorStateReason = e.Message;
+            SetState(IndexState.Error, ignoreWriteError: true);
+        }
+
+        internal void HandleDiskFullErrors(IndexingStatsScope stats, DiskFullException dfe)
+        {
+            if (_logger.IsOperationsEnabled)
+                _logger.Operations($"Disk full error occurred for '{Name}'.", dfe);
+
+            stats.AddDiskFullError(dfe);
+
+            var diskFullErrors = Interlocked.Increment(ref _diskFullErrors);
+
+            if (State == IndexState.Error || diskFullErrors < DiskFullErrorLimit)
+                return;
+
+            _errorStateReason = $"State was changed due to excessive number of disk full errors ({diskFullErrors}).";
             SetState(IndexState.Error, ignoreWriteError: true);
         }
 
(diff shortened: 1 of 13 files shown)


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1573,12 +1598,15 @@ public bool DoIndexingWork(IndexingStatsScope stats, CancellationToken cancellat
...
                             SlowWriteNotification.Notify(commitStats, DocumentDatabase);
-
                             stats.RecordCommitStats(commitStats.NumberOfModifiedPages, commitStats.NumberOf4KbsWrittenToDisk);
                         }
                     }
                     catch
                     {
+                        // need to call it here to the let the index continue running
+                        // we'll stop when we reach the index error threshold 
+                        _mre.Set();
```
ayende: This is called on _any_ exception, but isn't the intent just for disk full?

src/Raven.Server/Indexing/LuceneVoronDirectory.cs
```diff
@@ -127,6 +126,12 @@ public override void DeleteFile(string name, IState s)
             if (readResult == null)
                 throw new FileNotFoundException("Could not find file", name);
 
+            if (_indexOutputFilesSummary.HasVoronWriteErrors)
```
ayende: Move this to the top of the method.
Probably a good idea to do this on the Write as well

ayende: Also, on Dispose, so it wouldn't try to create a new stream on another file?

src/Raven.Server/Indexing/VoronIndexOutput.cs
```diff
@@ -73,11 +114,54 @@ protected override void Dispose(bool disposing)
...
                 PosixFile.DeleteOnClose(_fileTempPath);
             }
         }
+
+        private void ThrowDiskFullException()
+        {
+            var directory = Path.GetDirectoryName(_fileTempPath);
+            var driveInfo = DiskSpaceChecker.GetDiskSpaceInfo(directory);
+            throw new DiskFullException($"There isn't enough space to flush the buffer of the file: {_fileTempPath}. " +
+                                        $"Currently available space: {driveInfo?.TotalFreeSpace}");
```
ayende: Handle the case where this is null and say `space: N/A` or something

ayende (changes requested):

src/Raven.Server/Indexing/LuceneVoronDirectory.cs
grisha-kotler: done

src/Raven.Server/Indexing/VoronIndexOutput.cs
grisha-kotler: done

src/Raven.Server/Documents/Indexes/Index.cs
grisha-kotler: This is for any exception. We have error handling and after a couple of errors we set the index as errored. Unless we do that the index just stops

arekpalinski: This is okay for me but I'm a bit worried that we'll error indexes very quickly after getting any error here  (e.g. what is the chance we'll recover from the disk full exception, we'll repeat it 10 times and mark the index as errored).

I wonder if we should have a notion of fallback mode - e.g. after getting end of disk space then wait 10 seconds, and run indexing then.. next wait 20 seconds etc until the max fallback mode duration.

ayende: Sounds good to me.
We should also force a cleanup of our state in this case, so a flush / sync of Voron and then ask to remove all the old recylced journals, old scratches, etc.

grisha-kotler:
Done

src/Raven.Server/Documents/Indexes/Index.cs
grisha-kotler: done

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1385,23 +1389,51 @@ internal void HandleExcessiveNumberOfReduceErrors(IndexingStatsScope stats, Exce
...
-            if (State == IndexState.Error || diskFullErrors < DiskFullErrorLimit)
+                // force flush and sync
+                GlobalFlushingBehavior.GlobalFlusher.Value.MaybeFlushEnvironment(storageEnvironment);
+                if (_logsAppliedEvent.Wait(timeToWaitInSeconds))
+                {
+                    storageEnvironment.ForceSyncDataFile();
+                }
+
+                // wait for sync
+                Task.Delay(1000 * timeToWaitInSeconds, _indexingProcessCancellationTokenSource.Token).Wait();
```
ayende: Why are you doing this? We are effectively waiting twice.
In any case, worse case we can try again and we'll wait for the sync on the next iteration, no?

grisha-kotler: fixed the calculations

arekpalinski:
@grisha-kotler Please note failing test: http://jenkins.hibernatingrhinos.com/job/PR_Tests_v4.1/2280/

arekpalinski:
We have it reported as failing occasionally: http://issues.hibernatingrhinos.com/issue/RavenDB-12325

PR #8267, opened 2018-12-18
fix for RavenDB-12187

myarichuk:
In graph queries, support equals and not equals between aliases in "where" clauses, so the following queries *should* work now:
```
match (Dogs as d1)-[Likes]->(Dogs as d2) 
where d1 != d2
```
Note that the following won't work at this stage (if I understand correctly, @talweiss1982 is going to handle this use-case in one of his tasks)
```
match (Dogs as d1)-[Likes]->(Dogs as d2) 
where id(d1) != id(d2)
```
In general, any method call in "where" clauses won't be parsed correctly by the QueryParser (more specifically, this is because QueryParser::Operator() method lacks the necessary code to parse this)

### DIFF
--- a/src/Raven.Server/Documents/Queries/AST/QueryExpression.cs
+++ b/src/Raven.Server/Documents/Queries/AST/QueryExpression.cs
@@ -130,6 +130,11 @@ namespace Raven.Server.Documents.Queries.AST
         {
             switch (left)
             {
+                case BlittableJsonReaderObject leftBlittableJson:
+                    if (right is BlittableJsonReaderObject rightBlittableJson)
+                        return leftBlittableJson.Equals(rightBlittableJson, true);
+
+                    throw new NotSupportedException("Cannot compare objects with non-objects.");
                 case long l:
                     switch (right)
                     {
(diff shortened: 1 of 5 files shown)


src/Raven.Server/Documents/Queries/AST/QueryExpression.cs
```diff
@@ -130,6 +130,11 @@ private static bool AreEqual(object left, object right)
         {
             switch (left)
             {
+                case BlittableJsonReaderObject leftBlittableJson:
+                    if (right is BlittableJsonReaderObject rightBlittableJson)
+                        return leftBlittableJson.Equals(rightBlittableJson, true);
```
ayende, line 135: What is the `true` here stands for?

src/Raven.Server/Documents/Queries/AST/QueryExpression.cs
```diff
@@ -130,6 +130,11 @@ private static bool AreEqual(object left, object right)
...
+
+                    throw new NotSupportedException("Cannot compare objects with non-objects.");
```
ayende, line 137: Return false here, we don't want to throw based on the data of the values.

ayende:
Failing test fixed here: 
https://github.com/ravendb/ravendb/commit/b6247e57553e15bb5e77fd3a11b3b65fb31b70a8

PR #8268, opened 2018-12-18
RavenDB-12502 Graph query staleness

talweiss1982:
* also includes a fix for the Collection Destination optimization

### DIFF
--- a/src/Raven.Server/Documents/Queries/Graph/GraphQueryIndexGatherer.cs
+++ b/src/Raven.Server/Documents/Queries/Graph/GraphQueryIndexGatherer.cs
@@ -1,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using NCrontab.Advanced.Extensions;
+using Raven.Server.Documents.Indexes;
+
+namespace Raven.Server.Documents.Queries.Graph
+{
+    public class GraphQueryIndexNamesGatherer: QueryPlanVisitor
+    {
+        public List<string> Indexes { get; set; } = new List<string>();
+
+        public override void VisitQueryQueryStep(QueryQueryStep qqs)
+        {
+            var name = qqs.GetIndexName;
+            if(name.IsNullOrWhiteSpace() == false)
+                Indexes.Add(name);
+        }
+    }
+}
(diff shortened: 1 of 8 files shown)


src/Raven.Server/Documents/Queries/Graph/GraphQueryIndexGatherer.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Raven.Server.Documents.Queries.Graph
+{
+    public class GraphQueryIndexNamesGatherer: QueryPlanVisitor
+    {
+        public List<string> Indexes { get; set; } = new List<string>();
+
+        public override void VisitQueryQueryStep(QueryQueryStep qqs)
+        {
+            var name = qqs.GetIndexName;
+            if(name.IsNullOrWhiteSpace() == false)
```
ayende, line 16: I'm assuming that this filters out the collection queries?

src/Raven.Server/Documents/Queries/Graph/GraphQueryPlan.cs
```diff
@@ -196,5 +199,87 @@ public List<Match> Execute()
...
+                {
+                    //we might get here if all indexes have timedout 
+                    return indexTimedout;
+                }
+            }
+        }
+
+        private List<Index> GetStaleIndexes(List<Index> indexes, long etag)
+        {
+            var staleList = new List<Index>();
```
ayende, line 272: Why not re-use the same instance? I don't see it being used for other stuff?

talweiss1982: I can't modify the list while iterating on it, so I need a 'to delete' list anyway so I might as well just create a new one.

src/Raven.Server/Documents/Queries/Graph/GraphQueryIndexGatherer.cs
talweiss1982: I'm not sure I need to verify this.

talweiss1982: yes a collection query will return null

PR #8272, opened 2018-12-19
RavenDB-12524 After we completed recovery, we are now going to valida…

ayende:
…te the checksum of all modified pages to ensure that if we have corrupted state, we'll learn about this early

### DIFF
--- a/src/Voron/Impl/Journal/WriteAheadJournal.cs
+++ b/src/Voron/Impl/Journal/WriteAheadJournal.cs
@@ -143,6 +143,7 @@ namespace Voron.Impl.Journal
 
             return journal;
         }
+
         public bool RecoverDatabase(TransactionHeader* txHeader, Action<string> addToInitLog)
         {
             // note, we don't need to do any concurrency here, happens as a single threaded
@@ -173,6 +174,9 @@ namespace Voron.Impl.Journal
 
             var lastSyncedTransactionId = logInfo.LastSyncedTransactionId;
 
+
+            var modifiedPages = new HashSet<long>();
+
             var journalFiles = new List<JournalFile>();
             long lastSyncedTxId = -1;
             long lastSyncedJournal = logInfo.LastSyncedJournal;
@@ -187,7 +191,7 @@ namespace Voron.Impl.Journal
                     RecoverCurrentJournalSize(pager);
 
                     var transactionHeader = txHeader->TransactionId == 0 ? null : txHeader;
-                    using (var journalReader = new JournalReader(pager, _dataPager, recoveryPager, lastSyncedTransactionId, transactionHeader))
+                    using (var journalReader = new JournalReader(pager, _dataPager, recoveryPager, modifiedPages, lastSyncedTransactionId, transactionHeader))
                     {
                         var transactionHeaders = journalReader.RecoverAndValidate(_env.Options);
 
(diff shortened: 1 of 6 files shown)


src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -222,6 +226,33 @@ public bool RecoverDatabase(TransactionHeader* txHeader, Action<string> addToIni
...
+                {
+                    if (minPageChecked >= modifedPage)
+                        continue;
+
+                    var ptr = (PageHeader*)_dataPager.AcquirePagePointerWithOverflowHandling(tempTx, modifedPage, null);
+                    _env.ValidatePageChecksum(modifedPage, ptr);
+
+                    minPageChecked = modifedPage + VirtualPagerLegacyExtensions.GetNumberOfPages(ptr);
+
+                    tempTx.Dispose(); // release any resources, we just wanted to validate things
```
karmeli87, line 252: This should be outside the `foreach`, no?

ayende: No, we want to clear its state after every page, to avoid filling up the memory with mapped pages in the case of 32 bits

src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -222,6 +226,33 @@ public bool RecoverDatabase(TransactionHeader* txHeader, Action<string> addToIni
...
+                // to pages that overwrite one another. So we have to do this at the end, this will detect
+                // corruption when applying journals at recovery time rather than at usage.
+                var tempTx = new TempPagerTransaction();
+
+                var sortedPages = modifiedPages.ToArray();
+                Array.Sort(sortedPages);
+
+                long minPageChecked = -1;
+
+                foreach (var modifedPage in modifiedPages)
```
arekpalinski, line 242: Usage of wrong variable in the loop - `modifiedPages` instead of `sortedPages`. That's why it wasn't working.

Already handled differently in https://github.com/ravendb/ravendb/pull/8303 and removed the overflow checks from here in https://github.com/ravendb/ravendb/pull/8308

src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -222,6 +226,33 @@ public bool RecoverDatabase(TransactionHeader* txHeader, Action<string> addToIni
...
+                {
+                    if (minPageChecked >= modifedPage)
+                        continue;
+
+                    var ptr = (PageHeader*)_dataPager.AcquirePagePointerWithOverflowHandling(tempTx, modifedPage, null);
+                    _env.ValidatePageChecksum(modifedPage, ptr);
+
+                    minPageChecked = modifedPage + VirtualPagerLegacyExtensions.GetNumberOfPages(ptr);
```
arekpalinski, line 250: BTW we could unintentionally skip validation of some pages as this line should be:
`modifedPage + VirtualPagerLegacyExtensions.GetNumberOfPages(ptr) - 1` (note `- 1`)

src/Voron/Impl/Journal/WriteAheadJournal.cs
arekpalinski: Another case: https://github.com/ravendb/ravendb/pull/8313

PR #8276, opened 2018-12-19
RavenDB-12525 Failing test: SlowTests.Issues.RavenDB_5435.CanCompact

haludi:
When there was nothing to sync the sync operation stop to collect the state
but instead of exiting the operation is continued to sync and update
because the operation didn't collect the state it stayed with the default values that were 0
and the header was updated to that values

Main:
http://issues.hibernatingrhinos.com/issue/RavenDB-12525

Related:
http://issues.hibernatingrhinos.com/issue/RavenDB-12351
http://issues.hibernatingrhinos.com/issue/RavenDB-12517

### DIFF
--- a/src/Voron/Impl/Journal/WriteAheadJournal.cs
+++ b/src/Voron/Impl/Journal/WriteAheadJournal.cs
@@ -574,7 +574,7 @@ namespace Voron.Impl.Journal
 
                     ApplyJournalStateAfterFlush(token, lastProcessedJournal, lastFlushedTransactionId, unusedJournals, byteStringContext);
 
-                    _waj._env.SuggestSyncDataFileSyncDataFile();
+                    _waj._env.SuggestSyncDataFile();
                 }
                 finally
                 {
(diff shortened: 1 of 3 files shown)


src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -1000,21 +1017,17 @@ public bool WaitForTaskToBeDone(Action task)
                         return false;
                     }
                 }
-
+                
                 public void RunTaskIfNotAlreadyRan()
                 {
-                    Debug.Assert(Monitor.IsEntered(_lock));
-
-                    if (_token.IsCancellationRequested)
```
ppekrol, line 1008: Why we removed that?

haludi: I passed it to [_token.ThrowIfCancellationRequested();](https://github.com/ravendb/ravendb/pull/8276/files/cd455fc9b8f00db11d6453bba4822982afda6e07#diff-33af42ec0439152b32cc47c6739f3211R1029). 
For now it will throw exception the `LockTaskResponsible` will save it and [throw it again](https://github.com/ravendb/ravendb/pull/8276/files/cd455fc9b8f00db11d6453bba4822982afda6e07#diff-33af42ec0439152b32cc47c6739f3211R1010) in `WaitForTaskToBeDone` like all exceptions
and just in case of `OperationCanceledException` it will [catch it and return false](https://github.com/ravendb/ravendb/pull/8276/files/cd455fc9b8f00db11d6453bba4822982afda6e07#diff-33af42ec0439152b32cc47c6739f3211R1015)

arekpalinski (approved):

PR #8285, opened 2018-12-19
RavenDB-12015 - remove StoreNextMapResultId from the dispose method

grisha-kotler:


ppekrol (approved):

src/Raven.Server/Documents/Indexes/MapReduce/ReduceMapResultsBase.cs
```diff
@@ -147,6 +147,7 @@ static ReduceMapResultsBase()
             }
 
             WriteLastEtags(indexContext);
+            _mapReduceContext.StoreNextMapResultId();
```
ppekrol, line 150: @grisha-kotler @arekpalinski Do we need to call it also here?
https://github.com/ravendb/ravendb/blob/98d37f5a1f1754ee7062fc4523f4020a47f65cf5/src/Raven.Server/Documents/Indexes/MapReduce/ReduceMapResultsBase.cs#L81-L85

grisha-kotler: The MapResultId shouldn't change if we filtered all results

ppekrol: It looks like it, but wanted to double confirm

ayende (approved):

src/Raven.Server/Documents/Indexes/MapReduce/ReduceMapResultsBase.cs
arekpalinski: Yes, we didn't do any work during the map phase.

PR #8295, opened 2018-12-20
RavenDB-12537 Treat exceptions in the Apply of the Cluster State Machine with care

karmeli87:


src/Raven.Server/Rachis/Follower.cs
```diff
@@ -120,13 +117,46 @@ private void FollowerSteadyState()
...
+                                    entries.Clear();
+                                }
+                                catch
+                                {
+                                    var processed = 0;
+                                    try
+                                    {
+                                        foreach (var entry in entries)
+                                        {
+                                            (hasRemovedFromTopology, lastLogIndex, lastTruncate, lastCommit) = ApplyLeaderStateToLocalState(sp,
```
ayende, line 138: I don't follow what is going on when an error is thrown here

src/Raven.Server/Rachis/Follower.cs
```diff
@@ -120,13 +117,46 @@ private void FollowerSteadyState()
...
+                                            {
+                                                break;
+                                            }
+
+                                            processed++;
+                                        }
+                                    }
+                                    finally
+                                    {
+                                        entries.RemoveRange(0, processed);
```
ayende, line 156: We'll have entries from previous reads, then try to re-process them?
But we wait for the next network read?

src/Raven.Server/Rachis/Leader.cs
```diff
@@ -454,28 +454,83 @@ private void OnVoterConfirmation()
...
+                    if (_lastCommit >= maxIndexOnQuorum ||
+                        maxIndexOnQuorum == 0)
+                        return;
+
+                    if (_engine.GetTermForKnownExisting(context, maxIndexOnQuorum) < Term)
+                        return;
+
+                    var changedFromLeaderElectToLeader = _engine.TakeOffice();
+
+                    var lastCommit = _engine.Apply(context, _lastCommit + 1, this, sp);
```
ayende, line 517: And if this fail?

PR #8298, opened 2018-12-20
RavenDB-12096 support exact method in edge where clause

talweiss1982:


src/Raven.Server/Documents/Queries/AST/QueryExpression.cs
```diff
@@ -32,7 +32,7 @@ private static object Evaluate(QueryExpression q, BlittableJsonReaderObject valu
             }
         }
 
-        public static bool IsMatchedBy(this QueryExpression where, BlittableJsonReaderObject value, BlittableJsonReaderObject queryParameters)
+        public static bool IsMatchedBy(this QueryExpression where, BlittableJsonReaderObject value, BlittableJsonReaderObject queryParameters, bool shouldCaseSensitiveStringCompare = false)
```
ayende, line 35: Future note, this should be changed to an options object if/when we'll have more state here

ayende (approved):

PR #8301, opened 2018-12-20
RavenDB-12537 Treat exceptions in the Apply of the Cluster State Machine with care

karmeli87:


src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1014,58 +989,57 @@ private void UpdateDatabase(TransactionOperationContext context, string type, Bl
...
-                            {
-                                items.DeleteByKey(valueNameToDeleteLowered);
-                            }
-                        }
+                    updateCommand.Initialize(serverStore, context);
+                    string relatedRecordIdToDelete;
+
+                    try
+                    {
+                        relatedRecordIdToDelete = updateCommand.UpdateDatabaseRecord(databaseRecord, index);
```
ayende, line 1014: What about if there is an _older_ command in the same transaction?

karmeli87: Not sure I'm following. 
But as long as we end up with the same state across the cluster it is fine.

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -681,10 +680,8 @@ private unsafe void RemoveNodeFromDatabase(TransactionOperationContext context,
             using (Slice.From(context.Allocator, "db/" + databaseName, out Slice key))
             {
                 if (items.ReadByKey(lowerKey, out TableValueReader reader) == false)
-                {
-                    NotifyLeaderAboutError(index, leader, new InvalidOperationException($"The database {databaseName} does not exists"));
-                    return;
-                }
+                    throw new RachisApplyException($"The database {databaseName} does not exists");
```
talweiss1982, line 684: I wonder if we can get to a state where the database doesn't exist since we haven't created it yet on this node and if so this is a case that could be specific to some nodes and not consistent across the cluster.
I know that if there is a creation command then it should be executed before the deletion but we have a two-phase deletion process and I have a gut feeling that there is an edge case lurking here.

ayende: No, because this is running in the cluster, so this is running in the same sequence everywhere

src/Raven.Server/Rachis/Exceptions.cs
```diff
@@ -76,4 +76,24 @@ public static void Throw(string msg)
...
+        {
+        }
+
+        public RachisApplyException(string message, Exception innerException) : base(message, innerException)
+        {
+        }
+
+        public static void Throw(string msg)
+        {
+            throw new RachisConcurrencyException(msg);
```
ppekrol: Should be `RachisApplyException`

ppekrol: I think that this is not used?

karmeli87: Copy paste failure :/

ppekrol (approved):

PR #8304, opened 2018-12-20
RavenDB-10585: Check dataDir when MigratingFromLegacyData, make sure the tools is closed if we abort the operation, added ProcessExtensions.

garayx:


src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -1092,10 +1094,19 @@ public async Task MigrateDatabaseOffline()
             var overallProgress = result.Progress as SmugglerResult.SmugglerProgress;
             var operationId = ServerStore.Operations.GetNextOperationId();
             var processDone = new AsyncManualResetEvent(token.Token);
-
             // send new line to avoid issue with read key 
             process.StandardInput.WriteLine();
-            process.Exited += (sender, e) => { processDone.Set(); };
+            process.Exited += (sender, e) =>
```
ppekrol, line 1092: It would be better to attach to Exited before starting the process.

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -1173,12 +1178,34 @@ public async Task MigrateDatabaseOffline()
...
+                                onProgress.Invoke(result.Progress);
+
+                                var killed = ProcessExtensions.TryKill(process);
+                                throw new InvalidOperationException($"{errorString} Process pid: {process.Id}, killed: {killed}");
+                            }
                         }
                         finally
                         {
-                            if (string.IsNullOrEmpty(tmpFile) == false)
+                            if (process.HasExited && string.IsNullOrEmpty(tmpFile) == false)
```
ppekrol, line 1209: Are we expecting here that it might not be killed/exited earlier?

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
ayende: Better is an understatement. This is required

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
ayende: If it didn't, we need to kill it, and the code above certain suggest we try

ayende (changes requested):

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
garayx: done

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
garayx: done

PR #8305, opened 2018-12-21
RavenDB-12554 - Early out of memory notification is triggered too soon

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-12554

### DIFF
--- a/src/Raven.Server/Config/Categories/MemoryConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/MemoryConfiguration.cs
@@ -22,6 +22,12 @@ namespace Raven.Server.Config.Categories
         [ConfigurationEntry("Memory.LowMemoryLimitInMb", ConfigurationEntryScope.ServerWideOnly)]
         public Size LowMemoryLimit { get; set; }
 
+        [Description("The minimum amount of available commited memory RavenDB will attempt to achieve (free commited memory lower than this value will trigger low memory behavior)")]
+        [DefaultValue(512)]
+        [SizeUnit(SizeUnit.Megabytes)]
+        [ConfigurationEntry("Memory.LowMemoryCommitLimitInMb", ConfigurationEntryScope.ServerWideOnly)]
+        public Size LowMemoryCommitLimitInMb { get; set; }
+
         [Description("EXPERT: The minimum amount of committed memory percentage that RavenDB will attempt to ensure remains available. Reducing this value too much may cause RavenDB to fail if there is not enough memory available for the operation system to handle operations.")]
         [DefaultValue(0.05f)]
         [SizeUnit(SizeUnit.Megabytes)]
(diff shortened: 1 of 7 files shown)


src/Raven.Server/Config/Categories/MemoryConfiguration.cs
```diff
@@ -22,10 +22,16 @@ public MemoryConfiguration()
...
         [DefaultValue(0.05f)]
         [SizeUnit(SizeUnit.Megabytes)]
         [ConfigurationEntry("Memory.MinimumFreeCommittedMemoryPercentage", ConfigurationEntryScope.ServerWideOnly)]
-        public float MinimumFreeCommittedMemory { get; set; }
+        public float MinimumFreeCommittedMemoryPercentage { get; set; }
+
+        [Description("EXPERT: The maximum amount of committed memory that RavenDB will attempt to ensure remains available. Reducing this value too much may cause RavenDB to fail if there is not enough memory available for the operation system to handle operations.")]
+        [DefaultValue(128)]
+        [SizeUnit(SizeUnit.Megabytes)]
+        [ConfigurationEntry("Memory.MaxFreeCommittedMemoryToKeep", ConfigurationEntryScope.ServerWideOnly)]
```
ppekrol: MaxFreeCommittedMemoryToKeepInMb

src/Sparrow/LowMemory/LowMemoryNotification.cs
```diff
@@ -323,12 +318,7 @@ public bool IsLowMemory(MemoryInfoResult memInfo, SmapsReader smapsReader)
                 isLowMemory = IsAvailableMemoryBelowThreshold(memInfo);
             }
 
-            if (PlatformDetails.RunningOnPosix == false)
-            {
-                // this is relevant only on Windows
-                var commitChargePlusMinSizeToKeepFree = memInfo.CurrentCommitCharge + GetCommitChargeThreshold(memInfo);
-                isLowMemory |= memInfo.TotalCommittableMemory <= commitChargePlusMinSizeToKeepFree;
-            }
+            isLowMemory |= MemoryInformation.IsEarlyOutOfMemory(memInfo);
```
ppekrol: Not sure if this is correct, because we will have a low memory exactly when first EOOM can be thrown and we probably want to do it a bit earlier to give a chance to cleanup a bit and to not throw the EOOM?

src/Raven.Server/Config/Categories/MemoryConfiguration.cs
grisha-kotler: done

src/Sparrow/LowMemory/LowMemoryNotification.cs
grisha-kotler: I suggest adding another configuration: LowMemoryCommitLimitInMb (by default set to 3 times the MaxFreeCommittedMemoryToKeepInMb).
For a 16GB machine:
MaxFreeCommittedMemoryToKeepInMb: 128MB, lower than this will throw EOOM
LowMemoryCommitLimitInMb: 384MB, lower than this, will trigger low memory notification

ayende: That sounds fine, but it should be at least 512MB by default or 5% (smallest).

grisha-kotler: done

Issue #21, opened 2010-03-04
Keep track of indexing errors


ayende:
Simple numbers:
Number of total indexed documents (including re-indexing).
Number of failed indexing attempts (threw an exception)

Issue #20, opened 2010-03-03
Add user visible log


ayende:
Where they can see things like errors from index functions

PR #8313, opened 2018-12-21
RavenDB-12524 Recovery needs to take into account that a modified pag…

arekpalinski:
…e could be freed and overwritten in later transactions. So we need to filter them out. Eg.:

- tx1: allocate overflow on page #20
- tx2: free page #20
- tx3: allocate page #21

During the recovery we must not validate #20 as it isn't in use effectively (and it's content was overwritten by #21 in tx3)

### DIFF
--- a/src/Voron/Impl/Journal/WriteAheadJournal.cs
+++ b/src/Voron/Impl/Journal/WriteAheadJournal.cs
@@ -234,12 +234,32 @@ namespace Voron.Impl.Journal
                 // corruption when applying journals at recovery time rather than at usage.
                 var tempTx = new TempPagerTransaction();
 
-                foreach (var modifiedPage in modifiedPages)
-                {
-                    var ptr = (PageHeader*)_dataPager.AcquirePagePointerWithOverflowHandling(tempTx, modifiedPage, null);
-                    _env.ValidatePageChecksum(modifiedPage, ptr);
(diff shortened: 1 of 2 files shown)


src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -234,12 +234,32 @@ public bool RecoverDatabase(TransactionHeader* txHeader, Action<string> addToIni
...
+                        var modifiedPage = sortedPages[i];
+
+                        var ptr = (PageHeader*)_dataPager.AcquirePagePointerWithOverflowHandling(tempTx, modifiedPage, null);
+
+                        var maxPageRange = modifiedPage + VirtualPagerLegacyExtensions.GetNumberOfPages(ptr) - 1;
 
-                    tempTx.Dispose(); // release any resources, we just wanted to validate things
+                        if (minPageChecked != -1 && maxPageRange >= minPageChecked)
+                        {
+                            continue;
```
arekpalinski, line 256: I wonder if we should do better and ensure that `modifiedPage` was really modified in earlier transaction (by collecting not only the page numbers by also ids of transactions when a page was modified). 

@ayende ?

ayende: We don't care about that, just that the value of the page is valid.

PR #8320, opened 2018-12-23
RavenDB-12087: String method works on all types, when indexing [WIP DONT MERGE]

garayx:
[WIP DONT MERGE]

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Static/Extensions/DynamicExtensionMethods.cs
+++ b/src/Raven.Server/Documents/Indexes/Static/Extensions/DynamicExtensionMethods.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Globalization;
 using Raven.Client.Documents.Indexes;
 using Sparrow.Json;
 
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Documents/Indexes/Static/Extensions/DynamicExtensionMethods.cs
```diff
@@ -105,5 +106,163 @@ private static bool TryGetString(dynamic o, out string value)
             value = null;
             return false;
         }
+
+        public static string Substring(dynamic o, int value)
```
ayende: use `object`, not `dynamic` here

ppekrol:
1. This is a not good idea to use DynamicExtensionMethods for this, this will rewrite every from that class, no matter if fits (check failing test - you are using IndexOf from List, yet it was rewritten to use the method that you have added - now we need to handle LINQ methods also and all methods from built-in types that match the name of any method from string class). IMO we are asking our self for trouble if we will do it in that class. Or even this way.
2. We will need to add ALL string methods (check 'RavenDB_7953' on how to add a test that checks here) and this will complicate things even further.
3. We will not benefit from any optimizations from LSV for those methods (can probably be ignored because it is not a common case)
4. I still think that due to 1 & 2 we need to introduce classes that will wrap those types and handle methods for string and TimeSpan/DateTime/DateTimeOffset. This way we will also benefit from point 3.

src/Raven.Server/Documents/Indexes/Static/Extensions/DynamicExtensionMethods.cs
garayx: done

garayx:
> 1. This is a not good idea to use DynamicExtensionMethods for this, this will rewrite every from that class, no matter if fits (check failing test - you are using IndexOf from List, yet it was rewritten to use the method that you have added - now we need to handle LINQ methods also and all methods from built-in types that match the name of any method from string class). IMO we are asking our self for trouble if we will do it in that class. Or even this way.
> 2. We will need to add ALL string methods (check 'RavenDB_7953' on how to add a test that checks here) and this will complicate things even further.
> 3. We will not benefit from any optimizations from LSV for those methods (can probably be ignored because it is not a common case)
> 4. I still think that due to 1 & 2 we need to introduce classes that will wrap those types and handle methods for string and TimeSpan/DateTime/DateTimeOffset. This way we will also benefit from point 3.


ok I will change my code.

PR #8321, opened 2018-12-23
Raven db 12277

talweiss1982:


src/Raven.Server/Documents/Queries/Graph/EdgeQueryStep.cs
```diff
@@ -65,9 +65,16 @@ public bool IsEmpty()
             return _results.Count == 0;
         }
 
+        public bool CollectIntermediateResults { get; set; }
+
+        public List<Match> IntermediateResults => CollectIntermediateResults?_results: new List<GraphQueryRunner.Match>();
```
ayende, line 70: I don't like this. It just allocates and throw it away.
Why do we need this to be valid at all if `CollectIntermediateResults` is `false`?

src/Raven.Server/Documents/Queries/Graph/RecursionQueryStep.cs
```diff
@@ -109,9 +109,16 @@ public bool IsEmpty()
             return _results.Count == 0;
         }
 
+        public bool CollectIntermediateResults { get; set; }
+
+        public List<Match> IntermediateResults => CollectIntermediateResults ? _results : new List<Match>();
```
ayende, line 114: In fact, why do we need it to be different from `_results` ?

src/Raven.Server/Documents/Queries/Graph/EdgeQueryStep.cs
talweiss1982: we could return null but I would still want an indication that the intermediate results are needed, see: https://issues.hibernatingrhinos.com/issue/RavenDB-12572

src/Raven.Server/Documents/Queries/Graph/RecursionQueryStep.cs
talweiss1982: collection destination doesn't have _result and we need to collect the requested documents.

PR #8323, opened 2018-12-24
Allow to discard unused memory

ayende:


src/Sparrow/Platform/PlatformDetails.cs
```diff
@@ -50,17 +51,46 @@ private static bool IsWindows8OrNewer()
 
                 var ver = os.Substring(idx + winString.Length);
 
-                if (ver != null)
+                // remove second occurence of '.' (win 10 might be 10.123.456)
+                var index = ver.IndexOf('.', ver.IndexOf('.') + 1);
+                ver = string.Concat(ver.Substring(0, index), ver.Substring(index + 1));
+
+                if (decimal.TryParse(ver, out decimal output))
```
ayende: Here we need to make sure that this is done with InvariantCulture.

src/Sparrow/Platform/PlatformDetails.cs
```diff
@@ -50,17 +51,46 @@ private static bool IsWindows8OrNewer()
...
+                if (idx < 0)
+                    return false;
+
+                var ver = os.Substring(idx + winString.Length);
+
+                // remove second occurence of '.' (win 10 might be 10.123.456)
+                var index = ver.IndexOf('.', ver.IndexOf('.') + 1);
+                ver = string.Concat(ver.Substring(0, index), ver.Substring(index + 1));
+
+                if (decimal.TryParse(ver, out decimal output))
```
ayende: Same, we need invairant

src/Sparrow/Memory.cs
```diff
@@ -272,5 +274,22 @@ public static T Read<T>(byte* ptr)
         {
             return Unsafe.Read<T>(ptr);
         }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static void Discard(void* baseAddress, long size)
```
ayende, line 279: Let's move this to librvnpal

PR #8324, opened 2018-12-24
RavenDB-12556 Assign to `_journalsToDelete` moved from the task that passed to write transaction

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-12556

When the flush operation finishes flushing it try to update journals state under "write transaction lock"
Part of this is to update the `_journalsToDelete` dictionary
If the write transaction lock was taken the flush operation passes the update task to the "write transaction lock" owner.
Meanwhile, the flush operation looks if there is sync task that needs flush lock and runs it
One of the sync tasks that assigned to flush operation include iteration over `_journalsToDelete` dictionary
so that can lead to concurrency issue

### DIFF
--- a/src/Voron/Impl/Journal/WriteAheadJournal.cs
+++ b/src/Voron/Impl/Journal/WriteAheadJournal.cs
@@ -642,8 +642,18 @@ namespace Voron.Impl.Journal
                     _waitForJournalStateUpdateUnderTx.Reset();
                     ExceptionDispatchInfo edi = null;
                     var sp = Stopwatch.StartNew();
+
+                    foreach (var unused in unusedJournals)
+                    {
+                        _journalsToDelete[unused.Number] = unused;
+                    }
+
+                    var singleUseFlag = new SingleUseFlag();
                     Action<LowLevelTransaction> currentAction = txw =>
                     {
+                        if (singleUseFlag.Raise() == false)
+                            throw new InvalidOperationException("Tried to update journal state after flush twice");
+
                         try
                         {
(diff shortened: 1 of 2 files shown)


src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -1048,7 +1053,8 @@ public bool WaitForTaskToBeDone(Func<bool> task)
 
                             if (current.DoneFlag.IsRaised())
                             {
-                                current.Error?.Throw();
+                                if (current.Error != null)
+                                    throw new AggregateException("Unable to complete flush", current.Error);
```
ayende: `AggregateException` is for when you have multiple errors

ayende:
Check the failing tests, please: 
http://jenkins.hibernatingrhinos.com/job/PR_Tests_v4.1/2306/

src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -642,8 +642,18 @@ public void ApplyLogsToDataFile(CancellationToken token, TimeSpan timeToWait)
                     _waitForJournalStateUpdateUnderTx.Reset();
                     ExceptionDispatchInfo edi = null;
                     var sp = Stopwatch.StartNew();
+
+                    foreach (var unused in unusedJournals)
```
grisha-kotler, line 646: On Dispose, this will probably delete the _journalsToDelete before we updated the header files

ayende: This is _bad_

PR #8326, opened 2018-12-24
RavenDB-11871 Journal applicator keep applying journals even if the i…

ayende:
…ndexes didn't change for hours

The issue was that we tried to create journals that are power of 2, but if the transaction size was greater than the size of the journal, we would create a journal file that was bigger than the actual transaction size.
That, in turn, would lead to having available space in the journal file, so we would keep it around after sync.
This change is simply to ensure that the size of the journal file will be either the max journal size or the space the transaction requires. In this case, after a large transaction, we'll no longer have any space there and can clear the large journal file.
As a result, on recovery, we'll not have any journal files to apply and can reduce the speed we take to startup

### DIFF
--- a/src/Voron/Impl/Journal/WriteAheadJournal.cs
+++ b/src/Voron/Impl/Journal/WriteAheadJournal.cs
@@ -123,6 +123,9 @@ namespace Voron.Impl.Journal
             if (_currentJournalFileSize < minRequiredSize)
             {
                 _currentJournalFileSize = Bits.NextPowerOf2(minRequiredSize);
+                if (_currentJournalFileSize > _env.Options.MaxLogFileSize)
+                    _currentJournalFileSize = Math.Max(_env.Options.MaxLogFileSize, minRequiredSize);
+
                 actualLogSize = _currentJournalFileSize;
             }
 
--- a/test/SlowTests/Voron/RavenDB-11871.cs
+++ b/test/SlowTests/Voron/RavenDB-11871.cs
@@ -1,0 +1,68 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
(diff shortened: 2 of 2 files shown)


src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -123,6 +123,9 @@ private JournalFile NextFile(int numberOf4Kbs = 1)
             if (_currentJournalFileSize < minRequiredSize)
             {
                 _currentJournalFileSize = Bits.NextPowerOf2(minRequiredSize);
+                if (_currentJournalFileSize > _env.Options.MaxLogFileSize)
+                    _currentJournalFileSize = Math.Max(_env.Options.MaxLogFileSize, minRequiredSize);
```
talweiss1982, line 127: I think this can be changed to:
_currentJournalFileSize = minRequiredSize;
since in this block:
1. _currentJournalFileSize  > _env.Options.MaxLogFileSize
1. _currentJournalFileSize  < minRequiredSize
thus
minRequiredSize >  _currentJournalFileSize   > _env.Options.MaxLogFileSize => 
minRequiredSize >_env.Options.MaxLogFileSize

ayende: Assume `minRequiredSize` is 33MB in size and `MaxLogFileSize` is 48MB
So `_currentJournalFileSize` is 64MB initially and is larger than `MaxLogFileSize`

We want to set it to the 48MB rather than 33MB in this case, because we want to reduce the number of allocation on the FS.

talweiss1982: You're right

PR #8337, opened 2018-12-25
RavenDB-12407:

maximburyak:
* Fixed finalization order related issues
* Added DisposeOnce behavior allowing to enforce disposing once, but getting the "right" disposal state
https://issues.hibernatingrhinos.com/issue/RavenDB-12407

### DIFF
--- a/src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
+++ b/src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
@@ -1272,15 +1272,7 @@ more responsive application.
         }
 
         ~InMemoryDocumentSessionOperations()
-        {
-            try
-            {
-                Dispose(false);
-            }
-            catch (ObjectDisposedException)
-            {
-                // nothing can be done here
-            }
+        {          
 #if DEBUG
             Debug.WriteLine("Disposing a session for finalizer! It should be disposed by calling session.Dispose()!");
 #endif
--- a/src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
+++ b/src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
@@ -1271,13 +1271,6 @@ more responsive application.
             Dispose(true);
         }
 
-        ~InMemoryDocumentSessionOperations()
-        {          
-#if DEBUG
-            Debug.WriteLine("Disposing a session for finalizer! It should be disposed by calling session.Dispose()!");
-#endif
-        }
-
         public void RegisterMissing(string id)
         {
             if (NoTracking)
--- a/src/Raven.Client/Http/HttpCache.cs
+++ b/src/Raven.Client/Http/HttpCache.cs
@@ -101,17 +101,18 @@ namespace Raven.Client.Http
             {
                 ReleaseRef();
             }
-
+#if !RELEASE
             ~HttpCacheItem()
             {
                 Allocation = null;
-#if !RELEASE
+
                 // Hitting this on DEBUG and/or VALIDATE and getting a higher number than 0 means we have a leak.
                 // On release we will leak, but wont crash. 
                 if (_usages > 0)
                     throw new LowMemoryException("Detected a leak on HttpCache when running the finalizer. See: http://issues.hibernatingrhinos.com/issue/RavenDB-9737");
+
+        }
 #endif
-            }
         }
 
         /// <summary>
--- a/src/Raven.Client/Http/HttpCache.cs
+++ b/src/Raven.Client/Http/HttpCache.cs
@@ -104,6 +104,7 @@ namespace Raven.Client.Http
 
             ~HttpCacheItem()
             {
+                Allocation = null;
 #if !RELEASE
                 // Hitting this on DEBUG and/or VALIDATE and getting a higher number than 0 means we have a leak.
                 // On release we will leak, but wont crash. 
--- a/src/Sparrow/ByteString.cs
+++ b/src/Sparrow/ByteString.cs
@@ -1559,16 +1559,6 @@ namespace Sparrow
 
         private bool _disposed;
 
-        ~ByteStringContext()
-        {
-            _isFinalizerThread = true;
-            _wholeSegments.Clear();
-            _internalReadyToUseMemorySegments.Clear();
-        }
-
-
-
-
         public void Dispose()
         {
             lock (this)
--- a/src/Sparrow/ByteString.cs
+++ b/src/Sparrow/ByteString.cs
@@ -433,7 +433,10 @@ namespace Sparrow
         {
             try
             {
-                Dispose();
+                if (Segment == null)
+                    return;
+                NativeMemory.Free(Segment, Size, _thread);
+                Segment = null;
             }
             catch (ObjectDisposedException)
             {
@@ -490,11 +493,11 @@ namespace Sparrow
 
     public struct ByteStringMemoryCache : IByteStringAllocator
     {
-        private static readonly ThreadLocal<SegmentStack> SegmentsPool;
+        private static readonly ThreadLocal<StackHeader<UnmanagedGlobalSegment>> SegmentsPool;
         private static readonly SharedMultipleUseFlag LowMemoryFlag;
         private static readonly LowMemoryHandler LowMemoryHandlerInstance = new LowMemoryHandler();
 
-        public static readonly NativeMemoryCleaner<SegmentStack, UnmanagedGlobalSegment> Cleaner;
+        public static readonly NativeMemoryCleaner<StackHeader<UnmanagedGlobalSegment>, UnmanagedGlobalSegment> Cleaner;
 
         private class LowMemoryHandler : ILowMemoryHandler
         {
@@ -512,9 +515,9 @@ namespace Sparrow
 
         static ByteStringMemoryCache()
         {
-            SegmentsPool = new ThreadLocal<SegmentStack>(() => new SegmentStack(), trackAllValues: true);
+            SegmentsPool = new ThreadLocal<StackHeader<UnmanagedGlobalSegment>>(() => new StackHeader<UnmanagedGlobalSegment>(), trackAllValues: true);
             LowMemoryFlag = new SharedMultipleUseFlag();
-            Cleaner = new NativeMemoryCleaner<SegmentStack, UnmanagedGlobalSegment>(() => SegmentsPool.Values, LowMemoryFlag, TimeSpan.FromMinutes(1), TimeSpan.FromMinutes(1));
+            Cleaner = new NativeMemoryCleaner<StackHeader<UnmanagedGlobalSegment>, UnmanagedGlobalSegment>(() => SegmentsPool.Values, LowMemoryFlag, TimeSpan.FromMinutes(1), TimeSpan.FromMinutes(1));
 
             ThreadLocalCleanup.ReleaseThreadLocalState += CleanForCurrentThread;
 
@@ -612,35 +615,7 @@ namespace Sparrow
                 current.Value?.Dispose();
                 current = current.Next;
             }
-        }
-      
-        public class SegmentStack : StackHeader<UnmanagedGlobalSegment>
-        {
-            ~SegmentStack()
-            {
-                if (Environment.HasShutdownStarted)
-                    return; // no need
-
-                var current = Interlocked.Exchange(ref Head, HeaderDisposed);
-                try
-                {
-                    while (current != null)
-                    {
-                        var segment = current.Value;
-                        current = current.Next;
-                        if (segment == null)
-                            continue;
-                        if (!segment.InUse.Raise())
-                            continue;
-                        segment.Dispose();
-                    }
-                }
-                catch (ObjectDisposedException)
-                {
-                    // in case that in the future we will throw that exception
-                }
-            }
-        }
+        }       
     }
 
     public sealed class ByteStringContext : ByteStringContext<ByteStringMemoryCache>
@@ -755,7 +730,12 @@ namespace Sparrow
         public void Reset()
         {
             if (_disposed)
+            {
+                if (_isFinalizerThread)
+                    return;
                 ThrowObjectDisposed();
+            }
+                
 
             Array.Clear(_internalReusableStringPoolCount, 0, _internalReusableStringPoolCount.Length);
             foreach (var stack in _internalReusableStringPool)
@@ -1582,16 +1562,12 @@ namespace Sparrow
         ~ByteStringContext()
         {
             _isFinalizerThread = true;
-            try
-            {
-                Dispose();
-            }
-            catch (ObjectDisposedException)
-            {
-                // This is expected, we might be calling the finalizer on an object that
-                // was already disposed, we don't want to error here because of this
-            }
+            _wholeSegments.Clear();
+            _internalReadyToUseMemorySegments.Clear();
         }
+
+
+
 
         public void Dispose()
         {
@@ -1603,6 +1579,7 @@ namespace Sparrow
                 GC.SuppressFinalize(this);
 
                 _disposed = true;
+
 
                 foreach (var segment in _wholeSegments)
                 {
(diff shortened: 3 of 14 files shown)


src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -1272,15 +1272,7 @@ public virtual void Dispose()
...
-        {
-            try
-            {
-                Dispose(false);
-            }
-            catch (ObjectDisposedException)
-            {
-                // nothing can be done here
-            }
+        {
```
ayende: Need to change the `#if DEBUG` for the entire finalizer.

src/Raven.Client/Http/HttpCache.cs
```diff
@@ -104,6 +104,7 @@ public void Dispose()
 
             ~HttpCacheItem()
             {
+                Allocation = null;
 #if !RELEASE
```
ayende: Need to change the `#if` to include the whole finalizer. 
We don't need it otherwise

src/Sparrow/ByteString.cs
```diff
@@ -1582,17 +1562,13 @@ private void ValidateAndUnregister(ByteStringStorage* value)
...
-            try
-            {
-                Dispose();
-            }
-            catch (ObjectDisposedException)
-            {
-                // This is expected, we might be calling the finalizer on an object that
-                // was already disposed, we don't want to error here because of this
-            }
+            _wholeSegments.Clear();
```
ayende: Why do you need this?
Can we avoid having a finalizer here entirely?

src/Sparrow/Json/ArenaMemoryAllocator.cs
```diff
@@ -308,11 +309,18 @@ public override string ToString()
         }
 
         public void Dispose()
+        {
+            Dispose(true);
+        }
+        public void Dispose(bool disposing)
```
ayende, line 315: Probably not required. The only thing you do differently is the lock on the `this`, and we _know_ that can't happen here.

src/Sparrow/Json/ArenaMemoryAllocator.cs
```diff
@@ -330,19 +338,27 @@ public void Dispose()
...
+                    Monitor.Exit(this);
+            }            
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public void Return(AllocatedMemoryData allocation)
         {
             if (_isDisposed)
                 return;
+            GC.SuppressFinalize(allocation);
```
ayende: Why `AllocatedMemoryData ` have a finalizer?

src/Sparrow/Json/BlittableJsonDocumentBuilder.cs
```diff
@@ -170,6 +170,8 @@ public void Dispose()
 
             _writer.Dispose();
             GlobalCache.Free(_cacheItem);
+            GC.SuppressFinalize(this);
```
ayende: Does `BlittableJsonDocumentBuilder` need a finalizer?

src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -311,12 +311,12 @@ public static ManagedPinnedBuffer ShortLivedInstance()
...
         {
             return new JsonOperationContext(4096, 1024, SharedMultipleUseFlag.None);
-        }
+        }       
 
         public JsonOperationContext(int initialSize, int longLivedSize, SharedMultipleUseFlag lowMemoryFlag)
         {
             Debug.Assert(lowMemoryFlag != null);
-            _disposeOnceRunner = new DisposeOnce<ExceptionRetry>(() =>
+            _disposeOnceRunner = new DisposeOnce<SingleAttemptWithWaitForDisposeToFinish>(() =>
```
ayende: Let's just change the behavior of `SingleAttempt` to allow recursive calls

src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -423,6 +423,15 @@ public void Dispose()
                 _parent._managedBuffers.Push(_buffer);
                 _buffer = null;
             }
+            
+            public void Kill()
+            {
+                if (_buffer == null || _parent.Disposed)
+                    return;
+
+                _parent?._managedBuffers?.Push(_buffer);
```
ayende: This should not be called during the dispose, I think

src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -1004,7 +1016,7 @@ protected internal virtual unsafe void Reset(bool forceReleaseLongLivedAllocator
                 {
                     _arenaAllocatorForLongLivedValues.Return(mem.AllocatedMemoryData);
                     mem.AllocatedMemoryData = null;
-                    mem.Dispose();
+                    mem.Dispose(); // todo: probably we don't need this call
```
ayende: why not?

src/Sparrow/Json/UnmanagedWriteBuffer.cs
```diff
@@ -15,6 +15,7 @@ public unsafe interface IUnmanagedWriteBuffer : IDisposable
         void WriteByte(byte data);
         void EnsureSingleChunk(JsonParserState state);
         void EnsureSingleChunk(out byte* ptr, out int size);
+        void Kill();
```
ayende, line 18: I don't think this should be required

src/Sparrow/Threading/DisposeOnce.cs
```diff
@@ -7,6 +7,7 @@ namespace Sparrow.Threading
     public interface IDisposeOnceOperationMode {}
     public struct ExceptionRetry : IDisposeOnceOperationMode { }
     public struct SingleAttempt : IDisposeOnceOperationMode { }
+    public struct SingleAttemptWithWaitForDisposeToFinish : IDisposeOnceOperationMode { }
```
ayende: I don't like this, we should add a `disposing` flag for when we run the actual disposing

ayende (changes requested):

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
maximburyak: done

src/Raven.Client/Http/HttpCache.cs
maximburyak: done

src/Sparrow/ByteString.cs
maximburyak: yes, the cost of fnialization is greater then the benefit of clearing

src/Sparrow/Json/ArenaMemoryAllocator.cs
maximburyak: Because it holds buffers and it's main unmamanged memory that it should release (look at the Dispose code)

maximburyak: It doesn't have one, removing this

src/Sparrow/Json/ArenaMemoryAllocator.cs
maximburyak: what can't happen there? the locking of this?

src/Sparrow/Json/BlittableJsonDocumentBuilder.cs
maximburyak: it doesn't have one, removing this

src/Sparrow/Json/ArenaMemoryAllocator.cs
ayende: Yes, if we are in the finalizer, we know no one is locking on `this`

src/Sparrow/Json/JsonOperationContext.cs
maximburyak: it's not a dispose call, it's a resources release code (structs don't have finalizers..)

ayende: It is only _called_ from finalizer. That shouldn't be the case.
You are `guaranteed` that this is not a valid thing to do there.

src/Sparrow/Json/JsonOperationContext.cs
maximburyak: in the dispose function we do:
1. throw exception if already disposed (we don't really care about it here)
2. return memory to the context (we just did it)
3. raise the IsDisposed flag for the LSV (actually, this one is the more important part..)

What we really want to do is to raise the IsDisposed flag from outside (I'll do just that)

src/Sparrow/Json/UnmanagedWriteBuffer.cs
maximburyak: We need it for the UnmanagedStreamBuffer.Kill, which should destroy it's ManagedPinnedBuffer

src/Sparrow/Json/JsonOperationContext.cs
maximburyak: I'm changing it so it will just call the _buffer.Dispose, in order to prevent concurrency issues with the disposal of the context

ayende: No need, `ManagedPinnedBuffer` already have a finalizer

src/Sparrow/Json/JsonOperationContext.cs
maximburyak: done

src/Sparrow/Threading/DisposeOnce.cs
maximburyak: done

src/Sparrow/ByteString.cs
```diff
@@ -755,7 +730,12 @@ public ByteStringContext(SharedMultipleUseFlag lowMemoryFlag, int allocationBloc
         public void Reset()
         {
             if (_disposed)
+            {
+                if (_isFinalizerThread)
```
ayende, line 734: Can this now _ever_ be called from finalizer?

src/Sparrow/Json/BlittableWriter.cs
```diff
@@ -252,6 +252,12 @@ static BlittableWriter()
             ThreadLocalCleanup.ReleaseThreadLocalState += CleanPropertyArrayOffset;
         }
 
+        ~BlittableWriter()
+        {
+            _unmanagedWriteBuffer.Kill();
```
ayende, line 257: Still pretty sure we don't need this

src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -312,6 +296,8 @@ public void Dispose()
                 if (Parent == null)
                     return;// disposed already
 
+                GC.SuppressFinalize(this);
```
ayende, line 299: Probably no longer required

src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -423,6 +423,14 @@ public void Dispose()
                 _parent._managedBuffers.Push(_buffer);
                 _buffer = null;
             }
+            
+            public void Kill()
+            {
+                if (_buffer == null)
+                    return;
+                _buffer.Dispose();
```
ayende, line 431: Not required

src/Sparrow/Json/Parsing/ObjectJsonParser.cs
```diff
@@ -203,6 +203,8 @@ public void Dispose()
             _disposed = true;
             if (_currentStateBuffer != null)
                 _ctx.ReturnMemory(_currentStateBuffer);
+
+            GC.SuppressFinalize(this);
```
ayende, line 207: Do we even need a finalizer here?

src/Sparrow/Json/UnmanagedBuffersPool.cs
```diff
@@ -185,9 +185,10 @@ public static int GetIndexFromSize(int size)
 
         public void Return(AllocatedMemoryData returned)
         {
+            GC.SuppressFinalize(returned);
```
ayende, line 188: Not required

src/Sparrow/Json/UnmanagedWriteBuffer.cs
ayende: That is NOT safe to do. The `ManagedPinnedBuffer` has its own finalizer

src/Sparrow/Threading/DisposeOnce.cs
```diff
@@ -102,10 +144,10 @@ public bool Disposed
                 if (state.Item1 == false)
                     return false;
 
-                if (typeof(TOperationMode) == typeof(SingleAttempt))
+                if (typeof(TOperationMode) == typeof(SingleAttempt) && _operationModeData.DuringDispose == false)
                     return true;
 
-                if (typeof(TOperationMode) == typeof(ExceptionRetry))
+                if (typeof(TOperationMode) == typeof(ExceptionRetry) || typeof(TOperationMode) == typeof(SingleAttempt))
```
ayende, line 150: Is this required?

src/Sparrow/Threading/DisposeOnce.cs
```diff
@@ -102,10 +144,10 @@ public bool Disposed
                 if (state.Item1 == false)
                     return false;
 
-                if (typeof(TOperationMode) == typeof(SingleAttempt))
+                if (typeof(TOperationMode) == typeof(SingleAttempt) && _operationModeData.DuringDispose == false)
                     return true;
```
ayende, line 148: Better to change to: `return  _operationModeData.DuringDispose == false;`, no?

src/Sparrow/Threading/DisposeOnce.cs
```diff
@@ -214,10 +267,10 @@ public bool Disposed
                 if (state.Item1 == false)
                     return false;
 
-                if (typeof(TOperationMode) == typeof(SingleAttempt))
+                if (typeof(TOperationMode) == typeof(SingleAttempt) && _operationModeData.DuringDispose == false)
```
ayende, line 270: Same as above

ayende (changes requested):

src/Sparrow/Json/ArenaMemoryAllocator.cs
maximburyak: that's right, so we don't want to try to enter lock at all

src/Sparrow/Json/JsonOperationContext.cs
maximburyak: you are right

src/Sparrow/Json/UnmanagedWriteBuffer.cs
maximburyak: ok

src/Sparrow/ByteString.cs
maximburyak: no, we can remove _isFinalizerThread entirely

src/Sparrow/Json/BlittableWriter.cs
maximburyak: right

src/Sparrow/Json/JsonContextPoolBase.cs
maximburyak: done

src/Sparrow/Json/JsonOperationContext.cs
maximburyak: done

src/Sparrow/Json/Parsing/ObjectJsonParser.cs
maximburyak: done

src/Sparrow/Json/UnmanagedBuffersPool.cs
maximburyak: done

src/Sparrow/Threading/DisposeOnce.cs
maximburyak: no(relates to comment below)

src/Sparrow/Threading/DisposeOnce.cs
maximburyak: right

src/Sparrow/Threading/DisposeOnce.cs
maximburyak: yes

src/Sparrow/Threading/DisposeOnce.cs
ayende: I don't understand?

PR #8340, opened 2018-12-26
RavenDB-12374 Changes API fail over should take into account the data…

efratshenhar:
…base topology and fail to another node

### DIFF
--- a/src/Raven.Client/Documents/Changes/DatabaseChanges.cs
+++ b/src/Raven.Client/Documents/Changes/DatabaseChanges.cs
@@ -560,7 +560,7 @@ namespace Raven.Client.Documents.Changes
 
                             try
                             {
-                                if ((json.TryGet("TopologyChange", out string change)) && (string.Equals(change,"true", StringComparison.OrdinalIgnoreCase)))
+                                if ((json.TryGet(nameof(TopologyChange), out bool supports)) && (supports))
                                 {
                                     GetOrAddConnectionState("Topology", "watch-topology-change", "", "");
                                     await _requestExecutor.UpdateTopologyAsync(_serverNode, 0, true).ConfigureAwait(false);
--- a/src/Raven.Client/Documents/Changes/DatabaseChanges.cs
+++ b/src/Raven.Client/Documents/Changes/DatabaseChanges.cs
@@ -477,7 +477,7 @@ namespace Raven.Client.Documents.Changes
                     {
                         ConnectionStatusChanged?.Invoke(this, EventArgs.Empty);
 
-                        _serverNode = await _requestExecutor.HandleServerNotResponsive(_url.AbsoluteUri, _serverNode, _nodeIndex, e).ConfigureAwait(false);
+                        _serverNode = await _requestExecutor.HandleServerNotResponsive(_url.AbsoluteUri, _serverNode, _nodeIndex, e, _cts.Token).ConfigureAwait(false);
 
                         if (ReconnectClient() == false)
                             return;
@@ -638,7 +638,7 @@ namespace Raven.Client.Documents.Changes
                     {
                         Url = topologyChange.Url,
                         Database = topologyChange.Database
-                    }, 0, true);
+                    }, 0, true).ConfigureAwait(false);
                     break;
                 default:
                     throw new NotSupportedException(type);
--- a/src/Raven.Client/Documents/Changes/DatabaseChanges.cs
+++ b/src/Raven.Client/Documents/Changes/DatabaseChanges.cs
@@ -319,7 +319,15 @@ namespace Raven.Client.Documents.Changes
 
             _counters.Clear();
 
-            _task.Wait();
+            try
+            {
+                _task.Wait();
+            }
+            catch
+            {
+                // we're disposing the document store
+                // nothing we can do here
+            }
 
             ConnectionStatusChanged?.Invoke(this, EventArgs.Empty);
             ConnectionStatusChanged -= OnConnectionStatusChanged;
@@ -477,7 +485,7 @@ namespace Raven.Client.Documents.Changes
                     {
                         ConnectionStatusChanged?.Invoke(this, EventArgs.Empty);
 
-                        _serverNode = await _requestExecutor.HandleServerNotResponsive(_url.AbsoluteUri, _serverNode, _nodeIndex, e, _cts.Token).ConfigureAwait(false);
+                        _serverNode = await _requestExecutor.HandleServerNotResponsive(_url.AbsoluteUri, _serverNode, _nodeIndex, e).ConfigureAwait(false);
 
                         if (ReconnectClient() == false)
                             return;
--- a/src/Raven.Client/Documents/Changes/DatabaseChanges.cs
+++ b/src/Raven.Client/Documents/Changes/DatabaseChanges.cs
@@ -40,6 +40,10 @@ namespace Raven.Client.Documents.Changes
         private readonly AtomicDictionary<DatabaseConnectionState> _counters = new AtomicDictionary<DatabaseConnectionState>(StringComparer.OrdinalIgnoreCase);
         private int _immediateConnection;
 
+        private ServerNode _serverNode;
+        private int _nodeIndex;
+        private Uri _url;
+
         public DatabaseChanges(RequestExecutor requestExecutor, string databaseName, Action onDispose)
         {
             _requestExecutor = requestExecutor;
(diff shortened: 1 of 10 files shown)


src/Raven.Client/Documents/Changes/DatabaseChanges.cs
```diff
@@ -431,17 +435,17 @@ private async Task DoWork()
...
-
             while (_cts.IsCancellationRequested == false)
             {
                 try
                 {
                     if (Connected == false)
                     {
-                        await _client.ConnectAsync(url, _cts.Token).ConfigureAwait(false);
+                        _url = new Uri($"{_serverNode.Url}/databases/{_database}/changes"
+                            .ToLower()
```
ayende, line 453: why `ToLower`?

src/Raven.Client/Documents/Changes/DatabaseChanges.cs
```diff
@@ -553,6 +560,12 @@ private async Task ProcessChanges()
 
                             try
                             {
+                                if (json.TryGet("EnableTopologyChangeNotification", out string ver))
```
ayende: Let's also validate that the value is `true`

src/Raven.Client/Documents/Changes/DatabaseChanges.cs
```diff
@@ -466,12 +470,15 @@ private async Task DoWork()
...
                 catch (Exception e)
                 {
+
                     //We don't report this error since we can automatically recover from it and we can't
                     // recover from the OnError accessing the faulty WebSocket.
                     try
                     {
                         ConnectionStatusChanged?.Invoke(this, EventArgs.Empty);
 
+                        _serverNode = await _requestExecutor.HandleServerDown(_url.AbsoluteUri, _serverNode, _nodeIndex, e).ConfigureAwait(false);
```
ayende: I think you should move this to be around `_client.ConnectAsync`, instead.
This can happen because the connection just broke, and we don't want to report server down for that.

src/Raven.Server/Documents/ChangesClientConnection.cs
```diff
@@ -505,6 +536,18 @@ public void Confirm(int commandId)
             });
         }
 
+        public void EnableTopologyChangeNotification()
+        {
+            _sendQueue.Enqueue(new ChangeValue
+            {
+                ValueToSend = new DynamicJsonValue
+                {
+                    ["EnableTopologyChangeNotification"] = ""
```
ayende: Set to `true`

src/Raven.Server/Documents/ChangesClientConnection.cs
```diff
@@ -72,6 +74,11 @@ public ChangesClientConnection(WebSocket webSocket, DocumentDatabase documentDat
 
         public TimeSpan Age => SystemTime.UtcNow - _startedAt;
 
+        public void WatchTopology()
+        {
+            _watchTopology = true;
+        }
```
ayende, line 80: Also send the current topology here

test/SlowTests/Cluster/ClusterOperationTests.cs
```diff
@@ -90,6 +97,145 @@ public async Task ClusterWideIdentity()
             }
         }
 
+        [Fact]
+        public async Task ChangesApiFailOver()
+        {
+            DebuggerAttachedTimeout.DisableLongTimespan = true;
```
ayende: Doesn't this have global impact?

test/SlowTests/Cluster/ClusterOperationTests.cs
```diff
@@ -90,6 +97,145 @@ public async Task ClusterWideIdentity()
...
+                }
+
+                Assert.Equal(list.Count, 3);
+            }
+        }
+
+        [Fact]
+        public async Task ChangesApiReorderDatabaseNodes()
+        {
+            DebuggerAttachedTimeout.DisableLongTimespan = true;
```
ayende: Doesn't this have global impact?

test/SlowTests/Cluster/ClusterOperationTests.cs
```diff
@@ -90,6 +97,145 @@ public async Task ClusterWideIdentity()
...
+                    {
+                        toDispose = server;
+                        break;
+                    }
+                }
+
+                DisposeServerAndWaitForFinishOfDisposal(toDispose);
+                await taskObservable.EnsureConnectedNow();
+
+                await Task.Delay(12000);
```
ayende: Can we use something other than `Delay` here?

test/SlowTests/Cluster/ClusterOperationTests.cs
```diff
@@ -90,6 +97,145 @@ public async Task ClusterWideIdentity()
...
+                    await session.SaveChangesAsync();
+                }
+
+                await Task.Delay(100);
+                Assert.Equal(list.Count, 1);
+
+                string url1 = store.GetRequestExecutor().Url;
+                await ReverseOrderSuccessfully(store, db);
+
+                await Task.Delay(500);
```
ayende: Can we use something other than `Delay` here?

ayende (changes requested):

src/Raven.Client/Documents/Changes/DatabaseChanges.cs
```diff
@@ -553,6 +560,12 @@ private async Task ProcessChanges()
 
                             try
                             {
+                                if (json.TryGet("EnableTopologyChangeNotification", out string ver))
+                                {
+                                    GetOrAddConnectionState("Topology", "watch-topology-change", "", "");
```
ppekrol, line 573: Wouldn't it be easier just to use RequestExecutor.TopologyUpdated event, in worst case this happens every 5 minutes, but when there are requests to that database, then it happens just after first request that happened after topology changed

ayende: No, the problem is that if we don't get topology updates, in a mostly idle app that watches `Changes()`, we won't know about this soon enough

src/Raven.Server/Documents/Handlers/ChangesHandler.cs
```diff
@@ -111,6 +111,8 @@ private async Task HandleConnection(WebSocket webSocket, JsonOperationContext co
                     var jsonParserState = new JsonParserState();
                     using (var parser = new UnmanagedJsonParser(context, jsonParserState, debugTag))
                     {
+                        connection.EnableTopologyChangeNotification();
```
ppekrol: Can we rename that to 'SendSupportedFeatures' and have a single notification for supported features (more future proof)

```json
{
   "TopologyChange" : true
}
```

ayende: I like that, yes.

src/Raven.Client/Documents/Changes/DatabaseChanges.cs
ppekrol: Ok. Can we double check that this will not brake old clients (I think not because it lacks 'Type' property)?

ayende: Yes, it intentionally doesn't have `Type` properly.
The C# client is safe, need to check the others, though

src/Raven.Server/Documents/ChangesClientConnection.cs
```diff
@@ -536,13 +536,13 @@ public void Confirm(int commandId)
...
 
-        public void EnableTopologyChangeNotification()
+        public void SendSupportedFeatures()
         {
             _sendQueue.Enqueue(new ChangeValue
             {
                 ValueToSend = new DynamicJsonValue
                 {
-                    ["EnableTopologyChangeNotification"] = ""
+                    ["TopologyChange"] = "true"
```
ppekrol: just `true` not `"true"`

then you can do `json.TryGet(nameof(TopologyChange), out bool supports)`

src/Raven.Client/Documents/Changes/DatabaseChanges.cs
```diff
@@ -560,9 +560,10 @@ private async Task ProcessChanges()
 
                             try
                             {
-                                if (json.TryGet("EnableTopologyChangeNotification", out string ver))
+                                if ((json.TryGet("TopologyChange", out string change)) && (string.Equals(change,"true", StringComparison.OrdinalIgnoreCase)))
```
ppekrol: Use `nameof(TopologyChange)` and the `.TryGet(bool)`

ppekrol (approved):

ayende:
Please check the failing tests. It looks like aggressive caching is somehow broken

efratshenhar:
Ignore the last fix, tests still failing

src/Raven.Client/Documents/Changes/DatabaseChanges.cs
ml054: It broke java client for sure, see: https://issues.hibernatingrhinos.com/issue/RDBC-268

ayende: And very likely the rests as well, I would assume

PR #8351, opened 2018-12-27
fix for RavenDB-12591

myarichuk:
1) Make sure not to replace QueryQueryStep during query optimization if it has where clause (its replacement step ignores where clause).
2) fix a test with invalid assert (the test itself with invalid and was passing because of the query results were wrong due to https://issues.hibernatingrhinos.com/issue/RavenDB-12567)

### DIFF
--- a/src/Raven.Server/Documents/Queries/Graph/QueryQueryStep.cs
+++ b/src/Raven.Server/Documents/Queries/Graph/QueryQueryStep.cs
@@ -62,6 +62,9 @@ namespace Raven.Server.Documents.Queries.Graph
         {
             get
             {
+                if (HasWhereClause) //TODO: verify with Tal how good is this change
+                    return false;
+
                 if (_queryMetadata.IsCollectionQuery == false)
                     return false;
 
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Documents/Queries/Graph/QueryQueryStep.cs
```diff
@@ -62,6 +62,9 @@ public bool CanBeConsideredForDestinationOptimization
         {
             get
             {
+                if (HasWhereClause) //TODO: verify with Tal how good is this change
```
ayende, line 65: This is wrong, because collection queries should work, the problem is probably that we aren't filtering the ids.

talweiss1982: In the original code we used HasWhereClause and indeed it is needed for id queries since they are collection queries and have a where clause.
So the quick fix is to add HasWhereClause to the other conditions.
If we want to support id queries than we need to support it in the IsMatchedBy method and see if the asked id is matched by the where clause in the CollectionDestinationStep

PR #8355, opened 2018-12-28
RavenDB-12259 projections should not be registered in subscription session

ppekrol:


ppekrol:
Does not need to be applied to 4.0 because we do not register subscription results in opened session there

ayende (approved):

PR #8363, opened 2018-12-30
RavenDB-12264 [WIP][Don't Merge] identical query steps optimization

talweiss1982:
The purpose of this PR is to surface a few issues I ran into implementing this.
This code doesn't even work and needs a bit more debugging making it work completely
(The remaining issue is with the matches aliases not working as expected)
see TODOs in the PR, my main issue is that we need to clone and mutate Matches so they work nicely with the rest of the code, I have been struggling with a couple of approaches and decided its better to surface things up and maybe come up with a cleaner solution.

### DIFF
--- a/src/Raven.Server/Documents/Queries/Graph/CollectionDestinationQueryStep.cs
+++ b/src/Raven.Server/Documents/Queries/Graph/CollectionDestinationQueryStep.cs
@@ -109,5 +109,10 @@ namespace Raven.Server.Documents.Queries.Graph
         {
             throw new NotSupportedException();
         }
+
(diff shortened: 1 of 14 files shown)


talweiss1982:
I think that for this to work we will need to clone the result set :(
i'm currently looking at failing tests that don't include recursive

ayende (approved):

talweiss1982:
Don't merge this it is all wrong

talweiss1982:
even cloning the result set won't fix the problem since the GetSingleGraphStepExecution

PR #8365, opened 2018-12-31
fix for RavenDB-12542

myarichuk:
In graph queries, support vertices that query indexes without being defined in explicit WITH clause.
For example, the following query should properly work:
```match (from index 'Orders/Totals')-[Lines where PricePerUnit > 200 select Product]->(from index 'Product/Search')```

Under the hood, in this case the query parser will create WITH clauses that match the "from index" statements inside the vertex definitions


myarichuk:
Will close and reopen with proper changes (query syntax needs to be adjusted)

PR #8370, opened 2019-01-01
RavenDB-12264 caching query results

talweiss1982:


talweiss1982:
This is just a first draft, after trying different ways to implement identical steps merging I realize that getting it right is too complicated and defeats the purpose.
As I was struggling with just QueryQueryStep and there is no generic solution in sight without major refactoring my current thought is that it is just not worth it at this stage and for QueryQueryStep the suggested solution will work just fine.
@ayende what do you think?

src/Raven.Server/Documents/Queries/Graph/QueryQueryStep.cs
```diff
@@ -114,13 +114,20 @@ public ValueTask Initialize()
         {
             if (_index != -1)
                 return default;
-
-            var results = _queryRunner.ExecuteQuery(new IndexQueryServerSide(_queryMetadata)
+            //TODO: we can cache only queries we know has identical steps to save memory 
+            //TODO: we can also add ref count per step and clear the results once they are populated
+            var cache = _graphQueryPlan.QueryCache;
+            var key = _queryMetadata.Query.ToString();
```
ayende: I don't particularly like this, because of the allocations. But we can live with this for now

ayende (approved):

src/Raven.Server/Documents/Queries/Graph/QueryQueryStep.cs
talweiss1982: We can save the allocation if we implement equality for QueryExpression and its descendants but there are other places using so much more managed memory in graph query that I feel that it is pointless and better handled in v4.5 where we should really think more about the memory footprint of graph queries.

ayende: At this point, this isn't worth it, I agree. We can wait for the re-impl

PR #8374, opened 2019-01-01
RavenDB-4771: Apply Topology-Id header validation fix to attachments,…

maximburyak:
… as it's done with documents


talweiss1982:
Did we swallow the error somewhere? 
We should have at least seen the error on fiddler, are they running under cluster (i don't recall they said so)?

ayende:
No, it was in the fiddler output

PR #8376, opened 2019-01-01
RavenDB-12488 - Display the real path for all the nodes when creating a database

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-12488

### DIFF
--- a/src/Raven.Server/Web/Studio/DataDirectoryInfo.cs
+++ b/src/Raven.Server/Web/Studio/DataDirectoryInfo.cs
@@ -46,7 +46,8 @@ namespace Raven.Server.Web.Studio
             {
                 NodeTag = _serverStore.NodeTag,
                 FullPath = realPath,
-                TotalFreeSpaceHumane = diskSpaceInfo?.TotalFreeSpace.ToString()
+                TotalFreeSpaceHumane = diskSpaceInfo?.TotalFreeSpace.ToString(),
+                TotalSpaceHumane = diskSpaceInfo?.TotalSize.ToString()
             };
 
             if (_getNodesInfo == false)
@@ -179,6 +180,8 @@ namespace Raven.Server.Web.Studio
 
         public string FullPath { get; set; }
 
+        public string TotalSpaceHumane { get; set; }
+
         public string TotalFreeSpaceHumane { get; set; }
 
         public DynamicJsonValue ToJson()
@@ -187,7 +190,8 @@ namespace Raven.Server.Web.Studio
             {
                 [nameof(NodeTag)] = NodeTag,
                 [nameof(FullPath)] = FullPath,
-                [nameof(TotalFreeSpaceHumane)] = TotalFreeSpaceHumane
+                [nameof(TotalFreeSpaceHumane)] = TotalFreeSpaceHumane,
+                [nameof(TotalSpaceHumane)] = TotalSpaceHumane
             };
         }
     }
--- a/src/Raven.Server/Web/Studio/DataDirectoryInfo.cs
+++ b/src/Raven.Server/Web/Studio/DataDirectoryInfo.cs
@@ -5,7 +5,6 @@ using System.Linq;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
-using Raven.Client.Documents.Conventions;
 using Raven.Client.Http;
 using Raven.Server.Json;
 using Raven.Server.ServerWide;
@@ -26,20 +25,26 @@ namespace Raven.Server.Web.Studio
 
         private readonly ServerStore _serverStore;
         private readonly string _path;
+        private readonly string _name;
+        private readonly bool _isBackup;
         private readonly bool _getNodesInfo;
         private readonly int _requestTimeoutInMs;
         private readonly Stream _responseBodyStream;
 
-        public DataDirectoryInfo(ServerStore serverStore, string path, bool getNodesInfo, int requestTimeoutInMs, Stream responseBodyStream)
+        public DataDirectoryInfo(
+            ServerStore serverStore, string path, string name, bool isBackup,
+            bool getNodesInfo, int requestTimeoutInMs, Stream responseBodyStream)
         {
             _serverStore = serverStore;
             _path = path;
+            _name = name;
+            _isBackup = isBackup;
             _getNodesInfo = getNodesInfo;
             _requestTimeoutInMs = requestTimeoutInMs;
             _responseBodyStream = responseBodyStream;
         }
 
-        public async Task UpdateDirectoryResult(string urlPath, string databaseName)
+        public async Task UpdateDirectoryResult(string databaseName)
         {
             var drivesInfo = PlatformDetails.RunningOnPosix ? DriveInfo.GetDrives() : null;
             var driveInfo = DiskSpaceChecker.GetDriveInfo(_path, drivesInfo, out var realPath);
@@ -75,7 +80,7 @@ namespace Raven.Server.Web.Studio
 
             if (relevantNodes.Count > 1)
             {
-                await UpdateNodesDirectoryResult(relevantNodes, clusterTopology, urlPath, dataDirectoryResult);
+                await UpdateNodesDirectoryResult(relevantNodes, clusterTopology, dataDirectoryResult);
             }
 
             using (_serverStore.ContextPool.AllocateOperationContext(out JsonOperationContext context))
@@ -105,11 +110,8 @@ namespace Raven.Server.Web.Studio
             }
         }
 
-        private async Task UpdateNodesDirectoryResult(
-            IEnumerable<string> nodes, ClusterTopology clusterTopology,
-            string urlPath, DataDirectoryResult dataDirectoryResult)
+        private async Task UpdateNodesDirectoryResult(IEnumerable<string> nodes, ClusterTopology clusterTopology, DataDirectoryResult dataDirectoryResult)
         {
-            var httpClient = RequestExecutor.CreateHttpClient(_serverStore.Server.Certificate.Certificate, DocumentConventions.Default);
             var tasks = new List<Task<SingleNodeDataDirectoryResult>>();
 
             foreach (var nodeTag in nodes)
@@ -125,8 +127,7 @@ namespace Raven.Server.Web.Studio
 
                 tasks.Add(Task.Run(async () =>
                 {
-                    var url = $"{serverUrl}/{urlPath}";
-                    var singleNodeResult = await GetSingleNodeDataDirectoryInfo(url, httpClient);
+                    var singleNodeResult = await GetSingleNodeDataDirectoryInfo(serverUrl);
                     return singleNodeResult;
                 }, _serverStore.ServerShutdown));
             }
@@ -146,22 +147,19 @@ namespace Raven.Server.Web.Studio
             }
         }
 
-        private async Task<SingleNodeDataDirectoryResult> GetSingleNodeDataDirectoryInfo(string url, HttpClient httpClient)
+        private async Task<SingleNodeDataDirectoryResult> GetSingleNodeDataDirectoryInfo(string serverUrl)
         {
             using (var cts = new CancellationTokenSource(_requestTimeoutInMs))
             {
                 try
                 {
-                    var request = new HttpRequestMessage(HttpMethod.Get, url);
-                    var response = await httpClient.SendAsync(request, cts.Token);
-                    if (response.IsSuccessStatusCode == false)
-                        return null;
-
-                    using (var responseStream = await response.Content.ReadAsStreamAsync())
+                    using (var requestExecutor = ClusterRequestExecutor.CreateForSingleNode(serverUrl, _serverStore.Server.Certificate.Certificate))
                     using (_serverStore.ContextPool.AllocateOperationContext(out JsonOperationContext context))
                     {
-                        var singleNodeDataBlittable = await context.ReadForMemoryAsync(responseStream, "studio-tasks-full-data-directory");
-                        return JsonDeserializationServer.SingleNodeDataDirectoryResult(singleNodeDataBlittable);
+                        var dataDirectoryInfo = new GetDataDirectoryInfoCommand(_path, _name, _isBackup);
+                        await requestExecutor.ExecuteAsync(dataDirectoryInfo, context, token: cts.Token);
+
+                        return dataDirectoryInfo.Result;
                     }
                 }
                 catch (OperationCanceledException)
@@ -171,11 +169,47 @@ namespace Raven.Server.Web.Studio
                 catch (Exception e)
                 {
                     if (Logger.IsInfoEnabled)
-                        Logger.Info($"Failed to get directory info result from: {url}", e);
+                        Logger.Info($"Failed to get directory info result from: {serverUrl}", e);
 
                     return null;
                 }
             }
+        }
+
+        private class GetDataDirectoryInfoCommand : RavenCommand<SingleNodeDataDirectoryResult>
+        {
+            private readonly string _path;
+            private readonly string _name;
+            private readonly bool _isBackup;
+
+
+            public GetDataDirectoryInfoCommand(string path, string name, bool isBackup)
+            {
+                _path = path;
+                _name = name;
+                _isBackup = isBackup;
+            }
+
+            public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, ServerNode node, out string url)
+            {
+                var path = _isBackup
+                    ? $"databases/{_name}/admin/backup-data-directory?path={_path}"
+                    : $"admin/studio-tasks/full-data-directory?path={_path}&name={_name}";
+
+                url = $"{node.Url}/{path}";
+
+                return new HttpRequestMessage
+                {
+                    Method = HttpMethod.Get
+                };
+            }
+
+            public override void SetResponse(JsonOperationContext context, BlittableJsonReaderObject response, bool fromCache)
+            {
+                Result = JsonDeserializationServer.SingleNodeDataDirectoryResult(response);
+            }
+
+            public override bool IsReadRequest => true;
         }
     }
 
--- a/src/Raven.Server/Web/Studio/DataDirectoryInfo.cs
+++ b/src/Raven.Server/Web/Studio/DataDirectoryInfo.cs
@@ -11,6 +11,7 @@ using Raven.Server.Json;
 using Raven.Server.ServerWide;
 using Raven.Server.ServerWide.Context;
 using Raven.Server.Utils;
+using Sparrow;
 using Sparrow.Json;
 using Sparrow.Json.Parsing;
 using Sparrow.Logging;
@@ -46,7 +47,9 @@ namespace Raven.Server.Web.Studio
             {
                 NodeTag = _serverStore.NodeTag,
                 FullPath = realPath,
-                TotalFreeSpaceHumane = diskSpaceInfo?.TotalFreeSpace.ToString(),
+                FreeSpaceInBytes = diskSpaceInfo?.TotalFreeSpace.GetValue(SizeUnit.Bytes) ?? 0,
+                FreeSpaceHumane = diskSpaceInfo?.TotalFreeSpace.ToString(),
+                TotalSpaceInBytes = diskSpaceInfo?.TotalSize.GetValue(SizeUnit.Bytes) ?? 0,
                 TotalSpaceHumane = diskSpaceInfo?.TotalSize.ToString()
             };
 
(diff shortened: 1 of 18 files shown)


src/Raven.Studio/typescript/viewmodels/resources/createDatabase.ts
```diff
@@ -242,20 +245,106 @@ class createDatabase extends dialogViewModelBase {
...
+            const selectedNodes = this.databaseModel.replication.nodes().map(x => x.tag());
+            if (this.databaseModel.replication.manualMode()) {
+                if (selectedNodes.length === 0) {
+                    return null;
+                }
+
+                return `${message}:`;
+            }
+
+            const numberOfClusterNodes = this.clusterNodes.slice().length;
```
ml054: what's the purpose of slice here?

src/Raven.Studio/typescript/viewmodels/resources/createDatabase.ts
```diff
@@ -242,20 +245,106 @@ class createDatabase extends dialogViewModelBase {
...
+            if (replicationFactor === 0) {
+                return null;
+            }
+
+            if (replicationFactor === 1 ||
+                replicationFactor === numberOfClusterNodes) {
+                return `${message}:`;
+            }
+
+            return `${message} ${replicationFactor} of the following nodes:`;
```
ml054: message is: The data files will be created in
so we return here:
The data files will be created in 3 of the following nodes?

src/Raven.Studio/typescript/viewmodels/resources/createDatabase.ts
```diff
@@ -242,20 +245,106 @@ class createDatabase extends dialogViewModelBase {
...
+
+        this.databaseLocationInfoToDisplay = ko.pureComputed(() => {
+            const databaseLocationInfo = this.databaseLocationInfo();
+            const selectedNodes = this.databaseModel.replication.nodes().map(x => x.tag());
+            const replicationFactor = this.databaseModel.replication.replicationFactor();
+            if (this.databaseModel.replication.manualMode()) {
+                if (selectedNodes.length === 0) {
+                    return [];
+                }
+            }
```
ml054: please remove line break after } and before else ;-)

src/Raven.Studio/wwwroot/App/views/resources/createDatabase.html
```diff
@@ -353,21 +353,51 @@ <h3>Path Configuration</h3>
...
         </div>
         <div class="form-group row">
-            <div class="col-sm-12">
+            <div class="col-sm-12" data-bind="if: $root.databaseLocationInfoMessage">
                 <p class="text-info bg-info padding padding-sm">
-                    The data files will be created in:<br/>
-                    <strong><small data-bind="text: $root.databaseLocationShowing"></small></strong>
+                    <span data-bind="text: $root.databaseLocationInfoMessage"></span><br />
+                    <div data-bind="if: $root.spinners.databaseLocationInfoLoading">
+                        <span class="btn-spinner"></span>
```
ml054: for inline spinner you should use class: 'global-spinner' IIRC

src/Raven.Studio/wwwroot/App/views/resources/createDatabase.html
```diff
@@ -353,21 +353,51 @@ <h3>Path Configuration</h3>
...
                 
-                <ul class="dropdown-menu" role="menu"
-                    data-bind="autoComplete: '#dataPath', foreach: $root.recentPathsAutocomplete.createCompleter()">
-                    <li role="presentation" data-bind="click: _.partial($root.recentPathsAutocomplete.complete, $data)">
+                <ul class="dropdown-menu max-height" role="menu"
+                    data-bind="autoComplete: '#dataPath', foreach: $root.allAutoCompleteOptions">
+                    <li role="presentation" data-bind="click: $data.isRecent ? _.partial($root.recentPathsAutocomplete.complete, $data.path) : $parent.dataPath.bind($parent, $data.path)">
                         <a role="menuitem" tabindex="-1" href="#">
-                            <span data-bind="text: $data"></span>
+                            <!-- ko if: $data.isRecent -->
```
ml054: no need for $data.

src/Raven.Studio/wwwroot/App/views/resources/createDatabase.html
```diff
@@ -353,21 +353,51 @@ <h3>Path Configuration</h3>
...
+                            <strong data-bind="text: $data.path"></strong>
+                            <!-- /ko -->
+                            <!-- ko ifnot: $data.isRecent -->
```
ml054: same here.

src/Raven.Studio/typescript/viewmodels/resources/createDatabase.ts
```diff
@@ -242,20 +245,106 @@ class createDatabase extends dialogViewModelBase {
...
+            if (replicationFactor === 1 ||
+                replicationFactor === numberOfClusterNodes) {
+                return `${message}:`;
+            }
+
+            return `${message} ${replicationFactor} of the following nodes:`;
+        });
+
+        this.allAutoCompleteOptions = ko.pureComputed(() => {
+            const result: { path: string, isRecent: boolean }[] = [];
```
ml054, line 313: I wonder if we need 2 separate groups here: recent and not-recent - I know recent are bold on UI. I think we should have single style for both and make sure we don't have duplicates on this list.

ml054: however .... recent are sorted by usage date and other are sorted A-Z, so maybe we should leave it.

ml054 (changes requested):
please have a look at my comments

src/Raven.Server/Web/Studio/DataDirectoryInfo.cs
```diff
@@ -0,0 +1,157 @@
...
+            }
+
+            return dataDirectoryResult;
+        }
+
+        private async Task UpdateNodesDirectoryResult(
+            IEnumerable<string> nodes, ClusterTopology clusterTopology,
+            string urlPath, DataDirectoryResult dataDirectoryResult)
+        {
+            var httpClient = RequestExecutor.CreateHttpClient(_serverStore.Server.Certificate.Certificate, DocumentConventions.Default);
```
ayende: Shouldn't this be disposed?

src/Raven.Server/Web/Studio/DataDirectoryInfo.cs
```diff
@@ -0,0 +1,157 @@
...
+                if (singleNodeResult == null)
+                    continue;
+
+                dataDirectoryResult.List.Add(singleNodeResult);
+            }
+        }
+
+        private async Task<SingleNodeDataDirectoryResult> GetSingleNodeDataDirectoryInfo(string url, HttpClient httpClient)
+        {
+            using (var cts = new CancellationTokenSource(7 * 1000))
```
ayende: Why 7 seconds?

src/Raven.Server/Web/Studio/DataDirectoryInfo.cs
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    public class SingleNodeDataDirectoryResult : IDynamicJson
+    {
+        public string NodeTag { get; set; }
+
+        public string FullPath { get; set; }
+
+        public string TotalFreeSpaceHumane { get; set; }
+
```
ayende, line 269: Better to also include the value as a number

src/Raven.Server/Web/Studio/StudioTasksHandler.cs
```diff
@@ -50,19 +51,43 @@ public Task FullDataDirectory()
...
-                    result = $"The administrator has restricted databases to be created only under the {RavenConfiguration.GetKey(x => x.Core.DataDirectory)} directory: '{ServerStore.Configuration.Core.DataDirectory.FullPath}'.";
+                    result =
+                        $"The administrator has restricted databases to be created only under the {RavenConfiguration.GetKey(x => x.Core.DataDirectory)} directory: '{ServerStore.Configuration.Core.DataDirectory.FullPath}'.";
                 }
             }
-            using (ServerStore.ContextPool.AllocateOperationContext(out JsonOperationContext context))
-            using (var writer = new BlittableJsonTextWriter(context, ResponseBodyStream()))
+
+            var drivesInfo = PlatformDetails.RunningOnPosix ? DriveInfo.GetDrives() : null;
+            var driveInfo = DiskSpaceChecker.GetDriveInfo(result, drivesInfo);
```
ayende: Why are we passing the `drivesInfo`?

src/Raven.Server/Web/Studio/FolderPath.cs
```diff
@@ -0,0 +1,100 @@
...
+using System.Collections.Generic;
+using System.IO;
+using Raven.Server.Config;
+using Sparrow.Platform;
+
+namespace Raven.Server.Web.Studio
+{
+    public static class FolderPath
+    {
+        private static readonly string FolderPathSuffix = PlatformDetails.RunningOnPosix ? "/" : "\\";
```
ayende: Replace with Path.DirectorySeparator

src/Raven.Server/Web/Studio/FolderPath.cs
```diff
@@ -0,0 +1,100 @@
...
+            else if (string.IsNullOrWhiteSpace(path))
+            {
+                folderPathOptions.List.AddRange(GetAvailableDrives());
+            }
+            else if (Directory.Exists(path))
+            {
+                path = path.Trim();
+
+                if (path.EndsWith("/") == false && path.EndsWith("\\") == false)
+                    path = $"{path}{FolderPathSuffix}";
```
ayende: Use Path.Combine

src/Raven.Server/Web/Studio/FolderPath.cs
```diff
@@ -0,0 +1,100 @@
...
+using Raven.Server.Config;
+using Sparrow.Platform;
+
+namespace Raven.Server.Web.Studio
+{
+    public static class FolderPath
+    {
+        private static readonly string FolderPathSuffix = PlatformDetails.RunningOnPosix ? "/" : "\\";
+
+        public static FolderPathOptions GetOptions(string path, bool isBackupFolder, RavenConfiguration ravenConfiguration)
```
ayende, line 10: What is this for?

src/Raven.Server/Web/System/OngoingTasksHandler.cs
```diff
@@ -500,6 +478,60 @@ private void VerifyPeriodicBackupConfiguration(ref BlittableJsonReaderObject rea
...
+
+            string actualFullPath;
+
+            try
+            {
+                actualFullPath = GetActualFullPath(path, out _);
+            }
+            catch (Exception e)
+            {
+                actualFullPath = e.Message;
```
ayende: What?!

ayende (changes requested):

src/Raven.Studio/typescript/viewmodels/resources/createDatabase.ts
grisha-kotler: fixed

src/Raven.Studio/typescript/viewmodels/resources/createDatabase.ts
grisha-kotler: Yes. For example: 5 nodes in a cluster. Replication factor: 3.
The data will be created in 3 of the 5 nodes.

src/Raven.Studio/typescript/viewmodels/resources/createDatabase.ts
grisha-kotler: done

src/Raven.Studio/wwwroot/App/views/resources/createDatabase.html
grisha-kotler: done

src/Raven.Studio/wwwroot/App/views/resources/createDatabase.html
grisha-kotler: done

src/Raven.Studio/wwwroot/App/views/resources/createDatabase.html
grisha-kotler: done

src/Raven.Server/Web/Studio/DataDirectoryInfo.cs
grisha-kotler: The HttpClient is cached (by the certificate thumbprint)
https://github.com/ravendb/ravendb/blob/v4.1/src/Raven.Client/Http/RequestExecutor.cs#L132

src/Raven.Server/Web/Studio/DataDirectoryInfo.cs
grisha-kotler: No reason. Should it be lower?

src/Raven.Server/Web/Studio/DataDirectoryInfo.cs
grisha-kotler: done

src/Raven.Server/Web/Studio/StudioTasksHandler.cs
grisha-kotler: In linux we need it to find the root mount (drive name in Windows)

src/Raven.Server/Web/Studio/FolderPath.cs
grisha-kotler: done

src/Raven.Server/Web/Studio/FolderPath.cs
grisha-kotler: done

src/Raven.Server/Web/Studio/StudioTasksHandler.cs
ayende: I don't follow?

src/Raven.Server/Web/Studio/DataDirectoryInfo.cs
```diff
@@ -0,0 +1,216 @@
...
+            }
+        }
+
+        private async Task<SingleNodeDataDirectoryResult> GetSingleNodeDataDirectoryInfo(string url, HttpClient httpClient)
+        {
+            using (var cts = new CancellationTokenSource(7 * 1000))
+            {
+                try
+                {
+                    var request = new HttpRequestMessage(HttpMethod.Get, url);
```
ppekrol: We should use custom RavenCommand instead of doing all of this? We have a RequestExecutor here available.

ppekrol: Then you would not need any HttpClient here and all of those changes made there.

src/Raven.Server/Web/Studio/DataDirectoryInfo.cs
ppekrol: I do not see those?

src/Raven.Server/Web/Studio/FolderPath.cs
ppekrol: I do not see that change?

src/Raven.Server/Web/Studio/StudioTasksHandler.cs
```diff
@@ -50,16 +48,31 @@ public Task FullDataDirectory()
...
                 {
-                    result = $"The administrator has restricted databases to be created only under the {RavenConfiguration.GetKey(x => x.Core.DataDirectory)} directory: '{ServerStore.Configuration.Core.DataDirectory.FullPath}'.";
+                    result =
+                        $"The administrator has restricted databases to be created only under the {RavenConfiguration.GetKey(x => x.Core.DataDirectory)} directory: '{ServerStore.Configuration.Core.DataDirectory.FullPath}'.";
                 }
             }
+
+            var getNodesInfo = GetBoolValueQueryString("getNodesInfo", required: false) ?? false;
+            var info = new DataDirectoryInfo(ServerStore, result, getNodesInfo, ResponseBodyStream());
+            var urlPath = $"admin/studio-tasks/full-data-directory?path={path}&name={name}";
```
ppekrol: Name and Path should be parameters to command. Endpoint url should be hardcoded inside the command.

src/Raven.Server/Web/System/OngoingTasksHandler.cs
```diff
@@ -500,6 +478,60 @@ private void VerifyPeriodicBackupConfiguration(ref BlittableJsonReaderObject rea
...
+            }
+
+            var getNodesInfo = GetBoolValueQueryString("getNodesInfo", required: false) ?? false;
+            var info = new DataDirectoryInfo(ServerStore, actualFullPath, getNodesInfo, ResponseBodyStream());
+            var urlPath = $"databases/{Database.Name}/admin/backup-data-directory?path={path}";
+            await info.UpdateDirectoryResult(urlPath, Database.Name);
```
ppekrol: use command

ppekrol (changes requested):

src/Raven.Server/Web/Studio/DataDirectoryInfo.cs
ayende: Better to take that as an argument and not hard code it

src/Raven.Server/Web/Studio/DataDirectoryInfo.cs
ayende: Agree, we don't want to manually use these

src/Raven.Studio/typescript/viewmodels/resources/createDatabase.ts
ml054: Wait it will be ' in 3 of the following nodes?'
not: created in 3 of the 5 nodes.
I think 'created in 3 of the 5 nodes.' is better.

grisha-kotler: done

src/Raven.Server/Web/Studio/FolderPath.cs
grisha-kotler: pushed my changes

src/Raven.Server/Web/Studio/FolderPath.cs
grisha-kotler: This will display auto complete options when starting to write the path in the studio
image

src/Raven.Server/Web/Studio/DataDirectoryInfo.cs
grisha-kotler: done

src/Raven.Server/Web/Studio/StudioTasksHandler.cs
grisha-kotler: done

src/Raven.Server/Web/System/OngoingTasksHandler.cs
grisha-kotler: done

src/Raven.Server/Web/Studio/DataDirectoryInfo.cs
grisha-kotler: done

src/Raven.Server/Web/Studio/DataDirectoryInfo.cs
grisha-kotler: pushed my changes

src/Raven.Server/Web/Studio/StudioTasksHandler.cs
grisha-kotler: We need the DrivesInfo in Linux to find the root mount.
The reason for that is that the provided URL might not exists.
For example: /mnt/Data/NewDatabase. The root mount is: /mnt/Data/.
Running the DiskSpaceChecker on a folder that doesn't exist doesn't provide any result.

https://github.com/ravendb/ravendb/blob/v4.1/src/Sparrow/Utils/DiskSpaceChecker.cs#L141

src/Raven.Server/Web/System/OngoingTasksHandler.cs
grisha-kotler: Example: Nodes A & B.
Path: E:\Temp (Exists on Node A but doesn't exist on Node B).

This will output the free space on Node A and output that the path doesn't exist on Node B.

ayende: Not like this. Create a class that has the full path and an error, not overload the path in this manner.

grisha-kotler: done

src/Raven.Server/Web/Studio/DataDirectoryInfo.cs
```diff
@@ -90,6 +96,38 @@ public async Task UpdateDirectoryResult(string databaseName)
             }
         }
 
+        public static bool CanAccessPath(string folderPath, out string error)
+        {
+            var originalFolderPath = folderPath;
+            while (true)
+            {
+                var directoryInfo = new DirectoryInfo(folderPath);
+                if (directoryInfo.Exists == false)
```
ayende, line 105: Can this throw?

grisha-kotler: No. It will return false if any error occurs while trying to determine if the path exists.

PR #8380, opened 2019-01-02
RavenDB-12491 - Server side work

ayende:
Probably would need more work, but can now detect missing libsodium / librvnpal and change what we get from the server. 
Still TBD - making a studio page that shows it.


ppekrol:
Closing because Marcin is keeping that in his branch now: https://github.com/ml054/ravendb/tree/RavenDB-12491

PR #8383, opened 2019-01-02
RavenDB-12556 Last flushed properties binned together to prevent using half updated values

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-12556#focus=streamItem-67-33225-0-0

`GatherInformationToStartSync` method reads the "Last flushed" properties and can run concurrently with `UpdateJournalStateUnderWriteTransactionLock` method that changes thos properties. 
This can lead to a situation that the sync operation gathers half information before the update and half after.

### DIFF
--- a/src/Voron/Impl/Journal/WriteAheadJournal.cs
+++ b/src/Voron/Impl/Journal/WriteAheadJournal.cs
@@ -468,10 +468,15 @@ namespace Voron.Impl.Journal
             private readonly ManualResetEventSlim _waitForJournalStateUpdateUnderTx = new ManualResetEventSlim();
             private readonly LockTaskResponsible _flushLockTaskResponsible;
 
-            private long _lastFlushedTransactionId;
-            private long _lastFlushedJournalId;
+            public class LastFlushState
+            {
+               public long TransactionId;
+               public long JournalId;
+               public JournalFile Journal;
+            }
+
+            private LastFlushState _lastFlushed = new LastFlushState();
             private long _totalWrittenButUnsyncedBytes;
-            private JournalFile _lastFlushedJournal;
             private bool _ignoreLockAlreadyTaken;
             private Action<LowLevelTransaction> _updateJournalStateAfterFlush;
 
@@ -481,8 +486,8 @@ namespace Voron.Impl.Journal
                 action?.Invoke(tx);
             }
 
-            public long LastFlushedTransactionId => Interlocked.Read(ref _lastFlushedTransactionId);
-            public long LastFlushedJournalId => Interlocked.Read(ref _lastFlushedJournalId);
+            public long LastFlushedTransactionId => Interlocked.Read(ref _lastFlushed.TransactionId);
+            public long LastFlushedJournalId => Interlocked.Read(ref _lastFlushed.JournalId);
             public long TotalWrittenButUnsyncedBytes => Interlocked.Read(ref _totalWrittenButUnsyncedBytes);
             public int JournalsToDeleteCount => _journalsToDelete.Count;
 
@@ -527,7 +532,8 @@ namespace Voron.Impl.Journal
                     if (jrnls.Count == 0)
                         return; // nothing to do
 
-                    Debug.Assert(jrnls.First().Number >= _lastFlushedJournalId);
+                    var lastFlushed = _lastFlushed;
+                    Debug.Assert(jrnls.First().Number >= lastFlushed.JournalId);
 
                     var pagesToWrite = new Dictionary<long, PagePosition>();
 
@@ -540,7 +546,7 @@ namespace Voron.Impl.Journal
 
                     foreach (var journalFile in jrnls)
                     {
-                        if (journalFile.Number < _lastFlushedJournalId)
+                        if (journalFile.Number < lastFlushed.JournalId)
                             continue;
                         var currentJournalMaxTransactionId = -1L;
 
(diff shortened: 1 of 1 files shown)


src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -481,8 +486,8 @@ public void OnTransactionCommitted(LowLevelTransaction tx)
                 action?.Invoke(tx);
             }
 
-            public long LastFlushedTransactionId => Interlocked.Read(ref _lastFlushedTransactionId);
-            public long LastFlushedJournalId => Interlocked.Read(ref _lastFlushedJournalId);
+            public long LastFlushedTransactionId => Interlocked.Read(ref _lastFlushed.TransactionId);
```
ayende: No, you need:

```suggestion
            public long LastFlushedTransactionId => Interlocked.Read(ref _lastFlushed).TransactionId;
```

In all places.
The values are _never_ mutated

src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -738,9 +744,12 @@ public void ApplyLogsToDataFile(CancellationToken token, TimeSpan timeToWait)
 
             private void UpdateJournalStateUnderWriteTransactionLock(LowLevelTransaction txw, long lastProcessedJournal, long lastFlushedTransactionId, List<JournalFile> unusedJournals)
             {
-                _lastFlushedJournalId = lastProcessedJournal;
-                _lastFlushedTransactionId = lastFlushedTransactionId;
-                _lastFlushedJournal = _waj._files.First(x => x.Number == lastProcessedJournal);
+                _lastFlushed = new LastFlushState
```
ayende: Interlocked?

ayende (changes requested):

src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -468,10 +468,15 @@ public sealed class JournalApplicator : IDisposable
...
             private readonly LockTaskResponsible _flushLockTaskResponsible;
 
-            private long _lastFlushedTransactionId;
-            private long _lastFlushedJournalId;
+            public class LastFlushState
+            {
+               public long TransactionId;
+               public long JournalId;
+               public JournalFile Journal;
+            }
```
arekpalinski, line 484: As we never mutate those values (since `LastFlushState` is the snapshot) I'd mark all of those fields as `readonly` and create a dedicated constructor to assign them

src/Voron/Impl/Journal/WriteAheadJournal.cs
haludi: I think it is not necessary here because I don't really care what is the current values. It just needs to be all or nothing and in addition, we are under flush lock or write transaction lock so it will be updated anyway.
Maybe just to be explicit? or there is something else?

ayende: Great idea

src/Voron/Impl/Journal/WriteAheadJournal.cs
ayende: Yes, that's fine

haludi: O.k. except the `Journal` that assigned so the next sync will see there is nothing more to sync, right?

src/Voron/Impl/Journal/WriteAheadJournal.cs
haludi: So actually I don't need `Interlock` here, right?

src/Voron/Impl/Journal/WriteAheadJournal.cs
ayende: I'm not following?

src/Voron/Impl/Journal/WriteAheadJournal.cs
ayende: Yes, that would be fine without it

src/Voron/Impl/Journal/WriteAheadJournal.cs
arekpalinski: Can we set entire `LastFlushState` object as `null` then instead?

haludi: O.k. sound good

haludi: I checked it and this leads to some problems. We lose the information about the `TransactionId` and `JournalId` and the code rely on them in `ApplyLogsToDataFile` and in the gather details for reports

arekpalinski: OK. Then we can leave `Journal` field as not readonly or create a dedicated boolean property (not readonly) to mark that this has been already synced.

ayende: The journal is assigned only on `UpdateJournalStateUnderWriteTransactionLock`, with the rest
I'm not following on the issue here?

arekpalinski: See here:
https://github.com/ravendb/ravendb/pull/8383/files#diff-33af42ec0439152b32cc47c6739f3211R979

` lastFlushed.Journal = null;`

ayende: Better to add a flag, which we can set when we finished syncing.

haludi: I will use `new SingleUseFlag();`
O.k.?

ayende: ye

arekpalinski (approved):

ayende (approved):

PR #8385, opened 2019-01-02
RavenDB-11698 allow configuring different max log file size

ppekrol:


src/Raven.Server/Config/Categories/LogsConfiguration.cs
```diff
@@ -18,5 +19,10 @@ public class LogsConfiguration : ConfigurationCategory
         [DefaultValue(true)]
         [ConfigurationEntry("Logs.UseUtcTime", ConfigurationEntryScope.ServerWideOnly)]
         public bool UseUtcTime { get; set; }
+
+        [DefaultValue(128)]
```
ppekrol, line 23: Changed default from 256 to 128

ayende (approved):

PR #8393, opened 2019-01-03
RavenDB-9889 Added conversion events to DocumentStore

ppekrol:


src/Raven.Client/Documents/Session/Events.cs
```diff
@@ -83,4 +84,77 @@ public SessionCreatedEventArgs(InMemoryDocumentSessionOperations session)
...
+        }
+    }
+
+    public class AfterConversionToDocumentEventArgs : EventArgs
+    {
+        public string Id { get; }
+
+        public object Entity { get; }
+
+        public BlittableJsonReaderObject Document { get; }
```
ayende: Shouldn't we provide a setter to give the user the chance to modify the document?

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -1683,14 +1686,48 @@ protected void UpdateSessionAfterSaveChanges(BatchCommandResult result)
...
+        }
+
+        internal void OnAfterConversionToDocumentInvoke(string id, object entity, ref BlittableJsonReaderObject document)
+        {
+            var onAfterConversionToDocument = OnAfterConversionToDocument;
+            if (onAfterConversionToDocument != null)
+            {
+                onAfterConversionToDocument.Invoke(this, new AfterConversionToDocumentEventArgs(this, id, entity, document));
+
+                if (document.Modifications != null)
```
ayende: I'm not sure if this is a good idea. We might want to just have them do that and just set the new instance value, instead of relying on `Modifications`

ppekrol: Done

ayende (approved):

PR #8412, opened 2019-01-04
RavenDB-12629 InvalidJournalException: No such journal on database re…

arekpalinski:
…covery:

- updating _journalsToDelete collection under write tx lock
- making journalsToDelete collection to be concurrent dictionary since UpdateJournalStateUnderWriteTransactionLock can be run concurrently with UpdateDatabaseStateAfterSync / GatherInformationToStartSync
- including the collection of journals to delete into LastFlushState so we have the snapshot state after the flushing when syncing (as of sync was requested)
- we remove items from _journalsToDelete _after_ we successfully sync

### DIFF
--- a/src/Voron/Impl/Journal/WriteAheadJournal.cs
+++ b/src/Voron/Impl/Journal/WriteAheadJournal.cs
@@ -7,6 +7,7 @@
 using Sparrow;
 using Sparrow.Binary;
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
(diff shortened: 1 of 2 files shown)


arekpalinski:
@haludi Please review

src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -938,12 +939,11 @@ private bool UpdateDatabaseStateAfterSync()
...
                 {
-                    if (_lastSyncedJournal == -1
-                        && _currentTotalWrittenBytes == -1
-                        && _lastSyncedTransactionId == -1)
+                    if (_lastFlushed == null && _currentTotalWrittenBytes == -1)
+                    {
                         throw new InvalidOperationException(
-                            $"Try to {nameof(UpdateDatabaseStateAfterSync)} " +
-                            $"without to call {nameof(GatherInformationToStartSync)} before");
+                            $"Try to {nameof(UpdateDatabaseStateAfterSync)} without to call {nameof(GatherInformationToStartSync)} before");
```
ayende, line 945: Without "calling", not "to call"

arekpalinski: Fixed in https://github.com/ravendb/ravendb/pull/8424

PR #8418, opened 2019-01-06
RavenDB-12022 :  CountesStorage redesign 

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-12022


src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -117,31 +132,11 @@ public IEnumerable<CounterDetail> GetCountersFrom(DocumentsOperationContext cont
...
-            {
-                if (take-- <= 0)
-                    yield break;
-
-                yield return TableValueToCounterDetail(context, result.Reader);
-            }
-        }
-
-        public long GetNumberOfCountersToProcess(DocumentsOperationContext context, string collection, long afterEtag, out long totalCount)
+       public long GetNumberOfCountersToProcess(DocumentsOperationContext context, string collection, long afterEtag, out long totalCount)
```
maximburyak: GetNumberOfCountersToProcess does ont exactly number of counters, am I right? I think the name should be more accurate here, like get number of documents to process (which also sounds strange)

src/Raven.Client/Documents/Operations/Counters/CountersDetail.cs
```diff
@@ -46,4 +46,26 @@ public DynamicJsonValue ToJson()
             };
         }
     }
+
+    public class CounterGroupDetail
+    {
+        public string DocumentId { get; set; }
```
maximburyak: Can DocumentId and ChangeVector be LazyStringValue instead of string?

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -228,61 +203,325 @@ private static (LazyStringValue Doc, LazyStringValue Name) ExtractDocIdAndName(J
...
+                    }
+                    else
+                    {
+                        // counter doesn't exists 
+                        CreateNewCounter(context, documentId, name, dbIdIndex, value, newETag, data);
+                    }
+
+                    if (data.Modifications != null)
+                    {
+                        data = context.ReadObject(data, null, BlittableJsonDocumentBuilder.UsageMode.ToDisk);
```
maximburyak: I think we should dispose the "old" data here

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -228,61 +203,325 @@ private static (LazyStringValue Doc, LazyStringValue Name) ExtractDocIdAndName(J
...
+            if (context.Transaction == null)
+            {
+                DocumentPutAction.ThrowRequiresTransaction();
+                Debug.Assert(false);// never hit
+            }
+
+            var collectionName = _documentsStorage.ExtractCollectionName(context, collection);
+            var table = GetCountersTable(context.Transaction.InnerTransaction, collectionName);
+
+            using (DocumentIdWorker.GetLowerIdSliceAndStorageKey(context, documentId, out Slice lowerId, out _))
```
maximburyak: we can probably call it docLowerId

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -228,61 +203,325 @@ private static (LazyStringValue Doc, LazyStringValue Name) ExtractDocIdAndName(J
...
+
+                return result.ChangeVector;
+            }
+        }
+
+        private static int GetDbIdIndex(DocumentsOperationContext context, BlittableJsonReaderObject data, int dbIdIndex)
+        {
+            if (data.TryGet(DbIds, out BlittableJsonReaderArray dbIds))
+            {
+                for (dbIdIndex = 0; dbIdIndex < dbIds.Length; dbIdIndex++)
```
maximburyak: Can DbIDs be a sorted list and a binary search should be performed, or even an object, to do it using the existing blittable mechanisms?

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -296,29 +535,193 @@ private void PutCounterImpl(DocumentsOperationContext context, string documentId
...
                     }
+                }
+            }
+        }
 
-                    UpdateMetrics(counterKey, name, changeVector, collection);
+        private static Dictionary<LazyStringValue, int> DbIdToIndex(BlittableJsonReaderArray dbIds)
+        {
+            var count = 0;
+            return new Dictionary<LazyStringValue, int>(dbIds.Items.ToDictionary(x => (LazyStringValue)x, x => count++));
```
maximburyak: It's better to implement the iteration here, instead of using ToDictionary with a lambda call
Also, can we use some kind of dictionary caching/cleaning here as we did with _scopes?

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -296,29 +535,193 @@ private void PutCounterImpl(DocumentsOperationContext context, string documentId
...
+        }
+
+        private static int FindMaxDbIdIndex(int sourceCount, Dictionary<LazyStringValue, int> dict, BlittableJsonReaderArray localDbIds, BlittableJsonReaderArray sourceDbIds)
+        {
+            var maxDbIdIndex = -1;
+
+            for (int i = 0; i < sourceCount; i++)
+            {
+                var sourceDbId = (LazyStringValue)sourceDbIds[i];
+                if (dict.TryGetValue(sourceDbId, out var dbIdIndex))
```
maximburyak: Please add caching of the calculated HashCode to LazyStringValue, so this will cost a bit less

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -37,6 +38,12 @@ public unsafe class CountersStorage
 
         public static readonly string CountersTombstones = "Counters.Tombstones";
 
+        public const string DbIds = "@DbIds";
+
+        private long _countersCount;
+
+        private readonly List<ByteStringContext<ByteStringMemoryCache>.InternalScope> _scopes = new List<ByteStringContext<ByteStringMemoryCache>.InternalScope>();
```
maximburyak: Please find more meaningfull name for that field - something like _counterModificationMemoryScopes or better (probably).

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -228,61 +203,325 @@ private static (LazyStringValue Doc, LazyStringValue Name) ExtractDocIdAndName(J
...
-                        if (table.ReadByKey(counterKey, out var existing))
+                        var existingChangeVector = TableValueToChangeVector(context, (int)CountersTable.ChangeVector, ref existing);
+
+                        if (ChangeVectorUtils.GetConflictStatus(changeVector, existingChangeVector) == ConflictStatus.AlreadyMerged)
+                            return;
+
+                        data = new BlittableJsonReaderObject(existing.Read((int)CountersTable.Data, out int oldSize), oldSize, context);
+                        data = data.Clone(context);
+                        data.TryGet(DbIds, out BlittableJsonReaderArray dbIds);
+                        _scopes.Clear();
```
maximburyak: we should probably clear that field in a finally statement

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -296,29 +535,193 @@ private void PutCounterImpl(DocumentsOperationContext context, string documentId
...
+            using (GetCounterPartialKey(context, documentId, counterName, out var counterKey))
+            {
+                RemoveTombstoneIfExists(context, counterKey);
+            }
+
+            var sourceCount = counterValues.Length / SizeOfCounterValues;
+            var maxDbIdIndex = FindMaxDbIdIndex(sourceCount, dict, localDbIds, sourceDbIds);
+            var sourceValues = (CounterValues*)counterValues.Ptr;
+
+            _scopes.Add(context.Allocator.Allocate((maxDbIdIndex + 1) * SizeOfCounterValues, out var newVal));
```
maximburyak: it's probably best to pass _scopes as parameter - it's caching is only an implementation detail

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -296,29 +535,193 @@ private void PutCounterImpl(DocumentsOperationContext context, string documentId
...
+
+        private long PutExistingCounter(DocumentsOperationContext context, BlittableJsonReaderObject data, LazyStringValue counterName,
+            BlittableJsonReaderArray localDbIds,
+            BlittableJsonReaderArray sourceDbIds, Dictionary<LazyStringValue, int> dict, BlittableJsonReaderObject.RawBlob existingCounter,
+            BlittableJsonReaderObject.RawBlob source, out bool modified)
+        {
+            long value = 0;
+            modified = false;
+            var existingCount = existingCounter.Length / SizeOfCounterValues;
+            var sorted = SortValuesByDbIdIndex(dict, localDbIds, sourceDbIds, source.Length / SizeOfCounterValues);
```
maximburyak: This should be called lazily, so the "sorted" list will be held in the calling method

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -296,29 +535,193 @@ private void PutCounterImpl(DocumentsOperationContext context, string documentId
...
+            };
+
+            return scope;
+
+        }
+
+        private static int[] SortValuesByDbIdIndex(Dictionary<LazyStringValue, int> dict,
+            BlittableJsonReaderArray localDbIds, BlittableJsonReaderArray sourceDbIds, int sourceCount)
+        {
+            var sorted = new int[dict.Count + sourceDbIds.Length];
```
maximburyak: Use an array pool here, probably

src/Raven.Server/Documents/CountersStorage.cs
maximburyak: SortValuesByDbIdIndex's main purpose seems to be  making sure the dict holds all dbIDs, after that, the iteration in line 610 can be performed on the dictionary values, I don't think we need "sorted" at all here, or am I missing something?

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -296,29 +535,193 @@ private void PutCounterImpl(DocumentsOperationContext context, string documentId
...
+                    sorted[newIndex] = i;
+
+                    dict.Add(dbId, newIndex);
+                }
+            }
+
+            return sorted;
+        }
+
+        private static int FindMaxDbIdIndex(int sourceCount, Dictionary<LazyStringValue, int> dict, BlittableJsonReaderArray localDbIds, BlittableJsonReaderArray sourceDbIds)
```
maximburyak: I guess that this function can be merged with SortValuesByDbIdIndex, given that it won't need to be sorting anything, but just intializing the dict and modifying the localDbIds

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -495,59 +831,29 @@ public IEnumerable<(string ChangeVector, long Value)> GetCounterValues(Documents
         {
```
maximburyak, line 897: We can probably abandon the IEnumerable return type for this function and for GetCountersForDocument

maximburyak (changes requested):

src/Raven.Server/Documents/CountersStorage.cs
aviv86: no, it's not just to make sure the dict holds all dbIDs.
the main purpose here was to avoid calling `context.Allocator.Allocate()` multiple times and moving the data around.
by sorting the source dbIDs by their local dbId-index, then iterating the sorted array in reverse order, we assure that we have a single call to `context.Allocator.Allocate()` (at most) and that we're not moving the data more than once.

src/Raven.Server/Documents/CountersStorage.cs
aviv86: yes, this was used only in counters ETL,  which doesn't work at all now. when i'll fix ETL part i'll change the name as well. maybe GetNumberOfCounterGroupsToProcess ?

src/Raven.Server/Documents/CountersStorage.cs
aviv86: the order of the counter partial-values in a raw blob rely on the order of the the DbIDs, so i don't think a sorted set is a good fit here - inserting a new dbId in the middle of the list will force us to modify all the raw data. 
object can work.

maximburyak: let's go with object then, but first, can you profile that on a multiple dbIndex scenario and see if GetDbIdIndex method is expensive? maybe it's just a preliminary optimization..

src/Raven.Client/Exceptions/Documents/Counters/CounterOverflowException.cs
```diff
@@ -16,12 +16,12 @@ public CounterOverflowException(string message, Exception inner) : base(message,
         {
         }
 
-        internal static void ThrowFor(string docId, string counterName, long value, long delta, OverflowException inner)
+        public static void ThrowFor(string docId, string counterName, long value, long delta, OverflowException inner)
```
ppekrol: can be `internal` ?

src/Raven.Client/Exceptions/Documents/Counters/CounterOverflowException.cs
```diff
@@ -16,12 +16,12 @@ public CounterOverflowException(string message, Exception inner) : base(message,
...
         {
             throw new CounterOverflowException($"Could not increment counter '{counterName}' from document '{docId}' with value '{value}' by '{delta}'.", inner);
         }
 
-        internal static void ThrowFor(string docId, string counterName, OverflowException inner)
+        public static void ThrowFor(string docId, string counterName, OverflowException inner)
```
ppekrol: can be `internal` ?

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -228,61 +203,325 @@ private static (LazyStringValue Doc, LazyStringValue Name) ExtractDocIdAndName(J
...
+
+                var dbIdIndex = int.MaxValue;
+                var newETag = _documentsStorage.GenerateNextEtag();
+
+                if (data != null)
+                {
+                    // Common case is that we modify the data IN PLACE
+                    // as such, we must copy it before modification
+
+                    data = data.Clone(context);
```
ppekrol: Need to dispose the old data

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -228,61 +203,325 @@ private static (LazyStringValue Doc, LazyStringValue Name) ExtractDocIdAndName(J
...
+                    if (changeVector != null && table.ReadByKey(lowerId, out var existing))
                     {
-                        if (table.ReadByKey(counterKey, out var existing))
+                        var existingChangeVector = TableValueToChangeVector(context, (int)CountersTable.ChangeVector, ref existing);
+
+                        if (ChangeVectorUtils.GetConflictStatus(changeVector, existingChangeVector) == ConflictStatus.AlreadyMerged)
+                            return;
+
+                        data = new BlittableJsonReaderObject(existing.Read((int)CountersTable.Data, out int oldSize), oldSize, context);
+                        data = data.Clone(context);
```
ppekrol: Need to dispose the old 'data' ?

src/Raven.Server/Documents/CountersStorage.cs
ppekrol: I do not like that concept. It would be better to add method `RegisterForDispose(IDisposable toDispose)` to DocumentsOperationContext to be absolutely sure that we will dispose everything at the end, even when some exception occured. Thoughts?

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -602,72 +891,131 @@ public void DeleteCountersForDocument(DocumentsOperationContext context, string
...
+            return changeVectorForTombstone;
+        }
+
+        private void RemoveCounterFromBlittableAndUpdateTable(DocumentsOperationContext context, string counterName, BlittableJsonReaderObject data, CollectionName collectionName, Table table,
+            Slice lowerId)
+        {
+            data.Modifications = new DynamicJsonValue(data);
+            data.Modifications.Remove(counterName);
+
+            data = context.ReadObject(data, null, BlittableJsonDocumentBuilder.UsageMode.ToDisk);
```
ppekrol: dispose old 'data' ?

src/Raven.Server/Documents/Handlers/CountersHandler.cs
```diff
@@ -253,6 +229,218 @@ private static void ThrowInvalidBatchOperationType(CounterOperation operation)
...
+                _dictionary = new Dictionary<string, CounterGroup>();
+                _fromEtl =  false;
+            }
+
+            public void Add(string id, CounterGroup cg)
+            {
+                _dictionary.Add(id, cg);
+            }
+
+            public void Add(string id, CounterDetail counterDetail, out bool isNew)
```
ppekrol: Rename to `AddLegacy`

ppekrol (changes requested):

src/Raven.Server/Documents/CountersStorage.cs
ayende: I don't like this, you are holding resources here per transaction that are held by an object that outlive it.

ayende: @ppekrol That would be better, yes

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -45,14 +52,22 @@ public unsafe class CountersStorage
...
-            // lower document id, record separator, lower counter name, record separator, 16 bytes dbid
-            CounterKey = 0,
-            Name = 1, // format of lazy string key is detailed in GetLowerIdSliceAndStorageKey
-            Etag = 2,
-            Value = 3,
-            ChangeVector = 4,
-            Collection = 5,
-            TransactionMarker = 6
+            // lower document id, record separator, prefix 
+            DocumentId = 0,
```
ayende: CounterKey, instead. This isn't the document id

src/Raven.Client/Documents/Operations/Counters/CountersDetail.cs
ayende: We probably should, yes

src/Raven.Server/Documents/CountersStorage.cs
ayende: No, this is intentionally a very simple array. In 99.999% of the cases, the number here is going to be < 5, so linear scan is better.

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -228,61 +203,325 @@ private static (LazyStringValue Doc, LazyStringValue Name) ExtractDocIdAndName(J
...
+        private void IncrementExistingCounter(DocumentsOperationContext context, string documentId, string name, long delta, ref bool exists, BlittableJsonReaderObject.RawBlob existingCounter,
+            int dbIdIndex, long newETag, BlittableJsonReaderObject data, long value)
         {
-            PutCounterImpl(context, documentId, collection, name, changeVector, value);
+            var existingCount = existingCounter.Length / SizeOfCounterValues;
+
+            if (dbIdIndex < existingCount)
+            {
+                exists = true;
+                var counter = (CounterValues*)existingCounter.Ptr + dbIdIndex;
```
ayende, line 413:
```suggestion
                var counter = &((CounterValues*)existingCounter.Ptr)[dbIdIndex];
```

Is that more readable?

src/Raven.Server/Documents/CountersStorage.cs
ayende: Why do we need this at all? It would be better to do linear scan each time than pay the allocation / building price of the dictionary

src/Raven.Server/Documents/CountersStorage.cs
ayende: I really don't like `_scopes`, seems like it should be a value that is passed to the method from caller and then disposes

src/Raven.Server/Documents/CountersStorage.cs
ayende: I don't follow why this is required? But in general we should kill the dictionary and not do any sorting

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -375,116 +776,51 @@ public Table GetCountersTable(Transaction tx, CollectionName collection)
...
-                    tvb.Add(counterKey);
-                    tvb.Add(nameSlice);
-                    tvb.Add(Bits.SwapBytes(etag));
-                    tvb.Add(value);
-                    tvb.Add(cv);
-                    tvb.Add(collectionSlice);
-                    tvb.Add(context.TransactionMarkerOffset);
-
-                    table.Set(tvb);
+                    if (prop == DbIds)
```
ayende: Question, what happens if a user defines a counter named `DbIds` ? Won't it kill us?

ayende: Probably we should define the DbIds property as `@` and avoid a counter of this name

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -602,72 +891,131 @@ public void DeleteCountersForDocument(DocumentsOperationContext context, string
...
         }
 
-        public string DeleteCounter(DocumentsOperationContext context, string documentId, string collection, string counterName)
+        public string DeleteCounter(DocumentsOperationContext context, Slice counterKey, string collection, long lastModifiedTicks, bool forceTombstone)
+        {
+            var (doc, name) = ExtractDocIdAndName(context, counterKey);
+            return DeleteCounter(context, doc, collection, name, forceTombstone, lastModifiedTicks);
+        }
+
+        public string DeleteCounter(DocumentsOperationContext context, string documentId, string collection, string counterName, bool forceTombstone = false, long lastModifiedTicks = -1)
```
ayende: Do we _need_ a separate tombstone?
Can we mark the property itself on the object as deleted? 
Something like, if counter property is null, it is deleted?

src/Raven.Server/Documents/Handlers/CountersHandler.cs
```diff
@@ -253,6 +229,218 @@ private static void ThrowInvalidBatchOperationType(CounterOperation operation)
...
+            {
+                isNew = false;
+                _legacyDictionary = _legacyDictionary ?? new Dictionary<string, Dictionary<string, List<(string ChangeVector, long Value)>>>();
```
ayende: Case insensitive on the doc id

src/Raven.Server/Documents/Handlers/CountersHandler.cs
```diff
@@ -253,6 +229,218 @@ private static void ThrowInvalidBatchOperationType(CounterOperation operation)
...
+
+                if (doc != null)
+                    docCollection = CollectionName.GetCollectionName(doc.Data);
+
+                _database.DocumentsStorage.CountersStorage.PutCounters(context, counters.DocId, docCollection,
+                    counters.CounterGroup.ChangeVector, counters.CounterGroup.Values);
+
+                foreach (var counter in counters.CounterGroup.Values.GetPropertyNames())
+                {
+                    if (counter.Equals(CountersStorage.DbIds))
```
ayende: How can this be?

src/Raven.Server/Smuggler/Documents/StreamDestination.cs
```diff
@@ -407,34 +407,70 @@ public void WriteIndex(IndexDefinition indexDefinition)
...
+                Writer.WriteObject(counterDetail.Values);
 
                 Writer.WriteEndObject();
             }
 
-            public StreamCounterActions(BlittableJsonTextWriter writer, string propertyName) : base(writer, propertyName)
+            public void WriteLegacyCounter(CounterDetail counterDetail)
+            {
+                // Used only in Database Destination 
+                throw new NotImplementedException();
```
ayende: NotSupported, and with a proper message

ayende (changes requested):

src/Raven.Server/Documents/CountersStorage.cs
aviv86: I don't think we should kill the dictionary.
without a dictionary, doesn't it mean that for each partial value in each counter, 
we will need to do a liner scan on `localDbIds` (and if missing, look in `localDbIds.Modifications` ?) to get the DbId-index?

but yes, maybe sorting can be avoided.
I was think of a case like this one:
- local DbIds = [A, B, C, D]
- local counter 'likes' has partial value of A
- source sends counter 'likes' with partial values of B, C, D
- all of these partial values are missing
- if we don't do any sorting, each iteration will go to `AddPartialValueToExistingCounter()` , allocate `newVal`, copy old data, etc.
- with sorting by local DbId (O(N)), we will go to `AddPartialValueToExistingCounter()` in the first iteration and all the next iterations will modify the data in place.

ayende: Do this, first read the `localDbIds` to a List, then scan this.
Add to this list if needed with new values.

Same logic, simpler behavior. There is no need to do much here, we expect 3 - 5 max.

src/Raven.Server/Documents/CountersStorage.cs
maximburyak: I agree that there could be small difference, but why will the linear scan be better?

src/Raven.Server/Documents/CountersStorage.cs
maximburyak: For small sizes (~5) dictionary gives about 10% performance improvement in searching, while we could cache/pool the dictionary objects for memory performance.

src/Raven.Server/Documents/CountersStorage.cs
maximburyak: the list is "cleare" at the end of the usage - this is only a mean of reusing the list between requests

maximburyak: We could just use a finally statment, in which we'll call the disposes and .clear. In wost case finalizers will catch that..
In the bottom line, all usages of the _scopes are performed inside the PutCounters function

src/Raven.Server/Documents/CountersStorage.cs
ayende: Linear scan for < 5 items is going to always be cheaper than the setup cost of anything else

src/Raven.Server/Documents/CountersStorage.cs
ayende: I know, but it open us the possibility for missing it later on. In general, we do caching and things like that at the context level

src/Raven.Server/Documents/CountersStorage.cs
aviv86: yes `DbIds` already has `@` prefix. we can avoid a counter of this name, but maybe better to wrap all counters in a nested object and call that `@Counters` property ?

src/Raven.Server/Documents/CountersStorage.cs
aviv86: We can do that, but a couple of things to consider here:
- when deleting all counters in a group, we're left with a table entry that's all `null`s
- if a counter was deleted and then created again, how would we deal with that in replication? 
 probably need to keep a deleted etag to know which information is more up do date.

src/Raven.Server/Documents/CountersStorage.cs
ayende: Yes, but let's use shorter name? We care about space.

src/Raven.Server/Documents/CountersStorage.cs
ayende: Not sure, right now we are explicit that delete + modify on counter is something that user need to verify won't happen and might cause the modification to be lost or the delete to be lost

src/Raven.Server/Documents/CountersStorage.cs
maximburyak: Yes, sounds good

src/Raven.Client/Documents/Operations/Counters/CountersDetail.cs
aviv86: done

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs
aviv86: removed the dictionary

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs
aviv86: I removed this function

src/Raven.Server/Documents/CountersStorage.cs
aviv86: removed this function

src/Raven.Client/Exceptions/Documents/Counters/CounterOverflowException.cs
aviv86: done

src/Raven.Client/Exceptions/Documents/Counters/CounterOverflowException.cs
aviv86: done

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/Handlers/CountersHandler.cs
aviv86: done

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs
aviv86: I'm using `@dbIds` and `@vals`. okay?

src/Raven.Server/Documents/CountersStorage.cs
aviv86: removed CounterTombstones.
As discussed yesterday, deleted entries need to keep a change vector string

src/Raven.Server/Documents/Handlers/CountersHandler.cs
aviv86: done

src/Raven.Server/Smuggler/Documents/StreamDestination.cs
aviv86: done

src/Raven.Server/Documents/Handlers/CountersHandler.cs
aviv86: fixed

src/Raven.Client/Documents/Operations/Counters/CounterBatch.cs
```diff
@@ -146,4 +146,10 @@ public DynamicJsonValue ToJson()
             };
         }
     }
+
+    public class CounterGroup
```
ayende: Why do we need this?

src/Raven.Client/Documents/Smuggler/DatabaseItemType.cs
```diff
@@ -21,6 +21,8 @@ public enum DatabaseItemType
         Unknown = 1 << 12,
 
         Counters = 1 << 13,
```
ayende, line 24: Mark as `[Obsolete]`

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -163,162 +173,598 @@ public long GetNumberOfCountersToProcess(DocumentsOperationContext context, stri
...
+                }
+
+
+                var newETag = _documentsStorage.GenerateNextEtag();
+
+                if (data != null)
+                {
+                    var dbIdIndex = GetDbIdIndex(context, data);
+
+                    data.TryGet(Values, out BlittableJsonReaderObject counters);
```
ayende: Validate that this worked

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -163,162 +173,598 @@ public long GetNumberOfCountersToProcess(DocumentsOperationContext context, stri
...
+
+
+                    if (counters.TryGetMember(name, out object existingCounter) == false || 
+                        existingCounter is LazyStringValue ||
+                        !(existingCounter is BlittableJsonReaderObject.RawBlob blob) ||
```
ayende: Remove this line

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -163,162 +173,598 @@ public long GetNumberOfCountersToProcess(DocumentsOperationContext context, stri
...
-            Debug.Assert(size > DbIdAsBase64Size + 2 /* record separators */);
-            int sizeOfDocId = 0;
-            for (; sizeOfDocId < size; sizeOfDocId++)
+            var dbIdIndex = int.MaxValue;
+            if (data.TryGet(DbIds, out BlittableJsonReaderArray dbIds))
             {
-                if (p[sizeOfDocId] == SpecialChars.RecordSeparator)
+                for (dbIdIndex = 0; dbIdIndex < dbIds.Length; dbIdIndex++)
+                {
+                    if (dbIds[dbIdIndex].Equals(context.Environment.Base64Id) == false)
```
ayende: Reverse this

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -163,162 +173,598 @@ public long GetNumberOfCountersToProcess(DocumentsOperationContext context, stri
...
-            var size = counterTombstoneId.Size;
-
-            int sizeOfDocId = 0;
-            for (; sizeOfDocId < size; sizeOfDocId++)
+            using (context.Allocator.Allocate((dbIdIndex + 1) * SizeOfCounterValues, out var newVal))
             {
-                if (p[sizeOfDocId] == SpecialChars.RecordSeparator)
-                    break;
-            }
+                if (dbIdIndex > 0)
```
ayende: Remove this check

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -163,162 +173,598 @@ public long GetNumberOfCountersToProcess(DocumentsOperationContext context, stri
...
+                }
+            }
+            else
+            {
+                // counter exists , but not with local DbId
 
-        public void PutCounter(DocumentsOperationContext context, string documentId, string collection, string name, long value)
-        {
-            PutCounterImpl(context, documentId, collection, name, null, value);
+                using (AddPartialValueToExistingCounter(context, existingCounter, dbIdIndex, value, newETag))
```
ayende: This isn't valid, we are disposing the value while still using `existingCounter`

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -163,162 +173,598 @@ public long GetNumberOfCountersToProcess(DocumentsOperationContext context, stri
...
                         {
                             var existingChangeVector = TableValueToChangeVector(context, (int)CountersTable.ChangeVector, ref existing);
 
                             if (ChangeVectorUtils.GetConflictStatus(changeVector, existingChangeVector) == ConflictStatus.AlreadyMerged)
                                 return;
+
+                            data = GetData(context, ref existing);
+                            data = data.Clone(context);
+                            data.TryGet(DbIds, out BlittableJsonReaderArray dbIds);
+                            data.TryGet(Values, out BlittableJsonReaderObject localCounters);
```
ayende: Need to check these

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -163,162 +173,598 @@ public long GetNumberOfCountersToProcess(DocumentsOperationContext context, stri
...
+
+                            var localDbIdsList = DbIdsToList(dbIds);
+
+                            sourceData.TryGet(DbIds, out BlittableJsonReaderArray sourceDbIds);
```
ayende: check this

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -163,162 +173,598 @@ public long GetNumberOfCountersToProcess(DocumentsOperationContext context, stri
...
+                                        localCounters.Modifications = localCounters.Modifications ?? new DynamicJsonValue(localCounters);
+                                        localCounters.Modifications[counterName] = mergedCv;
+                                        continue;
+
+                                    }
+                                    if (loaclCounterValues != null)
+                                    {
+                                        // delete + blob => resolve conflict
+
+                                        using (var localValuesAsCv = RawBlobToChangeVector(context, data, loaclCounterValues, existingChangeVector))
```
ayende: You don't need this. Can do the check directly on the change vector by asking to read by id for each of the existing db ids there

ayende: for `loaclCounterValues`, iterate over the internal values, for each non zero etag, go get the db id and then check just that.
You only care if there are all after the delete or no, nothing else

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -163,162 +173,598 @@ public long GetNumberOfCountersToProcess(DocumentsOperationContext context, stri
...
+                                        }                                       
+                                    }
+                                    else
+                                    {
+                                        // put deleted counter
+
+                                        modified = true;
+                                        changeType = CounterChangeTypes.Delete;
+                                        localCounters.Modifications = localCounters.Modifications ?? new DynamicJsonValue(localCounters);
+                                        // todo add local dbId + etag ?
```
ayende: No need, the original is sufficent.

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -163,162 +173,598 @@ public long GetNumberOfCountersToProcess(DocumentsOperationContext context, stri
...
+                                        changeType = CounterChangeTypes.Delete;
+                                        localCounters.Modifications = localCounters.Modifications ?? new DynamicJsonValue(localCounters);
+                                        // todo add local dbId + etag ? 
+                                        localCounters.Modifications[counterName] = deletedSourceCounter;
+                                        _countersCount--;
+                                    }
+                                }
+
+                                if (prop.Value is BlittableJsonReaderObject.RawBlob sourceBlob)
+                                {
```
ayende, line 608: Extract this to a function that operate over delete cv and counter value.
Should't matter which direction

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -163,162 +173,598 @@ public long GetNumberOfCountersToProcess(DocumentsOperationContext context, stri
...
+            {
+                localDbIdsList.Add((LazyStringValue)dbIds[i]);
+            }
 
-                    using (Slice.From(context.Allocator, changeVector, out var cv))
-                    using (DocumentIdWorker.GetStringPreserveCase(context, collectionName.Name, out Slice collectionSlice))
+            return localDbIdsList;
+        }
+
+        private long InternalPutCounter(DocumentsOperationContext context, BlittableJsonReaderObject counters, string counterName,
```
ayende: `localDbIds` - extract to a struct that contains both the array and the list in one shot. And can apply operations to them

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -163,162 +173,598 @@ public long GetNumberOfCountersToProcess(DocumentsOperationContext context, stri
...
+        {
+            long value = 0;
+            var existingCount = existingCounter.Length / SizeOfCounterValues;
+            var sourceCount = source.Length / SizeOfCounterValues;
+            modified = false;
+
+            for (var index = 0; index < sourceCount; index++)
+            {
+                var sourceDbId = (LazyStringValue)sourceDbIds[index];
+                var sourceValue = ((CounterValues*)source.Ptr)[index];
```
ayende: Use pointer to avoid copying this

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -163,162 +173,598 @@ public long GetNumberOfCountersToProcess(DocumentsOperationContext context, stri
...
-                    using (Slice.From(context.Allocator, changeVector, out var cv))
-                    using (DocumentIdWorker.GetStringPreserveCase(context, collectionName.Name, out Slice collectionSlice))
+            return localDbIdsList;
+        }
+
+        private long InternalPutCounter(DocumentsOperationContext context, BlittableJsonReaderObject counters, string counterName,
+            BlittableJsonReaderArray localDbIds, BlittableJsonReaderArray sourceDbIds, List<LazyStringValue> localDbIdsList, 
+            BlittableJsonReaderObject.RawBlob existingCounter, BlittableJsonReaderObject.RawBlob source, out bool modified)
+        {
+            long value = 0;
```
ayende: Drop this, sum it up in the end which is easier

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -163,162 +173,598 @@ public long GetNumberOfCountersToProcess(DocumentsOperationContext context, stri
...
+                var sourceDbId = (LazyStringValue)sourceDbIds[index];
+                var sourceValue = ((CounterValues*)source.Ptr)[index];
+
+                int localDbIdIndex = GetOrAddDbIdIndex(localDbIds, localDbIdsList, sourceDbId);
+
+                if (localDbIdIndex < existingCount)
+                {
+                    var localValuePtr = (CounterValues*)existingCounter.Ptr + localDbIdIndex;
+                    if (localValuePtr->Etag >= sourceValue.Etag ||
+                        sourceDbId.Equals(context.Environment.Base64Id))
```
ayende: Remove the `sourceDbId` clause

src/Raven.Server/Documents/CountersStorage.cs
ayende: Should return `bool` here

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -163,162 +173,598 @@ public long GetNumberOfCountersToProcess(DocumentsOperationContext context, stri
...
+                BlittableJsonReaderObject data = null;
+                sourceData.TryGet(Values, out BlittableJsonReaderObject sourceCounters);
+
+                using (DocumentIdWorker.GetLowerIdSliceAndStorageKey(context, documentId, out Slice counterKey, out _))
                 {
-                    if (changeVector != null)
+                    using (table.Allocate(out TableValueBuilder tvb))
                     {
-                        if (table.ReadByKey(counterKey, out var existing))
+                        if (changeVector != null && table.ReadByKey(counterKey, out var existing))
```
ayende:
```suggestion
                        if ( table.ReadByKey(counterKey, out var existing))
```

Can't use the CV here

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -163,162 +173,598 @@ public long GetNumberOfCountersToProcess(DocumentsOperationContext context, stri
...
+            if (modified)
+            {
+                counters.Modifications = counters.Modifications ?? new DynamicJsonValue(counters);
+                counters.Modifications[counterName] = existingCounter;
             }
+
+            return value;
+        }
+
+        private static int GetOrAddDbIdIndex(BlittableJsonReaderArray localDbIds, List<LazyStringValue> localDbIdsList, LazyStringValue dbId)
```
ayende: Move this to dedicated struct

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -375,219 +809,116 @@ public Table GetCountersTable(Transaction tx, CollectionName collection)
...
-                }
-
-                RemoveTombstoneIfExists(context, documentId, name);
+                if (table.ReadByKey(key, out var existing) == false)
+                    yield break;
 
-                var etag = _documentsStorage.GenerateNextEtag();
-                var result = ChangeVectorUtils.TryUpdateChangeVector(_documentDatabase.ServerStore.NodeTag, _documentsStorage.Environment.Base64Id, etag, string.Empty);
+                var data = GetData(context, ref existing);
+                data.TryGet(Values, out BlittableJsonReaderObject counters);
```
ayende: Paranoia, please

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -375,219 +809,116 @@ public Table GetCountersTable(Transaction tx, CollectionName collection)
...
-                    Type = exists ? CounterChangeTypes.Increment : CounterChangeTypes.Put,
-                    Value = value
-                });
-
-                return result.ChangeVector;
             }
         }
 
-        public IEnumerable<string> GetCountersForDocument(DocumentsOperationContext context, string docId)
+        private static BlittableJsonReaderObject GetData(JsonOperationContext context, ref TableValueReader existing)
```
ayende:
```suggestion
        private static BlittableJsonReaderObject GetCounterValuesData(JsonOperationContext context, ref TableValueReader existing)
```

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -375,219 +809,116 @@ public Table GetCountersTable(Transaction tx, CollectionName collection)
...
-                    yield return val;
-                }
+                if (table.ReadByKey(key, out var tvr) == false)
+                    return false;
+
+                var data = GetData(context, ref tvr);
+                if (data.TryGet(Values, out BlittableJsonReaderObject counters) == false ||
+                    counters.TryGetMember(counterName, out object counterValues) == false ||
+                    counterValues is LazyStringValue || // deleted
+                    !(counterValues is BlittableJsonReaderObject.RawBlob counterValuesAsRawBlob))
```
ayende: Remove this line

ayende:
```suggestion
  
```

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -375,219 +809,116 @@ public Table GetCountersTable(Transaction tx, CollectionName collection)
...
-        }
-
-        private static LazyStringValue ExtractCounterName(JsonOperationContext context, TableValueReader tvr)
-        {
-            return TableValueToId(context, (int)CountersTable.Name, ref tvr);
-
-        }
-
-        public ByteStringContext.InternalScope GetCounterKey(DocumentsOperationContext context, string documentId, string name, string changeVector, out Slice partialKeySlice)
+        public IEnumerable<(string ChangeVector, long Value)> GetCounterValues(DocumentsOperationContext context, string docId, string counterName)
```
ayende, line 896: Check where this is used. This should only be returned as a debug info / specific call. 
Shouldn't be in general use

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -375,219 +809,116 @@ public Table GetCountersTable(Transaction tx, CollectionName collection)
...
-                byte* dest = buffer.Ptr + docIdLower.Size + 1;
-                nameLower.CopyTo(dest);
-                dest[nameLower.Size] = SpecialChars.RecordSeparator;
-                cv.CopyTo(cv.Size - DbIdAsBase64Size, dest, nameLower.Size + 1, DbIdAsBase64Size);
+                var data = GetData(context, ref tvr);
+                if (data.TryGet(DbIds, out BlittableJsonReaderArray dbIds) == false ||
+                    data.TryGet(Values, out BlittableJsonReaderObject counters) == false ||
+                    counters.TryGetMember(counterName, out object counterValues) == false ||
+                    counterValues is LazyStringValue || // deleted
+                    !(counterValues is BlittableJsonReaderObject.RawBlob blob))
```
ayende:
```suggestion
```

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -602,89 +933,170 @@ public void DeleteCountersForDocument(DocumentsOperationContext context, string
             if (table.NumberOfEntries == 0)
                 return;
 
-            using (GetCounterPartialKey(context, documentId, out var keyPrefix))
+            using (DocumentIdWorker.GetLowerIdSliceAndStorageKey(context, documentId, out Slice lowerId, out _))
             {
-                table.DeleteByPrimaryKeyPrefix(keyPrefix);
+                //todo update Count of Counters
+                table.DeleteByKey(lowerId /*, holder => _countersCount -= GetData(context, ref holder.Reader).Count -1*/);
```
ayende: Remove this, we won't track the number of counters for now

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -602,89 +933,170 @@ public void DeleteCountersForDocument(DocumentsOperationContext context, string
...
+                    return null;
+                
+                var data = GetData(context, ref existing);
+
+                if (data.TryGet(Values, out BlittableJsonReaderObject counters) == false ||
+                    counters.TryGetMember(counterName, out object counterToDelete) == false)                
+                    return null;
+
+                if (counterToDelete is LazyStringValue || // already deleted
+                    !(counterToDelete is BlittableJsonReaderObject.RawBlob blob))
```
ayende:
```suggestion
```

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -602,89 +933,170 @@ public void DeleteCountersForDocument(DocumentsOperationContext context, string
...
+                _countersCount--;
+
+                return cv;
+
+
             }
         }
 
-        public string DeleteCounter(DocumentsOperationContext context, Slice key, string collection, long lastModifiedTicks, bool forceTombstone)
+        private void RemoveCounterFromBlittableAndUpdateTable(DocumentsOperationContext context, string counterName, BlittableJsonReaderObject data,
```
ayende: See if you can trim the number of params here

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -602,89 +933,170 @@ public void DeleteCountersForDocument(DocumentsOperationContext context, string
...
+                tvb.Add(collectionSlice);
+                tvb.Add(context.TransactionMarkerOffset);
+
+                table.Set(tvb);
             }
+        }
 
-            var newEtag = _documentsStorage.GenerateNextEtag();
-            _documentsStorage.EnsureLastEtagIsPersisted(context, newEtag);
+        private static LazyStringValue RawBlobToChangeVector(DocumentsOperationContext context, BlittableJsonReaderObject data, BlittableJsonReaderObject.RawBlob counterToDelete, string changeVector,
```
ayende: drop this

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -602,89 +933,170 @@ public void DeleteCountersForDocument(DocumentsOperationContext context, string
...
+                        return new LazyStringValue(null, cvMem.Address, sb.Length, context);*/
+
+            var mem = context.GetMemory(sb.Length);
+            return new LazyStringValue(sb.ToString(), mem.Address, sb.Length, context);
 
-            return newChangeVector;
         }
 
-        private void CreateTombstone(DocumentsOperationContext context, Slice keySlice, string collectionName, long deletedEtag, long lastModifiedTicks, long newEtag, string newChangeVector)
+        private static void AddLocalNodeToChangeVector(DocumentsOperationContext context, StringBuilder sb)
```
ayende: drop this

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -797,5 +1209,38 @@ private static SortedSet<string> GetCountersForDocument(BlittableJsonReaderArray
...
+        {
+            counterGroupDetail.Values.TryGet(Values, out BlittableJsonReaderObject counters);
+            counters.Modifications = new DynamicJsonValue(counters);
+
+            var prop = new BlittableJsonReaderObject.PropertyDetails();
+
+            for (int i = 0; i < counters.Count; i++)
+            {
+                counters.GetPropertyByIndex(i, ref prop);
+
```
ayende, line 1207: Need to handle tombstones here

src/Raven.Server/Documents/Handlers/StatsHandler.cs
```diff
@@ -67,7 +67,7 @@ private void FillDatabaseStatistics(DatabaseStatistics stats, DocumentsOperation
                 stats.CountOfConflicts = Database.DocumentsStorage.ConflictsStorage.ConflictsCount;
                 stats.SizeOnDisk = size.Data;
                 stats.NumberOfTransactionMergerQueueOperations = Database.TxMerger.NumberOfQueuedOperations;
-                stats.CountOfCounters = Database.DocumentsStorage.CountersStorage.GetNumberOfCounterEntries(context);
+                stats.CountOfCounters = Database.DocumentsStorage.CountersStorage.GetNumberOfCounterEntries();
```
ayende: Remove `CountOfCounters`

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -840,21 +840,10 @@ private unsafe void ReadItemsFromSource(ref UnmanagedWriteBuffer writeBuffer, in
...
-                    item.KeyDispose = Slice.From(context.Allocator, ReadExactly(keySize), keySize, out item.Key);
-
-                    var collectionSize = *(int*)ReadExactly(sizeof(int));
-                    Debug.Assert(collectionSize > 0);
-                    item.Collection = Encoding.UTF8.GetString(ReadExactly(collectionSize), collectionSize);
+                    var sizeOfData = *(int*)ReadExactly(sizeof(int));
+                    item.DocumentSize = sizeOfData;
 
-                    item.LastModifiedTicks = *(long*)ReadExactly(sizeof(long));
+                    ReadExactly(sizeOfData, ref writeBuffer);
```
ayende, line 849: Need to handle 4.1 -> 4.2 replication counter (error, abort connection)
4.2 -> 4.1 want to replicate (error, abort connection)

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
```diff
@@ -926,6 +926,58 @@ private unsafe void WriteCounterToServer(DocumentsOperationContext context, Repl
...
+                                       sizeof(int) + // size of doc collection
+                                       item.Collection.Size + // doc collection
+                                       sizeof(int) // size of data
+                                       + item.Values.Size; // data
+
+                    if (requiredSize > _tempBuffer.Length)
+                        ThrowTooManyChangeVectorEntries(item);
+
+                    var tempBufferPos = 0;
+                    pTemp[tempBufferPos++] = (byte)item.Type;
```
ayende, line 943: ensure the value is different from the value in 41

ayende (changes requested):

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -312,34 +308,37 @@ public string PutCounter(DocumentsOperationContext context, string documentId, s
...
-        private static int GetDbIdIndex(DocumentsOperationContext context, BlittableJsonReaderObject data)
+        private int GetDbIdIndex(BlittableJsonReaderObject data)
         {
             var dbIdIndex = int.MaxValue;
             if (data.TryGet(DbIds, out BlittableJsonReaderArray dbIds))
             {
                 for (dbIdIndex = 0; dbIdIndex < dbIds.Length; dbIdIndex++)
                 {
-                    if (dbIds[dbIdIndex].Equals(context.Environment.Base64Id) == false)
+                    if (dbIds[dbIdIndex].Equals(_localNodeTagAndDbId) == false)
```
ayende: We should hold only the db id here, not the node tag

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -417,7 +412,7 @@ private BlittableJsonReaderObject WriteNewCountersDocument(DocumentsOperationCon
                     builder.WritePropertyName(DbIds);
 
                     builder.StartWriteArray();
-                    builder.WriteValue(context.Environment.Base64Id);
+                    builder.WriteValue(_localNodeTagAndDbId);
```
ayende: just db id here. The node tag may change over time, and we don't want to mess things up

src/Raven.Server/Documents/Handlers/StatsHandler.cs
maximburyak: @ayende can we have the number of "counter entries" here (instead of the percise counters count) this way we'll have an indication of the order of magnitude of counters. It's the same as tracking the attachment count - there could be 1kb attachment and 10mb attachment, we still treat the equally in it's count field

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -1051,10 +1034,10 @@ public string DeleteCounter(DocumentsOperationContext context, string documentId
...
 
 
-            if (count < dbIdIndex && addLocal)
+/*            if (count < dbIdIndex && addLocal)
             {
                 AddLocalNodeToChangeVector(context, sb);
-            }
+            }*/
 
             /*            var size = Encoding.UTF8.GetMaxByteCount(sb.Length);
```
maximburyak: let's try making this work

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -1242,5 +1220,74 @@ public static void ConvertFromBlobToNumbers(JsonOperationContext context, Counte
...
+
+            var sizeOfValues = counterValues.Length / SizeOfCounterValues;
+
+            int numOfMatches = 0;
+            for (int i = 0; i < sizeOfValues; i++)
+            {
+                bool found = false;
+
+                var current = (CounterValues*)counterValues.Ptr + i;
+                var dbId = dbIds[i].ToString().Substring(2);
```
maximburyak: let's find a way to do that without materialization

src/Raven.Server/Documents/Handlers/StatsHandler.cs
ayende: Fine by me, yes, but needs to be carefully explained.

src/Raven.Client/Documents/Smuggler/DatabaseItemType.cs
aviv86: done

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs
aviv86: done

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs
aviv86: done

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs
aviv86: fixed

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs
aviv86: fixed

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Client/Documents/Operations/Counters/CounterBatch.cs
aviv86: removed this. using `CounterGroupDetail` instead

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
aviv86: done

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
aviv86: done

src/Raven.Server/Documents/CountersStorage.cs

aviv86 (commented):
@arekpalinski please review ETL part

arekpalinski:
@aviv86 I'll do it tomorrow morning

test/Tests.Infrastructure/RavenTestBase.cs
```diff
@@ -603,7 +603,7 @@ protected IDisposable RestoreDatabase(IDocumentStore store, RestoreBackupConfigu
             var restoreOperation = new RestoreBackupOperation(config);
 
             var operation = store.Maintenance.Server.Send(restoreOperation);
-            operation.WaitForCompletion(timeout ?? TimeSpan.FromSeconds(30));
+            operation.WaitForCompletion(timeout ?? TimeSpan.FromSeconds(55530)); //todo revert
```
arekpalinski, line 606: TODO to revert timeout

ayende:
@aviv86 The PR is already pretty big with a lot of comments. Can you create a new clean one so we can review the final state?

PR #8419, opened 2019-01-06
fix for RavenDB-12397

myarichuk:


myarichuk:
Added a test to verify that RavenDB-11671 is resolved by changes in RavenDB-12397

ppekrol (approved):

test/SlowTests/Issues/RavenDB-12397.cs
```diff
@@ -0,0 +1,153 @@
...
+using FastTests;
+using Newtonsoft.Json.Linq;
+using Raven.Client.Documents;
+using Xunit;
+
+namespace SlowTests.Issues
+{
+    public class RavenDB_12397 : RavenTestBase
+    {
+        public class Beer
```
ppekrol: @myarichuk Try to use `private class` next time

myarichuk: ok

src/Raven.Server/Documents/Queries/Parser/QueryParser.cs
```diff
@@ -875,63 +875,83 @@ private static QueryExpression FinalProcessingOfMatchExpression(List<MatchPath>
...
             var clauses = new List<PatternMatchElementExpression>();
-            bool reversed = false;
-            for (int i = list.Count - 1; i >= 1; i--)
+
+            //the last path element cannot be a "cross-road"
+            while (matchPaths.Count > 1)
             {
-                if (list[i].EdgeType != list[i - 1].EdgeType)
+                var hasFoundJunction = false;
+                for (var i = 0; i < matchPaths.Count - 2; i++)
```
ayende, line 913: Why -2 here?

myarichuk: The last item in the path cannot be a "junction"
Also, since we are dealing with current and next items and compare between their directions to detect "junctions", there is no need to handle the last one

ayende: Can you add an assert or failure to check that it can't be, just to be safe?

myarichuk: Sure, I will add more precautions to that area of code.

test/SlowTests/Issues/RavenDB-12397.cs
myarichuk: changed it in commit after code review

myarichuk:
added a commit with changes related to the code review

ppekrol (approved):

ayende (approved):

PR #8422, opened 2019-01-07
RavenDB-12376: RvnMemoryMapPager - Posix 64Bit Part

aviviadi:


src/Voron/Impl/Paging/RvnMemoryMapPager.cs
```diff
@@ -0,0 +1,214 @@
...
+
+        public RvnMemoryMapPager(StorageEnvironmentOptions options, VoronPathSetting file, long? initialFileSize = null, bool canPrefetchAhead = true, bool usePageProtection = false)
+            : base(options, canPrefetchAhead, usePageProtection)
+        {
+            FileName = file;
+            _copyOnWriteMode = options.CopyOnWriteMode && FileName.FullPath.EndsWith(Constants.DatabaseFilename);
+            _logger = LoggingSource.Instance.GetLogger<StorageEnvironment>($"Pager-{file}");
+
+            if (initialFileSize.HasValue == false || initialFileSize.Value == 0) 
+                initialFileSize = SysInfo.PageSize * 16;
```
ayende: Validate here that the `initial page size` is a multiple of the page size and at least 64KB

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
```diff
@@ -0,0 +1,214 @@
...
+            var pager = new PagerState(this, Options.PrefetchSegmentSize, Options.PrefetchResetThreshold, allocationInfo);
+            SetPagerState(pager);
+        }
+
+        public override byte* AcquirePagePointer(IPagerLevelTransactionState tx, long pageNumber, PagerState pagerState = null)
+        {
+            var state = pagerState ?? _pagerState;
+
+            // ReSharper disable once InvertIf
+            if (SysInfo.CanPrefetch == PalDefinitions.True)
```
ayende: Why isn't this a bool?

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
```diff
@@ -0,0 +1,214 @@
...
+                out _handle,
+                out var baseAddress,
+                out _totalAllocationSize,
+                out var errorCode);
+
+            if (rc != 0)
+            {
+                try
+                {
+                    PalHelper.ThrowLastError(errorCode, $"rvn_create_and_mmap64_file failed on {(FAIL_CODES)rc}");
```
ayende: `FAIL_CODES` -> `PalFailCodes`, or something like that

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
```diff
@@ -0,0 +1,214 @@
...
+
+            if (initialFileSize.HasValue == false || initialFileSize.Value == 0) 
+                initialFileSize = SysInfo.PageSize * 16;
+
+            Debug.Assert(file != null);
+
+            var rc = rvn_create_and_mmap64_file(
+                file.FullPath,
+                initialFileSize.Value,
+                _copyOnWriteMode ? MMAP_OPTIONS.CopyOnWrite : MMAP_OPTIONS.None,
```
ayende: `MMAP_OPTIONS` -> `PalMmapFlags`

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
```diff
@@ -0,0 +1,214 @@
...
+
+        public override byte* AcquirePagePointer(IPagerLevelTransactionState tx, long pageNumber, PagerState pagerState = null)
+        {
+            var state = pagerState ?? _pagerState;
+
+            // ReSharper disable once InvertIf
+            if (SysInfo.CanPrefetch == PalDefinitions.True)
+            {
+                // ReSharper disable once UnusedVariable
+                if (_pagerState.ShouldPrefetchSegment(pageNumber, out void* virtualAddress, out long bytes))
```
ayende: `out _` instead of unused variable

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
```diff
@@ -0,0 +1,214 @@
...
+        {
+            var currentState = GetPagerStateAndAddRefAtomically();
+            try
+            {
+                using (var metric = Options.IoMetrics.MeterIoRate(FileName.FullPath, IoMetrics.MeterType.DataSync, 0))
+                {
+                    foreach (var alloc in currentState.AllocationInfos)
+                    {
+                        metric.IncrementFileSize(alloc.Size);
+                        if (rvn_memory_sync(alloc.BaseAddress, alloc.Size, MSYNC_OPTIONS.MS_SYNC, out var errorCode) != 0)
```
ayende: `MSYNC_OPTIONS` -> `MemorySyncFlags`

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
```diff
@@ -0,0 +1,214 @@
...
+                currentState.Release();
+            }
+        }
+
+        public override void ReleaseAllocationInfo(byte* baseAddress, long size)
+        {
+            base.ReleaseAllocationInfo(baseAddress, size);
+
+            var rc = rvn_unmap(baseAddress, size, DeleteOnClose ? PalDefinitions.True : PalDefinitions.False, out var errorCodeUnmap, out var errorCodeMadvise);
+            // TODO : is madvise don't need - should fail unmap call ? (currently we fail it)
```
ayende: I don't understand this TODO

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
```diff
@@ -0,0 +1,214 @@
...
+            if (initialFileSize.HasValue == false || initialFileSize.Value == 0) 
+                initialFileSize = SysInfo.PageSize * 16;
+
+            Debug.Assert(file != null);
+
+            var rc = rvn_create_and_mmap64_file(
+                file.FullPath,
+                initialFileSize.Value,
+                _copyOnWriteMode ? MMAP_OPTIONS.CopyOnWrite : MMAP_OPTIONS.None,
+                out _handle,
```
ayende, line 48: We need to implement a `SafeHandle` here

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
```diff
@@ -0,0 +1,214 @@
...
+            if (rc != 0)
+                PalHelper.ThrowLastError(errorCodeUnmap != 0 ? errorCodeUnmap : errorCodeMadvise,
+                    $"Failed to unmap {FileName.FullPath}. rc={rc}. DeleteOnClose={DeleteOnClose}, errorCodeUnmap={errorCodeUnmap}, errorCodeMadvise={errorCodeMadvise}");
+
+            NativeMemory.UnregisterFileMapping(FileName?.FullPath, new IntPtr(baseAddress), size);
+        }
+
+        protected override void DisposeInternal()
+        {
+            var rc = rvn_dispose_handle(FileName.FullPath, _handle, DeleteOnClose ? PalDefinitions.True : PalDefinitions.False, out var errorCodeClose, out var errorCodeUnlink);
```
ayende: Let do that through the `SafeHandle`, just make sure that we aren't throwing on dispose

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
```diff
@@ -0,0 +1,214 @@
...
+            // ignore results.. but warn! we cannot delete the file probably
+            // rc contains either FAIL_UNLINK, FAIL_CLOSE or both of them   
+            if (_logger.IsInfoEnabled)
+                _logger.Info($"Unable to dispose handle for {FileName.FullPath} (ignoring). rc={rc}. DeleteOnClose={DeleteOnClose}, errorCodeClose={errorCodeClose}, errorCodeUnlink={errorCodeUnlink}",
+                    new IOException($"Unable to dispose handle for {FileName.FullPath} (ignoring)."));
+        }
+
+        protected internal override void PrefetchRanges(Win32MemoryMapNativeMethods.WIN32_MEMORY_RANGE_ENTRY* list, int count)
+        {
+            // TODO : Get rid of WIN32_MEMORY_RANGE_ENTRY and use Pal's PrefetchRanges instead
```
ayende, line 144: Yes, very much so

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
```diff
@@ -0,0 +1,214 @@
...
+            // ReSharper disable once UnusedVariable
+            rvn_prefetch_ranges((PalDefinitions.PrefetchRanges*)list, count, out var errorCode);
```
ayende: `out _`

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
```diff
@@ -0,0 +1,214 @@
...
+
+        protected internal override PagerState AllocateMorePages(long newLength)
+        {
+            if (DisposeOnceRunner.Disposed)
+                ThrowAlreadyDisposedException();
+
+            var rc = rvn_allocate_more_space(newLength, _totalAllocationSize, FileName.FullPath, _handle, _copyOnWriteMode ? MMAP_OPTIONS.CopyOnWrite : MMAP_OPTIONS.None,
+                out var newAddress, out var newLengthAfterAdjustment, out var errorCode);
+
+            if (rc == (int)FAIL_CODES.FAIL_ALLOCATION_NO_RESIZE)
```
ayende: Why is this not an error?

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
```diff
@@ -0,0 +1,214 @@
...
+                return null;
+
+            if (rc != 0)
+            {
+                // TODO : can't allocate more pages, should we force throw out of mem (as was done before RvnMemoryMapPager introduced ?
```
ayende: This depends on what the failure was.
Did we fail to allocate more disk space ( `DiskFull` or fail to actually mmap the memory `OOM`)

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
```diff
@@ -0,0 +1,214 @@
...
+            return newPagerState;
+        }
+
+        internal override void ProtectPageRange(byte* start, ulong size, bool force = false)
+        {
+            if (size == 0 || 
+                UsePageProtection == false && force == false) 
+                return;
+
+            if (rvn_protect_range(start, (long)size, MPROTECT_OPTIONS.PROT_READ, out var errorCode) == 0)
```
ayende: `MPROTECT_OPTIONS` -> `MemoryProtectionFlags`

src/Voron/Platform/Pal.cs
```diff
@@ -1,24 +1,22 @@
...
-using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Runtime.InteropServices;
-using System.Text;
-using Sparrow.Platform;
-
-// ReSharper disable SwitchStatementMissingSomeCases
-// ReSharper disable InconsistentNaming
+using static Voron.Platform.PalDefinitions;
+// ReSharper disable BuiltInTypeReferenceStyle
```
ayende: ??

src/Voron/Platform/Pal.cs
```diff
@@ -75,7 +73,74 @@ static Pal()
...
-            out Int32 specialErrnoCodes
-            );
+            out Int32 specialErrnoCodes);
+
+        [DllImport(LIBRVNPAL, SetLastError = true)]
+        public static extern Int32 rvn_create_and_mmap64_file(
+            string filename,
+            Int64 initialFileSize,
+            PalFlags.MMAP_OPTIONS flags,
+            out void* handle,
```
ayende, line 82: Need to write our own `SafeHandle` impl to take care of this.

src/Voron/Platform/Pal.cs
```diff
@@ -75,7 +73,74 @@ static Pal()
...
+            Int64 size,
+            PalFlags.MSYNC_OPTIONS flags,
+            out Int32 errorCode);
+
+        [DllImport(LIBRVNPAL, SetLastError = true)]
+        public static extern Int32 rvn_dispose_handle(
+            string filepath,
+            void* handle,
+            Int32 deleteOnClose,
+            out Int32 errorCodeClose,
```
ayende: Only one error code should be returned, you can check the return value to check what was the actual error.

src/Voron/Platform/Pal.cs
```diff
@@ -75,7 +73,74 @@ static Pal()
...
+            out Int32 errorCodeUnlink);
+
+        [DllImport(LIBRVNPAL, SetLastError = true)]
+        public static extern Int32 rvn_unmap(
+            void* address,
+            Int64 size,
+            Int32 deleteOnClose,
+            out Int32 errorCodeUnmap,
```
ayende: Only one error code should be returned, you can check the return value to check what was the actual error.

src/Voron/Platform/PalDefinitions.cs
```diff
@@ -0,0 +1,26 @@
+using System;
+using System.Diagnostics.CodeAnalysis;
+// ReSharper disable BuiltInTypeReferenceStyle
+
+namespace Voron.Platform
+{
+    public unsafe class PalDefinitions
+    {
+        public static Int32 True => 1;
+        public static Int32 False => 0;
```
ayende: Kill these two, there is no such thing, we use `bool`

src/Voron/Platform/PalDefinitions.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        [SuppressMessage("ReSharper", "UnassignedField.Global")]
+        public struct SystemInformation
+        {
+            public Int32 PageSize;
+            public Int32 CanPrefetch;
```
ayende:
```suggestion
            public bool CanPrefetch;
```

src/Voron/Platform/PalDefinitions.cs
```diff
@@ -0,0 +1,26 @@
...
+// ReSharper disable BuiltInTypeReferenceStyle
+
+namespace Voron.Platform
+{
+    public unsafe class PalDefinitions
+    {
+        public static Int32 True => 1;
+        public static Int32 False => 0;
+
+        [SuppressMessage("ReSharper", "UnassignedField.Global")]
```
ayende: Why unused?

src/Voron/Platform/PalDefinitions.cs
```diff
@@ -0,0 +1,26 @@
...
+        public struct SystemInformation
+        {
+            public Int32 PageSize;
+            public Int32 CanPrefetch;
+        }
+
+        [SuppressMessage("ReSharper", "UnusedMember.Global")]
```
ayende: Why unused?

src/Voron/Platform/PalFlags.cs
```diff
@@ -1,13 +1,13 @@
-using System;
-// ReSharper disable InconsistentNaming
-// ReSharper disable UnusedMember.Local
 // ReSharper disable IdentifierTypo
+// ReSharper disable InconsistentNaming
+// ReSharper disable UnusedMember.Global
+
```
ayende: I really don't like the R# notifications, let's fix them

src/Voron/Platform/PalFlags.cs
```diff
@@ -1,13 +1,13 @@
...
+using System;
 
 namespace Voron.Platform
 {
     public static class PalFlags
     {
-        [Flags]
         public enum FAIL_CODES : uint
```
ayende:
```suggestion
        public enum FailureFlags : uint
```

src/Voron/Platform/PalFlags.cs
```diff
@@ -21,14 +21,50 @@ public enum FAIL_CODES : uint
             FAIL_RACE_RETRIES = 7,
             FAIL_PATH_RECURSION = 8,
             FAIL_FLUSH_FILE = 9,
+            FAIL_SYSCONF = 10,
```
ayende: Use C# naming conventions, not UPPER_CASE

ayende: This is about all of the stuff in this file

src/Voron/Platform/PalFlags.cs
```diff
@@ -21,14 +21,50 @@ public enum FAIL_CODES : uint
...
+        }
+
+        public enum MMAP_OPTIONS
+        {
+            None = 0,
+            CopyOnWrite = 1,
+        }
+
+        [Flags]
+        public enum MSYNC_OPTIONS
```
ayende: This is a bad idea, you are reflecting the OS options to higher level code.
Instead, let's say what we _want_.
And unless I miss something, we always pass the same flags.

src/Voron/Platform/PalFlags.cs
```diff
@@ -21,14 +21,50 @@ public enum FAIL_CODES : uint
...
+        {
+            None = 0,
+            MS_ASYNC = (1 << 0),
+            MS_SYNC = (1 << 1),
+            MS_INVALIDATE = (1 << 2)
+        }
+
+        [Flags]
+        public enum MPROTECT_OPTIONS
```
ayende: Same, we don't need to expose these flags, we need to say what we want.
`MemoryProtectionFlags` with `ReadOnly`, `ReadWrite` are the only things we use

src/Voron/Platform/PalHelper.cs
```diff
@@ -1,30 +1,34 @@
...
+                const int maxNativeErrorStr = 256;
+                var buf = stackalloc byte[maxNativeErrorStr];
+                var size = Pal.rvn_get_error_string(lastError, buf, maxNativeErrorStr, out int specialErrnoCodes);
 
                 var nativeMsg = size >= 0 ? Encoding.UTF8.GetString(buf, size) : lastError.ToString();
 
                 txt = $"Errno: {lastError}='{nativeMsg}' (rc={specialErrnoCodes}) - '{msg}'";
 
-                if ((specialErrnoCodes & (int)PalFlags.ERRNO_SPECIAL_CODES.ENOMEM) != 0)
+                if ((specialErrnoCodes & (int)PalFlags.ERRNO_SPECIAL_CODES.ENOMEM) != 0 || forceThrowOutOfMemory)
```
ayende:
```suggestion
                if (((PalFlags)specialErrnoCodes & PalFlags.ERRNO_SPECIAL_CODES.ENOMEM) != 0 || forceThrowOutOfMemory)
```

src/Voron/Platform/PalHelper.cs
```diff
@@ -1,30 +1,34 @@
...
                     throw new OutOfMemoryException(txt);
 
                 if ((specialErrnoCodes & (int)PalFlags.ERRNO_SPECIAL_CODES.ENOENT) != 0)
                     throw new FileNotFoundException(txt);
+
+                if ((specialErrnoCodes & (int)PalFlags.ERRNO_SPECIAL_CODES.ENOSPC) != 0)
```
ayende:
```suggestion
                if (((PalFlags)specialErrnoCodes & PalFlags.ERRNO_SPECIAL_CODES.ENOSPC) != 0)
```

src/Voron/Platform/Posix/PosixHelper.cs
```diff
@@ -27,7 +27,7 @@ public static void AllocateFileSpace(StorageEnvironmentOptions options, int fd,
             if (result == (int)Errno.ENOSPC)
             {
                 var diskSpaceResult = DiskSpaceChecker.GetDiskSpaceInfo(file);
-                throw new DiskFullException(file, size, diskSpaceResult?.TotalFreeSpace.GetValue(SizeUnit.Bytes));
+                throw new DiskFullException(file, size, diskSpaceResult?.TotalFreeSpace.GetValue(SizeUnit.Bytes), result.ToString());
```
ayende: Why are you calling `ToString` on the `result`? That wouldn't give a good error message

src/Voron/Platform/Win32/Win32NativeFileMethods.cs
```diff
@@ -128,7 +128,7 @@ public static void SetFileLength(SafeFileHandle fileHandle, long length)
                 if (lastError == (int) Win32NativeFileErrors.ERROR_DISK_FULL)
                 {
                     var driveInfo = DiskSpaceChecker.GetDiskSpaceInfo(filePath);
-                    throw new DiskFullException(filePath, length, driveInfo?.TotalFreeSpace.GetValue(SizeUnit.Bytes));
+                    throw new DiskFullException(filePath, length, driveInfo?.TotalFreeSpace.GetValue(SizeUnit.Bytes), lastError.ToString());
```
ayende: Same, `lastError` is a string, we need an actual error

src/librvnpal/inc/rvn.h
```diff
@@ -4,34 +4,88 @@
...
 #define true 1
 #define false 0
 
+#define max(x, y) ((x) >= (y)) ? (x) : (y)
+#define min(x, y) ((x) <= (y)) ? (x) : (y)
+
+EXPORT struct SYSTEM_INFORMATION
+{
+    int32_t page_size;
+    int32_t can_prefetch;
```
ayende:
```suggestion
    bool can_prefetch;
```

src/Voron/Platform/PalFlags.cs
ayende: Also, this probably should be `int`, not `uint`

src/librvnpal/inc/rvn.h
```diff
@@ -4,34 +4,88 @@
...
 
-int32_t /* internal use */
-flush_file (int32_t fd);
+EXPORT int32_t
+rvn_memory_sync(void *address, int64_t size, int32_t flags, int32_t *detailed_error_code);
 
-int32_t /* internal use */
-sync_directory_allowed (int dir_fd);
+EXPORT int32_t
+rvn_dispose_handle(const char *filepath, void *handle, int32_t delete_on_close, int32_t *unlink_error_code, int32_t *close_error_code);
```
ayende: Should have just one `int32_t` error. We can't expose our internal impl in this manner

src/librvnpal/inc/rvn.h
```diff
@@ -4,34 +4,88 @@
...
 
+EXPORT int32_t
+rvn_unmap(void *address, int64_t size, int32_t delete_on_close, int32_t *unmap_error_code, int32_t *madvise_error_code);
```
ayende: One error out var

src/librvnpal/inc/rvn.h
```diff
@@ -4,34 +4,88 @@
...
 
+EXPORT int32_t
+rvn_protect_range(void *start_address, int64_t size, int32_t flags, int32_t *detailed_error_code);
+
+EXPORT int32_t
+rvn_allocate_more_space(int64_t new_length, int64_t total_allocation_size, const char *filename, void *handle, int32_t flags, void **new_address, int64_t *new_length_after_adjustment, int32_t *detailed_error_code);
+
+/* For internal use: */
+PRIVATE int64_t
+rvn_nearest_size_to_page_size(int64_t orig_size, int64_t sys_page_size);
```
ayende: `rvn_` prefix is for public stuff, let's use a different one

src/librvnpal/inc/status_codes.h
```diff
@@ -2,26 +2,49 @@
...
+#define ERRNO_SPECIAL_CODES_ENOSPC      (1 << 3)
+
+#define SYNC_DIR_ALLOWED                0
+#define SYNC_DIR_NOT_ALLOWED            1
+
+#define MMOPTIONS_NONE                  0
+#define MMOPTIONS_COPY_ON_WRITE         1
+
+#define MSYNC_OPTIONS_MS_SYNC           (1 << 0)
+#define MSYNC_OPTIONS_MS_ASYNC          (1 << 1)
```
ayende: When do we use this?

src/librvnpal/inc/status_codes.h
```diff
@@ -2,26 +2,49 @@
...
+#define MSYNC_OPTIONS_MS_INVALIDATE     (1 << 2)
+
+#define MPROTECT_OPTIONS_PROT_NONE      0
+#define MPROTECT_OPTIONS_PROT_READ      (1 << 0)
+#define MPROTECT_OPTIONS_PROT_WRITE     (1 << 1)
+#define MPROTECT_OPTIONS_PROT_EXEC      (1 << 2)
```
ayende: When do we use this?

I really want to avoid copying the structure of the OS flags

src/librvnpal/src/posix/fileutils.c
```diff
@@ -0,0 +1,136 @@
...
+    {
+        return orig_size;
+    }
+    return ((orig_size / sys_page_size) + 1) * sys_page_size;
+}
+
+int64_t
+rvn_pwrite(int32_t fd, void *buffer, uint64_t count, uint64_t offset, int32_t *detailed_error_code)
+{
+    int64_t actually_written = 0;
```
ayende:
```suggestion
    uint64_t actually_written = 0;
```

src/librvnpal/src/posix/fileutils.c
```diff
@@ -0,0 +1,136 @@
...
+    int64_t cifs_retries = 3;
+    do
+    {
+        int64_t result = pwrite(fd, buffer, count - (uint64_t)actually_written, offset + actually_written);
```
ayende:
```suggestion
        int64_t result = pwrite(fd, buffer, count - actually_written, offset + actually_written);
```

src/librvnpal/src/posix/fileutils.c
```diff
@@ -0,0 +1,136 @@
...
+        if (result < 0) /* we assume zero cannot be returned at any case as defined in POSIX */
+        {
+            if (errno == EINVAL && rvn_sync_directory_allowed(fd) == false && --cifs_retries > 0)
+            {
+                /* cifs/nfs mount can sometimes fail on EINVAL after file creation
+                lets give it few retries with short pauses between them - RavenDB-11954 */
+
+                sleep(1);
```
ayende: Use `usleep` and the shorter sleep times

src/librvnpal/src/posix/fileutils.c
```diff
@@ -0,0 +1,136 @@
...
+            *detailed_error_code = errno;
+            return result;
+        }
+    }
+
+    return result; /* return EINTR */
+}
+
+int32_t
+rvn_pointer_to_int(void *ptr)
```
ayende: Why is this required? we can just explicitly cast to `int32_t`, no?
I believe that this could should be a macro with:

```
#define PtrToInt(p) ((int32_t)(intptr_t)p))

```

src/librvnpal/src/posix/fileutils.c
```diff
@@ -0,0 +1,136 @@
...
+{
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wint-conversion"
+    return ptr;
+#pragma GCC diagnostic pop
+}
+
+int32_t
+rvn_dispose_handle(const char *filepath, void *handle, int32_t delete_on_close, int32_t *unlink_error_code, int32_t *close_error_code)
```
ayende: Single out error

src/librvnpal/src/posix/geterrorstring.c
```diff
@@ -18,8 +23,11 @@ int32_t rvn_get_error_string(int32_t error, char* buf, int32_t buf_size, int32_t
 		case ENOENT:
 			*special_errno_flags = ERRNO_SPECIAL_CODES_ENOENT;
 			break;
+		case ENOSPC:
+			*special_errno_flags = ERRNO_SPECIAL_CODES_ENOSPC;
+			break;
 		default:
-			*special_errno_flags = ERRNO_SPECIAL_CODES_NONE;
+			*special_errno_flags = ERRNO_SPECIAL_CODES_NO_ERROR;
```
ayende: `SUCCESS` ?

src/librvnpal/src/posix/mapping.c
```diff
@@ -0,0 +1,174 @@
...
+                           void **handle,
+                           void **base_addr,
+                           int64_t *actual_file_size,
+                           int32_t *detailed_error_code)
+{
+    int32_t rc = SUCCESS;
+
+    assert(initial_file_size > 0);
+
+    /* TODO: rnv_ensure_file_exists(path); */
```
ayende: isn't this done by `O_CREAT` ?

src/librvnpal/src/posix/mapping.c
```diff
@@ -0,0 +1,174 @@
...
+    int64_t sys_page_size = sysconf(_SC_PAGE_SIZE);
+    if (sys_page_size == -1)
+    {
+        rc = FAIL_SYSCONF;
+        goto error_cleanup;
+    }
+
+    if (sz <= initial_file_size || sz % sys_page_size != 0)
+    {
+        sz = rvn_nearest_size_to_page_size(max(initial_file_size, sz), sys_page_size);
```
ayende: The min here isn't `sys_page_size` but `allocation granularity`, `4KB` on linux, 64Kb on windows, let's always make it 64KB

src/librvnpal/inc/rvn.h
```diff
@@ -4,34 +4,88 @@
...
 
 EXPORT int32_t
-rvn_write_header(const char *path, void *header, int32_t size, uint32_t *detailed_error_code);
+rvn_write_header(const char *path, void *header, int32_t size, int32_t *detailed_error_code);
 
+EXPORT int32_t
+rvn_get_error_string(int32_t error, char *buf, int32_t buf_size, int32_t *special_errno_flags);
+
+EXPORT int32_t
+rvn_create_and_mmap64_file(const char *path, int64_t initial_file_size, int32_t flags, void **handle, void **base_addr, int64_t *actual_file_size, int32_t *detailed_error_code);
```
ayende, line 50: Let's document here what this means, that we are getting read/write memory, etc. 
so the flags that we are actually using and their impact

src/librvnpal/src/posix/fileutils.c
```diff
@@ -0,0 +1,136 @@
...
+            if (unlink_rc != 0)
+            {
+                /* record the error and continue to close */
+                rc |= FAIL_UNLINK;
+                *unlink_error_code = errno;
+            }
+        }
+    }
+
+    int32_t close_rc = close(fd);
```
ayende: Check that fd != -1 ?

src/librvnpal/src/posix/mapping.c
```diff
@@ -0,0 +1,174 @@
...
+error_cleanup:
+    *detailed_error_code = errno;
+    if (fd != -1)
+        close(fd);
+
+    return rc;
+}
+
+int32_t
+rvn_allocate_more_space(int64_t new_length, int64_t total_allocation_size, const char *filename, void *handle, int32_t flags,
```
ayende: I don't understand the signature here.
You are operating of a file, so that should be first. 
what is the new_legnth and total allocation for?

src/librvnpal/src/posix/mapping.c
```diff
@@ -0,0 +1,174 @@
...
+    int64_t sys_page_size = sysconf(_SC_PAGE_SIZE);
+    if (sys_page_size == -1)
+    {
+        rc = FAIL_SYSCONF;
+        goto error_cleanup;
+    }
+
+    *new_length_after_adjustment = rvn_nearest_size_to_page_size(new_length, sys_page_size);
+
+    if (*new_length_after_adjustment <= total_allocation_size)
```
ayende: I assume that total_allocation_size is the current size?  Is this method increasing the size of the allocation?

src/librvnpal/src/posix/mapping.c
```diff
@@ -0,0 +1,174 @@
...
+error_cleanup:
+    *detailed_error_code = errno;
+    if (fd != -1)
+        close(fd);
+
+    return rc;
+}
+
+int32_t
+rvn_unmap(void *address, int64_t size, int32_t delete_on_close, int32_t *unmap_error_code, int32_t *madvise_error_code)
```
ayende: single out error

src/librvnpal/src/posix/sync.c
```diff
@@ -10,134 +10,163 @@
...
+
+    return rc;
+}
 
-  return rc;
+int32_t
+rvn_memory_sync(void *address, int64_t size, int32_t flags, int32_t *detailed_error_code)
+{
+    int32_t msync_flags = 0;
+    if (flags & MSYNC_OPTIONS_MS_SYNC)
```
ayende: Nope, we aren't going to expose the same interface as the OS has, we need to expose our own, much simpler interface

src/librvnpal/src/posix/virtualmemory.c
```diff
@@ -0,0 +1,68 @@
...
+    int32_t prefetch_errno = 0;
+    for (i = 0; i < count; i++)
+    {
+        struct RVN_RANGE_LIST record = range_list[i];
+        int32_t prefetch_rc = rvn_prefetch_virtual_memory(record.virtual_address, record.number_of_bytes, &prefetch_errno);
+        if (rc == 0) /* record first failure */
+        {
+            rc = prefetch_rc;
+            *detailed_error_code = prefetch_errno;
+        }
```
ayende: On any failure, we immediately return

src/librvnpal/src/posix/virtualmemory.c
```diff
@@ -0,0 +1,68 @@
...
+int32_t
+rvn_prefetch_ranges(struct RVN_RANGE_LIST *range_list, int32_t count, int32_t *detailed_error_code)
+{
+    int32_t i = 0;
+    int32_t rc = 0;
+    int32_t prefetch_errno = 0;
+    for (i = 0; i < count; i++)
+    {
+        struct RVN_RANGE_LIST record = range_list[i];
+        int32_t prefetch_rc = rvn_prefetch_virtual_memory(record.virtual_address, record.number_of_bytes, &prefetch_errno);
```
ayende:
```suggestion
        int32_t prefetch_rc = rvn_prefetch_virtual_memory(record.virtual_address, record.number_of_bytes, detailed_error_code);
```

src/librvnpal/src/posix/virtualmemory.c
```diff
@@ -0,0 +1,68 @@
...
+        }
+    }
+    return rc;
+}
+
+int32_t
+rvn_protect_range(void *start_address, int64_t size, int32_t flags, int32_t *detailed_error_code)
+{
+    int32_t mprotect_flags = 0;
+    if (flags & MPROTECT_OPTIONS_PROT_NONE)
```
ayende: Nope, we don't expose the OS state outside, only our own state / options

src/librvnpal/src/posix/writefileheader.c
```diff
@@ -10,65 +10,72 @@
...
-    goto error_cleanup;
-  }
-
-  close (fd);
-  fd = -1;
-
-  if (syncIsNeeded == true)
-    return sync_directory_for (path, detailed_error_code);
-  return SUCCESS;
+rvn_write_header(const char *path,
```
ayende, line 21: Does this include @haludi 's changes?

src/librvnpal/src/posix/writefileheader.c
```diff
@@ -10,65 +10,72 @@
...
+
+    if (fd == -1)
+    {
+        rc = FAIL_OPEN_FILE;
+        goto error_cleanup;
+    }
+
+    int32_t remaining = size;
+
+    /* do not call rvn_get_file_size to avoid 3 lseek calls, and use only 2 calls */
```
ayende: that isn't a meaningful optimization, we write this very rarely

src/librvnpal/src/posix/writefileheader.c
```diff
@@ -10,65 +10,72 @@
...
+        rc = FAIL_SEEK_FILE;
+        goto error_cleanup;
+    }
+
+    if (sz != remaining)
+        syncIsNeeded = true;
+
+    while (remaining > 0)
+    {
+        uint64_t written = write(fd, header, (uint64_t)remaining);
```
ayende, line 49: Can just use `pwrite` here and avoid setting calling `lseek` again.
Although, can call `stat` here, which likely will be fastest option

ayende (changes requested):

ayende:
And, to verify, this doesn't yet include windows impl, right?

src/librvnpal/src/posix/mapping.c
```diff
@@ -1,5 +1,11 @@
 #if defined(__unix__) || defined(__APPLE__)
 
+#ifdef __APPLE__
+#define MMAP mmap
```
ayende: We can probably just do `mmap64` ->`mmap` here, and avoid `MMAP`

src/librvnpal/src/posix/virtualmemory.c
```diff
@@ -53,10 +53,12 @@ rvn_protect_range(void *start_address, int64_t size, int32_t flags, int32_t *det
         mprotect_flags |= PROT_WRITE;
     if (flags & MPROTECT_OPTIONS_PROT_EXEC)
         mprotect_flags |= PROT_EXEC;
+#ifndef __APPLE__
```
ayende: All of this is not required. And this is exactly why not.
We don't need these options, and exposing them cause us problem

ayende (changes requested):

src/Voron/Platform/PalDefinitions.cs
aviviadi: As suggested from [this](https://stackoverflow.com/questions/20035826/why-dllimport-for-c-bool-as-unmanagedtype-i1-throws-but-as-byte-it-works?lq=1) and [this](https://docs.microsoft.com/en-us/cpp/dotnet/calling-native-functions-from-managed-code?view=vs-2017) C# `bool` must be marshaled (in our case to Int32, or maybe we want to change it to c89's byte?)

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
aviviadi: It is a flag to return null, in order to be consistent with previous implementation (like [here](https://github.com/ravendb/ravendb/blob/v4.1/src/Voron/Impl/Paging/AbstractPager.cs#L315))

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
aviviadi: unmap exec madvise(DONT_NEED) and then unmap.  madvise might fail and unmap succeed.
should we fail on that call, should we log it, or should we ignore it? (ignore sounds right to me)

src/Voron/Platform/PalDefinitions.cs
aviviadi: I ended up with `[MarshalAs(UnmanagedType.I4)]` for the bool type to make sure we write int32_t into that bool and not something else (as it might change between OSs and compilers)

src/Voron/Platform/Pal.cs
aviviadi: but we might have them both fail and we want to log both of them (to understand later if and why we have leftovers)

src/Voron/Platform/Pal.cs
aviviadi: here it makes sense to leave one error code if madvise error should be ignored and maybe even not returning madvise error at all if we are not logging it. Is this still the case? why not leaving the option to read all the failures? (if it is a signature issue, we can implement `(Int32 *)errorCodeArray`)

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
aviviadi: Is it ok to leave the SafeHandle implementation to the Rvn MMP's Windows part?

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
aviviadi: Will be done when we implement other pagers (added to [TODOs](https://issues.hibernatingrhinos.com/issue/RavenDB-12376#focus=streamItem-67-33365-0-0))

src/librvnpal/src/posix/mapping.c
aviviadi: no, it is not.

src/librvnpal/src/posix/virtualmemory.c
aviviadi: I want to record the first error but continue with the other ranges. The above suggestion will override the errno if second iteration successful.

aviviadi:
> And, to verify, this doesn't yet include windows impl, right?

no, it is not.

src/librvnpal/src/posix/mapping.c
aviviadi: as a precocious i would like to leave it this way, as :
```
void *mmap    (void *addr, size_t len, int prot, int flags, int fildes, off_t off);
void *mmap64(void *addr, size_t len, int prot, int flags, int fildes, off64_t off); 
``` 
in [here](https://www.mkssoftware.com/docs/man3/mmap.3.asp) it states : `The mmap64() function is a part of the large file extensions. `

src/librvnpal/src/posix/writefileheader.c
aviviadi: nop.. @haludi will merge later on using this function

src/librvnpal/src/posix/writefileheader.c
aviviadi: we don't have rvn_get_file_size anyhow..

src/librvnpal/src/posix/fileutils.c
aviviadi: Simple cast generates compiler warning which want to avoid just here

src/librvnpal/src/posix/mapping.c
aviviadi: total_allocation_size is the "previous" size and this is for knowing if we need to increase the file size or return "AlloctionNoResize". 
The new_length.. is the new length to resize to..

src/librvnpal/src/posix/mapping.c
aviviadi: > I assume that total_allocation_size is the current size? Is this method increasing the size of the allocation?

it returns new_length_after_adjustment and the caller sets the total_allocation_size to the new value.

src/Voron/Platform/PalDefinitions.cs
ayende: `I4` is fine, yes

src/librvnpal/src/posix/virtualmemory.c
ayende: No, it is a bad idea. We want to exit early

ayende: We got an error, and continuing work from there means that we don't know what is our state.
given that this is an optimization only, that doesn't matter

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
ayende: But this isn't actually an error. This is good, we don't need to do anything because we are big enough.
For that matter, can we move this check to the managed code?

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
ayende: No, when we want to `unmap`, that is what we care about.
If this is a temp file and the `madvice` failed, that is a missed optimization, not something we care about

src/Voron/Platform/Pal.cs
ayende: That doesn't matter, we fail early and return the earlier failure code.
Note that we still want to _close_ the file, but if we are already in error mode, we don't care if that failed.

ayende: The reasoning behind this is that we can't expose our internal state and behavior outside. That changes over time.
What we want to do is to have a common error handling interface.
Return code, `int32_t` with our own error code.
`detailed_error_code` with the system error code.

Any additional data you want, must be encoded in the return code, nothing else.

src/Voron/Platform/Pal.cs
ayende: Because you are baking the implementation behavior directly into the interface.
What if we have a new behavior, we want to not have to change the signature.
For example, if we decided that encrypted files need to be zeroed first, and that fails, do we now need _another_  error code?

That violates design guidelines and something we want to avoid.

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
ayende: No, `SafeHandle` is something that we want for all platforms.
Basically, this is just a way to ensure that if we leak a handle, the GC's finalizer will dispose it.

ayende: Basically, same as `SafeFileHandle`

src/librvnpal/src/posix/mapping.c
ayende: so call it `rvn_mmap` instead, I don't like `MMAP`

src/librvnpal/src/posix/fileutils.c
ayende:
```//gcc 5.4.0

#include  <stdio.h>


int main(void)
{
    
    void * a = (void*)5;
    int  b = (int)(long)a;
    printf("Hello, world! %i", b);
    return 0;
}```

Compiles with now warnings. Note the _two_ casts, we are being explicit about it

src/librvnpal/src/posix/mapping.c
ayende: But we already know that in the managed code, so why call out to here anyway?

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
aviviadi: if we move this back to managed, we need : _nearest_size_to_page_size.  It will not require pinvoke as we already have the page size from get_system_information.  
Do you want to re-implement _nearest_size_to_page_size in managed ?

src/Voron/Platform/PalDefinitions.cs
aviviadi: I4 doesn't work so I am left with U1

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
ayende: We already have the page size in managed code, so there is nothing else we need.
We can do that in managed code, yes.
Still need to _verify_ that the desired size is page aligned.

src/Voron/Platform/PalDefinitions.cs
ayende: The problem is that the moment we use `char` in here, that mess up alignment, padding, etc.
We want to ensure that this is not going to cause us (very subtle) problems down the line, so we want to make the compiler's job as simple as possible.
That means that we have to be strict about alignment, help the compiler avoid the need for padding, etc.

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
aviviadi: I ment - can we delay the implementation to the window's part PR. and then when we do that - also for posix.

ayende: Why, it is a pretty simple one

src/Voron/Platform/PalDefinitions.cs
aviviadi: so we have to cast from bool to int on the pInvoke calling

ayende: Or use `int`

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
aviviadi: So, we are left with the native error code handling `GetNativeErrorString(lastError, msg, out var specialErrnoCodes)` which can return NoMem and NoSpc.  In posix we rely on the native error code returned by the syscalls. in windows we will return NoMem or NoSpc according to exceptions I presume

src/librvnpal/src/posix/geterrorstring.c
aviviadi: NONE instead of SUCCESS (zero value). it is not Success necessarily

src/Voron/Platform/PalDefinitions.cs
aviviadi: I originally used Int32.. get back to it ?

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
ayende: What we get back (always) is out own codes.
If they happen to coincide in Linux with something, great.
But I would worry about that across distros and would rather use our own directly always

src/Voron/Platform/PalDefinitions.cs
ayende: yes

aviviadi:
will be continued at #8502

PR #8431, opened 2019-01-07
RavenDB-12137 [WIP] Add graph queries to RunningQueries

talweiss1982:


src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -1436,6 +1445,18 @@ public long GetEnvironmentsHash()
 
             return hash;
         }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void AddToCurrentlyRunningGraphQueries(ExecutingQueryInfo executingQueryInfo)
+        {
+            CurrentlyRunningGraphQueries.Add(executingQueryInfo);
```
ppekrol: Can we move it to GraphQueryRunner and acces it via Database.QueryRunner.Graph.CurrentlyRunningGraphQueries ?

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -392,7 +393,15 @@ public void NotifyOnPendingClusterTransaction()
         private long? _nextClusterCommand;
         private long _lastCompletedClusterTransaction;
         public long LastCompletedClusterTransaction => _lastCompletedClusterTransaction;
+        private int _numberOfGraphQueries;
+
+        public int GetNextGraphQueryId()
+        {
+            return Interlocked.Increment(ref _numberOfGraphQueries);
```
ppekrol: Can we move it to GraphQueryRunner and acces it via Database.QueryRunner.Graph.CurrentlyRunningGraphQueries ?

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
```diff
@@ -17,16 +18,23 @@ public class QueriesDebugHandler : DatabaseRequestHandler
         [RavenAction("/databases/*/debug/queries/kill", "POST", AuthorizationStatus.ValidUser)]
         public Task KillQuery()
         {
+            ExecutingQueryInfo query;
             var name = GetQueryStringValueAndAssertIfSingleAndNotEmpty("indexName");
             var id = GetLongQueryString("id");
+            if (name == "@graph")
```
ppekrol: '@graph' should be moved to constants e.g. Constants.Documents.Indexing.GraphIndexName

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
```diff
@@ -68,29 +98,32 @@ public Task RunningQueries()
...
-                        writer.WritePropertyName((nameof(query.QueryId)));
-                        writer.WriteInteger(query.QueryId);
-                        writer.WriteComma();
+                if(indexes.Count != 0)
+                    writer.WriteComma();
 
-                        writer.WritePropertyName((nameof(query.StartTime)));
-                        writer.WriteDateTime(query.StartTime, isUtc: true);
-                        writer.WriteComma();
+                writer.WritePropertyName("Graph queries");
```
ppekrol: Can we use property without space? Maybe `@graph` that you used before?

src/Raven.Server/Documents/Queries/GraphQueryDoneRunning.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+
+namespace Raven.Server.Documents.Queries
+{
+    public struct GraphQueryDoneRunning : IDisposable
+    {
+        private static readonly TimeSpan DefaultLockTimeout = TimeSpan.FromSeconds(3);
```
ppekrol: Is this used?

src/Raven.Server/Documents/Queries/GraphQueryDoneRunning.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        private static readonly TimeSpan ExtendedLockTimeout = TimeSpan.FromSeconds(30);
```
ppekrol: Is this used?

ppekrol (changes requested):

src/Raven.Server/Documents/DocumentDatabase.cs
ayende: Yes, this shouldn't be on `DocumentDatabase`

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -392,7 +393,15 @@ public void NotifyOnPendingClusterTransaction()
...
+        }
+
         private int _clusterTransactionDelayOnFailure = 1000;
+        public readonly ConcurrentSet<ExecutingQueryInfo> CurrentlyRunningGraphQueries = new ConcurrentSet<ExecutingQueryInfo>();
```
ayende: Why do we need a separate behavior for graph queries? Can't we use the existing running queries for the same thing?

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
ayende: I really don't like it being separate. 
These are just queries, and should be treated the same.

ayende (changes requested):
I don't like this impl.
What we want to do here is to use the exact same code and process that we have for the already existing running queries impl, as well as the same cancellation support.

That means that we call the regular `AddToRunningQueries` and (maybe) handle any changes there, but I think we probably shouldn't have to do much there.

ppekrol:
@ayende this is because running queries were only on index-level

@talweiss1982 I think we need to take this opportunity and move running queries to QueryRunner, so we can handle all types of queries (index, collection, graph). Thoughts?

ayende:
fine by me

maximburyak:
@ayende @ppekrol can we postpone the change proposed be @ppekrol to 4.2.x? I think that right now we want to concentrate on getting all graph features done

ppekrol:
We will do it in 2 stages.
1. @talweiss1982 will move the code to GraphQueryRunner (as was written in one of the first comments) and add the cancellation support
2. To allow @talweiss1982 to continue the work on Graph issues, I will do the unification

talweiss1982:
@ppekrol i added cancelation to graph queries and moved the running graph queries under the runner

ppekrol (approved):
Rebased and did some minor adjustments.

ayende (approved):

PR #8436, opened 2019-01-08
RavenDB-12511 Expose queries and how to kill them in the studio

ml054:


ml054:
cc: @arekpalinski

ppekrol (approved):

arekpalinski (approved):

src/Raven.Server/Documents/Handlers/Debugging/ServerWideQueriesDebugHandler.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Threading.Tasks;
+using Raven.Server.Documents.Queries;
+using Raven.Server.Routing;
+using Raven.Server.Web;
+
+namespace Raven.Server.Documents.Handlers.Debugging
+{
+    public class ServerWideQueriesDebugHandler : RequestHandler
+    {
+        [RavenAction("/debug/queries/running/live", "GET", AuthorizationStatus.ValidUser)]
```
ayende, line 14: No, this is wrong.
This is server wide, and we have to make sure that this is:
`/admin/debug` and that the permission level is `Operator`

ayende (changes requested):

src/Raven.Server/Documents/Handlers/Debugging/ServerWideQueriesDebugHandler.cs
ml054: We have single endpoint for every user, please have a look at https://github.com/ravendb/ravendb/pull/8436/files/2bb1368e4f0037cc356ebf5798659055073ce222#diff-e1eadb52a3455ec6c3d3be2636d09b4eR16  we filter to allowed databases. So you get only allowed db in this websocket. 

Is it ok then?

src/Raven.Server/Documents/Queries/LiveRunningQueriesCollector.cs
```diff
@@ -0,0 +1,138 @@
...
+using Raven.Server.ServerWide;
+using Raven.Server.Utils.Stats;
+using Sparrow.Json;
+
+namespace Raven.Server.Documents.Queries
+{
+    public class LiveRunningQueriesCollector : LivePerformanceCollector<LiveRunningQueriesCollector.ExecutingQueryCollection>
+    {
+        private readonly ServerStore _serverStore;
+        private readonly string[] _dbNames;
```
ayende: Turn this to a Set, instead.

src/Raven.Server/Documents/Handlers/Debugging/ServerWideQueriesDebugHandler.cs
ayende: yes

src/Raven.Server/Documents/Queries/LiveRunningQueriesCollector.cs
ml054: done

PR #8440, opened 2019-01-08
RavenDB-12610

ppekrol:
- WaitForExecutionOnRelevantNodes needs to wait for execution on all nodes, not the first one


src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -428,44 +429,67 @@ private async Task WaitForExecutionOnRelevantNodes(JsonOperationContext context,
...
+                        {
+                            // unwrapping exceptions for all of the faulted tasks
+                            await Task.WhenAll(faultedTasks);
+                        }
+                        catch (OperationCanceledException oce)
+                        {
+                            throw new TimeoutException(
+                                $"Waited too long for the raft command (number {index}) to be executed on any of the relevant nodes to this command.", oce);
+                        }
+                        catch (RavenException e) when (e.InnerException is HttpRequestException)
```
arekpalinski: What if two of them fail with this error, won't we get aggregate exception then?

ppekrol: Fixed

PR #8452, opened 2019-01-09
PrefetchingBehavior continuations can execute synchronously because they are short lived

danielmarbach:
I was investigating an issue for a client and had to look into the PrefetchBehavior. I did not look into many details there but I have a hunch the continuations scheduled don't actually need to be scheduled. Adding `ExecuteSynchronously` can reduce the thread pool pressure on the server.

> Specifies that the continuation task should be executed synchronously. With this option specified, the continuation runs on the same thread that causes the antecedent task to transition into its final state. If the antecedent is already complete when the continuation is created, the continuation will run on the thread that creates the continuation. If the antecedent's CancellationTokenSource is disposed in a finally block (Finally in Visual Basic), a continuation with this option will run in that finally block. Only very short-running continuations should be executed synchronously.    Because the task executes synchronously, there is no need to call a method such as Wait() to ensure that the calling thread waits for the task to complete.

https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskcontinuationoptions

Technically the `Task.WaitAll` in the dispose method

```
Task.WaitAll(futureIndexBatches.Values.Select(ObserveDiscardedTask).ToArray());
```

could be removed but I left it there since it doesn't hurt.

### DIFF
--- a/Raven.Database/Prefetching/PrefetchingBehavior.cs
+++ b/Raven.Database/Prefetching/PrefetchingBehavior.cs
@@ -1283,7 +1283,7 @@ namespace Raven.Database.Prefetching
                         t.AssertNotFailed();
                     }
                     return t.Result;
-                })
+                }, CancellationToken.None, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default)
             };
 
             futureIndexBatch.Task.ContinueWith(t =>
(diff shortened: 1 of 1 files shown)


talweiss1982:
I think that it is hard to determine the behavior of this change since under different scenario and access pattern doing this modification can yield a different result.
@ayende we need to test this on multiple configurations and access patterns, multiple databases multiple indexes, heavy write and more.

ayende:
Are we sure that all of these are really very short lived?

danielmarbach:
@ayende it looks like. One is only asserting the task state, another one calls a delegate that is only used by testing infra and the third one just logs

ayende:
@talweiss1982 I'm incline to say that this is an acceptable change, can you run it through its paces next week?

talweiss1982 (approved):
I have run some tests and didn't see any major difference in performance

danielmarbach:
🎆

danielmarbach:
Is this already released?

ayende:
yes

PR #8454, opened 2019-01-09
RavenDB-12650 Simplifying the implementation - no need to return inne…

arekpalinski:
…rMemberExpression


arekpalinski:
The change was discussed with Aviv

PR #8465, opened 2019-01-10
RavenDB-12623: Fixed non unique ids in _threadIds due to _releaser be…

maximburyak:
…ing static ThreadStatic, therefore it was "stolen" between context pools using the same thread

https://issues.hibernatingrhinos.com/issue/RavenDB-12623

### DIFF
--- a/src/Sparrow/Json/JsonContextPoolBase.cs
+++ b/src/Sparrow/Json/JsonContextPoolBase.cs
@@ -69,7 +69,7 @@ namespace Sparrow.Json
         }
 
         [ThreadStatic]
-        private static ContextStackThreadReleaser _releaser;
+        private ContextStackThreadReleaser _releaser;
         
 
         private void EnsureCurrentThreadContextWillBeReleased(int currentThreadId)
@@ -339,7 +339,6 @@ namespace Sparrow.Json
                     return;
             }
         }
-
         public void Dispose()
         {
             if (_disposed)
@@ -352,8 +351,19 @@ namespace Sparrow.Json
                 _disposed = true;
                 ThreadLocalCleanup.ReleaseThreadLocalState -= CleanThreadLocalState;
                 _nativeMemoryCleaner.Dispose();
+
+#if Debug
+                var z = new HashSet<ContextStack>();
+#endif
                 foreach (var kvp in EnumerateAllThreadContexts())
                 {
+#if Debug
+                    if (z.Add(kvp) == false)
+                    {
+                        throw new InvalidOperationException("threads list is not unique");
+                    }
+#endif
+
                     kvp.Dispose();
                 }
                 _contextStacksByThreadId.Clear();


src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -69,7 +69,7 @@ public bool Add(JsonContextPoolBase<T> parent)
         }
 
         [ThreadStatic]
-        private static ContextStackThreadReleaser _releaser;
+        private ContextStackThreadReleaser _releaser;
```
ayende, line 72: That doesn't actually work. You don't get a separate thread static per value.
An option here is to either use `ThreadLocal` or to have `ContextStackThreadReleaser` hold a list of associated pools

ayende (changes requested):

ayende:
Also, I'm not sure that I understand what the actual issue here is, can you explain ?

src/Sparrow/Json/JsonContextPoolBase.cs
maximburyak: The commit message explains it, basically it caused thread id change in one context to affect another one. We suspect it happened because of shared instances of ThreadId members in ContextStackThreadReleaser. After that change above, duplicates in _threadIds were not reproduced, I undesrstand now that the ThreadStatic was ignored, but looking at the releaser's ctor, i think we might not need the thread locality. But yes, we better review it again next week

talweiss1982:
@ayende I suspect that the scenario that is causing this is when you have one thread shared by two context pools and one context is invoking ensure context will be released twice and in between a different context belonging to a different pool is also invoking the method. This will cause the thread holder to appear twice in the thread ID list  and when disposing of the context pool it will dispose the context twice.
Now what is the real world scenario where this will happen, I suspect an endpoint using server store and database and jumping in between them. In this case scenario we have one thread two context pools. This might also happen when combining indexes stuff with database stuff in the same endpoint.

src/Sparrow/Json/JsonContextPoolBase.cs
ayende: How could that be? The only way cross thread stuff is when this is finalized.
I'm not seeing how that can happen.

ayende:
I'm able to reproduce this fairly easily by doing a lot of writes with a `patch` and then creating indexes.
After a while, this happens

@talweiss1982 / @maximburyak The whole _point_ of the thread static stuff is that we won't have any concurrency there to worry about, so I don't see how that can be possible.

The code is meant to hold multiple context pools from different locations, so I need to understand what the actual failure is.

ppekrol:
@arekpalinski and @ppekrol comments on the subject:
https://github.com/ppekrol/ravendb/commit/99acb4e904c3938bd1845300b3e67c96d1da98f0

ayende:
So the problem is thread I'd reuse, it seems. That certainly will cause issues,yes

talweiss1982:
@ayende I didn't say anything about concurrency the scenario I described is 100% serial

talweiss1982:
@pawel check demonstrates what happens when two pools are using the same thread (not cunncurently ofc...)
Our "solution" resolve the problem since pools can't snag thread ID from each other. 
I think that one pool will dispose twice of a context and the other will leak a context

maximburyak:
@ppekrol thread id reuse is anothet problem, but I tjink thst the issue we encountered did not include that. When in EnsureCurrentThreadContextWillBeReleased we added thread id uniqueness validation in various points, the earliest one was in the beginning of the function. It means the value of ThreadIdHolder was modified by another run of this function, in another thread.

ppekrol:
@talweiss1982 @maximburyak By removing 'static' you are just ensuring that the finalizer of ContextStackThreadReleaser will be called only when you dispose the JsonContextPoolBase (ThreadStaticAttribute should be removed BTW - it is only valid on static fields) and GC executes finalizer, not when thread is no longer there, so from operational point of view, this also disables the 'hijacking' mechanism from `EnsureCurrentThreadContextWillBeReleased` (Interlocked.CompareExchange part on array).

What does it give you to have a property inside JsonContextPoolBase  called `_releaser` with thread id set to the thread that created the instance?

maximburyak:
@ppekrol the solution we introduced is not good enough- that's clear. I would like to concentrate on the problem definition, which apparently has several origins. The work with the _releaser is probably not correct at the proposed code, especially given that @ayende was able to reproduce the issue.

ayende:
@talweiss1982 / @maximburyak I am missing something in your scenario, then, because I don't see any problem with two context pools adding themselves to the same releaser, it is explicitly designed to support that, after all.

maximburyak:
@ayende, here's a commit visualising what @talweiss1982 described above (https://github.com/ravendb/ravendb/pull/8465#issuecomment-453452055): https://github.com/maximburyak/ravendb/blob/d788bf9e4cc1741e7bddc4b0f67fc55833e8fdd7/src/Sparrow/Json/JsonContextPoolBase.cs#L115
You can run the fast tests in debug for demonstration, when it fails for the first time, it fails after setting the releaser's ThreadId to a new value, while failing in validating "fellow" context pulls (I tracked other context pools in order to be able to debug this)

ppekrol:
@maximburyak isn't this caused by 'CleanThreadLocalState' removing the record from the dict, so 'EnsureCurrentThreadContextWillBeReleased' is called again?

ayende:
Root cause is that the `_release.ThreadIdHolder` will share the `ThreadId` instance between `JsonContextPool` instances, leading to.... issues. 
Already fixed in @maximburyak 's end and now going through tests

PR #8467, opened 2019-01-10
Changes for RavenDB-12135

myarichuk:
Add regression test for the issue and made exception more informative

### DIFF
--- a/src/Raven.Server/Documents/Queries/QueryMetadata.cs
+++ b/src/Raven.Server/Documents/Queries/QueryMetadata.cs
@@ -985,7 +985,7 @@ namespace Raven.Server.Documents.Queries
 
         private void ThrowInvalidDuplicateAliasInSelectClause(BlittableJsonReaderObject parameters, string finalAlias)
         {
-            throw new InvalidQueryException($"Duplicate alias '{finalAlias}' detected", QueryText, parameters);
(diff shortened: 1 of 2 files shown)


src/Raven.Server/Documents/Queries/QueryMetadata.cs
```diff
@@ -985,7 +985,7 @@ private void FillSelectFields(BlittableJsonReaderObject parameters)
 
         private void ThrowInvalidDuplicateAliasInSelectClause(BlittableJsonReaderObject parameters, string finalAlias)
         {
-            throw new InvalidQueryException($"Duplicate alias '{finalAlias}' detected", QueryText, parameters);
+            throw new InvalidQueryException($"Duplicate alias '{finalAlias}' detected. Note: if an alias is not specified, the implicit alias equals field's name. So a select clause 'select employee.Name, manager.Name' is invalid because it has two field with the same implicit alias - 'Name'. In order to fix it, use explicit aliases - 'select employee.Name as EmployeeName, manager.Name as ManagerName'", QueryText, parameters);
```
ppekrol: fixed few typos: `Duplicate alias '{finalAlias}' detected. Note: if an alias is not specified, the implicit alias is equal to field name. So a select clause 'select employee.Name, manager.Name' is invalid because it has two fields with the same implicit alias - 'Name'. In order to fix it, use explicit aliases - 'select employee.Name as EmployeeName, manager.Name as ManagerName'`

PR #8473, opened 2019-01-13
RavenDB-12623: Fixing problem with same ThreadIdHolder being present …

maximburyak:
…in two context pools, causing thread ids duplicates and overruns.

Also, added assertion code to catch that kind of issues, if they reappear

### DIFF
--- a/src/Sparrow/Json/JsonContextPoolBase.cs
+++ b/src/Sparrow/Json/JsonContextPoolBase.cs
@@ -2,15 +2,41 @@
 using System.Buffers;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading;
+using Sparrow.Collections;
 using Sparrow.LowMemory;
 using Sparrow.Threading;
 using Sparrow.Utils;
 
 namespace Sparrow.Json
 {
+    public class ThreadIdHolder
+    {
+        public int ThreadId;
+        public override string ToString()
+        {
+            return ThreadId.ToString();
+        }
+    }
+
+#if DEBUG
+    public interface ThreadIDsHolderInterface
+    {
+        ThreadIdHolder[] ThreadIDs { get; }
+    }
+
+    public static class JsonContextPoolsHolder
+    {
+        public static ConcurrentSet<ThreadIDsHolderInterface> AllContexts = new ConcurrentSet<ThreadIDsHolderInterface>();
+    } 
+#endif
+
     public abstract class JsonContextPoolBase<T> : ILowMemoryHandler, IDisposable
+#if DEBUG
+        , ThreadIDsHolderInterface
+#endif
         where T : JsonOperationContext
     {
         /// <summary>
@@ -19,12 +45,11 @@ namespace Sparrow.Json
         /// average to the same overall type of contexts
         /// </summary>
         private ConcurrentDictionary<int, ContextStack> _contextStacksByThreadId = new ConcurrentDictionary<int, ContextStack>();
-        private ThreadIdHolder[] _threadIds = Array.Empty<ThreadIdHolder>();
+        private ThreadIdHolder[] _threadIds = Array.Empty< ThreadIdHolder>();
 
-        private class ThreadIdHolder
-        {
-            public int ThreadId;
-        }
+        private static ConcurrentBag<ThreadIdHolder[]> AllThreadIdHolders = new ConcurrentBag<ThreadIdHolder[]>();
+
+        public ThreadIdHolder[] ThreadIDs => _threadIds;
 
         private readonly NativeMemoryCleaner<ContextStack, T> _nativeMemoryCleaner;
         private bool _disposed;
@@ -38,20 +63,20 @@ namespace Sparrow.Json
         {
             private HashSet<JsonContextPoolBase<T>> _parents = new HashSet<JsonContextPoolBase<T>>();
             int _threadId;
-            public ThreadIdHolder ThreadIdHolder;
+            public List<ThreadIdHolder> ThreadIdHolders = new List<ThreadIdHolder>();            
 
             public ContextStackThreadReleaser()
             {
                 _threadId = NativeMemory.CurrentThreadStats.Id;
-                ThreadIdHolder = new ThreadIdHolder
-                {
-                    ThreadId = _threadId
-                };
             }
 
             ~ContextStackThreadReleaser()
             {
-                ThreadIdHolder.ThreadId = -1;
+                foreach (var item in ThreadIdHolders)
+                {
+                    Debug.Assert(item.ThreadId == _threadId);
+                    item.ThreadId = -1;
+                }
                 // we remove the references from the thread dictionary
                 // it is possible that the pool is no longer referenced and was collected, and as such
                 // the finalizers for the context stack values were already run, if not, they will run soon
@@ -82,6 +107,10 @@ namespace Sparrow.Json
             if (_releaser.Add(this) == false)
                 return;
 
+#if DEBUG
+            AssertThreadIDsUniqueness(); 
+#endif
+
             while (true)
             {
                 var copy = _threadIds;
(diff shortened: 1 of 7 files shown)


src/Raven.Studio/package-lock.json
```diff
@@ -10,8 +10,8 @@
       "integrity": "sha1-iQrnxdjId/bThIYCFazp1+yUW9o=",
       "dev": true,
       "requires": {
-        "normalize-path": "2.1.1",
-        "through2": "2.0.3"
+        "normalize-path": "^2.0.1",
```
ayende: Wh?

src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -2,15 +2,41 @@
...
+    {
+        public int ThreadId;
+        public override string ToString()
+        {
+            return ThreadId.ToString();
+        }
+    }
+
+#if DEBUG
+    public interface ThreadIDsHolderInterface
```
ayende, line 25: I don't like all the code associated with it.
I don't think we need to carry it forward

src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -38,20 +63,26 @@ private class ContextStackThreadReleaser
...
             int _threadId;
-            public ThreadIdHolder ThreadIdHolder;
+            public List<ThreadIdHolder> ThreadIdHolders = new List<ThreadIdHolder>();
+            public static HashSet<int> pastThreadIDs = new HashSet<int>();
 
             public ContextStackThreadReleaser()
             {
                 _threadId = NativeMemory.CurrentThreadStats.Id;
-                ThreadIdHolder = new ThreadIdHolder
+                if (pastThreadIDs.Add(_threadId) == false)
```
ayende: Need to remove this

src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -91,21 +126,58 @@ private void EnsureCurrentThreadContextWillBeReleased(int currentThreadId)
...
+                {
+                    _releaser.ThreadIdHolders.Add(threads[copy.Length]);
                     break;
+                }
             }
         }
-  
+
+#if DEBUG
+        private void AssertThreadIDsUniqueness()
```
ayende, line 155: Probably should remove this

src/Sparrow/Utils/NativeMemory.cs
```diff
@@ -71,7 +73,7 @@ public bool IsThreadAlive()
             public ThreadStats()
             {
                 _threadInstance = Thread.CurrentThread;
-                Id = _threadInstance.ManagedThreadId;
+                Id = Interlocked.Increment(ref _uniqueThreadId);
```
ayende: Need to also have the `ManagedThreadId` here.

ayende (changes requested):
There is a lot of debug code here that still needs to be removed

src/Sparrow/Json/JsonContextPoolBase.cs
maximburyak: I would actually feel more comfortable if we'll be running with this "debug" code for a while, at least until being a bit closer to 4.2 release

maximburyak:
The latest force-push addressed the failing test and the redendant package-lock.json changes

src/Sparrow/Json/JsonContextPoolBase.cs
maximburyak: @ayende ?

src/Sparrow/Utils/NativeMemory.cs
ppekrol: Rename maybe to `InternalId` and we need to check every place we are using this and probably change bunch of those to `ThreadId`

src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -348,12 +417,27 @@ public void Dispose()
...
+#if DEBUG
+                JsonContextPoolsHolder.AllContexts.TryRemove(this); 
+#endif
                 _cts.Cancel();
                 _disposed = true;
                 ThreadLocalCleanup.ReleaseThreadLocalState -= CleanThreadLocalState;
                 _nativeMemoryCleaner.Dispose();
+
+#if Debug
+                var z = new HashSet<ContextStack>();
```
ppekrol: If we want to keep it then `z` is probably not the best name :)

src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -91,21 +120,58 @@ private void EnsureCurrentThreadContextWillBeReleased(int currentThreadId)
                     {
                         if(Interlocked.CompareExchange(ref copy[i].ThreadId, currentThreadId, -1) == -1)
                         {
-                            _releaser.ThreadIdHolder = copy[i];
+                            _releaser.ThreadIdHolders.Add(copy[i]);
                             return;
                         }
                         return;
```
ppekrol: This `return` needs to be removed

src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -91,21 +120,58 @@ private void EnsureCurrentThreadContextWillBeReleased(int currentThreadId)
                     {
                         if(Interlocked.CompareExchange(ref copy[i].ThreadId, currentThreadId, -1) == -1)
```
ppekrol, line 131: I think that the problem with `CleanThreadLocalState` still exists? https://github.com/ppekrol/ravendb/commit/99acb4e904c3938bd1845300b3e67c96d1da98f0#diff-1aa997bd6d5750decc752ea2c476347bR186

src/Sparrow/Utils/NativeMemory.cs
maximburyak: That's too risky move, I think

src/Sparrow/Json/JsonContextPoolBase.cs
maximburyak: I don't like it either, @ayende ?

src/Sparrow/Json/JsonContextPoolBase.cs
maximburyak: @ppekrol I've run the slowTests-> kill in the middle scenario, and it worked well

src/Sparrow/Utils/NativeMemory.cs
maximburyak: But maybe we should rename it to InternalId, in order for the code to be more correct

src/Sparrow/Json/JsonContextPoolBase.cs
maximburyak: yes

src/Sparrow/Utils/NativeMemory.cs
arekpalinski: @maximburyak in https://github.com/ravendb/ravendb/pull/8473/commits/0804bbc5ca48fca2564df7293a2c4dd72370f0ba you introduced `ManagedThreadId` but note that nobody uses it right now. The issue is that `Id` filed, which was the managed thread id, was used in multiple places. I think that all of them should use `ManagedThreadId ` after changes made in this PR.

I'm for renaming `ThreadStats.Id` to `ThreadStats.InternalId`

arekpalinski: Some correction, maybe not all of them but at least _some of them_ e.g.

- https://github.com/ravendb/ravendb/blob/cd1c8edf84f957153e9442e73248c8e889e33427/src/Raven.Server/Utils/ThreadsUsage.cs#L71

- https://github.com/ravendb/ravendb/blob/cd1c8edf84f957153e9442e73248c8e889e33427/src/Raven.Server/Documents/Handlers/Debugging/MemoryDebugHandler.cs#L287

src/Sparrow/Json/JsonContextPoolBase.cs
arekpalinski: Yeah, but doesn't it look suspicious to you? We were wondering if that can cause other issues.

In `~ContextStackThreadReleaser() ` we mark "slot" as free by setting `ThreadId = -1; ` and later remove from the dict: `_contextStacksByThreadId.TryRemove`:
https://github.com/ravendb/ravendb/blob/1118d10015513c32c29c1b61dba1ff20958bd3fe/src/Sparrow/Json/JsonContextPoolBase.cs#L73-L87
 While in `CleanThreadLocalState` we only call `_contextStacksByThreadId.TryRemove`. Shouldn't `-1` be set as well?

src/Sparrow/Utils/NativeMemory.cs
maximburyak: @arekpalinski for debug stuff, we want to use the ManagedThreadId, for any other place, where we want to identify the thread uniquely, we need to use the InternalId

arekpalinski: Makes sense. So we just need to fix debug stuff usages.

src/Sparrow/Json/JsonContextPoolBase.cs
maximburyak: yes, we should call the ContextStackThreadReleaser release logic from there, assign null and suppress finalization - it might be more complex than that, updates in next commit

src/Sparrow/Utils/NativeMemory.cs
maximburyak: done

src/Sparrow/Json/JsonContextPoolBase.cs
ppekrol: The problem is that the `for` loop will not work as expected. When we fail to do compareExchange on first item, then we are returning and not registering our thread at all.

maximburyak: done

ppekrol (approved):

arekpalinski (approved):

src/Sparrow/Json/JsonContextPoolBase.cs
ayende: No, that is really heavy weight and it is something that we know will cause us issues.
We can always bring back this code

src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -167,10 +246,27 @@ private void CleanThreadLocalState()
...
                 current = MaybeGetCurrentContextStack()?.Head;
 
+                _contextStacksByThreadId.TryRemove(NativeMemory.CurrentThreadStats.InternalId, out _);
+
+                if (_releaser != null)
+                {
+                    // we want to clear our JsonContextPool's current thread's state from the _releaser, but to avoid touching any other states
+                    foreach (var threadIdHolder in _threadIds)
+                    {
+                        if (Interlocked.CompareExchange(ref threadIdHolder.ThreadId, -1, NativeMemory.CurrentThreadStats.InternalId) == NativeMemory.CurrentThreadStats.InternalId)
```
ayende, line 256: I don't think this is a  good idea. How do we know that this is from _our_ context pool and not from another?

ppekrol: Probably @maximburyak should confirm, but `_threadIds` are local to the context pool and we are no longer hijacking the ThreadIdHolders from other pools, we are always creating an instance for this specific pool in `EnsureCurrentThreadContextWillBeReleased` or reuse one from the ones that were previosly created by that pool.

ayende: Yes, you are correct.

src/Sparrow/Json/JsonContextPoolBase.cs
maximburyak: ok, created an issue for that, we'll let a few more nightlies to run with that and we'll get it removed https://issues.hibernatingrhinos.com/issue/RavenDB-12699

PR #8476, opened 2019-01-14
Raven db 12673 cert pass

iftahbe:


src/Raven.Server/Utils/Cli/RavenCli.cs
```diff
@@ -825,26 +825,46 @@ private static bool CommandReplaceClusterCert(List<string> args, RavenCli cli)
...
             try
             {
                 certBytes = File.ReadAllBytes(path);
-                cert = new X509Certificate2(path, password, X509KeyStorageFlags.Exportable | X509KeyStorageFlags.PersistKeySet | X509KeyStorageFlags.MachineKeySet);
+                cert.Import(certBytes, password, X509KeyStorageFlags.Exportable | X509KeyStorageFlags.MachineKeySet);
             }
             catch (Exception e)
             {
-                WriteError("Failed to load the provided certificate. Please check the path and password." + e, cli);
+                throw new ArgumentException("Failed to load the provided certificate. Please check the path and password.", e);
```
ppekrol: Don't we need to use `WriteError` here?

src/Raven.Server/Utils/Cli/RavenCli.cs
```diff
@@ -825,26 +825,46 @@ private static bool CommandReplaceClusterCert(List<string> args, RavenCli cli)
...
+            if (string.IsNullOrWhiteSpace(password) == false)
+            {
+                try
+                {
+                    // With the private key, without the password
+                    certBytes = cert.Export(X509ContentType.Pkcs12);
+                }
+                catch (Exception e)
+                {
+                    throw new ArgumentException("Failed to export the password provided certificate.", e);
```
ppekrol: Don't we need to use WriteError here?

ppekrol: And return `false` ?

src/Raven.Server/Utils/Cli/RavenCli.cs
ppekrol: And return `false` ?

ppekrol (approved):

ayende (approved):

PR #8477, opened 2019-01-14
Raven db 8355

ml054:


src/Raven.Server/Documents/Handlers/Admin/AdminSortersHandler.cs
```diff
@@ -53,6 +53,8 @@ public async Task Put()
 
                 await Database.RachisLogIndexNotifications.WaitForIndexNotification(index, ServerStore.Engine.OperationTimeout);
 
+                NoContentStatus();
```
ppekrol, line 56: We are setting 'Created' 2 lines below?

src/Raven.Server/Documents/Handlers/SortersHandler.cs
```diff
@@ -0,0 +1,65 @@
...
+                var sorters = record.Sorters;
+                
+                using (var writer = new BlittableJsonTextWriter(context, ResponseBodyStream()))
+                {
+                    
+                    writer.WriteStartObject();
+
+                    if (sorters != null)
+                    {
+                        writer.WriteArray(context, "Sorters", sorters.Values, (w, c, sorter) =>
```
ppekrol: we want to have a property `Sorters` always, in worst case with empty array

src/Raven.Server/Documents/Handlers/Admin/AdminSortersHandler.cs
ml054: no content also removed application/json. It is common pattern in controllers. Although the method name is a bit misleading.

src/Raven.Server/Documents/Handlers/SortersHandler.cs
ml054: done

PR #8478, opened 2019-01-14
RavenDB-12280 [WIP][MERGABLE] Client side support for graph query

talweiss1982:
@ppekrol @ayende @maximburyak @myarichuk 
Please review this PR, it is still incomplete, I need to add assertions and properly deal with aggregating parameters and even then we need to expend the functionality but first, we need a baseline to start with.

### DIFF
--- a/src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
+++ b/src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
@@ -92,6 +92,10 @@ namespace Raven.Client.Documents.Session
 
         protected LinkedList<QueryToken> OrderByTokens = new LinkedList<QueryToken>();
 
+        protected LinkedList<QueryToken> WithTokens = new LinkedList<QueryToken>();
+
+        protected QueryToken GraphRawQuery { get; private set; }
+
         /// <summary>
         ///   which record to start reading from
         /// </summary>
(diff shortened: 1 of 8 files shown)


talweiss1982:
I have a problem with adding IRavenQueryable in the "With" method, I can't access the DocumentQuery instance, therefore, I can't enforce rules on it nor can I aggregate the parameters into the parent DocumentQuery.
Thoughts?

ppekrol:
```
IGraphQuery<T> With<TOther>(string alias, IRavenQueryable<TOther> query);
```


You are reffering to this?
I think it should be `IDocumentQuery` instead of `IRavenQueryable`. We do not want to mix both of our query types (LINQ and IDocumentQuery).

talweiss1982:
@ppekrol yes this seems to be the viable solution but it is not as nice of an interface

ppekrol:
You should look into builders e.g.
https://github.com/ravendb/ravendb/blob/v4.1/src/Raven.Client/Documents/Queries/MoreLikeThis/MoreLikeThisFactory.cs
or
https://github.com/ravendb/ravendb/blob/v4.1/src/Raven.Client/Documents/Queries/Facets/FacetFactory.cs

ayende:
I think that users will want to mostly pass Linq expression into this.
We will need another overload for `IDocumetnQuery`, and there is `aysnc` version to consider :-(

ppekrol:
We should not mix both of them. LINQ expression can be added at later stage to session.Query, not session.Advanced.DocumentQuery

maximburyak:
more relevant pr exists: https://github.com/ravendb/ravendb/pull/8543

PR #8480, opened 2019-01-15
RavenDB-12689: Size might be taken from existing DataDir of old test,…

aviviadi:
… so delete dir before test starts


ppekrol:
Closing this one down. We will do 4.2 to 4.1 merge.

PR #8493, opened 2019-01-16
RavenDB-12686 moved all running queries to query runner

ppekrol:


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -2852,30 +2860,27 @@ private static void ThrowInvalidField(string f)
...
         }
 
-        private QueryDoneRunning MarkQueryAsRunning(IIndexQuery query, OperationCancelToken token)
+        private IndexQueryDoneRunning MarkQueryAsRunning(IIndexQuery query)
         {
-            var queryStartTime = DateTime.UtcNow;
-            var queryId = Interlocked.Increment(ref _numberOfQueries);
-
-            if (queryId == 1 && _didWork == false)
+            if (_firstQuery && _didWork == false)
```
ayende, line 2865: I assume that we don't need to worry about concurrent queries here?

ppekrol: This was done to set a timeout on first batch of newly created index (test RavenDB_8847). I've assumed that even when we have more then 1 concurrent query, then it is still an acceptable behavior.

ayende: ok

PR #8495, opened 2019-01-16
Raven db 12456

ml054:


ppekrol (approved):

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -921,17 +922,27 @@ private IEnumerable<FullBackup.StorageEnvironmentInformation> GetAllStoragesForB
...
                             smugglerResult = smuggler.Execute();
                         }
+
+                        switch (outputStream)
+                        {
+                            case EncryptingXChaCha20Poly1305Stream encryptedStream:
+                                encryptedStream.Flush(flushToDisk: true);
+                                break;
+                            case Stream fileStream:
+                                fileStream.Flush();
```
ayende, line 932: Also `flushToDisk: true` if the `outputStream` is a `FileStream`

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -968,6 +979,18 @@ private IEnumerable<FullBackup.StorageEnvironmentInformation> GetAllStoragesForB
                         // end of values
 
                         writer.WriteEndObject();
+                        writer.Flush();
+                        switch (outputStream)
+                        {
+                            case EncryptingXChaCha20Poly1305Stream encryptedStream:
+                                encryptedStream.Flush(flushToDisk: true);
+                                break;
+                            case Stream fileStream:
```
ayende, line 988: Same here, `FileStream` handling

src/Raven.Server/Documents/PeriodicBackup/Restore/RestoreBackupTask.cs
```diff
@@ -545,13 +553,32 @@ private List<string> GetFilesForRestore(string backupLocation)
...
-                return fileStream;
+            using (_serverStore.ContextPool.AllocateOperationContext(out TransactionOperationContext ctx))
+            using (ctx.OpenReadTransaction())
+            {
+                if (_restoreConfiguration.BackupEncryptionSettings?.Key != null)
+                {
+                    return new DecryptingXChaCha20Oly1305Stream(fileStream, Convert.FromBase64String(_restoreConfiguration.BackupEncryptionSettings.Key));
+                }
+                var key = _serverStore.GetSecretKey(ctx, database);
+                if (key != null)
```
ayende, line 566: Doesn't this mean that encrypted db always restore from encrypted database?
But we are restoring the db at this point, so I don't see how this can be possible? Where is that key coming from?

src/Raven.Server/Documents/DocumentDatabase.cs
efratshenhar: FileStream is a stream. changed the name in https://github.com/ravendb/ravendb/pull/8506

src/Raven.Server/Documents/DocumentDatabase.cs
efratshenhar: FileStream is a stream. changed the name in https://github.com/ravendb/ravendb/pull/8506

src/Raven.Server/Documents/PeriodicBackup/Restore/RestoreBackupTask.cs
efratshenhar: In https://github.com/ravendb/ravendb/pull/8506 I changed this method.
The key is from _restoreConfiguration.EncryptionKey . we do  '_serverStore.PutSecretKey(_restoreConfiguration.EncryptionKey,...' in Execute

ayende: So we need to check on `FileStream` and then call `Flush(flushToDisk: true)` here, no?

PR #8496, opened 2019-01-16
RavenDB-11818 cleanup old certificates from user intermediates store

iftahbe:


src/Raven.Server/ServerWide/SecretProtection.cs
```diff
@@ -543,10 +543,37 @@ private static void AddCertificateChainToTheUserCertificateAuthority(X509Certifi
...
                         continue;
+                    
+                    // Remove all expired certs which have the same subject name as our own
+                    var chain = new X509Chain();
+                    chain.Build(c);
 
-                    userIntermediateStore.Add(cert);
+                    foreach (var element in chain.ChainElements)
+                    {
+                        if (element.Certificate.NotAfter.ToUniversalTime() > DateTime.Now && element.Certificate.NotBefore.ToUniversalTime() < DateTime.Now)
```
ppekrol: I think we should use `DateTime.UtcNow` ?
Also it would be better to introduce variable before all of the loops and use one value

ppekrol (changes requested):

ppekrol (approved):

PR #8498, opened 2019-01-16
RavenDB-9755 BlittableJsonReaderObject.cs : remove GetPropertiesByIns…

ayende:
…ertionOrder()


ayende:
This is WIP, to see what affect it has on the rest of the system

PR #8505, opened 2019-01-17
RavenDB-12707 - DiskFullException is identified as an unexpected exception

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-12707


ayende (approved):

ayende:
Where are they expected / unexpected?

ppekrol:
We do not want to wrap those inside SystemException, because when rethrowing we will not handle them well in our own code

PR #8518, opened 2019-01-18
Optimizations for processes normally happening on the indexing process

redknightlois:


ayende:
Can you look at the test failures?

redknightlois:
Yes. Tomorrow, that is why I issue the PR running slow locally is hell. I rearranged some blocks, so that was kinda expected, but fast test didn't show any issues though.

src/Raven.Server/Utils/TypeConverter.cs
```diff
@@ -330,85 +332,119 @@ private static IEnumerable<object> Flatten(IEnumerable items)
...
+                return dto;
 
-                if (type == null)
-                    return blittableJsonObject;
+            if (LazyStringParser.TryParseTimeSpan(buffer, size, out TimeSpan ts))
+                return ts;
 
-                if (type.StartsWith(TypeList) == false)
-                    return blittableJsonObject;
+            return value; // ensure that the decompressed lazy string value is returned
```
maximburyak, line 385: I'm not sure how we can track that, but for the parsing above, we certainly should dispose of the LSV

src/Sparrow/Json/BlittableJsonReaderObject.cs
```diff
@@ -293,6 +296,102 @@ private static void ThrowFormatException(object value, string fromType, string t
...
+
+                try
+                {
+                    if (type.GetTypeInfo().IsEnum)
+                    {
+                        obj = (T)Enum.Parse(type, result.ToString());
+                    }
+                    else if (type == typeof(DateTime))
+                    {
+                        if (ChangeTypeToString(result, out string dateTimeString) == false)
```
maximburyak, line 331: Can we use span here?

src/Raven.Server/Utils/TypeConverter.cs
redknightlois: Type converter is not the owner. Caller should know best. The semantics of this method is unclear, so on an optimization pass I wouldn't change the semantics. We should open a new ticket to figure out if we can allow destructive semantics on type converters and modify all of them accordingly

src/Sparrow/Json/BlittableJsonReaderObject.cs
redknightlois: Probably, but wouldn't be clean anyways because right now the core is unaware or span. The spanification of the core should deal with that and treat span as first class citizen starting wit voron and sparrow

src/Raven.Server/Utils/TypeConverter.cs
ayende: For the most part, we expect this to be called from indexing / projection code, so the target isn't a dynamically allocated string but rather a pointer to a blittable

src/Sparrow/Json/BlittableJsonReaderObject.cs
ayende: We can already use span, we have already started to take dependency on it.
So we can technically get a `Span<byte>` and then use `Utf8Parser.TryParse` directly

ayende: Not sure how much that would help, the common case is a string, not date time

redknightlois: Yes, I have several cases where I am already working with spans already, but they are still a clutch that require the same method to be written in 2 different APIs. We need a definite solution. The generalized allocators are giving us spans for arbitrary memory with a decent API. That along with the ability to customize heavily the behavior is the big change there.

redknightlois: For example, if every call with a pointer you need to write `new ReadOnlySpan<byte>(ptr)` the code gets unwildly very fast.

src/Raven.Server/Utils/TypeConverter.cs
maximburyak: @redknightlois you are right, I was refering to the usage of the after compressed string decompression, thing is that "here" we know "for sure" that the value won't be used

redknightlois: The caller knows it won't be used, not the method. Callers responsibility to deal with it

redknightlois: Clear responsibility and usage uniformity is key when handling memory manually. Having a non uniform destroy semantics here is just asking for trouble

maximburyak: So is there a way we could avoid a "temporary leak" of the uncompressed string? or it's not a viable scenario in the first place (date string is too short to be compressed?)

ayende: what temporary leak? And strings are only compressed from 128bytes

maximburyak: The "possible" leak was when we inflated the compressed string, but returned a datetime, or timespan, but given that strings are compressed only from 128bytes size, it's irrelevant for those types of input

redknightlois: Also, remember that those leaks get healed up on context reset, so they should be typically small anyways.

maximburyak: @redknightlois that's correct

PR #8523, opened 2019-01-20
RavenDB-12721 - Wrong Threads CPU usage calculation

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-12721

### DIFF
--- a/src/Raven.Server/Utils/ThreadsUsage.cs
+++ b/src/Raven.Server/Utils/ThreadsUsage.cs
@@ -19,6 +19,11 @@ namespace Raven.Server.Utils
         {
             using (var process = Process.GetCurrentProcess())
             {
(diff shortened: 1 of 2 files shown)


src/Raven.Server/Utils/ThreadsUsage.cs
```diff
@@ -127,7 +133,7 @@ private static IEnumerable<ProcessThread> GetProcessThreads(Process process)
         {
             try
             {
-                return process.Threads.Cast<ProcessThread>();
+                return process.Threads.Cast<ProcessThread>().ToList();
```
ppekrol, line 136: method return type should be changed to `List<ProcessThread>`

grisha-kotler: done

PR #8525, opened 2019-01-21
RavenDB-12526 remove cell coloring on sorted tables

ml054:


ml054:
I will open this for 4.1

PR #8531, opened 2019-01-21
RavenDB-11818 ignore access denied when cleaning certs from intermedi…

iftahbe:
…ates store

### DIFF
--- a/src/Raven.Server/ServerWide/SecretProtection.cs
+++ b/src/Raven.Server/ServerWide/SecretProtection.cs
@@ -580,8 +580,8 @@ namespace Raven.Server.ServerWide
                         catch (CryptographicException e)
                         {
                             // Access denied?
-                            if (Logger.IsOperationsEnabled)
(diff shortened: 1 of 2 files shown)


src/Raven.Server/ServerWide/SecretProtection.cs
```diff
@@ -573,7 +573,16 @@ public static void AddCertificateChainToTheUserCertificateAuthorityStoreAndClean
...
-                        userIntermediateStore.Remove(element.Certificate);                        
+                        try
+                        {
+                            userIntermediateStore.Remove(element.Certificate);
+                        }
+                        catch (CryptographicException e)
+                        {
+                            // Access denied?
+                            if (Logger.IsOperationsEnabled)
+                                Logger.Operations($"Tried to clean expired certificates from the OS user intermediate store but got an exception when removing a certificate with subject name: {element.Certificate.SubjectName.Name}.", e);
```
ayende: Add the thumbprint as well.
And this should go to info, not ops

PR #8532, opened 2019-01-21
RavenDB-11818 ignore access denied when cleaning certs from intermedi…

iftahbe:
…ates store

### DIFF
--- a/src/Raven.Server/ServerWide/SecretProtection.cs
+++ b/src/Raven.Server/ServerWide/SecretProtection.cs
@@ -575,7 +575,16 @@ namespace Raven.Server.ServerWide
                     {
                         if (element.Certificate.NotAfter.ToUniversalTime() > utcNow && element.Certificate.NotBefore.ToUniversalTime() < utcNow)
                             continue;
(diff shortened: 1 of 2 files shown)


src/Raven.Server/ServerWide/SecretProtection.cs
```diff
@@ -575,7 +575,16 @@ public static void AddCertificateChainToTheUserCertificateAuthorityStoreAndClean
                     {
                         if (element.Certificate.NotAfter.ToUniversalTime() > utcNow && element.Certificate.NotBefore.ToUniversalTime() < utcNow)
                             continue;
-                        userIntermediateStore.Remove(element.Certificate);                        
+                        try
+                        {
+                            userIntermediateStore.Remove(element.Certificate);
```
ppekrol, line 580: @iftahbe the loop can continue after catching the exception? Can we succeed on some of the certificates and fail on others or we will just log a lot of exceptions?

iftahbe: Yes we can fail on one and succeed at the other. I believe this is a very rare case. Anyway, I'm moving the logging to Info.

PR #8536, opened 2019-01-22
RavenDB-12299: Snapshot restore should fail before actual restore, if…

garayx:
… there is not enough disk space.

### DIFF
--- a/src/Raven.Server/Web/System/AdminDatabasesHandler.cs
+++ b/src/Raven.Server/Web/System/AdminDatabasesHandler.cs
@@ -11,7 +11,6 @@ using System.IO;
 using System.IO.Compression;
 using System.Linq;
 using System.Net;
-using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http.Features.Authentication;
@@ -40,10 +39,10 @@ using Raven.Server.Smuggler.Documents;
 using Raven.Client.Extensions;
 using Raven.Client.ServerWide.Operations.Migration;
 using Raven.Server.Config;
+using Raven.Server.Config.Settings;
 using Raven.Server.Documents.Indexes;
 using Raven.Server.Documents.Indexes.Auto;
 using Raven.Server.Documents.PeriodicBackup.Restore;
-using Raven.Server.Extensions;
 using Raven.Server.ServerWide.Commands.Indexes;
 using Raven.Server.Utils;
 using Raven.Server.Web.Studio;
@@ -547,17 +546,29 @@ namespace Raven.Server.Web.System
                     catch (Exception e)
                     {
                         if (e is InvalidDataException)
-                            throw new InvalidDataException($"Invalid snapshot file: {restoreConfigurationJson.LastFileNameToRestore} at path : {zipPath}");
-                        if (e is FileNotFoundException)
-                            throw new FileNotFoundException($"Could not find file: {restoreConfigurationJson.LastFileNameToRestore} at path: {zipPath}");
+                        {
+                            if (Logger.IsOperationsEnabled)
+                                Logger.Operations($"Restore database from snapshot operation failed. Invalid snapshot file {restoreConfigurationJson.LastFileNameToRestore}", e);
+                            throw new InvalidDataException($"Invalid snapshot file {restoreConfigurationJson.LastFileNameToRestore} at {zipPath}", e);
+                        }
 
-                        throw new IOException($"Error reading snapshot file: {restoreConfigurationJson.LastFileNameToRestore}, at path: {zipPath}. Please provide a valid snapshot file.");
+                        if (e is FileNotFoundException)
+                        {
+                            if (Logger.IsOperationsEnabled)
+                                Logger.Operations($"Restore database from snapshot operation failed. Could not find file {restoreConfigurationJson.LastFileNameToRestore}", e);
+                            throw new FileNotFoundException($"Could not find file {restoreConfigurationJson.LastFileNameToRestore} at {zipPath}", e);
+                        }
+
+                        if (Logger.IsOperationsEnabled)
+                            Logger.Operations($"Restore database from snapshot operation failed. Error reading snapshot file {restoreConfigurationJson.LastFileNameToRestore}", e);
+
+                        throw new IOException($"Error reading snapshot file {restoreConfigurationJson.LastFileNameToRestore}, at {zipPath}. Please provide a valid snapshot file.", e);
                     }
 
                     var baseDataDirectory = ServerStore.Configuration.Core.DataDirectory.FullPath;
 
                     var destinationPath = string.IsNullOrEmpty(restoreConfigurationJson.DataDirectory) == false 
-                        ? PathUtil.ToFullPath(restoreConfigurationJson.DataDirectory, baseDataDirectory) 
+                        ? new PathSetting(restoreConfigurationJson.DataDirectory, baseDataDirectory).FullPath
                         : RavenConfiguration.GetDataDirectoryPath(ServerStore.Configuration.Core, databaseName, ResourceType.Database);
 
                     var drivesInfo = PlatformDetails.RunningOnPosix ? DriveInfo.GetDrives() : null;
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -531,6 +533,39 @@ public async Task RestoreDatabase()
...
+                {
+                    long backupSizeInBytes;
+                    try
+                    {
+                        using (var zip = ZipFile.OpenRead(Path.Combine(restoreConfigurationJson.BackupLocation, restoreConfigurationJson.LastFileNameToRestore)))
+                            backupSizeInBytes = zip.Entries.Sum(entry => entry.Length);
+                    }
+                    catch
+                    {
+                        throw new IOException($"Could not open the backup file, provided file : {Path.Combine(restoreConfigurationJson.BackupLocation, restoreConfigurationJson.LastFileNameToRestore)} Please provide a valid backup file.");
```
ayende: You need to also include _the actual reason you failed here_, otherwise, the user is left guessing

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -531,6 +533,39 @@ public async Task RestoreDatabase()
...
+                    }
+
+                    var destinationPath = string.IsNullOrWhiteSpace(restoreConfigurationJson.DataDirectory)
```
ayende: There is a LOT going on in this statement, make it into its own function and use `if` to clearly show what is going on there.

I'm not following why you have `Combine` there

ayende (changes requested):

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -531,6 +533,39 @@ public async Task RestoreDatabase()
...
+                        ? Path.Combine(ServerStore.Configuration.Core.DataDirectory.FullPath, restoreConfigurationJson.DatabaseName)
+                        : Directory.Exists(restoreConfigurationJson.DataDirectory)
+                            ? restoreConfigurationJson.DataDirectory
+                            : Path.Combine(ServerStore.Configuration.Core.DataDirectory.FullPath, restoreConfigurationJson.DataDirectory);
```
ppekrol: Are we combining this correctly? One `DataDirectory` with another one?

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -531,6 +533,39 @@ public async Task RestoreDatabase()
...
+                        using (var zip = ZipFile.OpenRead(Path.Combine(restoreConfigurationJson.BackupLocation, restoreConfigurationJson.LastFileNameToRestore)))
+                            backupSizeInBytes = zip.Entries.Sum(entry => entry.Length);
+                    }
+                    catch
+                    {
+                        throw new IOException($"Could not open the backup file, provided file : {Path.Combine(restoreConfigurationJson.BackupLocation, restoreConfigurationJson.LastFileNameToRestore)} Please provide a valid backup file.");
+                    }
+
+                    var destinationPath = string.IsNullOrWhiteSpace(restoreConfigurationJson.DataDirectory)
+                        ? Path.Combine(ServerStore.Configuration.Core.DataDirectory.FullPath, restoreConfigurationJson.DatabaseName)
```
ppekrol: instead of doing that you should use `ServerStore.Configuration.Core.DataDirectory.Combine('')`

ppekrol (changes requested):

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
garayx: done

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
garayx: done

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
garayx: done

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
garayx: done

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -531,6 +533,46 @@ public async Task RestoreDatabase()
...
+
+                    try
+                    {
+                        using (var zip = ZipFile.OpenRead(zipPath))
+                            backupSizeInBytes = zip.Entries.Sum(entry => entry.Length);
+                    }
+                    catch (Exception e)
+                    {
+                        if (e is InvalidDataException)
+                            throw new InvalidDataException($"Invalid snapshot file: {restoreConfigurationJson.LastFileNameToRestore} at path : {zipPath}");
```
ppekrol: I think we need to put `e` as inner exception in all of those exceptions here.

ppekrol: Let's avoid multiple `:` in the message. Maybe change to sth like that:
```
$"Invalid snapshot file '{restoreConfigurationJson.LastFileNameToRestore}' at '{zipPath}' path"
```

This should isolate the path nicely (need to apply to other messages also sth similar)

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -531,6 +533,46 @@ public async Task RestoreDatabase()
...
+                        if (e is FileNotFoundException)
+                            throw new FileNotFoundException($"Could not find file: {restoreConfigurationJson.LastFileNameToRestore} at path: {zipPath}");
+
+                        throw new IOException($"Error reading snapshot file: {restoreConfigurationJson.LastFileNameToRestore}, at path: {zipPath}. Please provide a valid snapshot file.");
+                    }
+
+                    var baseDataDirectory = ServerStore.Configuration.Core.DataDirectory.FullPath;
+
+                    var destinationPath = string.IsNullOrEmpty(restoreConfigurationJson.DataDirectory) == false 
+                        ? PathUtil.ToFullPath(restoreConfigurationJson.DataDirectory, baseDataDirectory)
```
ppekrol: can we use `new PathSetting(restoreConfigurationJson.DataDirectory, baseDataDirectory).FullPath` here?

We do some validation inside that class.

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -531,6 +533,46 @@ public async Task RestoreDatabase()
...
+                    var drivesInfo = PlatformDetails.RunningOnPosix ? DriveInfo.GetDrives() : null;
+                    var destinationDirInfo = DiskSpaceChecker.GetDriveInfo(destinationPath, drivesInfo, out _);
+                    var destinationDriveInfo = DiskSpaceChecker.GetDiskSpaceInfo(destinationDirInfo.DriveName);
+
+                    if (destinationDriveInfo == null)
+                        throw new ArgumentException($"Provided path starts with an invalid drive name. Please use a proper path. Drive name provided: {destinationDirInfo.DriveName}.");
+
+                    var desiredFreeSpace = Size.Min(new Size(512, SizeUnit.Megabytes), destinationDriveInfo.TotalSize * 0.01) + new Size(backupSizeInBytes, SizeUnit.Bytes);
+
+                    if (destinationDriveInfo.TotalFreeSpace <= desiredFreeSpace)
```
ppekrol: Should be `<` instead of `<=`

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
garayx: done

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
garayx: done

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
garayx: done

ppekrol (approved):

PR #8538, opened 2019-01-22
RavenDB-12725 If journal was not synced then we must not remove it on…

arekpalinski:
… recovery


arekpalinski:
`CanSurviveRestart` tests is failing - under investigation

ayende (approved):

ayende:
Okay, I have made some changes, we are not _not_ deleting the files on startup and we are scheduling them to be synced.
Fast tests pass on my machine, but I want to see what the slow tests are doing

ayende:
On my end, lots of failing tests:

* SlowTests.Voron.Backups.Incremental.IncrementalBackupShouldCreateConsecutiveJournalFiles
* SlowTests.Voron.Backups.Incremental.CanDoMultipleIncrementalBackupsAndRestoreOneByOne
* SlowTests.Voron.Backups.Incremental.IncrementalBackupShouldCopyJustNewPagesSinceLastBackup

ayende:
Hopefully this should fix the tests

src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -76,6 +76,10 @@ public WriteAheadJournal(StorageEnvironment env)
                 var currentJournal = journalFilesCount > 0 ? _journalIndex : -1;
                 header->Journal.CurrentJournal = currentJournal;
                 header->Journal.JournalFilesCount = journalFilesCount;
+                if(journalFilesCount == 0)
+                {
+                    header->Journal.LastSyncedJournal = -1;
+                }
```
arekpalinski: I'm a bit worried about that change. Note we call `_updateLogInfo` action from `UpdateFileHeaderAfterDataFileSync`

https://github.com/ravendb/ravendb/blob/d476f61ccb2ffbc77fe95772c3599183a1204097/src/Voron/Impl/Journal/WriteAheadJournal.cs#L1216-L1226

and this can override potentially `LastSyncedJournal` we set here:

https://github.com/ravendb/ravendb/blob/d476f61ccb2ffbc77fe95772c3599183a1204097/src/Voron/Impl/Journal/WriteAheadJournal.cs#L1221

BTW The case that `JournalFilesCount == 0` might be tricky for us during recovery. See here:
https://github.com/ravendb/ravendb/blob/d476f61ccb2ffbc77fe95772c3599183a1204097/src/Voron/Impl/Journal/WriteAheadJournal.cs#L163-L167
 I'm gonna double check that - I'm worried that we might still have some journals that were not synced and we should to run the recovery

arekpalinski: Please confirm, but I think you did it specifically for compaction, didn't you? For that purpose I think that would be ok because we force syncing, but might be not to accomplish that so probably only if we `synced` successfully:

https://github.com/ravendb/ravendb/blob/69018d823cb89f4d0bbb9c636b82d14110250f8d/src/Voron/Impl/Compaction/StorageCompaction.cs#L87-L88

src/Voron/Impl/FileHeaders/HeaderAccessor.cs
```diff
@@ -101,6 +96,28 @@ public bool Initialize()
                 }
                 _revision = _theHeader->HeaderRevision;
 
+                if (_theHeader->Version != Constants.CurrentVersion)
+                {
+                    _locker.ExitWriteLock();
```
ayende, line 102: Why are you exiting the write lock here?

arekpalinski: We're already under write lock so we release it so the updater can access the header accessor. We're in the constructor of storage env so nobody else can touch it anyway.

src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -934,6 +936,10 @@ private bool GatherInformationToStartSync()
                         return false; // we have already disposed, nothing to do here
 
                     _lastFlushed = _parent._lastFlushed;
+
+                    if (_lastFlushed.IsValid == false)
```
arekpalinski, line 940: I got NRE on `_lastFlushed.Journal.SetLastReadTxHeader` (`Journal` was `null`, test `RavenDB_10420`) after I added the suggestion to sync data file after recovery. I did it by adding the following at the end of ctor of `StorageEnvironment`:

```
if (isNew == false && _options.ManualSyncing == false)
	SuggestSyncDataFile(); // let's suggest syncing data file after the recovery
```

The issue was that doing a sync _before_ we flushed anything isn't actually valid because `_lastFlushed` wasn't defined in `UpdateJournalStateUnderWriteTransactionLock`. So I decided to not scheduling the sync immediately after loading a database. We'll do it after the first flush.

arekpalinski:
All changes squashed in https://github.com/ravendb/ravendb/pull/8556

src/Voron/Impl/Journal/WriteAheadJournal.cs
ayende: In that case, we need to do a suggested flush, anyway, no?
I don't think it would be triggered until the first commit otherwise, no?

arekpalinski: The recovery on startup is actually the same as flush - it writes data from journals to data file but it's unsynced stuff. So maybe we should set `JournalApplicator._lastFlushed` at the end of the recovery? Like we do at the end of flush?

Then we can safely suggest data sync after the recovery.

ayende: Yes, I think we should set that one.

PR #8541, opened 2019-01-22
RavenDB-12648 - update the ServicePoint ConnectionLimit

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-12648

### DIFF
--- a/src/Raven.Client/Http/RequestExecutor.cs
+++ b/src/Raven.Client/Http/RequestExecutor.cs
@@ -28,6 +28,7 @@ using Raven.Client.Properties;
 using Raven.Client.ServerWide.Commands;
 using Raven.Client.Util;
 using Sparrow;
+using Sparrow.Collections;
 using Sparrow.Json;
 using Sparrow.Logging;
 using Sparrow.Threading;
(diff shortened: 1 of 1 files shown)


ayende (commented):
I don't like that we modify something global for the user. 
Why do we need to do this and not let the user handle the situation?
Are there any issues with setting to to this high a value? That is very likely to cause a LOT of connections, no?

grisha-kotler:
This will allow more concurrent connections.
In .Net core the _MaxConnectionsPerServer_ is being used (in the HttpClientHandler). The default is int.MaxValue.
In .Net framework the _ServicePointManager.ConnectionLimit_ is being used. The default is: 2.

ayende:
Okay

ppekrol:
Also worth noting here is that we did the same thing in 3.x. We used `256` there as a number, but @grisha-kotler and I agreed, that since .NET Core runtime is using `int.Max` as a default, then it would be better to increase the value from 256 to 1024 * 10. Should be sufficient.

ppekrol (approved):

PR #8543, opened 2019-01-22
RavenDB-12280  Client side support for graph query

talweiss1982:


ayende (approved):
@talweiss1982 I _love_ it. This is simple, straightforward, and solve the problem very elegantly.
What is more, it doesn't run into many of the issues that I thought we would run into in this mode.

src/Raven.Client/Documents/Session/AsyncDocumentQuery.cs
```diff
@@ -1011,5 +1082,25 @@ public IAsyncDocumentQuery<TResult> AsyncQuery<TResult>(string indexName, string
...
+            var docQuery = (AsyncDocumentQuery<TOther>)queryInspector.GetAsyncDocumentQuery(x => x.ParameterPrefix = $"w{WithTokens.Count}p");
+            if (docQuery.SelectTokens?.Count > 0)
+            {
+                throw new NotSupportedException($"Select is not premitted in a 'With' clause in query:{docQuery}");
+            }
+
+
+            foreach (var keyValue in docQuery.QueryParameters)
+            {
+                QueryParameters.Add(keyValue.Key, keyValue.Value);
```
ppekrol, line 1145: Maybe we should use `AddParameter` method. I do not remember if we are doing some validation there or not.

src/Raven.Client/Documents/Session/DocumentQuery.cs
```diff
@@ -774,6 +844,26 @@ IDocumentQuery<TResult> IDocumentQuery<T>.OfType<TResult>()
...
+            var docQuery = (DocumentQuery<TOther>)queryInspector.GetDocumentQuery(x => x.ParameterPrefix = $"w{WithTokens.Count}p");
+            if (docQuery.SelectTokens?.Count > 0)
+            {
+                throw new NotSupportedException($"Select is not premitted in a 'With' clause in query:{docQuery}");
+            }
+
+
+            foreach (var keyValue in docQuery.QueryParameters)
+            {
+                QueryParameters.Add(keyValue.Key, keyValue.Value);
```
ppekrol, line 907: Maybe we should use `AddParameter` method. I do not remember if we are doing some validation there or not.

src/Raven.Client/Documents/Session/Tokens/WithToken.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        public WithToken(string alias, string query)
+        {
+            this.alias = alias;
+            _query = query;
+        }
+
+        public override void WriteTo(StringBuilder writer)
+        {
+            writer.Append($"with {{{_query}}} as {alias}");
```
ppekrol: Let's skip interpolation and split it to few writer calls.

src/Raven.Client/Documents/Session/IAsyncDocumentQuery.cs
```diff
@@ -67,6 +67,13 @@ public interface IAsyncRawDocumentQuery<T> :
     {
     }
 
+    public interface IAsyncGraphQuery<T> :
+        IQueryBase<T, IAsyncGraphQuery<T>>,
+        IAsyncDocumentQueryBase<T>
+    {
+        IAsyncGraphQuery<T> With<TOther>(string alias, IRavenQueryable<TOther> query, CancellationToken token = default);
```
ppekrol: I think we do not need cancellation token here

ppekrol: Let's remember to add the overload with `IDocumentQuery<TOther>`

ppekrol (commented):
Great that we figured out an elegant and simple way to handle this.

src/Raven.Client/Documents/Session/AsyncDocumentQuery.cs
```diff
@@ -1011,5 +1082,25 @@ public IAsyncDocumentQuery<TResult> AsyncQuery<TResult>(string indexName, string
...
         }
+
+        public IAsyncGraphQuery<T> With<TOther>(string alias, IRavenQueryable<TOther> query, CancellationToken token = default)
+        {
+            //TODO:Find a proper way of do this extraction
+            var queryInspector = (RavenQueryInspector<TOther>)query;
+            var docQuery = (AsyncDocumentQuery<TOther>)queryInspector.GetAsyncDocumentQuery(x => x.ParameterPrefix = $"w{WithTokens.Count}p");
+            if (docQuery.SelectTokens?.Count > 0)
+            {
+                throw new NotSupportedException($"Select is not premitted in a 'With' clause in query:{docQuery}");
```
ppekrol: typo:

```
$"Select is not permitted in a 'With' clause in query: {docQuery}"
```

src/Raven.Client/Documents/Session/DocumentQuery.cs
```diff
@@ -774,6 +844,26 @@ IDocumentQuery<TResult> IDocumentQuery<T>.OfType<TResult>()
...
         }
 
+        IGraphQuery<T> IGraphQuery<T>.With<TOther>(string alias, IRavenQueryable<TOther> query)
+        {
+            //TODO:Find a proper way of do this extraction
+            var queryInspector = (RavenQueryInspector<TOther>)query;
+            var docQuery = (DocumentQuery<TOther>)queryInspector.GetDocumentQuery(x => x.ParameterPrefix = $"w{WithTokens.Count}p");
+            if (docQuery.SelectTokens?.Count > 0)
+            {
+                throw new NotSupportedException($"Select is not premitted in a 'With' clause in query:{docQuery}");
```
ppekrol: typo:

```
$"Select is not permitted in a 'With' clause in query: {docQuery}"
```

src/Raven.Client/Documents/Session/IAsyncDocumentQuery.cs
ayende: No, we don't, that isn't executing

src/Raven.Client/Documents/Session/AsyncDocumentQuery.cs
talweiss1982: I need to set the key of the parameter in the "parent" query the AddParameter only gets the value of the parameter.
The keys are promised to be unique so this will not throw (since we never remove parameters from the dictionary and I use the count of values and with statements)

src/Raven.Client/Documents/Session/Tokens/WithToken.cs
talweiss1982: It will be less readable but okay

src/Raven.Client/Documents/Session/IAsyncDocumentQuery.cs
talweiss1982: I need to see if my solution works for IDocumentQuery since the tokens might be already populated with parameter names

talweiss1982: I might need to transform the DocumentQuery to a queryable and back so i could modify the tokens. writing a test to check this senario.

talweiss1982: @ayende @ppekrol 
what if we have something like so:
.With("Foo", session.Advanced.AsyncDocumentQuery<Foo>(), q => q.WhereIn(x=>x.Name, names))
The issue here is that i need a clean DocumentQuery, i could replace the empty AsyncDocumentQuery with those three parameters:
string indexName , string collectionName , bool isMapReduce
to construct the query myself but if not then we can't use my approach and we are back to iterating the tokens

ppekrol: We can either drop the support for IDocumentQuery or do:
```
.With("Foo",q => q.DocumentQuery<TIndex>().WhereIn(x=>x.Name, names));
```

The q will be our own 'builder' that will do:
```
public IFilterDocumentQuery<T> DocumentQuery<TIndex>()
{
    var dq = session.Advanced.DocumentQuery<TIndex>();
    dq.PrefixParameters = "abc";
    return dq;
}
```

talweiss1982: @ppekrol okay I have made modification they are still WIP please look at :
IAsyncGraphQuery<T> With<TOther>(string alias, Func<IAsyncDocumentBuilder,IAsyncDocumentQuery<TOther>> queryFactory);

src/Raven.Client/Documents/Session/IAsyncAdvancedSessionOperations.Query.cs
```diff
@@ -11,7 +11,11 @@ namespace Raven.Client.Documents.Session
     /// <summary>
     ///     Advanced async session operations
     /// </summary>
-    public partial interface IAsyncAdvancedSessionOperations
+    public partial interface IAsyncAdvancedSessionOperations : IAsyncDocumentBuilder
+    {
+    }
+
+    public interface IAsyncDocumentBuilder
```
ppekrol: rename to `IAsyncDocumentQueryBuilder`

ppekrol (commented):
I like that.

src/Raven.Client/Documents/Session/IDocumentQuery.cs
```diff
@@ -67,6 +67,16 @@ public interface IRawDocumentQuery<T> :
     {
     }
 
+    public interface IGraphQuery<T> :
+        IQueryBase<T, IGraphQuery<T>>,
+        IDocumentQueryBase<T>, IEnumerable<T>
```
ppekrol, line 72: We should not inherit from `IEnumerable<T>` but I guess this is related to https://issues.hibernatingrhinos.com/issue/RavenDB-10520 and for now we do not have much choice?

ppekrol (approved):
Can you check the failing test? I think it is matter of adding there `IAsyncGraphQuery`

maximburyak:
other PR: https://github.com/ravendb/ravendb/pull/8478 is more relevant

maximburyak:
It looks good, @talweiss1982 will add some more tests soon and we'll be ready to merge

talweiss1982:
okay i had an issue with the edges it is now fixed, added a test.

PR #8545, opened 2019-01-23
RavenDB-10842 Add percentiles to traffic watch

kwiato:


src/Raven.Studio/wwwroot/Content/css/icons.less
```diff
@@ -160,6 +160,7 @@
 .make-icon(lets-encrypt, "\e9df");
 .make-icon(chash, "\e9dd");
 .make-icon(minus, "\e9dc");
+.make-icon(average, "\e9dc");
```
ml054, line 163: Did you include other icon files in this PR? Or average icon was already present?

kwiato: It was already present, I've reused other icon but with different class name.

ml054: oh, I can see it uses the same code as icon-minus. Thanks

ml054 (approved):

PR #8546, opened 2019-01-23
RavenDB-12320 When smuggler imports revisions w/o documents, it creat…

efratshenhar:
…es orphans

### DIFF
--- a/src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
+++ b/src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
@@ -9,6 +9,7 @@ using Raven.Client.Documents.Indexes;
 using Raven.Client.Documents.Operations.Attachments;
 using Raven.Client.Documents.Operations.Counters;
 using Raven.Client.Documents.Smuggler;
+using Raven.Client.Exceptions.Documents;
 using Raven.Client.ServerWide;
 using Raven.Client.Util;
 using Raven.Server.Documents;
@@ -150,6 +151,21 @@ namespace Raven.Server.Smuggler.Documents
 
             public void WriteDocument(DocumentItem item, SmugglerProgressBase.CountsWithLastEtag progress)
             {
+                if (item.Document.NonPersistentFlags.Contain(NonPersistentDocumentFlags.FromSmuggler) &&
+                     ((item.Document.Flags & DocumentFlags.HasRevisions) != 0) &&
+                    ((item.Document.Flags & DocumentFlags.Revision) !=0) || 
+                     ((item.Document.Flags & DocumentFlags.DeleteRevision) != 0))
+                {
+                    using (_database.DocumentsStorage.ContextPool.AllocateOperationContext(out DocumentsOperationContext context))
+                    using (context.OpenReadTransaction())
+                    {
+                        if (_database.DocumentsStorage.Get(context, item.Document.Id) == null)
+                        {
+                            throw new DocumentDoesNotExistException();
+                        }
+                    }
+                }
+
                 if (item.Attachments != null)
                 {
                     if(_options.OperateOnTypes.HasFlag(DatabaseItemType.Attachments))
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Smuggler/Documents/DatabaseSmuggler.cs
```diff
@@ -498,6 +499,20 @@ private SmugglerProgressBase.Counts ProcessRevisionDocuments(SmugglerResult resu
 
                     item.Document.NonPersistentFlags |= NonPersistentDocumentFlags.FromSmuggler;
 
+                    if (item.Document.NonPersistentFlags.Contain(NonPersistentDocumentFlags.FromSmuggler))
```
ppekrol: We need to move this inside the `actions.WriteDocument` for DatabaseDestination.

You can throw `DocumentDoesNotExist` exception there, catch it here and increase the errored count.

src/Raven.Server/Smuggler/Documents/DatabaseSmuggler.cs
```diff
@@ -659,6 +674,16 @@ private SmugglerProgressBase.Counts ProcessCounters(SmugglerResult result)
                     if (result.Counters.ReadCount % 1000 == 0)
                         AddInfoToSmugglerResult(result, $"Read {result.Counters.ReadCount:#,#;;0} counters.");
 
+                    using (_database.DocumentsStorage.ContextPool.AllocateOperationContext(out DocumentsOperationContext context))
```
ppekrol: We need to move this inside the `actions.WriteCounter` for DatabaseDestination.

You can throw `DocumentDoesNotExist` exception there, catch it here and increase the errored count.

ppekrol (changes requested):

src/Raven.Server/Smuggler/Documents/DatabaseSmuggler.cs
efratshenhar: for StreamDestination we don't need it, right?

ppekrol: I do not see a reason.

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
```diff
@@ -150,6 +151,21 @@ public DatabaseDocumentActions(DocumentDatabase database, BuildVersionType build
 
             public void WriteDocument(DocumentItem item, SmugglerProgressBase.CountsWithLastEtag progress)
             {
+                if (item.Document.NonPersistentFlags.Contain(NonPersistentDocumentFlags.FromSmuggler) &&
+                     ((item.Document.Flags & DocumentFlags.HasRevisions) != 0) &&
+                    ((item.Document.Flags & DocumentFlags.Revision) !=0) || 
+                     ((item.Document.Flags & DocumentFlags.DeleteRevision) != 0))
+                {
+                    using (_database.DocumentsStorage.ContextPool.AllocateOperationContext(out DocumentsOperationContext context))
+                    using (context.OpenReadTransaction())
```
ppekrol: Can we avoid opening a ReadTransaction for every call?
It would be better to open transaction when we haven't opened it before and in HandleBatchOfDocumentsIfNecessary when we are actually adding the batch for processing, we can close the transaction

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
```diff
@@ -150,6 +151,21 @@ public DatabaseDocumentActions(DocumentDatabase database, BuildVersionType build
...
+                    {
+                        if (_database.DocumentsStorage.Get(context, item.Document.Id) == null)
+                        {
+                            throw new DocumentDoesNotExistException();
```
ayende: Also, you _must_ provide more details here. For example, what is the missing document?

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
```diff
@@ -710,6 +726,14 @@ private void AddToBatch(CounterDetail counter)
 
             public void WriteCounter(CounterDetail counterDetail)
             {
+                using (_database.DocumentsStorage.ContextPool.AllocateOperationContext(out DocumentsOperationContext context))
+                using (context.OpenReadTransaction())
+                {
+                    if (_database.DocumentsStorage.Get(context, counterDetail.DocumentId) == null)
+                    {
+                        throw new DocumentDoesNotExistException();
```
ayende: Same, not enough info on error

ayende (changes requested):

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
ppekrol: @efratshenhar just pass the Id in the ctor. It will generate nice message.

ppekrol: There is also a second parameter there, you can add a detailed message. We should add that to the progress so user will see what is the source of the error maybe? I think we have sth like `result.AddError` there available.

PR #8548, opened 2019-01-23
RavenDB-12375 more info on error message

iftahbe:


iftahbe:
I'll make a better error with the actual port that needs to be opened on the remote node

PR #8549, opened 2019-01-23
RavenDB-12657 & RavenDB-12665

ppekrol:


grisha-kotler:
Can you add a test with counter increment?

arekpalinski (approved):

src/Raven.Server/Documents/DocumentIdWorker.cs
```diff
@@ -197,19 +206,29 @@ private static ByteStringContext.InternalScope UnlikelyGetLowerUnicode(ByteStrin
...
 
                 _jsonParserState.FindEscapePositionsIn(ptr, ref strLength, escapePositionsSize);
+                if (strLength != originalStrLength)
+                {
+                    var anotherStrLength = originalStrLength;
+                    _jsonParserState.FindEscapePositionsIn(ptr + idSize + maxStrSize, ref anotherStrLength, escapePositionsSize);
+
+#if DEBUG
+                    if (strLength != anotherStrLength)
+                        throw new InvalidOperationException($"String length mismatch between Id ({str}) and it's lowercased counterpart after finding escape positions. Original: {anotherStrLength}. Lowercased: {strLength}");
```
ayende, line 216: But that is something that can happen, no?

ayende: https://stackoverflow.com/questions/7491680/unicode-characters-having-asymmetric-upper-lower-case-why

PR #8554, opened 2019-01-24
RavenDB-12540 Slow tests Server.Replication.ReplicationBasicTestsSlow…

efratshenhar:
….DisableExternalReplication

### DIFF
--- a/test/SlowTests/Server/Replication/ReplicationBasicTestsSlow.cs
+++ b/test/SlowTests/Server/Replication/ReplicationBasicTestsSlow.cs
@@ -113,15 +113,25 @@ namespace SlowTests.Server.Replication
                 Assert.Equal("John Dow", replicated1.Name);
                 Assert.Equal(30, replicated1.Age);
 
+                var db1 = await GetDocumentDatabaseInstanceFor(store1);
+                var db2 = await GetDocumentDatabaseInstanceFor(store2);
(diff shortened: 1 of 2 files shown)


test/SlowTests/Server/Replication/ReplicationBasicTestsSlow.cs
```diff
@@ -119,7 +121,17 @@ public async Task DisableExternalReplication()
                     Disabled = true
                 };
                 var res = await store1.Maintenance.SendAsync(new UpdateExternalReplicationOperation(external));
+                using (Server.ServerStore.ContextPool.AllocateOperationContext(out JsonOperationContext context))
+                {
+                    var value = WaitForValue(() =>
```
ayende: Why do you have this change?

efratshenhar: Wait for the topology to change

ppekrol: @efratshenhar check this PR https://github.com/ravendb/ravendb/pull/7688
It looks like we did not applied that to 4.0

ayende: I understand what it does, but why are you waiting for topology change when updating external replication?

PR #8555, opened 2019-01-24
RavenDB-12754: Prevent resize of compression pager when not needed

famoser:
The compression pager is reduced in size even when it is not needed. 

As far as I understood, the lines of code I have changed should prevent a resize if the max used pages by the buffer are close to the number of already allocated pages to prevent expensive reallocations. 

They did not so far, as a `return true` resizes the buffer; and as the default return also returned `true`. I changed this behaviour to reflect the behaviour described in the already existing comment.

### DIFF
--- a/src/Voron/Impl/Journal/WriteAheadJournal.cs
+++ b/src/Voron/Impl/Journal/WriteAheadJournal.cs
@@ -1725,16 +1725,13 @@ namespace Voron.Impl.Journal
             if ((DateTime.UtcNow - _lastCompressionBufferReduceCheck).TotalMinutes < 5)
                 return false;
 
-            if (_maxNumberOfPagesRequiredForCompressionBuffer > _compressionPager.NumberOfAllocatedPages / 2)
-            {
-                // we recently used at least half of the compression buffer,
-                // no point in reducing size yet, we'll try to do it again next time
-                return false;
-            }
+            
+            // prevent resize if we recently used at least half of the compression buffer
+            var preventResize = _maxNumberOfPagesRequiredForCompressionBuffer > _compressionPager.NumberOfAllocatedPages / 2;
 
             _maxNumberOfPagesRequiredForCompressionBuffer = 0;
             _lastCompressionBufferReduceCheck = DateTime.UtcNow;
-            return true;
+            return !preventResize;
         }
 
         public void TryReduceSizeOfCompressionBufferIfNeeded()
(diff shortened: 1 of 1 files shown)


src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -1725,11 +1725,11 @@ private bool ShouldReduceSizeOfCompressionPager(long initialSize, bool forceRedu
             if ((DateTime.UtcNow - _lastCompressionBufferReduceCheck).TotalMinutes < 5)
                 return false;
 
-            if (_maxNumberOfPagesRequiredForCompressionBuffer < _compressionPager.NumberOfAllocatedPages / 2)
+            if (_maxNumberOfPagesRequiredForCompressionBuffer > _compressionPager.NumberOfAllocatedPages / 2)
             {
                 // we recently used at least half of the compression buffer,
                 // no point in reducing size yet, we'll try to do it again next time
-                return true;
+                return false;
```
ayende: The code here is actually correct. It is stating that we didn't use enough of the file.
The problem is actually different. It is that we are not checking for the maximum size that we should have here.

src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -1725,16 +1725,13 @@ private bool ShouldReduceSizeOfCompressionPager(long initialSize, bool forceRedu
...
 
-            if (_maxNumberOfPagesRequiredForCompressionBuffer > _compressionPager.NumberOfAllocatedPages / 2)
-            {
-                // we recently used at least half of the compression buffer,
-                // no point in reducing size yet, we'll try to do it again next time
-                return false;
-            }
+            
+            // prevent resize if we recently used at least half of the compression buffer
+            var preventResize = _maxNumberOfPagesRequiredForCompressionBuffer > _compressionPager.NumberOfAllocatedPages / 2;
```
ayende, line 1729: That would still cause "bouncing". If you transactions are varying in size, that can cause you to grow & shrink things rapidly.
The problem is that we only have `initialSize` here, and we need the `maxSize` as well

famoser: The bounce would (with the latest change) happen only all five minutes, but I get your point. Should I close the PR, or attempt a "real" fix sometimes this week?

famoser: I see you are way ahead of me; I see no more issues with your latest commit

ayende:
Next step now is to let the CI server take the code for a spin.

ayende (approved):

ppekrol:
Merged. Thanks @famoser for your contribution!

PR #8556, opened 2019-01-24
RavenDB-12725 Check sum issue

arekpalinski:
- Changing the way we're detecting what range of journal files is necessary to properly recover on database startup. The issue was that that we relied on JournalFilesCount which was in-memory state and under some conditions could be invalid. In result, on recovery startup, we were deleting a journal that had page that were not applied to data file yet.
- Applying the same logic when taking a full backup so we won't miss any journal file
- Added validation to the recovery which will prevent from loading a database if _first_ journal file is missing
- We no longer delete any journals after we apply their data to the data file - we must not do it because we don't sync the data file during the recovery. The first flush will schedule the data file sync which will remove already synced journals.
- We no longer modify the headers after the recovery process - as we don't delete any journals there now
- Changed the format FileHeader (JournalInfo.JournalFilesCount changed to JournalInfo.Reserved and JournalInfo.Flags) and bumping Voron version. Added relevant schema update.
- Enhanced exception messages

Original PR: https://github.com/ravendb/ravendb/pull/8538

### DIFF
--- a/src/Voron/Schema/Updates/From22.cs
+++ b/src/Voron/Schema/Updates/From22.cs
@@ -1,0 +1,26 @@
+﻿using Sparrow;
+using Voron.Impl.FileHeaders;
+using Voron.Impl.Journal;
+
+namespace Voron.Schema.Updates
+{
+    public class From22 : IVoronSchemaUpdate
+    {
+        public unsafe bool Update(int currentVersion, StorageEnvironmentOptions options, HeaderAccessor headerAccessor, out int versionAfterUpgrade)
+        {
+            headerAccessor.Modify(header =>
+            {
+                Memory.Set(header->Journal.Reserved, 0, JournalInfo.NumberOfReservedBytes);
+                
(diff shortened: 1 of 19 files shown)


src/Voron/Schema/Updates/From22.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Voron.Schema.Updates
+{
+    public class From22 : IVoronSchemaUpdate
+    {
+        public unsafe bool Update(int currentVersion, StorageEnvironmentOptions options, HeaderAccessor headerAccessor, out int versionAfterUpgrade)
+        {
+            headerAccessor.Modify(header =>
+            {
+                Memory.Set(header->Journal.Reserved, 0, JournalInfo.NumberOfReservedBytes);
```
ayende, line 13: Replace `NumberOfReservedBytes` with `3`

arekpalinski: Fixed in https://github.com/ravendb/ravendb/pull/8576

PR #8557, opened 2019-01-24
RavenDB-12744 A resolved document must preserve the original change v…

karmeli87:
…ector to avoid ping-pong replication

but it must update the global change vector.

### DIFF
--- a/src/Raven.Server/Documents/AttachmentsStorage.cs
+++ b/src/Raven.Server/Documents/AttachmentsStorage.cs
@@ -1119,7 +1119,7 @@ namespace Raven.Server.Documents
             return (doc, name);
         }
 
-#if DEBUG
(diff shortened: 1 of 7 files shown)


ppekrol (approved):

ppekrol:
@karmeli87 Can you check the failing test?

karmeli87:
Already fixed here https://github.com/ravendb/ravendb/pull/8558

ppekrol:
I've rebased the PR. Let's do another CI round.

ayende:
I'm not sure that I'm following, the merged change vector for the conflict is the merged vectors of the documents, which should end up the same on all nodes, no?

karmeli87:
Yes, the problem was that, that merged change vector was set as the global change vector. Instead we need to merge it into the global one.

PR #8569, opened 2019-01-27
RavenDB-12035 Setup wizard fails in macOS Mojave 10.14

ml054:


ml054:
I will open this for 4.1

PR #8571, opened 2019-01-27
RavenDB-12759 - fix NRE on shutdown

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-12759

### DIFF
--- a/src/Raven.Server/ServerWide/ServerStore.cs
+++ b/src/Raven.Server/ServerWide/ServerStore.cs
@@ -200,90 +200,91 @@ namespace Raven.Server.ServerWide
                 var delay = 500;
                 while (ServerShutdown.IsCancellationRequested == false)
                 {
-                    CancellationTokenSource cts = null;
                     Task leaderChangedTask = null;
-
-                    try
+                    using (var cts = CancellationTokenSource.CreateLinkedTokenSource(ServerShutdown))
                     {
-                        _engine.WaitForState(RachisState.Follower, ServerShutdown).Wait(ServerShutdown);
-                        if (ServerShutdown.IsCancellationRequested)
-                            return;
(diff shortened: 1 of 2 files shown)


src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -301,6 +301,9 @@ public void UpdateTopologyChangeNotification()
 
         private void WaitForLeaderChangeTaskToComplete(Task leaderChangedTask, CancellationTokenSource cts)
         {
+            if (cts == null)
```
ppekrol: I think we need to initialize the CTS a bit earlier to avoid getting any possibility of passing `null` here. Also code in the `UpdateTopologyChangeNotification` should use cts.Token all the way down probably. Now it is mixed between cts.Token and ServerShutdown.

ayende: Why do you get a token _source_ here?

ppekrol (approved):

src/Raven.Server/ServerWide/ServerStore.cs
grisha-kotler: changed

grisha-kotler: changed

ayende:
Did you remove the NRE fix?

ppekrol:
@ayende @grisha-kotler I've managed to recover those

ppekrol:
Something is still not right...

ppekrol:
Should be good now. @grisha-kotler can you re-check?

ayende (approved):

grisha-kotler:
LGTM

PR #8572, opened 2019-01-27
4.1 to 4.2 merge

ppekrol:


src/Raven.Studio/typescript/viewmodels/database/query/query.ts
```diff
@@ -653,27 +653,34 @@ class query extends viewModelBase {
                     .done((queryResults: pagedResultExtended<document>) => {
                         this.hasMoreUnboundedResults(false);
                         
+                        if (queryResults.items.length < take + 1) {
```
ppekrol, line 656: @ml054 Can you review this part? I had a conflict here and I;m not sure if I merged this correctly.

Issue #23, opened 2010-03-04
Investigate making indexing a C# 4.0 dynamic feature


ayende:
No reason to write my own dialect of the bloody thing if C# 4.0 gives it for free

Issue #10, opened 2010-03-02
Query by document id


ayende:
Should be able to POST a list of document ids to the server and get back their content.

PR #8574, opened 2019-01-27
RavenDB-12742 - Update of big value in same tx fails

ayende:


ppekrol:
Rebased.

src/Voron/Data/BTrees/Tree.cs
```diff
@@ -1268,6 +1268,8 @@ private bool TryOverwriteOverflowPages(TreeNodeHeader* updatedNode, int len, out
                         _llt.FreePage(readOnlyOverflowPage.PageNumber + requestedOverflows + i);
```
arekpalinski, line 1268: I'm pretty sure that we don't need `_llt.FreePage` now

src/Voron/Impl/LowLevelTransaction.cs
```diff
@@ -803,14 +800,7 @@ internal void FreePage(long pageNumber)
...
 
-                if (_dirtyPages.Remove(pageNumber) == false &&
-                    _dirtyOverflowPages.TryGetValue(pageNumber, out long numberOfOverflowPages))
-                {
-                    _dirtyOverflowPages.Remove(pageNumber);
-
-                    if (numberOfOverflowPages > 1) // prevent adding range which length is 0
-                        _dirtyOverflowPages.Add(pageNumber + 1, numberOfOverflowPages - 1); // change the range of the overflow page
-                }
+                _dirtyPages.Remove(pageNumber);
```
arekpalinski: Can we just call `DiscardScratchModificationOn` here?

src/Voron/Data/BTrees/Tree.cs
ayende: Yes, you do. Consider the case where you _shrink_ the value.
You wrote 64 KB, now you write 16 KB. We need a test for this, though

src/Voron/Impl/LowLevelTransaction.cs
ayende: Yes, I belive that would be best

src/Voron/Data/BTrees/Tree.cs
arekpalinski: I don't see a reason to call `FreePage` since we dropped `_dirtyOverflowPages` concept.

The initial 64KB write allocated page #10 (#23 in scratch). Now we have 16KB to write and we're in `TryOverwriteOverflowPages` optimization. We just need to discard #23, and allocate a new page in scratch for page #10 which will be smaller than initially.

Test:

```
[Theory]
        [InlineData(32 * 1024, 32 * 1024)]
        [InlineData(32 * 1024, 16 * 1024)]
        public void Can_update_big_value_in_the_same_transaction(int initialSize, int updateSize)
        {
            var r = new Random();

            using (var tx = Env.WriteTransaction())
            {
                Tree tree = tx.CreateTree("tree");

                var initial = new byte[initialSize];

                r.NextBytes(initial);
                tree.Add("key", initial);

                var updated = new byte[initialSize];

                r.NextBytes(updated);
                tree.Add("key", updated);

                tx.Commit();
            }
        }
```

ppekrol:
Rebased

src/Voron/Data/BTrees/Tree.cs
arekpalinski: @ayende You added some tests to verify the behavior when updated value is shrink. Do you still think that `_llt.FreePage` is necessary here?

ayende: Yes, inside `FreePage`, we are calling 

```
_freeSpaceHandling.FreePage(this, pageNumber);
_freedPages.Add(pageNumber);
```

Otherwise, we'll lose the space

arekpalinski: Right.

arekpalinski (approved):

arekpalinski:
Any reason to not apply it to 4.1?

ayende:
I think we should, yes

arekpalinski:
Created PR against 4.1: https://github.com/ravendb/ravendb/pull/8600

PR #8580, opened 2019-01-28
RavenDB-12724 Exposing the reason a test failed due to licensing failure

talweiss1982:
I couldn't reproduce the failure, this could happen due to a few reasons from cluster timing issues to our licensing server not responding when the license expires.
What I did is expose the reason the license failed to load explicitly so the test will fail with the real reason.
I could easily ignore the test failure when this happens but I think it would be better to live this so we could know why we fail to activate the license.

### DIFF
--- a/src/Raven.Server/Commercial/LicenseManager.cs
+++ b/src/Raven.Server/Commercial/LicenseManager.cs
@@ -540,14 +540,14 @@ namespace Raven.Server.Commercial
             });
         }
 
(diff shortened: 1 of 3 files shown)


talweiss1982:
After talking with @ppekrol we decided to move the validation into the ensure not passive and throw on servers that are generated by the TestBase

talweiss1982:
see https://github.com/ravendb/ravendb/pull/8582

PR #8581, opened 2019-01-28
RavenDB-12771 fixing resolved document and revisions behaviour

karmeli87:
- modifing a resovled document should keep the HasRevisions flags
- A revision that has the 'Conflicted' flag shouldn't overwrite the document
- fix various database change vector merging
- adding counter to a resovled document should not generate a revision

### DIFF
--- a/src/Raven.Server/Documents/CountersStorage.cs
+++ b/src/Raven.Server/Documents/CountersStorage.cs
@@ -693,7 +693,7 @@ namespace Raven.Server.Documents
             if (hadModifications == false)
                 return;
 
-            var flags = doc.Flags.Strip(DocumentFlags.FromClusterTransaction);
+            var flags = doc.Flags.Strip(DocumentFlags.FromClusterTransaction | DocumentFlags.Resolved);
             if (counters.Count == 0)
             {
                 flags = flags.Strip(DocumentFlags.HasCounters);
--- a/src/Raven.Server/Documents/DocumentPutAction.cs
+++ b/src/Raven.Server/Documents/DocumentPutAction.cs
@@ -169,6 +169,8 @@ namespace Raven.Server.Documents
                     }
                 }
 
+                UpdateChangeVectorIfNeeded(context, nonPersistentFlags, ref changeVector, ref newEtag);
+
                 using (Slice.From(context.Allocator, changeVector, out var cv))
                 using (table.Allocate(out TableValueBuilder tvb))
                 {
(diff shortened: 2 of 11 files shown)


src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -250,9 +250,11 @@ public RevisionsCollectionConfiguration GetRevisionsConfiguration(string collect
             return Configuration.Default ?? _emptyConfiguration;
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
ayende: This is a big method and not likely to be inlined even with this.
Why is this here?

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -312,6 +321,7 @@ public RevisionsCollectionConfiguration GetRevisionsConfiguration(string collect
             return true;
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
ayende: remove this

ayende (commented):
Talk to me about this tomorrow, I want to go over this with you so I can be sure that I understand what you have done here

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
grisha-kotler: done

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
grisha-kotler: done

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -693,7 +693,7 @@ public long GetNumberOfCounterEntries(DocumentsOperationContext context)
             if (hadModifications == false)
                 return;
 
-            var flags = doc.Flags.Strip(DocumentFlags.FromClusterTransaction);
+            var flags = doc.Flags.Strip(DocumentFlags.FromClusterTransaction | DocumentFlags.Resolved);
```
ayende, line 696: Check what happens with attachments in the same scenario

src/Raven.Server/Documents/DocumentPutAction.cs
```diff
@@ -229,6 +222,20 @@ oldFlags .Contain(DocumentFlags.HasAttachments))
             }
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
ayende: Remove this

src/Raven.Server/Documents/CountersStorage.cs
karmeli87: done

src/Raven.Server/Documents/DocumentPutAction.cs
karmeli87: done

PR #8591, opened 2019-01-29
RavenDB-12778 fix

myarichuk:
Make sure to separate "color in console" and "use named pipes" flags in RavenCli. This makes the cli not to use "named pipes" mode when running in interactive mode, regardless if the "parent process id" is passed or not (RavenDB-12778)


ayende (approved):

ppekrol:
Closing because we will do a 4.2 -> 4.1 merge

PR #8598, opened 2019-01-29
RavenDB-12780 NRE when accessing Id of a loaded document in an index

ayende:


src/Raven.Server/Documents/Indexes/Static/StaticIndexBase.cs
```diff
@@ -57,6 +57,8 @@ public void AddReferencedCollection(string collection, string referencedCollecti
 
         public dynamic Id(dynamic doc)
         {
+            if (doc == null)
```
grisha-kotler, line 60: doc is DynamicNullObject.Null (the return value of LoadDocument when the document is missing)
https://github.com/ravendb/ravendb/pull/8599

PR #8599, opened 2019-01-29
RavenDB-12780 - NRE when accessing Id of a loaded document in an index

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-12780

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Static/StaticIndexBase.cs
+++ b/src/Raven.Server/Documents/Indexes/Static/StaticIndexBase.cs
@@ -282,7 +282,7 @@ namespace Raven.Server.Documents.Indexes.Static
         {
             throw new InvalidOperationException(
                 $"{funcName} may only be called with a document, " +
-                $"but was called with a parameter of type {doc.GetType().FullName}: {doc}");
+                $"but was called with a parameter of type {doc?.GetType().FullName}: {doc}");
         }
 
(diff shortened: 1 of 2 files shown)


src/Raven.Server/Documents/Indexes/Static/StaticIndexBase.cs
```diff
@@ -58,7 +58,7 @@ public void AddReferencedCollection(string collection, string referencedCollecti
         public dynamic Id(dynamic doc)
         {
             var json = (DynamicBlittableJson)doc;
```
ayende: If `doc` is `DynamicNullObject`, your cast will fail, need to do the check higher than this.

ayende (changes requested):

ppekrol:
According to `LoadDocument` code it would be better to check if value is `DynamicNullObject`. Check here: https://github.com/ravendb/ravendb/blob/v4.1/src/Raven.Server/Documents/Indexes/Static/StaticIndexBase.cs#L86

src/Raven.Server/Documents/Indexes/Static/StaticIndexBase.cs
grisha-kotler: Done

ppekrol (approved):

ayende (approved):

PR #8609, opened 2019-01-30
RavenDB-12725

arekpalinski:
- added an option allowing to ignore missing journals on startup 
- set JournalApplicator._lastFlushed after the recovery and suggest syncing an environment


src/Raven.Server/Program.cs
```diff
@@ -178,6 +178,24 @@ public static int Main(string[] args)
...
 
+                            if (configuration.Storage.IgnoreInvalidJournalErrors == true)
+                            {
+                                var message =
+                                    $"Server is running in dangerous mode because {RavenConfiguration.GetKey(x => x.Storage.IgnoreInvalidJournalErrors)} was set. " +
+                                    "It means that storages of databases, indexes and system one will be loaded regardless missing or corrupted journal files which " +
+                                    "are mandatory to properly load the storage. " +
+                                    "This switch is meant to be use only for recovery purposes. Please make sure that you won't use it on regular basis. ";
+
+                                if (Logger.IsInfoEnabled)
```
ayende: Should probably be operations

src/Raven.Server/Storage/Layout/StorageLoader.cs
```diff
@@ -0,0 +1,138 @@
...
+                                
+                                break;
+                            case StorageEnvironmentWithType.StorageEnvironmentType.Configuration:
+                                message += $"You can delete the configuration storage folder at '{basePath.FullPath}' and restart the server. It will be recreated on database startup.";
+                                break;
+                            case StorageEnvironmentWithType.StorageEnvironmentType.System:
+                                message += "You can start the server in dangerous mode temporary so it will ignore invalid journals on startup:" +
+                                           $"{ravenServer} --{RavenConfiguration.GetKey(x => x.Storage.IgnoreInvalidJournalErrors)}=true{Environment.NewLine}" +
+                                           $"Eventually you should delete the storage configuration at '{basePath.FullPath}' and create your databases again with the usage of existing data.";
+                                break;
```
arekpalinski, line 87: Depending on the storage env type I wanted to suggest different actions to recover from this state. Please let me know if you see the potential improvement of those messages.

src/Raven.Server/Storage/Layout/StorageLoader.cs
```diff
@@ -0,0 +1,138 @@
...
+                var preRtmJournalPath = basePath.Combine("Journal");
+                var journalsPath = options.JournalPath;
+                if (Directory.Exists(preRtmJournalPath.FullPath))
+                {
+                    TryMoveJournals(preRtmJournalPath, journalsPath);
+                    Directory.Delete(preRtmJournalPath.FullPath);
+                }
+                else
+                {
+                    if (TryMoveJournals(basePath, journalsPath) == false)
```
ayende, line 59: When would this ever be triggered because of the missing journal?
You may be missing journal 16, but 17 is there, so it wouldn't call it, no?

src/Raven.Server/Program.cs
arekpalinski: Fixed in https://github.com/ravendb/ravendb/pull/8609/commits/f82c40d830e0afbdf48da42e2457ecdf2f0964ee

src/Raven.Server/Storage/Layout/StorageLoader.cs
arekpalinski: The existing code, in particular, `TryMoveJournals` was created to handle the change when we moved journals files to `Journals` directory in the past.

If journals are already in `Journals` directory then we're sure that missing 16 will cause throwing exception here.

ayende (approved):

PR #8611, opened 2019-01-30
RavenDB-12421 unbounded compaction in 32-bit mode could sometime cras…

talweiss1982:
…h the storage

### DIFF
--- a/src/Raven.Server/Documents/TransactionOperationsMerger.cs
+++ b/src/Raven.Server/Documents/TransactionOperationsMerger.cs
@@ -16,6 +16,7 @@ using Sparrow;
 using Sparrow.Json;
 using Sparrow.Logging;
 using Sparrow.LowMemory;
+using Sparrow.Platform;
 using Sparrow.Utils;
 using Voron.Debugging;
 using Voron.Global;
@@ -826,7 +827,7 @@ namespace Raven.Server.Documents
                 var modifiedSize = llt.NumberOfModifiedPages * Constants.Storage.PageSize;
 
                 var canCloseCurrentTx = previousOperation == null || previousOperation.IsCompleted;
-                if (canCloseCurrentTx)
+                if (canCloseCurrentTx || llt.Environment.Options.ForceUsing32BitsPager || PlatformDetails.Is32Bits)
                 {
                     if (_operations.IsEmpty)
                         break; // nothing remaining to do, let's us close this work
--- a/src/Raven.Server/Documents/TransactionOperationsMerger.cs
+++ b/src/Raven.Server/Documents/TransactionOperationsMerger.cs
@@ -48,6 +48,8 @@ namespace Raven.Server.Documents
         private readonly long _maxTxSizeInBytes;
         private readonly double _maxTimeToWaitForPreviousTxBeforeRejectingInMs;
 
+        private readonly bool _is32Bits;
+
         public TransactionOperationsMerger(DocumentDatabase parent, CancellationToken shutdown)
         {
             _parent = parent;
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/TransactionOperationsMerger.cs
```diff
@@ -826,7 +827,7 @@ private enum PendingOperations
                 var modifiedSize = llt.NumberOfModifiedPages * Constants.Storage.PageSize;
 
                 var canCloseCurrentTx = previousOperation == null || previousOperation.IsCompleted;
-                if (canCloseCurrentTx)
+                if (canCloseCurrentTx || llt.Environment.Options.ForceUsing32BitsPager || PlatformDetails.Is32Bits)
```
ayende: Can we extract this computation to the constructor? 
I want to make it easy to ignore

ayende: Also, what about the behavior a bit below, where we do the check again in:

```
// if I can't close the tx, this means that the previous async operation is still in progress
// and there are incoming requests coming in. We'll accept them, to a certain limit
if (modifiedSize < _maxTxSizeInBytes)
    continue; // we can still process requests at this time, so let's do that...
```

src/Voron/Impl/Paging/Windows32BitsMemoryMapPager.cs
```diff
@@ -338,9 +339,18 @@ private LoadedPage MapPages(TransactionState state, long startPage, long size)
...
+
+                    const int ERROR_NOT_ENOUGH_MEMORY = 8;
+                    if (lastWin32Error == ERROR_NOT_ENOUGH_MEMORY)
+                    {
+                        throw new OutOfMemoryException($"Unable to map {size / Constants.Size.Kilobyte:#,#0} kb starting at {startPage} on {FileName}", new Win32Exception(lastWin32Error));
+                    }
+                    else
+                    {
+                        throw new Win32Exception(
+                            $"Unable to map {size / Constants.Size.Kilobyte:#,#0} kb starting at {startPage} on {FileName}",
```
ayende: Include the `lastWin32Error ` here too

ayende (changes requested):

src/Raven.Server/Documents/TransactionOperationsMerger.cs
talweiss1982: I'm not sure what should be the behavior in the merger, I don't want to start rejecting requests maybe it is better if @aviviadi  or @haludi will look into that?
_maxTxSizeInBytes controls the max transaction size it is set to 4MB for 32 bit
I'm not sure what the behavior is if we break from this loop...

talweiss1982: actually the merger is limiting this to 4MB for 32bit so we can remove this entirly

src/Raven.Server/Documents/TransactionOperationsMerger.cs
```diff
@@ -67,6 +68,7 @@ public TransactionOperationsMerger(DocumentDatabase parent, CancellationToken sh
 
         public void Start()
         {
+            Is32Bits = _parent.DocumentsStorage.Environment.Options.ForceUsing32BitsPager || PlatformDetails.Is32Bits;
```
ppekrol: We need to move this to ctor and mark it as readonly

ppekrol: You can do that by accessing database.Configuration.Storage.ForceUsing32BitsPager

ppekrol (approved):

PR #8620, opened 2019-01-31
RavenDB-12778 fix

myarichuk:
Make sure to separate "color in console" and "use named pipes" flags in RavenCli. This makes the cli not to use "named pipes" mode when running in interactive mode, regardless if the "parent process id" is passed or not.
Also make sure that commands that are not supported in "named pipes" mode throw exceptions PROPERLY.


ayende (approved):

ppekrol (approved):

ppekrol:
Merged to 4.1: https://github.com/ravendb/ravendb/pull/8622

PR #8623, opened 2019-01-31
RavenDB-12535: Asynchronous Global Prefetcher

redknightlois:
PrefetchVirtualMemory is a quite costly kernel call. We should just enqueue this on a dedicated worker thread and continue with our work, expecting the OS to do the right thing and when we need that memory we already have the memory in place.


src/Sparrow/Collections/RingBuffer.cs
```diff
@@ -0,0 +1,136 @@
...
+            {
+                // We will check if we potentially have the ability to push 1 more item. 
+                int cidx = Volatile.Read(ref _currentIdx);
+                int sidx = Volatile.Read(ref _startIdx);
+
+                // Check if we have empty spaces in the buffer.
+                if (cidx - sidx >= this._size)
+                    return false; // No space to do anything
+
+                Interlocked.CompareExchange(ref _currentIdx, _currentIdx + 1, _currentIdx);
```
ayende: Shouldn't you be checking if this succeeded?

src/Sparrow/Collections/RingBuffer.cs
```diff
@@ -0,0 +1,136 @@
...
+                
+                // We check if we were successful 
+                if (Volatile.Read(ref _currentIdx) > cidx)
```
ayende: Isn't there a race here, why no check the return value from `CompareExchange` ?

src/Voron/GlobalPrefetchingBehavior.cs
```diff
@@ -0,0 +1,105 @@
...
+            const int StackSpace = 32;
+
+            var toPrefetch = stackalloc MemorySegment[StackSpace];
+            try
+            {               
+                while (true)
+                {
+                    var commands = CommandQueue.Acquire();
+                    if (commands.IsEmpty)
+                        Thread.Sleep(1000); // If no command is in the queue, we are surely not busy.
```
ayende: That is a bad idea. If we are sleeping and suddenly want to prefetch, that is going to _not_ prefetch and we have piecemeal I/O
Should use an event, instead.

src/Voron/GlobalPrefetchingBehavior.cs
```diff
@@ -0,0 +1,105 @@
...
+                IsBackground = true,
+                Name = PrefetchingThreadName
+            };
+            thread.Start();
+            return prefetcher;
+        });
+
+        private readonly Logger _log = LoggingSource.Instance.GetLogger<GlobalPrefetchingBehavior>("Global Prefetcher");
+         
+        public readonly SingleConsumerRingBuffer<PrefetchCommand> CommandQueue = new SingleConsumerRingBuffer<PrefetchCommand>(128);
```
ayende: Why do we want something like this?
It would be much easier to use a `ConcurrentQueue`, instead.
This is going to be triggering I/O, so I don't think we need to worry that much about efficiencies.

src/Sparrow/Platform/Win32/Win32MemoryMethods.cs
```diff
@@ -1,9 +1,10 @@
 ﻿using System;
 using System.Runtime.InteropServices;
+using System.Security;
 
 namespace Sparrow.Platform.Win32
 {
-    public static unsafe class Win32MemoryProtectMethods
+    public static unsafe class Win32MemoryMethods
```
ayende: why this change?

ayende (changes requested):
This is very going to conflict with the librvnpal changes, pay attention to that.

src/Sparrow/Collections/RingBuffer.cs
redknightlois: This is a reservation, I have code here that is using `Interlocked.Increment` instead. Needed to run slow tests, that is why I did this PR...

src/Sparrow/Collections/RingBuffer.cs
redknightlois: Not really. Because it works as a reservation system.

src/Voron/GlobalPrefetchingBehavior.cs
redknightlois: Nope. Because the caller is high frequency and we want to avoid the actual cost on the caller side. We dont care if we miss a few, in fact it goes way beyond that... we dont care if we miss a ton of them because we dont have enough space on the reservation queue.

src/Voron/GlobalPrefetchingBehavior.cs
redknightlois: Yes we do, because the one that is going to pay for it is the caller... The whole point of the optimization is to move the slow parts outside of the critical path, even at the expense of losing a few hits here and there.

src/Voron/GlobalPrefetchingBehavior.cs
ayende: The caller is already protected by `ShouldPrefetch`, so I don't think we are going to see too many of that

src/Sparrow/Platform/Win32/Win32MemoryMethods.cs
redknightlois: Because it was a mess, all methods there are Memory methods, not protect methods. They happen to be useful for that, but we have many things scattered all over the place.

ayende: Break it to a separate PR, it make it really hard to review the changes for this PR, which are already complex enough to require careful review

redknightlois: Its just a rename, nothing more. Had to move stuff around to access that. So you will still see the changes in places that used that anyway. To me is a simple refactor rename.

src/Voron/GlobalPrefetchingBehavior.cs
redknightlois: You will. Don't trust me, trust the profiling trace :)

redknightlois: After some changes, the profiling trace suggests we can do wake up based on quantity. I tried though to get down the number of milliseconds down, with barely any change in hard page faults though.

ayende: Can you show the numbers?

redknightlois: 180 hard faults per second either way.

redknightlois: For reference we were in the 100ish range. But I didn't have time today to check the performance targets as I don't know the behavior yet of 4.2

redknightlois: Implemented using a BlockingCollection instead. The wait was too much on average.

src/Voron/GlobalPrefetchingBehavior.cs
redknightlois: Using `BlockingCollection`

src/Sparrow/Platform/Win32/Win32ProcessMethods.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace Sparrow.Platform.Win32
```
ppekrol: Needs to be moved to Sparrow.Server

src/Sparrow/Collections/RingBuffer.cs
```diff
@@ -0,0 +1,132 @@
+﻿using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using System.Threading;
+
+namespace Sparrow.Collections
```
ppekrol, line 7: Better to move it to Sparrow.Server if this is not used in Raven.Client

redknightlois: Botched merge, that is not used. removing.

src/Sparrow/Platform/Win32/Win32ProcessMethods.cs
redknightlois: Botched merge, that is not used. Removing

src/Sparrow/Collections/RingBuffer.cs
redknightlois: That is going to be used by RavenDB-12799 (https://issues.hibernatingrhinos.com/issue/RavenDB-12799) on `Sparrow`

src/Sparrow.Server/Platform/Win32/Win32MemoryProtectMethods.cs
```diff
@@ -24,8 +25,12 @@ public static unsafe class Win32MemoryProtectMethods
...
         );
 
-        [DllImport("kernel32.dll", SetLastError = true)]
-        public static extern bool DiscardVirtualMemory(void* lpAddress, UIntPtr size);
+        //[DllImport("kernel32.dll", SetLastError = true, CallingConvention = CallingConvention.Winapi)]
+        //[SecurityCritical]
+        //public static extern bool PrefetchVirtualMemory(IntPtr hProcess, UIntPtr NumberOfEntries, MemorySegment* VirtualAddresses, ulong Flags);
+
+        //[DllImport("kernel32.dll", SetLastError = true)]
+        //public static extern bool DiscardVirtualMemory(void* lpAddress, UIntPtr size);
```
ayende: Better to delete, rather than comment the code

src/Sparrow/Collections/RingBuffer.cs
ayende: What is this going to be used _in sparrow_ ?

src/Sparrow/Collections/RingBuffer.cs
```diff
@@ -0,0 +1,157 @@
...
+            int sidx = Volatile.Read(ref _startIdx);
+
+            // Check if we have empty spaces in the buffer.
+            if (cidx - sidx >= this._size)
+                return false; // No space to do anything
+
+            int ticker = Interlocked.Increment(ref _currentIdx) - 1;
+
+            // Assign the value
+            ref var cl = ref _buffer[ticker % this._size];
```
ayende: The compiler can't assume power of 2 here, no?
Can we turn this to a shift?

src/Sparrow/Collections/RingBuffer.cs
```diff
@@ -0,0 +1,157 @@
...
+
+            // We assign the item
+            cl.Item = item;
+
+            // We then write the volatile value.
+            Volatile.Write(ref cl.IsReady, true);
+            return true;
+
+            // Very rarely we will go through and try again, the only condition is when we have a huge contention and traffic.
```
ayende, line 72: what make us go through here again?

src/Sparrow/Collections/RingBuffer.cs
ayende: Okay, this is used by the logging infrastructure, which we want to keep in sparrow

src/Voron/Impl/Paging/AbstractPager.cs
```diff
@@ -529,45 +542,27 @@ public void MaybePrefetchMemory(long pageNumber, int pagesToPrefetch)
...
-            // the CLR has to zero the stack space, and WIN32_MEMORY_RANGE_ENTRY is 16 bytes on 64 bits
-            // therefor, we want to avoid having too large a value here. We are already zeroing 256 bytes
-            // and with the default behavior can get 16MB prefetched per sys call, so that is quite enough.
-            // We expect that we'll only rarely need to do more than 16 at a go, anyway
-            const int StackSpace = 16;
+            var prefetcher = GlobalPrefetchingBehavior.GlobalPrefetcher.Value;
 
-            int prefetchIdx = 0;
-            var toPrefetch = stackalloc PalDefinitions.PrefetchRanges[StackSpace];
+            var command = default(PalDefinitions.PrefetchRanges);
```
ayende:
```suggestion
            PalDefinitions.PrefetchRanges command = default;
```

ayende: This looks clearer (minor)

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
```diff
@@ -150,11 +150,11 @@ protected override void DisposeInternal()
             }
         }
 
-        protected internal override void PrefetchRanges(PrefetchRanges* list, int count)
-        {
-            rvn_prefetch_ranges(list, count, out _);
-            // we explicitly ignore the return code here, this is optimization only
-        }
+        //protected internal override void PrefetchRanges(PrefetchRanges* list, int count)
```
ayende: Delete commented code

src/Voron/StorageEnvironment.cs
```diff
@@ -158,6 +158,9 @@ public StorageEnvironment(StorageEnvironmentOptions options)
 
                 if (isNew == false && _options.ManualSyncing == false)
                     SuggestSyncDataFile(); // let's suggest syncing data file after the recovery
+
+                // Ensure we are always have the prefetcher available.
+                var prefetcher = GlobalPrefetchingBehavior.GlobalPrefetcher.Value;
```
ayende:
```suggestion
                GC.KeepAlive(GlobalPrefetchingBehavior.GlobalPrefetcher.Value);
```

We use this commonly, and it is guaranteed that the compiler won't elide this

src/Sparrow.Server/Platform/Win32/Win32MemoryProtectMethods.cs
redknightlois: Yeah. Was merging and wanted to be sure it was not needed and forgot to delete afterwards

src/Sparrow/Collections/RingBuffer.cs
redknightlois: We can if we control the size. Easy change if we want to do so.

src/Voron/StorageEnvironment.cs
redknightlois: Ok, for readability purposes it makes a lot of sense. But, there is no way thr compiler could elide it, because of the static side effects. If that would happen, there are a few invariants violated (major regression).

src/Sparrow/Collections/RingBuffer.cs
redknightlois: The ring buffer works in a acquire, release mode. You first acquire, do what you need to do and then release. On very small buffers with very high traffic you could end up in a situation where you have hit the bounded size. You end up there when that happens.

redknightlois:
Done. Added power of 2 assumptions and therefore better codegen.

test/FastTests/Sparrow/RingBuffer.cs
```diff
@@ -0,0 +1,147 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Sparrow.Collections;
+using Xunit;
+
+namespace FastTests.Sparrow
+{
+    public class RingBufferTests
```
arekpalinski: Should inherit from `NoDisposalNeeded`

ppekrol:
Rebased, did a cleanup and added an option to disable prefetching

ppekrol:
Why it needs to be public? it is used only by us internally.

redknightlois:
That is in sparrow, I am working on POCs on voron that use that structure

ppekrol:
Internals are visible to Voron and Sparrow.Server

redknightlois:
Didn't know that we have setup a friend. Great

PR #8629, opened 2019-02-01
RavenDB-12761 Fixing race condition in test - make sure that Index.Re…

arekpalinski:
…setErrors isn't called meanwhile when indexing thread starts

### DIFF
--- a/test/FastTests/Server/Documents/Indexing/Auto/BasicAutoMapIndexing.cs
+++ b/test/FastTests/Server/Documents/Indexing/Auto/BasicAutoMapIndexing.cs
@@ -420,8 +420,14 @@ namespace FastTests.Server.Documents.Indexing.Auto
                         new[] { new AutoIndexField { Name = "Name", Storage = FieldStorage.No } }),
                     database))
                 {
+                    var mre = new ManualResetEvent(false);
+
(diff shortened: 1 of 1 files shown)


test/FastTests/Server/Documents/Indexing/Auto/BasicAutoMapIndexing.cs
```diff
@@ -423,6 +423,12 @@ public void WriteErrors()
                     index.Start();
                     Assert.Equal(IndexRunningStatus.Running, index.Status);
 
+                    var mre = new ManualResetEvent(false);
+
+                    database.IndexStore.IndexBatchCompleted = x => { mre.Set(); };
```
ppekrol, line 425: Shouldn't we attach to this before doing `index.Start()` ?

arekpalinski: Yes, we should. Thx

arekpalinski: Fixed. Force pushed

ppekrol (approved):

PR #8632, opened 2019-02-03
RavenDB-12762 use current data dir when continuing cluster setup

iftahbe:


src/Raven.Server/Commercial/SetupManager.cs
```diff
@@ -1336,24 +1336,55 @@ public static IEnumerable<string> GetCertificateAlternativeNames(X509Certificate
...
-                settingsJsonObject.TryGet(dataDirProp, out string dataDirFromZip);
+                // During setup we use the System database to store cluster configurations as well as the trusted certificates.
+                // We need to make sure that the currently used data dir will be the one written (or not written) in the resulting settings.json
+                var dataDirKey = RavenConfiguration.GetKey(x => x.Core.DataDirectory);
+                var currentDataDir = serverStore.Configuration.GetServerWideSetting(dataDirKey) ?? serverStore.Configuration.GetSetting(dataDirKey);
 
-                var dataDirFromZipPath = new PathSetting(dataDirFromZip);
-                
-                if (currentDataDir.Equals(dataDirFromZipPath) == false)
+                if (settingsJsonObject.TryGet(dataDirKey, out string dataDirFromZip))
```
ppekrol: check also here if `dataDirFromZip` is not null or empty

src/Raven.Server/Commercial/SetupManager.cs
```diff
@@ -1336,24 +1336,55 @@ public static IEnumerable<string> GetCertificateAlternativeNames(X509Certificate
...
+                        settingsJsonObject = context.ReadObject(settingsJsonObject, "removingDataDir");
+                    }
+                    else
+                    {
+                        var currentDataDirPath = new PathSetting(currentDataDir);
+                        if (currentDataDirPath.Equals(dataDirFromZipPath) == false)
+                        {
+                            settingsJsonObject.Modifications = new DynamicJsonValue(settingsJsonObject)
+                            {
+                                [dataDirKey] = currentDataDirPath.FullPath
```
ppekrol: I think we want to put here the original currentDataDir value, not the evaluated one?

src/Raven.Server/Commercial/SetupManager.cs
```diff
@@ -1336,24 +1336,55 @@ public static IEnumerable<string> GetCertificateAlternativeNames(X509Certificate
...
+                    {
+                        // nothing to do, neither has a data dir defined
+                    }
+                    else
+                    {
+                        var currentDataDirPath = new PathSetting(currentDataDir);
+                        
+                        settingsJsonObject.Modifications = new DynamicJsonValue(settingsJsonObject)
+                        {
+                            [dataDirKey] = currentDataDirPath.FullPath
```
ppekrol: I think we want to put here the original currentDataDir value, not the evaluated one?

src/Raven.Server/Commercial/SetupManager.cs
iftahbe: yes

ppekrol (approved):

PR #8633, opened 2019-02-03
RavenDB-12623: Removing debug code

maximburyak:


ppekrol (approved):

redknightlois:
Unless there is a bug there why are we removing the debug safeguards?

maximburyak:
@redknightlois that's code that was temporary in the first place. Is there a _part_ of it you think we should keep?

redknightlois:
You are essentially checking an invariant there. If we don't want to introduce that error again it would make sense to run that at least on validation. I would probably switched that to validation if it is too costly to run on debug.

maximburyak:
@ayende  thoughts?

ayende:
This was temporary code that I didn't like having in the first place. 
I get the point of having the validation, but we need to draw the line somewhere otherwise even `VALIDATE` builds will become unusable.

PR #8635, opened 2019-02-03
Raven.client added ToArrayAsync()

valeriob:
The most common use of materialized queries is to use an immutable data structure, like an array instead of list.
This pull requests add support to async version of the equivalent async methos query.ToArrayAsync()
It also leverage the queryResult.TotalResults to allocate the array upfront instead of letting the list grow little by little.

### DIFF
--- a/src/Raven.Client/Documents/LinqExtensions.cs
+++ b/src/Raven.Client/Documents/LinqExtensions.cs
@@ -314,6 +314,19 @@ namespace Raven.Client.Documents
         }
 
         /// <summary>
+        /// Returns an array of results for a query asynchronously. 
+        /// </summary>
+        public static Task<T[]> ToArrayAsync<T>(this IQueryable<T> source, CancellationToken token = default)
+        {
+            var provider = source.Provider as IRavenQueryProvider;
+            if (provider == null)
+                throw new ArgumentException("You can only use Raven Queryable with ToArrayAsync");
+
+            var documentQuery = provider.ToAsyncDocumentQuery<T>(source.Expression);
+            return documentQuery.ToArrayAsync(token);
+        }
+
+        /// <summary>
         /// Determines whether a sequence contains any elements.
         /// </summary>
         /// 
--- a/src/Raven.Client/Documents/Session/AsyncDocumentQuery.cs
+++ b/src/Raven.Client/Documents/Session/AsyncDocumentQuery.cs
@@ -771,7 +771,14 @@ namespace Raven.Client.Documents.Session
         }
 
         /// <inheritdoc />
-        Task<List<T>> IAsyncDocumentQueryBase<T>.ToListAsync(CancellationToken token)
+        async Task<List<T>> IAsyncDocumentQueryBase<T>.ToListAsync(CancellationToken token)
+        {
+            var items = await ExecuteQueryOperation(null, token).ConfigureAwait(false);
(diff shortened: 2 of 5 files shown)


src/Raven.Client/Documents/LinqExtensions.cs
```diff
@@ -313,6 +313,19 @@ public static Task<List<T>> ToListAsync<T>(this IQueryable<T> source, Cancellati
...
+        /// Returns an array of results for a query asynchronously. 
+        /// </summary>
+        public static Task<T[]> ToArrayAsync<T>(this IQueryable<T> source, CancellationToken token = default)
+        {
+            var provider = source.Provider as IRavenQueryProvider;
+            if (provider == null)
+                throw new ArgumentException("You can only use Raven Queryable with ToArrayAsync");
+
+            var documentQuery = provider.ToAsyncDocumentQuery<T>(source.Expression);
+            return documentQuery.ToArrayAsync(token);
```
ppekrol, line 326: @valeriob It is actually a big change, because now all queries that were doing `ToList()` will be allocating an array and after that allocating a second list.

I think it would be better to just invoke `ToListAsync` extension method here and issue `ToArray` on the results here as a shortcut.

src/Raven.Client/Documents/Session/AsyncDocumentQuery.cs
```diff
@@ -771,7 +771,14 @@ Lazy<Task<int>> IAsyncDocumentQueryBase<T>.CountLazilyAsync(CancellationToken to
         }
 
         /// <inheritdoc />
-        Task<List<T>> IAsyncDocumentQueryBase<T>.ToListAsync(CancellationToken token)
+        async Task<List<T>> IAsyncDocumentQueryBase<T>.ToListAsync(CancellationToken token)
+        {
+            var items = await ExecuteQueryOperation(null, token).ConfigureAwait(false);
```
ppekrol, line 776: Just invoke previous implementation and then do `ToArray` on the results as a shortcut.

ppekrol (changes requested):

src/Raven.Client/Documents/LinqExtensions.cs
valeriob: Yes and no: everybody will allocate an array , and only who calls ToList() will have the overhead of allocating a new list from it. 
I think that query should as default return arrays and not lists (like Load() does) and ppl should opt in into List or other structures. I've never used ToList() in any db (ORM, uORM or ravendb 2/3/4) query. I think that ravendb4 i so relative new that this change is still possible, maybe we can add some comment on the ToList version to warn ppl of that allocation, even though it should be expected 😄.
Consider also that the gains obtained from initializing the array upfront should be more than the performances spent doing a single copy in case of ToList(), i think it's a net gain afterall (i did not measure, just counting copies of memory, consider that EnsureCapacity doubles the array every time, so can be pretty inefficient).

https://referencesource.microsoft.com/#mscorlib/system/collections/generic/list.cs,eb66b6616c6fd4ef,references
https://referencesource.microsoft.com/#mscorlib/system/collections/generic/list.cs,9808f1f5ef16c436,references

ppekrol:
@valeriob I've modified a bit your solution. Check it here: https://github.com/ravendb/ravendb/pull/8640

PR #8637, opened 2019-02-03
RavenDB-12822 propagating the wait for non stale result in graph queries

talweiss1982:


src/Raven.Client/Documents/Session/DocumentQuery.cs
```diff
@@ -908,6 +908,16 @@ private IGraphQuery<T> WithInternal<TOther>(string alias, DocumentQuery<TOther>
             }
 
             WithTokens.AddLast(new WithToken(alias, docQuery.ToString()));
+
+            if (docQuery.TheWaitForNonStaleResults)
```
ppekrol, line 912: We need to add it to AsyncDocumentQuery also IIRC.
Can you create test for both async and sync?

ppekrol: Probably it would be cleaner to do:
```csharp
            if (docQuery.TheWaitForNonStaleResults)
                WaitForNonStaleResults(docQuery.Timeout);
```

talweiss1982: @ppekrol TheWaitForNonStaleResults and Timeout are both protected not sure if we should expose them, will it be okay to check it with reflection?

ayende: No, I don't want that

ppekrol: No need that, just let's do the async version and add some simple tests

talweiss1982: how to test it?

ppekrol: I've amended your commit. Check what I've did.

ppekrol (approved):

PR #8640, opened 2019-02-03
RavenDB-12826 added ToArrayAsync support to async queries

ppekrol:


valeriob:
I thought about that, thanks @ppekrol !

PR #8648, opened 2019-02-04
RavenDB-12022 : CountersStorage redesign 

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-12022


src/Raven.Server/Documents/ETL/EtlProcess.cs
```diff
@@ -229,16 +229,9 @@ public virtual IEnumerable<TExtracted> Extract(DocumentsOperationContext context
 
                             counters = new FilterCountersEnumerator(counters, stats, Database.DocumentsStorage, context);
 
-                            var counterTombstones = Database.DocumentsStorage.GetTombstonesFrom(context, CountersStorage.CountersTombstones, fromEtag, 0, int.MaxValue)
-                                .GetEnumerator();
-                            scope.EnsureDispose(counterTombstones);
-
-                            counterTombstones = new FilterTombstonesEnumerator(counterTombstones, stats, Tombstone.TombstoneType.Counter, context, maxEtag: lastDocEtag);
```
arekpalinski, line 236: We drop the concept of counter tombstones completely, right? What is the approach now?

src/Raven.Server/Documents/ETL/EtlProcess.cs
```diff
@@ -1100,14 +1069,11 @@ public override EtlProcessProgress GetProgress(DocumentsOperationContext documen
                 if (result.NumberOfDocumentTombstonesToProcess > 0)
                     result.NumberOfDocumentTombstonesToProcess -= performance.NumberOfTransformedTombstones[EtlItemType.Document];
 
-                if (result.NumberOfCountersToProcess > 0)
-                    result.NumberOfCountersToProcess -= performance.NumberOfTransformedItems[EtlItemType.Counter];
-
-                if (result.NumberOfCounterTombstonesToProcess > 0)
-                    result.NumberOfCounterTombstonesToProcess -= performance.NumberOfTransformedTombstones[EtlItemType.Counter];
+                if (result.NumberOfCounterGroupsToProcess > 0)
+                    result.NumberOfCounterGroupsToProcess -= performance.NumberOfTransformedItems[EtlItemType.Counter];
```
arekpalinski: Here we have counter group vs number of particular items in a group, don't we? E.g. we have a one group but we processed 2 counters from that group, possible?

If it's gonna work differently then maybe renaming `EtlItemType.Counter` to `EtlItemType.CounterGroup` would make more sense?

src/Raven.Server/Documents/ETL/FilterCountersEnumerator.cs
```diff
@@ -29,7 +29,7 @@ public bool MoveNext()
             {
                 var current = _counters.Current;
 
-                var doc = _docsStorage.Get(_context, current.DocumentId);
+                var doc = _docsStorage.Get(_context, current.CounterKey);
```
arekpalinski, line 32: What is actually `CounterKey` ? Is it ID of a related document?

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
```diff
@@ -356,6 +310,97 @@ bool ShouldFilterOutDeletion()
...
+            }
+
+            return counterOperations;
+        }
+
+        private bool GetCounterValueAndCheckIfShouldSkip(LazyStringValue docId, string function, BlittableJsonReaderObject.PropertyDetails prop, out long value, out bool delete)
+        {
+            value = 0;
+
+            if (prop.Value is LazyStringValue)
```
arekpalinski, line 381: Does value being `LazyStringValue` mean that counter was deleted?

src/Raven.Server/Documents/ETL/EtlProcess.cs
aviv86: yes, we dropped counter tombstones completely. 
now we handle deletes by keeping a 'tombstone-change-vector' (dbid + etag of each partial value) inside the counter-group document . in replication, when we have a delete/increment conflict we use this 'change vector' to decide which more up to date

src/Raven.Server/Documents/ETL/EtlProcess.cs
aviv86: yes `EtlItemType.CounterGroup` makes more sense

src/Raven.Server/Documents/ETL/FilterCountersEnumerator.cs
aviv86: yes for now `CounterKey` is just the document id. 
but we are planning to limit the size of a counter group to improve performance. 
and then format of `CounterKey` will be : docId | record separator | prefix

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
aviv86: yes

src/Raven.Server/Documents/ETL/EtlProcess.cs
arekpalinski: Do we ever remove those from the counter group?

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
arekpalinski: That's tricky. Could you put a comment so one won't be surprised as I was?

src/Raven.Server/Documents/ETL/EtlProcess.cs
aviv86: no we don't remove them, at least for now. we've discussed about trying to use the `TombstoneCleaner` to remove deleted counters that their 'change vector' is already merged. 
but this is not implemented yet

arekpalinski: ok

src/Raven.Server/Smuggler/Documents/StreamSource.cs
```diff
@@ -540,19 +621,10 @@ private IEnumerable<BlittableJsonReaderObject> ReadArray(INewDocumentActions act
 
                     if (_state.CurrentTokenType == JsonParserToken.EndArray)
                         break;
-                    if (actions != null)
```
aviv86: @maximburyak can you please verify that this is correct ?

aviv86: reverted this change

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
```diff
@@ -921,38 +921,31 @@ private class CounterActions : ICounterActions
...
+            private int _countersCount;
+            private readonly int _maxBatchSize;
 
             public CounterActions(DocumentDatabase database)
             {
                 _database = database;
-                _docCount = 0;
                 _cmd = new CountersHandler.SmugglerCounterBatchCommand(_database);
+
+                _maxBatchSize = sizeof(int) == IntPtr.Size || database.Configuration.Storage.ForceUsing32BitsPager
```
ppekrol: PlatformDetails.Is32bits

aviv86: done

src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
```diff
@@ -309,7 +310,13 @@ public class ReplicationFeatures
                     {
                         Replication = new SupportedFeatures.ReplicationFeatures
                         {
+
+                            // 'ReplicationFeatures.Counters' feature is NOT supported in v4.2 (and above),
+                            // it was replaced by 'ReplicationFeatures.CountersBatch'.
+                            // However, when negotiating with v4.1.x we want to agree
+                            // on a replication version that supports ClusterTransactions.
                             Counters = true,
```
ayende, line 318: I thought that we had the ability to negotiate on specific features, why do we need to enable counters as well?

src/Raven.Server/Documents/DocumentPutAction.cs
```diff
@@ -481,7 +481,6 @@ private static void ThrowInvalidDocumentId(string id)
         {
             var countersStorage = context.DocumentDatabase.DocumentsStorage.CountersStorage;
             var onDiskCounters = countersStorage.GetCountersForDocument(context, id).ToList();
```
ayende, line 483: Can we avoid the list allocation?

src/Raven.Server/Documents/DocumentPutAction.cs
```diff
@@ -499,6 +498,7 @@ private static void ThrowInvalidDocumentId(string id)
             }
 
             flags |= DocumentFlags.HasCounters;
+            onDiskCounters.Sort(StringComparer.OrdinalIgnoreCase);
```
ayende, line 501: Why are we using case insensitive sort?
For that matter, why is this sorted?

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
```diff
@@ -924,6 +926,58 @@ private unsafe void WriteCounterToServer(DocumentsOperationContext context, Repl
             }
         }
 
+        private unsafe void WriteCountersToServer(DocumentsOperationContext context, ReplicationBatchItem item)
+        {
+            using (Slice.From(context.Allocator, item.ChangeVector, out var cv))
```
ayende, line 931: Add {} here

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
```diff
@@ -924,6 +926,58 @@ private unsafe void WriteCounterToServer(DocumentsOperationContext context, Repl
...
+                                       cv.Size + // change vector
+                                       sizeof(short) + // transaction marker
+                                       sizeof(int) + // size of doc id
+                                       item.Id.Size +
+                                       sizeof(int) + // size of doc collection
+                                       item.Collection.Size + // doc collection
+                                       sizeof(int) // size of data
+                                       + item.Values.Size; // data
+
+                    if (requiredSize > _tempBuffer.Length)
```
ayende, line 945: For very large number of counters, don't we have an issue here?
That _should_ be handled by the counter splitting, though.

src/Raven.Server/Storage/Schema/Updates/Documents/From16.cs
```diff
@@ -0,0 +1,518 @@
...
+            if (readTable.GetTree(LegacyCountersSchema.Key) != null)
+            {
+                _dbId = ReadDbId(step);
+                using (step.DocumentsStorage.ContextPool.AllocateOperationContext(out DocumentsOperationContext context))
+                {
+                    string currentDocId = null;
+                    var batch = new Dictionary<string, List<CounterDetail>>();
+                    var dbIds = new HashSet<string>();
+
+                    foreach (var counterDetail in GetAllCounters(readTable, context))
```
ayende, line 97: This can be a very big list (licensing has 31 millions). We can't do that in a single transaction.
We have to break it into multiple transactions.

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -163,148 +208,142 @@ public long GetNumberOfCountersToProcess(DocumentsOperationContext context, stri
...
-                    var etag = _documentsStorage.GenerateNextEtag();
+                        if (propIndex != -1)
+                        {
+                            var prop = new BlittableJsonReaderObject.PropertyDetails();
+                            counters.GetPropertyByIndex(propIndex, ref prop);
+                            existingCounter = prop.Value;
+                            name = prop.Name; // use original casing
+                        }
+
+                        if (existingCounter == null ||
```
ayende, line 300: Can we reverse this logic?
Here we check what the value is _not_, then in the `else` use `existingCounter as BlittableJsonReaderObject.RawBlob`
I would rather we did the check here and reverse the if/else

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -314,280 +353,601 @@ private void PutCounterImpl(DocumentsOperationContext context, string documentId
...
+            {
+                counters.Modifications = counters.Modifications ?? new DynamicJsonValue(counters);
+                counters.Modifications[counterName] = existingCounter;
+            }
+
+            return modified;
         }
 
-        public long? GetCounterValue(DocumentsOperationContext context, string docId, string counterName)
+        private void AddPartialValueToExistingCounter(DocumentsOperationContext context,
```
ayende, line 786: This mutate the arg in a non obvious way, better to return `RawBlob` here

aviv86: done

src/Raven.Server/Documents/CountersStorage.cs
aviv86: done

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
aviv86: done

src/Raven.Server/Documents/DocumentPutAction.cs
aviv86: removed this
properties are already sorted in blittable

src/Raven.Server/Documents/DocumentPutAction.cs
aviv86: not sure if we can avoid that, because we need to know the number of counters (excluding deleted counters)

src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
aviv86: not exactly. we negotiate on a set of supported features. 
for `OperationTypes.Replication` we have to choose between : `ReplicationWithPullOption` (unknown to 4.1), `ReplicationAttachmentMissingVersion41`, `ReplicationAttachmentMissing`, `ReplicationBaseLine`
so if we want to have `Cluster Transactions` supported for replication with v4.1.x, our only option is `ReplicationAttachmentMissingVersion41`

src/Raven.Server/Documents/DocumentPutAction.cs
ppekrol: Can I get more details about this? Blittable does not sort properties in any way, but you might be referring to a fact than during insertion to blittable those properties are sorted. Want to double check that.

ayende: This is an array property inside the metadata. The sorting is on the values in the array

src/Raven.Client/ServerWide/Tcp/TcpConnectionHeader.cs
ayende: Okay, I see

src/Raven.Server/Documents/DocumentPutAction.cs
aviv86: Yes sorry I wasn't very clear about it.

I was referring to the `CounterGroup` blittable that I fetch the counter names from.
Each counter-name is a property name in this blittable. 

during insertion to blittable those properties are sorted. 
also when iterating over them with `GetPropertyByIndex(i, ref prop)`

PR #8656, opened 2019-02-05
RavenDB-12823 fix failing tests

karmeli87:


src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -378,7 +378,10 @@ public static void SetDatabaseChangeVector(DocumentsOperationContext context, st
         private static void ThrowOnNotUpdatedChangeVector(DocumentsOperationContext context, string changeVector)
         {
             var globalChangeVector = GetDatabaseChangeVector(context);
-            if (changeVector != globalChangeVector && 
+
+            if (globalChangeVector != changeVector && 
+                globalChangeVector != null &&
```
ppekrol, line 383: Can global change vector be null is some cases?

ayende:
LGTM

PR #8682, opened 2019-02-10
RavenDB-12855: Pal now handles pure memory mode for journal + remove …

aviviadi:
…win and posix open flags from managed code


src/librvnpal/src/win/journal.c
```diff
@@ -23,6 +23,8 @@ rvn_open_journal_for_writes(const char* file_name, int32_t transaction_mode, int
             break;
         default:
             access_flags = FILE_FLAG_NO_BUFFERING | FILE_FLAG_WRITE_THROUGH;
+            if (durability_support == DURABILITY_NOT_SUPPORTED)
+                access_flags = FILE_FLAG_NO_BUFFERING;
```
ayende: In Windows, these two flags go together, we want either both or none.

src/librvnpal/src/posix/journal.c
```diff
@@ -36,15 +36,18 @@ _free_journal_handle(struct journal_handle* handle)
...
 {
     assert(initial_file_size > 0);
 
     int32_t rc;
     
     int32_t flags = O_DSYNC | O_DIRECT;
-    if (transaction_mode == JOURNAL_MODE_DANGER)
-        flags = 0;    
+    if (durability_support == DURABILITY_NOT_SUPPORTED)
+        flags = O_DSYNC;
```
ayende, line 47: Why? If we aren't durable, we should have no flags.

ayende (changes requested):

src/librvnpal/src/posix/journal.c
aviviadi: Non Durable is different then "Danger Mode", and the reason is FS doesn't supporting O_DIRECT flag. But we do use O_DSYNC for minimum guarantee we can achieve. Otherwise we can use danger mode.   O_DIRECT is what we are missing here and it is purely the ability to ask for ack when the data is physically on the disk. 

We do want at least know the data is on the disk cache and not the OS cache with O_DSYNC..

ayende: If we don't have `O_DIRECT` then we don't really need `O_DSYCN` either, I think. 
Same as the change you made for Windows.

aviviadi: I disagree, because O_DSYNC provides a guarantee of returning from the write call only AFTER passing all the data to the disk.  
This is useful, i.e., on ungraceful reset/boot to the machine, without AC power loss. The write was done and written to the journal (but stays in the _device_ memory buffer), the system reboots and the device still have time to write is buffers physically. 

So, unless we suffer a power loss during that time to the disk, O_DSYNC can protect us from thinking we wrote something we didn't

ayende: Actually, `O_DSYNC` force a write beyond the device buffer. In what cases are we not using `O_DIRECT`, though? What cases isn't it supported?

aviviadi: NFS (also zfs iirc, at least the old versions)

aviviadi: I am changing this to zero.

ayende: Okay, I follow. Don't change it to 0 in this case, your reasoning is better.

PR #8685, opened 2019-02-10
RavenDB-12764 create a revision from incoming replication

karmeli87:
If we have an extrnal replication to a destination with revisions enabled, we have to generate a new revision upon arrival.
Since revisions keep the original change vector which is also the primary key, we can create a revision directly.


ppekrol:
We will add a test here.

ayende (approved):

PR #8687, opened 2019-02-11
RavenDB-10098 trust certificates by public key pinning hash

iftahbe:


src/Raven.Server/RavenServer.cs
```diff
@@ -931,39 +937,105 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
+                        var certKey = Constants.Certificates.Prefix + certificate.Thumbprint;
+                        var cert = ServerStore.Cluster.Read(ctx, certKey) ??
+                                   ServerStore.Cluster.GetLocalState(ctx, certKey);
+
+                        if (cert == null)
                         {
-                            authenticationStatus.Status = AuthenticationStatus.Operator;
+                            // The certificate is not explicitly registered in our server, let's see if we have a certificate
+                            // with the same public key pinning hash.
+                            var allCertKeys = ServerStore.Cluster.ItemKeysStartingWith(ctx, Constants.Certificates.Prefix, 0, int.MaxValue).ToList();
```
ayende: No. We may have a lot of them, and we don't want to process them by hand.
Instead, add a Voron index for the public key pinning hash and check on that.

src/Raven.Server/RavenServer.cs
```diff
@@ -931,39 +937,105 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
+                            allCertKeys.AddRange(ServerStore.Cluster.GetCertificateKeysFromLocalState(ctx));
+
+                            foreach (var key in allCertKeys)
+                            {
+                                var currentCert = ServerStore.Cluster.Read(ctx, key) ??
+                                                  ServerStore.Cluster.GetLocalState(ctx, key);
+
```
ayende: Don't you already have that, from when you loaded them, just before this code?

src/Raven.Server/RavenServer.cs
```diff
@@ -931,39 +937,105 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
+                                if (currentCert == null)
+                                    continue;
+
+                                if (currentCert.TryGet(nameof(CertificateDefinition.PublicKeyPinningHash), out string hash) == false)
+                                    continue;
+
+                                if (CertificateUtils.GetPublicKeyPinningHash(certificate).Equals(hash) == false)
+                                    continue;
+
+                                if (currentCert.TryGet(nameof(CertificateDefinition.Thumbprint), out string thumbprint) && certificate.Thumbprint.Equals(thumbprint))
```
ayende: Move this up?

src/Raven.Server/RavenServer.cs
```diff
@@ -931,39 +937,105 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
+                                var certBytes = certificate.Export(X509ContentType.Cert);
+
+                                var newCopy = new CertificateDefinition
+                                {
+                                    Name = currentCertDef.Name,
+                                    Certificate = Convert.ToBase64String(certBytes),
+                                    Permissions = currentCertDef.Permissions,
+                                    SecurityClearance = currentCertDef.SecurityClearance,
+                                    Thumbprint = certificate.Thumbprint,
+                                    PublicKeyPinningHash = CertificateUtils.GetPublicKeyPinningHash(certificate),
```
ayende: Do we need to compute the hash again?

src/Raven.Server/RavenServer.cs
```diff
@@ -931,39 +937,105 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
+                                    NotAfter = certificate.NotAfter,
+                                };
+
+                                lock (_certificatesLock)
```
ayende: Why? This isn't the right place to handle the actual logic of this.
You detected, based on your current state, that the certificate should be added.
Now you call to `PutCertificateCollectionInCluster`, which is going to add it to the cluster, which already make it a sync operation.
Use a dedicated command for this, that will handle  the check if it exists, remove previous copies, etc.
Note that you should also be explicit that the certificate has already been replaced, so registering this in the cluster may _fail_, and should be handled.

ayende (changes requested):

src/Raven.Server/RavenServer.cs
iftahbe: Certificate definitions are stored in the "Items" table along with other objects. Do you want to define an index on that table or move the certs to a table of their own?

ayende: Yes, given that we have new behavior for that, we should store it separately and have a dedicated index for that.

iftahbe:
For code review only, I still have to write the schema upgrade.

src/Raven.Server/RavenServer.cs
iftahbe: Added a dedicated command.

src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
```diff
@@ -461,7 +462,7 @@ public async Task AddNode()
                         using (ctx.OpenReadTransaction())
                         {
                             var key = Constants.Certificates.Prefix + certificate.Thumbprint;
-                            var readCert = ServerStore.Cluster.Read(ctx, key);
+                            var readCert = ServerStore.Cluster.GetCertificateByPrimaryKey(ctx, key);
```
ayende:
```suggestion
                            var readCert = ServerStore.Cluster.GetCertificateByThumbprint(ctx, key);
```

src/Raven.Server/RavenServer.cs
```diff
@@ -920,7 +926,7 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
             {
                 authenticationStatus.Status = AuthenticationStatus.ClusterAdmin;
             }
-            else if (wellKnown != null && wellKnown.Contains(certificate.Thumbprint, StringComparer.OrdinalIgnoreCase))
+            else if (wellKnown != null && wellKnown.Contains(certificate.Thumbprint, StringComparer.Ordinal))
```
ayende: Revert this change, thumbprints are base 16 and case insensitive

src/Raven.Server/RavenServer.cs
```diff
@@ -929,45 +935,188 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
-                            authenticationStatus.Status = AuthenticationStatus.ClusterAdmin;
-                        }
-                        else if (definition.SecurityClearance == SecurityClearance.ClusterNode)
-                        {
-                            authenticationStatus.Status = AuthenticationStatus.ClusterAdmin;
-                        }
-                        else if (definition.SecurityClearance == SecurityClearance.Operator)
+                        var certKey = Constants.Certificates.Prefix + certificate.Thumbprint;
+                        var cert = ServerStore.Cluster.GetCertificateByPrimaryKey(ctx, certKey) ??
+                                   ServerStore.Cluster.GetLocalState(ctx, certKey);
```
ayende:
```suggestion
                                   ServerStore.Cluster.GetLocalStateByThumbprint(ctx, certKey);
```

src/Raven.Server/RavenServer.cs
```diff
@@ -929,45 +935,188 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
+
+                        if (cert == null)
                         {
-                            authenticationStatus.Status = AuthenticationStatus.Operator;
+                            // The certificate is not explicitly registered in our server, let's see if we have a certificate
+                            // with the same public key pinning hash.
+                            var pinningHash = CertificateUtils.GetPublicKeyPinningHash(certificate);
```
ayende: Move to a dedicated method

src/Raven.Server/RavenServer.cs
```diff
@@ -929,45 +935,188 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
+                            var certsWithSameHash = ServerStore.Cluster.GetCertificatesByPinningHashSortedByExpiration(ctx, pinningHash);
+
+                            if (certsWithSameHash.Count > 0)
+                            {
+                                // Hash is good, let's validate it was signed by same issuer, otherwise users can use the private key to register a new cert with a different issuer.
+                                if (CertHasKnownIssuer(certificate, certsWithSameHash) == false)
+                                {
+                                    if (Logger.IsOperationsEnabled)
+                                        Logger.Operations($"You tried to use a certificate: '{certificate.Subject} ({certificate.Thumbprint})' which is not " +
```
ayende: You?

ayende: Should go to the audit log.
Should have ip / port for the connection as well

src/Raven.Server/RavenServer.cs
```diff
@@ -929,45 +935,188 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
+                                                          "registered in the cluster explicitly but is trusted implicitly by its HTTP Public Key Pinning Hash (HPKP). " +
```
ayende:
```suggestion
                                                          "registered in the cluster explicitly but is trusted implicitly by its Public Key Pinning Hash" +
```

src/Raven.Server/RavenServer.cs
```diff
@@ -929,45 +935,188 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
+                                                          "Your certificate was signed by an unknown issuer. This is not allowed - closing the connection. " +
+                                                          "To fix this, the admin can register the thumbprint of the *issuer* certificate (e.g. Let's Encrypt) in the 'Security.WellKnownIssuers.Admin' configuration entry.");
```
ayende: Provide the value that needs to be set here.

ayende: The message needs to also be shown to the user

src/Raven.Server/RavenServer.cs
```diff
@@ -929,45 +935,188 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
+                                    SecurityClearance = existingCert.SecurityClearance,
+                                    Password = existingCert.Password,
+                                    Thumbprint = certificate.Thumbprint,
+                                    PublicKeyPinningHash = pinningHash,
+                                    NotAfter = certificate.NotAfter
+                                };
+                                
+                                // This command will discard leftover certificates after the new certificate is saved.
+                                ServerStore.SendToLeaderAsync(new PutCertificateWithSamePinningHashCommand(certKey, newCertDef))
+                                    .Wait(ServerStore.ServerShutdown);
```
ayende: This isn't correct. We are now waiting for the cluster to go through.
We _must_ be able to proceed normally even without the cluster.

Note that the cluster command need to handle the case of repeated registration of the same cert.

src/Raven.Server/RavenServer.cs
```diff
@@ -929,45 +935,188 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
+
+                                if (_authAuditLog.IsInfoEnabled)
+                                    _authAuditLog.Info($"Got connection with new certificate: '{certificate.Subject} ({certificate.Thumbprint})' but it's not registered in the cluster. " +
+                                                       "Allowing the connection based on certificate's HTTP Public Key Pinning Hash (HPKP) which is trusted by the cluster. " +
```
ayende: Need ip / port.
Remove HTTP here.

src/Raven.Server/RavenServer.cs
```diff
@@ -929,45 +935,188 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
+                    {
+                        ctx.CloseTransaction();
+                    }
                 }
             }
 
             return authenticationStatus;
         }
 
+        private bool CertHasKnownIssuer(X509Certificate2 certificate, List<CertificateDefinition> certsWithSameHash)
```
ayende: Chain of server and user must be:

* Same length
* Same order
* Public key hash same for all

ayende:
```
for(var i = 0; i < userChain.Length; i++)
{
    if(userChain[i].PublicKeyHash != serverChain[i].PublicKeyHash)
         return false;
}
```

ayende: Alternatively, the user chain issuer is in one of the well known issuers.

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -89,11 +89,16 @@ public class SecurityConfiguration : ConfigurationCategory
         [ConfigurationEntry("Security.UnsecuredAccessAllowed", ConfigurationEntryScope.ServerWideOnly)]
         public UnsecuredAccessAddressRange UnsecuredAccessAllowed { get; set; }
 
-        [Description("Allow to specify well known certificate thumbprints that will be trusted by the server as cluster admins.")]
+        [Description("Well known certificate thumbprints that will be trusted by the server as cluster admins.")]
         [DefaultValue(null)]
         [ConfigurationEntry("Security.WellKnownCertificates.Admin", ConfigurationEntryScope.ServerWideOnly)]
         public string[] WellKnownAdminCertificates { get; set; }
 
+        [Description("Well known issuer thumbprints that will be used to validate a new certificate's HPKP if the issuer's certificate has changed.")]
```
ayende:
```suggestion
        [Description("Well known issuer public key hashes that will be used to validate a new certificate's PKP if the issuer's certificate has changed.")]
```

src/Raven.Server/RavenServer.cs
```diff
@@ -929,45 +935,188 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
+                catch (Exception e)
+                {
+                    if (Logger.IsOperationsEnabled)
+                        Logger.Operations(
+                            $"Cannot validate new certificate '{certificate.FriendlyName} {certificate.Thumbprint}'. Failed to build chain for the existing client certificate '{c.Thumbprint}'", e);
+                    return false;
+                }
+            }
+
+            var wellKnown = ServerStore.Configuration.Security.WellKnownIssuersCertificates;
```
ayende: openssl s_client -servername $1 -connect $1:443 | openssl x509 -pubkey -noout | openssl rsa -pubin -outform der | openssl dgst -sha256 -binary | openssl enc -base64

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
ayende: `openssl s_client -servername $1 -connect $1:443 | openssl x509 -pubkey -noout | openssl rsa -pubin -outform der | openssl dgst -sha256 -binary | openssl enc -base64`

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -950,6 +1005,61 @@ private T PutValue<T>(TransactionOperationContext context, string type, Blittabl
...
+            {
+                var certs = context.Transaction.InnerTransaction.OpenTable(CertificatesSchema, CertificatesSlice);
+                var command = (PutCertificateCommand)CommandBase.CreateFrom(cmd);
+
+                using (Slice.From(context.Allocator, command.PublicKeyPinningHash, out Slice hashSlice))
+                using (Slice.From(context.Allocator, command.Name, out Slice keySlice))
+                using (var rec = context.ReadObject(command.ValueToJson(), "inner-val"))
+                {
+                    if (_clusterAuditLog.IsInfoEnabled)
+                        _clusterAuditLog.Info($"Registering new certificate '{command.Value.Thumbprint}' in the cluster.");
```
ayende: What about permissions?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1236,6 +1351,23 @@ public IEnumerable<string> GetCertificateKeysFromLocalState(TransactionOperation
             }
         }
 
+        public IEnumerable<string> GetCertificateKeysFromCluster(TransactionOperationContext context)
```
ayende:
```suggestion
        public IEnumerable<string> GetCertificateThumbprintFromCluster(TransactionOperationContext context)
```

Change cert key to cert thumprint globally

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1439,6 +1571,57 @@ private static unsafe (string, BlittableJsonReaderObject) GetCurrentItem(Transac
...
+        public IEnumerable<CertificateDefinition> GetCertificatesByPinningHash(TransactionOperationContext context, string hash)
+        {
+            var certs = context.Transaction.InnerTransaction.OpenTable(CertificatesSchema, CertificatesSlice);
+
+            using (Slice.From(context.Allocator, hash, out Slice hashSlice))
+            {
+                foreach (var tvr in certs.SeekForwardFrom(CertificatesSchema.Indexes[CertificatesHashSlice], hashSlice, 0))
+                {
+                    var def = GetCertificateDefinition(context, tvr.Result);
+                    if (def.PublicKeyPinningHash.Equals(hash))
```
ayende: Inverse this, if this isn't right, break immediately

src/Raven.Server/ServerWide/Commands/PutCertificateWithSamePinningHashCommand.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+
+        }
+
+        public override DynamicJsonValue ValueToJson()
+        {
+            return Value?.ToJson();
+        }
+
+        public override void VerifyCanExecuteCommand(ServerStore store, TransactionOperationContext context, bool isClusterAdmin)
```
ayende, line 25: This is only valid for admin callers.
This may be _generated_ by any user, but the command is sent by the cluster itself and require cluster admin permissions

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
```diff
@@ -591,22 +621,24 @@ public Task WhoAmI()
...
                                 Thumbprint = Server.Certificate.Certificate.Thumbprint,
+                                PublicKeyPinningHash = CertificateUtils.GetPublicKeyPinningHash(Server.Certificate.Certificate),
                                 NotAfter = Server.Certificate.Certificate.NotAfter
                             };
 
                             certificate = ctx.ReadObject(serverCertDef.ToJson(), "Server/Certificate/Definition");
                         }
                     }
-                    else if (wellKnown != null && wellKnown.Contains(clientCert.Thumbprint, StringComparer.OrdinalIgnoreCase))
+                    else if (wellKnown != null && wellKnown.Contains(clientCert.Thumbprint, StringComparer.Ordinal))
```
ayende: thumbprints are case insensitive

test/SlowTests/Authentication/AuthenticationClusterTests.cs
```diff
@@ -104,5 +115,271 @@ public async Task CanReplaceClusterCert()
...
+                res = await store1.Maintenance.SendAsync(new UpdateExternalReplicationOperation(external));
+                Assert.Equal(externalList.First().TaskId, res.TaskId);
+
+                using (var session = store1.OpenAsyncSession())
+                {
+                    await session.StoreAsync(new User { Name = "Avivush" }, "users/2");
+                    await session.SaveChangesAsync();
+                }
+
+                var replicated = WaitForDocumentToReplicate<User>(store2, "users/2", 10000);
```
ayende, line 359: Check for replication _errors_, not wait for this

ayende (changes requested):

test/SlowTests/Authentication/AuthenticationClusterTests.cs
iftahbe: Currently auth errors aren't caught here. Opened a separate issue: https://issues.hibernatingrhinos.com/issue/RavenDB-13010

src/Raven.Server/ServerWide/Commands/PutCertificateWithSamePinningHashCommand.cs
iftahbe: done

src/Raven.Server/ServerWide/ClusterStateMachine.cs
iftahbe: done

src/Raven.Server/RavenServer.cs
iftahbe: done

src/Raven.Server/RavenServer.cs
iftahbe: done

src/Raven.Server/RavenServer.cs
iftahbe: done. (Comparing against a known good client cert and not the server cert)

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
iftahbe: It's too much to put in the configuration descriptiopn. I'll document it instead.

iftahbe:
Failing test: Server_store_basic_read_write_should_work
Fixing.

src/Raven.Server/Routing/RequestRouter.cs
```diff
@@ -302,6 +304,11 @@ private static void ThrowUnknownAuthStatus(RouteInformation route)
                 {
                     message = "The supplied client certificate '" + name + "' is unknown to the server. In order to register your certificate please contact your system administrator.";
                 }
+                else if (feature.Status == RavenServer.AuthenticationStatus.UnfamiliarIssuer)
+                {
+                    message = "The supplied client certificate '" + name + $"' is unknown to the server but has a known Public Key Pinning Hash. Will not use it to authenticate because the issuer is unknown. " +
```
ppekrol: 1. Use interpolation
2. use `RavenConfiguration.GetKey` to generate the configuration key

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -950,6 +1057,69 @@ private T PutValue<T>(TransactionOperationContext context, string type, Blittabl
...
+            {
+                NotifyValueChanged(context, type, index);
+            }
+        }
+
+        private void DiscardLeftoverCertsWithSamePinningHash(TransactionOperationContext context, string hash, string type, long index)
+        {
+            var certsWithSameHash = GetCertificatesByPinningHashSortedByExpiration(context, hash);
+
+            var keysToDelete = certsWithSameHash.Select(x => x.Thumbprint).ToList();
```
ppekrol: certsWithSameHash.Select(x => x.Thumbprint).Skip(Constants.Certificates.MaxNumberOfCertsWithSameHash).ToList();

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -950,6 +1057,69 @@ private T PutValue<T>(TransactionOperationContext context, string type, Blittabl
...
+            {
+                keysToDelete = keysToDelete.GetRange(Constants.Certificates.MaxNumberOfCertsWithSameHash,
+                    keysToDelete.Count - Constants.Certificates.MaxNumberOfCertsWithSameHash);
+            }
+
+            try
+            {
+                var certs = context.Transaction.InnerTransaction.OpenTable(CertificatesSchema, CertificatesSlice);
+
+                foreach (var key in keysToDelete)
```
ppekrol: what if there are no keys to delete? Do we need to notify that value changed?

src/Raven.Server/ServerWide/Commands/DeleteMultipleValuesCommand.cs
```diff
@@ -22,8 +22,8 @@ public override DynamicJsonValue ToJson(JsonOperationContext context)
 
         public override void VerifyCanExecuteCommand(ServerStore store, TransactionOperationContext context, bool isClusterAdmin)
         {
-            if (Names.Contains(ServerStore.LicenseStorageKey) || Names.Any(name => name.StartsWith(Constants.Certificates.Prefix)))
-                throw new RachisApplyException("Attempted to use DeleteMultipleValuesCommand to delete certificates or a license, use dedicated commands for this.");
+            if (Names.Contains(ServerStore.LicenseStorageKey))
+                throw new RachisApplyException("Attempted to use DeleteMultipleValuesCommand to delete a license, use dedicated command for this.");
```
ppekrol: use `nameof(DeleteMultipleValuesCommand)`

src/Raven.Server/ServerWide/Commands/DeleteValueCommand.cs
```diff
@@ -20,8 +20,8 @@ public override DynamicJsonValue ToJson(JsonOperationContext context)
 
         public override void VerifyCanExecuteCommand(ServerStore store, TransactionOperationContext context, bool isClusterAdmin)
         {
-            if (Name == ServerStore.LicenseStorageKey || Name.StartsWith(Constants.Certificates.Prefix))
-                throw new RachisApplyException("Attempted to use DeleteValueCommand to delete a certificate or a license, use dedicated commands for this.");
+            if (Name == ServerStore.LicenseStorageKey)
+                throw new RachisApplyException("Attempted to use DeleteValueCommand to delete a license, use dedicated command for this.");
```
ppekrol: use `nameof(DeleteValueCommand)`

src/Raven.Server/ServerWide/Commands/PutValueCommand.cs
```diff
@@ -26,9 +26,8 @@ public override DynamicJsonValue ToJson(JsonOperationContext context)
         public override void VerifyCanExecuteCommand(ServerStore store, TransactionOperationContext context, bool isClusterAdmin)
         {
             if (Name == ServerStore.LicenseStorageKey ||
-                Name == ServerStore.LicenseLimitsStorageKey ||
-                Name.StartsWith(Constants.Certificates.Prefix))
-                throw new InvalidOperationException("Attempted to use PutValueCommand to delete a certificate or license, use dedicated commands for this.");
+                Name == ServerStore.LicenseLimitsStorageKey)
+                throw new InvalidOperationException("Attempted to use PutValueCommand to delete a license, use dedicated command for this.");
```
ppekrol: use `nameof(PutValueCommand)`

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
```diff
@@ -348,6 +349,33 @@ public static async Task PutCertificateCollectionInCluster(CertificateDefinition
             }
         }
 
+        [RavenAction("/admin/certificates/purge", "DELETE", AuthorizationStatus.Operator)]
+        public async Task PurgeExpired()
+        {
+            var feature = HttpContext.Features.Get<IHttpAuthenticationFeature>() as RavenServer.AuthenticateConnection;
+            var clientCert = feature?.Certificate;
```
ppekrol: are we using `feature` and `clientCert` here?

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
```diff
@@ -431,10 +458,7 @@ private async Task DeleteInternal(List<string> keys)
         public Task GetCertificates()
```
ppekrol, line 455: just a note:
1. if there is a client command then we need to remove the paging
2. need to check if this affects studio? paging there?

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
ppekrol: We need to expose this in the studio? If yes, it would be 'nice' to know how many were purged.

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -553,7 +553,7 @@ public void Initialize()
             }
 
             options.SchemaVersion = SchemaUpgrader.CurrentVersion.ServerVersion;
-            options.SchemaUpgrader = SchemaUpgrader.Upgrader(SchemaUpgrader.StorageType.Server, null, null);
+            options.SchemaUpgrader = SchemaUpgrader.Upgrader(SchemaUpgrader.StorageType.Server, null, null, this);
```
ppekrol, line 558: do we need to pass ServerStore?

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
iftahbe: Not going to expose it in Studio.

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
iftahbe: I didn't remove the paging, just moved it to the inner method: GetAllRegisteredCertificates

src/Raven.Server/ServerWide/ServerStore.cs
iftahbe: Not anymore, I used a short term context instead.

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -89,11 +89,16 @@ public class SecurityConfiguration : ConfigurationCategory
...
-        [Description("Allow to specify well known certificate thumbprints that will be trusted by the server as cluster admins.")]
+        [Description("Well known certificate thumbprints that will be trusted by the server as cluster admins.")]
         [DefaultValue(null)]
         [ConfigurationEntry("Security.WellKnownCertificates.Admin", ConfigurationEntryScope.ServerWideOnly)]
         public string[] WellKnownAdminCertificates { get; set; }
 
+        [Description("Well known issuer 'Public Key Pinning Hashes' that will be used to validate a new client certificate when the issuer's certificate has changed.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.WellKnownIssuerHashes.Admin", ConfigurationEntryScope.ServerWideOnly)]
+        public string[] WellKnownIssuerHashes { get; set; }
```
ayende, line 100: When documenting this, also include the OpenSSL commands to generate this hash

src/Raven.Server/Https/AuthenticatingAdapter.cs
```diff
@@ -108,7 +110,9 @@ public async Task<IAdaptedConnection> OnConnectionAsync(ConnectionAdapterContext
 
             var tls = context.Features.Get<ITlsConnectionFeature>();
             var certificate = tls?.ClientCertificate;
-            var authenticationStatus = _server.AuthenticateConnectionCertificate(certificate);
+            var conn = context.Features.Get<IHttpConnectionFeature>();
+            var address = $"{conn?.RemoteIpAddress}:{conn?.RemotePort}";
```
ayende: Why do we need to get the address here? I assume that this is for logging / error generation?
If so, better to pass these details to the `AuthenticateConnectionCertificate` method.
I want to avoid this allocation if possible (most connections are successful)

src/Raven.Server/RavenServer.cs
```diff
@@ -880,6 +887,8 @@ public bool CanAccess(string db, bool requireAdmin)
 
             public AuthenticationStatus StatusForAudit => _status;
 
+            public string IssuerHash;
```
ayende, line 890: Why do we need this?

src/Raven.Server/RavenServer.cs
```diff
@@ -929,45 +938,179 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
+                                foreach (var kvp in definition.Permissions)
+                                {
+                                    authenticationStatus.AuthorizedDatabases.Add(kvp.Key, kvp.Value);
+                                }
                             }
                         }
                     }
+                    finally
+                    {
+                        ctx.CloseTransaction();
```
ayende: why do we need to explicitly close the transaction here? Isn't this closed by the `using` above?

src/Raven.Server/RavenServer.cs
```diff
@@ -929,45 +938,179 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
 
+        private void MaybeAllowConnectionBasedOnPinningHash(X509Certificate2 certificate, TransactionOperationContext ctx,
+            ref AuthenticateConnection authenticationStatus, ref BlittableJsonReaderObject cert, string remoteAddress)
+        {
+            // The certificate is not explicitly registered in our server, let's see if we have a certificate
+            // with the same public key pinning hash.
+            var pinningHash = certificate.GetPublicKeyPinningHash();
+            var certsWithSameHash = ServerStore.Cluster.GetCertificatesByPinningHashSortedByExpiration(ctx, pinningHash);
+
+            if (certsWithSameHash.Count > 0)
```
ayende: Let's reverse this if statement and reduce nesting in this way

src/Raven.Server/RavenServer.cs
```diff
@@ -929,45 +938,179 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
             return authenticationStatus;
         }
 
+        private void MaybeAllowConnectionBasedOnPinningHash(X509Certificate2 certificate, TransactionOperationContext ctx,
+            ref AuthenticateConnection authenticationStatus, ref BlittableJsonReaderObject cert, string remoteAddress)
+        {
+            // The certificate is not explicitly registered in our server, let's see if we have a certificate
+            // with the same public key pinning hash.
+            var pinningHash = certificate.GetPublicKeyPinningHash();
+            var certsWithSameHash = ServerStore.Cluster.GetCertificatesByPinningHashSortedByExpiration(ctx, pinningHash);
```
ayende: Here, let's just return a single item, the latest, instead of the full list. We are never use anything else here, anyway

src/Raven.Server/RavenServer.cs
```diff
@@ -929,45 +938,179 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
+                    NotAfter = certificate.NotAfter
+                };
+
+                cert = ServerStore.Cluster.GetCertificateByThumbprint(ctx, certificate.Thumbprint) ??
+                           ServerStore.Cluster.GetLocalStateByThumbprint(ctx, certificate.Thumbprint);
+                if (cert != null)
+                    return;
+
+                // This command will discard leftover certificates after the new certificate is saved.
+                ServerStore.SendToLeaderAsync(new PutCertificateWithSamePinningHashCommand(certificate.Thumbprint, newCertDef));
```
ayende: Don't we get a warning here about ignoring the return value?

src/Raven.Server/RavenServer.cs
```diff
@@ -929,45 +938,179 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
+                
+                if (_authAuditLog.IsInfoEnabled)
+                    _authAuditLog.Info($"Connection from {remoteAddress} with new certificate '{certificate.Subject} ({certificate.Thumbprint})' which is not registered in the cluster. " +
+                                       "Allowing the connection based on the certificate's Public Key Pinning Hash which is trusted by the cluster. " +
```
ayende: Let's include the previous certificate's thumbprint as well here.

src/Raven.Server/RavenServer.cs
```diff
@@ -929,45 +938,179 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
-                        }
-                        else if (definition.SecurityClearance == SecurityClearance.Operator)
+                        var cert = ServerStore.Cluster.GetCertificateByThumbprint(ctx, certificate.Thumbprint) ??
+                                   ServerStore.Cluster.GetLocalStateByThumbprint(ctx, certificate.Thumbprint);
+
+                        if (cert == null)
                         {
-                            authenticationStatus.Status = AuthenticationStatus.Operator;
+                            // If connection will be approved, cert won't be null anymore and will be assigned the relevant permissions
+                            MaybeAllowConnectionBasedOnPinningHash(certificate, ctx, ref authenticationStatus, ref cert, remoteAddress);
```
ayende: This can be a pretty expensive operation. We read certificates, parse them, compute PK, etc.
I wonder if we should put some rate limiting here to prevent ourselves from being hammered ?

ayende: Note, if we do, that should be in another PR

src/Raven.Server/RavenServer.cs
```diff
@@ -929,45 +938,179 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
+            var knownCertChain = new X509Chain();
+
+            try
+            {
+                userChain.Build(userCertificate);
+            }
+            catch (Exception e)
+            {
+                if (Logger.IsOperationsEnabled)
+                    Logger.Operations($"Cannot validate new client certificate '{userCertificate.FriendlyName} {userCertificate.Thumbprint}', failed to build the chain.", e);
```
ayende: This should be `Info` level

src/Raven.Server/RavenServer.cs
```diff
@@ -929,45 +938,179 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
+            {
+                issuerCertificate = userChain.ChainElements.Count > 1
+                    ? userChain.ChainElements[1].Certificate
+                    : userChain.ChainElements[0].Certificate;
+                issuerPinningHash = issuerCertificate.GetPublicKeyPinningHash();
+            }
+            catch (Exception e)
+            {
+                if (Logger.IsOperationsEnabled)
+                    Logger.Operations($"Cannot extract pinning hash from the client certificate's issuer '{issuerCertificate?.FriendlyName} {issuerCertificate?.Thumbprint}'.", e);
```
ayende: Info level

src/Raven.Server/RavenServer.cs
```diff
@@ -929,45 +938,179 @@ internal AuthenticateConnection AuthenticateConnectionCertificate(X509Certificat
...
+                return true;
+
+            try
+            {
+                knownCertChain.Build(knownCertificate);
+            }
+            catch (Exception e)
+            {
+                if (Logger.IsOperationsEnabled)
+                    Logger.Operations($"Cannot validate new client certificate '{userCertificate.FriendlyName} {userCertificate.Thumbprint}'. Found a known certificate '{knownCertificate.Thumbprint}' with the same hash but failed to build its chain.", e);
```
ayende: Info level

src/Raven.Server/RavenServer.cs
```diff
@@ -1314,7 +1457,7 @@ private void ListenToNewTcpConnection(TcpListener listener)
                         supported);
                 }
 
-                bool authSuccessful = TryAuthorize(Configuration, tcp.Stream, header, out var err);
+                bool authSuccessful = TryAuthorize(Configuration, tcp.Stream, header, tcpClient.Client.RemoteEndPoint.ToString(), out var err);
```
ayende: Same here, let's avoid the string allocation if we can avoid it

src/Raven.Server/RavenServer.cs
```diff
@@ -1560,15 +1703,15 @@ private bool TryAuthorize(RavenConfiguration configuration, Stream stream, TcpCo
...
 
             var certificate = (X509Certificate2)sslStream.RemoteCertificate;
-            var auth = AuthenticateConnectionCertificate(certificate);
+            var auth = AuthenticateConnectionCertificate(certificate, remoteAddress);
 
             switch (auth.Status)
             {
                 case AuthenticationStatus.Expired:
-                    msg = "The provided client certificate " + certificate.FriendlyName + " is expired on " + certificate.NotAfter;
+                    msg = "The provided client certificate " + certificate?.FriendlyName + " is expired on " + certificate?.NotAfter;
```
ayende:
```suggestion
                    msg = "The provided client certificate " + (certificate?.FriendlyName  ?? "<Missing cert>")+ " is expired on " + certificate?.NotAfter;
```

As well as elsewhere around here

ayende: and how can this be null?

src/Raven.Server/RavenServer.cs
```diff
@@ -1591,11 +1734,15 @@ private bool TryAuthorize(RavenConfiguration configuration, Stream stream, TcpCo
...
                             if (auth.CanAccess(header.DatabaseName, requireAdmin: false))
                                 return true;
-                            msg = "The certificate " + certificate.FriendlyName + " does not allow access to " + header.DatabaseName;
+                            msg = "The certificate " + certificate?.FriendlyName + " does not allow access to " + header.DatabaseName;
                             return false;
                         default:
                             throw new InvalidOperationException("Unknown operation " + header.Operation);
                     }
+                case AuthenticationStatus.UnfamiliarIssuer:
+                    msg = $"The client certificate {certificate?.FriendlyName} is not registered in the cluster explicitly but is trusted implicitly by its Public Key Pinning Hash. " +
```
ayende: It still isn't trusted. It _could_ be, but need the additional step.
Alternatively, let's tell the admin to just register the new cert

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -110,6 +128,24 @@ static ClusterStateMachine()
                 StartIndex = 0,
                 Count = 1, // Database, Separator, Commands count
             });
+
+            // We use the follow format for the certificates data
+            // { key, hash, data }
```
ayende:
```suggestion
            // { thumbprint, public key hash, data }
```

Isn't that the case?

ayende: key & hash are too generic

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -276,8 +316,15 @@ protected override void Apply(TransactionOperationContext context, BlittableJson
                     case nameof(PutLicenseLimitsCommand):
                         PutValue<LicenseLimits>(context, type, cmd, index, leader);
                         break;
+                    case nameof(PutCertificateWithSamePinningHashCommand):
+                        PutCertificate(context, type, cmd, index, serverStore);
+                        if (cmd.TryGet(nameof(PutCertificateCommand.Name), out string certKey))
```
ayende:
```suggestion
                        if (cmd.TryGet(nameof(PutCertificateWithSamePinningHashCommand.Name), out string certKey))
```

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -276,8 +316,15 @@ protected override void Apply(TransactionOperationContext context, BlittableJson
...
+                            DeleteLocalState(context, certKey);
+                        if (cmd.TryGet(nameof(PutCertificateCommand.PublicKeyPinningHash), out string hash))
```
ayende:
```suggestion
                        if (cmd.TryGet(nameof(PutCertificateWithSamePinningHashCommand.PublicKeyPinningHash), out string hash))
```

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -862,6 +921,45 @@ private void DeleteValue(TransactionOperationContext context, string type, Blitt
...
+
+        private void DeleteMultipleCertificates(TransactionOperationContext context, string type, BlittableJsonReaderObject cmd, long index, Leader leader)
+        {
+            try
+            {
+                var certs = context.Transaction.InnerTransaction.OpenTable(CertificatesSchema, CertificatesSlice);
+                var command = (DeleteCertificateCollectionFromClusterCommand)CommandBase.CreateFrom(cmd);
+
+                foreach (var name in command.Names)
+                {
```
ayende, line 945: why aren't you calling `DeleteCertificate` here?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -862,6 +921,45 @@ private void DeleteValue(TransactionOperationContext context, string type, Blitt
...
         }
 
+        private void DeleteCertificate(TransactionOperationContext context, string type, BlittableJsonReaderObject cmd, long index)
+        {
+            try
+            {
+                var certs = context.Transaction.InnerTransaction.OpenTable(CertificatesSchema, CertificatesSlice);
+                var command = (DeleteCertificateFromClusterCommand)CommandBase.CreateFrom(cmd);
+
+                using (Slice.From(context.Allocator, command.Name, out Slice keySlice))
```
ayende: why aren't you calling `DeleteCertificate` here?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -950,6 +1057,67 @@ private T PutValue<T>(TransactionOperationContext context, string type, Blittabl
...
+            {
+                var certs = context.Transaction.InnerTransaction.OpenTable(CertificatesSchema, CertificatesSlice);
+                var command = (PutCertificateCommand)CommandBase.CreateFrom(cmd);
+
+                using (Slice.From(context.Allocator, command.PublicKeyPinningHash, out Slice hashSlice))
+                using (Slice.From(context.Allocator, command.Name, out Slice keySlice))
+                using (var cert = context.ReadObject(command.ValueToJson(), "inner-val"))
+                {
+                    var existing = serverStore.Cluster.GetCertificateByThumbprint(context, command.Name) ??
+                                  serverStore.Cluster.GetLocalStateByThumbprint(context, command.Name);
```
ayende, line 1064: Is this correct? If this is registered locally, and we register it on the cluster, shouldn't we remove it now from the local?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -950,6 +1057,67 @@ private T PutValue<T>(TransactionOperationContext context, string type, Blittabl
...
+            if (keysToDelete.Count == 0)
+                return;
+
+            try
+            {
+                var certs = context.Transaction.InnerTransaction.OpenTable(CertificatesSchema, CertificatesSlice);
+
+                foreach (var key in keysToDelete)
+                {
+                    using (Slice.From(context.Allocator, key.ToLowerInvariant(), out Slice lowerKey))
```
ayende: why aren't you calling `DeleteCertificate` here?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
ayende: Still need to change this, no?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1236,6 +1409,29 @@ public IEnumerable<string> GetCertificateKeysFromLocalState(TransactionOperation
...
+            foreach (var result in certTable.SeekByPrimaryKeyPrefix(Slices.Empty, Slices.Empty, start))
+            {
+                if (take-- <= 0)
+                    yield break;
+
+                yield return GetCurrentItem(context, result.Value);
+            }
+        }
+
+        public IEnumerable<string> GetCertificateKeysFromCluster(TransactionOperationContext context)
```
ayende: What is this method used for?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1439,6 +1634,58 @@ private static unsafe (string, BlittableJsonReaderObject) GetCurrentItem(Transac
             return (key, doc);
         }
 
+        public BlittableJsonReaderObject GetCertificateByThumbprint(TransactionOperationContext context, string thumbprint)
+        {
+            var certs = context.Transaction.InnerTransaction.OpenTable(CertificatesSchema, CertificatesSlice);
+
+            using (Slice.From(context.Allocator, thumbprint, out var thumbprintSlice))
```
ayende: I think you need to call `ToLower` here, no?

src/Raven.Server/ServerWide/Commands/PutCertificateWithSamePinningHashCommand.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            if (Value.SecurityClearance == SecurityClearance.ClusterAdmin || Value.SecurityClearance != SecurityClearance.ClusterNode)
```
ayende: I don't understand this logic? You are either a cluster admin or not a cluster node?

ayende (changes requested):

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
iftahbe: OK

src/Raven.Server/Https/AuthenticatingAdapter.cs
iftahbe: OK

src/Raven.Server/RavenServer.cs
iftahbe: We need this so that it can be displayed to the user in: UnlikelyFailAuthorization() in the case of AuthenticationStatus.UnfamiliarIssuer.
I chose to put it in the connection feature just like we do with the AuthenticationStatus.

src/Raven.Server/RavenServer.cs
iftahbe: Fixed

src/Raven.Server/RavenServer.cs
iftahbe: OK

src/Raven.Server/RavenServer.cs
iftahbe: Ok

src/Raven.Server/RavenServer.cs
iftahbe: Fixed

src/Raven.Server/RavenServer.cs
iftahbe: Ok

src/Raven.Server/RavenServer.cs
iftahbe: I agree

src/Raven.Server/RavenServer.cs
iftahbe: Ok

src/Raven.Server/RavenServer.cs
iftahbe: OK

src/Raven.Server/RavenServer.cs
iftahbe: OK

src/Raven.Server/RavenServer.cs
iftahbe: OK

src/Raven.Server/RavenServer.cs
iftahbe: It cannot be null here. 
It can be null in general but then we'll geth auth.Status = NoCertificateProvided which will make us fall to the default case in the switch.

I believe I put the '?' there because of a compiler warning about the casting a few lines above.
var certificate = (X509Certificate2)sslStream.RemoteCertificate;

src/Raven.Server/RavenServer.cs
iftahbe: Yes

src/Raven.Server/ServerWide/ClusterStateMachine.cs
iftahbe: ok

src/Raven.Server/ServerWide/ClusterStateMachine.cs
iftahbe: fixed

src/Raven.Server/ServerWide/ClusterStateMachine.cs
iftahbe: fixed

src/Raven.Server/ServerWide/ClusterStateMachine.cs
iftahbe: fixed

src/Raven.Server/ServerWide/ClusterStateMachine.cs
iftahbe: fixed

src/Raven.Server/ServerWide/ClusterStateMachine.cs
iftahbe: We handle this in the caller, DeleteLocalState is called right after PutCertificate

src/Raven.Server/ServerWide/ClusterStateMachine.cs
iftahbe: fixed

src/Raven.Server/ServerWide/ClusterStateMachine.cs
iftahbe: Renamed it to GetCertificateThumbprintsFromCluster. It is used to get all the keys so that we can delete them all from local state in OnSnapshotInstalledAsync() and in DeleteAllExistingCertificates()

src/Raven.Server/ServerWide/ClusterStateMachine.cs
iftahbe: yes, I went over all the occurrences of thumbprint as a key and changed them to lower

src/Raven.Server/ServerWide/Commands/PutCertificateWithSamePinningHashCommand.cs
iftahbe: I don't get it either, seems strange (it was copied from PutCertificateCommand). Fixing both.

src/Raven.Server/ServerWide/ClusterStateMachine.cs
iftahbe: fixed

src/Raven.Server/RavenServer.cs
ayende: OK

src/Raven.Server/RavenServer.cs
ayende: That shouldn't cause it to warn about null, I think.
Let's drop this `?`

ayende (approved):

PR #8688, opened 2019-02-11
RavenDB-12853 prevent the invocation of GetDocumentStore from within …

talweiss1982:
…TestClass constructor +  modifying tests

### DIFF
--- a/test/SlowTests/Server/Documents/Queries/Dynamic/Map/MatchingAutoMapIndexesForDynamicQueries.cs
+++ b/test/SlowTests/Server/Documents/Queries/Dynamic/Map/MatchingAutoMapIndexesForDynamicQueries.cs
@@ -19,7 +19,7 @@ namespace SlowTests.Server.Documents.Queries.Dynamic.Map
 
         public void Initialize([CallerMemberName] string caller = null)
         {
-            _documentDatabase = CreateDocumentDatabase(caller: $"{GetType().Name}.{caller}");
+            _documentDatabase = CreateDocumentDatabase(caller: caller);
 
             _sut = new DynamicQueryToIndexMatcher(_documentDatabase.IndexStore);
         }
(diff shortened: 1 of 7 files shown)


test/SlowTests/Server/Documents/Queries/Dynamic/Map/MatchingAutoMapIndexesForDynamicQueries.cs
```diff
@@ -13,19 +14,20 @@ namespace SlowTests.Server.Documents.Queries.Dynamic.Map
...
-        private readonly DocumentDatabase _documentDatabase;
-        private readonly DynamicQueryToIndexMatcher _sut;
+        private DocumentDatabase _documentDatabase;
+        private DynamicQueryToIndexMatcher _sut;
 
-        public MatchingAutoMapIndexesForDynamicQueries()
+        public void Initialize([CallerMemberName] string caller = null)
         {
-            _documentDatabase = CreateDocumentDatabase();
+            _documentDatabase = CreateDocumentDatabase(caller: $"{GetType().Name}.{caller}");
```
ppekrol: This can generate very long names. Just throw if the 'caller' is null.

ppekrol (changes requested):

ppekrol:
Avoid `GetType().Name`, it is not needed and will cause path errors (128 character limit IIRC)

test/SlowTests/Server/Documents/Queries/Dynamic/Map/MatchingAutoMapIndexesForDynamicQueries.cs
talweiss1982: it is the class name . invoking method it is not too long I have run the tests and didn't get any error about too long path names (on windows) I think it is more informative but if you want I can cut it down to the caller (which is not null since I send it explicitly)

PR #8694, opened 2019-02-12
RavenDB-12672 - change incoming replication documents handling

grisha-kotler:


src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -891,6 +893,105 @@ private unsafe void ReadItemsFromSource(ref UnmanagedWriteBuffer writeBuffer, in
...
+            {
+                private readonly byte* _ptr;
+                private readonly long _allocationSize;
+                private readonly NativeMemory.ThreadStats _threadStats;
+                private long _used;
+                public long Free => _allocationSize - _used;
+
+                public Allocation(long allocationSize)
+                {
+                    _ptr = NativeMemory.AllocateMemory(allocationSize, out var threadStats);
```
maximburyak, line 966: I'm not sure we want to use not pooled memory here always.
Maybe we better store some usage statistics and change this to some pooled solution, if we have a pathological high memory usage during replication

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -683,12 +681,12 @@ public struct ReplicationItem : IDisposable
             #region Document
 
             public string Id;
-            public int Position;
             public string ChangeVector;
             public int DocumentSize;
             public string Collection;
             public long LastModifiedTicks;
             public DocumentFlags Flags;
+            public BlittableJsonReaderObject Document;
```
ppekrol, line 689: This struct is disposable. We probably need to dispose this?

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -891,6 +893,105 @@ private unsafe void ReadItemsFromSource(ref UnmanagedWriteBuffer writeBuffer, in
...
+
+                _maxSizeForContextUseInBytes = maxSizeForContextUse.GetValue(SizeUnit.Bytes);
+                var minSizeToNonContextAllocationInMb = PlatformDetails.Is32Bits ? 4 : 16;
+                _minSizeToAllocateNonContextUseInBytes = new Size(minSizeToNonContextAllocationInMb, SizeUnit.Megabytes).GetValue(SizeUnit.Bytes);
+            }
+
+            public byte* AllocateMemoryForDocument(int size)
+            {
+                TotalDocumentsSizeInBytes += size;
+                if (TotalDocumentsSizeInBytes < _maxSizeForContextUseInBytes)
```
ppekrol: Should be `<=` in this case

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -891,6 +893,105 @@ private unsafe void ReadItemsFromSource(ref UnmanagedWriteBuffer writeBuffer, in
...
+                {
+                    _context.Allocator.Allocate(size, out var output);
+                    return output.Ptr;
+                }
+
+                if (_currentAllocation == null || _currentAllocation.Free < size)
```
ppekrol, line 927: `<=` ?

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
ppekrol: This only kicks-in after (assuming default configuration) 128mb

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
grisha-kotler: done

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
grisha-kotler: Here we don't have enough space on the currently allocated chunk (added a comment)

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
grisha-kotler: Added disposal for it.
Pretty sure that this is disposed during context dispose anyway.

grisha-kotler: We do dispose it in MergedDocumentReplicationCommand

ppekrol (approved):

PR #8704, opened 2019-02-13
fix for RavenDB-12784

myarichuk:
During graph queries that need to "wait for non-stale", make sure to generate auto-indexes.
In order to do that, DynamicQueryRunnier::MatchIndex() is used to get relevant index to wait for "non-stale", and generate auto-index if needed.
Also added precaution so if the query runner will attempt to close transaction for a query that is a part of QueryQueryStep it will properly throw.


src/Sparrow/Json/BlittableJsonReaderObject.cs
```diff
@@ -26,6 +26,10 @@ public unsafe class BlittableJsonReaderObject : BlittableJsonReaderBase, IDispos
         private readonly bool _isRoot;
         private byte* _objStart;
 
+        #if DEBUG || VALIDATE
```
talweiss1982: Is this debug leftovers?

myarichuk: yes, but I'd like to leave this - as it may be useful for debugging memory issues in the future

ayende: No, lte's remove that, debug builds are already expensive enough

ayende (changes requested):
We need a test case

ppekrol (approved):

ppekrol:
@myarichuk can you check failing tests?

myarichuk:
The most recent commit should handle the failing tests.

PR #8705, opened 2019-02-13
RavenDB-12842 Capturing local stack traces throws sillently

ml054:


ayende (approved):

src/Raven.Server/Documents/Handlers/Debugging/ThreadsHandler.cs
```diff
@@ -24,11 +24,19 @@ public Task StackTrace()
...
+            using (var stream = new MemoryStream())
             {
-                OutputResultToStream(sw, threadIds.ToHashSet(), includeStackObjects);
+                using (var sw = new StreamWriter(stream))
+                {
+                    OutputResultToStream(sw, threadIds.ToHashSet(), includeStackObjects);
+                 
+                    sw.Flush();
+                    
+                    stream.Position = 0;
```
ml054, line 35: here I wanted to avoid case when we start writing to response - and HTTP 200 was sent, then we encounter an error. Right now we either send error (with 5xx status) which is valid JSON - or valid response.

PR #8716, opened 2019-02-14
RavenDB-12793 [WIP] Failed to start database 

talweiss1982:
what i have found out is that we have a convoy of notification tasks been held by the TP capturing disposed of DocumentDatabases those inside contains scratch buffer pools that seems to be holding on to dictionaries.
i have check the GCRoot and it is the ScratchBufferPool._current that is holding the data while the pool is disposed of. 
not sure my fix is the right one, i would like @arekpalinski @ayende to review this.
Anyway i'm going to run this code over the weekend and see if it still leaks.

### DIFF
--- a/src/Voron/Impl/Scratch/ScratchBufferPool.cs
+++ b/src/Voron/Impl/Scratch/ScratchBufferPool.cs
@@ -81,6 +81,8 @@ namespace Voron.Impl.Scratch
 
                     _recycleArea.RemoveFirst();
                 }
+
+                _current = null;
             });
 
             _env = env;
--- a/src/Voron/Impl/Scratch/ScratchBufferFile.cs
+++ b/src/Voron/Impl/Scratch/ScratchBufferFile.cs
@@ -48,9 +48,10 @@ namespace Voron.Impl.Scratch
             scratchPager.AllocatedInBytesFunc = () => AllocatedPagesCount * Constants.Storage.PageSize;
 
             _disposeOnceRunner = new DisposeOnce<SingleAttempt>(() =>
-            {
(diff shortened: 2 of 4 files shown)


src/Voron/Impl/Scratch/ScratchBufferPool.cs
```diff
@@ -81,6 +81,12 @@ public ScratchBufferPool(StorageEnvironment env)
 
                     _recycleArea.RemoveFirst();
                 }
+
+                if (_current.File.IsDisposed == false)
+                {
+                    _current.File.Dispose();
+                    _scratchSpaceMonitor.Decrease(_current.File.NumberOfAllocatedPages * Constants.Storage.PageSize);
+                }
```
arekpalinski: It's safe to add this but I don't think it's the root cause and might not guarantee `_current` will be _always_ disposed.

The current implementation assumes that `_current` file should be on `_scratchBuffers` dict - see `AddScratchBufferFile()` method.

I wonder if we might have a race condition between iterating over `_scratchBuffers` and `NextFile()` call - `NextFile()` called at the time of `ScratchBufferPool` being disposed? 

The only scenario I could think about is that `NextFile()` is called from `Free()` which is called in the background by the flusher - so maybe concurrently with environment dispose?

talweiss1982: Yes, my fix is wrong, I forgot what I found before and even wrote about in the issue...
The problem is that scratch file dispose of isn't cleaning the dictionaries and the _current reference is keeping those dictionaries alive

PR #8721, opened 2019-02-16
RavenDB-10999 rich cron editor

ml054:


src/Raven.Server/Web/Studio/StudioTasksHandler.cs
```diff
@@ -217,6 +222,53 @@ public Task Format()
...
+        {
+            var expression = GetQueryStringValueAndAssertIfSingleAndNotEmpty("expression");
+            CrontabSchedule crontabSchedule;
+            try
+            {
+                // will throw if the cron expression is invalid
+                crontabSchedule = CrontabSchedule.Parse(expression);
+            }
+            catch (Exception e)
+            {
```
ppekrol, line 237: It's better to throw `BadRequestException` here and let the base handler to do the work.

ml054: I will handle this in different way (like we do in other endpoints). 

I will respond with HTTP 200 if it passes validation and then it doesn't pass, but I return status in property: IsValid. 

Then we:
- avoid having error in console 
- we can distinguish between connection error (http 5xx) and parse error (http 200 and isvalid = false)

This is how we handle such things in any other controller in the same file.

PR #8727, opened 2019-02-17
RavenDB-12823 should mutate the revision flags to include `Conflicted…

karmeli87:
…` once conflict is resolved

### DIFF
--- a/src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
+++ b/src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
@@ -332,15 +332,17 @@ namespace Raven.Server.Documents.Revisions
             if (configuration == null)
                 configuration = GetRevisionsConfiguration(collectionName.Name);
 
-            using (DocumentIdWorker.GetLowerIdSliceAndStorageKey(context, id, out Slice lowerId, out Slice idPtr))
+            using (DocumentIdWorker.GetLowerIdSliceAndStorageKey(context, id, out Slice lowerId, out Slice idSlice))
             using (Slice.From(context.Allocator, changeVector, out Slice changeVectorSlice))
             {
                 var table = EnsureRevisionTableCreated(context.Transaction.InnerTransaction, collectionName);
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -721,10 +723,13 @@ private static void DeleteCountOfRevisions(DocumentsOperationContext context, Sl
 
             using (Slice.From(context.Allocator, changeVector, out var changeVectorSlice))
             {
-                var revisionExists = table.VerifyKeyExists(changeVectorSlice);
+                var revisionExists = table.ReadByKey(changeVectorSlice,out var tvr);
                 // Revisions are immutable.
```
ayende: wrong comment?

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -769,6 +774,52 @@ private static void DeleteCountOfRevisions(DocumentsOperationContext context, Sl
             }
         }
 
+        private void MarkRevisionsAsConflictedIfNeeded(DocumentsOperationContext context, Slice lowerId, Slice idSlice, DocumentFlags flags, TableValueReader tvr, Table table,
+            Slice changeVectorSlice)
+        {
+            // Revisions are immutable, but if there was a conflict we need to update the flags accordingly with the `Conflicted` flag.
```
ayende, line 779: Why is that the case? The revision itself isn't conflicted, after all. 
Or is it that it was conflicted on another node?

karmeli87: This is relevant in the case when revisions are enabled.
So node A generates a document with a revision, and node B also.
When the conflict is resolved, we need to add the `Conflicted` flag to an already existing revision.

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
karmeli87: yes, removed

PR #8728, opened 2019-02-17
RavenDB-12730: use already implemented rvn_get_current_thread_id thro…

aviviadi:
…ugh Raven.Server (with fallback to old way if using Raven.Client and add ARM64 SYS-gettid unique value)

### DIFF
--- a/src/Raven.Server/Utils/PoolOfThreads.cs
+++ b/src/Raven.Server/Utils/PoolOfThreads.cs
@@ -325,7 +325,7 @@ namespace Raven.Server.Utils
                     return;
                 }
 
-                _currentUnmanagedThreadId = PlatformDetails.GetCurrentThreadId();
+                _currentUnmanagedThreadId = NativeMemory.GetCurrentUnmanagedThreadId.Invoke();
                 _currentProcess = Process.GetCurrentProcess();
 
                 if (PlatformDetails.RunningOnLinux)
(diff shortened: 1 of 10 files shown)


src/Raven.Server/Utils/PoolOfThreads.cs
```diff
@@ -325,7 +325,7 @@ private void InitializeProcessThreads()
                     return;
                 }
 
-                _currentUnmanagedThreadId = PlatformDetails.GetCurrentThreadId();
+                _currentUnmanagedThreadId = NativeMemory.GetCurrentUnmanagedThreadId.Invoke();
```
ayende, line 328: Why? This is the same.

src/Sparrow/Platform/Posix/PerPlatformValues.cs
```diff
@@ -8,10 +8,11 @@ public class PerPlatformValues
         public class SyscallNumbers
         {
             public static long SYS_gettid =
-            (RuntimeInformation.OSArchitecture == Architecture.Arm ||
-             RuntimeInformation.OSArchitecture == Architecture.Arm64)
+            (RuntimeInformation.OSArchitecture == Architecture.Arm)
```
ayende, line 11: Why do we still need this?

src/Raven.Server/Utils/PoolOfThreads.cs
ppekrol: Server uses PAL for this. Client needs to use PlatformDetails.

PR #8731, opened 2019-02-18
RavenDB-12793 

talweiss1982:
@ayende I have reverted the task runner to its original form since the modifications made by you still kept references to the DocumentDatabase and I could see the leak in the dump we took

### DIFF
--- a/src/Raven.Server/Documents/DocumentDatabase.cs
+++ b/src/Raven.Server/Documents/DocumentDatabase.cs
@@ -289,36 +289,18 @@ namespace Raven.Server.Documents
                 SubscriptionStorage.Initialize();
                 _addToInitLog("Initializing SubscriptionStorage completed");
 
-                TaskExecutor.Execute((state) =>
+                var databaseTaskRunner = new DatabaseTaskRunner
                 {
-                    try
-                    {
-                        NotifyFeaturesAboutStateChange(record, index);
-                    }
-                    catch
-                    {
-                        // We ignore the exception since it was caught in the function itself
-                    }
-                }, null);
+                    Landlord = ServerStore.DatabasesLandlord,
+                    DatabaseName = Name,
+                    Logger = _logger,
+                    Cts = _databaseShutdown.Token,
+                    Index = index,
+                    Record = record
+                };
 
-                Task.Factory.StartNew(async _ =>
-                {
-                    try
-                    {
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -289,44 +289,90 @@ public void Initialize(InitializeOptions options = InitializeOptions.None)
...
+                }
+                catch
+                {
+                    // We ignore the exception since it was caught in the function itself
+                }
+            }
+
+            public void ClusterTransaction()
+            {
+                Task.Factory.StartNew(ExecuteClusterTransaction, Cts, TaskCreationOptions.LongRunning);
```
ayende:
```suggestion
                Task.Factory.StartNew(ExecuteClusterTransaction, Cts);
```

No reason for this.

src/Voron/Impl/Scratch/ScratchBufferFile.cs
```diff
@@ -48,9 +48,10 @@ public ScratchBufferFile(AbstractPager scratchPager, int scratchNumber)
             scratchPager.AllocatedInBytesFunc = () => AllocatedPagesCount * Constants.Storage.PageSize;
 
             _disposeOnceRunner = new DisposeOnce<SingleAttempt>(() =>
-            {
+            {                
                 _scratchPager.PagerState.DiscardOnTxCopy = true;
                 _scratchPager.Dispose();
+                Reset();
```
ayende: What is this for here?

talweiss1982: Tens of millions of dictionary entries showed up in the original dump. The leak was probably from the _current of the scratch buffer pool but still clearing the dictionaries will release the memory faster.

arekpalinski: The intend is to clear the dictionaries, right? I'd rather prefer to call `.Clear()` on those dicts explicitly in `Dispose()` method or create method `ClearDictionaries()` and using from both: `Reset()` and `Dispose()` 

I don't expect any changes in `Reset()` method in the future but just in case let's try to avoid calling it during the disposal as it's supposed to reset the state of the scratch buffer file and continue using it, not to dispose / clear the resources.

ppekrol (approved):

PR #8734, opened 2019-02-19
RavenDB-12906 fix studio compilation related to update of counter ref…

ml054:
…actor


src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
```diff
@@ -206,7 +206,10 @@ public override void Transform(RavenEtlItem item)
                             if (_script.HasLoadCounterBehaviors && _script.TryGetLoadCounterBehaviorFunctionFor(item.Collection, out var function))
                             {
                                 var counterGroupDetail = GetCounterGroupFor(item);
-                                AddCounters(item.DocumentId, counterGroupDetail.Values, function);
+                                if (counterGroupDetail != null)
```
ml054, line 209: @aviv86 please verify if this change is correct. It was throwing when document doesn't have counter. Maybe we need test for that?

aviv86: yes this is correct

ppekrol (approved):

arekpalinski (approved):

PR #8740, opened 2019-02-19
RavenDB-12854 We must not use current tx when dealing with compressio…

arekpalinski:
…n buffer pager if encryption is on. The issue was that dispose of previous transaction could zero data in compression.buffers (CryptoPager.TxOnDispose) while it was already in use by another transaction (note that tx write lock is released before the tx dispose). The solution is to use temporary tx while working with compression.buffers pager so it won't be accessed concurrently.


arekpalinski:
`Can_setup_etl_from_encrypted_to_non_encrypted_db` failed this time

PR #8742, opened 2019-02-19
RavenDB-12911 Concurrency exception during spawning CandidateAmbassad…

karmeli87:
…ors can cause the cluster to hang

### DIFF
--- a/src/Raven.Server/Rachis/Candidate.cs
+++ b/src/Raven.Server/Rachis/Candidate.cs
@@ -75,18 +75,7 @@ namespace Raven.Server.Rachis
                             continue; // we already voted for ourselves
                         var candidateAmbassador = new CandidateAmbassador(_engine, this, voter.Key, voter.Value);
                         _voters.Add(candidateAmbassador);
-                        try
-                        {
-                            _engine.AppendStateDisposable(this, candidateAmbassador);
-                        }
-                        catch (ConcurrencyException)
-                        {
-                            foreach (var ambassador in _voters)
-                            {
-                                ambassador.Dispose();
-                            }
-                            return; // we lost the election, because someone else changed our state to follower
-                        }
+                        _engine.AppendStateDisposable(this, candidateAmbassador); // concurrency exception here will dispose the current candidate and it ambassadors
                         candidateAmbassador.Start();
                     }
                     while (_running && _engine.CurrentState == RachisState.Candidate)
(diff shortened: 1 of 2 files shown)


src/Raven.Server/Rachis/CandidateAmbassador.cs
```diff
@@ -72,7 +72,7 @@ public void Dispose()
                     {
                         _engine.Log.Info(
                             $"CandidateAmbassador for {_tag}: Waited for a full second for thread {_candidateAmbassadorLongRunningWork.ManagedThreadId} " +
-                            $"({(_candidateAmbassadorLongRunningWork.Join(0) ? "running" : "finished")}) to finish, after the elections were {_candidate.ElectionResult}");
+                            $"({(_candidateAmbassadorLongRunningWork.Join(0) ? "finished" : "running")}) to finish, after the elections were {_candidate.ElectionResult}");
```
grisha-kotler, line 75: finished to finish? running to finish?

src/Raven.Server/Rachis/Candidate.cs
```diff
@@ -75,18 +75,7 @@ private void Run()
...
-                        }
-                        catch (ConcurrencyException)
-                        {
-                            foreach (var ambassador in _voters)
-                            {
-                                ambassador.Dispose();
-                            }
-                            return; // we lost the election, because someone else changed our state to follower
-                        }
+                        _engine.AppendStateDisposable(this, candidateAmbassador); // concurrency exception here will dispose the current candidate and it ambassadors
```
grisha-kotler, line 78: where is that disposed?

grisha-kotler (changes requested):

src/Raven.Server/Rachis/CandidateAmbassador.cs
ayende: The issue was that we had an a polarity issue in the log statement.
If `Join` return `true`, we are finished, not running

src/Raven.Server/Rachis/Candidate.cs
ayende: When the state of the engine changes, IIRC

ayende (approved):

src/Raven.Server/Rachis/CandidateAmbassador.cs
karmeli87: The message looks something like this:
`Waited for a full second for thread 123 (running) to finish`

grisha-kotler (approved):

PR #8752, opened 2019-02-20
RavenDB-12913 reduce memory fragmentation under 32bit system

talweiss1982:


maximburyak (approved):

ppekrol (approved):

src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -1629,6 +1635,30 @@ private CompressedPagesResult PrepareToWriteToJournal(LowLevelTransaction tx, IP
...
+        /// for allocation larger than 1MB we will alligned them to be MB alligned 
+        /// </summary>
+        /// <param name="pagesRequired"></param>
+        /// <returns></returns>
+        private static int AdjustPagesRequiredFor32Bits(int pagesRequired)
+        {
+            var bytes = pagesRequired * Constants.Storage.PageSize;
+            if (bytes < Constants.Size.Megabyte / 2)
+            {
+                pagesRequired = Bits.NextPowerOf2(bytes / Constants.Storage.PageSize + bytes % Constants.Storage.PageSize == 0 ? 0 : 1);
```
ayende:
```suggestion
                pagesRequired = Bits.NextPowerOf2(bytes) / Constants.Storage.PageSize ;
```

Shouldn't this be just this? There can't be any remainder here.

PR #8754, opened 2019-02-20
fix for RavenDB-12494

myarichuk:
In RequestRouter, make sure to check the Request and not Response headers for cluster-wide transaction's LastKnownClusterTransactionIndex. (otherwise we would not be waiting for the cluster transaction _at all_, because the waiting code never executes)

### DIFF
--- a/src/Raven.Server/Routing/RequestRouter.cs
+++ b/src/Raven.Server/Routing/RequestRouter.cs
@@ -114,7 +114,7 @@ namespace Raven.Server.Routing
                 {
                     using (reqCtx.Database.DatabaseInUse(tryMatch.Value.SkipUsagesCount))
                     {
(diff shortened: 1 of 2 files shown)


ppekrol (approved):

src/Raven.Server/Routing/RequestRouter.cs
```diff
@@ -114,7 +114,7 @@ public async ValueTask HandlePath(RequestHandlerContext reqCtx)
                 {
                     using (reqCtx.Database.DatabaseInUse(tryMatch.Value.SkipUsagesCount))
                     {
-                        if (context.Response.Headers.TryGetValue(Constants.Headers.LastKnownClusterTransactionIndex, out var value)
+                        if (context.Request.Headers.TryGetValue(Constants.Headers.LastKnownClusterTransactionIndex, out var value)
```
ayende, line 117: OUCH!

PR #8767, opened 2019-02-21
fix related to RavenDB-12494

myarichuk:
Make sure to properly wait for correct Rachis index - the condition was incorrect

### DIFF
--- a/src/Raven.Server/Routing/RequestRouter.cs
+++ b/src/Raven.Server/Routing/RequestRouter.cs
@@ -116,7 +116,7 @@ namespace Raven.Server.Routing
                     {
                         if (context.Request.Headers.TryGetValue(Constants.Headers.LastKnownClusterTransactionIndex, out var value)
                             && long.TryParse(value, out var index)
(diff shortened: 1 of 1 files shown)


ayende (approved):

ppekrol (approved):

ppekrol:
`SlowTests.Cluster.ClusterTransactionTests.SessionCompareExchangeCommands` is hanging because:
1. `ClusterStateMachine` writes the CompareExchange commands to `(CompareExchangeSchema, CompareExchange)`
2. We are notifying client about index of that command
3. Client is waiting for this command to be executed on database. but it isn't because `ExecuteClusterTransactionTask` is reading commands from `(ClusterStateMachine.TransactionCommandsSchema, ClusterStateMachine.TransactionCommands)` so last processed raft command index in database is never bumped

Also I think there might be a race with usage of `_hasClusterTransaction` and resetting it. We should probably use `SetAndResetAtomically` without resetting it inside the `ExecuteClusterTransactionTask`.

cc @karmeli87

karmeli87:
Couldn't push to @myarichuk so created https://github.com/ravendb/ravendb/pull/8778

PR #8789, opened 2019-02-25
RDBC-282 Changes API doesn't reconnect after server failure

efratshenhar:


src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1106,7 +1106,16 @@ public async Task<ServerNode> HandleServerNotResponsive(string url, ServerNode c
...
             _nodeSelector?.OnFailedRequest(nodeIndex);
             var (_, serverNode) = await GetPreferredNode().ConfigureAwait(false);
-            await UpdateTopologyAsync(serverNode, 0, true, debugTag: "handle-server-not-responsive").ConfigureAwait(false);
+            try
+            {
+                await UpdateTopologyAsync(serverNode, 0, true, debugTag: "handle-server-not-responsive").ConfigureAwait(false);
+            }
+            catch (Exception)
+            {
+                //We don't want to stop observe for changes if server down. we will wait for one to be up
```
ml054: I think you mix classes responsibilities. You are in  HandleServerNotResponsive method in request executor, but comment is about changes api. So I think we should handle that in changes api instead.

src/Raven.Client/Documents/Changes/DatabaseChanges.cs
```diff
@@ -483,8 +484,10 @@ private async Task DoWork()
                     // recover from the OnError accessing the faulty WebSocket.
                     try
                     {
-                        ConnectionStatusChanged?.Invoke(this, EventArgs.Empty);
+                        if (wasConnected)
+                            ConnectionStatusChanged?.Invoke(this, EventArgs.Empty);
 
+                        wasConnected = false;
                         _serverNode = await _requestExecutor.HandleServerNotResponsive(_url.AbsoluteUri, _serverNode, _nodeIndex, e).ConfigureAwait(false);
```
ml054: I think we should catch here and handle the exception.

ml054:
@efratshenhar Could you also write test case for that? It is quite important to avoid regressions in this piece of code.

ml054:
by test case I mean:
1. STart single server
2. Connect to changes api
3. kill server
4. wait a bit
5. start server
6. modify document
7. wait for changes api to get notification

efratshenhar:
yes

ppekrol (approved):

ppekrol:
Merging, still need a test.

talweiss1982:
https://github.com/ravendb/ravendb/pull/8799/

PR #8790, opened 2019-02-25
RavenDB-12912 Reduce Voron.Data.Tables.NewPageAllocator.Create() allo…

ayende:
…cation from 2.06MB on 32 bits

### DIFF
--- a/src/Voron/Data/Tables/NewPageAllocator.cs
+++ b/src/Voron/Data/Tables/NewPageAllocator.cs
@@ -1,9 +1,11 @@
 ﻿using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using Sparrow;
 using Sparrow.Binary;
+using Sparrow.Platform;
 using Voron.Data.BTrees;
 using Voron.Data.Fixed;
 using Voron.Global;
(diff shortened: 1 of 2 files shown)


src/Voron/Data/Tables/NewPageAllocator.cs
```diff
@@ -66,18 +67,35 @@ public void Create()
 
         private unsafe Page AllocateMoreSpace(FixedSizeTree fst)
         {
-            var allocatePage = _llt.AllocatePage(NumberOfPagesInSection);
+            int numberOfPagesToAllocate;
+            if (_llt.Environment.Options.ForceUsing32BitsPager || PlatformDetails.Is32Bits)
+                numberOfPagesToAllocate = BitmapSize;             // 256 KB
+            else
+                numberOfPagesToAllocate = NumberOfPagesInSection; // 2 MB
```
arekpalinski: Why not to set `NumberOfPagesInSection` accordingly instead? Note that have multiple usages of `NumberOfPagesInSection` there (e.g. `FreePage()`) - isn't this a problem?

ayende: The issue is that we need to keep the same structure between 32 bits & 64 bits systems.
So we can't change the size, because then we wouldn't be able to open the files between platforms.
That is why I'm just marking some of the data as busy upfront.
And in FreePage, the check is merely to catch bad data, which is fine

arekpalinski: Oh, right.  It's ok then.

Minor issue - let's extract this code to a method and use it in `GetNumberOfPreAllocatedFreePages()` to avoid code duplication

PR #8794, opened 2019-02-25
RDBC-285 Allow to define boost 0 from the client API

ayende:


ppekrol:
Can we write a test for that? Non only the client-api but also what results we would expect from boosting by 0.

src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
```diff
@@ -950,7 +950,7 @@ public void Boost(decimal boost)
             if (whereToken == null)
                 throw new InvalidOperationException("Missing where clause");
 
-            if (boost <= 0m)
+            if (boost < 0m)
                 throw new ArgumentOutOfRangeException(nameof(boost), "Boost factor must be a positive number");
```
ml054, line 954: probably we need to update this message as well.

PR #8802, opened 2019-02-26
RDBC-285 Add test for zero boosting

ayende:


test/SlowTests/Issues/RDBC-285.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace SlowTests.Issues
+{
+    public class RDBC_285 : RavenTestBase
+    {
+        [Fact]
+        public void CanQueryWithBoostEqZero()
+        {
+            using (var store = GetDocumentStore())
+            {
+                store.Maintenance.Send(new CreateSampleDataOperation());
```
ppekrol, line 20: Test failed. We need to wait for indexing after this I think?

ayende: Yep

PR #8803, opened 2019-02-26
RavenDB-12598 adding test for changes-api with aggressive caching

talweiss1982:


ayende:
I don't understand the test. You are only testing the aggressive caching on node 1, but probably should check that reads from the other nodes are also going through and refreshing aggressive caching, no?

talweiss1982:
The claim was that other nodes in the database group aren't getting changes when using aggressive caching. I have set aggressive caching to all 3 stores making sure each can communicate only to its original server.

maximburyak:
@ayende the test makes sure that the changes api receives the right notifications, assuming that the aggressive caching will respond correctly. We probably should also check the behavior of the caching as well

ayende:
You are checking this on a single store , for both the modifications and the changes.
We need to do that on different stores

talweiss1982:
no, I'm not I'm checking on all stores, i have added the load too and test still passes.

ppekrol (approved):

talweiss1982:
@ayende i have modified the test, it still  passes.

ayende (approved):

PR #8810, opened 2019-02-27
RavenDB-12875, RavenDB-12887: normal parallelism in tests

aviviadi:


aviviadi:
https://github.com/ravendb/ravendb/pull/8812 instead

PR #8818, opened 2019-02-27
RavenDB-12967 Race condition in data exploration means that we aren't…

ayende:
… sending any results out

- Underlying issue was that we were disposing the cancellation token too early

### DIFF
--- a/Raven.Database/Server/Controllers/StreamsController.cs
+++ b/Raven.Database/Server/Controllers/StreamsController.cs
@@ -246,69 +246,67 @@ namespace Raven.Database.Server.Controllers
 
             var msg = GetEmptyMessage();
 
-            using (var cts = new CancellationTokenSource())
+            var cts = new CancellationTokenSource();
+            var timeout = cts.TimeoutAfter(TimeSpan.FromSeconds(timeoutSeconds));
+            var indexQuery = new IndexQuery
             {
-                var timeout = cts.TimeoutAfter(TimeSpan.FromSeconds(timeoutSeconds));
-                var indexQuery = new IndexQuery
+                PageSize = pageSize,
+                Start = 0,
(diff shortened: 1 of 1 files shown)


ayende:
The test failure isn't related to the changes

Raven.Database/Server/Controllers/StreamsController.cs
```diff
@@ -246,69 +246,67 @@ public async Task<HttpResponseMessage> Exploration(string collection)
 
             var msg = GetEmptyMessage();
 
-            using (var cts = new CancellationTokenSource())
+            var cts = new CancellationTokenSource();
```
talweiss1982, line 249: Shouldn't we dispose of the Cts? maybe we need to override the Dispose of the HttpContent and dispose of the Cts there.

Raven.Database/Server/Controllers/StreamsController.cs
```diff
@@ -246,69 +246,67 @@ public async Task<HttpResponseMessage> Exploration(string collection)
...
+            var timeout = cts.TimeoutAfter(TimeSpan.FromSeconds(timeoutSeconds));
```
talweiss1982, line 250: We are also seems to be leaking a timer here

Raven.Database/Server/Controllers/StreamsController.cs
ayende: No, `StreamQueryContent` will handle that for us.

Raven.Database/Server/Controllers/StreamsController.cs
ayende: This is handled by `StreamQueryContent`, if there is an error here that isn't caught, we'll let the finalizer handle this.
Expected to be very rare

PR #8823, opened 2019-02-28
RavenDB-12943 added QueryMethodConverter

ppekrol:


src/Raven.Client/Documents/Linq/QueryMethodConverter.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.Linq.Expressions;
+using Raven.Client.Documents.Session;
+
+namespace Raven.Client.Documents.Linq
+{
+    public abstract class QueryMethodConverter
+    {
+        public struct Parameters<T>
```
lahma, line 9: Could be a `readonly struct` with readonly fields and then `TryConvertQueryMethod` could take as `in` parameter?

ayende (approved):

src/Raven.Client/Documents/Linq/QueryMethodConverter.cs
ppekrol: Wouldn't the implication be that the person referencing the client API would need to switch the project to C# 7.2+ in order to implement the converter?

arekpalinski:
@ppekrol LGTM

PR #8824, opened 2019-02-28
Raven db 12955

ml054:


src/Raven.Server/Web/RequestHandler.cs
```diff
@@ -573,18 +574,57 @@ private static void ThrowInvalidAuthStatus(RavenServer.AuthenticationStatus? sta
...
 
+        private static bool IsOriginAllowed(string origin, ServerStore serverStore)
+        {
+            if (serverStore.Server.Certificate.Certificate == null)
+            {
+                // running in unsafe mode - since server can be access via multiple urls/aliases accept them 
+                return true;
+            }
+            
+            var topology = serverStore.GetClusterTopology();
```
ayende, line 604: Do we need to worry about case sensitivity here?

ml054: Good question, I will check that using empirical mode. :)

ml054:
I've found one bug. Closing for now.

src/Raven.Server/Web/RequestHandler.cs
ml054: It needs to be case insensitive. Thanks. I will fix that.

PR #8827, opened 2019-02-28
RavenDB-12957 fixing concurrent access to context in IncomingReplicat…

talweiss1982:
…ionHandler

### DIFF
--- a/src/Sparrow/Json/JsonOperationContext.cs
+++ b/src/Sparrow/Json/JsonOperationContext.cs
@@ -403,16 +403,11 @@ namespace Sparrow.Json
             }
         }
 
-
-#if DEBUG || VALIDATE
-        private readonly LongReference _getMemory = new LongReference();
-#endif
-
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public AllocatedMemoryData GetMemory(int requestedSize)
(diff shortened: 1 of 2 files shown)


src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -403,47 +403,66 @@ private static void ThrowParentWasDisposed()
             }
         }
 
+
+#if DEBUG || VALIDATE
+        private readonly LongReference _getMemory = new LongReference();
```
ayende: Why do we have multiple of them? We don't actually care if the concurrent usage is between methods or just in one method. I think we should have just one of these

ayende (changes requested):

src/Sparrow/Json/JsonOperationContext.cs
talweiss1982: They call each other

ayende: That being the case, let's do this differently and say that we want to do:
`Interlocked.CompareExchange(ref val, 0, CurrentThreadId)`

This way, in the same thread, it will work, but you can catch it if it is between threads.

PR #8830, opened 2019-03-01
Raven db 12955

ml054:


ppekrol (approved):

arekpalinski (approved):

src/Raven.Server/Web/RequestHandler.cs
```diff
@@ -573,18 +574,72 @@ private static void ThrowInvalidAuthStatus(RavenServer.AuthenticationStatus? sta
...
+        public static void SetupCORSHeaders(HttpContext httpContext, ServerStore serverStore)
         {
+            httpContext.Response.Headers.Add("Vary", "Origin");
+            
+            var requestedOrigin = httpContext.Request.Headers["Origin"];
 
-            httpContext.Response.Headers.Add("Access-Control-Allow-Origin", httpContext.Request.Headers["Origin"]);
+            if (requestedOrigin.Count == 0)
+            {
+                // no CORS headers needed
```
ayende, line 585: I don't think this is the case. Here we are running with no security, but we if we return no results, then the request won't work, would it?
In that case, a cluster using HTTP won't work either, unless I'm missing something?

ml054: Request contains origin when it is cross-site request, or request with not-simple methods (GET, HEAD). 

So, if it is same origin - no headers needed - origin isn't passed. 
If it is https with cross-site, then origin is filled in - we go further and filter based on topology
If it is http with cross-site, origin is filled in  - but we send back what we have.

ayende: Okay, I see.

ayende (approved):

PR #8837, opened 2019-03-01
RavenDB-12811 dashboard updates for non cluster admin operators

ml054:


src/Raven.Server/Dashboard/DatabasesInfo.cs
```diff
@@ -21,6 +22,18 @@ public override DynamicJsonValue ToJson()
             json[nameof(Items)] = new DynamicJsonArray(Items.Select(x => x.ToJson()));
             return json;
         }
+
+        public override DynamicJsonValue ToJsonWithFilter(Func<string, bool> filter)
+        {
+            var items = new DynamicJsonArray(Items.Where(x => filter(x.Database)).Select(x => x.ToJson()));
```
ppekrol: let's avoid LINQ, some of these notifications are created often when Studio is connected

ml054: done

arekpalinski (approved):

src/Raven.Server/Dashboard/DatabasesInfo.cs
```diff
@@ -21,6 +22,25 @@ public override DynamicJsonValue ToJson()
             json[nameof(Items)] = new DynamicJsonArray(Items.Select(x => x.ToJson()));
             return json;
         }
+
+        public override DynamicJsonValue ToJsonWithFilter(Func<string, bool> filter)
```
ayende, line 26: Why not return an empty array?

src/Raven.Server/NotificationCenter/NotificationCenterWebsocketWriter.cs
```diff
@@ -56,18 +55,6 @@ public async Task WriteNotifications(Func<string, bool> shouldWriteByDb)
                             continue;
                         }
 
-                        if (shouldWriteByDb != null &&
-                            shouldWriteByDb((string)tuple.Item2["Database"]) == false)
-                        {
-                            if (sp.ElapsedMilliseconds > 5000)
-                            {
-                                sp.Restart();
-                                await SendHeartbeat();
```
ayende, line 65: We actually probably need this here. If we have a lot of events / messages, but none that we can send, we need to do the heartbeat here.

src/Raven.Server/Dashboard/DatabasesInfo.cs
ml054: if we return null, then we don't append anything to the queue

src/Raven.Server/NotificationCenter/NotificationCenterWebsocketWriter.cs
ml054: see note above - when we return null such message is not queued, so we timeout on queue.get and heartbird is sent

src/Raven.Server/Dashboard/DatabasesInfo.cs
ml054: moreover we need to filter inner structures and we don't want to send database names users can't access.

src/Raven.Server/NotificationCenter/NotificationCenterWebsocketWriter.cs
ml054: here: https://github.com/ravendb/ravendb/pull/8837/files/a22b9775326dfcd390fe541d5d948d731f9990ac#diff-88c47ac394f9600c53afb59060c78863L52

ayende: Okay, looks good, then

PR #8858, opened 2019-03-03
RavenDB-12874: CanPreformSeveralClusterTransactions - add NoTimeouts(…

aviviadi:
…) to prevent flackiness due to slow response while slow testing

### DIFF
--- a/test/SlowTests/Cluster/ClusterTransactionTests.cs
+++ b/test/SlowTests/Cluster/ClusterTransactionTests.cs
@@ -110,69 +110,78 @@ namespace SlowTests.Cluster
         public async Task CanPreformSeveralClusterTransactions(int numberOfNodes)
         {
             NoTimeouts();
-            var numOfSessions = 10;
-            var docsPerSession = 2;
-            var leader = await CreateRaftClusterAndGetLeader(numberOfNodes);
-            using (var store = GetDocumentStore(new Options
+            try
             {
-                Server = leader,
-                ReplicationFactor = numberOfNodes
-            }))
-            {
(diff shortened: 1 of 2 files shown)


test/SlowTests/Cluster/ClusterTransactionTests.cs
```diff
@@ -109,6 +109,7 @@ private string RandomString(int length)
         [InlineData(5)]
         public async Task CanPreformSeveralClusterTransactions(int numberOfNodes)
         {
+            NoTimeouts();
```
karmeli87, line 112: Can we do that? I think we are reusing the servers, no?

aviviadi: so? are we relying on timeouts in the tests?

aviviadi: I can set no timeouts only for the server in hand, and restore the timeout once the test is over

test/SlowTests/Cluster/ClusterTransactionTests.cs
```diff
@@ -109,70 +109,80 @@ private string RandomString(int length)
         [InlineData(5)]
         public async Task CanPreformSeveralClusterTransactions(int numberOfNodes)
         {
```
ayende, line 111: We should do this only on our server, so we need `UseLocalServer` here too, I think

aviviadi: Not needed: `var leader = await CreateRaftClusterAndGetLeader(numberOfNodes);` is being used in this test (creates new servers)

PR #8861, opened 2019-03-03
RavenDB-11725: Fix test LZ4

aviviadi:


test/StressTests/Voron/HugeTransactions.cs
```diff
@@ -131,7 +131,7 @@ private static unsafe void VerifyData
             Assert.Equal(desired, val);
         }
 
-        [Theory64Bit(Skip = "RavenDB-11725")]
+        [Theory64Bit]
```
ayende, line 134: What is the difference?

aviviadi: at v4.2 it passes.  at 4.1 it fails.  I am searching the problem for 4.1 right now.. Please wait before merging..

aviviadi:
still failing on v4.1

PR #8863, opened 2019-03-03
RavenDB-11470 Restore should not generate new change vectors

karmeli87:


ayende (approved):

src/Raven.Client/Documents/Smuggler/DatabaseSmugglerOptions.cs
```diff
@@ -32,6 +32,8 @@ public DatabaseSmugglerOptions()
 
         public bool SkipRevisionCreation { get; set; }
 
+        public bool KeepOriginalChangeVector { get; set; }
```
ppekrol: Let's move it to server-side options

ppekrol (approved):

PR #8867, opened 2019-03-04
RavenDB-12736 cluster stabilization

karmeli87:


src/Raven.Server/Rachis/Follower.cs
```diff
@@ -411,7 +411,74 @@ private void NegotiateWithLeader(TransactionOperationContext context, LogLengthN
...
+            using (var cts = new CancellationTokenSource())
+            {
+                var readAndCommitSnapshotTask = Task.Run(() => ReadAndCommitSnapshot(context, snapshot, cts.Token), cts.Token);
+                try
+                {
+                    WaitForTaskCompletion(readAndCommitSnapshotTask, "ReadAndCommitSnapshot");
+                }
+                catch
+                {
+                    cts.Cancel();
```
ayende, line 426: In here, do we want to wait until `readAndCommitSnapshotTask` is done?
Because otherwise we might dispose the `cts` before it is noticed by the task

src/Raven.Server/Rachis/Follower.cs
```diff
@@ -411,7 +411,74 @@ private void NegotiateWithLeader(TransactionOperationContext context, LogLengthN
...
+                    throw;
+                }
+            }
+
+            _debugRecorder.Record("Snapshot was received and committed");
+
+            // notify the state machine, we do this in an async manner, and start
+            // the operator in a separate thread to avoid timeouts while this is
+            // going on
+            var task = Task.Run(() => _engine.SnapshotInstalledAsync(snapshot.LastIncludedIndex));
```
ayende, line 436: Do we need `cts` here as well?

src/Raven.Server/Rachis/Follower.cs
```diff
@@ -411,7 +411,74 @@ private void NegotiateWithLeader(TransactionOperationContext context, LogLengthN
...
+            _engine.Timeout.Defer(_connection.Source);
+        }
+
+        private void WaitForTaskCompletion(Task task, string debug)
+        {
+            var sp = Stopwatch.StartNew();
+
+            var timeToWait = (int)(_engine.ElectionTimeout.TotalMilliseconds / 4);
+
+            using (_engine.ContextPool.AllocateOperationContext(out JsonOperationContext context))
```
ayende, line 463: I think it might be a better idea to move the context into the while.
The reason is that we want to reset / renew it on each iteration, rather than risk it getting too big if we wait for a lot of time

src/Raven.Server/Rachis/Follower.cs
```diff
@@ -425,9 +492,9 @@ private void NegotiateWithLeader(TransactionOperationContext context, LogLengthN
                     }
 
                     //This is okay to ignore because we will just get the committed entries again and skip them
-                    ReadInstallSnapshotAndIgnoreContent(context);
+                    ReadInstallSnapshotAndIgnoreContent();
```
ayende, line 495: Reading the content may take a while (2GB, frex). I think we ought to provide the `cts` here as well

src/Raven.Server/Rachis/FollowerAmbassador.cs
```diff
@@ -472,18 +472,22 @@ private void SendSnapshot(Stream stream)
             using (context.OpenReadTransaction())
             {
                 var earliestIndexEntry = _engine.GetFirstEntryIndex(context);
-                if (_followerMatchIndex >= earliestIndexEntry)
+               
+                if (_followerMatchIndex >= earliestIndexEntry ||
+                    _followerMatchIndex == earliestIndexEntry - 1) // In case the first entry is the next entry to send
```
ayende, line 477: `_followerMatchIndex >= earliestIndexEntry-1` ?

src/Raven.Server/Rachis/Follower.cs
karmeli87: done

src/Raven.Server/Rachis/Follower.cs
karmeli87: `SnapshotInstalledAsync` has two possible blocking calls.
1. opening a write tx
2. reload license

In both cases I'm not sure if cancellation is necessary. 
though we can cancel between the two calls.

src/Raven.Server/Rachis/Follower.cs
karmeli87: done

src/Raven.Server/Rachis/Follower.cs
karmeli87: done

src/Raven.Server/Rachis/FollowerAmbassador.cs
karmeli87: This was only to make the code more clear and to add the comment in the same line.

ayende: ok

src/Raven.Server/Rachis/Follower.cs
ayende: ok, not needed, then

PR #8869, opened 2019-03-04
RavenDB-12860 [For Review][WIP] Tcp multiple ips

talweiss1982:
@ayende @ppekrol please review this PR since there is a lot of margin for errors here

### DIFF
--- a/src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
+++ b/src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
@@ -182,8 +182,8 @@ namespace Raven.Client.Documents.Subscriptions
                     await requestExecutor.ExecuteAsync(command, context, sessionInfo: null, token: token).ConfigureAwait(false);
                 }
 
-                string choosenUrl = null;
-                (_tcpClient, choosenUrl) = await TcpUtils.ConnectAsyncWithPriority(command.Result, requestExecutor.DefaultTimeout).ConfigureAwait(false);
+                string chosenUrl = null;
+                (_tcpClient, chosenUrl) = await TcpUtils.ConnectAsyncWithPriority(command.Result, requestExecutor.DefaultTimeout).ConfigureAwait(false);
                 _tcpClient.NoDelay = true;
                 _tcpClient.SendBufferSize = 32 * 1024;
                 _tcpClient.ReceiveBufferSize = 4096;
@@ -199,7 +199,7 @@ namespace Raven.Client.Documents.Subscriptions
                     Version = SubscriptionTcpVersion ?? TcpConnectionHeaderMessage.SubscriptionTcpVersion,
                     ReadResponseAndGetVersionCallback = ReadServerResponseAndGetVersion,
                     DestinationNodeTag = CurrentNodeTag,
-                    DestinationUrl = command.Result.Url
+                    DestinationUrl = chosenUrl
                 };
                 _supportedFeatures = TcpNegotiation.NegotiateProtocolVersion(context, _stream, parameters);
 
--- a/src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
+++ b/src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
@@ -157,7 +157,7 @@ namespace Raven.Client.Documents.Subscriptions
 
         private async Task<Stream> ConnectToServer(CancellationToken token)
         {
-            var command = new GetTcpInfoForRemoteTaskCommand("Subscription/" + _dbName, _dbName, _options?.SubscriptionName);
+            var command = new GetTcpInfoForRemoteTaskCommand("Subscription/" + _dbName, _dbName, _options?.SubscriptionName, verifyDatabase:true);
 
             var requestExecutor = _store.GetRequestExecutor(_dbName);
 
@@ -266,8 +266,6 @@ namespace Raven.Client.Documents.Subscriptions
             {
                 if (connectionStatus.Exception.Contains(nameof(DatabaseDoesNotExistException)))
                     DatabaseDoesNotExistException.ThrowWithMessage(_dbName, connectionStatus.Message);
-                if (connectionStatus.Exception.Contains(nameof(DatabaseDisabledException)))
-                    throw new DatabaseDisabledException(_dbName + " is disabled");
             }
             if (connectionStatus.Type != SubscriptionConnectionServerMessage.MessageType.ConnectionStatus)
                 throw new Exception("Server returned illegal type message when expecting connection status, was: " +
--- a/src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
+++ b/src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
@@ -12,6 +12,7 @@ using System.Net.Sockets;
 using System.Threading;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
+using Raven.Client.Documents.Commands;
 using Raven.Client.Exceptions;
 using Raven.Client.Exceptions.Database;
 using Raven.Client.Exceptions.Documents.Subscriptions;
@@ -156,7 +157,7 @@ namespace Raven.Client.Documents.Subscriptions
 
         private async Task<Stream> ConnectToServer(CancellationToken token)
         {
-            var command = new GetTcpInfoCommand("Subscription/" + _dbName, _dbName);
+            var command = new GetTcpInfoForRemoteTaskCommand("Subscription/" + _dbName, _dbName, _options?.SubscriptionName);
 
             var requestExecutor = _store.GetRequestExecutor(_dbName);
 
@@ -181,7 +182,8 @@ namespace Raven.Client.Documents.Subscriptions
                     await requestExecutor.ExecuteAsync(command, context, sessionInfo: null, token: token).ConfigureAwait(false);
                 }
 
-                _tcpClient = await TcpUtils.ConnectAsync(command.Result.Url, requestExecutor.DefaultTimeout).ConfigureAwait(false);
+                string choosenUrl = null;
+                (_tcpClient, choosenUrl) = await TcpUtils.ConnectAsyncWithPriority(command.Result, requestExecutor.DefaultTimeout).ConfigureAwait(false);
                 _tcpClient.NoDelay = true;
                 _tcpClient.SendBufferSize = 32 * 1024;
                 _tcpClient.ReceiveBufferSize = 4096;
--- a/src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
+++ b/src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
@@ -266,6 +266,8 @@ namespace Raven.Client.Documents.Subscriptions
             {
                 if (connectionStatus.Exception.Contains(nameof(DatabaseDoesNotExistException)))
                     DatabaseDoesNotExistException.ThrowWithMessage(_dbName, connectionStatus.Message);
+                if (connectionStatus.Exception.Contains(nameof(DatabaseDisabledException)))
+                    throw new DatabaseDisabledException(_dbName + " is disabled");
             }
             if (connectionStatus.Type != SubscriptionConnectionServerMessage.MessageType.ConnectionStatus)
                 throw new Exception("Server returned illegal type message when expecting connection status, was: " +
--- a/src/Raven.Client/ServerWide/Commands/TcpConnectionInfo.cs
+++ b/src/Raven.Client/ServerWide/Commands/TcpConnectionInfo.cs
@@ -7,7 +7,7 @@ namespace Raven.Client.ServerWide.Commands
         public int Port;
         public string Url;
         public string Certificate;
-        public string[] TcpServerUrls;
+        public string[] ServerUrls;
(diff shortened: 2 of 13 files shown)


src/Raven.Client/ServerWide/Commands/TcpConnectionInfo.cs
```diff
@@ -7,15 +7,26 @@ public class TcpConnectionInfo
...
         public DynamicJsonValue ToJson()
         {
-            return new DynamicJsonValue
+            var res =  new DynamicJsonValue
             {
                 [nameof(Port)] = Port,
                 [nameof(Url)] = Url,
                 [nameof(Certificate)] = Certificate
             };
+            if (TcpPriorityServerUrls != null)
```
ayende: Under what conditions will this be `null` ? 
I think we want to always have this, and at a minimum, use the current url as the only entry here.
That simplify most of the logic for the client (even though we'll need to check both for backward compact reasons)

src/Raven.Server/Config/Categories/CoreConfiguration.cs
```diff
@@ -22,6 +22,11 @@ public class CoreConfiguration : ConfigurationCategory
         [ConfigurationEntry("ServerUrl.Tcp", ConfigurationEntryScope.ServerWideOnly)]
         public string[] TcpServerUrls { get; set; }
 
+        [Description("When provided those urls are going to be published in the tcp info and be used for internal tcp communication with increasing priority where the first item is the most prioritize")]
```
ayende:
```suggestion
        [Description("When provided those URLs are used for all TCP communications to this node. The list is used in the order provided, and the first URL that can be successfully connected to will be used. This is meant to allow inter-cluster communication via private IPs, while external clients will use external IPs.  The provided URLs can be raw IPs or utilize hostnames.")]
```

src/Raven.Server/Config/Categories/CoreConfiguration.cs
```diff
@@ -22,6 +22,11 @@ public class CoreConfiguration : ConfigurationCategory
         [ConfigurationEntry("ServerUrl.Tcp", ConfigurationEntryScope.ServerWideOnly)]
         public string[] TcpServerUrls { get; set; }
```
ayende, line 23: Why do we need this and `TcpPriorityServerUrls` ? 
Can we have just one such configuration? What happens if we provide multiple ips here?

src/Raven.Server/RavenServer.cs
```diff
@@ -995,8 +995,28 @@ private void StartSnmp()
 
         public TcpListenerStatus StartTcpListener()
         {
+            void EnsureTcpPriorityServerUrlsAreListenTo(int defaultPort, HashSet<(string Host, int Port)> valueTuples, TcpListenerStatus tcpListenerStatus)
+            {
+                if (Configuration.Core.TcpPriorityServerUrls != null)
```
ayende: These guys are meant for external use, not internal one.
Internally, we'll usually bind to `0.0.0.0`, then we'll use the hostname to decide how to connect.

Another issue is that we probably should use `TcpServerUrl.Public`

ayende (changes requested):

src/Raven.Client/ServerWide/Commands/TcpConnectionInfo.cs
talweiss1982: The default is null.
This is something that bypass the public tcp url and as such i'm not sure we won't to always use it, we want to use it only for server->server communication.
If we want to use this always we might as well just use 'TcpServerUrls' (which i don't think is a good idea)

src/Raven.Server/RavenServer.cs
talweiss1982: I don't understand what you mean by "...we probably should use TcpServerUrl.Public"
Here I chose to make sure that the hostname that is in the priority list listens to I could throw if they are not part of the {TcpServerUrls U PublicTcpServerUrl}

src/Raven.Server/Config/Categories/CoreConfiguration.cs
talweiss1982: That is what I understood the request was if not then the issue should be more clear.
TcpServerUrls are automatically generated and populated and have validations that there is no more than one item there for some cases.
I found it wrong to use it.
We want to have multiple ips that is the whole point of this code, that server could have a priority list of ips they can connect to and it contains the private tcp url by the order of which you prefer to connect.
This is what i did now the server->server communication is using those ips to open a tcp client and not the public tcp url.

talweiss1982:
@ayende Please clarify the behavior you intended because this review is confusing and I'm blocked

talweiss1982:
@maximburyak please review this PR

src/Raven.Client/ServerWide/Commands/TcpConnectionInfo.cs
```diff
@@ -7,15 +7,25 @@ public class TcpConnectionInfo
...
-            return new DynamicJsonValue
+            var res =  new DynamicJsonValue
             {
                 [nameof(Port)] = Port,
                 [nameof(Url)] = Url,
                 [nameof(Certificate)] = Certificate
             };
+
+            var array = new DynamicJsonArray();
+            foreach (var url in TcpServerUrls)
```
ayende: Don't we need to handle a null here?

src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
```diff
@@ -156,7 +157,7 @@ private Task Run(CancellationToken ct)
 
         private async Task<Stream> ConnectToServer(CancellationToken token)
         {
-            var command = new GetTcpInfoCommand("Subscription/" + _dbName, _dbName);
+            var command = new GetTcpInfoForRemoteTaskCommand("Subscription/" + _dbName, _dbName, _options?.SubscriptionName, verifyDatabase:true);
```
ayende, line 160: Is there a compatibility concern here?
https://ravendb.net/docs/article-page/4.1/csharp/client-api/faq/backward-compatibility

So a 4.2 client will not work with a 4.1 or 4.0 server, so this is fine, according to the docs.

src/Raven.Client/Util/TcpUtils.cs
```diff
@@ -144,5 +144,34 @@ private static TcpClient NewTcpClient(TimeSpan? timeout, bool useIPv6)
             Debug.Assert(tcpClient.Client != null);
             return tcpClient;
         }
+
+        internal static async Task<(TcpClient Client, string Url)> ConnectAsyncWithPriority(TcpConnectionInfo info, TimeSpan? tcpConnectionTimeout)
+        {
+            string selectedUrl = null;
+            TcpClient tcpClient = null;
+            
+            foreach (var url in info.TcpServerUrls)
```
ayende: We need to either fall back to the just the `Url` if the `TcpServerUrls` are not specified or give a good error here.
Otherwise, we'll get NRE when connecting to an old server.

src/Raven.Client/ServerWide/Commands/TcpConnectionInfo.cs
```diff
@@ -7,15 +7,25 @@ public class TcpConnectionInfo
         public int Port;
         public string Url;
         public string Certificate;
-
+        public string[] TcpServerUrls;
```
ayende: `Tcp` is already in the class name.

src/Raven.Client/Util/TcpUtils.cs
```diff
@@ -144,5 +144,34 @@ private static TcpClient NewTcpClient(TimeSpan? timeout, bool useIPv6)
...
+            {
+                try
+                {
+                    tcpClient = await ConnectAsync(url, tcpConnectionTimeout).ConfigureAwait(false);
+                    selectedUrl = url;
+                    break;
```
ayende: Just return here, no?

src/Raven.Server/Web/System/TcpConnectionInfoHandler.cs
```diff
@@ -70,22 +72,25 @@ public Task GetRemoteTaskTopology()
         }
 
         [RavenAction("/info/remote-task/tcp", "GET", AuthorizationStatus.RestrictedAccess)]
-        public Task GetRemoteTaskTcp()
+        public async Task GetRemoteTaskTcp()
         {
             var remoteTask = GetStringQueryString("remote-task");
             var database = GetStringQueryString("database");
+            var verifyDatabase = GetBoolValueQueryString("verify-database", false);
+            if (verifyDatabase.HasValue && verifyDatabase.Value)
```
ayende, line 80: Why do we have this? Worst case, we'll get an error on the TCP connection, no?

ayende (changes requested):

src/Raven.Client/ServerWide/Commands/TcpConnectionInfo.cs
talweiss1982: No, because we always populate the server list and if it is not set it will contain the ServerTcpUrl as a sole entry

src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
talweiss1982: It was the requirement that i'll move subscription to use the remote task, i could modify the database specific info/tcp to run the new logic and have a new entry point for the database/info/tcp that will handle this (this means duplicating code between the handlers), would that be better?

src/Raven.Client/Util/TcpUtils.cs
talweiss1982: for the subscription you're correct we might have null need to handle it because of the client

src/Raven.Client/Util/TcpUtils.cs
talweiss1982: for the case where the original server url was not tested we test it below this might mean we will try it twice if it appears in the url list, we could check if it contains it and return

src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
ayende: When did we get the `/tcp/info/remote` endpoint?

src/Raven.Client/ServerWide/Commands/TcpConnectionInfo.cs
ayende: What happens if you are reading from a 4.1 or a 4.0 server?

src/Raven.Client/Util/TcpUtils.cs
ayende: Unless I miss something, this function returns the first working url.
I don't see any double checks here

ayende (changes requested):

src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
talweiss1982: We got it for Pull replication not sure when that was added

src/Raven.Client/Util/TcpUtils.cs
talweiss1982: you're correct

src/Raven.Server/Web/System/TcpConnectionInfoHandler.cs
talweiss1982: There is a failing subscription test where we disabled the database and the get tcp command is not failing and expected to fail.
The subscription code assumes that the get tcp info will fail if the database is disabled.
I have fixed it so if the active connection could deal with disabling a database but @maximburyak  reviewed it and thought it is the wrong fix and that the tcp info call should fail if the database is disabled and this is why i added this.

talweiss1982:
@ayende i need your comment regarding the get tcp info, you requested that i'll use the same connection as Pull replication and this is why i use the different command:
1. should I take care of backward compatibility against v4.0, v4.1 server?
2. should i modify subscription to recognize disabled database after the connection is established and remove the validation for the database been active?

ayende:
1) Yes
2) what is the current behavior in this context?

src/Raven.Client/Util/TcpUtils.cs
```diff
@@ -149,27 +149,27 @@ internal static async Task<(TcpClient Client, string Url)> ConnectAsyncWithPrior
         {
             string selectedUrl = null;
             TcpClient tcpClient = null;
-            
-            foreach (var url in info.TcpServerUrls)
+
+            if (info.TcpServerUrls != null)
```
ayende, line 153: Shouldn't this be `ServerUrls`

src/Raven.Client/Util/TcpUtils.cs
```diff
@@ -149,27 +149,27 @@ internal static async Task<(TcpClient Client, string Url)> ConnectAsyncWithPrior
         {
             string selectedUrl = null;
```
ayende, line 150: Can remove this var

ayende (changes requested):

src/Raven.Client/Util/TcpUtils.cs
talweiss1982: not sure what's going on here it seems not to fully push my changes...

talweiss1982:
The current behavior is subscription expects to fail to get TCP info if the database is disabled

talweiss1982:
I'm going to close this PR and open a clean one since something seems to have gone wrong here

PR #8875, opened 2019-03-05
RavenDB-6931 making sure we don't modify @attachments via patch

talweiss1982:


src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
```diff
@@ -233,6 +235,73 @@ BlittableObjectInstance UpdateOriginalDocument()
...
+                    if(Sparrow.Memory.Compare(attachmentsForOriginalDocument.BasePointer, attachmentsForModifiedDocument.BasePointer, attachmentsForOriginalDocument.Count) != 0)
+                    {
+                        ThrowPatchModificationToAttachments(id);
+                    }
+                }
+
+            }
+        }
+
+        private static Dictionary<string, BlittableJsonReaderObject> GetAttachmentsFromArray(BlittableJsonReaderArray attachmentsArray)
```
ayende: I don't think that this method is used anywhere?

ayende:
Looks good, otherwise

src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
```diff
@@ -185,6 +186,7 @@ protected PatchResult ExecuteOnDocument(DocumentsOperationContext context, strin
                         else if (DocumentCompare.IsEqualTo(originalDoc, modifiedDocument, tryMergeMetadataConflicts: true) != DocumentCompareResult.Equal)
```
ppekrol: I think it would be better to check why `DocumentCompare.IsEqual` does not return `DocumentCompareResult.NotEqual | DocumentCompareResult.AttachmentsNotEqual` flag

src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
talweiss1982: seems i forgot to clean this one up

src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
talweiss1982: I did check it and it is tailored for put document and not generalized.
If I'll change it I'll mess up conflicts...

talweiss1982:
I'll check the failing tests but its wired since I didn't change anything regarding parameters

src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
ppekrol: I think we need to look into that, to avoid iterating through document twice

talweiss1982: I already looked into it and its no good

talweiss1982: i'm not iterating the document i'm accessing a single property

talweiss1982: There seems to be wrong logic behind the compare code i'm afraid this has high probabbility to break revisions and counter code but i guess we'll need to see the test result to know

maximburyak:
That feels like a fundamental change for 4.2, maybe we should better merge it post 4.2 release?

ayende:
I would rather do that in a point release, rather than in 4.2.1, or something like that.

ayende:
Note that this just validate something that would cause the user grief if they did (and we expect none would)

talweiss1982:
Breaks behavior

PR #8879, opened 2019-03-06
RavenDB-13025: Add stacktrace to CatastrophicFailure

aviviadi:


ayende (commented):
What about when the exception is re-thrown? Do we have it as well?

ayende:
Consider the case of a failure to commit, which is catastrophic. 
The next write transaction would throw that we failed and in an invalid state. But would we have that stack trace there?

src/Voron/StorageEnvironmentOptions.cs
```diff
@@ -238,7 +238,7 @@ protected StorageEnvironmentOptions(VoronPathSetting tempPath, IoChangesNotifica
         public void SetCatastrophicFailure(ExceptionDispatchInfo exception)
         {
             _catastrophicFailure = exception;
-            _catastrophicFailureNotification.RaiseNotificationOnce(_environmentId, ToString(), exception.SourceException);
+            _catastrophicFailureNotification.RaiseNotificationOnce(_environmentId, ToString(), exception.SourceException, Environment.StackTrace);
```
arekpalinski: Why do we need to use `Environment.Stacktrace`? Wouldn't it be the same as `ExceptionDispatchInfo.SourceException` that we have here (`exception.SourceException`) ?

aviviadi:
> Consider the case of a failure to commit, which is catastrophic.
> The next write transaction would throw that we failed and in an invalid state. But would we have that stack trace there?

now it does : https://github.com/ravendb/ravendb/pull/8879/commits/61f1ca56deba68706685259905ed3134337b739b

src/Voron/StorageEnvironmentOptions.cs
ayende: That capture the case _until_ this point, with `Environment.Stacktrace` catching _From_ this point

PR #8883, opened 2019-03-06
RavenDB-12946 Restore CheckIfIdAlreadyIncluded for load with includes

efratshenhar:


src/Raven.Client/Documents/Session/Operations/LoadOperation.cs
```diff
@@ -52,7 +52,7 @@ public LoadOperation ById(string id)
             if (_ids == null)
                 _ids = new[] { id };
 
-            if (_session.IsLoadedOrDeleted(id))
+            if (_session.IsLoadedOrDeleted(id) && (_includes == null))
```
ayende: This does only part of the work. In particular, id doesn't check that the included documents are already there. 
If they are, we don't need to hit the server at all

ayende (changes requested):

src/Raven.Client/Documents/Session/Operations/LoadOperation.cs
ppekrol: This is to error prone. Sooner or later someone will do `ById` first and then `Include`. Need to move that logic to `CreateRequest` method.

maximburyak: @ayende this actually already works, not treated here, see gist: https://gist.github.com/maximburyak/af6438d1847fa999960b99d2dcd1576d
@ppekrol I think that this a topic of a seperate issue.

ayende: where is the code that actually handles this?

ppekrol: It's in the `CreateRequest` method. I will sit with @efratshenhar on Sunday and we will figure it out.

src/Raven.Client/Documents/Session/Operations/LoadOperation.cs
```diff
@@ -12,11 +13,10 @@ internal class LoadOperation
         private readonly InMemoryDocumentSessionOperations _session;
         private static readonly Logger Logger = LoggingSource.Instance.GetLogger<LoadOperation>("Client");
 
-        private string[] _ids;
+        private readonly List<string> _ids = new List<string>();
```
ppekrol: better to use array only, then you will not have to do `.ToArray()` in `CreateRequest`

ppekrol (changes requested):

ppekrol (approved):

PR #8885, opened 2019-03-06
RavenDB-11346: Add systemd daemon of ravendb.service

aviviadi:


gregolsky:
We don't bundle those scripts in release packages at the moment (just have run.sh right now). Should we?

ppekrol:
Yes, but I think those scripts needs to be more interactive and they need to check for the actual OS before running?

gregolsky:
Nowadays it's `systemd` for daemons and `systemctl` for management pretty much everywhere AFAIK - CentOS, Ubuntu and other major distros. I think we can check if `systemctl` is there and continue if so, exit with a msg otherwise. Do you concur @aviviadi ?

aviviadi:
> Nowadays it's `systemd` for daemons and `systemctl` for management pretty much everywhere AFAIK - CentOS, Ubuntu and other major distros. I think we can check if `systemctl` is there and continue if so, exit with a msg otherwise. Do you concur @aviviadi ?

I agree : https://github.com/ravendb/ravendb/pull/8885/commits/0442420f6ddee8c82d4c5fe2aa5255371a4cf748

scripts/linux/install-daemon-upstart.sh
```diff
@@ -0,0 +1,205 @@
+#!/bin/bash
+
+SCRIPT_TITLE="install-daemon"
+PKG_INSTALLER="apt-get" # supports : sudo ${PKG_INSTALLER} install <pkgname> 
+
+CHK_PKGS=( "bzip2" "tar" )
```
gregolsky, line 6: Wondering if we should check if `dotnet` prerequisites are installed in this script or is it out of scope?

Here's the ref from install script https://github.com/ravendb/ravendb/blob/v4.1/install_runtime_prerequisites.sh

scripts/linux/install-daemon-upstart.sh
```diff
@@ -0,0 +1,205 @@
...
+
+REPORT_FILE='/tmp/install-daemon.ravendb.report.log'
+OUT_FILE='/tmp/install-daemon.ravendb.out.log'
+
+RAVEN_UBUNTU_PKG="../../artifacts/ubuntu.`lsb_release -r | awk '{print $2}'`-x64"
+RDB_WATCHDOG="ravendb.watchdog.sh"
+
+if [ "$1" == "" ]
+then
+	RAVEN_PARAMS="--Raven/ServerUrl=http://localhost:8080 --Raven/LogsDirectory=RDB_RAVENDB_PATH/../Logs --Raven/DataDir=RDB_RAVENDB_PATH/../Databases --Raven/StudioDirectory=RDB_RAVENDB_PATH/Server"
```
gregolsky: Shouldn't we just follow what's in `settings.json` instead of overriding `ServerUrl` and other stuff?
rvn service installation does not override anything with command line opts IIRC.

Going further (not for this time around I think) we probably should think of lay out stuff in Linux as people are used to - you know "when in Rome.." :) I mean using places like `/etc/ravendb/settings.json` and `/var/lib/ravendb` for data etc.

ayende: Ideally, I want to have just one place for these settings, and `settings.json` is the correct one, I think

scripts/linux/install-daemon-upstart.sh
aviviadi: dotnet is not a prerequisite any longer as RavenDB downloaded tarball contains it..

scripts/linux/install-daemon-upstart.sh
aviviadi: https://github.com/ravendb/ravendb/pull/8885/commits/57495de8703a719eac044db2deea831ef96ef552

scripts/linux/install-daemon-upstart.sh
gregolsky: Of course. .netcore runtime is part of the package. I meant .netcore prerequisites like `libicu` and friends.

ppekrol: @aviviadi ?

aviviadi: I am right now on it. answer in 10-15 minutes

aviviadi: @gregolsky @ppekrol Verified on scratch install of latest raspbian stretch and RavenDB v4.1.4 - nothing was needed at all..

gregolsky: Raspbian ok. However they have listed these as "required" on e.g. Ubuntu and CentOS:
https://github.com/dotnet/docs/blob/master/docs/core/linux-prerequisites.md#ubuntu
https://github.com/dotnet/docs/blob/master/docs/core/linux-prerequisites.md#centos-and-fedora

On the other hand we don't install or check C++ redist for people on Windows when doing `install-as-service.ps1` so maybe we should *not* do that here to stay consistent.
Related docs page: https://ravendb.net/docs/article-page/4.1/csharp/start/getting-started#prerequisites

Thoughts?

ppekrol: Fine by me

aviviadi: @gregolsky Required for SDK, not for ready to run published package.  It is like taking the safe side, but most of the packages are actually not needed in order to run RavenDB.

gregolsky: well just heads up and no action required I think, but this is what I got on a fresh CentOS 7
```
FailFast:                                                                                      │
Couldn't find a valid ICU package installed on the system. Set the configuration flag System.Gl│
obalization.Invariant to true if you want to run with no globalization support.                │
                                                                                               │
   at System.Environment.FailFast(System.String)                                               │
   at System.Globalization.GlobalizationMode.GetGlobalizationInvariantMode()                   │
   at System.Globalization.GlobalizationMode..cctor()                                          │
   at System.Globalization.CultureData.CreateCultureWithInvariantData()                        │
   at System.Globalization.CultureData.get_Invariant()                                         │
   at System.Globalization.CultureInfo..cctor()                                                │
   at System.StringComparer..cctor()                                                           │
   at System.AppDomain.InitializeCompatibilityFlags()                                          │
   at System.AppDomain.Setup(System.Object)                                                    │
./run.sh: line 30: 11823 Aborted                 ./Raven.Server --browser                      │
[centos@ip-XXXXXX RavenDB]$
```

PR #8886, opened 2019-03-06
RavenDB-12898 Limit size of counter groups

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-12898

### DIFF
--- a/src/Raven.Server/Documents/CountersStorage.cs
+++ b/src/Raven.Server/Documents/CountersStorage.cs
@@ -702,9 +702,9 @@ namespace Raven.Server.Documents
 
                             using (var counterGroupKey = TableValueToString(context, (int)CountersTable.CounterKey, ref tvr))
                             {
-                                if (entriesToUpdate.TryGetValue(counterGroupKey, out var tuple))
+                                if (entriesToUpdate.TryGetValue(counterGroupKey, out var putCountersData))
                                 {
-                                    data = tuple.Data;
+                                    data = putCountersData.Data;
                                 }
                                 else
                                 {
@@ -722,13 +722,11 @@ namespace Raven.Server.Documents
                                         throw new InvalidDataException(
                                             $"Local Counter-Group document '{counterKeySlice}' is missing '{DbIds}' property. Shouldn't happen");
                                     }
-
-                                  
-
+                                 
                                     // clone counter group key  
                                     var scope = context.Allocator.Allocate(counterGroupKey.Size, out var output);
 
-                                    tuple = new PutCountersData
+                                    putCountersData = new PutCountersData
                                     {
                                         Data = data,
                                         DbIdsHolder = new DbIdsHolder(dbIds),
@@ -740,7 +738,7 @@ namespace Raven.Server.Documents
                                     counterGroupKey.CopyTo(output.Ptr);
                                     var clonedKey = context.AllocateStringValue(null, output.Ptr, output.Length);
 
-                                    entriesToUpdate.Add(clonedKey, tuple);
+                                    entriesToUpdate.Add(clonedKey, putCountersData);
                                 }
 
                                 if (data.TryGet(Values, out BlittableJsonReaderObject localCounters) == false)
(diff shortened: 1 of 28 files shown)


src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -502,45 +631,43 @@ private BlittableJsonReaderObject WriteNewCountersDocument(DocumentsOperationCon
...
-                            if (data.TryGet(DbIds, out BlittableJsonReaderArray dbIds) == false)
-                            {
-                                throw new InvalidDataException($"Local Counter-Group document '{counterKey}' is missing '{DbIds}' property. Shouldn't happen");
-                            }
-                            var dbIdsHolder = new DbIdsHolder(dbIds);
-
-                            if (data.TryGet(Values, out BlittableJsonReaderObject localCounters) == false)
+                        if (table.SeekOnePrimaryKeyPrefix(documentKeyPrefix, out _) == false)
+                        {
+                            data = context.ReadObject(sourceData, documentId, BlittableJsonDocumentBuilder.UsageMode.ToDisk);
```
ayende:
```suggestion
                            // simplest case of having no counters for this documents, can use the raw data from the source as-is
                            data = context.ReadObject(sourceData, documentId, BlittableJsonDocumentBuilder.UsageMode.ToDisk);
```

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -502,45 +631,43 @@ private BlittableJsonReaderObject WriteNewCountersDocument(DocumentsOperationCon
...
+                                tvb.Add(collectionSlice);
+                                tvb.Add(context.TransactionMarkerOffset);
+
+                                table.Set(tvb);
                             }
+                        }
 
+                        else
+                        {
                             if (sourceData.TryGet(DbIds, out BlittableJsonReaderArray sourceDbIds) == false)
```
ayende: Let's add this validation _before_ the previous `if`. So we validate the structure before we save it as-is

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -493,6 +565,63 @@ private BlittableJsonReaderObject WriteNewCountersDocument(DocumentsOperationCon
...
+                builder.WriteObjectEnd();
+
+                builder.WriteObjectEnd();
+                builder.FinalizeDocument();
+
+                return builder.CreateReader();
+            }
+        }
+
         public void PutCounters(DocumentsOperationContext context, string documentId, string collection, string changeVector,
```
ayende, line 639: This is pretty complex, and I think this requires additional code review (and probably some comments).
Let's sit on this tomorrow.

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -411,8 +414,14 @@ private static void SplitCounterGroup(DocumentsOperationContext context, Collect
                     tvb.Add(fst.BasePointer, fst.Size);
                     tvb.Add(collectionSlice);
                     tvb.Add(context.TransactionMarkerOffset);
-                    table.Update(storageId, tvb);
-
+                    if (storageId.HasValue)
+                    {
+                        table.Update(storageId.Value, tvb);
```
ayende: Why do you need that?

aviv86: when I call `SplitCounterGroup` from `PutCounters`  I do not have the table value reader at hand anymore (because now I do that at the end, after the loop). 
when I call `SplitCounterGroup` from `IncrementCounter` I do have the value reader

ayende: Yes, but why does this matter? Shouldn't we just always call `Set` ?

arekpalinski:
@ayende Regarding this commit: 1aa9d80b7da7d665d70e36dd1856836d5e5a638b. I'm currently rewriting `From16.cs` schema upgrade as it requires much more work - we need to split it into multiple transactions, we need to delete relevant tables, indexes etc

arekpalinski:
Oh, I see `From16.cs` contains more changes there. I'll rebase my work against this branch then.

arekpalinski:
@aviv86 @ayende 
I've submitted the following PR https://github.com/aviv86/ravendb/pull/1 which splits the counters migration into multiple commits. Please review it there.

Also please have a look at the following comment: https://issues.hibernatingrhinos.com/issue/RavenDB-12898#focus=streamItem-67-35156.0-0

grisha-kotler (approved):

ppekrol (approved):

PR #8888, opened 2019-03-07
RavenDB-6931 assert @attachments not modified

talweiss1982:


ppekrol (approved):

talweiss1982:
SlowTests.Authentication.AuthenticationLetsEncryptTests.CanGetLetsEncryptCertificateAndRenewIt failure seems unrelated

src/Raven.Server/Documents/DocumentCompare.cs
```diff
@@ -69,19 +76,30 @@ private static DocumentCompareResult IsMetadataEqualTo(BlittableJsonReaderObject
...
                 return DocumentCompareResult.NotEqual;
             }
 
-            return ComparePropertiesExceptStartingWithAt(currentMetadata, objMetadata, true, tryMergeConflicts);
+            return ComparePropertiesExceptStartingWithAt(currentMetadata, objMetadata, true, options);
+        }
+
+        private static void ThrowAttachmentsModificationsDetected(string documentId)
+        {
+            throw new InvalidOperationException("Illegal modifications of '@attachmetns' detected for document: " + documentId);
```
ayende:
```suggestion
            throw new InvalidOperationException("Illegal modifications of '@attachments' detected for document: " + documentId);
```

src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
```diff
@@ -182,7 +182,12 @@ protected PatchResult ExecuteOnDocument(DocumentsOperationContext context, strin
 
                             result.Status = PatchStatus.Created;
                         }
-                        else if (DocumentCompare.IsEqualTo(originalDoc, modifiedDocument, tryMergeMetadataConflicts: true) != DocumentCompareResult.Equal)
+                        else if (DocumentCompare.IsEqualTo(originalDoc, modifiedDocument, 
+                                    new DocumentCompare.DocumentCompareOptions
```
ayende: Let's make this into a constant, like the `Default` field there

src/Raven.Server/Documents/Replication/ConflictManager.cs
```diff
@@ -199,7 +199,7 @@ private static void InvalidConflictWhenThereIsNone(string id)
 
             if (existingDoc != null)
             {
-                var compareResult = DocumentCompare.IsEqualTo(existingDoc.Data, incomingDoc, true);
+                var compareResult = DocumentCompare.IsEqualTo(existingDoc.Data, incomingDoc, new DocumentCompare.DocumentCompareOptions{TryMergeMetadataConflicts = true});
```
ayende: Let's make this into a constant, too

src/Raven.Server/Documents/DocumentCompare.cs
```diff
@@ -11,8 +11,15 @@ namespace Raven.Server.Documents
 {
     public static class DocumentCompare
     {
+        public struct DocumentCompareOptions
+        {
+            public bool TryMergeMetadataConflicts;
+            public bool ThrowOnAttachmentModifications;
+            public string DocumentId;
```
ayende: Am I missing something, or is this never set?

ayende (changes requested):

src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
talweiss1982: done

src/Raven.Server/Documents/Replication/ConflictManager.cs
talweiss1982: done

src/Raven.Server/Documents/DocumentCompare.cs
talweiss1982: I forgot to set it but if we will use this than we can't use constant options so its allocation vs better erroring, i think it is best to catch the exception on patch only and add a better message there

src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
```diff
@@ -182,17 +182,30 @@ protected PatchResult ExecuteOnDocument(DocumentsOperationContext context, strin
...
+                            try
+                            {
+                                compareResult = DocumentCompare.IsEqualTo(originalDoc, modifiedDocument,
+                                    DocumentCompare.DocumentCompareOptions.MergeMetadataAndThrowOnAttachmentModification);
+                            }
+                            catch (InvalidOperationException ioe)
                             {
-                                putResult = _database.DocumentsStorage.Put(context, originalDocument.Id,
-                                    originalDocument.ChangeVector, modifiedDocument, null, null, originalDocument.Flags);
+                                throw new InvalidOperationException("Illegal modifications of '@attachments' detected for document: " + id);
```
ayende: Better to just state that there was an error processing this and use the `ioe` as inner exception

ayende (changes requested):

src/Raven.Server/Documents/DocumentCompare.cs
ayende: Sounds good, yes

src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
ppekrol: Done

PR #8893, opened 2019-03-07
RavenDB-13039 Shared global indexes can become out of sync if a tx mo…

ayende:
…ve them between embedded / large states on different tables.

Avoid heap corruption / AVE when we are using two tables with a shared index and a particular access pattern

### DIFF
--- a/src/Voron/Data/Fixed/FixedSizeTree.cs
+++ b/src/Voron/Data/Fixed/FixedSizeTree.cs
@@ -35,13 +35,22 @@ namespace Voron.Data.Fixed
         private readonly int _maxEmbeddedEntries;
 
         private NewPageAllocator _newPageAllocator;
-        private RootObjectType? _type;
         private FastStack<FixedSizeTreePage> _cursor;
         private int _changes;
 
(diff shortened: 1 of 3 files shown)


src/Voron/Data/Fixed/FixedSizeTree.cs
```diff
@@ -35,13 +35,22 @@ public unsafe partial class FixedSizeTree : IDisposable
...
         private int _changes;
 
         public LowLevelTransaction Llt => _tx;
 
-        internal RootObjectType? Type => _type;
+        internal RootObjectType? Type
+        {
+            get
+            {
+                var header = _parent.DirectRead(_treeName);
```
ppekrol, line 47: Is there an option to calculate that only when we are switching from one RootObjectType to another?

ayende: We did that, but the issue is that we may have multiple _instances_ of this pointing to the same location, that is why we have to read it anew every time

ppekrol (approved):

arekpalinski (approved):

PR #8894, opened 2019-03-07
RavenDB-13043 Studio is an installable progressive web app (PWA)

JudahGabriel:
To see it in action, you need:

- The Studio running under HTTPS. (Required for PWA and service worker APIs.)
- Use Chrome (Other browsers currently have spotty support for this.)
- Use the Studio for more than 30 seconds or so. (Chrome heuristic.)

Once you've done this, you can install the Studio either through Chrome's app menu or through the Studio UI.

Chrome menu:
image

Studio UI:
image

Doing either will cause the browser to prompt for install:

image

If the user accepts installation, Raven Studio will launch in its own window without browser chrome or address bar:

image

And the app will be in the start menu and on the taskbar:

image


src/Raven.Studio/typescript/common/pwaInstaller.ts
```diff
@@ -0,0 +1,70 @@
+﻿/// <reference path="../../typings/tsd.d.ts" />
+
+/**
+ * Service that aids in performing installation for progressive web apps (PWAs).
+ * For more info, see https://developers.google.com/web/fundamentals/app-install-banners
+ * */
+class pwaInstaller {
+    
+    private deferredInstallPrompt: BeforeInstallPromptEvent | null = null;
```
ml054: we don't use explicit null notation in current version of studio. Let's be consistent.

src/Raven.Studio/typescript/common/pwaInstaller.ts
```diff
@@ -0,0 +1,70 @@
...
+        if (installPrompt) {
+            this.deferredInstallPrompt = installPrompt;
+        }
+    }
+
+    get canInstallApp(): boolean {
+        return !!this.deferredInstallPrompt;
+    }
+
+    promptInstallApp(): Promise<InstallPromptResult> | null {
```
ml054: remove |null

src/Raven.Studio/typescript/viewmodels/database/settings/studioConfiguration.ts
```diff
@@ -3,6 +3,7 @@ import getStudioConfigurationCommand = require("commands/resources/getStudioConf
 import databaseStudioConfigurationModel = require("models/database/settings/databaseStudioConfigurationModel");
 import eventsCollector = require("common/eventsCollector");
 import saveStudioConfigurationCommand = require("commands/resources/saveStudioConfigurationCommand");
+import pwaInstaller = require("../../../common/pwaInstaller");
```
ml054: let's avoid relative paths - r.js has issues with packing this properly

src/Raven.Studio/typescript/viewmodels/database/settings/studioConfiguration.ts
```diff
@@ -37,6 +42,16 @@ class studioConfiguration extends viewModelBase {
             .execute()
             .always(() => this.spinners.save(false));
     }
+
+    installApp() {
```
ml054: which has to moved to server wide studio configuration. Current it is in db specific studio configuration.

src/Raven.Studio/wwwroot/index.html
```diff
@@ -1,60 +1,61 @@
 ﻿<!DOCTYPE html>
 <html>
-    <head>
-        <title>Raven.Studio</title>
+<head>
```
ml054: can we roll back formatting here? It hurts git blame.

src/Raven.Studio/wwwroot/manifest.json
```diff
@@ -1,18 +1,31 @@
 {
-    "name": "RavenDB",
+    "name": "Raven Studio",
```
ml054, line 2: I wonder if our studio could have some node name in it's name. What if you have cluster - then you end up with few apps with the same name. Btw. did you check if you can install more than one app?

ml054:
@JudahGabriel  Btw what about links with target blank, or fetching data from different cluster nodes? Will it work?

src/Raven.Studio/typescript/common/pwaInstaller.ts
JudahGabriel: Boo.

src/Raven.Studio/typescript/viewmodels/database/settings/studioConfiguration.ts
JudahGabriel: Typescript did that for me automatically. I'll fix it.

src/Raven.Studio/wwwroot/index.html
JudahGabriel: Hmmm. I didn't do any formatting myself; Visual Studio must have done this.

src/Raven.Studio/wwwroot/manifest.json
JudahGabriel: I'm aware of this limitation, but it means we'd need to generated the manifest on the server side. I decided against adding that in this first iteration, but I can add it.

JudahGabriel: And yeah, you can install multiple Raven Studios.

src/Raven.Studio/typescript/common/pwaInstaller.ts
JudahGabriel: :-(

src/Raven.Studio/typescript/viewmodels/database/settings/studioConfiguration.ts
JudahGabriel: You want this moved to the server-wide Studio Configuration? OK.

src/Raven.Studio/wwwroot/manifest.json
JudahGabriel: Marcin, to do this, we need to generate the manifest on the server. To do that, we could either:

- Leave the manifest.json file in the solution, but when it's requested we read it, change the name, then return it.
- Delete manifest.json file and just generate it on the fly from the server.

Do you have a preference?

JudahGabriel: Actually, it looks like we can [generate the manifest dynamically on the client](https://medium.com/@alshakero/how-to-setup-your-web-app-manifest-dynamically-using-javascript-f7fbee899a61). So this is option 3.

JudahGabriel:
> 
> 
> @JudahGabriel Btw what about links with target blank, or fetching data from different cluster nodes? Will it work?

Yeah, the web manifest spec has now updated to standardize on this behavior. Accessing URLs outside of the app will launch in a separate, full browser chrome window, IIRC.

ml054:
@JudahGabriel let me play with this next week. I'm curious about limitations of this solution (if any).

ml054:
You don't register service worker so it doesn't work by default.

ml054:
I've managed to install and test this and my humble opinion is: The PWA standard is not yet mature to deploy. The problem is we have few places in which we open new tabs - and new tabs opens as standard chrome window so having an app is fancy but it breaks entire idea of PWA.

JudahGabriel:
Service worker: looks like when I reverted the formatting changes to index.html, I inadvertently removed the service worker registration. I've added it back in.

JudahGabriel:
> The PWA standard is not yet mature to deploy. The problem is we have few places in which we open new tabs - and new tabs opens as standard chrome window so having an app is fancy but it breaks entire idea of PWA.

Isn't that what a normal native app would do? Click an external link (or link to target="_blank"), it opens in a browser tab. That's the desired behavior, no?

ml054:
the problem is when "external link" is different node link. Which is and isn't external to us. Then we end up with native and non-native windows. This is behaviour we have in studio - once we navigate to non-relevant node we open this in new window.

JudahGabriel:
So the current behavior is, inside the Raven Studio app, you click a different node link, and it opens in a browser tab.

You don't like this behavior. What's the alternative? Open in another app instance? Or open inside the current app?

One thing we could do is navigate to other nodes inside the app. This could be done with web app manifest [scopes](https://developer.mozilla.org/en-US/docs/Web/Manifest#scope).

ml054:
Scopes are relative to root url, right? In cluster we change entire dns name. 

For native application I'd detect that and show one tab per cluster node. Optionally we could detect ctrl+click or going to external node and open new "soft" tab inside our application, but it is huge change.

JudahGabriel:
Scopes just specify what domains can load in-app. But, I don't think multiple scopes are permitted by the spec, so that might be a dead-end for us.

I still think the PWA standard is doing the right thing: click an external link, open in a new browser tab. This is a good behavior for even our nodes, IMO. I think this behavior is unlikely to change in the future versions of the PWA web standard.

OK Marcin, I gave it a shot. :-) If you don't think it's worth adding to Raven Studio, I understand.

PR #8904, opened 2019-03-11
RavenDB-13037 cluster stabilization

karmeli87:


src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -1059,55 +1065,92 @@ public void StateChanged(long index)
...
+            while (taken == false)
             {
-                DatabaseShutdown.ThrowIfCancellationRequested();
-
-                Debug.Assert(string.Equals(Name, record.DatabaseName, StringComparison.OrdinalIgnoreCase),
-                    $"{Name} != {record.DatabaseName}");
-
-                // index and LastDatabaseRecordIndex could have equal values when we transit from/to passive and want to update the tasks. 
-                if (LastDatabaseRecordIndex > index)
+                Monitor.TryEnter(_clusterLocker, TimeSpan.FromSeconds(5), ref taken);
```
ayende, line 1074: Why are you first trying to take the lock, then try to check even if you didn't get it?
Isn't it better to put the `CanSkip` first?

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -1059,55 +1065,92 @@ public void StateChanged(long index)
...
-                EtlLoader?.HandleDatabaseValueChanged(record);
+        private void NotifyFeaturesAboutValueChange(DatabaseRecord record, long index)
+        {
+            var taken = false;
+            while (taken == false)
+            {
+                Monitor.TryEnter(_clusterLocker, TimeSpan.FromSeconds(5), ref taken);
+                try
+                {
+                    if (CanSkip(record.DatabaseName, index))
```
ayende, line 1144: Same here

src/Raven.Server/Rachis/CandidateAmbassador.cs
```diff
@@ -109,12 +109,19 @@ private void Run()
                         Action disconnect;
                         try
                         {
-                            var connection = _engine.ConnectToPeer(_url, _tag, _engine.ClusterCertificate).Result;
+                            var connectTask = _engine.ConnectToPeer(_url, _tag, _engine.ClusterCertificate);
+                            while (connectTask.Wait(1000) == false)
```
ayende, line 113: Why are we doing it in this manner?
And can we use a cancellation token here, instead?

src/Raven.Server/Rachis/FollowerAmbassador.cs
```diff
@@ -177,7 +177,21 @@ private unsafe void Run()
                                 using (_engine.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
                                 {
                                     _engine.RemoveAndDispose(_leader, _connection);
-                                    var connection = _engine.ConnectToPeer(_url, _tag, _engine.ClusterCertificate).Result;
+                                    var connectTask = _engine.ConnectToPeer(_url, _tag, _engine.ClusterCertificate);
+
+                                    while (connectTask.Wait(1000) == false)
```
ayende, line 182: Same, better to use cancellation token

ayende (changes requested):

src/Raven.Server/Rachis/CandidateAmbassador.cs
karmeli87: the await is consisted from basically three parts:

1. fetch tcp info (can pass a cancellation token to the request executor)
2. tcp client tries to connect (haven't saw an overload with cancellation)
3. wrap ssl stream (no token either).

ayende: See: https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.wait?view=netframework-4.7.2#System_Threading_Tasks_Task_Wait_System_Int32_System_Threading_CancellationToken_

src/Raven.Server/Documents/DocumentDatabase.cs
karmeli87: We waited for 5 seconds for the lock and got it (or not).. it might be that we don't need to do anything because in that time a newer index already updated the features.

src/Raven.Server/Rachis/CandidateAmbassador.cs
karmeli87: and for (1), we cancel it after 15 seconds (the default value of `OperationTimeout`).

src/Raven.Server/Documents/DocumentDatabase.cs
ayende: Why not put the `CanSkip` before trying to get the lock, as well as after?

src/Raven.Server/Rachis/CandidateAmbassador.cs
karmeli87: Okay... I thought of passing a token to the `ConnectToPeer` method.
I can pass a token, but I will still need to validate the term (more relevant for the follower ambassador)

src/Raven.Server/Documents/DocumentDatabase.cs
karmeli87: Done

src/Raven.Server/Rachis/CandidateAmbassador.cs
karmeli87: We will deal with it in a different issue by limiting the concurrent number of ambassadors:
https://issues.hibernatingrhinos.com/issue/RavenDB-13067

ayende (approved):

PR #8914, opened 2019-03-12
RavenDB-12104 Recover from exception during accepting TCP connection

karmeli87:


src/Raven.Server/RavenServer.cs
```diff
@@ -1157,9 +1157,16 @@ private void ListenToNewTcpConnection(TcpListener listener)
                 }
                 catch (Exception e)
                 {
-                    if (_tcpLogger.IsInfoEnabled)
+                    if (_tcpLogger.IsOperationsEnabled)
```
ayende: Move the log statement to the `if`

ayende: Also, we need normal exception to be in info, and the retries in operations

ayende (changes requested):

src/Raven.Server/RavenServer.cs
karmeli87: So the first time should be logged as `info` and the rest as `operations`?

src/Raven.Server/RavenServer.cs
```diff
@@ -1143,26 +1143,56 @@ private void ListenToNewTcpConnection(TcpListener listener)
...
+                Logger tcpAuditLog;
+
+                while (true)
                 {
-                    if (_tcpLogger.IsInfoEnabled)
+                    try
                     {
-                        _tcpLogger.Info("Failed to accept new tcp connection", e);
+                        tcpClient = await listener.AcceptTcpClientAsync();
+                        tcpAuditLog = LoggingSource.AuditLog.IsInfoEnabled ? LoggingSource.AuditLog.GetLogger("TcpConnections", "Audit") : null;
```
ppekrol: do we need to create that each time?

src/Raven.Server/RavenServer.cs
```diff
@@ -1143,26 +1143,56 @@ private void ListenToNewTcpConnection(TcpListener listener)
...
+                            backoffSecondsDelay = 1;
+                            continue;
+                        }
+
+                        if (_tcpLogger.IsOperationsEnabled)
+                        {
+                            _tcpLogger.Operations($"Failed to accept new tcp connection again, will wait {backoffSecondsDelay} seconds before retrying", e);
+                        }
+
+                        await Task.Delay(TimeSpan.FromSeconds(backoffSecondsDelay), ServerStore.ServerShutdown);
```
ppekrol: this can throw OperationCanceledException and/or ObjectDisposedException probably?

ayende: We want to handle `OperationCanceledException ` as well, yes

src/Raven.Server/RavenServer.cs
karmeli87: We will create it only once, after accepting the tcp was successful.

src/Raven.Server/RavenServer.cs
```diff
@@ -1143,26 +1143,56 @@ private void ListenToNewTcpConnection(TcpListener listener)
...
+
+                        if (listener.Server.Connected)
+                        {
+                            backoffSecondsDelay = Math.Min(backoffSecondsDelay * 2, 60);
+                            continue;
+                        }
+
+                        if (_tcpLogger.IsOperationsEnabled)
+                        {
+                            _tcpLogger.Operations($"The socket on {listener.LocalEndpoint} is no longer connected.");
```
ayende: Let's include here the exception

ayende: This probably deserve an alert

karmeli87: done

src/Raven.Server/RavenServer.cs
karmeli87: Not sure why this is necessary... we are shutting down in that case, but done.

src/Raven.Server/RavenServer.cs
```diff
@@ -1332,6 +1315,80 @@ private void ListenToNewTcpConnection(TcpListener listener)
...
+
+                    if (_tcpLogger.IsOperationsEnabled)
+                    {
+                        _tcpLogger.Operations(msg, e);
+                    }
+
+                    var alert = AlertRaised.Create(
+                        null,
+                        msg,
+                        $"There are four threads listening for a new TCP connections, if any of them is down it is a severe error that can affect performance, when all of them are down it will cause TCP connectivity issues.{Environment.NewLine}" +
```
ayende: This is describing implementation details. Let's just explain what is going on.
Unable to accept connections from TCP and the listen socket was rudely disconnected or something like that

karmeli87: I though what should I write to avoid support calls :)

What do you suggest? should what you proposed be enough `Unable to accept connections from TCP and the listen socket was rudely disconnected` ?

ppekrol (approved):

PR #8920, opened 2019-03-12
Preparation for generalized allocators 

redknightlois:
I pondered PowerOf2 and NearestPowerOf2 as potential alternatives. I settled for the former, but the latter is fine too if anyone has sound objections to the one chosen.


ayende:
I don't have any issues with this change, but what is the point of this?
What does this gives us?

redknightlois:
Semantic consistency mainly. It was motivated because I made many mistakes over time just because of that expecting Next and not getting it (allocators code use power of 2 assumptions all over the place and made many difficult to diagnose error because of that).

ayende (approved):

PR #8921, opened 2019-03-12
Generalized Allocators

redknightlois:


ppekrol:
@redknightlois the RavenDB-12068 seems way wrong. Did you mean RavenDB-8951?

redknightlois:
How did I get 12068??? Yes, it's very wrong.

redknightlois:
Apparently, I picked up that number 6 months ago... dont understand why.

ppekrol:
This PR is for review purposes only. Not to be merged with 4.2. Target release is: 5.0

src/Sparrow.Server/Allocators.Arena.cs
```diff
@@ -0,0 +1,345 @@
...
+            // The reasons to consider current buffer size and also the used amount of memory is very important to figure out
+            // what is the optimal policy for the arena size on the next round.
+            int preferredSize = allocator._options.GrowthStrategy.GetPreferredSize(allocator._currentBuffer.Size, allocator.InUse);
+
+            if (allocator._olderBuffers != null)
+            {
+                // Free old buffers not being used anymore. A single arena ought to be enough for anybody, you know ;) 
+                foreach (var unusedBuffer in allocator._olderBuffers)
+                {
+                    _used += unusedBuffer.Size;
```
ayende, line 246: Why are you _increasing_ the `_used` amount here?

src/Sparrow.Server/Allocators.Arena.cs
```diff
@@ -0,0 +1,345 @@
...
+
+                    Pointer ptr = unusedBuffer;
+
+                    allocator.Allocated -= ptr.Size;
+                    allocator._nativeAllocator.Release(ref allocator._nativeAllocator, ref ptr);
```
ayende, line 251:
```suggestion
                    allocator._nativeAllocator.Release(ref ptr);
```

Isn't this better?

src/Sparrow.Server/Allocators.Arena.cs
```diff
@@ -0,0 +1,345 @@
...
+            if (allocator.InUse <= allocator._currentBuffer.Size || Allocator.LowMemoryFlag.IsRaised())
+            {
+                // Override with the current setup
+                preferredSize = allocator._currentBuffer.Size;
+            }
+
+            if (allocator._currentBuffer.IsValid)
+            {
+                allocator.Allocated -= _currentBuffer.Size;
+                allocator._nativeAllocator.Release(ref allocator._nativeAllocator, ref _currentBuffer);
```
ayende, line 266: On each `Reset` you'll return the memory? Isn't that going to be expensive with malloc / free on each usage?

src/Sparrow.Server/Allocators.Arena.cs
```diff
@@ -0,0 +1,345 @@
...
+            allocator._used = 0;
+
+            allocator.TotalAllocated = 0;
+            allocator.Allocated = 0;
+            allocator.InUse = 0;
+        }
+
+        public void Renew(ref ArenaAllocator<TOptions> allocator)
+        {
+            if (!allocator._currentBuffer.IsValid)
```
ayende, line 285: If you previously `Reset`, then this is never going to be true, isn't it?

src/Sparrow.Server/Allocators.Arena.cs
```diff
@@ -0,0 +1,345 @@
...
+            allocator._currentBuffer = new Pointer();
+            allocator._ptrStart = allocator._ptrCurrent = null;
+            allocator._used = 0;
+
+            allocator.TotalAllocated = 0;
+            allocator.Allocated = 0;
+            allocator.InUse = 0;
+        }
+
+        public void Renew(ref ArenaAllocator<TOptions> allocator)
```
ayende, line 283: Did you change the concept of reset / renew as they are currently used?
This seems to assume that reset hasn't been called.

src/Sparrow.Server/Allocators.Arena.cs
```diff
@@ -0,0 +1,345 @@
...
+        }
+
+        public void Dispose(ref ArenaAllocator<TOptions> allocator, bool disposing)
+        {            
+            if (allocator._olderBuffers != null)
+            {
+                // Free the remnant buffers on local storage from last cycle regrowth
+                foreach (var unusedBuffer in allocator._olderBuffers)
+                {
+                    _used += unusedBuffer.Size;
```
ayende, line 329: Why _increasing_ the amount here?

src/Sparrow.Server/Allocators.Arena.cs
```diff
@@ -0,0 +1,345 @@
...
+                allocator._currentBuffer = allocator._nativeAllocator.Allocate(ref allocator._nativeAllocator, (int)allocator._preferredSize);
+                allocator.Allocated += allocator._currentBuffer.Size;
+            }
+
+            if (allocator._olderBuffers != null)
+            {
+                // Free old buffers not being used anymore. A single arena ought to be enough for anybody, you know ;) 
+                foreach (var unusedBuffer in allocator._olderBuffers)
+                {
+                    _used += unusedBuffer.Size;
```
ayende, line 296: Why _increase_ the amount here?

src/Sparrow.Server/Allocators.Fragment.cs
```diff
@@ -0,0 +1,272 @@
...
+                Pointer ptr = new Pointer(allocator._ptrCurrent, size);
+                allocator._ptrCurrent += size;
+                allocator._used += size;
+                allocator.TotalAllocated += size;
+                allocator.InUse += size;
+
+                return ptr;
+            }
+
+            // The allocation is too big to fit, the memory will be used afterwards as a segment when released.
```
ayende, line 134: For allocations that are larger than the block size (which are likely rare), shouldn't we free them immediately?

src/Sparrow.Server/Allocators.Fragment.cs
```diff
@@ -0,0 +1,272 @@
...
+            if (bytesLeft > allocator._options.ReuseBlocksBiggerThan)
+            {
+                // The allocation is bigger than ReuseBlocksBiggerThan and therefore we can leave this one available to be used later.
+                allocator._internalPartialReusedSegments.Add(bytesLeft, new Pointer(allocator._ptrCurrent, bytesLeft));
+            }
+
+            // TODO: Time the impact of using a heap structure instead of a SortedList.
+            // Get the biggest segment on the Heap and if size < the biggest one, we avoid the allocation.
+            Pointer segment;
+            if (allocator._internalPartialReusedSegments.Count > 0 && allocator._internalPartialReusedSegments.Values[0].Size >= size)
```
ayende, line 174: This is pretty much by definition never going to work. 
You just put the current segment there, because it was too small, and now you are checking the smallest one for this.
That is going to be the size of the current segment or smaller, so not big enough

ayende: Better to do binary search for the first value that is equal or greater to the size, no?

src/Sparrow.Server/Allocators.Fragment.cs
```diff
@@ -0,0 +1,272 @@
...
+
+            allocator.TotalAllocated = 0;
+            allocator.Allocated = 0;
+            allocator.InUse = 0;
+        }
+
+        public void OnAllocate(ref FragmentAllocator<TOptions> allocator, Pointer ptr) {}
+        public void OnRelease(ref FragmentAllocator<TOptions> allocator, Pointer ptr) {}
+
+        public void Dispose(ref FragmentAllocator<TOptions> allocator, bool disposing)
```
ayende, line 247: `Reset` is identical to this, why?

src/Sparrow.Server/Allocators.Native.cs
```diff
@@ -0,0 +1,209 @@
...
+            return new Pointer(memory, size);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void Release(ref NativeAllocator<TOptions> allocator, ref Pointer ptr)
+        {
+            Interlocked.Add(ref allocator._allocated, -ptr.Size);
+
+            // PERF: Given that for the normal use case the INativeOptions we will use returns constants the
+            //       JIT will be able to fold all this if sequence into a branchless single call.
```
ayende, line 141: Do we want to zero the memory on release as well (although might be only valid for secured)?

src/Sparrow.Server/Allocators.Native.cs
```diff
@@ -0,0 +1,209 @@
...
+        [Conditional("DEBUG")]
+        [Conditional("VALIDATE")]
+        private void CheckForLeaks(ref NativeAllocator<TOptions> allocator)
+        {
+            if (allocator.Allocated != 0)
+            {
+#if VALIDATE || DEBUG
+
+                StringBuilder stackFrameAppender = new StringBuilder();
+                using (var reader = new StringReader(allocator._initializeStackTrace.ToString()))
```
ayende: Why do you need this? Isn't this skipping the first three lines? if so, `Split().Skip(3)` is more readable and this is debug code

src/Sparrow.Server/Allocators.PointerTypes.cs
```diff
@@ -0,0 +1,527 @@
...
+    public interface IPointerType { }
+    public interface IPointerType<T> : IPointerType where T : struct { }
+
+    [StructLayout(LayoutKind.Sequential)]
+#if !VALIDATE
+    // We do this to have the ability to perform extensive validation for pointers.
+    readonly
+#endif
+    public unsafe struct Pointer : IPointerType, Meta.IDescribe
+        {
```
ayende: Need reformating

src/Sparrow.Server/Allocators.PointerTypes.cs
```diff
@@ -0,0 +1,527 @@
...
+        public Span<T> AsSpan(int start, int length)
+        {
+            if (start + length > Size)
+                throw new ArgumentException($"{nameof(length)} cannot be bigger than block size.");            
+
+            return new Span<T>((byte*)Address + start * Unsafe.SizeOf<T>(), length);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public ReadOnlySpan<byte> AsReadOnlySpan()
```
ayende: Pretty sure you meant `ReadOnlySpan<T>` here, no?

src/Sparrow.Server/Allocators.PointerTypes.cs
```diff
@@ -0,0 +1,527 @@
...
+            return new ReadOnlySpan<T>((byte*)Address + start * Unsafe.SizeOf<T>(), length);
+        }
+
+        public int Size
+        {
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            get
+            {
+                EnsureIsNotBadPointerAccess();
+                // We move to the actual header which is behind.
```
ayende: Don't understand this comment

src/Sparrow.Server/Allocators.PointerTypes.cs
```diff
@@ -0,0 +1,527 @@
...
+            }
+        }
+
+        public int SizeAsBytes
+        {
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            get
+            {
+                EnsureIsNotBadPointerAccess();
+                // We move to the actual header which is behind.
```
ayende: I don't understand this comment

src/Sparrow.Server/Allocators.PointerTypes.cs
```diff
@@ -0,0 +1,527 @@
...
+
+#if VALIDATE
+        public uint Generation;        
+#endif
+
+        [MethodImpl(MethodImplOptions.NoOptimization)]
+        static BlockPointer()
+        {
+            // If T is not a blittable type, fail on type loading unrecoverably.
+            // We need to ensure the JIT wont optimize away the call to Marshal.SizeOf<T> for correctness either now or in the future.
```
ayende: Can probably do that by setting a static member value here, no?

src/Sparrow.Server/Allocators.PointerTypes.cs
```diff
@@ -0,0 +1,527 @@
...
+        public Span<T> AsSpan(int start, int length)
+        {
+            if (start + length > Size)
+                throw new ArgumentException($"{nameof(length)} cannot be bigger than block size.");            
+            
+            return new Span<T>((byte*)Address + start * Unsafe.SizeOf<T>(), length);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public ReadOnlySpan<byte> AsReadOnlySpan()
```
ayende: `ReadOnlySpan<T> `, no?

src/Sparrow.Server/Allocators.PointerTypes.cs
```diff
@@ -0,0 +1,527 @@
...
+            return new ReadOnlySpan<T>((byte*)Address + start * Unsafe.SizeOf<T>(), length);
+        }
+
+        public int Size
+        {
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            get
+            {
+                EnsureIsNotBadPointerAccess();
+                // We move to the actual header which is behind.
```
ayende: comment?

src/Sparrow.Server/Allocators.PointerTypes.cs
```diff
@@ -0,0 +1,527 @@
...
+            }
+        }
+
+        public int SizeAsBytes
+        {
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            get
+            {
+                EnsureIsNotBadPointerAccess();
+                // We move to the actual header which is behind.
```
ayende: comment?

src/Sparrow.Server/Allocators.PointerTypes.cs
```diff
@@ -0,0 +1,527 @@
...
+        }
+
+        public override string ToString()
+        {
+            return Describe();
+        }
+
+        public static implicit operator BlockPointer(BlockPointer<T> ptr)
+        {
+            return new BlockPointer(ptr.Address, ptr.BlockSize, ptr.Size);
```
ayende:
```suggestion
            return new BlockPointer(ptr.Address, ptr.BlockSize, ptr.SizeAsBytes);
```

ayende: And I'm not sure about the `BlockSize` here either, is that in bytes, in elements?

src/Sparrow.Server/Allocators.Pool.FixedSize.cs
```diff
@@ -0,0 +1,270 @@
...
+            if (ptr.Size != allocator._options.MaxBlockSize)
+                ThrowMemoryDoesNotBelongToAllocator(ptr);
+
+            if (allocator.InUse > allocator._options.MaxPoolSizeInBytes || Allocator.LowMemoryFlag.IsRaised())
+                goto UnlikelyRelease;
+
+            int originalSize = ptr.Size;
+
+            var section = allocator._freed;
+            if (section.IsValid)
```
ayende, line 138: Can't we spare that and always copy the `_freed` unconditionally?

src/Sparrow.Server/Allocators.Pool.FixedSize.cs
```diff
@@ -0,0 +1,270 @@
...
+            // This should be an unlikely case, when you are running out of memory or over allocated,
+            // all performance guarantees go down the drain. So we dont care if we hit expensive calls
+            // that allows us to release some steam at the cost of hitting cold code. 
+            // https://github.com/dotnet/coreclr/issues/6024
+
+            allocator.InUse -= ptr.Size;
+            allocator.Allocated -= ptr.Size;
+
+            Pointer nakedPtr = new Pointer(ptr.Address, ptr.Size);
+            allocator._internalAllocator.Release(ref nakedPtr);
```
ayende, line 166: Why not send the `ptr` directly?
Also, didn't reset `ptr`

src/Sparrow.Server/Allocators.Pool.FixedSize.cs
```diff
@@ -0,0 +1,270 @@
...
+                ReleaseMemoryPool(ref allocator);
+                allocator.Allocated = 0;
+            }                
+
+            allocator._internalAllocator.Renew();
+            allocator.InUse = 0;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void Reset(ref FixedSizePoolAllocator<TOptions> allocator)
```
ayende, line 193: I don't understand the usage of reset & renew in this PR

src/Sparrow.Server/Allocators.Pool.FixedSizeThreadAffine.cs
```diff
@@ -0,0 +1,371 @@
...
+            /// By default whenever we create an allocator we are going to dispose it too when the time comes.
+            /// </summary>
+            /// <param name="allocator">the allocator to dispose.</param>
+            public void ReleaseAllocator(IAllocatorComposer<Pointer> allocator, bool disposing)
+            {
+                allocator?.Dispose(disposing);
+            }
+        }
+
+        public struct Static : IFixedSizeThreadAffinePoolOptions, INativeOptions
```
ayende, line 58: What is the difference from `Default `?

src/Sparrow.Server/Allocators.Pool.FixedSizeThreadAffine.cs
```diff
@@ -0,0 +1,371 @@
...
+            if (allocator._options.ItemsPerLane < 1)
+                throw new ArgumentOutOfRangeException($"{nameof(allocator._options.ItemsPerLane)} cannot be smaller than 1.");
+            if (allocator._options.ItemsPerLane > 4)
+                throw new ArgumentOutOfRangeException($"{nameof(allocator._options.ItemsPerLane)} cannot be bigger than 4.");            
+        }
+
+        public Pointer Allocate(ref FixedSizeThreadAffinePoolAllocator<TOptions> allocator, int size)
+        {
+            if (allocator._options.AcceptOnlyBlocks && size != allocator._options.BlockSize)
+                throw new InvalidOperationException(
```
ayende, line 191: Move to dedicated method?

src/Sparrow.Server/Allocators.Pool.FixedSizeThreadAffine.cs
```diff
@@ -0,0 +1,371 @@
...
+
+            if (allocator._options.ItemsPerLane < 1)
+                throw new ArgumentOutOfRangeException($"{nameof(allocator._options.ItemsPerLane)} cannot be smaller than 1.");
+            if (allocator._options.ItemsPerLane > 4)
+                throw new ArgumentOutOfRangeException($"{nameof(allocator._options.ItemsPerLane)} cannot be bigger than 4.");            
+        }
+
+        public Pointer Allocate(ref FixedSizeThreadAffinePoolAllocator<TOptions> allocator, int size)
+        {
+            if (allocator._options.AcceptOnlyBlocks && size != allocator._options.BlockSize)
```
ayende, line 190: How can we ever support something that isn't fixed size here?

src/Sparrow.Server/Allocators.Pool.NonBlocking.cs
```diff
@@ -0,0 +1,308 @@
...
+    /// The idea is that we will store in concurrent stacks the blocks of memory and use non-blocking APIs to release and allocate from the pool.
+    /// </summary>
+    /// <typeparam name="TOptions">The options to use for the allocator.</typeparam>
+    /// <remarks>The Options object must be properly implemented to achieve performance improvements. (use constants as much as you can on configuration)</remarks>
+    public unsafe struct NonBlockingPoolAllocator<TOptions> : IAllocator<NonBlockingPoolAllocator<TOptions>, BlockPointer>, ILowMemoryHandler<NonBlockingPoolAllocator<TOptions>>, IRenewable<NonBlockingPoolAllocator<TOptions>>
+        where TOptions : struct, IPoolAllocatorOptions
+    {
+        private TOptions _options;        
+
+        // PERF: This should be devirtualized.
```
ayende, line 23: How, this is a interface field, so have to be boxed and have to go through the method dispatch

src/Sparrow.Server/Allocators.Pool.cs
```diff
@@ -0,0 +1,312 @@
...
+
+    public static class PoolAllocator
+    {
+        public struct Default : IPoolAllocatorOptions, INativeOptions
+        {
+            public bool UseSecureMemory => false;
+            public bool ElectricFenceEnabled => false;
+            public bool Zeroed => false;
+
+            public int MaxBlockSize => 64 * Constants.Size.Megabyte;
```
ayende, line 24: All these constants need to take into account 32 bits as well

src/Sparrow.Server/Allocators.Pool.cs
```diff
@@ -0,0 +1,312 @@
...
+            /// By default whenever we create an allocator we are going to dispose it too when the time comes.
+            /// </summary>
+            /// <param name="allocator">the allocator to dispose.</param>
+            public void ReleaseAllocator(IAllocatorComposer<Pointer> allocator, bool disposing)
+            {
+                allocator?.Dispose(disposing);
+            }
+        }
+
+        public struct Static : IPoolAllocatorOptions, INativeOptions
```
ayende, line 46: This is identical do `Default`?

src/Sparrow.Server/Allocators.Pool.cs
```diff
@@ -0,0 +1,312 @@
...
+        UnlikelyRelease:
+            // This should be an unlikely case, when you are running out of memory or over allocated,
+            // all performance guarantees go down the drain. So we dont care if we hit expensive calls
+            // that allows us to release some steam at the cost of hitting cold code. 
+            // https://github.com/dotnet/coreclr/issues/6024
+
+            allocator.InUse -= ptr.Size;
+            allocator.Allocated -= ptr.BlockSize;
+
+            Pointer nakedPtr = new Pointer(ptr.Address, ptr.BlockSize);
```
ayende: Why do the copy?

src/Sparrow.Server/Allocators.cs
```diff
@@ -0,0 +1,681 @@
...
+        {
+
+            if (_allocator is ILifecycleHandler<TAllocator> a)
+                a.BeforeInitialize(ref _allocator);
+
+            _allocator.Initialize(ref _allocator);
+            _allocator.Configure(ref _allocator, ref options);
+
+#if VALIDATE
+            this._generation =  (uint)new Random().Next() & 0xFFF00000;
```
ayende, line 147: Huh?

ayende (changes requested):

src/Sparrow.Server/Allocators.Arena.cs
redknightlois: Probably, but I am not entirely sure of all the potential semantic implications with the interplay between structs inside structs... Until I am not entirely sure about the potential implications I prefer to err on the side of caution.

src/Sparrow.Server/Allocators.Arena.cs
redknightlois: The idea is that you don't use the `Arena` without a caching backend (unless you really need it like that). Always remember that the design is composable.

src/Sparrow.Server/Allocators.Arena.cs
redknightlois: Yes, `Reset` and `Renew` as they are currently used overlap in their responsibility. `Reset` and `Renew` now have clear implications.

src/Sparrow.Server/Allocators.Pool.cs
redknightlois: Until we have everything integrated those numbers are sensible defaults. After that we will need to looks at all different variants of allocators we end up using (I will create custom configurations for each) until we can then go there and find the actual 'default' case.

src/Sparrow.Server/Allocators.Pool.NonBlocking.cs
redknightlois: Nope, this particular case will actually hit a special JIT code-path as long as you never return a TOptions based naked interface.

src/Sparrow.Server/Allocators.Pool.cs
redknightlois: No, the static one overrides the `ReleaseAllocator` disposing mechanism because it is designed to deal with the `readonly static` lifecycle instead of a normal method stored allocator.

src/Sparrow.Server/Allocators.Pool.cs
redknightlois: Because when running in validation mode, that pointer may actually be `tracked` by a different allocator. For the purpose of the `AllocatorPool` the pointer is `new` no allocator tracking is needed.

src/Sparrow.Server/Allocators.cs
redknightlois: Each `Reset` and `Renew` will increase the generation number. But in order to be able to track pointer crossovers between allocators, the first 12 bits are the allocator id and the rest are the actual generation. Both conditions are checked. Will add a comment there.

src/Sparrow.Server/Allocators.Arena.cs
redknightlois: Actually, that is a no-op bug translated because of copy-pasta from Reset and Renew where it is added in the loop and then setup to 0 at the end of the method. Good catch. Removing from the 3 sites.

src/Sparrow.Server/Allocators.Arena.cs
ayende: I can't see how this can be anything else.

src/Sparrow.Server/Allocators.Arena.cs
ayende: Please state clearly what they are, because I currently can't understand when each is use, it's semantics meaning, etc.

src/Sparrow.Server/Allocators.Fragment.cs
redknightlois: It will actually reuse them until we hit a `Reset` boundary so they are going to be either short-lived or amortized on long-running processes.

src/Sparrow.Server/Allocators.Arena.cs
redknightlois: That is actually a no-op bug in the 3 places it happens. On `Reset` and `Renew` I add that and then store `0`. On `Dispose` well it is not going to be used anyways. Good catch.

src/Sparrow.Server/Allocators.Pool.NonBlocking.cs
ayende: This is `Allocator<TAllocator>`, which is a class, unless I missed something?
And the JIT don't know what the actual type will be here.

src/Sparrow.Server/Allocators.Native.cs
redknightlois: We can configure that from the caller site. For secure that is part of the normal operation of secure memory, so while implicitly you will setup it like false at the allocator level.

src/Sparrow.Server/Allocators.Native.cs
redknightlois: Those 3 lines are the actual leak checker code, skipping them is more readable. From the code point of view, yeah `Split().Skip(3)` would do the trick.

src/Sparrow.Server/Allocators.PointerTypes.cs
redknightlois: Yeah, still every time you change something it gets screwed again... VS support for #define statements is quite crude for C#.

src/Sparrow.Server/Allocators.PointerTypes.cs
redknightlois: Yes it is.

src/Sparrow.Server/Allocators.PointerTypes.cs
redknightlois: Comment no relevant anymore. Removing

src/Sparrow.Server/Allocators.PointerTypes.cs
redknightlois: No need that is why we use the `[MethodImpl(MethodImplOptions.NoOptimization)]` attribute. Modifying the comment to account for the misunderstanding.

src/Sparrow.Server/Allocators.PointerTypes.cs
redknightlois: Yes

src/Sparrow.Server/Allocators.PointerTypes.cs
redknightlois: Actually, it looks like I've never used/tested those conversions... most of them are wrong.

src/Sparrow.Server/Allocators.Pool.FixedSize.cs
redknightlois: Same rationale as before.

src/Sparrow.Server/Allocators.Pool.FixedSizeThreadAffine.cs
redknightlois: No need now. that is Pre-2.0 JIT behavior. Until we are sure that we are only using 2.0 JIT we need to keep it, but this is going to be executed on hopefully 3.0 JIT so no need to do that anymore.

src/Sparrow.Server/Allocators.Pool.FixedSizeThreadAffine.cs
redknightlois: You can, you shouldn't because it is going to be inefficient. But still I left the option available because there were a few use cases that were like that.

src/Sparrow.Server/Allocators.Arena.cs
redknightlois: Exactly

src/Sparrow.Server/Allocators.Pool.FixedSize.cs
redknightlois: I will add some information about it. But the concept is the following. 
Every allocator, when asked to reset, will get rid of as much memory as it possibly can. Allocators when asked to `Reset` wont try to speculatively hoard on memory, they will release as much as they can to the parent allocator. On the other hand, calling Renew may be a 'no-op' in the sense that the allocator is free to speculatively hoard the memory for its own use. For example in the case of the `PoolAllocator` when asked to Reset it will `clean-up` its internal storage and release the pools to the parent allocator or release the memory if it has ownership. While when asked to `Renew` it will only release the internal store if there is a Low Memory Flag activated and keep the memory in hot standby to use on the next cycle. 

That is the reason why we track not only the memory that the allocator has given to the callers (which could hint on internal fragmentation) but also the total allocated memory on stores. If those numbers are big, it would make sense to issue a `Reset`, or for example after an indexing call, etc... While for most cases you will end up only calling `Renew`.

src/Sparrow.Server/Allocators.Fragment.cs
redknightlois: So this goes like this... we insert in biggest to smallest in order (unless I made a mistake on `DuplicateKeyComparer`)... If the biggest chunk is not big enough for the allocation, we just go an get a brand new segment of size BlockSize... remember that if the size is actually bigger than block size, we have a dedicated path to deal with it on `Allocate`

Weirdly enough that is actually the proven way to limit internal fragmentation. The most efficient strategy (while it may sound strange) is to take the biggest chunk instead of the smallest. Someone at the math department of one of the top Israel universities was able to figure that one out (if I am not mixing the location of 2 pretty related papers on fragmentation).

src/Sparrow.Server/Allocators.Fragment.cs
redknightlois: For all uses and purposes, we just can call reset. Check my earlier comments on the difference between `Reset` and `Renew`

redknightlois: It was missing a call to the internal allocator `Dispose` method though... Fixed.

redknightlois:
@ayende By now we can say with evidence that allocating memory is a pain in the ass... :D

src/Sparrow.Server/Allocators.Arena.cs
redknightlois: I lean with your viewpoint, but we are playing razor thin on what the language/JIT allows you to do here. This is internal API detail that can be changed on an afternoon, so I wouldn't do it until we have enough code to ensure that is the case... Last thing I want is to have to deal with weird feature interplay while stabilizing allocators at the same time.

src/Sparrow.Server/Allocators.Pool.NonBlocking.cs
redknightlois: The trick is that it knows when it gets resolved, which aint by the time the code is emitted for `Sparrow.Server` but when it is used like in https://github.com/ravendb/ravendb/pull/8921/files#diff-ae8fcae82d91040bb8935f9585e85fbfR87

src/Sparrow.Server/Allocators.Pool.FixedSize.cs
ayende: In that case, let's give these method different names, because they are conflicting with the existing terminology in the codebase.

Another issue is who call them, and when? 
When is Reset and when is Renew called?

src/Sparrow.Server/Allocators.Fragment.cs
ayende: You are correct, but let's rename `DuplicateKeyComparer` to `InverseDuplicateKeyComparer` to make it clearer what is going on.

src/Sparrow.Server/Allocators.Pool.FixedSize.cs
redknightlois: Makes a lot of sense to change the same to avoid mistakes. Any suggestion there?
Those are called by the owner, which in our case and on our current design is `JsonOperationContext` and the indexing threads... I am not entirely sure we should keep that that way, even though it makes sense to have a single `OperationContext` manager which will decide based on a sensible heuristic if to call one or the other.

src/Sparrow.Server/Allocators.Fragment.cs
redknightlois: Done

src/Sparrow.Server/Allocators.Pool.FixedSize.cs
ayende: `Renew` -> `ResetState` and `Reset` -> `ReleaseResources` ?

And who calls what is critical to the design.
We currently checkout a context, use it and return it.
Who does what in this case?

redknightlois: Not fond of those names because `Reset` and `ResetState` can become false friends, but not as bad as `Reset` and `Renew` semantically speaking.

redknightlois: In the process of returning it, the general manager should decide before checkout of the context what to do with the allocator... a lease on the allocators themselves is not a bad idea either.

ayende: That is too generic. Consider the scenarios we have.
For indexing, we know if we'll have a new indexing round or if we are going to rest, so that make sense.
But for requests, we have no such information, how do we deal with it then?

redknightlois: For that how old or how much memory it has allocated makes sense as a possible heuristic.

ppekrol:
Let's reopen this or open a new one when needed.

PR #8926, opened 2019-03-13
Raven db 12860

talweiss1982:


src/Raven.Client/ServerWide/Commands/TcpConnectionInfo.cs
```diff
@@ -7,15 +7,27 @@ public class TcpConnectionInfo
         public int Port;
         public string Url;
         public string Certificate;
-
+        public string[] ServerUrls;
```
ppekrol: can we rename that to `Urls` the single Url property is called `Url`, do not see a good reason to not call this property `Urls`

ppekrol (approved):

src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
```diff
@@ -156,17 +157,49 @@ private Task Run(CancellationToken ct)
 
         private async Task<Stream> ConnectToServer(CancellationToken token)
         {
-            var command = new GetTcpInfoCommand("Subscription/" + _dbName, _dbName);
+            async Task<TcpConnectionInfo> LegacyTryGetTcpInfo(RequestExecutor requestExecutor1, JsonOperationContext context, ServerNode node = null)
```
ayende: `requestExecutor1` ?

ayende: And can we move this to a separate method?

src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
```diff
@@ -156,17 +157,49 @@ private Task Run(CancellationToken ct)
...
+            {
+                var tcpCommand = new GetTcpInfoCommand("Subscription/" + _dbName, _dbName);
+                try
+                {
+                    if(node != null)
+                    {
+                        await requestExecutor1.ExecuteAsync(node, null, context, tcpCommand, shouldRetry: false, sessionInfo: null, token: token)
```
ayende: why `shouldRetry`  is `false`?

src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
```diff
@@ -156,17 +157,49 @@ private Task Run(CancellationToken ct)
...
                 if (_redirectNode != null)
                 {
                     try
                     {
-                        await requestExecutor.ExecuteAsync(_redirectNode, null, context, command, shouldRetry: false, sessionInfo: null, token: token).ConfigureAwait(false);
+                        await requestExecutor.ExecuteAsync(_redirectNode, null, context, command, shouldRetry: false, sessionInfo: null, token: token)
+                            .ConfigureAwait(false);
+                        tcpInfo = command.Result;
+                    }
+                    catch (ClientVersionMismatchException)
```
ayende, line 175: what would throw this error?

src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
talweiss1982: resharper... sure

src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
talweiss1982: I don't know this was the original logic, I think this is the case where you've already been redirected

src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
talweiss1982: the router when it doesn't find the path for the endpoint, I have tested this by modifying the remote task URL to something that doesn't exist and it propagate nicely to the client

ppekrol:
https://github.com/ravendb/ravendb/pull/8948

PR #8933, opened 2019-03-13
RavenDB-13093 Node not being deleted and hangs on pending for deletion

karmeli87:
- In case the observer replace a promotable node with another promotable node, the redundant node it will not be deleted from the topology after catching up.
- Deleting a node create a raft command for removing it from the database group, in case the command failed we need to notify about it.

### DIFF
--- a/src/Raven.Server/Documents/DatabasesLandlord.cs
+++ b/src/Raven.Server/Documents/DatabasesLandlord.cs
@@ -197,20 +197,19 @@ namespace Raven.Server.Documents
         public bool ShouldDeleteDatabase(string dbName, DatabaseRecord record)
         {
             var deletionInProgress = DeletionInProgressStatus.No;
-            var directDelete = record.DeletionInProgress != null &&
-                               record.DeletionInProgress.TryGetValue(_serverStore.NodeTag, out deletionInProgress) &&
+            var directDelete = record.DeletionInProgress?.TryGetValue(_serverStore.NodeTag, out deletionInProgress) == true &&
                                deletionInProgress != DeletionInProgressStatus.No;
 
-            if (directDelete &&
-                record.Topology.Count == record.Topology.ReplicationFactor)
-            // If the deletion was issued form the cluster observer to maintain the replication factor we need to make sure
-            // the all the documents were replicated from this node, therefor the deletion will be called from the replication code.
-            {
-                DeleteDatabase(dbName, deletionInProgress, record);
-                return true;
-            }
+            if (directDelete == false)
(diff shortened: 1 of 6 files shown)


src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -287,10 +286,23 @@ private void NotifyLeaderAboutRemoval(string dbName)
...
                         {
-                            _logger.Info($"Failed to notify leader about removal of node {_serverStore.NodeTag} from database {dbName}", t.Exception);
+                            _logger.Info(message, t.Exception);
                         }
+
+                        var alert = AlertRaised.Create(
+                            null,
+                            message,
+                            $"The database was deleted, but we probably failed to notify the cluster about it.{Environment.NewLine} " +
+                            $"If the node is still shown as pending for deletion, it will require the admin manually to remove the node.",
```
ayende: How are they going to be able to do that? 
Can we have the cluster observer re-validate this occasionally? Not notifying the cluster observer is actually pretty common in some failure scenarios, so we want a retry on that, and better if the cluster observer does this explicitly.

src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
```diff
@@ -911,21 +909,24 @@ private void RemoveOtherNodesIfNeeded(DatabaseObservationState state, ref List<D
...
+                ErrorOnDatabaseDoesNotExists = false,
+                DatabaseName = dbName,
+                FromNodes = nodesToDelete.ToArray(),
+                HardDelete = _hardDeleteOnReplacement,
+                UpdateReplicationFactor = false,
+            };
+
+            if (deletions == null)
+                deletions = new List<DeleteDatabaseCommand>();
+            deletions.Add(deletionCmd);
```
ayende, line 931: And here is the code that does that. Does this run only once?

karmeli87: It is too frequent to be called in the observer and here we can't tell between `failed to notify` and `deletion in progress`.

On the relevant node we will retry to send the command again after 15 seconds.

ayende: "too frequent" ? I'm not sure that I follow. Isn't that twice a second?

ayende: Okay, if there is a retry, that is fine

PR #8941, opened 2019-03-14
RavenDB-13029 GetCollections throws NRE

talweiss1982:
* Note while this commit seems to make the problem go away there might be a more complex issue here that is why i added a debug assert to make sure this surface in tests.


talweiss1982:
i'm seeing debug assertion failing after this change so don't merge this yet

talweiss1982:
debug assert fails

ayende:
You cannot change the order of operations here. It is important for a lot of factors. 

What is the cause of the NRE in the first place?

talweiss1982:
@ayende  note that in the async commit the order is replaced:
https://github.com/ravendb/ravendb/blob/v4.1/src/Voron/Impl/LowLevelTransaction.cs#L944
see my comment on the issue for further details

ayende:
I think that we can get away with it there because we are still holding the tx lock from the _current_ transaction in the `EndAsyncCommit`. 
Can you also try reversing it there and see if that works?

PR #8948, opened 2019-03-16
RavenDB-12860 we need to retry the connections in SubscriptionWorker in case of a node that is down

ppekrol:


ayende (approved):

ppekrol:
After reviewing the code, it looks like the proper value there is `false`. We want to try to connect only to the given node (`chosenNode`). If it fails, the code determines new node elsewhere (in `ShouldTryToReconnect`)

PR #8950, opened 2019-03-16
RavenDB-12945 removed GlobalHttpClientTimeout

ppekrol:


src/Raven.Client/Http/IdleTimeoutHttpContent.cs
```diff
@@ -0,0 +1,73 @@
...
+            var sw = Stopwatch.StartNew();
+
+            var innerStream = await _content.ReadAsStreamAsync().ConfigureAwait(false);
+            while (true)
+            {
+                var read = innerStream.Read(_buffer, 0, BufferSize);
+                if (read == 0)
+                    break;
+
+                await stream.WriteAsync(_buffer, 0, read).ConfigureAwait(false);
```
ayende: Don't we need to pass a cancellation token here with the configured timeout.

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -639,220 +641,226 @@ protected virtual bool TryLoadFromCache(JsonOperationContext context)
...
+                            else
                             {
-                                if (cts.IsCancellationRequested && token.IsCancellationRequested == false) // only when we timed out
-                                {
-                                    var timeoutException = new TimeoutException($"The request for {request.RequestUri} failed with timeout after {timeout}", e);
-                                    if (shouldRetry == false)
-                                        throw timeoutException;
-
-                                    sp.Stop();
+                                cts.CancelAfter(DefaultHttpClientTimeout);
```
ayende, line 761: What does this gives us that the usual timeout doesn't?

Also, note that this doesn't handle timeouts on read, which is the root cause of our problem

ayende (changes requested):

PR #8961, opened 2019-03-18
RavenDB-13029 fixing an issue where a read transaction invoking GetCo…

talweiss1982:
…llections could throw NRE

### DIFF
--- a/src/Raven.Server/Documents/DocumentsStorage.cs
+++ b/src/Raven.Server/Documents/DocumentsStorage.cs
@@ -1570,7 +1570,9 @@ namespace Raven.Server.Documents
             foreach (var kvp in _collectionsCache)
             {
                 var collectionTable = context.Transaction.InnerTransaction.OpenTable(DocsSchema, kvp.Value.GetTableName(CollectionTableType.Documents));
-
+                //This is the case where a read transaction reading a collection cached by a later write transaction we can safly ignore it.
+                if (collectionTable == null && context.Transaction.InnerTransaction.IsWriteTransaction == false)
+                    continue;
                 yield return new CollectionStats
                 {
                     Name = kvp.Key,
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1570,7 +1570,9 @@ public IEnumerable<CollectionStats> GetCollections(DocumentsOperationContext con
             foreach (var kvp in _collectionsCache)
             {
                 var collectionTable = context.Transaction.InnerTransaction.OpenTable(DocsSchema, kvp.Value.GetTableName(CollectionTableType.Documents));
-
+                //This is the case where a read transaction reading a collection cached by a later write transaction we can safly ignore it.
+                if (collectionTable == null && context.Transaction.InnerTransaction.IsWriteTransaction == false)
+                    continue;
```
ayende: If this happens in write tx, let's write a meaningful exception

src/Voron/Impl/LowLevelTransaction.cs
```diff
@@ -940,8 +940,8 @@ public void EndAsyncCommit()
             if (AsyncCommit.Result)
                 Environment.LastWorkTime = DateTime.UtcNow;
 
-            CommitStage3_DisposeTransactionResources();
             BeforeCommitFinalization?.Invoke(this);
+            CommitStage3_DisposeTransactionResources();
```
ppekrol, line 944: This shouldn't be moved, right?

talweiss1982: It should to be aligned with regular commit. When moving this for commit it broke things pretty badly and async commit have the order backwards. This leads me to believe there is a very rare case where async commit will fail horribly.

ayende: It isn't _actually_ a problem, because there is a lock on the _next_ transaction, but that is only true right now and is a bug waiting to happen. 
We're better of _with_ this change.

PR #8965, opened 2019-03-18
RavenDB-13124: Correctly restore an updated Compare Exchange value.

garayx:


src/Raven.Server/ServerWide/Commands/CompareExchangeCommands.cs
```diff
@@ -225,10 +224,6 @@ private unsafe void WriteCompareExchangeTombstone(TransactionOperationContext co
         public override DynamicJsonValue ToJson(JsonOperationContext context)
```
ppekrol: Don't need to override the method now

garayx: done

ppekrol (approved):

PR #8970, opened 2019-03-19
RavenDB-12520 _ensure_path_exists implemented for posix

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-12520

### DIFF
--- a/src/librvnpal/src/posix/fileutils.c
+++ b/src/librvnpal/src/posix/fileutils.c
@@ -95,8 +95,8 @@ _ensure_path_exists(const char *path, int32_t *detailed_error_code)
     char* dup_path = strdup(path);
     if(dup_path == NULL)
     {
-        *detailed_error_code = errno;
-        return FAIL_NOMEM;
+        rc = FAIL_NOMEM;
+        goto error_cleanup;
     }
     char *current_end = dup_path;
 
@@ -107,7 +107,7 @@ _ensure_path_exists(const char *path, int32_t *detailed_error_code)
         
         if (current_end != NULL)
             *current_end = '\0';
-        
+
         if (stat(dup_path, &sb) == -1)
         {
             if (errno != ENOENT)
@@ -116,37 +116,44 @@ _ensure_path_exists(const char *path, int32_t *detailed_error_code)
                 goto error_cleanup;
             }
 
-            if (mkdir(dup_path, 0700) == -1)
+            if (mkdir(dup_path, 0755) == -1)
             {
-                rc = FAIL_CREATE_DIRECTORY;
-                goto error_cleanup;
+                *detailed_error_code = errno;
+                char buf[1];
+                if (readlink(dup_path, buf, 1) != -1 && errno != EINVAL)
+                    rc = FAIL_BROKEN_LINK;
+                else
+                    rc = FAIL_CREATE_DIRECTORY;
+                goto cleanup;
             }
 
             rc = _sync_directory_for_internal(dup_path, detailed_error_code);
             if (rc != SUCCESS)
-                goto error_cleanup_with_error;
+                goto cleanup;
 
             rc = _sync_directory_for(dup_path, detailed_error_code);
             if (rc != SUCCESS)
-                goto error_cleanup_with_error;
+                goto cleanup;
         }
         else if(S_ISDIR(sb.st_mode) == 0)
         {
             *detailed_error_code = ENOTDIR;
             rc = FAIL_NOT_DIRECTORY;
-            goto error_cleanup_with_error;
+            goto cleanup;
         }
 
         if (current_end != NULL)
             *current_end = '/';
     }
 
-    return SUCCESS;
+    rc = SUCCESS;
+    goto cleanup;
 
 error_cleanup:
     *detailed_error_code = errno;
-error_cleanup_with_error:
-    free(dup_path);
+cleanup:
+    if (dup_path != NULL)
+        free(dup_path);
     return rc;
 }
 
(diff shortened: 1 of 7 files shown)


src/librvnpal/src/posix/fileutils.c
```diff
@@ -86,10 +87,67 @@ _allocate_file_space(int32_t fd, int64_t size, int32_t *detailed_error_code)
...
+            *detailed_error_code = ENOTDIR;
+            rc = FAIL_NOT_DIRECTORY;
+            goto error_cleanup_with_error;
+        }
+
+        if (current_end != NULL)
+            *current_end = '/';
+    }
+
+    return SUCCESS;
```
ayende:
```suggestion
    rc = SUCCESS;
    goto error_cleanup_with_error;
```

Here you are leaking `dup_path` on success.

src/librvnpal/src/posix/mapping.c
```diff
@@ -50,15 +51,20 @@ rvn_create_and_mmap64_file(
...
-    struct map_file_handle* mfh = calloc(1, sizeof(struct map_file_handle));
+    struct map_file_handle *mfh = calloc(1, sizeof(struct map_file_handle));
     *handle = mfh;
     if(mfh == NULL)
     {
         rc = FAIL_CALLOC;
         goto error_cleanup;
     }
+
+    char* directory = strdup(path);
```
ayende: Need to check if we failed here

ayende (changes requested):

src/librvnpal/src/posix/fileutils.c
aviviadi: free dup_path missing

src/librvnpal/src/posix/fileutils.c
```diff
@@ -86,10 +87,67 @@ _allocate_file_space(int32_t fd, int64_t size, int32_t *detailed_error_code)
...
 {
-    /* TODO: implement */
-    return SUCCESS;   
+    assert(path != NULL);
+
+    int32_t rc;
+    char* dup_path = strdup(path);
+    if(dup_path == NULL)
+    {
+        *detailed_error_code = errno;
```
aviviadi: goto errorCleanup

src/librvnpal/src/posix/fileutils.c
```diff
@@ -86,10 +87,67 @@ _allocate_file_space(int32_t fd, int64_t size, int32_t *detailed_error_code)
...
+        if (current_end != NULL)
+            *current_end = '/';
+    }
+
+    return SUCCESS;
+
+error_cleanup:
+    *detailed_error_code = errno;
+error_cleanup_with_error:
+    free(dup_path);
```
aviviadi: verify dup_path != NULL

src/librvnpal/src/posix/fileutils.c
ayende: Don't think that would be a good idea. It calls to `free(dup_path)`, and that is NULL.
Not sure what will happen, but not valid option to do

src/librvnpal/src/posix/fileutils.c
```diff
@@ -86,10 +87,67 @@ _allocate_file_space(int32_t fd, int64_t size, int32_t *detailed_error_code)
...
+        
+        if (stat(dup_path, &sb) == -1)
+        {
+            if (errno != ENOENT)
+            {
+                rc = FAIL_STAT_FILE;
+                goto error_cleanup;
+            }
+
+            if (mkdir(dup_path, 0700) == -1)
```
aviviadi: use 0755 permissions (like dotnet's System.IO.Directory.CreateDirectory)

src/librvnpal/src/posix/mapping.c
```diff
@@ -50,15 +51,20 @@ rvn_create_and_mmap64_file(
...
+    int ret = _ensure_path_exists(dirname(directory), detailed_error_code);
```
aviviadi: 1. verify path != null (however we do allow empty string)
2. dirname will return "." - verify _ensure_path_exists can handle string "."

ayende:
Also, write some tests here for adversarial input. Can we pass non path arguments here somehow? How does it handle it?

src/librvnpal/src/posix/mapping.c
```diff
@@ -48,17 +49,32 @@ rvn_create_and_mmap64_file(
...
-    struct map_file_handle* mfh = calloc(1, sizeof(struct map_file_handle));
+    struct map_file_handle *mfh = calloc(1, sizeof(struct map_file_handle));
     *handle = mfh;
     if(mfh == NULL)
     {
         rc = FAIL_CALLOC;
         goto error_cleanup;
     }
+
+    if(path[0] == '\0')
```
ayende: How can that be? What would be passing us something like this?
And if they are passing us an empty string, wouldn't this cause issues with `strdup` never ending?

ayende (changes requested):

src/librvnpal/src/posix/mapping.c
haludi: https://github.com/ravendb/ravendb/pull/8970#discussion_r266767782
@avivadi commented that we should accept an empty string and it should reference to current directory as `"."`. As I understand it `strdup` has no problem with an empty string but `stat` do
so I switch it with `"."`.

ayende: Consider the case where we send an empty C# string `""`.
This is translated by pinvoke to `"\0"` as a C string.
You are writing changing this to be: `"."`, but you _don't have a null terminator.

Now you pass a string without null terminator to `strdup` and will cause _issues_.

You can _change_ the value, so `if (path[0] == '\0') path = "."; `, which _is_ null terminated.

ayende: Also add a comment explaining this explicitly

ayende (approved):

ppekrol:
@haludi can you check failing tests?

PR #8977, opened 2019-03-19
RavenDB-12735 catch NotSupportedException and rethrow as IOException

karmeli87:
remove this workaround after updating to .net core 3
https://github.com/dotnet/corefx/issues/36141

### DIFF
--- a/src/Sparrow/Json/BlittableJsonTextWriter.cs
+++ b/src/Sparrow/Json/BlittableJsonTextWriter.cs
@@ -755,6 +755,12 @@ namespace Sparrow.Json
             {
                 //we are disposing, so this exception doesn't matter
             }
+            // TODO: remove when we update to .net core 3
+            // https://github.com/dotnet/corefx/issues/36141
+            catch (NotSupportedException e)
+            {
+                throw new IOException("The stream was closed by the peer.", e);
+            }
             finally
             {
                 _returnBuffer.Dispose();


src/Sparrow/Json/BlittableJsonTextWriter.cs
```diff
@@ -755,6 +755,17 @@ public virtual void Dispose()
             {
                 //we are disposing, so this exception doesn't matter
             }
+            // TODO: remove when we update to .net core 3
+            // https://github.com/dotnet/corefx/issues/36141
+            catch (NotSupportedException e)
+            {
+                if (e.Message.Contains("The WriteAsync method cannot be called when another write operation is pending"))
```
ayende: Not a good idea, may get different values under different cultures. 
Just wrap the exception

karmeli87: Okay, but what if we getting a different `NotSupportedException`? what about to check only for the word `WriteAsync` ?

ayende: I don't consider this likely, and the idea is that we'll restart anyway

karmeli87: okay, done.

ppekrol (approved):

PR #8979, opened 2019-03-19
RavenDB-11800 Don't allow adding node to the cluster when the leader initialize with port 0

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-11800

### DIFF
--- a/src/Raven.Server/ServerWide/ServerStore.cs
+++ b/src/Raven.Server/ServerWide/ServerStore.cs
@@ -186,7 +186,8 @@ namespace Raven.Server.ServerWide
         private PoolOfThreads.LongRunningWork _clusterMaintenanceSetupTask;
         private PoolOfThreads.LongRunningWork _updateTopologyChangeNotification;
 
-        public bool ValidateNotRandomPort = true;
+        public bool ValidateFixedPort = true;
+
         public Dictionary<string, ClusterNodeStatusReport> ClusterStats()
         {
             if (_engine.LeaderTag != NodeTag)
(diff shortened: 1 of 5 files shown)


src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -186,6 +186,7 @@ public long LastRaftCommitIndex
         private PoolOfThreads.LongRunningWork _clusterMaintenanceSetupTask;
         private PoolOfThreads.LongRunningWork _updateTopologyChangeNotification;
 
+        public bool ValidateNotRandomPort = true;
```
ayende:
```suggestion
        public bool ValidateNonRandomPort = true;
```

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -417,6 +418,18 @@ public ClusterTopology GetClusterTopology(TransactionOperationContext context)
...
         {
+
+            if (ValidateNotRandomPort)
+            {
+                if (Configuration.Core.ServerUrls != null &&
+                    Uri.TryCreate(Configuration.Core.ServerUrls[0], UriKind.Absolute, out var uri) &&
+                    uri.Port == 0)
+                {
+                    throw new InvalidOperationException($"Failed to add node '{nodeUrl}' to cluster. " +
+                                                        "Adding nodes to cluster is forbidden when the leader has port '0' in 'Configuration.Core.ServerUrls' setting");
```
ayende:
```suggestion
                                                        "Adding nodes to cluster is forbidden when the leader has port '0' in 'Configuration.Core.ServerUrls' setting. Define a fixed port for the node to enable cluster creation.");
```

ayende (commented):
Also, what about the other side, so we can't add a node that was started with port 0?

karmeli87:
Not sure if you can tell that other side started with port 0..
What we can do is to reject requests to add us to a cluster, so the only valid cluster for starting with port 0 will be single node.

ayende:
We can give that indication in `RachisAdminHandler.GetNodeInfo`, and check that in `AddNode`

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -415,8 +417,20 @@ public ClusterTopology GetClusterTopology(TransactionOperationContext context)
             return _engine.GetTopology(context);
         }
 
+        public bool HasFixedPort =>
+            Configuration.Core.ServerUrls == null ||
```
ppekrol, line 421: can this be `null`? thought that by default it is `localhost:8080`?

ppekrol (approved):

PR #8982, opened 2019-03-20
RavenDB-13164: Update exception message for stack trace collector on …

garayx:
…linux

### DIFF
--- a/src/Raven.Server/Documents/Handlers/Debugging/ThreadsHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/Debugging/ThreadsHandler.cs
@@ -10,6 +10,7 @@ using Raven.Server.Utils;
 using Raven.Server.Web;
 using Sparrow.Json;
 using Sparrow.Json.Parsing;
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Documents/Handlers/Debugging/ThreadsHandler.cs
```diff
@@ -18,6 +19,9 @@ public class ThreadsHandler : RequestHandler
         [RavenAction("/admin/debug/threads/stack-trace", "GET", AuthorizationStatus.Operator)]
         public Task StackTrace()
         {
+            if (PlatformDetails.RunningOnPosix)
+                throw new NotSupportedException("Stack Traces Capture is not supported on linux.");
```
arekpalinski: Maybe "not supported on POSIX" since it applies to Linux but macOS as well

garayx: done

arekpalinski (approved):

PR #8983, opened 2019-03-20
RavenDB-13161

arekpalinski:


src/Raven.Server/Web/System/ResourceNameValidator.cs
```diff
@@ -69,6 +69,11 @@ public static bool IsValidResourceName(string name, string dataDirectory, out st
                 errorMessage = $"The name '{name}' is forbidden for use!";
                 return false;
             }
+            if (name.Contains(".") && name.Any(char.IsLetterOrDigit) == false)
```
ayende: What about `/` or `\` ?

src/Raven.Client/Constants.cs
```diff
@@ -69,8 +69,6 @@ private Windows()
                 }
 
                 public static readonly int MaxPath = (int)typeof(Path).GetField("MaxLongPath", BindingFlags.Static | BindingFlags.NonPublic).GetValue(null);
-
-                public static readonly string[] ReservedFileNames = { "con", "prn", "aux", "nul", "com1", "com2", "com3", "com4", "com5", "com6", "com7", "com8", "com9", "lpt1", "lpt2", "lpt3", "lpt4", "lpt5", "lpt6", "lpt7", "lpt8", "lpt9", "clock$" };
```
ayende, line 73: Why did you remove this?

ml054: I wasn't used here. There is another place we have those constants.

src/Raven.Server/Web/System/ResourceNameValidator.cs
arekpalinski: `/` and `\` aren't valid character for database name and are checked by earlier validation check. 

Although `/` character is allowed for index name so I added that index name must not start with `.` if it contains `/` in  https://github.com/ravendb/ravendb/pull/8983/commits/d9c51cdf63e21211f202b21e3fb4357131a8fd2c

ppekrol (approved):

ayende (approved):

PR #8984, opened 2019-03-20
RavenDB-13152: invalid path in setup instruction

garayx:


src/Raven.Server/Commercial/SetupManager.cs
```diff
@@ -1902,7 +1902,7 @@ public static async Task SimulateRunningServer(X509Certificate2 serverCertificat
                     {
                         var ravenPath = typeof(RavenServer).Assembly.Location;
                         if (ravenPath.EndsWith(".dll", StringComparison.OrdinalIgnoreCase))
-                            ravenPath = ravenPath.Substring(ravenPath.Length - 4);
+                            ravenPath = ravenPath.Substring(0, ravenPath.Length - 4);
```
arekpalinski: minor: can we define `".dll"` as a const and use `.Length` instead of `4` to be explicit what is it from?

garayx: done

ayende (approved):

PR #8985, opened 2019-03-20
RavenDB-13168: Avoid scary users with regexps on invalid index name

garayx:


src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -703,7 +703,7 @@ public static void ValidateIndexName(string name, bool isStatic)
                 throw new ArgumentException($"Index name '{name}' not permitted. Static index name cannot start with 'Auto/'", nameof(name));
 
             if (isStatic && NameUtils.IsValidIndexName(name) == false)
-                throw new ArgumentException($"Index name '{name}' is not permitted. Only letters, digits and characters that match regex '{NameUtils.ValidIndexNameCharacters}' are allowed.", nameof(name));
+                throw new ArgumentException($"Index name '{name}' is not permitted. Only letters, digits and characters ('_', '-', '/', '.') are allowed.", nameof(name));
```
arekpalinski: Can we have a test which will ensure that if someone changes the regex we'll be aware of the need to change this text as well?

garayx: mb I can place debug assert inside that function?
Debug.Assert(NameUtils.ValidIndexNameCharacters == @"([_\/\-\.]+)")

arekpalinski (approved):

PR #8986, opened 2019-03-20
RavenDB_12868_3.5: Use reflection in order to find indexed documents …

maximburyak:
…dynamic objects representation __document_id field


ayende (approved):

Raven.Database/Indexing/Index.cs
```diff
@@ -1070,7 +1070,20 @@ public static string TryGetDocKey(object current)
         {
             var dic = current as DynamicJsonObject;
             if (dic == null)
-                return null;
+            {                
+                var properites = current.GetType().GetProperties();
+                var idProperty = properites.FirstOrDefault(x => x.Name == "__document_id");
```
arekpalinski: 1) Minor: please use `Constants.DocumentIdFieldName`
2) Should we check for `Constants.ReduceKeyFieldName` as well? (see below code)

ppekrol: Done

PR #8987, opened 2019-03-21
RavenDB-13180: Correctly escape characters in url on compare exchange…

garayx:
… operations.

### DIFF
--- a/test/SlowTests/Client/UniqueValues.cs
+++ b/test/SlowTests/Client/UniqueValues.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using FastTests;
@@ -10,7 +9,6 @@ using Raven.Client.Documents.Smuggler;
 using Raven.Client.ServerWide.Operations;
 using Raven.Server.ServerWide.Context;
 using Raven.Tests.Core.Utils.Entities;
-using Sparrow.Json;
 using Xunit;
 
 namespace SlowTests.Client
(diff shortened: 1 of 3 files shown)


test/SlowTests/Client/UniqueValues.cs
```diff
@@ -295,5 +295,55 @@ public async Task CompareExchangeShouldBeRemovedFromStorageWhenDbGetsDeleted()
...
             }
         }
+
+        [Fact]
+        public async Task EnsureCanCompareExchangeWithCorrectCharsEscapeInKey()
+        {
+            DoNotReuseServer();
+            var store = GetDocumentStore();
+
+            const string cmpXchgKey1 = "emails/foo+test@email.com";
```
ml054: can't we wrap this with a method and call it with arg and expected result to avoid code duplication?

garayx: ofc I can, but I am not sure if this is necessary for test.

JudahGabriel:
Egor, thanks for the fix! Can you confirm: if I store a cmpxchg with "emails/foo+test@ab.com", **will I be able to get that value using GetCompareExchangeValueOperation(keyWithPlusSign), or will I need to manually escape the keyWithPlusSign**?

ml054:
@JudahGabriel according to test https://github.com/ravendb/ravendb/pull/8987/files#diff-d7adabe65d19221ae0f622e52c1b2751R307

you don't have to escape nothing.

garayx:
> Egor, thanks for the fix! Can you confirm: if I store a cmpxchg with "emails/foo+test@ab.com", will I be able to get that value using GetCompareExchangeValueOperation(keyWithPlusSign), or will I need to manually escape the keyWithPlusSign?

Yes, the same key should be used for Put, Get, Delete operations despite 'special characters' used

arekpalinski (approved):

grisha-kotler (approved):

PR #8988, opened 2019-03-21
[WIP] RavenDB-10956 Support remove CV entries from a specific database

karmeli87:
The removal occurs in 2 phases.
First we send command for removal of specific CVs. Upon committing the command each node will strip those CVs from _outgoing_ replication. So the peers will not receive documents with those CVs.

Only when all outgoing connections of a node stripping those CVs the node will confirm the removal command.
In the second phase when all nodes confirmed the 1st phase it is safe to remove those CVs from the global database change vector.

### DIFF
--- a/src/Raven.Client/ServerWide/DatabaseRecord.cs
+++ b/src/Raven.Client/ServerWide/DatabaseRecord.cs
@@ -15,6 +15,17 @@ using Raven.Client.Exceptions.Documents.Indexes;
 
 namespace Raven.Client.ServerWide
 {
+    public class RemoveChangeVectors
+    {
+        public List<string> IgnoredList;
+
+        public long Index;
+
+        public List<string> Confirmations;
+
+        public bool ShouldUpdateGlobal;
+    }
+
     public class DatabaseRecordWithEtag : DatabaseRecord
     {
         public long Etag { get; set; }
@@ -64,6 +75,8 @@ namespace Raven.Client.ServerWide
         public ExpirationConfiguration Expiration;
 
         public List<PeriodicBackupConfiguration> PeriodicBackups = new List<PeriodicBackupConfiguration>();
+
+        public RemoveChangeVectors RemoveChangeVectors;
 
         public List<ExternalReplication> ExternalReplications = new List<ExternalReplication>();
 
--- a/src/Raven.Server/Documents/DocumentDatabase.cs
+++ b/src/Raven.Server/Documents/DocumentDatabase.cs
@@ -6,6 +6,7 @@ using System.IO.Compression;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Raven.Client.Documents.Linq;
 using Raven.Client.Documents.Operations.Configuration;
 using Raven.Client.Documents.Session;
 using Raven.Client.Documents.Smuggler;
@@ -29,6 +30,7 @@ using Raven.Server.Documents.TcpHandlers;
 using Raven.Server.Json;
 using Raven.Server.NotificationCenter.Notifications;
 using Raven.Server.NotificationCenter.Notifications.Details;
+using Raven.Server.Rachis;
 using Raven.Server.Routing;
 using Raven.Server.ServerWide;
 using Raven.Server.ServerWide.Commands;
@@ -245,6 +247,8 @@ namespace Raven.Server.Documents
         public readonly QueryMetadataCache QueryMetadataCache;
 
         public long LastTransactionId => DocumentsStorage.Environment.CurrentReadTransactionId;
+
+        public List<string> IgnoredChangeVectors;
 
         public void Initialize(InitializeOptions options = InitializeOptions.None)
         {
@@ -1124,6 +1128,7 @@ namespace Raven.Server.Documents
                         EtlLoader?.HandleDatabaseRecordChange(record);
                         OnDatabaseRecordChanged(record);
                         SubscriptionStorage?.HandleDatabaseRecordChange(record);
+                        HandleIgnoreChangeVectors(record);
 
                         if (_logger.IsInfoEnabled)
                             _logger.Info($"Finish to process record {index} for {record.DatabaseName}.");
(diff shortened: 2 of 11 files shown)


karmeli87:
This PR is still missing tests and some of the logic, but I would like to have feedback for what I've done so far.

src/Raven.Client/ServerWide/DatabaseRecord.cs
```diff
@@ -15,6 +15,17 @@
 
 namespace Raven.Client.ServerWide
 {
+    public class RemoveChangeVectors
```
ayende, line 18: Should be either nested class or a separate file

src/Raven.Client/ServerWide/DatabaseRecord.cs
```diff
@@ -65,6 +76,8 @@ public DatabaseRecord(string databaseName)
 
         public List<PeriodicBackupConfiguration> PeriodicBackups = new List<PeriodicBackupConfiguration>();
 
+        public RemoveChangeVectors RemoveChangeVectors;
```
ayende, line 79: What if I have multiple removals?

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -1143,6 +1148,89 @@ private void NotifyFeaturesAboutStateChange(DatabaseRecord record, long index)
...
+                    return;
+            }
+
+            SetDatabaseChangeVectorIfNeeded(newRecord.RemoveChangeVectors?.ShouldUpdateGlobal);
+
+            IgnoredChangeVectors = newIgnoreList;
+            if (IgnoredChangeVectors == null)
+                return;
+
+            ConfirmChangeVectorRemoval(newRecord);
```
ayende, line 1170: Shouldn't we confirm anyway if we removed the ignored list? Or is it expected that this will be empty list, then?

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -1143,6 +1148,89 @@ private void NotifyFeaturesAboutStateChange(DatabaseRecord record, long index)
...
+        private void ConfirmChangeVectorRemoval(DatabaseRecord newRecord)
+        {
+            var index = newRecord.RemoveChangeVectors?.Index ?? 0;
+            if (index > 0)
+            {
+                Task.Run(async () =>
+                {
+                    while (DatabaseShutdown.IsCancellationRequested == false)
+                    {
+                        if (ReplicationLoader.OutgoingHandlers.Any() == false || ReplicationLoader.OutgoingHandlers.Min(h => h.RaftIndex) >= index)
```
ayende, line 1182: I don't understand why you are involving the `OutgoingHandlers` here?
I think that I'm missing something here in terms of logical flow of the operation.

In terms of what is actually going on, the `OutgoingHandlers` may not contain tasks that have just been added (this can happen on db startup, mind) or tasks that failed.

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -1143,6 +1148,89 @@ private void NotifyFeaturesAboutStateChange(DatabaseRecord record, long index)
...
+                        {
+                            try
+                            {
+                                var res = await ServerStore.SendToLeaderAsync(new ConfirmRemoveChangeVectorCommand(Name, ServerStore.NodeTag));
+                                await ServerStore.Engine.WaitForCommitIndexChange(RachisConsensus.CommitIndexModification.GreaterOrEqual, res.Index);
+                                break;
```
ayende, line 1188: You need cancellation tokens here, no?
Or is that handled internally?

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -1143,6 +1148,89 @@ private void NotifyFeaturesAboutStateChange(DatabaseRecord record, long index)
...
+        }
+
+        private void SetDatabaseChangeVectorIfNeeded(bool? shouldUpdate)
+        {
+            var ignoredList = IgnoredChangeVectors;
+            if (shouldUpdate == true && ignoredList != null)
+            {
+                Task.Run(async () =>
+                {
+                    while (DatabaseShutdown.IsCancellationRequested == false)
```
ayende, line 1212: Why is this a loop?

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -1234,6 +1322,23 @@ private void InitializeFromDatabaseRecord(DatabaseRecord record)
...
 
+        public static void RemoveIgnoredChangeVectors(List<string> list, ref string changeVector)
+        {
+            if (list == null || list.Count == 0)
+                return;
+
+            if (changeVector == null)
+                return;
+
+            //TODO: optimize this
```
ayende, line 1333: Probably *rare* operation, don't really need to optimize it

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1833,6 +1833,39 @@ public static LazyStringValue TableValueToId(JsonOperationContext context, int i
...
+
+            public SetDatabaseChangeVectorCommand(List<string> ignoredList)
+            {
+                _ignoredList = ignoredList;
+            }
+
+            protected override int ExecuteCmd(DocumentsOperationContext context)
+            {
+                Debug.Assert(_ignoredList.Count > 0);
+                DocumentDatabase.RemoveIgnoredChangeVectors(_ignoredList, ref context.LastDatabaseChangeVector);
```
ayende, line 1849: first use `context.LastDatabaseChangeVector = context.LastDatabaseChangeVector ?? GetDatabaseChangeVector()`

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1833,6 +1833,39 @@ public static LazyStringValue TableValueToId(JsonOperationContext context, int i
...
+                var global = GetDatabaseChangeVector(context);
+                if (global != null)
+                {
+                    DocumentDatabase.RemoveIgnoredChangeVectors(_ignoredList, ref global);
+                    var tree = context.Transaction.InnerTransaction.ReadTree(GlobalTreeSlice);
```
ayende, line 1854: I'm pretty sure we have `SaveDatabseChangeVector`, no? If no, create it, this leaks the storage abstraction

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
```diff
@@ -475,6 +479,8 @@ private void Replicate()
             }
         }
 
+        public long RaftIndex;
```
ayende, line 482: I don't understand what is the purpose of this. 
This just tells you what is the last raft index that we sent this with, not whatever we properly updated anything internally about it.

Note that the `LastRaftCommitIndex` is updated *without* waiting for the commands to actually finish applying (async)

ayende (changes requested):
I don't follow why you are using the replication for that.

The flow that I had in mind was that you register the list of CV to filter.
You then immediately update the global change vector.
You then filter out this db ids whenever we generate a new CV.

The list of db ids to ignore is static and remains the same until admin edits it again.

Another possible impl is to just filter the change vectors at _global db cv time_, which is probably cheaper, but need some verification to see that it does the right thing.

ppekrol:
Closing. Moved to 4.2.x

PR #9014, opened 2019-04-01
RavenDB-13163 creating documents with very long names

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-13163


grisha-kotler (changes requested):
Need another test for attachments

src/Raven.Server/Documents/AttachmentsStorage.cs
```diff
@@ -166,6 +166,11 @@ public IEnumerable<DynamicJsonValue> GetAttachmentsMetadataForDocumentWithCounts
                 if (TableValueToFlags((int)DocumentsTable.Flags, ref tvr).HasFlag(DocumentFlags.Artificial))
                     throw new InvalidOperationException($"Cannot put attachment {name} on artificial document '{documentId}'.");
 
+                if (documentId.Length + name.Length > DocumentIdWorker.MaxIdSize)
```
ppekrol: Can we check this earlier? After checking if transaction is null?

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -775,12 +772,32 @@ private static SortedSet<string> GetCountersForDocument(BlittableJsonReaderArray
...
+            {
+                ThrowStorageKeyTooBig(documentId, counterName);
+            }
+        }
+
         private static void ThrowCounterNameTooBig(string name)
         {
             throw new ArgumentException(
-                $"Counter name cannot exceed {MaxCounterNameSize} bytes, but counter name was {Encoding.UTF8.GetByteCount(name)} bytes. The invalid counter name is '{name}'.",
+                $"Counter name cannot exceed {DocumentIdWorker.MaxIdSize} bytes, but counter name was {Encoding.UTF8.GetByteCount(name)} bytes. The invalid counter name is '{name}'.",
```
ppekrol: I think we can get rid of the GetByteCount here? We are using string.Length for that now?

karmeli87:
We also need to consider if those exceptions are thrown during the `Apply` of the `ClusterStateMachine`.
Is it okay to commit?
if it isn't we have a problem, because we cannot recover from this.
If it is fine to commit, a custom exception `KeyTooBigException` would be better, so we can catch it and treat as "Expected".

src/Raven.Server/Documents/AttachmentsStorage.cs
aviv86: done

src/Raven.Server/Documents/CountersStorage.cs

grisha-kotler (approved):

ppekrol (approved):

PR #9015, opened 2019-04-01
RavenDB-13163 creating documents with very long names

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-13163


src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -796,12 +812,14 @@ private struct PutCountersData
 
                         using (Slice.External(context.Allocator, kvp.Key, out var countersGroupKey))
                         {
-                            if (currentData.Size > MaxCounterDocumentSize)
+                            currentData.TryGet(Values, out BlittableJsonReaderObject localCounters);
+
+                            if (currentData.Size > MaxCounterDocumentSize &&
+                                localCounters.Count > 1)
```
ayende: why are you checking the count here?

src/Raven.Server/Documents/DocumentIdWorker.cs
```diff
@@ -35,16 +38,15 @@ static DocumentIdWorker()
             var maxStrSize = Encoding.GetMaxByteCount(strLength);
             var escapePositionsSize = JsonParserState.FindEscapePositionsMaxSize(id);
 
-            
+            if (strLength > MaxIdSize)
```
ayende, line 40: Isn't that too big? We are testing limitting this to 512 to have more breathing room

ayende: The key issue, BTW, is that we use the document id in many places.
For example, a doc id that is max size + counter id that is max size, etc.

I think that we should have a hard limit of 512 and that is it.

src/Raven.Server/Documents/CountersStorage.cs
aviv86: if counter name is huge then it's possible to exceed `MaxCounterDocumentSize` while having just one counter in counter-group document

ayende: I don't see how that can be. We should have a much lower limit to the size of a counter name

ayende (approved):

PR #9016, opened 2019-04-01
RavenDB-13319 - Missing cast on index creation

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-13319


ppekrol:
Workaround is simple and the chance to reset user indexes during simple patch upgrade is too high. Move this to 4.2.

PR #9023, opened 2019-04-02
RavenDB-13228 Recover from an invalid cluster state where one node ha…

karmeli87:
…s higher term while the other has more appended entries

### DIFF
--- a/src/Raven.Server/Rachis/FollowerAmbassador.cs
+++ b/src/Raven.Server/Rachis/FollowerAmbassador.cs
@@ -732,7 +732,7 @@ namespace Raven.Server.Rachis
 
         private long InitialNegotiationWithFollower()
         {
-            UpdateLastMatchFromFollower(0);
+            Interlocked.Exchange(ref _followerMatchIndex, 0);
             using (_engine.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
             {
                 LogLengthNegotiation lln;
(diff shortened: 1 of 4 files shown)


ppekrol (approved):

src/Raven.Server/Rachis/FollowerAmbassador.cs
```diff
@@ -732,7 +732,7 @@ public void Copy(byte* ptr, int size)
 
         private long InitialNegotiationWithFollower()
         {
-            UpdateLastMatchFromFollower(0);
+            Interlocked.Exchange(ref _followerMatchIndex, 0);
```
ayende, line 735: Why not call `UpdateLastMatchFromFollower` here?

ayende (commented):
Other than that, LGTM

src/Raven.Server/Rachis/FollowerAmbassador.cs
karmeli87: It updates also the `_lastReplyFromFollower` which might cause the leader to think that this follower is alive even when no message was received from it. 
If this happens in a loop, the leader is not moving to a candidate state, when he should.

ayende: The point there was to give us a leg up if the initial connection attempt is a long one. 
`InitialNegotiationWithFollower` is not called in a loop, does it?

karmeli87: We are giving a leg up on spinning the follower ambassador thread.

In the rachis test we fail on the first read from the follower (which is in `InitialNegotiationWithFollower`), because the follower sends an error message. So the ambassador try again by obtaining the connection and fails again on the first read.. and this happens in a loop.

PR #9024, opened 2019-04-02
RavenDB-13165 add new features to about page

ml054:


src/Raven.Studio/wwwroot/App/views/shell/about.html
```diff
@@ -137,11 +137,17 @@ <h4>External Replication</h4>
                     <div class="name">Immediate</div><div class="value"><i data-bind="attr: { class: licenseAttribute('HasExternalReplication') }"></i></div>
                     <div class="name">Delayed</div><div class="value"><i data-bind="attr: { class: licenseAttribute('HasDelayedExternalReplication') }"></i></div>
 
+                    <h4>Pull Replication</h4>
```
ppekrol, line 141: Can we add Counters to `Extensions`? We have `Attachments` there

ml054: done

PR #9026, opened 2019-04-02
RavenDB-13237 We must not add a configuration name of ETL task which …

arekpalinski:
…is non relevant for a given node to unique configuration names hash set. That resulted in non working ETL task failover after node restart.

### DIFF
--- a/src/Raven.Server/Documents/ETL/EtlLoader.cs
+++ b/src/Raven.Server/Documents/ETL/EtlLoader.cs
@@ -24,7 +24,7 @@ namespace Raven.Server.Documents.ETL
         private const string AlertTitle = "ETL loader";
 
         private EtlProcess[] _processes = new EtlProcess[0];
-        private readonly HashSet<string> _uniqueNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase); // read and modified under a lock.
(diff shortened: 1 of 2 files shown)


src/Raven.Server/Documents/ETL/EtlLoader.cs
```diff
@@ -75,16 +75,18 @@ public void Initialize(DatabaseRecord record)
                     foreach (var process in toRemove)
                     {
                         processes.Remove(process);
-                        _uniqueNames.Remove(process.ConfigurationName);
+                        _uniqueConfigurationNames.Remove(process.ConfigurationName);
                     }
                 }
 
+                var ensureUniqueConfigurationNames = _uniqueConfigurationNames.ToHashSet();
```
ayende: This is already a `HashSet`, but I _think_ that in this case you'll be using one that is using case _sensitive_ comparer, which can cause conflicts later.

arekpalinski: You're right. Fixed in https://github.com/ravendb/ravendb/pull/9026/commits/5a9f0ff0c6846effe215c3b8dade4e56b89ea6bf. 

I'm making a copy of the global `HashSet` for validation purposes to avoid adding names of ETL tasks during the validation which aren't going to be started on this node anyway.

PR #9027, opened 2019-04-02
RavenDB-13309 can't get subscription information via client api

efratshenhar:


src/Raven.Server/Web/System/OngoingTasksHandler.cs
```diff
@@ -950,9 +951,18 @@ public Task GetOngoingTaskInfo()
...
+                            {
+                                var subscriptions = Database.SubscriptionStorage.GetAllSubscriptions(context, false, 0, int.MaxValue);
+                                foreach (var subscription in subscriptions)
+                                {
+                                    if (subscription.SubscriptionId != key) continue;
+                                    name = subscription.SubscriptionName;
+                                    break;
+                                }
+                            }
+                            itemKey = SubscriptionState.GenerateSubscriptionItemKeyName(record.DatabaseName, name);
```
ppekrol, line 962: What if name will be null after checking all of the available subscriptions?

ppekrol (approved):

PR #9029, opened 2019-04-02
RavenDB-13234 Fixing an issue where we register command for disposal …

talweiss1982:
…to the wrong context


ayende (approved):
Very nice catch!

ppekrol (approved):

PR #9034, opened 2019-04-02
 RavenDB-13321: fix concurrency in test

garayx:


src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -428,6 +460,12 @@ public bool IsDatabaseLoaded(StringSegment databaseName)
             return false;
         }
 
+        public bool DatabaseRecordExists(string databaseName)
+        {
+            var record = _serverStore.LoadDatabaseRecord(databaseName, out _);
```
ppekrol: better to use `ReadRawDatabase` from ServerStore.Cluster

src/Raven.Client/Documents/Identity/AsyncHiLoIdGenerator.cs
```diff
@@ -76,7 +77,18 @@ public RangeValue(long min, long max)
         /// <returns></returns>
         public Task<string> GenerateDocumentIdAsync(object entity)
         {
-            return NextIdAsync().ContinueWith(task => GetDocumentIdFromId(task.Result));
+            return NextIdAsync().ContinueWith(task =>
```
ppekrol: What is the relation between this failing test and this fix?
Would it be better to do await NextIdAsync and then execute second operation?

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -168,7 +168,32 @@ private void HandleClusterDatabaseChanged(string databaseName, long index, strin
             }
         }
 
-        private void UnloadDatabase(string databaseName)
+        private void UnloadDatabase(string databaseName, bool dbRecordIsNull = false)
+        {
+            if (dbRecordIsNull)
+            {
+                UnloadDatabaseInternal(databaseName);
```
ppekrol, line 176: Need a description here explaining the situation

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -178,6 +203,13 @@ private void UnloadDatabase(string databaseName)
             {
                 // this is already in the process of being deleted, we can just exit and let the other thread handle it
             }
+            catch (Exception e)
+            {
```
ppekrol, line 236: We should rethrow here? Now we are ignoring all the exceptions.

src/Raven.Client/Documents/Identity/AsyncHiLoIdGenerator.cs
```diff
@@ -76,7 +77,18 @@ public RangeValue(long min, long max)
...
+            {
+                try
+                {
+                    return GetDocumentIdFromId(task.Result);
+                }
+                catch (Exception e)
```
ayende: This obfuscate the actual exception stack trace. 
I don't follow why this is needed?

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -168,7 +168,32 @@ private void HandleClusterDatabaseChanged(string databaseName, long index, strin
...
+            if (DatabasesCache.TryGetValue(databaseName, out var databaseTask) == false)
+            {
+                // database was already unloaded or deleted
+                return;
+            }
+
+            if (databaseTask.IsCompleted == false)
+            {
+                // This case is when an unload was issued prior to the actual loading of a database.
+                databaseTask.ContinueWith(t => { UnloadDatabaseInternal(databaseName); });
```
ayende: Here you need to check that the load itself was _successful_.
Another issue here is that here you are handling the unload in an _async_ manner.
I'm not sure that this is the right thing to do, and it may cause us concurrency issues.
In some cases, we'll issue a request to unload, but it takes time for this to be processed, but the request that was supposed to unload already returned, leading to file busy issues?

src/Raven.Server/Documents/DatabasesLandlord.cs
ayende: Yes, I don't like this. The error handling above is very explicit about why it is handling this. Why add this `catch`?

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -260,7 +260,8 @@ public async Task Put()
                     throw licenseLimit;
                 }
 
-                if (ServerStore.DatabasesLandlord.IsDatabaseLoaded(name) == false)
+                if (ServerStore.DatabasesLandlord.IsDatabaseLoaded(name) == false && 
+                    ServerStore.DatabasesLandlord.DatabaseRecordExists(name) == false)
```
ayende: Why did you add the second check? Any update to the actual database record will not trigger this now, would it?

ayende (changes requested):

src/Raven.Client/Documents/Identity/AsyncHiLoIdGenerator.cs
grisha-kotler: This threw `AggregatedException` instead of `DatabaseDisabledException` as expected.
Before the change the `HiLo` request succeeded while the `SaveChanges` with `DatabaseDisabledException`.

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
grisha-kotler: It should be only triggered for newly created databases on this node.

src/Raven.Client/Documents/Identity/AsyncHiLoIdGenerator.cs
ayende: So handle that at the caller, or turn this into an `async` method, which will do that too

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
ayende: Why does that matter?

src/Raven.Server/Documents/DatabasesLandlord.cs

src/Raven.Server/Documents/DatabasesLandlord.cs
garayx: done

src/Raven.Client/Documents/Identity/AsyncHiLoIdGenerator.cs
garayx: done

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
garayx: removed

src/Raven.Server/Documents/DatabasesLandlord.cs
garayx: there is a comment on line 73, I can also add it here.

src/Raven.Server/Documents/DatabasesLandlord.cs
garayx: > Here you need to check that the load itself was successful.

added

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -182,14 +182,18 @@ private void UnloadDatabase(string databaseName, bool dbRecordIsNull = false)
...
             }
-            else
+            else if (databaseTask.IsCompleted == false)
             {
-                UnloadDatabaseInternal(databaseName);
+                // This case is when an unload was issued prior to the actual loading of a database.
+                databaseTask.ContinueWith(t =>
+                {
+                    if (databaseTask.IsCompletedSuccessfully)
+                        UnloadDatabaseInternal(databaseName);
```
ayende: This can be delayed arbitrarily, no?

src/Raven.Client/Documents/Identity/AsyncMultiDatabaseHiLoIdGenerator.cs
```diff
@@ -25,11 +25,11 @@ public AsyncMultiDatabaseHiLoIdGenerator(DocumentStore store, DocumentConvention
         }
 
 
-        public Task<string> GenerateDocumentIdAsync(string dbName, object entity)
+        public async Task<string> GenerateDocumentIdAsync(string dbName, object entity)
         {
             var db = dbName ?? Store.Database;
             var generator = _generators.GetOrAdd(db, GenerateAsyncMultiTypeHiLoFunc);
-            return generator.GenerateDocumentIdAsync(entity);
+            return await generator.GenerateDocumentIdAsync(entity).ConfigureAwait(false);
```
ppekrol: redundant `await` ?

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -168,15 +168,62 @@ private void HandleClusterDatabaseChanged(string databaseName, long index, strin
...
+                // This case is when an unload was issued prior to the actual loading of a database.
+                databaseTask.ContinueWith(t =>
+                {
+                    if (databaseTask.IsCompletedSuccessfully)
+                    {
+                        using (_serverStore.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
+                        using (context.OpenReadTransaction())
+                        {
+                            var databaseRecord = _serverStore.Cluster.ReadRawDatabase(context, databaseName, out _);
+                            databaseRecord.TryGet(nameof(DatabaseRecord.Disabled), out bool dbDisabled);
```
ppekrol: Check DatabasesInfoNotificationSender.IsDatabaseDisabled implementation. It takes more factors into account (e.g. DatabaseState).

ppekrol (changes requested):

src/Raven.Client/Documents/Identity/AsyncMultiDatabaseHiLoIdGenerator.cs
ayende: Yes, we can just pass it through.

src/Raven.Client/Documents/Identity/AsyncMultiTypeHiLoIdGenerator.cs
```diff
@@ -30,27 +29,35 @@ public AsyncMultiTypeHiLoIdGenerator(DocumentStore store, string dbName, Documen
...
+            }
 
-            lock (_generatorLock)
+            await _generatorLock.WaitAsync().ConfigureAwait(false);
+
+            try
             {
                 if (_idGeneratorsByTag.TryGetValue(tag, out value))
-                    return value.GenerateDocumentIdAsync(entity);
+                    return await value.GenerateDocumentIdAsync(entity).ConfigureAwait(false);
```
ayende: This is likely a problem. 
You are doing the expensive ( network call wait) _while_ you are holding the lock.
This needs to be moved outside the lock

ayende (changes requested):

src/Raven.Server/Documents/DatabasesLandlord.cs

src/Raven.Client/Documents/Identity/AsyncMultiTypeHiLoIdGenerator.cs
garayx: done

src/Raven.Client/Documents/Identity/AsyncMultiDatabaseHiLoIdGenerator.cs
garayx: done

src/Raven.Client/Documents/Identity/AsyncMultiTypeHiLoIdGenerator.cs
grisha-kotler: _idGeneratorsByTag is a ConcurrentDictionary can't we use GetOrAdd?

ayende: no, the `lock` there is fine

ayende (approved):

ppekrol (approved):

PR #9039, opened 2019-04-03
RavenDB-12973 Limit subscription batch size in MB

efratshenhar:


src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
```diff
@@ -82,6 +87,7 @@ public class SubscriptionDocumentsFetcher
                     0,
                     int.MaxValue))
                 {
+                    _threadAllocations = NativeMemory.CurrentThreadStats;
```
ayende: Shouldn't this be _before_ the `foreach` ? This would cause the value to reset per iteration

src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
```diff
@@ -129,10 +135,43 @@ public class SubscriptionDocumentsFetcher
...
+        public bool CanContinueBatch(DocumentsOperationContext ctx)
+        {
+            var totalAllocated = _threadAllocations.TotalAllocated;
+            _threadAllocations.CurrentlyAllocatedForProcessing = totalAllocated;
+            var currentlyInUse = new Size(totalAllocated, SizeUnit.Bytes);
+            if (currentlyInUse > _currentMaximumAllowedMemory)
+            {
+                if (MemoryUsageGuard.TryIncreasingMemoryUsageForThread(_threadAllocations, ref _currentMaximumAllowedMemory,
+                        currentlyInUse,
+                        _db.DocumentsStorage.Environment.Options.RunningOn32Bits, _logger, out var memoryUsage) ) return true;
```
ayende: `return true` on a separate line

ayende (commented):
I think that a much simpler method here would be to just count the size of the data we send on the wire and cut the batch after some MB size.
What you are doing here relates more to memory management than anything else, and we shouldn't be sending a batch of 32 MB to the client anyway.
Also note that the code as written allow the server to send much larger sizes if the memory budget allows it.

src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
```diff
@@ -90,6 +95,7 @@ public class SubscriptionDocumentsFetcher
                     _threadAllocations = NativeMemory.CurrentThreadStats;
                     using (doc.Data)
                     {
+                        size += new Size(doc.Data.Size, SizeUnit.Megabytes);
```
ayende: Do we need to consider tombstones here?

src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
```diff
@@ -34,7 +35,7 @@ public class SubscriptionDocumentsFetcher
         private readonly SubscriptionState _subscription;
         private readonly SubscriptionPatchDocument _patch;
         private NativeMemory.ThreadStats _threadAllocations;
-        private Size _currentMaximumAllowedMemory = new Size(32, SizeUnit.Megabytes);
+        private Size _MaximumAllowedMemory;
```
ayende: `_MaximumAllowedMemory` -> `_maximumAllowedMemory`

Need to add a comment explaining why we have this limit

src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
```diff
@@ -77,7 +78,11 @@ public class SubscriptionDocumentsFetcher
              IncludeDocumentsCommand includesCmd,
             long startEtag)
         {
-            int size = 0;
+            int numberOfDocs = 0;
+            Size size = new Size(0 , SizeUnit.Megabytes);
+            _MaximumAllowedMemory = new Size((PlatformDetails.Is32Bits ||
```
ayende: This should be initialized once, not each time

src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
```diff
@@ -133,51 +139,24 @@ public class SubscriptionDocumentsFetcher
...
-
         private IEnumerable<(Document Doc, Exception Exception)> GetRevisionsToSend(
             DocumentsOperationContext docsContext,
             IncludeDocumentsCommand includesCmd,
             long startEtag)
         {
-            int size = 0;
+            int numberOfDocs = 0;
+            Size size = new Size(0, SizeUnit.Megabytes);
+            _MaximumAllowedMemory = new Size((PlatformDetails.Is32Bits ||
```
ayende: This should be initialized once, not each time

ayende (changes requested):

ppekrol (approved):

src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
ppekrol: done

src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
ppekrol: done

src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
ppekrol: tombstones? we are not processing them here?

src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
ppekrol: done

ppekrol (approved):

PR #9041, opened 2019-04-03
RavenDB-13247 changes in certificate executable

iftahbe:
The test passes locally both in Windows and Linux. @ppekrol, I wish we could somehow run those tests in Jenkins anyway (maybe just run one of them separately in its own test suite?).

### DIFF
--- a/src/Raven.Server/Config/Categories/SecurityConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/SecurityConfiguration.cs
@@ -49,14 +49,39 @@ namespace Raven.Server.Config.Categories
 
         [Description("A command or executable providing a .pfx certificate file. If specified, RavenDB will use HTTPS/SSL for all network activities. The certificate path setting takes precedence over executable configuration option.")]
         [DefaultValue(null)]
-        [ConfigurationEntry("Security.Certificate.Exec.V2", ConfigurationEntryScope.ServerWideOnly)]
-        public string CertificateExecV2 { get; set; }
+        [ConfigurationEntry("Security.Certificate.Exec.Load", ConfigurationEntryScope.ServerWideOnly)]
+        public string CertificateExecLoad { get; set; }
+
+        [Description("A command or executable providing a .pfx certificate file. If specified, RavenDB will use HTTPS/SSL for all network activities. The certificate path setting takes precedence over executable configuration option.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.Certificate.Exec.Renew", ConfigurationEntryScope.ServerWideOnly)]
+        public string CertificateExecRenew { get; set; }
+
+        [Description("A command or executable providing a .pfx certificate file. If specified, RavenDB will use HTTPS/SSL for all network activities. The certificate path setting takes precedence over executable configuration option.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.Certificate.Exec.OnCertificateChange", ConfigurationEntryScope.ServerWideOnly)]
+        public string CertificateExecOnCertificateChange { get; set; }
 
         [Description("The command line arguments for the 'Security.Certificate.Exec' command or executable.")]
         [DefaultValue(null)]
         [ConfigurationEntry("Security.Certificate.Exec.Arguments", ConfigurationEntryScope.ServerWideOnly)]
         public string CertificateExecArguments { get; set; }
 
+        [Description("The command line arguments for the 'Security.Certificate.Exec.Load' command or executable.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.Certificate.Exec.Load.Arguments", ConfigurationEntryScope.ServerWideOnly)]
+        public string CertificateExecLoadArguments { get; set; }
+
+        [Description("The command line arguments for the 'Security.Certificate.Exec.Renew' command or executable.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.Certificate.Exec.Renew.Arguments", ConfigurationEntryScope.ServerWideOnly)]
+        public string CertificateExecRenewArguments { get; set; }
+
+        [Description("The command line arguments for the 'Security.Certificate.Exec.OnCertificateChange' command or executable.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.Certificate.Exec.OnCertificateChange.Arguments", ConfigurationEntryScope.ServerWideOnly)]
+        public string CertificateExecOnCertificateChangeArguments { get; set; }
+        
         [Description("The number of seconds to wait for the certificate executable to exit. Default: 30 seconds")]
         [DefaultValue(30)]
         [TimeUnit(TimeUnit.Seconds)]
@@ -110,7 +135,7 @@ namespace Raven.Server.Config.Categories
 
         public bool IsCertificateConfigured =>
             string.IsNullOrWhiteSpace(CertificatePath) == false ||
-            string.IsNullOrWhiteSpace(CertificateExecV2) == false;
+            string.IsNullOrWhiteSpace(CertificateExecLoad) == false;
 
         public bool AuthenticationEnabled => IsCertificateConfigured;
 
@@ -133,7 +158,7 @@ namespace Raven.Server.Config.Categories
                 {
                     if (isServerUrlHttps == false)
                         throw new InvalidOperationException(
-                            $"When the server certificate in either `{RavenConfiguration.GetKey(x => x.Security.CertificatePath)}` or `{RavenConfiguration.GetKey(x => x.Security.CertificateExecV2)}`  is specified, the `{RavenConfiguration.GetKey(x => x.Core.ServerUrls)}` must be using https, but was " +
+                            $"When the server certificate in either `{RavenConfiguration.GetKey(x => x.Security.CertificatePath)}` or `{RavenConfiguration.GetKey(x => x.Security.CertificateExecLoad)}` is specified, the `{RavenConfiguration.GetKey(x => x.Core.ServerUrls)}` must be using https, but was " +
                             serverUrl);
                 }
                 else
--- a/src/Raven.Server/Config/Categories/SecurityConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/SecurityConfiguration.cs
@@ -42,45 +42,45 @@ namespace Raven.Server.Config.Categories
         [ConfigurationEntry("Security.Certificate.Password", ConfigurationEntryScope.ServerWideOnly)]
         public string CertificatePassword { get; set; }
 
-        [Description("Deprecated. Use Security.Certificate.Exec.Load along with Security.Certificate.Exec.Renew and Security.Certificate.Exec.OnCertificateChange")]
+        [Description("Deprecated. Use Security.Certificate.Load.Exec along with Security.Certificate.Renew.Exec and Security.Certificate.Change.Exec")]
         [DefaultValue(null)]
         [ConfigurationEntry("Security.Certificate.Exec", ConfigurationEntryScope.ServerWideOnly)]
         public string CertificateExec { get; set; }
 
         [Description("A command or executable providing a .pfx cluster certificate when invoked by RavenDB. If specified, RavenDB will use HTTPS/SSL for all network activities. The certificate path setting takes precedence over executable configuration option.")]
         [DefaultValue(null)]
-        [ConfigurationEntry("Security.Certificate.Exec.Load", ConfigurationEntryScope.ServerWideOnly)]
-        public string CertificateExecLoad { get; set; }
+        [ConfigurationEntry("Security.Certificate.Load.Exec", ConfigurationEntryScope.ServerWideOnly)]
+        public string CertificateLoadExec { get; set; }
 
         [Description("A command or executable to handle automatic renewals, providing a renewed .pfx cluster certificate. The leader node will invoke the executable once every hour and if a new certificate is received, it will be sent to the other nodes. The executable specified in Security.Certificate.Exec.OnCertificateChange will then be used to persist the certificate across the cluster.")]
         [DefaultValue(null)]
-        [ConfigurationEntry("Security.Certificate.Exec.Renew", ConfigurationEntryScope.ServerWideOnly)]
-        public string CertificateExecRenew { get; set; }
+        [ConfigurationEntry("Security.Certificate.Renew.Exec", ConfigurationEntryScope.ServerWideOnly)]
+        public string CertificateRenewExec { get; set; }
 
         [Description("A command or executable handling a change in the cluster certificate. When invoked, RavenDB will send the new cluster certificate to this executable, giving the follower nodes a way to persist the new certificate.")]
         [DefaultValue(null)]
-        [ConfigurationEntry("Security.Certificate.Exec.OnCertificateChange", ConfigurationEntryScope.ServerWideOnly)]
-        public string CertificateExecOnCertificateChange { get; set; }
+        [ConfigurationEntry("Security.Certificate.Change.Exec", ConfigurationEntryScope.ServerWideOnly)]
+        public string CertificateChangeExec { get; set; }
 
-        [Description("The command line arguments for the 'Security.Certificate.Exec' command or executable.")]
+        [Description("Deprecated. Use Security.Certificate.Load.Exec and Security.Certificate.Load.Exec.Arguments")]
         [DefaultValue(null)]
         [ConfigurationEntry("Security.Certificate.Exec.Arguments", ConfigurationEntryScope.ServerWideOnly)]
         public string CertificateExecArguments { get; set; }
 
-        [Description("The command line arguments for the 'Security.Certificate.Exec.Load' command or executable.")]
+        [Description("The command line arguments for the 'Security.Certificate.Load.Exec' command or executable.")]
         [DefaultValue(null)]
-        [ConfigurationEntry("Security.Certificate.Exec.Load.Arguments", ConfigurationEntryScope.ServerWideOnly)]
-        public string CertificateExecLoadArguments { get; set; }
+        [ConfigurationEntry("Security.Certificate.Load.Exec.Arguments", ConfigurationEntryScope.ServerWideOnly)]
+        public string CertificateLoadExecArguments { get; set; }
 
-        [Description("The command line arguments for the 'Security.Certificate.Exec.Renew' command or executable.")]
+        [Description("The command line arguments for the 'Security.Certificate.Renew.Exec' command or executable.")]
         [DefaultValue(null)]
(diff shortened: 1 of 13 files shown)


src/Raven.Server/RavenServer.cs
```diff
@@ -585,6 +516,105 @@ private async Task DoActualCertificateRefresh(CertificateHolder currentCertifica
...
+        {
+            try
+            {
+                var certHolder = ServerStore.Secrets.LoadCertificateWithExecutable(SecretProtection.CertificateRequestType.Renew, Configuration.Security.CertificateExecV2, Configuration.Security.CertificateExecArguments, ServerStore);
+
+                return certHolder.Certificate.Export(X509ContentType.Pfx); // With the private key
+            }
+            catch (Exception e)
+            {
+                throw new InvalidOperationException("Unable to refresh certificate with executable.", e);
```
ayende, line 529: Add the command here?

src/Raven.Server/RavenServer.cs
```diff
@@ -644,7 +664,10 @@ public async Task StartCertificateReplicationAsync(string base64CertWithoutPassw
                 }
 
                 if (Logger.IsOperationsEnabled)
-                    Logger.Operations("Got new certificate from Lets Encrypt! Starting certificate replication.");
+                {
+                    var source = string.IsNullOrEmpty(Configuration.Security.CertificateExecV2) ? "Let's Encrypt" : $"executable ({Configuration.Security.CertificateExecV2})";
```
ayende: Just the exec wouldn't be good enough `/bin/bash` ? 
We need the args here as well.

Do you think we can assume that there isn't anything sensitive in the args?

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -47,6 +47,11 @@ public class SecurityConfiguration : ConfigurationCategory
         [ConfigurationEntry("Security.Certificate.Exec", ConfigurationEntryScope.ServerWideOnly)]
         public string CertificateExec { get; set; }
 
+        [Description("A command or executable providing a .pfx certificate file. If specified, RavenDB will use HTTPS/SSL for all network activities. The certificate path setting takes precedence over executable configuration option.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.Certificate.Exec.V2", ConfigurationEntryScope.ServerWideOnly)]
```
ayende: I don't like this path name. It make sense, but it doesn't explain what the difference from the usual are.
It also doesn't handle users who are already using exec config, and we need to do something about it.

We can error (this is for 4.2) if this `Exec` alone is specified, but something there is required.

ayende: Maybe something like, if the `Exec` is specified, and we need to load, we'll strip the first argument and call anyway.
And error on refresh / update ?

src/Raven.Server/ServerWide/SecretProtection.cs
```diff
@@ -319,10 +321,20 @@ private static byte[] DecryptProtectedData(byte[] secret, byte[] key)
...
+            Renew,
+            CertificateChanged
+        }
+
+        public RavenServer.CertificateHolder LoadCertificateWithExecutable(CertificateRequestType type, string executable, string userArgs, ServerStore serverStore)
         {
             Process process;
 
+            
+            var args = $"{userArgs} {type}";
```
ayende: Might be better to do the `type` first ? How would that look like to the script?
Because we don't know how many args the user passes, but can say that the first arg to the script is the op type

src/Raven.Server/ServerWide/SecretProtection.cs
```diff
@@ -417,6 +429,76 @@ string GetStdError()
             };
         }
 
+        public void NotifyExecutableOfCertificateChange(string executable, string userArgs, string newCertificateBase64, ServerStore serverStore)
+        {
+            Process process;
+
+            var args = $"{userArgs} {CertificateRequestType.CertificateChanged} {newCertificateBase64}";
```
ayende: Looking at this, I think that this gets too complex for the script.
We might want to simply provide three different exec options

ayende: Also, not sure if we should pass the cert in the command line. That is actually publicly visible info.
`Get-CimInstance Win32_Process | select CommandLine`

ayende (changes requested):

src/Raven.Server/ServerWide/SecretProtection.cs
ppekrol: I personally would prefer 1 script that can handle all the cases, not X-number of different ones. What is the industry standard for this?

ayende: I don't think there _is_ such a standard. And the problem with having a single script is that you turn something fairly simple (use certbot) to complex shell script, which I think we should avoid.
A single exec also means versioning issues.

src/Raven.Server/RavenServer.cs
iftahbe: Yes we need the args too.
Theoretically yes, there can be sensitive stuff there. For example credentials to a DNS service. We can document and warn about this.
I assume most users will just pass a path or a thumbprint or friendly name.

src/Raven.Server/ServerWide/SecretProtection.cs
iftahbe: That's a problem. For example if the user does this:
Exec: powershell
Args: -NoProfile /path/to/script

Then if I put the type first, it will run 'powershell type -NoProfile /path/to/script' but type is an argument to the script and not for powershell.

src/Raven.Server/RavenServer.cs
ayende: Let's do the args. The arguments are publicly exposed anyway, and user can create a script to hide them anyway

src/Raven.Server/ServerWide/SecretProtection.cs
iftahbe: How should we pass the cert?

src/Raven.Server/ServerWide/SecretProtection.cs
ayende: Okay, that make me favor having separate config options more for this

src/Raven.Server/ServerWide/SecretProtection.cs
iftahbe: Redirect standard input?

ayende: Standard Input

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
iftahbe: Maybe you missed this but I added an exception to RavenServer.LoadCertificate which doesn't allow to use the old value. It tells them to switch to V2 and go to the documentation

src/Raven.Server/RavenServer.cs
iftahbe: The command is in the inner exception

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -42,16 +42,46 @@ public class SecurityConfiguration : ConfigurationCategory
...
 
-        [Description("A command or executable providing a .pfx certificate file. If specified, RavenDB will use HTTPS/SSL for all network activities. The certificate path setting takes precedence over executable configuration option.")]
+        [Description("Deprecated. Use Security.Certificate.Exec.Load along with Security.Certificate.Exec.Renew and Security.Certificate.Exec.OnCertificateChange")]
         [DefaultValue(null)]
         [ConfigurationEntry("Security.Certificate.Exec", ConfigurationEntryScope.ServerWideOnly)]
         public string CertificateExec { get; set; }
 
+        [Description("A command or executable providing a .pfx cluster certificate when invoked by RavenDB. If specified, RavenDB will use HTTPS/SSL for all network activities. The certificate path setting takes precedence over executable configuration option.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.Certificate.Exec.Load", ConfigurationEntryScope.ServerWideOnly)]
```
ppekrol: Let's rename to `Security.Certificate.Load.Exec`

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -42,16 +42,46 @@ public class SecurityConfiguration : ConfigurationCategory
...
+        public string CertificateExecLoad { get; set; }
+
+        [Description("A command or executable to handle automatic renewals, providing a renewed .pfx cluster certificate. The leader node will invoke the executable once every hour and if a new certificate is received, it will be sent to the other nodes. The executable specified in Security.Certificate.Exec.OnCertificateChange will then be used to persist the certificate across the cluster.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.Certificate.Exec.Renew", ConfigurationEntryScope.ServerWideOnly)]
```
ppekrol: Let's rename to `Security.Certificate.Renew.Exec`

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -42,16 +42,46 @@ public class SecurityConfiguration : ConfigurationCategory
...
+        public string CertificateExecRenew { get; set; }
+
+        [Description("A command or executable handling a change in the cluster certificate. When invoked, RavenDB will send the new cluster certificate to this executable, giving the follower nodes a way to persist the new certificate.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.Certificate.Exec.OnCertificateChange", ConfigurationEntryScope.ServerWideOnly)]
```
ppekrol: Let's rename to `Security.Certificate.Change.Exec`

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -42,16 +42,46 @@ public class SecurityConfiguration : ConfigurationCategory
...
+        public string CertificateExecOnCertificateChange { get; set; }
+
         [Description("The command line arguments for the 'Security.Certificate.Exec' command or executable.")]
         [DefaultValue(null)]
         [ConfigurationEntry("Security.Certificate.Exec.Arguments", ConfigurationEntryScope.ServerWideOnly)]
         public string CertificateExecArguments { get; set; }
 
+        [Description("The command line arguments for the 'Security.Certificate.Exec.Load' command or executable.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.Certificate.Exec.Load.Arguments", ConfigurationEntryScope.ServerWideOnly)]
```
ppekrol: Let's rename to `Security.Certificate.Load.Exec.Arguments`

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -42,16 +42,46 @@ public class SecurityConfiguration : ConfigurationCategory
...
+        public string CertificateExecLoadArguments { get; set; }
+
+        [Description("The command line arguments for the 'Security.Certificate.Exec.Renew' command or executable.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.Certificate.Exec.Renew.Arguments", ConfigurationEntryScope.ServerWideOnly)]
```
ppekrol: Let's rename to `Security.Certificate.Renew.Exec.Arguments`

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -42,16 +42,46 @@ public class SecurityConfiguration : ConfigurationCategory
...
+        public string CertificateExecRenewArguments { get; set; }
+
+        [Description("The command line arguments for the 'Security.Certificate.Exec.OnCertificateChange' command or executable.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.Certificate.Exec.OnCertificateChange.Arguments", ConfigurationEntryScope.ServerWideOnly)]
```
ppekrol: Let's rename to `Security.Certificate.OnCertificateChange.Exec.Arguments`

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -42,16 +42,46 @@ public class SecurityConfiguration : ConfigurationCategory
...
-        [Description("A command or executable providing a .pfx certificate file. If specified, RavenDB will use HTTPS/SSL for all network activities. The certificate path setting takes precedence over executable configuration option.")]
+        [Description("Deprecated. Use Security.Certificate.Exec.Load along with Security.Certificate.Exec.Renew and Security.Certificate.Exec.OnCertificateChange")]
         [DefaultValue(null)]
         [ConfigurationEntry("Security.Certificate.Exec", ConfigurationEntryScope.ServerWideOnly)]
         public string CertificateExec { get; set; }
 
+        [Description("A command or executable providing a .pfx cluster certificate when invoked by RavenDB. If specified, RavenDB will use HTTPS/SSL for all network activities. The certificate path setting takes precedence over executable configuration option.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.Certificate.Exec.Load", ConfigurationEntryScope.ServerWideOnly)]
+        public string CertificateExecLoad { get; set; }
```
ppekrol: Let's rename to `CertificateLoadExec`

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -42,16 +42,46 @@ public class SecurityConfiguration : ConfigurationCategory
...
+
+        [Description("A command or executable to handle automatic renewals, providing a renewed .pfx cluster certificate. The leader node will invoke the executable once every hour and if a new certificate is received, it will be sent to the other nodes. The executable specified in Security.Certificate.Exec.OnCertificateChange will then be used to persist the certificate across the cluster.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.Certificate.Exec.Renew", ConfigurationEntryScope.ServerWideOnly)]
+        public string CertificateExecRenew { get; set; }
```
ppekrol: Let's rename to `CertificateRenewExec`

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -42,16 +42,46 @@ public class SecurityConfiguration : ConfigurationCategory
...
+
+        [Description("A command or executable handling a change in the cluster certificate. When invoked, RavenDB will send the new cluster certificate to this executable, giving the follower nodes a way to persist the new certificate.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.Certificate.Exec.OnCertificateChange", ConfigurationEntryScope.ServerWideOnly)]
+        public string CertificateExecOnCertificateChange { get; set; }
```
ppekrol: Let's rename to `CertificateChangeExec`

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -42,16 +42,46 @@ public class SecurityConfiguration : ConfigurationCategory
...
+
         [Description("The command line arguments for the 'Security.Certificate.Exec' command or executable.")]
         [DefaultValue(null)]
         [ConfigurationEntry("Security.Certificate.Exec.Arguments", ConfigurationEntryScope.ServerWideOnly)]
         public string CertificateExecArguments { get; set; }
 
+        [Description("The command line arguments for the 'Security.Certificate.Exec.Load' command or executable.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.Certificate.Exec.Load.Arguments", ConfigurationEntryScope.ServerWideOnly)]
+        public string CertificateExecLoadArguments { get; set; }
```
ppekrol: Let's rename to `CertificateLoadExecArguments `

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -42,16 +42,46 @@ public class SecurityConfiguration : ConfigurationCategory
...
+
+        [Description("The command line arguments for the 'Security.Certificate.Exec.Renew' command or executable.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.Certificate.Exec.Renew.Arguments", ConfigurationEntryScope.ServerWideOnly)]
+        public string CertificateExecRenewArguments { get; set; }
```
ppekrol: Let's rename to `CertificateRenewExecArguments `

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -42,16 +42,46 @@ public class SecurityConfiguration : ConfigurationCategory
...
+
+        [Description("The command line arguments for the 'Security.Certificate.Exec.OnCertificateChange' command or executable.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.Certificate.Exec.OnCertificateChange.Arguments", ConfigurationEntryScope.ServerWideOnly)]
+        public string CertificateExecOnCertificateChangeArguments { get; set; }
```
ppekrol: Let's rename to `CertificateChangeExecArguments `

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -128,7 +158,7 @@ internal static void Validate(RavenConfiguration configuration)
                 {
                     if (isServerUrlHttps == false)
                         throw new InvalidOperationException(
-                            $"When the server certificate in either `{RavenConfiguration.GetKey(x => x.Security.CertificatePath)}` or `{RavenConfiguration.GetKey(x => x.Security.CertificateExec)}`  is specified, the `{RavenConfiguration.GetKey(x => x.Core.ServerUrls)}` must be using https, but was " +
+                            $"When the server certificate in either `{RavenConfiguration.GetKey(x => x.Security.CertificatePath)}` or `{RavenConfiguration.GetKey(x => x.Security.CertificateExecLoad)}` is specified, the `{RavenConfiguration.GetKey(x => x.Core.ServerUrls)}` must be using https, but was " +
```
ppekrol: https -> HTTPS

test/Tests.Infrastructure/ClusterTestBase.cs
```diff
@@ -446,22 +446,37 @@ protected async Task<RavenServer> SetupRaftClusterOnExistingServers(params Raven
             bool createNewCert = false,
             string serverCertPath = null,
             IDictionary<string, string> customSettings = null,
+            List<IDictionary<string, string>> customSettingsList = null,
```
ppekrol, line 449: Can we avoid having this?

ppekrol (approved):

PR #9042, opened 2019-04-03
RavenDB-13291 Improving counters schema update:

arekpalinski:
- making sure we delete only legacy counter related tables
- deleting counter tombstones table
- marking CollectionTableType.Counters as obsolete


ppekrol:
@arekpalinski Ping me tomorrow about this

PR #9043, opened 2019-04-03
RavenDB-13302: Due to non thread safe _activeTx.ScanOldest call, we "…

aviviadi:
…refreshing" OldestActiveTransaction member during flush

### DIFF
--- a/src/Voron/Impl/Journal/WriteAheadJournal.cs
+++ b/src/Voron/Impl/Journal/WriteAheadJournal.cs
@@ -565,7 +565,6 @@ namespace Voron.Impl.Journal
                     if (_waj._env.Disposed)
                         return;
 
-
                     var jrnls = GetJournalSnapshots();
 
                     if (jrnls.Count == 0)
@@ -581,6 +580,8 @@ namespace Voron.Impl.Journal
 
                     long lastFlushedTransactionId = -1;
 
+                    // RavenDB-13302: we need to force a re-check this before we make decisions here
+                    _waj._env.ActiveTransactions.RecheckOldestTransaction();
                     long oldestActiveTransaction = _waj._env.ActiveTransactions.OldestTransaction;
 
                     foreach (var journalFile in jrnls)
--- a/src/Voron/Impl/Journal/WriteAheadJournal.cs
+++ b/src/Voron/Impl/Journal/WriteAheadJournal.cs
@@ -581,7 +581,7 @@ namespace Voron.Impl.Journal
                     long lastFlushedTransactionId = -1;
 
                     // RavenDB-13302: we need to force a re-check this before we make decisions here
-                    _waj._env.ActiveTransactions.RecheckOldestTransaction();
+                    _waj._env.ActiveTransactions.ForceRecheckingOldestTransactionByFlusherThread();
                     long oldestActiveTransaction = _waj._env.ActiveTransactions.OldestTransaction;
 
                     foreach (var journalFile in jrnls)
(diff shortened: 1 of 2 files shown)


ayende (approved):

src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -581,6 +580,8 @@ public void ApplyLogsToDataFile(CancellationToken token, TimeSpan timeToWait)
 
                     long lastFlushedTransactionId = -1;
 
+                    // RavenDB-13302: we need to force a re-check this before we make decisions here
+                    _waj._env.ActiveTransactions.RecheckOldestTransaction();
```
arekpalinski: Why we don't need to call that before other usages of `ActiveTransactions.OldestTransaction` ?

ayende: This happens after write transactions, and will do the fixup for everything, basically.
So we only need to have this in one location

arekpalinski: Can you elaborate on "This happens after write transactions"? Why is that the case?

ayende: This is the flusher, and it is going to run after write transactions, so that will ensure that this gets update eventually.
The actual value we have here isn't really that important, the only thing that matter here is that we make some progress over time

arekpalinski: Okay, so this way we prevent from getting stuck. Makes sense. Thanks for clarification.

arekpalinski: Okay, so this way we prevent from getting stuck. Makes sense. Thanks for clarification.

src/Voron/Util/ActiveTransactions.cs
```diff
@@ -30,6 +30,22 @@ public void Add(LowLevelTransaction tx)
             }
         }
 
+        public void RecheckOldestTransaction()
```
arekpalinski: I wonder if we can have a more specific name for this guy.  Maybe `ForceRecheckingOldestTransactionByFlusherThread` ?
I'd like to have something more explicit here.

ayende: Fine by me

redknightlois:
LGTM. On a side note, this is an effect of a race condition; the question is now that we know this one can happen, we probably should explicitely mark (even if with comments) which variables on the execution codepath are shared between threads. The rationale is that while this code doesnt change much, we have to be extra careful when modifying code on this general area. Having an audit of the shared state can diminish the probability of this kind of race conditions happening again.

ayende:
The `_activeTxs` is _literally_ called `RacyConcucrrentBag`

redknightlois:
@ayende Yes, but the audit should be on the chain of methods that rely on that shared state. The shared state is secondary.

redknightlois:
We have a racy bag we know and we need to be racy (for performance reasons)...
Now we have the state which is 'shared' and 'racy'
And we have methods `A` and `B` that takes decisions from that `racy` state.
Therefore `A` and `B` are tainted, whatever lever they pull, it has to be considered as dangerous.
If `C` uses either one of those, it has to have mitigation steps in place to deal with the `racyness`  properties of `A` and `B` which are transitive. If it does not, it has to be tainted too.

ppekrol:
We will create a 4.1 PR for this and do a 4.1 -> 4.2 merge

aviviadi:
cont at : https://github.com/ravendb/ravendb/pull/9050

PR #9044, opened 2019-04-03
RavenDB-13185 global handling of CORS headers

ml054:
@arekpalinski @ppekrol @ayende 
Since I'm playing with security here, please take special attention during review. 

Feel free to ping me if something is unclear.

### DIFF
--- a/src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
@@ -220,11 +220,9 @@ namespace Raven.Server.Documents.Handlers
             }
         }
 
-        [RavenAction("/databases/*/subscriptions/connection-details", "GET", AuthorizationStatus.ValidUser)]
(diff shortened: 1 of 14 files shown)


src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
```diff
@@ -220,11 +220,9 @@ public Task GetSubscriptionState()
             }
         }
 
-        [RavenAction("/databases/*/subscriptions/connection-details", "GET", AuthorizationStatus.ValidUser)]
+        [RavenAction("/databases/*/subscriptions/connection-details", "GET", AuthorizationStatus.ValidUser, CorsMode = CorsMode.Cluster)]
```
ayende, line 223: When are we accessing this across the cluster?

ml054: from ongoing tasks in studio - we always query responsible node directly.

ppekrol (approved):

PR #9045, opened 2019-04-03
RavenDB-13276 creating subscription with long name makes server unres…

efratshenhar:
…ponsive

### DIFF
--- a/src/Raven.Server/ServerWide/Commands/Subscriptions/PutSubscriptionCommand.cs
+++ b/src/Raven.Server/ServerWide/Commands/Subscriptions/PutSubscriptionCommand.cs
@@ -1,8 +1,10 @@
 ﻿using System;
 using Raven.Client;
 using Raven.Client.Documents.Subscriptions;
+using Raven.Client.Exceptions.Documents.Subscriptions;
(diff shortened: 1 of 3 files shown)


src/Raven.Server/ServerWide/Commands/Subscriptions/PutSubscriptionCommand.cs
```diff
@@ -52,6 +54,8 @@ public override unsafe void Execute(TransactionOperationContext context, Table i
             while (tryToSetName)
             {
                 var subscriptionItemName = SubscriptionState.GenerateSubscriptionItemKeyName(DatabaseName, SubscriptionName);
+                if (subscriptionItemName.Length > AbstractPager.MaxKeySize)
```
ppekrol: I think it would be better to use the same constraint as for documents. Which is 512. Check here: https://github.com/ravendb/ravendb/pull/9014

That should be applied to initial subscription name only, not the generated one.

ayende: The generated subscription name should never be too long, I think

ppekrol (approved):

PR #9046, opened 2019-04-03
RavenDB-13278 : revision configuration options for collections should be case insensitive

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-13278

### DIFF
--- a/src/Raven.Server/Documents/Handlers/RevisionsHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/RevisionsHandler.cs
@@ -12,6 +12,7 @@ using System.Threading.Tasks;
 using Microsoft.Extensions.Primitives;
 using Raven.Client;
 using Raven.Client.Documents.Commands;
+using Raven.Client.Documents.Operations.Revisions;
 using Raven.Client.Exceptions.Documents.Revisions;
 using Raven.Server.Documents.Revisions;
 using Raven.Server.Json;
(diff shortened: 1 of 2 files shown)


grisha-kotler (approved):

src/Raven.Server/Documents/Handlers/RevisionsHandler.cs
```diff
@@ -61,7 +62,29 @@ public Task GetRevisionsConfig()
         [RavenAction("/databases/*/admin/revisions/config", "POST", AuthorizationStatus.DatabaseAdmin)]
         public async Task ConfigRevisions()
         {
-            await DatabaseConfigurations(ServerStore.ModifyDatabaseRevisions, "read-revisions-config");
+            await DatabaseConfigurations(
+                ServerStore.ModifyDatabaseRevisions,
+                "read-revisions-config",
+                beforeSetupConfiguration: (string name, ref BlittableJsonReaderObject configuration, JsonOperationContext context) =>
```
ayende, line 68: I'm assuming that this happens before we go to the cluster state machine, right?

ayende: What happens if we have two separate requests like this? Is this a whole sale replacement over time, or individual ops?

ppekrol: Full replacement: https://github.com/ravendb/ravendb/blob/v4.1/src/Raven.Server/ServerWide/Commands/EditRevisionsConfigurationCommand.cs#L28

aviv86: yes,  this happens before we go to the cluster state machine

ayende (approved):

PR #9049, opened 2019-04-03
RavenDB-13181 Counters replication discrepancies

karmeli87:
The underlying issue is that we resotred not in the insertion order in which the document, revisions, coutners, etc. were putted in the original database.
In the restored database this could cause that a document with higher change vector will have a lower local etag a therefore will skip the replication of doucments with lower change vector.

- Remove keep original change vector
- Insert the counters by insertion order

### DIFF
--- a/src/Raven.Client/Documents/Operations/Counters/CounterBatch.cs
+++ b/src/Raven.Client/Documents/Operations/Counters/CounterBatch.cs
@@ -88,6 +88,7 @@ namespace Raven.Client.Documents.Operations.Counters
         public long Delta;
 
         internal string ChangeVector;
+        public string DocumentId;
 
         public static CounterOperation Parse(BlittableJsonReaderObject input)
         {
--- a/src/Raven.Client/Documents/Operations/Counters/CounterBatch.cs
+++ b/src/Raven.Client/Documents/Operations/Counters/CounterBatch.cs
@@ -89,7 +89,6 @@ namespace Raven.Client.Documents.Operations.Counters
         public long Delta;
 
         internal string ChangeVector;
-        [JsonProperty]
         internal string DocumentId;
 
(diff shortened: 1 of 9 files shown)


src/Raven.Server/Documents/Handlers/CountersHandler.cs
```diff
@@ -39,12 +39,13 @@ public class ExecuteCounterBatchCommand : TransactionOperationsMerger.MergedTran
             private readonly bool _replyWithAllNodesValues;
             private readonly bool _fromEtl;
             private readonly bool _fromSmuggler;
-            private readonly Dictionary<string, List<CounterOperation>> _dictionary;
+            private readonly List<CounterOperation> _list;
+            //private readonly Dictionary<string, List<CounterOperation>> _dictionary;
```
grisha-kotler: can be removed

grisha-kotler (changes requested):

src/Raven.Client/Documents/Operations/Counters/CounterBatch.cs
```diff
@@ -88,6 +88,7 @@ public class CounterOperation
         public long Delta;
 
         internal string ChangeVector;
+        public string DocumentId;
```
ayende: Shouldn't this be internal?

karmeli87: It fails on the TxReplay, since serializing this ignore the internal.

ayende: Can we fix the serialization, instead?
This is sitting on the client, and I don't want it exposed to users

aviv86 (approved):

ppekrol (approved):

ayende (approved):

PR #9054, opened 2019-04-04
RavenDB-13338: Add SkipAddingIdFieldToDynamicObject to DocumentConven…

garayx:
…tions

### DIFF
--- a/src/Raven.Client/Documents/Conventions/DocumentConventions.cs
+++ b/src/Raven.Client/Documents/Conventions/DocumentConventions.cs
@@ -129,6 +129,8 @@ namespace Raven.Client.Documents.Conventions
             MaxHttpCacheSize = new Size(httpCacheSizeInMb, SizeUnit.Megabytes);
 
             OperationStatusFetchMode = OperationStatusFetchMode.ChangesApi;
+
+            SkipAddingIdFieldToDynamicObject = false;
         }
 
         private bool _frozen;
@@ -154,6 +156,7 @@ namespace Raven.Client.Documents.Conventions
         private Func<string, BlittableJsonReaderObject, string> _findClrType;
         private bool _useOptimisticConcurrency;
         private bool _throwIfQueryPageSizeIsNotSet;
+        private bool _skipAddingIdFieldToDynamicObject;
         private int _maxNumberOfRequestsPerSession;
         private Action<JsonSerializer> _customizeJsonSerializer;
         private TimeSpan? _requestTimeout;
@@ -243,6 +246,21 @@ namespace Raven.Client.Documents.Conventions
             {
                 AssertNotFrozen();
                 _customizeJsonSerializer = value;
+            }
+        }
+
+        /// <summary>
+        ///     If 'true' Id field will not be added to dynamic objects.
+        ///     If 'false' Id field will be added to dynamic objects.
+        ///     Default value is 'false'
+        /// </summary>
+        public bool SkipAddingIdFieldToDynamicObject
+        {
+            get => _skipAddingIdFieldToDynamicObject;
+            set
+            {
+                AssertNotFrozen();
+                _skipAddingIdFieldToDynamicObject = value;
             }
         }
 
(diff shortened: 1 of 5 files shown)


src/Raven.Client/Documents/Conventions/DocumentConventions.cs
```diff
@@ -246,6 +249,21 @@ public Action<JsonSerializer> CustomizeJsonSerializer
             }
         }
 
+        /// <summary>
+        ///     If 'true' Id field will not be added to dynamic objects.
+        ///     If 'false' Id field will be added to dynamic objects.
```
ayende: Can remove this line

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
```diff
@@ -246,6 +249,21 @@ public Action<JsonSerializer> CustomizeJsonSerializer
             }
         }
 
+        /// <summary>
+        ///     If 'true' Id field will not be added to dynamic objects.
```
ayende:
```suggestion
        ///     By default, the field 'Id' field will be added to dynamic objects, this allows to disable this behavior.
```

That said, let's avoid the double negative here.
Make the convention: "AddIdFieldToDynamicObjects" and default to true

ayende (changes requested):

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
garayx: done

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
garayx: done

ayende (approved):

src/Raven.Client/Documents/Session/EntityToBlittable.cs
```diff
@@ -283,7 +285,26 @@ private static bool TryRemoveIdentityProperty(BlittableJsonReaderObject document
...
+                {
+                    try
+                    {
+                        if (document.Modifications == null)
+                            document.Modifications = new DynamicJsonValue(document);
+
+                        document.Modifications.Remove("Id");
+                        return true;
+                    }
+                    catch (RuntimeBinderException)
```
ppekrol: No need to catch RuntimeBinderException. We are not performing any operations on 'dynamic'

src/Raven.Client/Documents/Session/EntityToBlittable.cs
```diff
@@ -283,7 +285,26 @@ private static bool TryRemoveIdentityProperty(BlittableJsonReaderObject document
         {
             var identityProperty = conventions.GetIdentityProperty(entityType);
             if (identityProperty == null)
+            {
+                if (conventions.AddIdFieldToDynamicObjects && entityType.GetInterfaces().Contains(typeof(IDynamicMetaObjectProvider)))
```
ppekrol: We need to cache this, it is too expensive. Probably better to pass the entity from the calling method and just to `entity is IDynamicMetaObjectProvider`

ppekrol (changes requested):

src/Raven.Client/Documents/Session/EntityToBlittable.cs
ayende: Yes, passing it there is much better

garayx: done

src/Raven.Client/Documents/Session/EntityToBlittable.cs
garayx: done

test/FastTests/Issues/RavenDB_13338.cs
```diff
@@ -40,5 +40,53 @@ public void SkipAddingIdFieldToDynamicObjectShouldWork()
...
+                    Assert.Equal("1", loadJObject.User.ToString());
+                    Assert.Equal(jObjectId, loadJObject.Id.ToString());
+                }
+
+                using (var commands = store.Commands())
+                {
+                    const string idField = "Id";
+
+                    var bJsonExpandoObject = commands.Get(expandoObjectId);
+                    Assert.Equal(false, bJsonExpandoObject.Contains(idField));
```
ppekrol: 1. Use `ContainsKey` method
2. `Assert.False(bJsonExpandoObject.ContainsKey(idField))`

garayx: done

ppekrol (approved):

PR #9057, opened 2019-04-04
RavenDB-13226 cluster observer should skip check index progress if it…

karmeli87:
…s state is 'Idle'


ayende (approved):

karmeli87:
Apparently Idle index is still indexing

ayende:
The fact that it is still indexing shouldn't matter, should it? 
We still shouldn't wait for it for rehab purposes, I think.

ppekrol:
Depends, Idle index is still indexing as it would be Normal one. It is just a flag that helps to indentify them and clean them up by the indexing cleanup mechanism. From one side we can say that it is idle so we can ignore it, from other if we will ignore it and move the node out of rehab then if someone that queries it (and converts it to Normal) might get way different results depending on the node that was hit.

PR #9060, opened 2019-04-04
RavenDB-13215: If db being deleted from group, display it as deleted …

garayx:
…only.

### DIFF
--- a/src/Raven.Studio/typescript/viewmodels/resources/manageDatabaseGroup.ts
+++ b/src/Raven.Studio/typescript/viewmodels/resources/manageDatabaseGroup.ts
@@ -206,8 +206,14 @@ class manageDatabaseGroup extends viewModelBase {
             this.nodes([]);
             this.clearNodesList(false);
         }
-        
-        this.updateNodes(incomingDbInfo.nodes());
+
+        if (incomingDbInfo.deletionInProgress().length > 0) {
+            let nodeToDisplay = incomingDbInfo.nodes().filter(x => incomingDbInfo.deletionInProgress().indexOf(x.tag()) < 0);
+            this.updateNodes(nodeToDisplay);
+        } else {
+            this.updateNodes(incomingDbInfo.nodes());
+        }
+
         this.deletionInProgress(incomingDbInfo.deletionInProgress());
         this.isEncrypted(incomingDbInfo.isEncrypted());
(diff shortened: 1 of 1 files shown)


src/Raven.Studio/typescript/viewmodels/resources/manageDatabaseGroup.ts
```diff
@@ -206,8 +206,14 @@ class manageDatabaseGroup extends viewModelBase {
             this.nodes([]);
             this.clearNodesList(false);
         }
-        
-        this.updateNodes(incomingDbInfo.nodes());
+
+        if (incomingDbInfo.deletionInProgress().length > 0) {
+            let nodeToDisplay = incomingDbInfo.nodes().filter(x => incomingDbInfo.deletionInProgress().indexOf(x.tag()) < 0);
```
grisha-kotler: const

grisha-kotler (changes requested):

src/Raven.Studio/typescript/viewmodels/resources/manageDatabaseGroup.ts
grisha-kotler: *nodesToDisplay

grisha-kotler (changes requested):

src/Raven.Studio/typescript/viewmodels/resources/manageDatabaseGroup.ts
```diff
@@ -206,8 +206,14 @@ class manageDatabaseGroup extends viewModelBase {
             this.nodes([]);
             this.clearNodesList(false);
         }
-        
-        this.updateNodes(incomingDbInfo.nodes());
+
+        if (incomingDbInfo.deletionInProgress().length > 0) {
```
grisha-kotler: const deletionInProgress = incomingDbInfo.deletionInProgress();

grisha-kotler (changes requested):

ml054 (changes requested):
we also have RavenDB-13089 which targets for 4.1, so the question is if we should PR this against 4.1 as well? Additionally I think this should be server side issue - not studio side - since currently you are filtering some state you get from server. Why can't we filter that on server side?

garayx:
will open a new PR with server side fix

PR #9075, opened 2019-04-06
RavenDB-13140 Using realpath() function

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-13140


src/librvnpal/src/posix/sync.c
```diff
@@ -69,77 +69,31 @@ _sync_directory_for_internal(char *dir_path, int32_t *detailed_error_code)
...
-    int32_t rc;
 
-    int32_t steps = 10;
 
-    while (1)
+int32_t
+_sync_directory_maybe_symblink(char *dir_path, int32_t *detailed_error_code)
+{
+    int32_t rc;
+    char* real_path = realpath(dir_path, NULL);
```
haludi, line 78: > The resolved_path == NULL
> feature, not standardized in POSIX.1-2001, but standardized in
> POSIX.1-2008, allows this design problem to be avoided.
http://man7.org/linux/man-pages/man3/realpath.3.html

Is it ok to use a feature that not standardized in POSIX.1-2001 but do in POSIX.1-2008
The other solution is to use `PATH_MAX` or `pathconf` that are problematic too in some file systems

ayende: Yes, that is fine. We are going to learn about any incompatibilities very quickly, after all

ayende (approved):

src/librvnpal/src/posix/sync.c
haludi: Maybe, if the changes are approved we should add commit with recompile before merge

ppekrol: Yes, please recompile and push to this PR the changes. cc @aviviadi @haludi

PR #9081, opened 2019-04-07
RavenDB-9813 : test replication of expired and deleted revisions

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-9813

### DIFF
--- a/src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
+++ b/src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
@@ -482,9 +482,18 @@ namespace Raven.Server.Documents.Revisions
                 configuration.MinimumRevisionAgeToKeep.HasValue == false)
                 return;
 
-            var numberOfRevisionsToDelete = revisionsCount - configuration.MinimumRevisionsToKeep ?? 0;
-            if (numberOfRevisionsToDelete <= 0)
-                return;
+            long numberOfRevisionsToDelete;
+            if (configuration.MinimumRevisionsToKeep.HasValue)
+            {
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -482,9 +482,18 @@ private BlittableJsonReaderObject RecreateCountersIfNeeded(DocumentsOperationCon
...
+            if (configuration.MinimumRevisionsToKeep.HasValue)
+            {
+                numberOfRevisionsToDelete = revisionsCount - configuration.MinimumRevisionsToKeep.Value;
+                if (numberOfRevisionsToDelete <= 0)
+                    return;
+            }
+            else
+            {
+                // delete all revisions which age has passed
+                numberOfRevisionsToDelete = long.MaxValue;
```
ppekrol, line 495: I have some doubts about that. If both of the settings `MinimumRevisionsToKeep` and `MinimumRevisionAgeToKeep` are `null` then we want to keep unlimited amount of revisions. What will happen now?

grisha-kotler: If both are null it won't reach this point.

ppekrol: Ok. Makes sense. I see the code now. The descriptions in the Studio match all our behaviors?

grisha-kotler: Yes

PR #9083, opened 2019-04-08
RavenDB-13257 indexing performance  - duplicate items when replacemen…

ml054:
…t is applied

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -1106,6 +1106,20 @@ namespace Raven.Server.Documents.Indexes
                                         var originalName = Name.Replace(Constants.Documents.Indexing.SideBySideIndexNamePrefix, string.Empty, StringComparison.InvariantCultureIgnoreCase);
                                         _isReplacing = true;
 
+                                        if (batchCompleted)
+                                        {
+                                            // this side-by-side index will be replaced in a second, notify about indexing success
+                                            // so we know that indexing batch is no longer in progress
+                                            DocumentDatabase.Changes.RaiseNotifications(new IndexChange
+                                            {
+                                                Name = Name,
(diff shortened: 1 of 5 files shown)


ppekrol (approved):

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1106,6 +1106,20 @@ protected void ExecuteIndexing()
...
+                                            // this side-by-side index will be replaced in a second, notify about indexing success
+                                            // so we know that indexing batch is no longer in progress
+                                            DocumentDatabase.Changes.RaiseNotifications(new IndexChange
+                                            {
+                                                Name = Name,
+                                                Type = IndexChangeTypes.BatchCompleted
+                                            });
+                                            
+                                            var batchCompletedAction = DocumentDatabase.IndexStore.IndexBatchCompleted;
+                                            batchCompletedAction?.Invoke((Name, didWork));
```
arekpalinski: Let's extract this piece of code to a method. We have exactly the same a few lines code below.

ml054: done

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1238,6 +1228,22 @@ protected void ExecuteIndexing()
             }
         }
 
+        private void NotifyAboutCompletedBatch(bool didWork)
+        {
+            // this side-by-side index will be replaced in a second, notify about indexing success
+            // so we know that indexing batch is no longer in progress
```
arekpalinski: This comment should not be here but when `NotifyAboutCompletedBatch` it's called

ml054: done - sorry

arekpalinski (approved):

PR #9111, opened 2019-04-10
RavenDB-13265 Fixing UnauthorizedAccessException during disposal or d…

arekpalinski:
…eletion of a database or an index. The issue was that LowLevelTransaction.EnsurePagerState did not add a pager state to the collection which we use during tx dispose to release pager states. In result we were still holding a reference to Voron data file. Note it required a data pager state to be changed in the background during EnsurePagerState (journals flushing) so it was not very likely to happen.

### DIFF
--- a/src/Voron/Impl/LowLevelTransaction.cs
+++ b/src/Voron/Impl/LowLevelTransaction.cs
@@ -31,6 +31,7 @@ namespace Voron.Impl
         private readonly ByteStringContext _allocator;
         private readonly PageLocator _pageLocator;
         private bool _disposeAllocator;
+        private TestingStuff _forTestingPurposes;
 
         private Tree _root;
         public Tree RootObjects => _root;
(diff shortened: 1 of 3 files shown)


src/Voron/Impl/LowLevelTransaction.cs
```diff
@@ -1132,14 +1133,19 @@ public void EnsurePagerStateReference(PagerState state)
...
             {
                 _lastState = state;
                 return;
             }
 
             state = state.CurrentPager.GetPagerStateAndAddRefAtomically(); // state might hold released pagerState, and we want to add ref to the current (i.e. data file was re-allocated and a new state is now available). RavenDB-6950
+
             _lastState = state;
+
+            _pagerStates.Add(state);
```
arekpalinski, line 1148: Fix is here - we didn't add `state` returned by `GetPagerStateAndAddRefAtomically()` which takes into account that we have a new state

ayende (approved):

PR #9113, opened 2019-04-10
RavenDB-13212 Concurrency issue when `TryPush` and `TryAcquired` was fixed

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-13212

### DIFF
--- a/src/Sparrow/Collections/RingBuffer.cs
+++ b/src/Sparrow/Collections/RingBuffer.cs
@@ -50,25 +50,23 @@ namespace Sparrow.Collections
         public bool TryPush(ref T item)
         {
             // We will check if we potentially have the ability to push 1 more item. 
-            int cidx = Volatile.Read(ref _currentIdx);
             int sidx = Volatile.Read(ref _startIdx);
 
+            //_currentIdx modified only by producer and there for no need of Volatile.Read
             // Check if we have empty spaces in the buffer.
-            if (cidx - sidx >= _size)
+            if (_currentIdx - sidx >= _size)
                 return false; // No space to do anything
 
-            int ticker = Interlocked.Increment(ref _currentIdx) - 1;
-
             // Assign the value
-            ref var cl = ref _buffer[ticker & _mask];
-
+            ref var cl = ref _buffer[_currentIdx & _mask];
             // We assign the item
             cl.Item = item;
-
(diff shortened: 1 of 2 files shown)


src/Sparrow/Collections/RingBuffer.cs
```diff
@@ -89,6 +89,13 @@ public bool TryAcquireSingle(out RingItem<T> item)
             }
 
             item = _buffer[sidx & _mask];
+            if (item.Item == null)
```
ayende: Why? That is explicitly something that can happen, this is not an error

src/Sparrow/Collections/RingBuffer.cs
```diff
@@ -50,25 +50,23 @@ public SingleConsumerRingBuffer(int size)
...
                 return false; // No space to do anything
 
-            int ticker = Interlocked.Increment(ref _currentIdx) - 1;
-
             // Assign the value
-            ref var cl = ref _buffer[ticker & _mask];
-
+            ref var cl = ref _buffer[_currentIdx & _mask];
             // We assign the item
             cl.Item = item;
```
haludi, line 70: The write and the read to the array are not by `Volatile` except for the writing to `RingItem<T>.IsReady.
Is it a problem?

ayende: @redknightlois Can you take a look?

redknightlois:
Volatile is needed. This is single consumer / multi producer buffer. The fact that we have a single producer on that case, is not true everywhere

redknightlois:
That would cause on very rare occasions to lose a single push because it will get overwritten during 2 two trypush that happens to read the same idx. If I understand correctly the issue my hypothesis is that we are checking and we allowing to move forward at a risky position while we shouldn't

redknightlois:
After a second look I stand corrected, it is not a rare occurrence, it is very likely on high contention. This is definitely NOT a solution, to solve it look into IsReady interaction with TryAdd

ayende:
@redknightlois any reason not to merge this?

redknightlois:
Yes, it doesnt fix the race condition. It creates a different one far more difficult to diagnose where messages are going to get overwritten (therefore will cause actions to get lost).

ayende (approved):

PR #9116, opened 2019-04-10
RavenDB-13375 Adding logs for all command apply

talweiss1982:


src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1051,42 +1175,79 @@ private void UpdateDatabase(TransactionOperationContext context, string type, Bl
...
+            }
+            finally
+            {
+                LogCommand(type, index, exception, updateCommand?.AdditionalDebugInformation(exception));
                 NotifyDatabaseAboutChanged(context, databaseName, index, type, DatabasesLandlord.ClusterDatabaseChangeType.RecordChanged);
             }
         }
-        private void UpdateDatabaseRecordEtagForBackup(TransactionOperationContext context, string type, BlittableJsonReaderObject cmd, long index, Leader leader, ServerStore serverStore)
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
ayende, line 1190: We assume that `info` is usually off, if you want to aggressively inline this, you can, but then the contents of the `if` need to be in another method, to reduce the amount of code that is inlined

talweiss1982: should that method need to be marked for inline too?

ayende (approved):

PR #9120, opened 2019-04-11
RavenDB-13287 fix runtime config

gregolsky:


ayende (commented):
Where is the fix here?

ayende (commented):
You are reverting the condition, as I understand it, but why didn't it work before?

gregolsky:
It was passing `AnyCPU` instead of blank. I think made a mistake when checking the build results.

PR #9131, opened 2019-04-14
RavenDB-13046 Null result for swap definition fixed

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-13046

### DIFF
--- a/src/Sparrow/LowMemory/CheckPageFileOnHdd.cs
+++ b/src/Sparrow/LowMemory/CheckPageFileOnHdd.cs
@@ -397,9 +397,7 @@ namespace Sparrow.LowMemory
                         continue; // we do not check swap file, only partitions
 
                     // remove numbers at end of string (i.e.: /dev/sda5 ==> sda)
-                    var disk = swaps[i].DeviceName != null
-                        ? _regExRemoveNumbers.Replace(swaps[i].DeviceName, "").Replace("/dev/", "")
-                        : _regExRemoveNumbers.Replace("/dev/", "");
+                    var disk = _regExRemoveNumbers.Replace(swaps[i].DeviceName, "").Replace("/dev/", "");
                     var filename = $"/sys/block/{disk}/queue/rotational";
                     var isHdd = KernelVirtualFileSystemUtils.ReadNumberFromFile(filename);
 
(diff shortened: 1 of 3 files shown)


src/Sparrow/LowMemory/CheckPageFileOnHdd.cs
```diff
@@ -397,9 +397,7 @@ public static string PosixIsSwappingOnHddInsteadOfSsd()
                         continue; // we do not check swap file, only partitions
 
                     // remove numbers at end of string (i.e.: /dev/sda5 ==> sda)
-                    var disk = swaps[i].DeviceName != null
-                        ? _regExRemoveNumbers.Replace(swaps[i].DeviceName, "").Replace("/dev/", "")
-                        : _regExRemoveNumbers.Replace("/dev/", "");
+                    var disk = _regExRemoveNumbers.Replace(swaps[i].DeviceName, "").Replace("/dev/", "");
```
aviviadi, line 400: Do we have here swaps[i].DeviceName null error handling?

test/FastTests/Sparrow/FileSystemUtilsTest.cs
```diff
@@ -0,0 +1,33 @@
...
+            var swap = KernelVirtualFileSystemUtils.ReadSwapInformationFromSwapsFile();
+            Assert.All(swap, s => Assert.NotNull(s.DeviceName));
+        }
+        
+        [LinuxFact]
+        public void WriteAndReadPageUsingCryptoPager_WhenTwoSwapDefined_AllResultShouldBeNotNull()
+        {
+            var fileName = GetTempFileName();
+            
+            string[] lines =
```
aviviadi, line 21: Please add also more `lines` to test:
```
{
                "Filename				Type		Size	Used	Priority"
}
```
and
```
{
                "Filename				Type		Size	Used	Priority",
                ""
}

and 
```
{
                "Filename				Type		Size	Used	Priority"
                "/dev/sda                               partition	999420	56320	-2", 
}
```
and also with filename

src/Sparrow/LowMemory/CheckPageFileOnHdd.cs
haludi: We can't get null at `DeviceName` as it implemented now and we shouldn't. 
I added a test to check it.
I think if we still want to check this it should be by `Debug.Assert`

aviviadi (approved):

src/Sparrow/LowMemory/CheckPageFileOnHdd.cs
aviviadi: PR Approved

PR #9132, opened 2019-04-14
RavenDB-13013 Using EnablePrefetching

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-13013

### DIFF
--- a/src/Voron/Impl/Paging/RvnMemoryMapPager.cs
+++ b/src/Voron/Impl/Paging/RvnMemoryMapPager.cs
@@ -85,11 +85,19 @@ namespace Voron.Impl.Paging
             SetPagerState(pager);
         }
 
+        protected override bool CanPrefetchQuery()
+        {
+            if (SysInfo.CanPrefetch == false || CanPrefetchAhead == false)
+                return false; // not supported
+
+            return true;
+        }
+        
         public override byte* AcquirePagePointer(IPagerLevelTransactionState tx, long pageNumber, PagerState pagerState = null)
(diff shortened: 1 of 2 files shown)


src/Voron/Impl/Paging/RvnMemoryMapPager.cs
```diff
@@ -89,7 +89,7 @@ public RvnMemoryMapPager(StorageEnvironmentOptions options, VoronPathSetting fil
         {
             var state = pagerState ?? _pagerState;
 
-            if (SysInfo.CanPrefetch)
+            if (Options.EnablePrefetching && SysInfo.CanPrefetch)
```
ppekrol: I think only one is needed?

aviviadi: @ppekrol you should not prefetch in non prefetchable environment even if EnablePrefetching is enabled.  So I think the above change is fine.

aviviadi: ok.. @ppekrol - you are right.
@haludi -  `override protected virtual bool CanPrefetchQuery()`
Implement it using both `SysInfo.CanPrefetch` and `_canPrefetchAhead` (make sure it _can_ be overriden by c'tor, i.e. - even if we are on Win10 and we are able to prefetch, but 32 bit pager wants this feature off)

ayende (approved):

ppekrol (approved):

PR #9134, opened 2019-04-14
RavenDB-13284 fixing an issue where replication could get into corrup…

talweiss1982:
…t in memory state due to modification of the definition after it was used by a collection


ayende (approved):

ayende (commented):
Just to verify, we also check the types in the `Equals`, right?

ppekrol:
All of them override `IsEqualTo` which is used in `Equals`

PR #9135, opened 2019-04-14
RavenDB-13334 In Test Subscription view change Results limit headline

efratshenhar:


src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
```diff
@@ -147,6 +147,77 @@ public class SubscriptionDocumentsFetcher
             }
         }
 
+        public IEnumerable<(Document Doc, Exception Exception)> GetDocumentsToSendWithTimeLimit(DocumentsOperationContext docsContext,
```
ppekrol: Can we add the timeLimit to the `GetDataToSend` but when it is `null` then not apply the logic? This will avoid a lot of code redundancy I think?

src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
```diff
@@ -147,6 +147,77 @@ public class SubscriptionDocumentsFetcher
...
+                                    };
+
+                                    yield return (projection, null);
+                                }
+                            }
+                            if (++numberOfDocs >= _maxBatchSize)
+                                yield break;
+                        }
+                    }
+                if (sp.Elapsed >= TimeSpan.FromSeconds(time))
```
ayende: This method should accept a TimeSpan, not a `time` value

ayende (changes requested):
I don't like this change.
This should _not_ be in the `GetDocumentsToSendWithTimeLimit`, it should be in the subscription handler itself, where you'll do the check.
Remember, this is a yielding method, so you can do the break in the calling code

maximburyak:
@efratshenhar , I would like us to also be able to "page forward" if we did not find any documents, by bumping up the etag to start from while also starting a new transaction.

src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
```diff
@@ -40,15 +41,18 @@ public async Task Try()
                 if (sub.Collection == null)
                     throw new ArgumentException("Collection must be specified");
 
+                const int maxPageSize = 1024;
                 var pageSize = GetIntValueQueryString("pageSize") ?? 1;
+                if (pageSize > maxPageSize)
```
ayende, line 46: Why have this limit? We already have a time limit that also applies, no?

ppekrol: I do not want users to request 1M results here and kill the Studio. This is used only for Studio/Subscription/Test functionality and the 1024 seams more than fine?

ppekrol (approved):

src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
ayende: Why bother? This isn't happening on the studio, it happens on the server, and limited by the time.
It is actually going to be pretty important to have large numbers there because we may need to scan a lot of documents to find a useful value if there is a where there.

ppekrol: This is not related to the number of documents to scan (it is int.Max). This is a number of results to return to the user in Studio and display in the grid there. We can remove this, but I think 1024 is more than enough?

ayende: Oh, number of docs to return is fine, then.

PR #9139, opened 2019-04-14
RavenDB-8775 - local license update

grisha-kotler:


src/Raven.Server/Commercial/LicenseHelper.cs
```diff
@@ -0,0 +1,230 @@
...
+        }
+
+        public void UpdateLocalLicense(License newLicense, RSAParameters rsaParameters)
+        {
+            try
+            {
+                if (RavenConfiguration.EnvironmentVariableLicenseString != null)
+                {
+                    // update the environment variable string value
+                    UpdateEnvironmentVariableLicenseString(newLicense, rsaParameters);
```
ayende: This usually won't be persistent, though

src/Raven.Server/Commercial/LicenseHelper.cs
```diff
@@ -0,0 +1,230 @@
...
+                    return;
+                }
+
+                // update the license string in settings.json
+                if (TryUpdatingSettingsJson(newLicense, rsaParameters))
```
ayende: Is this also the priority order for reading the license?

src/Raven.Server/Commercial/LicenseHelper.cs
```diff
@@ -0,0 +1,230 @@
...
+            if (licenseString == null)
+            {
+                preferredTarget = EnvironmentVariableTarget.User;
+                licenseString = GetLicenseString();
+            }
+
+            if (licenseString == null)
+            {
+                if (Logger.IsInfoEnabled)
+                    Logger.Info($"Failed to access the environment variable: `{RavenConfiguration.EnvironmentVariableLicenseString}`");
```
ayende, line 168: Won't this also trigger if we don't have the env var setup at all?

src/Raven.Server/Commercial/LicenseHelper.cs
```diff
@@ -0,0 +1,230 @@
...
+                return;
+            }
+
+            var oldLicense = DeserializeLicense(licenseString);
```
ayende: shouldn't this be on a try/catch?

src/Raven.Server/Commercial/LicenseHelper.cs
```diff
@@ -0,0 +1,230 @@
...
+            if (ValidateLicense(newLicense, rsaParameters, oldLicense) == false)
+                return;
+
+            var newLicenseString = JsonConvert.SerializeObject(newLicense);
+            Environment.SetEnvironmentVariable(RavenConfiguration.EnvironmentVariableLicenseString, newLicenseString, preferredTarget);
```
ayende, line 179: This is non persistent set.
For that matter, I'm not sure that there is something that we _can_ do properly here.
Consider a docker instance, which may have the env setup externally completely.

Shouldn't we keep a local copy of the license inside the server store instead as well?

src/Raven.Server/Commercial/LicenseHelper.cs
```diff
@@ -0,0 +1,230 @@
...
+            }
+            catch (Exception e)
+            {
+                if (Logger.IsInfoEnabled)
+                    Logger.Info($"Failed to validate license: `{oldLicense}`", e);
+
+                return true;
+            }
+
+            if (oldLicense.Id != newLicense.Id)
```
ayende, line 208: Why does this matter?

ayende (changes requested):

ayende (approved):

PR #9141, opened 2019-04-14
RavenDB-13396 Fix BlittableJsonReaderObject.TryGetMember performance …

maximburyak:
…degradation due to switch to framework's StringSegment

### DIFF
--- a/src/Sparrow/StringComparer.cs
+++ b/src/Sparrow/StringComparer.cs
@@ -1,8 +1,98 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Runtime.CompilerServices;
+using Microsoft.Extensions.Primitives;
+using Sparrow.Binary;
+using Sparrow.Utils;
 
 namespace Sparrow
 {
+
+    public struct StringSegmentEqualityStructComparer : IEqualityComparer<StringSegment>
+    {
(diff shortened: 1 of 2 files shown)


ppekrol (approved):

src/Sparrow/StringComparer.cs
```diff
@@ -1,8 +1,98 @@
...
+using Microsoft.Extensions.Primitives;
+using Sparrow.Binary;
+using Sparrow.Utils;
 
 namespace Sparrow
 {
+
+    public struct StringSegmentEqualityStructComparer : IEqualityComparer<StringSegment>
+    {
+        public static IEqualityComparer<StringSegment> BoxedInstance = new StringSegmentEqualityStructComparer();
```
ayende, line 13: Why are you using a `BoxedInstance` and not just an `Instance` ?

src/Sparrow/StringComparer.cs
```diff
@@ -1,8 +1,98 @@
...
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public int GetHashCode(StringSegment x)
+        {
+            int xStart = x.Offset;
+            int xSize = x.Length;
+            string xStr = x.Buffer;
+
+            uint hash = 0;
```
ayende, line 46: Should we start this at the length of the segment?

PR #9145, opened 2019-04-16
RavenDB-13404 When importing 4.1 dump file with counters on 4.2, need…

ayende:
… to import legacy counters


ayende:
No need, we have a specific field for that

PR #9151, opened 2019-04-17
RavenDB-12708: Fix clean shutdown on SIGTERM in Linux plus clean shut…

aviviadi:
…down also while in interactive mode

### DIFF
--- a/src/Raven.Server/Program.cs
+++ b/src/Raven.Server/Program.cs
@@ -268,6 +268,7 @@ namespace Raven.Server
                 {
                     if (Logger.IsOperationsEnabled)
                         Logger.OperationsAsync("Server has shut down").Wait(TimeSpan.FromSeconds(15));
+                    ShutdownCompleteMre.Set();
                 }
             } while (rerun);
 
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Program.cs
```diff
@@ -321,6 +323,7 @@ public static bool RunAsNonInteractive()
                     return; // already done
                 Console.WriteLine("Received graceful exit request...");
                 ShutdownServerMre.Set();
+                ShutdownCompleteMre.WaitOne(TimeSpan.FromSeconds(30)); // on linux we have to keep Unloading event exit last
```
ayende, line 326: Where are you ever setting this?

ayende (changes requested):

src/Raven.Server/Program.cs
aviviadi: https://github.com/ravendb/ravendb/pull/9151/files#diff-4539864a63aaae59904fb3d5a47f8e22R271

ayende (approved):

PR #9154, opened 2019-04-17
RavenDB-12624 Generate an error for cluster transaction on documents …

karmeli87:
…containing counters or attachments

### DIFF
--- a/src/Raven.Server/Documents/Handlers/BatchHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/BatchHandler.cs
@@ -131,7 +131,7 @@ namespace Raven.Server.Documents.Handlers
             }
         }
 
-        private static void ValidateCommand(MergedBatchCommand command)
+        private static void ValidateCommandForClusterWideTransaction(MergedBatchCommand command)
         {
             foreach (var document in command.ParsedCommands)
             {
--- a/src/Raven.Server/Documents/Handlers/BatchHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/BatchHandler.cs
@@ -73,7 +73,7 @@ namespace Raven.Server.Documents.Handlers
                     if (Server.Configuration.Core.FeaturesAvailability == FeaturesAvailability.Stable)
                         FeaturesAvailabilityException.Throw("Cluster Transactions");
 
-                    ValidateCommand(command);
+                    ValidateCommandForClusterWideTransaction(command);
 
                     using (Database.ClusterTransactionWaiter.CreateTask(out var taskId))
                     {
--- a/src/Raven.Server/Documents/Handlers/BatchHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/BatchHandler.cs
@@ -73,6 +73,8 @@ namespace Raven.Server.Documents.Handlers
                     if (Server.Configuration.Core.FeaturesAvailability == FeaturesAvailability.Stable)
                         FeaturesAvailabilityException.Throw("Cluster Transactions");
 
+                    ValidateCommand(command);
+
                     using (Database.ClusterTransactionWaiter.CreateTask(out var taskId))
                     {
                         // Since this is a cluster transaction we are not going to wait for the write assurance of the replication.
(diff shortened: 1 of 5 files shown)


src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -129,6 +131,31 @@ public async Task BulkDocs()
             }
         }
 
+        private static void ValidateCommand(MergedBatchCommand command)
```
ayende:
```suggestion
        private static void ValidateCommandForClusterWideTransaction(MergedBatchCommand command)
```

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -129,6 +131,31 @@ public async Task BulkDocs()
...
+        {
+            foreach (var document in command.ParsedCommands)
+            {
+                switch (document.Type)
+                {
+                    case CommandType.PUT:
+                    case CommandType.DELETE:
```
ayende, line 141: Do we need it for anything but `PUT` ?

src/Raven.Server/Documents/Handlers/BatchRequestParser.cs
```diff
@@ -315,7 +318,7 @@ private async Task<CommandData> MoveNextUnlikely(JsonOperationContext ctx)
                 if (_state.CurrentTokenType == JsonParserToken.EndArray)
                     return new CommandData { Type = CommandType.None };
 
-                return await ReadSingleCommand(ctx, _stream, _state, _parser, _buffer, _token);
+                return await ReadSingleCommand(ctx, _stream, _state, _parser, _buffer, modifier, _token);
```
ayende, line 321: can we not pass a null here for most cases?

src/Raven.Server/Documents/Handlers/BatchHandler.cs
karmeli87: Don't think so. 
no silent merges if the document is deleted.

src/Raven.Server/Documents/Handlers/BatchRequestParser.cs
karmeli87: I wanted to be on the safe side with this change. 
But we probably later, can do a better reuse for the modifier.
Or I'm missing the point?

ayende: The idea is that you are changing a lot of places, for single usecase.
Better to reduce the number of changes.

src/Raven.Server/Documents/Handlers/BulkInsertHandler.cs
```diff
@@ -56,7 +56,7 @@ private async Task<IOperationResult> DoBulkInsert(Action<IOperationProgress> onP
                             long totalSize = 0;
                             while (true)
                             {
-                                var task = parser.MoveNext(docsCtx);
+                                var task = parser.MoveNext(docsCtx, new BlittableMetadataModifier(docsCtx));
```
ppekrol, line 59: can we reuse the modifier here?

ayende (approved):

src/Raven.Server/Documents/Handlers/BulkInsertHandler.cs
karmeli87: The modifier is bounded to the context. 
I wanted to avoid complexity for now, since we switch the context here (few times).

ppekrol: BlittableMetadataModifier needs to be disposed. We need to detect if context has switched.

ppekrol (changes requested):

src/Raven.Server/Documents/Handlers/BulkInsertHandler.cs
karmeli87: The `BlittableMetadataModifier` is disposed when the context is and the behavior is not different from what we had before this change.
However we do need to do that properly, so I opened an issue for that:
https://issues.hibernatingrhinos.com/issue/RavenDB-13445

ppekrol (approved):

PR #9156, opened 2019-04-17
RavenDB-12568 preventing inserting of documents with @attachments fro…

talweiss1982:
…m DocumentHandler.MergedPutCommand

### DIFF
--- a/src/Raven.Server/Documents/Handlers/DocumentHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/DocumentHandler.cs
@@ -9,6 +9,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Net;
+using System.Runtime.CompilerServices;
 using System.Runtime.ExceptionServices;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Primitives;
@@ -17,6 +18,7 @@ using Raven.Client.Documents.Changes;
 using Raven.Client.Documents.Commands;
 using Raven.Client.Documents.Commands.Batches;
 using Raven.Client.Documents.Operations;
+using Raven.Client.Documents.Operations.Attachments;
 using Raven.Client.Documents.Operations.Counters;
 using Raven.Client.Exceptions;
 using Raven.Server.Documents.Includes;
@@ -31,6 +33,7 @@ using Sparrow.Extensions;
 using Sparrow.Json;
 using Sparrow.Json.Parsing;
 using Sparrow.Utils;
+using Voron;
 using DeleteDocumentCommand = Raven.Server.Documents.TransactionCommands.DeleteDocumentCommand;
 using PatchRequest = Raven.Server.Documents.Patch.PatchRequest;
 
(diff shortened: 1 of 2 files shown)


ayende (commented):
Are we going to do the wiring of the attachment references here?
In other words, consider:

* Create doc `a`
* Add attachment `b`
* Clone `a` doc to `c`
* Delete `a`
* Access `c` and then try to get `b`

Does this work?

talweiss1982:
@ayende it will if clone will be done right and increase the ref count of the attachment.
At any rate, my changes to prevent insertion of @attachments are only for direct put using a command (exposed by the client but not used) if you are using a session then you will go through the batch handler.

ayende:
@talweiss1982 What is the actual behavior right now?

talweiss1982:
there is no clone as much as I know

ayende:
From the studio

talweiss1982:
you don't have a way to do this with a single transaction from the studio.
I'm not sure you can with a session either as the reading of the attachment won't be in the same transaction either.
Only with a patch, I think you could achieve this scenario and i don't see how my changes will affect the behavior of this scenario.

test/FastTests/Issues/RavenDB-12568.cs
```diff
@@ -0,0 +1,57 @@
...
+            {
+                var requestExecutor = store.GetRequestExecutor();
+
+                using (var context = JsonOperationContext.ShortTermSingleUse())
+                using (var stringStream = new MemoryStream(System.Text.Encoding.UTF8.GetBytes(Json)))
+                using (var blittableJson = context.Read(stringStream, "Reading of foo/bar"))
+                {
+                    try
+                    {
+                        requestExecutor.Execute(new PutDocumentCommand("foo/bar", null, blittableJson), context);
```
ppekrol, line 29: Use Assert.Throws.
if this command will not throw then this test will still pass.

src/Raven.Server/Documents/Handlers/DocumentHandler.cs
```diff
@@ -551,24 +554,57 @@ public class MergedPutCommand : TransactionOperationsMerger.MergedTransactionCom
...
+        {
+            if (document.TryGet(Constants.Documents.Metadata.Key, out BlittableJsonReaderObject metadata)
+                && metadata.TryGet(Constants.Documents.Metadata.Attachments, out BlittableJsonReaderArray attachments))
+            {
+                ValidateAttachmentHashInternal(context, id, attachments);
+            }
+        }
+
+        private void ValidateAttachmentHashInternal(DocumentsOperationContext context, string id, BlittableJsonReaderArray attachments)
+        {
```
ppekrol, line 584: Just in case can we check if attachments array is not null? I think that `{ '@attachments' : null }` might be a problem here?

ppekrol:
Studio is using `bulk_docs` for that IIRC. @talweiss1982 are we covering bulkInsert?

ayende:
@ppekrol The issue is probably not visible via the `bulk_docs`, we are talking about the behavior when we invoke `PUT /docs?id=doc` directly, for example, via fiddler

test/FastTests/Issues/RavenDB-12568.cs
talweiss1982: It throws RavenException which is a very general exception

ppekrol: Which is fine, but if it will not throw anything (e.g. someone will change sth in the future) then this test will still pass

talweiss1982:
The scenario of using PutDocument puts us in a bad position since it doesn't clean unwanted properties i talk with @ayende regarding this one and we decided that we will just clean the unwanted properties.
I'm closing this PR since its code will no longer be relevant after my changes.

PR #9157, opened 2019-04-17
RavenDB-13304: WaitForIndexNotification return index only after command complete

garayx:


src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1094,21 +1084,36 @@ private void NotifyDatabaseAboutChanged(TransactionOperationContext context, str
...
-                    }, null);
+                {
+                    ExecuteAsyncTask(index, () => DatabaseChanged?.Invoke(this, (databaseName, index, type, change)));
+                }
             };
         }
 
+        private void ExecuteAsyncTask(long index, Action action)
+        {
+            _rachisLogIndexNotifications.AddTask(index);
```
grisha-kotler, line 1096: A comment is needed here to explain why we do this _before_ starting the async task

test/SlowTests/Cluster/ClusterIndexNotificationsTest.cs
```diff
@@ -0,0 +1,134 @@
...
+
+                using (testingStuff.CallDuringDocumentDatabaseInternalDispose(() =>
+                {
+                    Thread.Sleep(12345);
+                }))
+                {
+                    var stats = store2.Maintenance.Send(new GetStatisticsOperation());
+
+                    var cts = new CancellationTokenSource();
+                    Parallel.Invoke(() => BackgroundWork(store2, cts));
```
grisha-kotler: - why do you need `Parallel.Invoke` here?

grisha-kotler: - You aren't waiting for the task to complete

test/SlowTests/Cluster/ClusterIndexNotificationsTest.cs
```diff
@@ -0,0 +1,134 @@
...
+
+                    while (stats.CountOfIndexes == 0)
+                    {
+                        Thread.Sleep(2000);
```
grisha-kotler: Use `WaitForValue` here

test/SlowTests/Cluster/ClusterIndexNotificationsTest.cs
```diff
@@ -0,0 +1,134 @@
...
+
+                using (testingStuff.CallDuringDocumentDatabaseInternalDispose(() =>
+                {
+                    Thread.Sleep(18 * 1000);
+                }))
+                {
+                    var stats = store2.Maintenance.Send(new GetStatisticsOperation());
+
+                    var cts = new CancellationTokenSource();
+                    Parallel.Invoke(() => BackgroundWork(store2, cts));
```
grisha-kotler: Same here

test/SlowTests/Cluster/ClusterIndexNotificationsTest.cs
```diff
@@ -0,0 +1,134 @@
...
+
+                    while (stats.CountOfIndexes == 0)
+                    {
+                        Thread.Sleep(2000);
```
grisha-kotler: Same here

grisha-kotler (changes requested):

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1486,16 +1493,20 @@ private void OnTransactionDispose(TransactionOperationContext context, long inde
         {
             context.Transaction.InnerTransaction.LowLevelTransaction.OnDispose += transaction =>
             {
+                _rachisLogIndexNotifications.AddTask(index);
```
karmeli87: this should be added only if the tx is committed

grisha-kotler (approved):

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -2138,15 +2159,58 @@ public void NotifyListenersAbout(long index, Exception e)
...
+
+        public void SetTaskCompleted(long index, Exception e)
+        {
+            if (_tasksDictionary.TryGetValue(index, out var tcs))
+            {
+                try
+                {
+                    // set the task as finished
+                    if (e == null)
+                        tcs.SetResult(null);
```
ayende: Use `TrySet` here

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -2138,15 +2159,58 @@ public void NotifyListenersAbout(long index, Exception e)
...
+                    else
+                        tcs.SetException(e);
+                }
+                catch (Exception ex)
```
ayende: What errors can be raised here?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -2138,15 +2159,58 @@ public void NotifyListenersAbout(long index, Exception e)
...
+                {
+                    if (Log.IsInfoEnabled)
+                        Log.Info($"Failed to set result of task with index {index}", ex);
+                }
+            }
+
+            _tasksDictionary.TryRemove(index, out _);
```
ayende, line 2204: Where are we putting the error in the queue?

ayende: For that matter, shouldn't we just use the dictionary, instead of doing linear search?
We can keep the tasks there until they are either:
* done (no error)
* done with error and observed
* done with error, unobserved and too much time have passed

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1487,17 +1495,12 @@ private void OnTransactionDispose(TransactionOperationContext context, long inde
...
-                        try
-                        {
-                            _rachisLogIndexNotifications.NotifyListenersAbout(index, null);
-                        }
-                        catch (Exception e)
-                        {
-                            _rachisLogIndexNotifications.NotifyListenersAbout(index, e);
-                        }
-                    }, null);
+                {
```
karmeli87, line 1498: Removing the `TaskExecutor.Execute` do make sense, since there is no actual work here.
But why removing the try/catch ?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
grisha-kotler: done

src/Raven.Server/ServerWide/ClusterStateMachine.cs
grisha-kotler: Nothing can throw there, except OperationCanceledException. Fixed.

src/Raven.Server/ServerWide/ClusterStateMachine.cs
grisha-kotler: SetResult and SetException

grisha-kotler: changed

src/Raven.Server/ServerWide/ClusterStateMachine.cs
karmeli87: But we do want to cancel the task, no?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
grisha-kotler: There are a couple of scenarios:
- We are waiting on the task (no error)
- We are waiting on the task (with error)
- We start waiting after the task was already completed (and removed from the dictionary).
In that case we look in the last errors queue (this is the only case that we do that):
https://github.com/ravendb/ravendb/blob/5e5be6d9c3da176b04242dfb760bdca526440f71/src/Raven.Server/ServerWide/ClusterStateMachine.cs#L2076

src/Raven.Server/ServerWide/ClusterStateMachine.cs
grisha-kotler: Done

ayende (approved):

ppekrol (approved):

PR #9158, opened 2019-04-17
RavenDB-13417 Set exception to the `topologyModification` task

karmeli87:
- If an exception occurred during `TryModifyTopology` we must set the exception to notifiy about it to other waiting threads.
- Don't be eager on update the promotable.

### DIFF
--- a/src/Raven.Server/Rachis/Leader.cs
+++ b/src/Raven.Server/Rachis/Leader.cs
@@ -609,7 +609,6 @@ namespace Raven.Server.Rachis
 
                 TryModifyTopology(ambassador.Key, ambassador.Value.Url, TopologyModification.Voter, out Task _);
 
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Rachis/Leader.cs
```diff
@@ -609,7 +609,6 @@ private void CheckPromotables()
 
                 TryModifyTopology(ambassador.Key, ambassador.Value.Url, TopologyModification.Voter, out Task _);
 
-                _promotableUpdated.Set();
```
ayende, line 612: why did you remove this?

karmeli87: There was a tight loop when the task from `TryModifyTopology` wasn't completed.

ayende (approved):

PR #9160, opened 2019-04-18
RavenDB-13380 can't disable subscription w/o name

efratshenhar:


src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -1346,11 +1346,26 @@ public Task<(long Index, object Result)> DeleteOngoingTask(long taskId, string t
...
-                    new ToggleTaskStateCommand(taskId, type, disable, dbName);
-
+            CommandBase disableEnableCommand;
+            switch (type)
+            {
+                case OngoingTaskType.Subscription:
+                    if (taskName == null)
+                    {
+                        if (_server.ServerStore.DatabasesLandlord.DatabasesCache.TryGetValue(dbName, out var databaseTask) == false)
+                            throw new RachisApplyException($"Value cannot be null or empty. Param: {nameof(dbName)}");
```
ayende: Huh? That exception is _really_ misleading. The `dbName` isn't empty, it is just not a loaded database for this server.

ayende (changes requested):
This should handle the case the db is not loaded and give an appropriate error

ayende:
Error if the db does not exists

PR #9163, opened 2019-04-18
RavenDB-13260 request snapshot if we try to add node from a different cluster

karmeli87:


src/Raven.Server/Rachis/Follower.cs
```diff
@@ -60,11 +61,22 @@ private void FollowerSteadyState()
...
                 {
+                    using (context.OpenReadTransaction())
+                    {
+                        if (_engine.GetSnapshotRequest(context))
+                        {
+                            if (_engine.Log.IsInfoEnabled)
+                            {
+                                _engine.Log.Info($"{ToString()}: Snapshot was requested, so we close this follower.");
+                            }
+                            return;
```
ayende, line 76: This happens on each and every read from the network, which seems excessive.
We need the state persistent, but we can do well by just checking a property on the engine, instead.

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -781,6 +790,20 @@ public Task<(long Index, object Result)> PutAsync(CommandBase cmd)
...
+            using (context.OpenReadTransaction())
+            {
+                if (GetSnapshotRequest(context))
+                {
+                    // we aren't allowed to be elected for leadership if we requested a snapshot 
+                    if (Log.IsInfoEnabled)
+                    {
+                        Log.Info("we aren't allowed to be elected for leadership if we requested a snapshot");
+                    }
+                    return;
```
ayende: Shouldn't we throw here?

ayende: Like in `SetNewStateInTx` ?
Or is this because this is called from the timeout?

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -1138,6 +1161,9 @@ public unsafe void ClearLogEntriesAndSetLastTruncate(TransactionOperationContext
         }
         public unsafe void TruncateLogBefore(TransactionOperationContext context, long upto)
         {
+            return;
```
ayende: This looks like debug code that wasn't removed

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -1879,7 +1911,25 @@ public void UpdateNodeTag(TransactionOperationContext context, string newTag)
                 var state = context.Transaction.InnerTransaction.CreateTree(GlobalStateSlice);
                 state.Add(TagSlice, str);
             }
+        }
+
+        public void SetSnapshotRequest(TransactionOperationContext context, bool request)
+        {
+            using (Slice.From(context.Transaction.InnerTransaction.Allocator, request.ToString(), out Slice str))
```
ayende: Why are you storing it as a string? Can just store a byte.

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -439,6 +441,52 @@ public async Task RemoveFromClusterAsync(string nodeTag)
             await _engine.RemoveFromClusterAsync(nodeTag).WithCancellation(_shutdownNotification.Token);
         }
 
+        public async Task RequestSnapshot()
+        {
+            var topology = GetClusterTopology();
+
+            if (topology.AllNodes.Count == 1)
+                throw new InvalidOperationException("Can't force snapshot, since I'm the only node in the cluster.");
```
ayende, line 448: But I can, if I'm forcing myself back to passive?

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -439,6 +441,52 @@ public async Task RemoveFromClusterAsync(string nodeTag)
...
+                    }
+
+                    await WaitForState(RachisState.Follower, ServerShutdown);
+
+                    // demote to non-voter, so we will not participate in the quorum.
+                    var demote = new DemoteClusterNodeCommand(NodeTag);
+                    await requestExecutor.ExecuteAsync(demote, ctx, token: ServerShutdown);
+                }
+
+                using (var tx = ctx.OpenWriteTransaction())
```
ayende: No, not to all of this.
It is _really_ complex and have too many steps in the middle.
Instead, use:
* cannot call request snapshot on the leader (user can demote manually).

And that solves the whole issue for you.

ayende (changes requested):
I'm

ayende:
Also, who can _call_ RequestSnapshot_ ?

src/Raven.Server/Rachis/Follower.cs
karmeli87: done.

src/Raven.Server/Rachis/RachisConsensus.cs
karmeli87: yes, It is a timeout callback.

src/Raven.Server/Rachis/RachisConsensus.cs
karmeli87: oops :/

src/Raven.Server/Rachis/RachisConsensus.cs
karmeli87: done.

karmeli87:
The `RequestSnapshot` property can be set by the following:
1. From the JS admin console
2. When we hard reset to passive

ayende:
Let's create a docs issue with all these commands, so we'll know what options we have

ayende (approved):

src/Raven.Client/ServerWide/Commands/Cluster/AddClusterNodeCommand.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System.Net.Http;
+using Raven.Client.Http;
+using Sparrow.Json;
+
+namespace Raven.Client.ServerWide.Commands.Cluster
+{
+    public class AddClusterNodeCommand : RavenCommand
```
ppekrol: switch to internal

src/Raven.Client/ServerWide/Commands/Cluster/DemoteClusterNodeCommand.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Text;
+using Raven.Client.Documents.Conventions;
+using Raven.Client.Http;
+using Sparrow.Json;
+
+namespace Raven.Client.ServerWide.Commands.Cluster
+{
+    public class DemoteClusterNodeCommand : RavenCommand
```
ppekrol: switch to internal

src/Raven.Client/ServerWide/Commands/Cluster/PromoteClusterNodeCommand.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Text;
+using Raven.Client.Documents.Conventions;
+using Raven.Client.Http;
+using Sparrow.Json;
+
+namespace Raven.Client.ServerWide.Commands.Cluster
+{
+    public class PromoteClusterNodeCommand : RavenCommand
```
ppekrol: switch to internal

src/Raven.Client/ServerWide/Commands/Cluster/RemoveClusterNodeCommand.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Text;
+using Raven.Client.Documents.Conventions;
+using Raven.Client.Http;
+using Sparrow.Json;
+
+namespace Raven.Client.ServerWide.Commands.Cluster
+{
+    public class RemoveClusterNodeCommand : RavenCommand
```
ppekrol: switch to internal

src/Raven.Client/ServerWide/Commands/Cluster/AddClusterNodeCommand.cs
```diff
@@ -0,0 +1,37 @@
...
+        public AddClusterNodeCommand(string url, string tag, bool watcher)
+        {
+            _url = url;
+            _tag = tag;
+            _watcher = watcher;
+        }
+
+        public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, ServerNode node, out string url)
+        {
+            url = $"{node.Url}/admin/cluster/node?url={_url}&watcher={_watcher}";
```
ppekrol: do we need to escape the parameters?

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -296,6 +297,7 @@ static RachisConsensus()
                 Slice.From(ctx, "Topology", out TopologySlice);
                 Slice.From(ctx, "LastTruncated", out LastTruncatedSlice);
                 Slice.From(ctx, "Entries", out EntriesSlice);
+                Slice.From(ctx, "SnapshotRequestSlice", out SnapshotRequestSlice);
```
ppekrol: rename string to `SnapshotRequest`

ppekrol (changes requested):
@karmeli87 can we write a test for this?

ppekrol (approved):

PR #9164, opened 2019-04-19
RavenDB-7498 Changes in aggressive cache:

arekpalinski:
-  added caching mode options to aggressive cache (TrackChangesAndDoBackgroundRefresh - not implemented yet)
- removed _aggressiveCachingUsed as we might provide different 'database' name

### DIFF
--- a/src/Raven.Client/Http/AggressiveCacheMode.cs
+++ b/src/Raven.Client/Http/AggressiveCacheMode.cs
@@ -1,0 +1,10 @@
+﻿namespace Raven.Client.Http
+{
+    public enum AggressiveCacheMode
+    {
+        Unknown = 0,
+        TrackChanges = 1,
+        TrackChangesAndDoBackgroundRefresh = 2, // TODO arek - NotImplementedException
+        DoNotTrackChanges = 3
+    }
+}
--- a/src/Raven.Client/Documents/Changes/EvictItemsFromCacheBasedOnChanges.cs
+++ b/src/Raven.Client/Documents/Changes/EvictItemsFromCacheBasedOnChanges.cs
@@ -16,7 +16,7 @@ namespace Raven.Client.Documents.Changes
         private readonly IDatabaseChanges _changes;
         private readonly IDisposable _documentsSubscription;
         private readonly IDisposable _indexesSubscription;
-        private RequestExecutor _requestExecutor;
+        private readonly RequestExecutor _requestExecutor;
 
         public EvictItemsFromCacheBasedOnChanges(DocumentStore store, string databaseName)
         {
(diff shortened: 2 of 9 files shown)


src/Raven.Client/Http/AggressiveCacheMode.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Raven.Client.Http
+{
+    public enum AggressiveCacheMode
+    {
+        Unknown = 0,
+        TrackChanges = 1,
+        TrackChangesAndDoBackgroundRefresh = 2, // TODO arek - NotImplementedException
```
ayende, line 7: I don't like this. Background refresh can be _really_ expensive and we don't want to do things behind the user's back

src/Raven.Client/Http/AggressiveCacheMode.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Raven.Client.Http
+{
+    public enum AggressiveCacheMode
+    {
+        Unknown = 0,
```
ppekrol, line 5: Do we need `Unknown` ?

Better to switch to
- `Live`
- `Passive` maybe?

src/Raven.Client/Http/AggressiveCacheMode.cs
ppekrol: This will make things easier. Arek added this because it was written in the issue to implement this.

src/Raven.Client/Http/AggressiveCacheMode.cs
ayende: We typically have _something_ as `0` value in flags. Such as `None`.
I would rather have something that is disabled.

src/Raven.Client/Http/AggressiveCacheMode.cs
arekpalinski: Yeah, the idea of having the background update was taken from https://issues.hibernatingrhinos.com/issue/RavenDB-7498 but I had some doubts how to do it properly so I left it without the implementation.

I'll drop this option entirely then.

src/Raven.Client/Http/AggressiveCacheMode.cs
arekpalinski: 1) I removed `Unknown` in another PR (https://github.com/ravendb/ravendb/pull/9165/files#diff-950a76aa4eb57746fcc07663c915c377R5) and made `TrackChanges` option as `0` because it's our default and didn't want to allow users to choose `Unknown / None` which isn't a viable option anyway.

2) `Live / Passive` aren't really meaningful. Let me stay with current names as they are more descriptive I think.

src/Raven.Client/Http/AggressiveCacheMode.cs
arekpalinski: Fixed in https://github.com/ravendb/ravendb/pull/9179

PR #9165, opened 2019-04-19
RavenDB-7498 Wrapping aggressive cache options into convention class …

arekpalinski:
…wrapper

### DIFF
--- a/src/Raven.Client/Documents/Conventions/DocumentConventions.cs
+++ b/src/Raven.Client/Documents/Conventions/DocumentConventions.cs
@@ -54,6 +54,40 @@ namespace Raven.Client.Documents.Conventions
 
         public readonly BulkInsertConventions BulkInsert;
 
+        public readonly AggressiveCacheConventions AggressiveCache;
+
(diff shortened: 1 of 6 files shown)


src/Raven.Client/Documents/Conventions/DocumentConventions.cs
```diff
@@ -54,6 +54,40 @@ public class DocumentConventions : Client.Conventions
...
         public readonly BulkInsertConventions BulkInsert;
 
+        public readonly AggressiveCacheConventions AggressiveCache;
+
+        public class AggressiveCacheConventions
+        {
+            private readonly DocumentConventions _conventions;
+            private readonly AggressiveCacheOptions _aggressiveCacheOptions;
+            
+            public AggressiveCacheConventions(DocumentConventions conventions)
```
ppekrol, line 64: should be internal

arekpalinski: Fixed in https://github.com/ravendb/ravendb/pull/9179

PR #9166, opened 2019-04-19
RavenDB-13421 Fixing rvn_prefetch_ranges Windows implementation (PAL)

arekpalinski:


arekpalinski:
@haludi @aviv86 Please review

haludi:
I think this is the issue too (and for sure an issue).
In addition, we maybe missed it so far because we don't check the error value when we call the function `rvn_prefetch_ranges`
https://github.com/ravendb/ravendb/blob/v4.2/src/Voron/GlobalPrefetchingBehavior.cs#L69
https://github.com/ravendb/ravendb/blob/v4.2/src/Voron/GlobalPrefetchingBehavior.cs#L77

arekpalinski:
@haludi What do you mean by "I think this is the issue too" ? Yes, I think we need to check the result of `rvn_prefetch_ranges` call

PR #9184, opened 2019-04-23
RavenDB-12568 using BlittableMetadataModifier when putting document d…

talweiss1982:
…irectly to avoid deviation of behavior between batch insert and manual insert

### DIFF
--- a/src/Raven.Server/Documents/Handlers/DocumentHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/DocumentHandler.cs
@@ -322,7 +322,7 @@ namespace Raven.Server.Documents.Handlers
                 // and the identity generation needs to take into account that the identity 
                 // generation can fail and will leave the reading task hanging if we abort
                 // easier to just do in synchronously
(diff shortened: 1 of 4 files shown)


ayende (approved):

ppekrol (approved):

talweiss1982:
@ppekrol those tests are doing something they shouldn't resolving conflicts by inserting raw blittables with unchecked metadata, i'm trying to see if i can adjust them to work properly but its a catch 22...

ppekrol:
We need to look into those, maybe those tests need to be rewritten now.

talweiss1982:
@ppekrol after looking at it we actually have no other way to resolve conflicts containing attachments manually. We could add a special command for manually choose the change vector that should win but I think this is breaking changes since people may have already relayed on this method of conflict resolving.
I think we should just leave PutDocument as it is document that it should avoid messing with the metadata and should only contain metadata for the purpose of resolving conflicts.
@ayende thoughts?

ayende:
@talweiss1982 At a minimum, we *must* ensure that we can't create attachments that do not exists.

ayende:
Or without references to them

PR #9185, opened 2019-04-23
fix for RavenDB-13340

myarichuk:


myarichuk:
My bad, previous fix _should_ be enough

PR #9186, opened 2019-04-23
RavenDB-13443 fixing an issue where deleting revisions where none exi…

talweiss1982:
…st would throw

### DIFF
--- a/src/Voron/Data/Tables/Table.cs
+++ b/src/Voron/Data/Tables/Table.cs
@@ -906,7 +906,7 @@ namespace Voron.Data.Tables
         public TableValueHolder SeekOneForwardFromPrefix(TableSchema.SchemaIndexDef index, Slice value)
         {
             var tree = GetTree(index);
-            if (tree == null)
+            if (tree == null || tree.State.NumberOfEntries == 0)
                 return null;
 
             using (var it = tree.Iterate(true))
--- a/src/Voron/Data/Tables/Table.cs
+++ b/src/Voron/Data/Tables/Table.cs
@@ -872,6 +872,9 @@ namespace Voron.Data.Tables
         public IEnumerable<SeekResult> SeekForwardFrom(TableSchema.SchemaIndexDef index, Slice value, int skip, bool startsWith = false)
         {
             var tree = GetTree(index);
+            if (tree == null || tree.State.NumberOfEntries == 0)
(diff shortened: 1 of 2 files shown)


src/Voron/Data/Tables/Table.cs
```diff
@@ -903,6 +906,9 @@ public IEnumerable<SeekResult> SeekForwardFrom(TableSchema.SchemaIndexDef index,
         public TableValueHolder SeekOneForwardFromPrefix(TableSchema.SchemaIndexDef index, Slice value)
         {
             var tree = GetTree(index);
+            if (tree == null)
```
ppekrol: should we check that number of entries in 0 here?

ppekrol (approved):

src/Voron/Data/Tables/Table.cs
talweiss1982: it's a perf thingy but yeah

src/Voron/Data/Tables/Table.cs
```diff
@@ -872,6 +872,9 @@ public long GetNumberOfEntriesFor(TableSchema.FixedSizeSchemaIndexDef index)
         public IEnumerable<SeekResult> SeekForwardFrom(TableSchema.SchemaIndexDef index, Slice value, int skip, bool startsWith = false)
         {
             var tree = GetTree(index);
+            if (tree == null || tree.State.NumberOfEntries == 0)
```
ayende, line 875: `|| tree.State.NumberOfEntries == 0` is already handled by the `Iterate` method

src/Voron/Data/Tables/Table.cs
ayende: I don't think we should do that, it is an already handled scenario

ppekrol: Small optimization to avoid creating iterator, setting prefix, seeking, etc. I do not see a good reason not to have this?

ayende: I can't think of any scenario where this would be meaningful performance optimization.
Null / empty trees are rare, but you've added a new conditional for every call

ppekrol: Removed this from 6 locations: https://github.com/ravendb/ravendb/pull/9188

PR #9189, opened 2019-04-24
RavenDB-12568 preventing the insertion of a document with attachment …

talweiss1982:
…if those do not exist

Note that the scenario of "cloning" a document and inserting it will not automatically increase the ref count of the attachment and this is by design.

Conflict resolution is relying on this behavior to resolve conflicts!

If you wish to clone a document with an attachment you should clone the attachment too like it is done in the studio.


ayende (commented):
Will this handle creating the reference to the attachment?

talweiss1982:
No, and it shouldn't.
Creating an attachment in this manner is not allowed as I explained in detail above.
If you wish to add a reference you need to do it manually doing this automatically will break our logic.
The only time you want to use `PutDocument` with existing metadata is when you're using it to resolve conflict and when you do that you don't want to increment the ref count of an attachment.

ayende (approved):

PR #9193, opened 2019-04-26
RavenDB-11883 added spatial index to sample data

ppekrol:


ayende:
We probably want to add some sample queries for this.

ayende (approved):

ppekrol:
We will take leverage of this index in our Demo app.

PR #9194, opened 2019-04-26
RavenDB-12669 prevent adding sorter with same code and different name

ppekrol:


test/SlowTests/Issues/RavenDB_8355.cs
```diff
@@ -89,6 +90,18 @@ public void CanUseCustomSorterWithOperations()
                     Code = sorterCode
                 }));
 
+                var e = Assert.Throws<SorterCompilationException>(() =>
+                {
+                    // We should not be able to add sorter with non-matching name
+                    store.Maintenance.Send(new PutSortersOperation(new SorterDefinition
```
ayende, line 96: Why not? This is really common thing. You start from the same definition you have and change it.

ppekrol: Then class name needs to be changed. We are validating that when first sorter is created. Sorters is a Dict of `<name, definition>` and the initial agreement was that the name must match the class name inside the definition.

PR #9199, opened 2019-04-28
[RavenDB-13458] change backup upload time calculation

Or-Geva:


src/Raven.Client/Documents/Operations/Backups/BackupStatus.cs
```diff
@@ -138,13 +137,21 @@ public UploadProgress()
 
         public double BytesPutsPerSec { get; set; }
 
-        public long UploadTimeInMs => _sw.ElapsedMilliseconds;
+        public long UploadTimeInMs => _sw?.ElapsedMilliseconds ?? 0;
 
         public void ChangeState(UploadState newState)
         {
             UploadState = newState;
             if (newState == UploadState.Done)
```
grisha-kotler: can be removed

grisha-kotler (changes requested):

grisha-kotler (approved):

src/Raven.Client/Documents/Operations/Backups/BackupStatus.cs
```diff
@@ -138,13 +137,20 @@ public UploadProgress()
...
             UploadState = newState;
-            if (newState == UploadState.Done)
-                _sw.Stop();
+            switch (newState)
+            {
+                case UploadState.PendingUpload:
+                    _sw = Stopwatch.StartNew();
+                    break;
+                case UploadState.Done:
+                    _sw.Stop();
```
ayende, line 152: Can this be null?
Can we have the install at all times and just call Restart on PendingUpdate?

grisha-kotler: Shouldn't be since we set UploadState to PendingUpload before starting the upload

ayende: I'm not sure we can always ensure that this happens.
Better to avoid the potential NRE

Or-Geva: Done.

src/Raven.Client/Documents/Operations/Backups/BackupStatus.cs
Or-Geva: Removed

PR #9200, opened 2019-04-28
Raven db 13320

talweiss1982:


src/Raven.Server/Documents/Handlers/AttachmentHandler.cs
```diff
@@ -305,23 +301,16 @@ public class MergedPutAttachmentCommand : TransactionOperationsMerger.MergedTran
...
             public AttachmentDetails Result;
             public string ContentType;
             public Stream Stream;
             public string Hash;
 
             protected override int ExecuteCmd(DocumentsOperationContext context)
             {
-                try
-                {
-                    Result = Database.DocumentsStorage.AttachmentsStorage.PutAttachment(context, DocumentId, Name,
```
ayende, line 318: Why do we need this change if we already have the change in the db init?

talweiss1982: if the command failed we should not commit the transaction.
I don't know what bugs are lurking over here but surely somebody might add some more persistent state in the future that should never be commited so if a command fails it should not persist anything.

ayende: We actually do want to do it.
The logic here is that this is an independent operation, so if it fails, it doesn't impact the rest of the merged tx.
We want to keep the same behavior to make sure that the next time we don't actually change the underlying datastructure but change the database change vector, we aren't getting NRE again

talweiss1982: if you want this behavior than i'll just open a new PR with just that

PR #9201, opened 2019-04-29
RavenDB-13320

talweiss1982:


ayende (commented):
Can you add the test as well?

talweiss1982:
@ayende the test is irrelevant since it was just checking the concurrency throw and the fact that the change vector didn't increment now it won't throw and will increment the change vector

PR #9204, opened 2019-04-29
RavenDB-11840 Implemention of compress log file and configuration for compression and retention by time and size

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-11840

### DIFF
--- a/src/Raven.Server/Config/Categories/LogsConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/LogsConfiguration.cs
@@ -26,10 +26,23 @@ namespace Raven.Server.Config.Categories
         [ConfigurationEntry("Logs.MaxFileSizeInMb", ConfigurationEntryScope.ServerWideOnly)]
         public Size MaxFileSize { get; set; }
 
-        [Description("How far back we should retain log entries")]
-        [DefaultValue(3 * 24)]
-        [TimeUnit(TimeUnit.Hours)]
-        [ConfigurationEntry("Logs.RetentionTimeInHrs", ConfigurationEntryScope.ServerWideOnly)]
-        public TimeSetting RetentionTime { get; set; }
+        [Description("How far back we should retain log entries in days")]
+        [DefaultValue(null)]
+        [MinValue(1)]
+        [TimeUnit(TimeUnit.Days)]
+        [ConfigurationEntry("Logs.RetentionTimeInDays", ConfigurationEntryScope.ServerWideOnly)]
+        public TimeSetting? RetentionTime { get; set; }
+
+        [Description("The maximum size of the log after which the old files will be deleted")]
+        [DefaultValue(null)]
+        [MinValue(256)]
+        [SizeUnit(SizeUnit.Megabytes)]
+        [ConfigurationEntry("Logs.RetentionSizeInMb", ConfigurationEntryScope.ServerWideOnly)]
+        public Size? RetentionSize { get; set; }
+
+        [Description("Will determine whether to compress the log files")]
+        [DefaultValue(false)]
+        [ConfigurationEntry("Logs.Compress", ConfigurationEntryScope.ServerWideOnly)]
+        public bool Compress { get; set; }
     }
 }
--- a/src/Raven.Server/Config/Categories/SecurityConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/SecurityConfiguration.cs
@@ -32,12 +32,12 @@ namespace Raven.Server.Config.Categories
         [DefaultValue(null)]
         [MinValue(256)]
         [SizeUnit(SizeUnit.Megabytes)]
-        [ConfigurationEntry("Logs.RetentionSizeInMb", ConfigurationEntryScope.ServerWideOnly)]
+        [ConfigurationEntry("Security.AuditLog.RetentionSizeInMb", ConfigurationEntryScope.ServerWideOnly)]
         public Size? AuditLogRetentionSize { get; set; }
 
         [Description("Will determine whether to compress the audit log files")]
         [DefaultValue(false)]
-        [ConfigurationEntry("Logs.Compress", ConfigurationEntryScope.ServerWideOnly)]
+        [ConfigurationEntry("Security.AuditLog.Compress", ConfigurationEntryScope.ServerWideOnly)]
         public bool AuditLogCompress { get; set; }
 
         [Description("The path to .pfx certificate file. If specified, RavenDB will use HTTPS/SSL for all network activities. Certificate setting priority order: 1) Path 2) Executable")]
--- a/src/Raven.Server/Config/Categories/SecurityConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/SecurityConfiguration.cs
@@ -5,6 +5,7 @@ using System.Net;
 using Raven.Server.Config.Attributes;
 using Raven.Server.Config.Settings;
 using Raven.Server.Utils;
+using Sparrow;
 
 namespace Raven.Server.Config.Categories
 {
@@ -25,7 +26,19 @@ namespace Raven.Server.Config.Categories
         [TimeUnit(TimeUnit.Hours)]
         [ConfigurationEntry("Security.AuditLog.RetentionTimeInHrs", ConfigurationEntryScope.ServerWideOnly)]
         [ConfigurationEntry("Security.AuditLog.RetentionTimeInHours", ConfigurationEntryScope.ServerWideOnly)]
-        public TimeSetting AuditLogRetention { get; set; }
+        public TimeSetting AuditLogRetentionTime { get; set; }
+
+        [Description("The maximum size of the audit log after which the old files will be deleted")]
+        [DefaultValue(null)]
+        [MinValue(256)]
+        [SizeUnit(SizeUnit.Megabytes)]
+        [ConfigurationEntry("Logs.RetentionSizeInMb", ConfigurationEntryScope.ServerWideOnly)]
+        public Size? AuditLogRetentionSize { get; set; }
+
+        [Description("Will determine whether to compress the audit log files")]
+        [DefaultValue(false)]
+        [ConfigurationEntry("Logs.Compress", ConfigurationEntryScope.ServerWideOnly)]
+        public bool AuditLogCompress { get; set; }
 
         [Description("The path to .pfx certificate file. If specified, RavenDB will use HTTPS/SSL for all network activities. Certificate setting priority order: 1) Path 2) Executable")]
         [DefaultValue(null)]
--- a/src/Raven.Server/Program.cs
+++ b/src/Raven.Server/Program.cs
@@ -82,7 +82,6 @@ namespace Raven.Server
 
             LoggingSource.UseUtcTime = configuration.Logs.UseUtcTime;
             LoggingSource.Instance.MaxFileSizeInBytes = configuration.Logs.MaxFileSize.GetValue(SizeUnit.Bytes);
-            LoggingSource.AuditLog.MaxFileSizeInBytes = configuration.Logs.MaxFileSize.GetValue(SizeUnit.Bytes);
             LoggingSource.Instance.SetupLogMode(
                 configuration.Logs.Mode, 
                 configuration.Logs.Path.FullPath, 
--- a/src/Raven.Server/Program.cs
+++ b/src/Raven.Server/Program.cs
@@ -81,8 +81,15 @@ namespace Raven.Server
             configuration.Initialize();
 
             LoggingSource.UseUtcTime = configuration.Logs.UseUtcTime;
-            LoggingSource.MaxFileSizeInBytes = configuration.Logs.MaxFileSize.GetValue(SizeUnit.Bytes);
-            LoggingSource.Instance.SetupLogMode(configuration.Logs.Mode, configuration.Logs.Path.FullPath, configuration.Logs.RetentionTime.AsTimeSpan);
+            LoggingSource.Instance.MaxFileSizeInBytes = configuration.Logs.MaxFileSize.GetValue(SizeUnit.Bytes);
+            LoggingSource.AuditLog.MaxFileSizeInBytes = configuration.Logs.MaxFileSize.GetValue(SizeUnit.Bytes);
+            LoggingSource.Instance.SetupLogMode(
+                configuration.Logs.Mode, 
+                configuration.Logs.Path.FullPath, 
+                configuration.Logs.RetentionTime?.AsTimeSpan,
+                configuration.Logs.RetentionSize?.GetValue(SizeUnit.Bytes),
+                configuration.Logs.Compress
+                );
 
             if (Logger.IsInfoEnabled)
                 Logger.Info($"Logging to {configuration.Logs.Path} set to {configuration.Logs.Mode} level.");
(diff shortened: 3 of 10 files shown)


src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -239,12 +274,51 @@ private FileStream GetNewStream(long maxFileSize)
...
+            foreach (var log in logGzFilesInfo.Reverse())
+            {
+                if (totalLogSize > RetentionSize)
+                    try
+                    {
+                        log.Delete();
+                    }
+                    catch (Exception e)
+                    {
+                        //TODO What can we do in this situation??
```
haludi: How can we handle a situation when can't be deleted?
in case of:
1) The file is in use
2) The remote connection to file lost

src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -204,28 +220,47 @@ private void StartNewLoggingThread()
...
+            _readyToCompress.Set();
+            _tokenSource.Cancel();
             _loggingThread.Join(TimeToWaitForLoggingToEndInMilliseconds);
-
+            _compressLoggingThread.Join();
         }
 
         private FileStream GetNewStream(long maxFileSize)
         {
+            var logFiles = Directory.GetFiles(_path, $"{_dateString}.*.log");
```
haludi: Should I consider a network error?

src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -204,28 +220,47 @@ private void StartNewLoggingThread()
...
+            Array.Sort(logFiles);
+            var logGzFiles = Directory.GetFiles(_path, $"{_dateString}.*.log.gz");
```
haludi: When compressed is it can be too much to get all the file names at once?

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -25,7 +26,19 @@ public class SecurityConfiguration : ConfigurationCategory
         [TimeUnit(TimeUnit.Hours)]
         [ConfigurationEntry("Security.AuditLog.RetentionTimeInHrs", ConfigurationEntryScope.ServerWideOnly)]
         [ConfigurationEntry("Security.AuditLog.RetentionTimeInHours", ConfigurationEntryScope.ServerWideOnly)]
-        public TimeSetting AuditLogRetention { get; set; }
+        public TimeSetting AuditLogRetentionTime { get; set; }
```
haludi, line 29: Should the configuration of the audit log be the same as the regular log

src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -204,28 +220,47 @@ private void StartNewLoggingThread()
                 Name = _name + " Thread"
             };
             _loggingThread.Start();
+            if (compress)
+            {
+                _compressLoggingThread = new Thread(BackgroundLoggerCompress)
+                {
+                    IsBackground = true,
+                    Name = _name + "Compress Thread"
```
ayende:
```suggestion
                    Name = _name + " Log Compression Thread"
```

src/Sparrow/Logging/LoggingSource.cs
ayende: What do you mean, when we are failing to get the directory is on the network?

src/Sparrow/Logging/LoggingSource.cs
ayende: We can usually state that losing I/O connectivity is _bad_. In this case, we can ignore the error and not handle it at all.
It will be fixed the next time we deal with it.

src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -239,12 +274,51 @@ private FileStream GetNewStream(long maxFileSize)
...
+                    try
+                    {
+                        log.Delete();
+                    }
+                    catch (Exception e)
+                    {
+                        //TODO What can we do in this situation??
+                    }
+                else
+                    return;
```
ayende, line 318: Use `{}` here

src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -239,12 +274,51 @@ private FileStream GetNewStream(long maxFileSize)
...
+            }
+
+            foreach (var log in logFilesInfo.Reverse())
+            {
+                if (totalLogSize > RetentionSize)
```
ayende: Use `{}`

src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -239,12 +274,51 @@ private FileStream GetNewStream(long maxFileSize)
...
+                    try
+                    {
+                        log.Delete();
+                    }
+                    catch (Exception e)
+                    {
+                        //TODO What can we do in this situation??
+                    }
+                else
+                    return;
```
ayende: Use `{}`

src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -239,12 +274,51 @@ private FileStream GetNewStream(long maxFileSize)
...
+            foreach (var log in logFilesInfo.Reverse())
+            {
+                if (totalLogSize > RetentionSize)
+                    try
+                    {
+                        log.Delete();
+                    }
+                    catch (Exception e)
+                    {
+                        //TODO What can we do in this situation??
```
ayende: Nothing

src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -284,24 +358,58 @@ private int GetNextLogNumberForToday()
...
-                // we use GetFiles because we don't expect to have a massive amount of files, and it is 
-                // not sure what kind of iteration order we get if we run and modify using Enumerate
-                existingLogFiles = Directory.GetFiles(_path, "*.log");
+                logFiles = Directory.GetFiles(_path, "*.log.gz");
             }
-            catch (Exception)
+            catch (IOException)
             {
-                return; // this can fail for various reasons, we don't really care for that in most cases
+                //Probably because a network error has occurred.
```
ayende: Network error is _really_ strange. We typically log to local disk

src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -522,6 +631,85 @@ private void BackgroundLogger()
...
+            {
+                string[] logFiles;
+                try
+                {
+                    logFiles = Directory.GetFiles(_path, "*.log");
+                }
+                catch (IOException)
+                {
+                    //Probably because a network error has occurred.
+                    Thread.Sleep(TimeSpan.FromMinutes(1));
```
ayende: That is a LOT
And network error is rare

ayende: You are also blocking the shutdown of the server while this is blocked

src/Sparrow/Logging/LoggingSource.cs
haludi: and what about when the file in use? should I continue deleting the next log?
To delete until we rich the limit of log size or it should be considered as we deleted it?

src/Sparrow/Logging/LoggingSource.cs
haludi: yes, when the path to the log directory is on a remote computer.
This is the only exception I can think we can suddenly get from `GetFIiles` after we already use it
(accept to when someone deleting the directory and create a file with the same name)

src/Sparrow/Logging/LoggingSource.cs
ayende: Ignore it, we'll be retrying this anyway

src/Sparrow/Logging/LoggingSource.cs
ayende: There are a lot of things that this can be.
Out of file descriptors, I/O errors, permissions issues, etc.

src/Raven.Server/Config/Categories/LogsConfiguration.cs
```diff
@@ -26,10 +26,23 @@ public class LogsConfiguration : ConfigurationCategory
...
+        [Description("How far back we should retain log entries in days")]
+        [DefaultValue(null)]
+        [MinValue(1)]
+        [TimeUnit(TimeUnit.Days)]
+        [ConfigurationEntry("Logs.RetentionTimeInDays", ConfigurationEntryScope.ServerWideOnly)]
+        public TimeSetting? RetentionTime { get; set; }
+
+        [Description("The maximum size of the log after which the old files will be deleted")]
+        [DefaultValue(null)]
+        [MinValue(256)]
```
ppekrol, line 38: Just a sanity check here... having this attribute and providing `null` will not set the min value, right?

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -25,7 +26,19 @@ public class SecurityConfiguration : ConfigurationCategory
...
+
+        [Description("The maximum size of the audit log after which the old files will be deleted")]
+        [DefaultValue(null)]
+        [MinValue(256)]
+        [SizeUnit(SizeUnit.Megabytes)]
+        [ConfigurationEntry("Logs.RetentionSizeInMb", ConfigurationEntryScope.ServerWideOnly)]
```
ppekrol: Needs to be `Security.AuditLog.RetentionSizeInMb`

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -25,7 +26,19 @@ public class SecurityConfiguration : ConfigurationCategory
...
+        public Size? AuditLogRetentionSize { get; set; }
+
+        [Description("Will determine whether to compress the audit log files")]
+        [DefaultValue(false)]
+        [ConfigurationEntry("Logs.Compress", ConfigurationEntryScope.ServerWideOnly)]
```
ppekrol: Same here. Wrong prefix.

src/Raven.Server/Program.cs
```diff
@@ -81,8 +81,15 @@ public static int Main(string[] args)
             configuration.Initialize();
 
             LoggingSource.UseUtcTime = configuration.Logs.UseUtcTime;
-            LoggingSource.MaxFileSizeInBytes = configuration.Logs.MaxFileSize.GetValue(SizeUnit.Bytes);
-            LoggingSource.Instance.SetupLogMode(configuration.Logs.Mode, configuration.Logs.Path.FullPath, configuration.Logs.RetentionTime.AsTimeSpan);
+            LoggingSource.Instance.MaxFileSizeInBytes = configuration.Logs.MaxFileSize.GetValue(SizeUnit.Bytes);
+            LoggingSource.AuditLog.MaxFileSizeInBytes = configuration.Logs.MaxFileSize.GetValue(SizeUnit.Bytes);
```
ppekrol: shouldn't this be moved to `ConfigureAuditLog` method?

src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -74,6 +81,8 @@ public class WebSocketContext
 
         public LogMode LogMode { get; private set; }
         public TimeSpan RetentionTime { get; private set; }
+        public long RetentionSize { get; private set; }
+        public bool Compress => _compressLoggingThread != null;
```
ppekrol: rename to `Compressing`

src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -204,28 +220,57 @@ private void StartNewLoggingThread()
...
+            string[] logGzFiles;
+            try
+            {
+                logFiles = Directory.GetFiles(_path, $"{_dateString}.*.log");
+                logGzFiles = Directory.GetFiles(_path, $"{_dateString}.*.log.gz");
+            }
+            catch (Exception)
+            {
+                // Something went wrong we will try again later
+                return null;
```
ppekrol: This will cause NREs. We need to continue the execution of this method, but ignore the cleanup parts that use both of those variables

src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -239,12 +284,61 @@ private FileStream GetNewStream(long maxFileSize)
...
 
+        private void LimitLogSize(string[] logFiles, string[] logGzFiles)
+        {
+            var logFilesInfo = logFiles.Select(f => new FileInfo(f));
+            var logGzFilesInfo = logGzFiles.Select(f => new FileInfo(f));
+            var totalLogSize = logFilesInfo.Sum(i => i.Length) + logGzFilesInfo.Sum(i => i.Length);
+
+            foreach (var log in logGzFilesInfo.Reverse())
+            {
+                if (totalLogSize > RetentionSize)
```
ppekrol: This should be before both of the loops to avoid checking that each time

ppekrol (changes requested):

src/Raven.Server/Config/Categories/LogsConfiguration.cs
haludi: Do you mean if I can set the value below 256 because of the `DefaultValue` is set to null?

src/Sparrow/Logging/LoggingSource.cs
haludi: Actually, I forgot to decrement the `totalLogSize` value every deletion...

src/Raven.Server/Config/Categories/LogsConfiguration.cs
ppekrol: No, what will happen when I will set the value to `null` explicitly in `settings.json` ?

src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -423,7 +557,14 @@ private void BackgroundLogger()
                     try
                     {
                         var maxFileSize = MaxFileSizeInBytes;
-                        using (var currentFile = GetNewStream(maxFileSize))
+                        if (GetNewStream(maxFileSize, out var currentFile) == false)
```
ppekrol: I do not like this approach, because logging will stop entirely when we will not be able to get the list of files in the directory. I think it would be safer just to omit the cleanup part. Also let's rename the method, now it does a bit more than the name suggests.

src/Sparrow/Logging/LoggingSource.cs
ppekrol: Can we have a test for this? We can create new instance of LoggingSource inside a test and check if everything is working as expected.

src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -282,26 +386,39 @@ private int GetNextLogNumberForToday()
...
-                // we use GetFiles because we don't expect to have a massive amount of files, and it is 
-                // not sure what kind of iteration order we get if we run and modify using Enumerate
-                existingLogFiles = Directory.GetFiles(_path, "*.log");
-            }
-            catch (Exception)
+            var now = DateTime.Now;
+            foreach (var logFile in logFiles)
             {
-                return; // this can fail for various reasons, we don't really care for that in most cases
+                var logDateTime = File.GetLastWriteTime(logFile);
```
ppekrol: I do not like this. We were extracting a date from file name here, for me it was better approach than rely on the last write time.

PR #9205, opened 2019-04-29
fix for RavenDB-8243

myarichuk:
Refactor code to use centralized timeout settings for TCP connections

### DIFF
--- a/src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
+++ b/src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
@@ -196,14 +196,19 @@ namespace Raven.Client.Documents.Subscriptions
                         tcpInfo = await LegacyTryGetTcpInfo(requestExecutor, context, token).ConfigureAwait(false);
                     }
                 }
+                
+                string chosenUrl; 
+                (_tcpClient, chosenUrl) = await TcpUtils.ConnectAsyncWithPriority(tcpInfo, 
+                    requestExecutor.DefaultTimeout.GetValueOrDefault(RequestExecutor.GlobalHttpClientTimeout), 
+                    requestExecutor.DefaultTimeout.GetValueOrDefault(RequestExecutor.GlobalHttpClientTimeout)).ConfigureAwait(false);
 
-                string chosenUrl;
-                (_tcpClient, chosenUrl) = await TcpUtils.ConnectAsyncWithPriority(tcpInfo, requestExecutor.SendTimeout, requestExecutor.ReceiveTimeout).ConfigureAwait(false);
                 _tcpClient.NoDelay = true;
                 _tcpClient.SendBufferSize = _options?.SendBufferSizeInBytes ?? SubscriptionWorkerOptions.DefaultSendBufferSizeInBytes;
                 _tcpClient.ReceiveBufferSize = _options?.ReceiveBufferSizeInBytes ?? SubscriptionWorkerOptions.DefaultReceiveBufferSizeInBytes;
                 _stream = _tcpClient.GetStream();
-                _stream = await TcpUtils.WrapStreamWithSslAsync(_tcpClient, tcpInfo, _store.Certificate, requestExecutor.SendTimeout, requestExecutor.ReceiveTimeout).ConfigureAwait(false);
+                _stream = await TcpUtils.WrapStreamWithSslAsync(_tcpClient, tcpInfo, _store.Certificate, 
+                    requestExecutor.DefaultTimeout.GetValueOrDefault(RequestExecutor.GlobalHttpClientTimeout), 
+                    requestExecutor.DefaultTimeout.GetValueOrDefault(RequestExecutor.GlobalHttpClientTimeout)).ConfigureAwait(false);
 
                 var databaseName = _dbName ?? _store.Database;
 
--- a/src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
+++ b/src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
@@ -198,12 +198,12 @@ namespace Raven.Client.Documents.Subscriptions
                 }
 
                 string chosenUrl;
-                (_tcpClient, chosenUrl) = await TcpUtils.ConnectAsyncWithPriority(tcpInfo, requestExecutor.DefaultTimeout).ConfigureAwait(false);
+                (_tcpClient, chosenUrl) = await TcpUtils.ConnectAsyncWithPriority(tcpInfo, requestExecutor.SendTimeout, requestExecutor.ReceiveTimeout).ConfigureAwait(false);
                 _tcpClient.NoDelay = true;
                 _tcpClient.SendBufferSize = _options?.SendBufferSizeInBytes ?? SubscriptionWorkerOptions.DefaultSendBufferSizeInBytes;
                 _tcpClient.ReceiveBufferSize = _options?.ReceiveBufferSizeInBytes ?? SubscriptionWorkerOptions.DefaultReceiveBufferSizeInBytes;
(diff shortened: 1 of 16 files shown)


ayende (approved):

maximburyak:
@myarichuk, please take a look at the failing test

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -109,6 +109,19 @@ public string Url
 
         internal string LastServerVersion;
 
+        public TimeSpan SendTimeout
```
ppekrol: This is TCP related. I do not see a reason for us having that in RequestExecutor

src/Raven.Server/Config/Categories/ServerConfiguration.cs
```diff
@@ -6,6 +6,18 @@ namespace Raven.Server.Config.Categories
 {
     public class ServerConfiguration : ConfigurationCategory
     {
+        [Description("Receive timeout for all TCP connections")]
+        [DefaultValue(30)]
+        [TimeUnit(TimeUnit.Seconds)]
+        [ConfigurationEntry("Server.Tcp.ReceiveTimeout", ConfigurationEntryScope.ServerWideOrPerDatabase)]
```
ppekrol: Should be `Server.Tcp.ReceiveTimeoutInSec` and be `ServerWideOnly`

src/Raven.Server/Config/Categories/ServerConfiguration.cs
```diff
@@ -6,6 +6,18 @@ namespace Raven.Server.Config.Categories
...
+        public TimeSetting ReceiveTimeout { get; set; }
+
+        [Description("Sending timeout for all TCP connections")]
+        [DefaultValue(30)]
+        [TimeUnit(TimeUnit.Seconds)]
+        [ConfigurationEntry("Server.Tcp.SendTimeout", ConfigurationEntryScope.ServerWideOrPerDatabase)]
```
ppekrol: Should be `Server.Tcp.SendTimeoutInSec` and be `ServerWideOnly`

src/Raven.Server/Config/Categories/ClusterConfiguration.cs
```diff
@@ -61,10 +61,11 @@ public class ClusterConfiguration : ConfigurationCategory
...
         public TimeSetting MoveToRehabGraceTime{ get; set; }
         
-        [Description("Tcp connection read/write timeout.")]
+        [Description("Tcp connection read/write timeout. (obsolete, use Server.Tcp.SendTimeout and Server.Tcp.ReceiveTimeout instead")]
         [DefaultValue(15 * 1000)]
         [TimeUnit(TimeUnit.Milliseconds)]
-        [ConfigurationEntry("Cluster.TcpTimeoutInMs", ConfigurationEntryScope.ServerWideOnly)]
+        [ConfigurationEntry("Server.Tcp.SendTimeout", ConfigurationEntryScope.ServerWideOrPerDatabase)]
+        [ConfigurationEntry("Cluster.TcpTimeoutInMs", ConfigurationEntryScope.ServerWideOrPerDatabase)] //changed to ServerWideOrPerDatabase to prevent incompatibility
         public TimeSetting TcpConnectionTimeout { get; set; }
```
ppekrol: You removed all the occurences of this I think? What would be the reason to keep that configuration?

ppekrol: Where is this used now?

src/Raven.Server/Config/Categories/ServerConfiguration.cs
```diff
@@ -6,6 +6,18 @@ namespace Raven.Server.Config.Categories
...
     public class ServerConfiguration : ConfigurationCategory
     {
+        [Description("Receive timeout for all TCP connections")]
+        [DefaultValue(30)]
+        [TimeUnit(TimeUnit.Seconds)]
+        [ConfigurationEntry("Server.Tcp.ReceiveTimeout", ConfigurationEntryScope.ServerWideOrPerDatabase)]
+        public TimeSetting ReceiveTimeout { get; set; }
+
+        [Description("Sending timeout for all TCP connections")]
+        [DefaultValue(30)]
```
ppekrol, line 16: This changes previous 15 * 1000 (ms) to 30 sec. Do we want that? cc @karmeli87

src/Raven.Client/Http/RequestExecutor.cs
myarichuk: We need a timeout configuration in the RequestExecutor for TCP - and we need it BEFORE we initialize the TcpClient

ppekrol: Still it is not related in any way to RequestExecutor. Maybe it can be moved to conventions, but only place where this is used in Client API is subscriptions and I'm not sure if this is something that we want to have configurable in the Client API

src/Raven.Server/Config/Categories/ClusterConfiguration.cs
myarichuk: It is used by users, possibly.
Perhaps we should throw a "not supported" exception (with informative error) if it is set a value so users will know that this config value is no longer valid?

src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
```diff
@@ -196,14 +196,19 @@ await requestExecutor.ExecuteAsync(_redirectNode, null, context, command, should
                         tcpInfo = await LegacyTryGetTcpInfo(requestExecutor, context, token).ConfigureAwait(false);
                     }
                 }
+                
+                string chosenUrl; 
+                (_tcpClient, chosenUrl) = await TcpUtils.ConnectAsyncWithPriority(tcpInfo, 
+                    requestExecutor.DefaultTimeout.GetValueOrDefault(RequestExecutor.GlobalHttpClientTimeout),
```
ppekrol, line 202: we should not use `GlobalHttpClientTimeout`. It is set to 12h.

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1310,6 +1310,8 @@ protected virtual Task PerformHealthCheck(ServerNode serverNode, int nodeIndex,
         protected Task _firstTopologyUpdate;
         protected string[] _lastKnownUrls;
         private readonly DisposeOnce<ExceptionRetry> _disposeOnceRunner;
+        private TimeSpan? _sendTimeout;
```
ppekrol, line 1313: can be removed?

src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
ppekrol: We should use default TcpClient timeouts then?

src/Raven.Server/Config/Categories/ClusterConfiguration.cs
ppekrol: Let's remove this configuration property entirely.

ppekrol (changes requested):

src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
myarichuk: Thought about it, but at line 201 we have not yet created TcpClient.
Perhaps we should add the timeouts to client Conventions? 
The DefaultTimeout is there I think, is there a reason we shouldn't have RequestTimeout and ResponseTimeout there?

src/Raven.Client/Http/RequestExecutor.cs
myarichuk: probably yes. I will move Request/Response timout to Conventions I think

PR #9209, opened 2019-04-30
RavenDB-13449 fix

myarichuk:
Make sure to register notification event only once per Raft command (even if it is a batch command that executes multiple Raft commands under the same index)

### DIFF
--- a/src/Raven.Server/ServerWide/ClusterStateMachine.cs
+++ b/src/Raven.Server/ServerWide/ClusterStateMachine.cs
@@ -43,6 +43,7 @@ using Raven.Server.ServerWide.Context;
 using Raven.Server.Utils;
 using Sparrow;
 using Sparrow.Binary;
+using Sparrow.Collections;
 using Sparrow.Json;
 using Sparrow.Json.Parsing;
 using Sparrow.Logging;
@@ -244,7 +245,7 @@ namespace Raven.Server.ServerWide
                             throw new RachisApplyException($"'{nameof(PutSubscriptionBatchCommand.Commands)}' is missing in '{nameof(PutSubscriptionBatchCommand)}'.");
                         }
                         foreach (BlittableJsonReaderObject command in subscriptionCommands)
-                        {
+                        {                            
                             Apply(context, command, index, leader, serverStore);
                         }
                         break;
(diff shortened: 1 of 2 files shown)


src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1442,26 +1443,43 @@ public override void EnsureNodeRemovalOnDeletion(TransactionOperationContext con
             _parent.InsertToLeaderLog(context, term, context.ReadObject(djv, "remove"), RachisEntryFlags.StateMachineCommand);
         }
 
+        private readonly ConcurrentSet<long> _scheduledValueChangedNotificationIndexes = new ConcurrentSet<long>();
+
         private void NotifyValueChanged(TransactionOperationContext context, string type, long index)
         {
-            context.Transaction.InnerTransaction.LowLevelTransaction.OnDispose += transaction =>
+            //in case of batches that do multiple changes with the same index, we need to notify about those changes only once
+            if (_scheduledValueChangedNotificationIndexes.TryAdd(index))
```
karmeli87, line 1451: What if you have a batch command that contains 3 commands.
After committing every command does something asyncly (`ExecuteAsyncTask`).
So you will notify the waiter that you are done only after the first command is completed. not all three, right?

ayende: All of them should happen in the same tx, no?
In that case, we'll register for this only once, and notify once the tx is completely committed

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1442,26 +1443,43 @@ public override void EnsureNodeRemovalOnDeletion(TransactionOperationContext con
...
 
-        private void NotifyDatabaseAboutChanged(TransactionOperationContext context, string databaseName, long index, string type, DatabasesLandlord.ClusterDatabaseChangeType change)
+        private readonly ConcurrentSet<long> _scheduledDatabaseChangedNotificationIndexes = new ConcurrentSet<long>();
+
+        private void NotifyDatabaseAboutChanged(TransactionOperationContext context, string databaseName, long index, string type,
+            DatabasesLandlord.ClusterDatabaseChangeType change)
         {
-            context.Transaction.InnerTransaction.LowLevelTransaction.OnDispose += transaction =>
+            //in case of batches that do multiple changes with the same index, we need to notify about those changes only once
+            if (_scheduledDatabaseChangedNotificationIndexes.TryAdd(index))
```
ayende, line 1471: What clears this value if the transaction _failed_ ?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
karmeli87: After the tx is committed, we run an async task.
So I have a command to notify db1 and another command (with the same index) to notify db2.
In this case, db2 wouldn't be notified.

Can't we use the `ExecuteManyOnDispose` for this purpose?

ayende: _can_ we have a batch command on mulitiple databases?

karmeli87: I don't know if we are using it this way. 
But I don't see a reason why not.

ppekrol:
Closed via: https://github.com/ravendb/ravendb/pull/9239

PR #9210, opened 2019-04-30
RavenDB-13487_v4.1: Migrating from v3.5 RavenFS throw NRE

garayx:


src/Raven.Server/Smuggler/Migration/Migrator_V3.cs
```diff
@@ -167,10 +173,10 @@ private async Task<string> MigrateRavenFs(string lastEtag)
                         WriteDocumentWithAttachment(documentActions, context, dataStream, key, metadata);
 
                         Parameters.Result.Documents.ReadCount++;
-                        if (Parameters.Result.Documents.Attachments.ReadCount % 50 == 0 || sp.ElapsedMilliseconds > 3000)
+                        if (Parameters.Result.Documents.ReadCount % 50 == 0 || sp.ElapsedMilliseconds > 3000)
```
grisha-kotler, line 176: The attachment count needs to incremented

grisha-kotler (changes requested):

src/Raven.Server/Smuggler/Migration/Migrator_V3.cs
garayx: :+1:

grisha-kotler (approved):

src/Raven.Server/Smuggler/Migration/Migrator_V2.cs
```diff
@@ -108,7 +115,7 @@ private async Task MigrateAttachments(string lastEtag)
                     var attachmentsArray = await GetAttachmentsList(lastEtag, transactionOperationContext);
                     if (attachmentsArray.Length == 0)
                     {
-                        var count = Parameters.Result.Documents.Attachments.ReadCount;
+                        var count = Parameters.Result.Documents.ReadCount;
```
ppekrol, line 118: This looks bizarre? Why are we not using the Attachments here?

grisha-kotler: It is the same number. Migrating legacy attachment creates a new document + attachment.

garayx: we write the attachment as dummy doc therefore its proper to increase docs count.
https://github.com/ravendb/ravendb/blob/v4.1/src/Raven.Server/Smuggler/Migration/AbstractLegacyMigrator.cs#L104

ppekrol: How this is displayed in the Studio? Shouldn't we increase both numbers?

garayx: we increase the docs count here and the attachments count increased by DatabaseItemType.Attachments so in studio we see both docs and attachments
https://github.com/ravendb/ravendb/blob/v4.1/src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs#L171

PR #9213, opened 2019-05-02
RavenDB-11856 Informative exception were added if reset of ceched properties occured while writing a document

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-11856

### DIFF
--- a/src/Sparrow/Json/CachedProperties.cs
+++ b/src/Sparrow/Json/CachedProperties.cs
@@ -347,7 +347,7 @@ namespace Sparrow.Json
         }
 
         private bool _hasDuplicates;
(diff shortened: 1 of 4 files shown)


src/Sparrow/Json/CachedProperties.cs
```diff
@@ -347,8 +347,7 @@ private void UpdatePropertiesSortOrder()
         }
 
         private bool _hasDuplicates;
-
-
+        internal int DocNo { get; private set; }
```
ayende: What is this property? I assume that this is meant to be `DocumentNumber`?

PR #9216, opened 2019-05-02
RavenDB-13448 Prevent adding new node from a different schema

karmeli87:
…ne by one instead of snapshot


aviviadi (approved):
This gives `Raven.Client.Exceptions.RavenException : System.InvalidOperationException: Can't add node with mismatched storage schema version.` 
We still need to make the test work by "upgrading" or command one by one replicating to a different schema server of another version

karmeli87:
This is the expected exception in this case. 

When we are adding a new node we can't do a command one by one, because we truncate the log.. so the commands aren't there any longer.

So the proper way would be to build a 4.1 cluster and upgrade the nodes. (we truncate the log only when _all_ nodes had applied the command)
So it could send the commands one by one.

src/Raven.Client/ServerWide/Commands/Cluster/AddClusterNodeCommand.cs
```diff
@@ -5,15 +5,15 @@
 
 namespace Raven.Client.ServerWide.Commands.Cluster
 {
-    internal class AddClusterNodeCommand : RavenCommand
+    public class AddClusterNodeCommand : RavenCommand
```
ppekrol: internal :)

src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
```diff
@@ -349,6 +351,13 @@ public async Task AddNode()
 
                 nodeInfo = infoCmd.Result;
 
+                if (SchemaUpgrader.CurrentVersion.ServerVersion != nodeInfo.ServerSchemaVersion)
```
ppekrol, line 354: What if we have a 4.2 cluster with schema 15 and we want to add a new node with schema 16 (also 4.2). This should be supported, right? We cannot only add node with 'older' schema?

karmeli87: No it shouldn't... you don't know what changes there were.
for example changing tree to table.. like we did with identities.

PR #9219, opened 2019-05-02
RavenDB-13476 - add TableRenaming option

grisha-kotler:


src/Voron/Impl/Transaction.cs
```diff
@@ -128,16 +128,16 @@ public void EndAsyncCommit()
...
             Slice nameSlice;
             using (Slice.From(Allocator, name, ByteStringType.Immutable, out nameSlice))
             {
-                return OpenTable(schema, nameSlice);
+                return OpenTable(schema, nameSlice, doSchemaValidation);
             }
         }
 
-        public Table OpenTable(TableSchema schema, Slice name)
+        public Table OpenTable(TableSchema schema, Slice name, bool doSchemaValidation = false)
```
ayende, line 140: What are you doing with this param?

src/Voron/Impl/Transaction.cs
```diff
@@ -311,6 +311,74 @@ public void DeleteTree(Slice name)
...
+        public Table RenameTable(string fromName, string toName, bool doSchemaValidation = false)
+        {
+            Slice.From(Allocator, fromName, ByteStringType.Immutable, out var fromNameSlice);
+            Slice.From(Allocator, toName, ByteStringType.Immutable, out var toNameSlice);
+            return RenameTable(fromNameSlice, toNameSlice, doSchemaValidation);
+        }
+
+        public Table RenameTable(Slice fromName, Slice toName, bool doSchemaValidation = false)
+        {
+            if (_lowLevelTransaction.Flags == TransactionFlags.ReadWrite == false)
```
ayende, line 324:
```suggestion
            if (_lowLevelTransaction.Flags != TransactionFlags.ReadWrite)
```

src/Voron/Impl/Transaction.cs
```diff
@@ -311,6 +311,74 @@ public void DeleteTree(Slice name)
...
+                var treeName = item.Key;
+
+                using (toTable.DirectAdd(treeName, sizeof(TreeRootHeader), out var ptr))
+                {
+                    var header = (TreeRootHeader*)ptr;
+                    tree.State.CopyTo(header);
+                }
+            }
+
+            _lowLevelTransaction.RootObjects.Delete(fromName);
```
ayende, line 373: We need to validate that there are no global indexes here, no?

grisha-kotler:
Moving this to 4.2.x since this isn't currently being used

PR #9220, opened 2019-05-02
RavenDB-13506 Fix that certificates wasn't appear when adding a new n…

karmeli87:
…ode to the cluster

### DIFF
--- a/src/Raven.Server/ServerWide/ClusterStateMachine.cs
+++ b/src/Raven.Server/ServerWide/ClusterStateMachine.cs
@@ -1626,9 +1626,9 @@ namespace Raven.Server.ServerWide
 
         private readonly (ByteString Entry, int Version)[] _snapshotEntries =
         {
-            (Items.Content, 40),
-            (CompareExchange.Content, 40),
-            (Identities.Content, 40),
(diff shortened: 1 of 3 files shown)


karmeli87:
Certificates and Compare Exchange Tombstones had the same issue.
The test verifies only the Compare Exchange Tombstones, because it is easier to test.

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1624,23 +1624,31 @@ private void UpdateEtagForBackup(DatabaseRecord databaseRecord, string type, lon
             }
         }
 
-        public override bool ShouldSnapshot(Slice slice, RootObjectType type)
+        private readonly (ByteString Entry, int Version)[] _snapshotEntries =
         {
-            var baseVersion = slice.Content.Match(Items.Content)
-                            || slice.Content.Match(CompareExchange.Content)
-                            || slice.Content.Match(Identities.Content);
+            (Items.Content, 40),
```
ppekrol: Shouldn't this be `Base40CommandsVersion` ?

PR #9228, opened 2019-05-05
RavenDB-13432: Included counters should affect the query ETag

garayx:


grisha-kotler:
Need a test with counter deletion

ayende (commented):
How is this going to _handle_ counter deletion?

src/Raven.Server/Documents/Queries/Dynamic/CollectionQueryRunner.cs
```diff
@@ -203,19 +204,20 @@ private unsafe void FillCountOfResultsAndIndexEtag(QueryResultServerSide<Documen
                 var numberOfDocuments = Database.DocumentsStorage.GetNumberOfDocuments(context);
                 buffer[0] = DocumentsStorage.ReadLastDocumentEtag(context.Transaction.InnerTransaction);
                 buffer[1] = DocumentsStorage.ReadLastTombstoneEtag(context.Transaction.InnerTransaction);
-                buffer[2] = numberOfDocuments;
+                buffer[2] = DocumentsStorage.ReadLastCountersEtag(context.Transaction.InnerTransaction);
```
ppekrol: This needs to only be applied when we have include on counters inside the query. We need to detect that and put that information inside the QueryMetadata. If we will not do that then counters will impact query caching mechanism pretty badly.

ppekrol (changes requested):

src/Raven.Server/Documents/Queries/Dynamic/CollectionQueryRunner.cs
garayx: done

grisha-kotler:
Need a test for: 
Query1: from Users where Name = $p0 (Params: $p0 = "Egor")
Query2: from Users where Name = $p0 (Params: $p0 = "Grisha")

The second query will return 304 (before the change)

ppekrol (approved):

src/Raven.Client/Documents/Session/AbstractDocumentQuery.CounterIncludes.cs
```diff
@@ -25,15 +24,20 @@ public abstract partial class AbstractDocumentQuery<T, TSelf>
...
-
                 if (kvp.Value.Counters?.Count > 0 == false)
                     continue;
 
-                CounterIncludesTokens.Add(CounterIncludesToken.Create(
-                    kvp.Key,
-                    kvp.Value.Counters.Count == 1
-                        ? AddQueryParameter(kvp.Value.Counters.First())
-                        : AddQueryParameter(kvp.Value.Counters)));
+                if (kvp.Value.Counters.Count == 1)
```
ayende: We do you have a special  for a single item? Isn't this going to end up with the same thing in the end?

ayende (commented):
How are tombstones for counters handled?

src/Raven.Client/Documents/Session/AbstractDocumentQuery.CounterIncludes.cs
garayx: done

garayx:
> How are tombstones for counters handled?

when we delete a counter we increment the etag, so the next query(after delete) will get 'fresh' results.
there is a test for this.

PR #9259, opened 2019-05-12
RavenDB-13535 An item with the same key has already been added on outgoing replication

karmeli87:
The issue here is that reading the last etag on database startup yield wrong value, because the attachment voron index for reading the last etag wasn't global.

The case where this could happened:
1. Sending a replication batch where the attachment is the last item 
2. The destination restarts the database after receiving that batch.


karmeli87:
This fix also require schema upgrade, but since it affects both 4.1 and 4.2 we can't simply bump the version and need a more generic solution, which is discussed here: https://issues.hibernatingrhinos.com/issue/RavenDB-13542

ayende (changes requested):
Shouldn't you be handling the migration of the schema here?

karmeli87:
I realized that we keep the attachments metadata in a single table (and not per collection)
So we may avoid changing the schema.

PR #9265, opened 2019-05-14
RavenDB-13528 / RavenDB-13555 Argument of WriteFile function - nNumbe…

arekpalinski:
…rOfBytesToWrite is actually DWORD (A 32-bit unsigned integer. The range is 0 through 4294967295 decimal). So we can write almost 4GB at once although the value passed there needs to be 4KB aligned.


ayende (approved):

src/Raven.Pal/src/win/fileutils.c
```diff
@@ -74,7 +74,7 @@ PRIVATE int32_t
 _write_file(void* handle, const void* buffer, int64_t size, int64_t offset, int32_t* detailed_error_code)
 {
     const int32_t WRITE_INCREMENT = 4096;
-    const int32_t NUMBER_OF_BYTES_TO_WRITE = (INT32_MAX / WRITE_INCREMENT) * WRITE_INCREMENT;
+    const int32_t NUMBER_OF_BYTES_TO_WRITE = (UINT32_MAX / WRITE_INCREMENT) * WRITE_INCREMENT;
```
arekpalinski, line 77: @ayende Do you think that it can be a potential source of problems on 32bits? (that we switched from max 2GB to 4GB)

ayende: Shouldn't matter, you can't pass it there anyway in 32 bits

arekpalinski: Okay. I just wanted to make sure.

PR #9271, opened 2019-05-15
RavenDB-1111 - Making sure exception got from reading the response of…

grisha-kotler:
… the failed replication request won't override the actual exception. We will log both.


ayende (commented):
where are you logging both if there is a read failure? I'm seeing you log the reading error, but not the actual exception that caused it

grisha-kotler:
fixed

PR #9273, opened 2019-05-15
 RavenDB-13556 Improve client interaction with the cluster [WIP]

karmeli87:
Since this is a large PR the work is splitted into 2 commits:

1. enforcing the guid to all raft command (many files and some tests)
2. core changes

This is still WIP because the client side is still missing, each operation (that make sense) has to be altered to add a guid to it.

Only the cluster tx implemented the guid on the client-side for testing purpose.

### DIFF
--- a/src/Raven.Client/Documents/Operations/OngoingTasks/ToggleOngoingTaskStateOperation.cs
+++ b/src/Raven.Client/Documents/Operations/OngoingTasks/ToggleOngoingTaskStateOperation.cs
@@ -56,6 +56,7 @@ namespace Raven.Client.Documents.Operations.OngoingTasks
             }
 
             public override bool IsReadRequest => false;
+            public override bool IsClusterCommand => true;
         }
     }
 }
--- a/src/Raven.Client/Http/RequestExecutor.cs
+++ b/src/Raven.Client/Http/RequestExecutor.cs
@@ -988,8 +988,19 @@ namespace Raven.Client.Http
 
             request.RequestUri = new Uri(url);
 
-            if (!request.Headers.Contains(Constants.Headers.ClientVersion))
+            if (request.Headers.Contains(Constants.Headers.ClientVersion) == false)
                 request.Headers.Add(Constants.Headers.ClientVersion, ClientVersion);
+
+            if (command.IsClusterCommand)
+            {
+                if (command.Guid == null)
+                {
+                    command.Guid = Guid.NewGuid().ToString();
+                }
+
+                if (request.Headers.Contains(Constants.Headers.RaftCommandGuid) == false)
+                    request.Headers.Add(Constants.Headers.RaftCommandGuid, command.Guid);
+            }
 
             return request;
         }
--- a/src/Raven.Server/Documents/DatabaseRequestHandler.cs
+++ b/src/Raven.Server/Documents/DatabaseRequestHandler.cs
@@ -45,8 +45,9 @@ namespace Raven.Server.Documents
         protected delegate void RefAction(string databaseName, ref BlittableJsonReaderObject configuration, JsonOperationContext context);
 
         protected async Task DatabaseConfigurations(Func<TransactionOperationContext, string,
-           BlittableJsonReaderObject, Task<(long, object)>> setupConfigurationFunc,
+           BlittableJsonReaderObject, string, Task<(long, object)>> setupConfigurationFunc,
            string debug,
+           string guid,
            RefAction beforeSetupConfiguration = null,
            Action<DynamicJsonValue, BlittableJsonReaderObject, long> fillJson = null,
            HttpStatusCode statusCode = HttpStatusCode.OK)
@@ -64,7 +65,7 @@ namespace Raven.Server.Documents
                 var configurationJson = await context.ReadForMemoryAsync(RequestBodyStream(), debug);
                 beforeSetupConfiguration?.Invoke(Database.Name, ref configurationJson, context);
 
-                var (index, _) = await setupConfigurationFunc(context, Database.Name, configurationJson);
+                var (index, _) = await setupConfigurationFunc(context, Database.Name, configurationJson, guid);
                 DatabaseRecord dbRecord;
                 using (context.OpenReadTransaction())
                 {
--- a/src/Raven.Server/Documents/Handlers/BatchHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/BatchHandler.cs
@@ -188,8 +188,10 @@ namespace Raven.Server.Documents.Handlers
 
         private async Task HandleClusterTransaction(DocumentsOperationContext context, MergedBatchCommand command, ClusterTransactionCommand.ClusterTransactionOptions options)
         {
+            var guid = GetStringQueryString("guid", required: false);
+
             var record = ServerStore.LoadDatabaseRecord(Database.Name, out _);
-            var clusterTransactionCommand = new ClusterTransactionCommand(Database.Name, record.Topology.DatabaseTopologyIdBase64, command.ParsedCommands, options);
+            var clusterTransactionCommand = new ClusterTransactionCommand(Database.Name, record.Topology.DatabaseTopologyIdBase64, command.ParsedCommands, options, guid);
             var result = await ServerStore.SendToLeaderAsync(clusterTransactionCommand);
 
             if (result.Result is List<string> errors)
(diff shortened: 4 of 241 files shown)


src/Raven.Client/Documents/Operations/OngoingTasks/ToggleOngoingTaskStateOperation.cs
```diff
@@ -56,6 +56,7 @@ public override void SetResponse(JsonOperationContext context, BlittableJsonRead
             }
 
             public override bool IsReadRequest => false;
+            public override bool IsClusterCommand => true;
```
ayende, line 59: I wonder if it wouldn't be more appropriate to handle this via a marker interface, instead of a property.

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -988,9 +988,20 @@ private HttpRequestMessage CreateRequest<TResult>(JsonOperationContext ctx, Serv
...
                 request.Headers.Add(Constants.Headers.ClientVersion, ClientVersion);
 
+            if (command.IsClusterCommand)
+            {
+                if (command.Guid == null)
+                {
+                    command.Guid = Guid.NewGuid().ToString();
+                }
+
+                if (request.Headers.Contains(Constants.Headers.RaftCommandGuid) == false)
```
ayende, line 1001: That isn't a good idea. We need to have the Guid directly in the command json itelf.
Certain commands may do a wire hop. Go over from node A to node B (which is the leader) and we want to Guid to be part of that.
For that matter, not crazy about the `Guid` as a name. A UniqueCommandId is better

src/Raven.Server/Documents/DatabaseRequestHandler.cs
```diff
@@ -45,8 +45,9 @@ public override void Init(RequestHandlerContext context)
         protected delegate void RefAction(string databaseName, ref BlittableJsonReaderObject configuration, JsonOperationContext context);
 
         protected async Task DatabaseConfigurations(Func<TransactionOperationContext, string,
```
ayende, line 47: Let's go ahead and declare a `delegate` here.

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -188,8 +188,10 @@ private void BatchTrafficWatch(ArraySegment<BatchRequestParser.CommandData> pars
 
         private async Task HandleClusterTransaction(DocumentsOperationContext context, MergedBatchCommand command, ClusterTransactionCommand.ClusterTransactionOptions options)
         {
+            var guid = GetRaftGuidFromHeaders();
```
ayende, line 191: This is a great example of why we don't want it in the headers. 
It is very easy to forget to do that in various places, and lose the protection it gives us

src/Raven.Server/Documents/Handlers/BatchRequestParser.cs
```diff
@@ -241,7 +241,7 @@ private static unsafe bool ReadClusterTransactionProperty(JsonParserState state)
         private static async Task GetIdentitiesValues(JsonOperationContext ctx, DocumentDatabase database, ServerStore serverStore,
             List<string> identities, List<int> positionInListToCommandIndex, CommandData[] cmds)
         {
-            var newIds = await serverStore.GenerateClusterIdentitiesBatchAsync(database.Name, identities);
+            var newIds = await serverStore.GenerateClusterIdentitiesBatchAsync(database.Name, identities, Guid.NewGuid().ToString());
```
ayende, line 244: Why is the guid a string? Why no use an actual guid here?

src/Raven.Server/Documents/Handlers/IndexHandler.cs
```diff
@@ -571,7 +571,7 @@ public async Task SetLockMode()
 
                 foreach (var name in parameters.IndexNames)
                 {
-                    await Database.IndexStore.SetLock(name, parameters.Mode);
+                    await Database.IndexStore.SetLock(name, parameters.Mode, string.Empty); // TODO: can't use single guid here for multiple commands
```
ayende, line 574: This is actually a good _reason_ to use a string for the unique id.
You can append to it `id + "/" + i`

src/Raven.Server/Rachis/Leader.cs
```diff
@@ -710,19 +710,42 @@ private void EmptyQueue()
...
+                            if (lastCommitted >= index) 
+                            {
+                                if (exception != null)
+                                {
+                                    cmd.Tcs.TrySetException(new Exception(exception));
+                                }
+                                else
+                                {
+                                    result = GetConvertResult(cmd.Command)?.Apply(result) ?? cmd.Command.FromRemote(result);
+                                    var completed = new ValueTask<(long, object)>((index, result));
```
ayende, line 725: Why create value task just to create a task immediately?

src/Raven.Server/Rachis/Leader.cs
```diff
@@ -1083,6 +1107,7 @@ public class CommandState
             public ConvertResultAction ConvertResult;
             public TaskCompletionSource<(long, object)> TaskCompletionSource;
             public Action<TaskCompletionSource<(long, object)>> OnNotify;
+            public string Debug;
```
ayende, line 1110: what's that?

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -312,6 +335,146 @@ static RachisConsensus()
...
+        {
+            var guid = GetGuidFromCommand(cmd);
+            if (guid == null) // shouldn't happened in new cluster version!
+                return;
+
+            if (guid == string.Empty)
+                return;
+
+            var table = context.Transaction.InnerTransaction.OpenTable(LogHistoryTable, LogHistorySlice);
+            var type = GetTypeFromCommand(cmd);
```
ayende, line 362: why do you care about the type?

ayende (changes requested):

src/Raven.Client/Documents/Operations/OngoingTasks/ToggleOngoingTaskStateOperation.cs
karmeli87: The reasons I did it as a property is:
1. Not to forget any command
2. Enforce future command not to forget it :-)

src/Raven.Client/Http/RequestExecutor.cs
karmeli87: The reason I've passed it via the headers is because it is easier to append it on the request.
Wasn't aware that redirecting doesn't preserve the headers...

src/Raven.Server/Documents/Handlers/BatchHandler.cs
karmeli87: But even when passed via the url it has to be retrieved.. so it is the same, no?

src/Raven.Server/Documents/Handlers/IndexHandler.cs
karmeli87: Nice idea! keep the uniqueness and the injective property.

src/Raven.Server/Rachis/RachisConsensus.cs
karmeli87: mainly for debug... if we will ever wonder what the hell this command was about, at least we will have the type.

src/Raven.Client/Documents/Operations/OngoingTasks/ToggleOngoingTaskStateOperation.cs
ayende: 1) That required to change a lot of unrelated code.
2) Can do that more easily by validating the commands that go to the cluster directly.

src/Raven.Client/Http/RequestExecutor.cs
ayende: Putting it in header / url means that it is outside of the data itself. We need it _in_ the command.

src/Raven.Server/Documents/Handlers/BatchHandler.cs
ayende: Yes, so don't put it in the URL. Make it in the command body itself.

karmeli87:
Not relevant for 4.1

PR #9277, opened 2019-05-16
[RavenDB-12658] refactoring + tests

Or-Geva:


grisha-kotler (approved):

src/Raven.Client/Documents/Session/DocumentSessionAttachmentsBase.cs
```diff
@@ -71,20 +71,30 @@ public void Store(string documentId, string name, Stream stream, string contentT
...
             Store(document.Id, name, stream, contentType);
         }
 
-        protected void ThrowEntityNotInSession(object entity)
+        protected void ThrowEntityNotInSessionOrMissingId(object entity)
         {
             throw new ArgumentException($"{entity} is not associated with the session. Use documentId instead or track the entity in the session.", nameof(entity));
         }
 
+        protected void ThrowEntityNotInSession(object entity)
```
ppekrol, line 87: I do not like it, create 2 overloads

ppekrol (changes requested):

Or-Geva:
Better error message when trying to get names of attachment from a document on in the session

src/Raven.Client/Documents/Session/DocumentSessionAttachmentsBase.cs
Or-Geva: I cannot overload with an object,
pulled out the if to implementation of getNames

src/Raven.Client/Documents/Session/DocumentSessionAttachmentsBase.cs
```diff
@@ -27,6 +27,9 @@ public AttachmentName[] GetNames(object entity)
             if (entity == null)
                 throw new ArgumentNullException(nameof(entity));
 
+            if (entity is string)
+                throw new ArgumentException($"You need to track the entity in the session for document id {entity}.", nameof(entity));
```
ppekrol: Can we be more explicit here? I've read that few times and I'm still having issues to understand what would be the action that I should take to not get this exception when this method is used.

Or-Geva: GetNames is the only one that has no overload by 'documentId', as opposed to other attachments' methods such as get, store, etc.

ppekrol: I understand, but the exception message is not 100% clear

Or-Geva: Done.

ppekrol (approved):

PR #9282, opened 2019-05-20
 RavenDB-13556 Improve client interaction with the cluster

karmeli87:
This PR touch many files, so I've split it into several logical commits.

The most interesting one, is probably the "core changes".

### DIFF
--- a/src/Raven.Client/Documents/Commands/Batches/BatchCommand.cs
+++ b/src/Raven.Client/Documents/Commands/Batches/BatchCommand.cs
@@ -12,7 +12,7 @@ using Sparrow.Json;
 
 namespace Raven.Client.Documents.Commands.Batches
 {
-    public class BatchCommand : RavenCommand<BatchCommandResult>, IDisposable
+    public class BatchCommand : RavenCommand<BatchCommandResult>, IRaftCommand, IDisposable
     {
         private readonly BlittableJsonReaderObject[] _commands;
         private readonly List<Stream> _attachmentStreams;
@@ -121,7 +121,7 @@ namespace Raven.Client.Documents.Commands.Batches
             if (_options == null)
                 return;
 
-            sb.AppendLine("?");
+            sb.Append("?");
 
             var replicationOptions = _options.ReplicationOptions;
             if (replicationOptions != null)
@@ -159,5 +159,7 @@ namespace Raven.Client.Documents.Commands.Batches
             foreach (var command in _commands)
                 command?.Dispose();
         }
+
+        public string RaftUniqueRequestId { get; } = Guid.NewGuid().ToString();
     }
 }
--- a/src/Raven.Client/Documents/Commands/Batches/BatchCommand.cs
+++ b/src/Raven.Client/Documents/Commands/Batches/BatchCommand.cs
@@ -160,6 +160,6 @@ namespace Raven.Client.Documents.Commands.Batches
                 command?.Dispose();
         }
 
-        public string RaftUniqueRequestId { get; } = RaftIdGenerator.NewId;
+        public string RaftUniqueRequestId { get; } = RaftIdGenerator.NewId();
     }
 }
--- a/src/Raven.Client/Documents/Commands/Batches/BatchCommand.cs
+++ b/src/Raven.Client/Documents/Commands/Batches/BatchCommand.cs
@@ -160,6 +160,6 @@ namespace Raven.Client.Documents.Commands.Batches
                 command?.Dispose();
         }
 
-        public string RaftUniqueRequestId { get; } = Guid.NewGuid().ToString();
+        public string RaftUniqueRequestId { get; } = RaftIdGenerator.NewId;
     }
 }
--- a/src/Raven.Client/Documents/Commands/Batches/BatchCommand.cs
+++ b/src/Raven.Client/Documents/Commands/Batches/BatchCommand.cs
@@ -8,6 +8,7 @@ using Raven.Client.Documents.Session;
 using Raven.Client.Http;
 using Raven.Client.Json;
 using Raven.Client.Json.Converters;
+using Raven.Client.Util;
 using Sparrow.Json;
 
 namespace Raven.Client.Documents.Commands.Batches
--- a/src/Raven.Client/Documents/Commands/DropSubscriptionConnectionCommand.cs
+++ b/src/Raven.Client/Documents/Commands/DropSubscriptionConnectionCommand.cs
@@ -4,7 +4,7 @@ using Sparrow.Json;
 
 namespace Raven.Client.Documents.Commands
 {
-    public class DropSubscriptionConnectionCommand:RavenCommand
+    public class DropSubscriptionConnectionCommand : RavenCommand
     {
         private readonly string _name;
 
@@ -12,6 +12,9 @@ namespace Raven.Client.Documents.Commands
         {
             _name = name;
         }
+
+        
+
         public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, ServerNode node, out string url)
         {
             url = $"{node.Url}/databases/{node.Database}/subscriptions/drop?name={_name}";
--- a/src/Raven.Client/Documents/Commands/DropSubscriptionConnectionCommand.cs
+++ b/src/Raven.Client/Documents/Commands/DropSubscriptionConnectionCommand.cs
@@ -13,8 +13,6 @@ namespace Raven.Client.Documents.Commands
             _name = name;
         }
 
-        
-
         public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, ServerNode node, out string url)
         {
             url = $"{node.Url}/databases/{node.Database}/subscriptions/drop?name={_name}";
--- a/src/Raven.Client/Http/RequestExecutor.cs
+++ b/src/Raven.Client/Http/RequestExecutor.cs
@@ -1030,11 +1030,21 @@ namespace Raven.Client.Http
         {
             var request = command.CreateRequest(ctx, node, out url);
 
-            request.RequestUri = new Uri(url);
+            var builder = new UriBuilder(url);
 
-            if (!request.Headers.Contains(Constants.Headers.ClientVersion))
+            if (command is IRaftCommand raftCommand)
+            {
+                Debug.Assert(raftCommand.RaftUniqueRequestId != null, $"Forget to create an id for {command.GetType()}?");
+
+                var raftRequestString = "raft-request-id=" + raftCommand.RaftUniqueRequestId;
+                builder.Query = builder.Query?.Length > 1 ? $"{builder.Query.Substring(1)}&{raftRequestString}" : raftRequestString;
+            }
+
+            request.RequestUri = builder.Uri;
+
+            if (request.Headers.Contains(Constants.Headers.ClientVersion) == false)
                 request.Headers.Add(Constants.Headers.ClientVersion, ClientVersion);
-
+            
             return request;
         }
 
--- a/src/Raven.Server/Commercial/LicenseManager.cs
+++ b/src/Raven.Server/Commercial/LicenseManager.cs
@@ -250,7 +250,7 @@ namespace Raven.Server.Commercial
             return _serverStore.GetClusterTopology().AllNodes.Count;
         }
 
-        public async Task ChangeLicenseLimits(string nodeTag, int newAssignedCores, string guid)
+        public async Task ChangeLicenseLimits(string nodeTag, int newAssignedCores, string raftRequestId)
         {
             if (_serverStore.IsLeader() == false)
                 throw new InvalidOperationException("Only the leader is allowed to change the license limits");
@@ -340,7 +340,7 @@ namespace Raven.Server.Commercial
                     };
                 }
 
-                await _serverStore.PutLicenseLimitsAsync(licenseLimits, guid);
+                await _serverStore.PutLicenseLimitsAsync(licenseLimits, raftRequestId);
             }
             finally
             {
@@ -429,7 +429,7 @@ namespace Raven.Server.Commercial
             }
         }
 
-        public async Task Activate(License license, bool skipLeaseLicense, string guid, bool ensureNotPassive = true, bool forceActivate = false)
+        public async Task Activate(License license, bool skipLeaseLicense, string raftRequestId, bool ensureNotPassive = true, bool forceActivate = false)
         {
             var newLicenseStatus = GetLicenseStatus(license);
             if (newLicenseStatus.Expiration.HasValue == false)
@@ -438,7 +438,7 @@ namespace Raven.Server.Commercial
             if (forceActivate == false)
             {
                 if (await ContinueActivatingLicense(
-                        license, skipLeaseLicense, guid, 
+                        license, skipLeaseLicense, raftRequestId, 
                         ensureNotPassive, 
                         newLicenseStatus).ConfigureAwait(false) == false)
                     return;
@@ -451,7 +451,7 @@ namespace Raven.Server.Commercial
 
                 try
                 {
-                    await _serverStore.PutLicenseAsync(license, guid).ConfigureAwait(false);
+                    await _serverStore.PutLicenseAsync(license, raftRequestId).ConfigureAwait(false);
 
                     SetLicense(license.Id, newLicenseStatus.Attributes);
                 }
@@ -492,7 +492,7 @@ namespace Raven.Server.Commercial
             };
         }
 
-        private async Task<bool> ContinueActivatingLicense(License license, bool skipLeaseLicense, string guid, bool ensureNotPassive, LicenseStatus newLicenseStatus)
+        private async Task<bool> ContinueActivatingLicense(License license, bool skipLeaseLicense, string raftRequestId, bool ensureNotPassive, LicenseStatus newLicenseStatus)
         {
             if (newLicenseStatus.Expired)
             {
@@ -509,7 +509,7 @@ namespace Raven.Server.Commercial
                     license = await GetUpdatedLicenseInternal(license);
                     if (license != null)
                     {
-                        await Activate(license, skipLeaseLicense: true, guid, ensureNotPassive: ensureNotPassive);
+                        await Activate(license, skipLeaseLicense: true, raftRequestId, ensureNotPassive: ensureNotPassive);
                         return false;
                     }
                 }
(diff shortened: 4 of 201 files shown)


src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1025,11 +1025,21 @@ private HttpRequestMessage CreateRequest<TResult>(JsonOperationContext ctx, Serv
...
             var request = command.CreateRequest(ctx, node, out url);
 
-            request.RequestUri = new Uri(url);
+            var builder = new UriBuilder(url);
 
-            if (!request.Headers.Contains(Constants.Headers.ClientVersion))
-                request.Headers.Add(Constants.Headers.ClientVersion, ClientVersion);
+            if (command is IRaftCommand raftCommand)
+            {
+                Debug.Assert(raftCommand.RaftUniqueRequestId != null, $"Forget to create an id for {command.GetType()}?");
```
ayende, line 1037: Also check for empty?

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1025,11 +1025,21 @@ private HttpRequestMessage CreateRequest<TResult>(JsonOperationContext ctx, Serv
...
+
+                var raftRequestString = "raft-request-id=" + raftCommand.RaftUniqueRequestId;
+                builder.Query = builder.Query?.Length > 1 ? $"{builder.Query.Substring(1)}&{raftRequestString}" : raftRequestString;
```
ayende, line 1040: Why are you skipping the first char in the existing `Query` ?

src/Raven.Server/Commercial/LicenseManager.cs
```diff
@@ -429,7 +429,7 @@ private void AssertCanAssignCores(int assignedCores)
             }
         }
 
-        public async Task Activate(License license, bool skipLeaseLicense, bool ensureNotPassive = true, bool forceActivate = false)
+        public async Task Activate(License license, bool skipLeaseLicense, string guid, bool ensureNotPassive = true, bool forceActivate = false)
```
ayende: `guid` of what? It looks like this related to the license id. Better to name it properly

src/Raven.Server/Commercial/LicenseManager.cs
```diff
@@ -703,7 +704,7 @@ private async Task ExecuteTasks()
             }
         }
 
-        public async Task LeaseLicense(bool forceUpdate = false)
+        public async Task LeaseLicense(string guid, bool forceUpdate = false)
```
ayende:
```suggestion
        public async Task LeaseLicense(string raftReqId, bool forceUpdate = false)
```

Or something like that

src/Raven.Server/Documents/DatabaseRequestHandler.cs
```diff
@@ -44,9 +44,11 @@ public override void Init(RequestHandlerContext context)
 
         protected delegate void RefAction(string databaseName, ref BlittableJsonReaderObject configuration, JsonOperationContext context);
 
-        protected async Task DatabaseConfigurations(Func<TransactionOperationContext, string,
-           BlittableJsonReaderObject, Task<(long, object)>> setupConfigurationFunc,
+        protected delegate Task<(long, object)> SetupFunc(TransactionOperationContext context, string databaseName, BlittableJsonReaderObject json, string guid);
+
+        protected async Task DatabaseConfigurations(SetupFunc setupConfigurationFunc,
            string debug,
+           string guid,
```
ayende: Meaningful name

src/Raven.Server/Documents/Handlers/IndexHandler.cs
```diff
@@ -638,14 +639,15 @@ public async Task SetLockMode()
...
+            var guid = GetRaftRequestIdFromQuery();
             using (ContextPool.AllocateOperationContext(out DocumentsOperationContext context))
             {
                 var json = await context.ReadForMemoryAsync(RequestBodyStream(), "index/set-priority");
                 var parameters = JsonDeserializationServer.Parameters.SetIndexPriorityParameters(json);
 
                 foreach (var name in parameters.IndexNames)
                 {
-                    await Database.IndexStore.SetPriority(name, parameters.Priority);
+                    await Database.IndexStore.SetPriority(name, parameters.Priority, $"{guid}/{name}");
```
ayende: Are they guaranteed to be unique? Also elsewhere where you do this

src/Raven.Server/Documents/PeriodicBackup/Restore/RestoreBackupTask.cs
```diff
@@ -180,7 +180,8 @@ public async Task<IOperationResult> Execute(Action<IOperationProgress> onProgres
                         // we are currently restoring, shouldn't try to access it
                         databaseRecord.DatabaseState = DatabaseStateStatus.RestoreInProgress;
 
-                        var (index, _) = await _serverStore.WriteDatabaseRecordAsync(databaseName, databaseRecord, null, restoreSettings.DatabaseValues, isRestore: true);
+                        // TODO: does string.Empty make sense here?
```
ayende: Probably not, get the req id from the top?

src/Raven.Server/Rachis/Leader.cs
```diff
@@ -710,19 +710,41 @@ private void EmptyQueue()
...
 
-                        var index = _engine.InsertToLeaderLog(context, Term, cmdJson, RachisEntryFlags.StateMachineCommand);
+                        if (_engine.HasHistoryLog(context, cmdJson, out var index, out var result, out var exception))
+                        {
+                            // if this command is already committed, we can skip it and notify the caller about it
+                            if (lastCommitted >= index) 
+                            {
+                                if (exception != null)
+                                {
+                                    cmd.Tcs.TrySetException(new Exception(exception));
```
ayende: I don't like this, however. You are throwing `Exception`, and we probably should be throwing the actual exception type that was raised.

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -282,9 +284,29 @@ private set
         private static readonly Slice TopologySlice;
         private static readonly Slice TagSlice;
         private static readonly Slice SnapshotRequestSlice;
-
+        internal static readonly Slice LogHistorySlice;
+        private static readonly Slice LogHistoryIndexSlice;
         internal static readonly Slice EntriesSlice;
         internal static readonly TableSchema LogsTable;
+        internal static readonly TableSchema LogHistoryTable;
+        private const int LogHistoryMaxEntries = 2048;
```
ayende: This should be configurable

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -282,9 +284,29 @@ private set
...
+        internal enum LogHistoryColumn
+        {
```
ayende: For debug purposes, also include the time.
We can also expose this as an admin debug endpoint, which can be really useful.

src/Raven.Server/Web/RequestHandler.cs
```diff
@@ -250,6 +250,27 @@ protected bool IsWebsocketRequest()
             return HttpContext.WebSockets.IsWebSocketRequest;
         }
 
+        protected string GetRaftRequestIdFromQuery()
+        {
+            var guid = GetStringQueryString("raft-request-id", required: false);
+
+            if (RequestRouter.TryGetClientVersion(HttpContext, out var clientVersion))
+            {
+                if (clientVersion.Build <= 42_008) // TODO: what should be the behaviour with the nightly version
```
ayende: The nightly also have a version indication, no?

src/Raven.Server/Web/RequestHandler.cs
```diff
@@ -250,6 +250,27 @@ protected bool IsWebsocketRequest()
...
+            if (RequestRouter.TryGetClientVersion(HttpContext, out var clientVersion))
+            {
+                if (clientVersion.Build <= 42_008) // TODO: what should be the behaviour with the nightly version 
+                {
+                    guid = guid ?? string.Empty;
+                }
+            }
+
+#if DEBUG
+            return guid;
```
ayende, line 267: I don't understand why you have this?

ayende: If the version match, and no request id provided, should throw

src/Raven.Client/Http/RequestExecutor.cs
karmeli87: I keep the empty for "don't care"

src/Raven.Client/Http/RequestExecutor.cs
karmeli87: followed this : https://docs.microsoft.com/en-us/dotnet/api/system.uribuilder.query?view=netframework-4.8

src/Raven.Server/Documents/Handlers/IndexHandler.cs
karmeli87: guid is unique.. as long as the name of the index is also unique (or any other suffix for the particular operation), it should be fine.

src/Raven.Server/Rachis/Leader.cs
karmeli87: Agree, but I keep the exception as a string and not sure how to convert it back to the original one.

src/Raven.Server/Web/RequestHandler.cs
karmeli87: IIRC nightlys are simply 42

src/Raven.Server/Web/RequestHandler.cs
karmeli87: I want to avoid throwing any exception in `release`. 
In order not to break any client..

src/Raven.Server/Documents/Handlers/IndexHandler.cs
ayende: I don't think we _require_ that the name will be unique, though.
Better to add an index number there too

src/Raven.Server/Rachis/Leader.cs
ayende: Keep the exception time and create it dynamically?

src/Raven.Server/Web/RequestHandler.cs
ayende: Right now you aren't throwing anything anyway, though?

src/Raven.Server/Documents/Handlers/IndexHandler.cs
karmeli87: Would it make sense or even possible to pass 2 indexes with the same name in the same operation?

ayende: Sense, no? 
Possible...

src/Raven.Server/Web/RequestHandler.cs
karmeli87: If the guid is null, I do throw before appending to the log (both release and debug)
I want to throw only in debug builds, in release even if we have a bug we should put `string.Empty`.

ayende (approved):

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -313,6 +342,225 @@ static RachisConsensus()
...
+                yield return ReadHistoryLog(context, entryHolder);
+            }
+        }
+
+        private unsafe DynamicJsonValue ReadHistoryLog(TransactionOperationContext context, Table.TableValueHolder entryHolder)
+        {
+            var djv = new DynamicJsonValue();
+
+            var ticks = Bits.SwapBytes(*(long*)entryHolder.Reader.Read((int)(LogHistoryColumn.Ticks), out _));
+            djv["Date"] = new DateTime(ticks);
```
ppekrol: nameof

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -313,6 +342,225 @@ static RachisConsensus()
...
+
+            var guidPtr = entryHolder.Reader.Read((int)(LogHistoryColumn.Guid), out var size);
+            djv[LogHistoryColumn.Guid.ToString()] = Encoding.UTF8.GetString(guidPtr, size);
```
ppekrol: nameof

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -313,6 +342,225 @@ static RachisConsensus()
...
+
+            var index = Bits.SwapBytes(*(long*)entryHolder.Reader.Read((int)(LogHistoryColumn.Index), out _));
+            djv[LogHistoryColumn.Index.ToString()] = index;
```
ppekrol: nameof

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -313,6 +342,225 @@ static RachisConsensus()
...
+
+            djv[LogHistoryColumn.Term.ToString()] = *(long*)entryHolder.Reader.Read((int)(LogHistoryColumn.Term), out _);
```
ppekrol: nameof

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -313,6 +342,225 @@ static RachisConsensus()
...
+
+            var typeString = entryHolder.Reader.Read((int)(LogHistoryColumn.Type), out size);
+            djv[LogHistoryColumn.Type.ToString()] = Encoding.UTF8.GetString(typeString, size);
```
ppekrol: nameof

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -313,6 +342,225 @@ static RachisConsensus()
...
+
+            djv[LogHistoryColumn.State.ToString()] = (*(HistoryStatus*)entryHolder.Reader.Read((int)(LogHistoryColumn.State), out _)).ToString();
```
ppekrol: nameof

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -313,6 +342,225 @@ static RachisConsensus()
...
+
+            var resultPtr = entryHolder.Reader.Read((int)(LogHistoryColumn.Result), out size);
+            if (size > 0)
+            {
+                var blittableResult = new BlittableJsonReaderObject(resultPtr, size, context);
+                djv[LogHistoryColumn.Result.ToString()] = blittableResult.ToString();
```
ppekrol: nameof

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -313,6 +342,225 @@ static RachisConsensus()
...
+                blittableResult.Dispose();
+            }
+            else
+            {
+                djv[LogHistoryColumn.Result.ToString()] = null;
```
ppekrol: nameof

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -313,6 +342,225 @@ static RachisConsensus()
...
+            }
+
+            var exTypePtr = entryHolder.Reader.Read((int)(LogHistoryColumn.ExceptionType), out size);
+            djv[LogHistoryColumn.ExceptionType.ToString()] = size > 0 ? Encoding.UTF8.GetString(exTypePtr, size) : null;
```
ppekrol: nameof

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -313,6 +342,225 @@ static RachisConsensus()
...
+
+            var exMsg = entryHolder.Reader.Read((int)(LogHistoryColumn.ExceptionMessage), out size);
+            djv[LogHistoryColumn.ExceptionMessage.ToString()] = size > 0 ? Encoding.UTF8.GetString(exMsg, size) : null;
```
ppekrol: nameof

src/Raven.Server/ServerWide/Commands/DeleteDatabaseCommand.cs
```diff
@@ -14,12 +14,11 @@ public class DeleteDatabaseCommand : UpdateDatabaseCommand
         public string[] FromNodes;
         public bool UpdateReplicationFactor = true;
 
-        public DeleteDatabaseCommand() : base(null)
+        public DeleteDatabaseCommand()
```
ppekrol, line 17: ErrorOnDatabaseDoesNotExists will be `false` now?

src/Raven.Server/Web/RequestHandler.cs
ppekrol: Should be `42_009` ?

src/Raven.Client/Documents/Commands/Batches/BatchCommand.cs
```diff
@@ -159,5 +159,7 @@ public void Dispose()
             foreach (var command in _commands)
                 command?.Dispose();
         }
+
+        public string RaftUniqueRequestId { get; } = Guid.NewGuid().ToString();
```
ppekrol: Since we are using that all over the place and there are some magic string here e.g. `string.Empty`. I would say that we need to create internal static class called `RaftIdentifier` with two methods:
1) `NewId` which will return this
2) `OtherNameThatExplainsWhatItDoes` which will return `string.Empty`

src/Raven.Client/Documents/Commands/DropSubscriptionConnectionCommand.cs
```diff
@@ -4,14 +4,17 @@
...
+    public class DropSubscriptionConnectionCommand : RavenCommand
     {
         private readonly string _name;
 
         public DropSubscriptionConnectionCommand(string name)
         {
             _name = name;
         }
+
+
```
ppekrol: what happened here? :)

src/Raven.Client/Documents/Commands/Batches/BatchCommand.cs
ppekrol: This will also help centralize the identifiers.

src/Raven.Server/ServerWide/Commands/DeleteDatabaseCommand.cs
karmeli87: This `ctor` is only for de-serialization. No code path uses it. 
What we do missing is to include the `ErrorOnDatabaseDoesNotExists ` in the `ToJson` of the `UpdateDatabaseCommand`

ppekrol: Depends for what purpose it is used. You know better.

ppekrol (approved):

ppekrol:
Rebased.

ayende (approved):

PR #9298, opened 2019-05-22
RavenDB-12674 Renew License

Danielle9897:


grisha-kotler (approved):

ppekrol (approved):

src/Raven.Studio/typescript/viewmodels/shell/registration.ts
```diff
@@ -182,6 +226,45 @@ class registration extends dialogViewModelBase {
...
+                            !this.renewWhenNotExpired() ) {
+                                app.closeDialog(this);
+                        }
+                        license.fetchSupportCoverage();
+                    });
+                
+                if (result.Error) {
+                    this.renewError(result.Error);
+                    this.renewMessage(null);
+                }
```
ml054: format code:
} else {

src/Raven.Studio/wwwroot/Content/css/modals.less
```diff
@@ -224,6 +224,12 @@
 }
 
 #licenseModal {
+    .licenseModalImg {
```
ml054, line 227: why do we need this?

src/Raven.Studio/typescript/viewmodels/shell/registration.ts
Danielle9897: done

src/Raven.Studio/wwwroot/Content/css/modals.less
Danielle9897: well - we don't have to - I just thought the image looks nicer in the dialog when it is a bit **smaller**

ml054 (approved):

PR #9300, opened 2019-05-23
RavenDB-13542 redesigned schema upgrades

ppekrol:


arekpalinski (approved):

ayende:
I'm pretty sure that we shouldn't be merging this to 4.1, _maybe_ 4.2 ?

arekpalinski:
We need to introduce it to both 4.1 and 4.2. 

This PR is for 4.1, for 4.2 we'll need to have a separate one. Merging this PR and preparing the one for 4.2 can wait since we don't have a need to introduce any schema upgrade to 4.1

ayende:
Di we need to merge it to 4.1? I don't think we are going to be doing much on it going forward

arekpalinski:
Yes, we need to merge it eventually (not right now) because we need to have consistent naming of schema upgrades between any 4.x version. Once Pawel will be back we'll get back to this and decide if we want to introduce this at this point - IMO I think we're good

arekpalinski (approved):

ppekrol:
Moved this to 5.0 and bumped all storages to 50_000

PR #9307, opened 2019-05-26
RDBS-11 Retrieving ids for a patch by index modified to use information from Lucene without accessing documents table

haludi:

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -2288,168 +2288,239 @@ namespace Raven.Server.Documents.Indexes
                 throw new NotSupportedException("Explanations are not supported by this type of query.");
 
             using (var marker = MarkQueryAsRunning(query, token))
+            using (var queryArea = await OpenTransactionsForNonStaleResultIfNeeded(query, documentsContext, marker))
             {
-                var queryDuration = Stopwatch.StartNew();
-                AsyncWaitForIndexing wait = null;
-                (long? DocEtag, long? ReferenceEtag)? cutoffEtag = null;
+                FillQueryResult(resultToFill, queryArea.IsStale, query.Metadata, documentsContext, queryArea.Context);
 
-                var stalenessScope = query.Timings?.For(nameof(QueryTimingsScope.Names.Staleness), start: false);
-
-                while (true)
+                using (var reader = IndexPersistence.OpenIndexReader(queryArea.Context.Transaction.InnerTransaction))
+                using (var queryScope = query.Timings?.For(nameof(QueryTimingsScope.Names.Query)))
                 {
-                    AssertIndexState();
-                    marker.HoldLock();
+                    QueryTimingsScope gatherScope = null;
+                    QueryTimingsScope fillScope = null;
 
-                    // we take the awaiter _before_ the indexing transaction happens, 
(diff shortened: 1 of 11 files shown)


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -2288,168 +2288,239 @@ public IndexDefinition GetIndexDefinition()
...
+                }
+
+                return;
+            }
+        }
 
-                                var totalResults = new Reference<int>();
-                                var skippedResults = new Reference<int>();
-                                IncludeCountersCommand includeCountersCommand = null;
+        protected class QueryContext : IDisposable
```
ayende, line 2380: Can you make this a struct? Will this cause allocations?

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -2288,168 +2288,239 @@ public IndexDefinition GetIndexDefinition()
...
+            public TransactionOperationContext Context;
+            public bool IsStale;
+            public IDisposable ReturnContextToPool { get; set; }
 
-                                var fieldsToFetch = new FieldsToFetch(query, Definition);
+            public void Dispose()
+            {
+                using (ReturnContextToPool)
+                {
+                    Context.CloseTransaction();
```
ayende, line 2390: IIRC, doesn't this already happen?

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -2288,168 +2288,239 @@ public IndexDefinition GetIndexDefinition()
...
+                    GetOrAddSpatialField,
+                    token.Token);
+            }
 
-                                var retriever = GetQueryResultRetriever(query, queryScope, documentsContext, fieldsToFetch, includeDocumentsCommand);
+            return documents;
+        }
 
-                                IEnumerable<(Document Result, Dictionary<string, Dictionary<string, string[]>> Highlightings, ExplanationResult Explanation)> documents;
+        protected async Task<QueryContext> OpenTransactionsForNonStaleResultIfNeeded(
```
ayende, line 2460: Can we use a ValueTask here?

src/Raven.Server/Documents/Queries/DocumentQueryResult.cs
```diff
@@ -80,4 +80,44 @@ private static string CreateExplanation(Lucene.Net.Search.Explanation explanatio
             return explanation.ToString();
         }
     }
+
+    public class IdsQueryResult : QueryResultServerSide<string>
+    {
+        public Queue<string> Ids { get; } = new Queue<string>();
```
ayende, line 86: Why a queue and not a list?

src/Raven.Server/Documents/Indexes/Index.cs
haludi: As it was implemented so far the read transaction is closed after we retrieve all the ids
I just refactoring this so I can reuse the same code in query for documents and in query for ids

src/Raven.Server/Documents/Queries/DocumentQueryResult.cs
haludi: `ExecuteRateLimitedOperations` using a queue

ppekrol:
We will take a different approach on the subject. Please open a new PR when ready.

PR #9314, opened 2019-05-28
RavenDB-13617 At GetLower function, missing cast added in optimization

haludi:


ayende (approved):

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-13617

src/Raven.Server/Documents/DocumentIdWorker.cs
```diff
@@ -95,7 +95,10 @@ public static ByteStringContext.InternalScope GetLower(ByteStringContext byteStr
 
                 // PERF: Trick to avoid multiple compare instructions on hot loops. 
                 //       This is the same as (ch >= 65 && ch <= 90)
-                if (ch - 65 <= 90 - 65) 
+                const byte max = 90 - 65;
+                // We need to use byte here, to avoid automatic promotion to int32
+                var remaining = (byte)(ch - 65);
+                if (remaining <= max)
```
ppekrol: I think we also need to handle https://github.com/ravendb/ravendb/blob/9e63a0ad28db7bede401782cb2a882ef7d249e89/src/Raven.Server/Documents/DocumentIdWorker.cs#L214 ?

ppekrol: Inside `GetLowerIdSliceAndStorageKey`

ayende: yes!

redknightlois:
Need to see the assembly compare of both to be sure. That change may introduce another temporal that depending on the context may spill registers and kill the optimization impact for a different reason.

haludi:
before vs after
before
after

haludi:
We discussed it and there were arisen an assumption that this bug didn't reveal so far because it is internal
and a change to it can damage backward compatibility (to ids with @)
As I see it so far the second spot (at `GetLowerIdSliceAndStorageKey` function) that @ppekrol [mention earlier](https://github.com/ravendb/ravendb/pull/9314#discussion_r288004571) is not a problem because there is a cast to `uint` before (maybe this was the intention in the first spot).
The first spot at `GetSliceFromId` function is used to handle reference.
I am checking now the theory that the fix can destroy backward compatibility
so far it looks like this is not the case but I need more time to be sure.

haludi:
Also, I added here the disassembly of before ([second picture](https://user-images.githubusercontent.com/37696258/58535358-74060400-81f6-11e9-91e4-83846f51dc4e.png)) and after ([third picture](https://user-images.githubusercontent.com/37696258/58535360-74060400-81f6-11e9-953b-87c0c3f7a602.png)) the fix.
In addition, I apply the code from `GetLowerIdSliceAndStorageKey` to `GetSliceFromId` and compared it to my fix (int the PR), the code without optimization and the code with the bug.
The result after run each one 500,000,000 times are :

GetLowerIdSliceAndStorageKey  00:00:09.7379055
My fix                                           00:00:10.7540067
Without optimization                   00:00:09.4131990
With bug                                      00:00:09.9451849

haludi:
image

Disassembly of optimization from `GetLowerIdSliceAndStorageKey`

ayende:
At a glance, the assembly looks the same in both cases.

ayende:
I don't know why you see such numbers, but as I see it, removing the optimizations gives the right behavior without perf cost, no?

haludi:
image

I try to cover different scenarios to check the optimizations against the bug code and unoptimized code
I pass to `ToLower` function different strings
"person" - just lowers
"PERSON" - just upper
"p#####1" - has char under 'A' char (# = 35)
"P_____1" - ha char above 'Z' char (_ = 95)

* Other function code - is the solution in `GetLowerIdSliceAndStorageKey` function
* The time is loop of 5,000,000,000 in my machine
* Any string scenario ran in parallel for all versions
* The scenario ran in sequential

haludi:
In addition, the test ran on
<TargetFramework>netcoreapp2.1</TargetFramework>
<RuntimeFrameworkVersion>2.1.11</RuntimeFrameworkVersion>

redknightlois:
> n't know why you see such numbers, but as I see it, removing the optimizations gives the right behavior without perf cost, no?

The `movzx r10d, r10b` is likely the one because it adds a second read after write hazard there... The information in the issue doesn't allow me to understand the actual error and how it related to the promotion to `uint32` because under a compare scenario there shouldn't be any difference. 

Having said that, there has been some work on the JIT to recognize those cases which would make that optimization not needed anymore; however, I dont remember if it was in the 2.x or 3.0 branch. If it hits, we don't need it anymore as this is internal and we control the runtime. If I understood properly the comparison, that looks like to be the case.

ayende:
Let's go with the simpler code, then

haludi:
image

Disassembly without optimization

haludi:
I apply a simple solution to `GetLower` function. @redknightlois please review.
I also checked for the possibility this fix of internal bug will cause a problem with backward compatibility
and as I see it so far this is not the case. But I need more time to be sure (maybe in another issue?)

ayende (approved):

PR #9316, opened 2019-05-28
RavenDB-13584 Adding a test for node distribution when node goes down

talweiss1982:
Also added operation for setting the dynamic redistribution feature


test/SlowTests/Cluster/NodeRedistribution.cs
```diff
@@ -0,0 +1,43 @@
...
+            var notInDbGroup = cluster.Nodes.First(s => res.Servers.Contains(s) == false);
+            using (var store = new DocumentStore
+            {
+                Database = databaseName,
+                Urls = new[] { cluster.Leader.WebUrl }
+            }.Initialize())
+            {
+                store.Maintenance.Server.Send(new SetDatabaseDynamicDistribution(store.Database, true));
+                await DisposeAndRemoveServer(res.Servers.First());     
+                Thread.Sleep((moveToRehabGraceTime + replicaTimeout)* 2 * 1000);
```
ayende: Can we avoid the sleep somehow? Can we do an active check?
This is likely to fail if the cluster observer is a bit slow.

ayende (changes requested):

test/SlowTests/Cluster/NodeRedistribution.cs
talweiss1982: I can change it if you want

ayende: please do

talweiss1982: now it seems to fail on debug

ayende: why?

talweiss1982: not sure why

src/Raven.Client/ServerWide/Operations/SetDatabaseDynamicDistribution.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Net.Http;
+using System.Text;
+using Raven.Client.Documents.Conventions;
+using Raven.Client.Http;
+using Raven.Client.Util;
+using Sparrow.Json;
+
+namespace Raven.Client.ServerWide.Operations
+{
+    public class SetDatabaseDynamicDistribution : IServerOperation<bool>
```
ppekrol: rename to `SetDatabaseDynamicDistributionOperation`

src/Raven.Client/ServerWide/Operations/SetDatabaseDynamicDistribution.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        private bool _allowDynamicDistribution;
+        private string _databaseName;
+
+        public SetDatabaseDynamicDistribution(string databaseName, bool allowDynamicDistribution)
+        {
+            _allowDynamicDistribution = allowDynamicDistribution;
+            _databaseName = databaseName;
```
ppekrol: Let's assert if this is not null or empty

src/Raven.Client/ServerWide/Operations/SetDatabaseDynamicDistribution.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        public RavenCommand<bool> GetCommand(DocumentConventions conventions, JsonOperationContext context)
+        {
+            return new SetDatabaseDynamicDistributionCommand(_databaseName, _allowDynamicDistribution);
+        }
+    }
+
+    public class SetDatabaseDynamicDistributionCommand : RavenCommand<bool>, IRaftCommand
```
ppekrol: mark as private and put inside the operation class

src/Raven.Client/ServerWide/Operations/SetDatabaseDynamicDistribution.cs
ppekrol: Do we want to expose this as a part of public API? Maybe internal is enough?

src/Raven.Client/ServerWide/Operations/SetDatabaseDynamicDistribution.cs
talweiss1982: what is null?

src/Raven.Client/ServerWide/Operations/SetDatabaseDynamicDistribution.cs
talweiss1982: why not? I think it was just missing and it is not breaking anything

src/Raven.Client/ServerWide/Operations/SetDatabaseDynamicDistribution.cs
ppekrol: databaseName, string

src/Raven.Client/ServerWide/Operations/SetDatabaseDynamicDistributionOperation.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Net.Http;
+using System.Text;
+using Raven.Client.Documents.Conventions;
+using Raven.Client.Http;
+using Raven.Client.Util;
+using Sparrow.Json;
+
+namespace Raven.Client.ServerWide.Operations
+{
+    public class SetDatabaseDynamicDistributionOperation : IServerOperation<bool>
```
ppekrol: `IServerOperation<bool>` needs to be changed to `IServerOperation`
The endpoint does not return anything

src/Raven.Client/ServerWide/Operations/SetDatabaseDynamicDistribution.cs
ppekrol: ok

src/Raven.Client/ServerWide/Operations/SetDatabaseDynamicDistributionOperation.cs
talweiss1982: I missed that interface, was looking for it, was below the other one

ppekrol (approved):

PR #9324, opened 2019-05-28
RavenDB-13620 Moving disk size calculation to PAL

ayende:


src/Raven.Pal/pal.ver
```diff
@@ -1 +1 @@
-42007
+42014
```
ppekrol: Can we bump only by 1?

ayende: Yes, auto generated

ayende:
Wait, need to rebuilt the binaries, too

ppekrol:
Can we squash everything to one commit to avoid pushing DLLs few times?

ayende:
Yes, I'll rebase when done

aviviadi (approved):

PR #9325, opened 2019-05-29
RavenDB-13612 follower should report back the latest incoming entry index

karmeli87:
The underlying issue was that the follower reported back his latest entry, which might be from a different term. (if this node was more up-to-date in the previous term)
This would cause the leader and the follower to commit different entries for the same index.

another cause that increased the probability to this issue is that the leader started to negotiate before appending 'noop' to his own log.
that might lead to a more up-to-date follower not to rewind his log.

### DIFF
--- a/src/Raven.Server/Rachis/Elector.cs
+++ b/src/Raven.Server/Rachis/Elector.cs
@@ -24,6 +24,10 @@ namespace Raven.Server.Rachis
             _electorLongRunningWork = PoolOfThreads.GlobalRavenThreadPool.LongRunning(x => HandleVoteRequest(), null, $"Elector for candidate {_connection.Source}");
         }
 
+        public override string ToString()
+        {
+            return $"Elector {_engine.Tag} for {_connection.Source}";
+        }
         public void HandleVoteRequest()
         {
             try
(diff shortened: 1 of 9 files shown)


src/Raven.Server/Rachis/Elector.cs
```diff
@@ -203,13 +220,13 @@ public void HandleVoteRequest()
                                 continue;
                             }
 
-                            if (lastTerm == rv.LastLogTerm && lastIndex > rv.LastLogIndex)
+                            if (lastLogTerm == rv.LastLogTerm && lastLogIndex > rv.LastLogIndex)
```
ayende, line 223: Why is this being checked only for trial election?

src/Raven.Server/Rachis/Elector.cs
```diff
@@ -289,17 +306,33 @@ private class HandleVoteResult
...
             }
 
+            if (lastLogTermUnderWriteLock > rv.LastLogTerm)
+            {
+                result.DeclineVote = true;
+                result.DeclineReason = $"My last log term {lastLogTermUnderWriteLock}, is higher than yours {rv.LastLogTerm}.";
+                return result;
+            }
+
+            if (lastLogIndexUnderWriteLock > rv.LastLogIndex)
```
ayende, line 329: Okay, this is being checked here again for real elections, right?

src/Raven.Server/Rachis/Follower.cs
```diff
@@ -297,23 +294,23 @@ private async Task Concurrent_SendAppendEntriesPendingToLeaderAsync(Cancellation
                     }
                 }
 
-                lastLogIndex = _engine.GetLastEntryIndex(context);
-
                 var lastEntryIndexToCommit = Math.Min(
-                    lastLogIndex,
+                    _engine.GetLastEntryIndex(context),
```
ayende, line 298: Shouldn't we say, my last index entry or (if you gave me any entries, the last entry index you gave me) or the leader commit?

ayende (approved):

PR #9327, opened 2019-05-29
RavenDB-12677 Provide a way to manually create a revision of a document

Danielle9897:
Note: This commit solves both: RavenDB-12677 and RavenDB-13420

### DIFF
--- a/src/Raven.Client/Documents/Commands/Batches/ICommandData.cs
+++ b/src/Raven.Client/Documents/Commands/Batches/ICommandData.cs
@@ -32,16 +32,14 @@ namespace Raven.Client.Documents.Commands.Batches
         AttachmentCOPY,
         CompareExchangePUT,
         CompareExchangeDELETE,
-
+        ForceRevisionCreation, 
         Counters,
-
         BatchPATCH,
         // NOTE: When you add anything here, such as additional 
         // operation relating to a document but not operating on it 
         // directly, be sure to also update DeferInternal
         // to recognize that these are allows, like with AttachmentPUT and
         // AttachmentDELETE
-
         ClientAnyCommand,
         ClientModifyDocumentCommand
     }
--- a/src/Raven.Client/Documents/Commands/Batches/PutCommandDataBase.cs
+++ b/src/Raven.Client/Documents/Commands/Batches/PutCommandDataBase.cs
@@ -8,8 +8,8 @@ namespace Raven.Client.Documents.Commands.Batches
 {
     internal class PutCommandDataWithBlittableJson : PutCommandDataBase<BlittableJsonReaderObject>
     {
-        public PutCommandDataWithBlittableJson(string id, string changeVector, BlittableJsonReaderObject document, bool forceRevisionCreation = false)
-            : base(id, changeVector, document, forceRevisionCreation)
+        public PutCommandDataWithBlittableJson(string id, string changeVector, BlittableJsonReaderObject document, ForceRevisionStrategy forceRevisionCreationStrategy = ForceRevisionStrategy.None)
+            : base(id, changeVector, document, forceRevisionCreationStrategy)
         {
         }
 
@@ -20,8 +20,8 @@ namespace Raven.Client.Documents.Commands.Batches
 
     public class PutCommandData : PutCommandDataBase<DynamicJsonValue>
     {
-        public PutCommandData(string id, string changeVector, DynamicJsonValue document, bool forceRevisionCreation = false)
-            : base(id, changeVector, document, forceRevisionCreation)
+        public PutCommandData(string id, string changeVector, DynamicJsonValue document, ForceRevisionStrategy forceRevisionCreationStrategy = ForceRevisionStrategy.None)
+            : base(id, changeVector, document, forceRevisionCreationStrategy)
         {
         }
 
@@ -32,7 +32,7 @@ namespace Raven.Client.Documents.Commands.Batches
 
     public abstract class PutCommandDataBase<T> : ICommandData
     {
-        protected PutCommandDataBase(string id, string changeVector, T document, bool forceRevisionCreation = false)
+        protected PutCommandDataBase(string id, string changeVector, T document, ForceRevisionStrategy forceRevisionCreationStrategy = ForceRevisionStrategy.None)
         {
             if (document == null)
                 throw new ArgumentNullException(nameof(document));
@@ -40,7 +40,7 @@ namespace Raven.Client.Documents.Commands.Batches
             Id = id;
             ChangeVector = changeVector;
             Document = document;
-            ForceRevisionCreation = forceRevisionCreation;
+            ForceRevisionCreationStrategy = forceRevisionCreationStrategy;
         }
 
         public string Id { get; }
@@ -48,18 +48,24 @@ namespace Raven.Client.Documents.Commands.Batches
         public string ChangeVector { get; }
         public T Document { get; }
         public CommandType Type { get; } = CommandType.PUT;
-        public bool ForceRevisionCreation { get; }
+        public ForceRevisionStrategy ForceRevisionCreationStrategy { get; }
 
         public DynamicJsonValue ToJson(DocumentConventions conventions, JsonOperationContext context)
         {
-            return new DynamicJsonValue
+            var json = new DynamicJsonValue
             {
                 [nameof(Id)] = Id,
                 [nameof(ChangeVector)] = ChangeVector,
                 [nameof(Document)] = Document,
-                [nameof(Type)] = Type.ToString(),
-                [nameof(ForceRevisionCreation)] = ForceRevisionCreation
+                [nameof(Type)] = Type.ToString()
             };
+            
+            if (ForceRevisionCreationStrategy != ForceRevisionStrategy.None)
+            {
+                json[nameof(ForceRevisionCreationStrategy)] = ForceRevisionCreationStrategy;
+            }
+
+            return json;
         }
 
         public abstract void OnBeforeSaveChanges(InMemoryDocumentSessionOperations session);
--- a/src/Raven.Client/Documents/Commands/Batches/PutCommandDataBase.cs
+++ b/src/Raven.Client/Documents/Commands/Batches/PutCommandDataBase.cs
@@ -8,8 +8,8 @@ namespace Raven.Client.Documents.Commands.Batches
 {
     internal class PutCommandDataWithBlittableJson : PutCommandDataBase<BlittableJsonReaderObject>
     {
-        public PutCommandDataWithBlittableJson(string id, string changeVector, BlittableJsonReaderObject document, ForceRevisionStrategy forceRevisionCreationStrategy = ForceRevisionStrategy.None)
-            : base(id, changeVector, document, forceRevisionCreationStrategy)
+        public PutCommandDataWithBlittableJson(string id, string changeVector, BlittableJsonReaderObject document, ForceRevisionStrategy strategy = ForceRevisionStrategy.None)
+            : base(id, changeVector, document, strategy)
         {
         }
 
@@ -20,8 +20,8 @@ namespace Raven.Client.Documents.Commands.Batches
 
     public class PutCommandData : PutCommandDataBase<DynamicJsonValue>
     {
-        public PutCommandData(string id, string changeVector, DynamicJsonValue document, ForceRevisionStrategy forceRevisionCreationStrategy = ForceRevisionStrategy.None)
-            : base(id, changeVector, document, forceRevisionCreationStrategy)
+        public PutCommandData(string id, string changeVector, DynamicJsonValue document, ForceRevisionStrategy strategy = ForceRevisionStrategy.None)
+            : base(id, changeVector, document, strategy)
         {
         }
 
@@ -32,7 +32,7 @@ namespace Raven.Client.Documents.Commands.Batches
 
     public abstract class PutCommandDataBase<T> : ICommandData
     {
-        protected PutCommandDataBase(string id, string changeVector, T document, ForceRevisionStrategy forceRevisionCreationStrategy = ForceRevisionStrategy.None)
+        protected PutCommandDataBase(string id, string changeVector, T document, ForceRevisionStrategy strategy = ForceRevisionStrategy.None)
         {
             if (document == null)
                 throw new ArgumentNullException(nameof(document));
@@ -40,7 +40,7 @@ namespace Raven.Client.Documents.Commands.Batches
             Id = id;
             ChangeVector = changeVector;
             Document = document;
-            ForceRevisionCreationStrategy = forceRevisionCreationStrategy;
+            ForceRevisionCreationStrategy = strategy;
         }
 
         public string Id { get; }
--- a/src/Raven.Client/Documents/Commands/Batches/PutCommandDataBase.cs
+++ b/src/Raven.Client/Documents/Commands/Batches/PutCommandDataBase.cs
@@ -8,8 +8,8 @@ namespace Raven.Client.Documents.Commands.Batches
 {
     internal class PutCommandDataWithBlittableJson : PutCommandDataBase<BlittableJsonReaderObject>
     {
-        public PutCommandDataWithBlittableJson(string id, string changeVector, BlittableJsonReaderObject document)
-            : base(id, changeVector, document)
+        public PutCommandDataWithBlittableJson(string id, string changeVector, BlittableJsonReaderObject document, bool forceRevisionCreation = false)
+            : base(id, changeVector, document, forceRevisionCreation)
         {
         }
 
@@ -20,8 +20,8 @@ namespace Raven.Client.Documents.Commands.Batches
 
     public class PutCommandData : PutCommandDataBase<DynamicJsonValue>
     {
-        public PutCommandData(string id, string changeVector, DynamicJsonValue document)
-            : base(id, changeVector, document)
+        public PutCommandData(string id, string changeVector, DynamicJsonValue document, bool forceRevisionCreation = false)
+            : base(id, changeVector, document, forceRevisionCreation)
         {
         }
 
@@ -32,7 +32,7 @@ namespace Raven.Client.Documents.Commands.Batches
 
     public abstract class PutCommandDataBase<T> : ICommandData
     {
-        protected PutCommandDataBase(string id, string changeVector, T document)
+        protected PutCommandDataBase(string id, string changeVector, T document, bool forceRevisionCreation = false)
         {
             if (document == null)
                 throw new ArgumentNullException(nameof(document));
@@ -40,6 +40,7 @@ namespace Raven.Client.Documents.Commands.Batches
             Id = id;
             ChangeVector = changeVector;
             Document = document;
+            ForceRevisionCreation = forceRevisionCreation;
         }
 
         public string Id { get; }
@@ -47,6 +48,7 @@ namespace Raven.Client.Documents.Commands.Batches
         public string ChangeVector { get; }
         public T Document { get; }
         public CommandType Type { get; } = CommandType.PUT;
+        public bool ForceRevisionCreation { get; }
 
         public DynamicJsonValue ToJson(DocumentConventions conventions, JsonOperationContext context)
         {
@@ -55,7 +57,8 @@ namespace Raven.Client.Documents.Commands.Batches
                 [nameof(Id)] = Id,
                 [nameof(ChangeVector)] = ChangeVector,
                 [nameof(Document)] = Document,
-                [nameof(Type)] = Type.ToString()
+                [nameof(Type)] = Type.ToString(),
+                [nameof(ForceRevisionCreation)] = ForceRevisionCreation
             };
         }
 
--- a/src/Raven.Client/Documents/Commands/Batches/PutCommandDataBase.cs
+++ b/src/Raven.Client/Documents/Commands/Batches/PutCommandDataBase.cs
@@ -8,7 +8,12 @@ namespace Raven.Client.Documents.Commands.Batches
 {
     internal class PutCommandDataWithBlittableJson : PutCommandDataBase<BlittableJsonReaderObject>
     {
-        public PutCommandDataWithBlittableJson(string id, string changeVector, BlittableJsonReaderObject document, ForceRevisionStrategy strategy = ForceRevisionStrategy.None)
+        public PutCommandDataWithBlittableJson(string id, string changeVector, BlittableJsonReaderObject document)
+            : base(id, changeVector, document)
+        {
+        }
+        
+        public PutCommandDataWithBlittableJson(string id, string changeVector, BlittableJsonReaderObject document, ForceRevisionStrategy strategy)
             : base(id, changeVector, document, strategy)
         {
         }
@@ -20,7 +25,12 @@ namespace Raven.Client.Documents.Commands.Batches
 
     public class PutCommandData : PutCommandDataBase<DynamicJsonValue>
     {
-        public PutCommandData(string id, string changeVector, DynamicJsonValue document, ForceRevisionStrategy strategy = ForceRevisionStrategy.None)
+        public PutCommandData(string id, string changeVector, DynamicJsonValue document)
+            : base(id, changeVector, document)
+        {
+        }
+        
+        public PutCommandData(string id, string changeVector, DynamicJsonValue document, ForceRevisionStrategy strategy)
             : base(id, changeVector, document, strategy)
         {
         }
--- a/src/Raven.Client/Documents/Commands/Batches/RevisionsCommandData.cs
+++ b/src/Raven.Client/Documents/Commands/Batches/RevisionsCommandData.cs
@@ -8,23 +8,21 @@ namespace Raven.Client.Documents.Commands.Batches
 {
     public class RevisionsCommandData : ICommandData
     {
-        public RevisionsCommandData(string id, string changeVector)
+        public RevisionsCommandData(string id)
         {
             Id = id ?? throw new ArgumentNullException(nameof(id));
-            ChangeVector = changeVector;
         }
 
+        public CommandType Type { get; } = CommandType.ForceRevisionCreation;
         public string Id { get; }
         public string Name { get; } = null;
         public string ChangeVector { get; }
-        public CommandType Type { get; } = CommandType.ForceRevisionCreation;
-
+    
         public virtual DynamicJsonValue ToJson(DocumentConventions conventions, JsonOperationContext context)
         {
             return new DynamicJsonValue
             {
                 [nameof(Id)] = Id,
-                [nameof(ChangeVector)] = ChangeVector,
                 [nameof(Type)] = Type.ToString()
             };
         }
--- a/src/Raven.Client/Documents/Commands/Batches/RevisionsCommandData.cs
+++ b/src/Raven.Client/Documents/Commands/Batches/RevisionsCommandData.cs
@@ -1,0 +1,36 @@
+﻿using System;
+using Raven.Client.Documents.Conventions;
+using Raven.Client.Documents.Session;
+using Sparrow.Json;
+using Sparrow.Json.Parsing;
+
+namespace Raven.Client.Documents.Commands.Batches
+{
+    public class RevisionsCommandData : ICommandData
+    {
+        public RevisionsCommandData(string id, string changeVector)
+        {
+            Id = id ?? throw new ArgumentNullException(nameof(id));
+            ChangeVector = changeVector;
+        }
+
+        public string Id { get; }
+        public string Name { get; } = null;
+        public string ChangeVector { get; }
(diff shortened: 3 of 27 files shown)


src/Raven.Studio/typescript/commands/database/documents/saveDocumentCommand.ts
```diff
@@ -5,26 +5,37 @@ import endpoints = require("endpoints");
 
 class saveDocumentCommand extends commandBase {
 
-    constructor(private id: string, private document: document, private db: database, private reportSaveProgress = true) {
+    constructor(private id: string, private document: document, private db: database, private reportSaveProgress: boolean, private forceRevisionCreation: boolean) {
```
ml054: let's make default value for this parameter = true, as forceRevisionCreation should be rare case when calling this function

src/Raven.Studio/typescript/commands/database/documents/saveDocumentCommand.ts
```diff
@@ -5,26 +5,37 @@ import endpoints = require("endpoints");
...
+        const urlArgs = {
+           forceRevisionCreation: this.forceRevisionCreation  
+        };
+              
+        const url = endpoints.databases.batch.bulk_docs + this.urlEncodeArgs(urlArgs);
+        
         const saveTask = this.post<saveDocumentResponseDto>(url, args, this.db);
 
-        if (this.reportSaveProgress) {
+        if (this.reportSaveProgress && !this.forceRevisionCreation) {
```
ml054: can we refactor to:
    
if (this.reportSaveProgress) {
    if (this.forceRevisionCreation) {
       ....
     } else {
        ....
     } 
}

src/Raven.Studio/typescript/common/shell/continueTest.ts
```diff
@@ -28,7 +28,7 @@ class continueTest {
     continue() {
         const doc = document.empty();
         const db = databasesManager.default.getDatabaseByName(this.databaseName());
-        new saveDocumentCommand("Debug/Done", doc, db, false)
+        new saveDocumentCommand("Debug/Done", doc, db, false, false)
```
ml054: no need to add this extra param - use default

src/Raven.Studio/wwwroot/App/views/database/documents/editDocument.html
```diff
@@ -42,7 +42,8 @@
             </div>
             <div class="btn-bar toolbar">
                 <div class="pull-left-sm">
-                    <button type="submit" class="btn btn-primary save-btn" data-bind="enable: isSaveEnabled, css: { 'btn-spinner': isSaving }, visible: !inReadOnlyMode()">
+                    <button type="submit" class="btn btn-primary save-btn" data-bind="enable: isSaveEnabled, css: { 'btn-spinner': isSaving }, visible: !inReadOnlyMode()"
+                            title="Save changes. A revision will be created only if 'Settings | Document Revisions' is configured.">
```
ml054: I'm not sure I understand this title

src/Raven.Studio/wwwroot/App/views/database/documents/editDocument.html
```diff
@@ -225,11 +223,11 @@ <h3 class="margin-none">Properties</h3>
                             <div class="badge" data-bind="text: $root.crudActionsProvider().countersCount() || ''"></div>
                         </button>
                     </div>
-                    <div class="btn-group" id="revisions_pane" role="group" data-bind="visible: revisionsEnabled">
-                        <button type="button" class="btn btn-xs btn-default doc-revisions" data-bind="css: { active: isRevisionsActive() }, click: _.partial(activateRevisions, false)">
+                    <div class="btn-group" id="revisions_pane" role="group">
+                        <button type="button" class="btn btn-xs btn-default doc-revisions" data-bind="css: { active: isRevisionsActive() }, click: _.partial(activateRevisions, false), disable: !$root.canViewRevisions()">
```
ml054, line 227: do we want to hide or disable revisions tab? What's the logic for other tabs?

src/Raven.Studio/wwwroot/App/views/database/documents/editDocument.html
```diff
@@ -258,6 +256,16 @@ <h3 class="margin-none">Properties</h3>
                         </button>
                     </div>
                 </div>
+                <div class="flex-window-head" data-bind="visible: isRevisionsActive() && $root.canViewRevisions() && !$root.inReadOnlyMode(),
+                     attr: { title: $root.dirtyFlag().isDirty() ? 'Save your changes before creating a revision' : 'A revision will be created even if Revisions are Not configured for the collection' }">
+                    <div class="padding padding-xs margin-top margin-top-sm">
+                        <button type="button" class="btn btn-sm btn-primary btn-block" 
+                                data-bind="click: _.partial($root.forceCreateRevision, null), disable: $root.dirtyFlag().isDirty(), css: { 'btn-spinner': $root.isSaving }">
```
ml054: why do need partial here? forceCreateRevision doesn't take any arguments.

ml054:
in general: really nice work!

ayende (commented):
This is only through the UI, right?
We don't have a way to do this through code?

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -41,8 +41,10 @@ public class BatchHandler : DatabaseRequestHandler
         [RavenAction("/databases/*/bulk_docs", "POST", AuthorizationStatus.ValidUser)]
         public async Task BulkDocs()
         {
+            bool forceRevisionCreation = GetBoolValueQueryString("forceRevisionCreation", required: false) ?? false;
```
ppekrol: This works well from Studio, but it will be problematic from Client API, because it is not per item, but per whole batch.
It needs to be either a new command or put command with a flag.

ppekrol (changes requested):

src/Raven.Studio/typescript/commands/database/documents/saveDocumentCommand.ts
Danielle9897: done

src/Raven.Studio/typescript/commands/database/documents/saveDocumentCommand.ts
Danielle9897: done

src/Raven.Studio/typescript/common/shell/continueTest.ts
Danielle9897: done

src/Raven.Studio/wwwroot/App/views/database/documents/editDocument.html
Danielle9897: Well, I added this because:

Now we have the 'Create Revision' Button, which is very visible/noticeable, in the Properties pane. 
Un-aware/new users might think this button is the only way to create a revision.
I sort of wanted this title to achieve both the following:
  1. Make them aware that a config option also exists
  2. Clarify & point out the difference between the 2 revision creation options (as the new button also has a tooltip)

Either way - this title can be always removed if it gets too annoying......

src/Raven.Studio/wwwroot/App/views/database/documents/editDocument.html
Danielle9897: done - removed partial

src/Raven.Studio/wwwroot/App/views/database/documents/editDocument.html
Danielle9897: 1. Revisions tab is always _visible_, never _hidden_.
2. Revisions tab is _disabled_ upon 'new document' & 'clone'

src/Raven.Studio/wwwroot/App/views/database/documents/editDocument.html
Danielle9897: Removed - as discussed

src/Raven.Server/Documents/DocumentPutAction.cs
```diff
@@ -262,6 +263,9 @@ private static bool ShouldVersionOldDocument(DocumentFlags flags, BlittableJsonR
             if (flags.Contain(DocumentFlags.Resolved))
                 return false; // we already versioned it with the a conflicted flag
 
+            if (nonPersistentDocumentFlags.Contain(NonPersistentDocumentFlags.ForceRevisionCreation))
+                return false; // no need to create 2 identical revisions
```
karmeli87: If I understand this correctly we want to version only the old document, so this should return `true`, no?

src/Raven.Server/Documents/Handlers/BatchHandler.cs
Danielle9897: Changed this flag to be part of the commandData item

src/Raven.Client/Documents/Session/IRevisionsSessionOperations.cs
```diff
@@ -10,6 +10,20 @@
...
+    {
+        // Don't force revision
+        None,
+        
+        // Create a forced revision from the document that is currently in store - BEFORE applying any changes made by the user 
+        // The only exception is: with current implementation the forced revision for *new* document only is 'after', not 'before'  
+        Before,
+        
+        // After is currently Not implemented on the server side ! TBD only if requested by customer
+        After
```
grisha-kotler: It can be removed since this isn't implemented yet.

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -900,6 +901,17 @@ private static bool UpdateMetadataModifications(DocumentInfo documentInfo)
 
             return true;
         }
+        
+        private void PrepareForCreatingRevisionsFromIDs(SaveChangesData result)
+        {
+            // Note: here there is not point checking 'Before' or 'After' because if there were changes then forced revision is done from the PUT command....
```
grisha-kotler:
```suggestion
            not -> no
```

grisha-kotler (changes requested):

src/Raven.Client/Documents/Session/IRevisionsSessionOperations.cs
Danielle9897: done

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
Danielle9897: done

src/Raven.Client/Documents/Session/DocumentSessionRevisions.cs
```diff
@@ -68,5 +68,33 @@ public T Get<T>(string id, DateTime date)
...
+            if (ReferenceEquals(entity, null))
+                throw new ArgumentNullException(nameof(entity));
+           
+            if (DocumentsByEntity.TryGetValue(entity, out DocumentInfo documentInfo) == false)
+            {
+                throw new InvalidOperationException("Cannot create a revision for the requested entity because it is not tracked by the session");
+            }
+
+            if (Session.IDsForCreatingForcedRevisions.ContainsKey(documentInfo.Id))
+            {
```
ayende: If they are doing the same operation, why throw? Can we just check that the same `ForceRevisionStrategy` is applied and treat is a noop?

src/Raven.Client/Documents/Session/DocumentSessionRevisions.cs
```diff
@@ -68,5 +68,33 @@ public T Get<T>(string id, DateTime date)
...
+                throw new InvalidOperationException("A request for creating a revision was already made for the document in the current session.");
+            }
+            
+            Session.IDsForCreatingForcedRevisions.Add(documentInfo.Id, forceRevisionCreationStrategy);
+        } 
+        
+        public void ForceRevisionCreationFor(string id, ForceRevisionStrategy forceRevisionCreationStrategy = ForceRevisionStrategy.Before)
+        {
+            if (Session.IDsForCreatingForcedRevisions.ContainsKey(id))
+            {
```
ayende, line 93: Same

src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
```diff
@@ -365,6 +369,44 @@ private void HandleDeleteInternal(BlittableJsonReaderObject batchResult, Command
...
+            // because ForceRevisionCreation can also create (put) new documents in the server:
+            // when forcing a revision for a document that did not have any revisions yet (original document is updated/saved with has-revisions flag)
+            
+            if (GetBooleanField(batchResult, CommandType.ForceRevisionCreation, "RevisionCreated") == false)
+            { 
+                // no forced revision was created...nothing to update.
+                return;
+            }
+            
+            var id = GetLazyStringField(batchResult, CommandType.ForceRevisionCreation, Constants.Documents.Metadata.Id);
```
ayende, line 383: Isn't this code duplicate from `HandleEntityPut`, can we avoid this duplication?

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -752,7 +750,30 @@ protected override int ExecuteCmd(DocumentsOperationContext context)
...
+                                var flags = DocumentFlags.None;
+                                
+                                //if (cmd.ForceRevision)
+                                if (cmd.ForceRevisionCreationStrategy == ForceRevisionStrategy.Before)
+                                // Note: we currently only handle 'Before'. 
+                                // Creating the revision 'After' will be done only upon customer demand.
+                                {
+                                    var existingDocument = Database.DocumentsStorage.Get(context, cmd.Id);
+                                    if (existingDocument == null)
+                                    {
```
ayende, line 763: Am I missing something, the comment on `HandleForceRevisionCreation` says that we can force revision on document creation, but here it will throw?

src/Raven.Server/Documents/Handlers/BatchRequestParser.cs
```diff
@@ -805,33 +808,73 @@ private static unsafe CommandPropertyName GetPropertyType(JsonParserState state)
...
             }
         }
 
+        private static unsafe ForceRevisionStrategy GetEnumValue(JsonParserState state, JsonOperationContext ctx)
+        {
+            switch (state.StringSize)
+            {
+                case 6:
+                    if (*(int*)state.StringBuffer == 1868981570 &&
+                        state.StringBuffer[5] == (byte)'e')
```
ayende: Missing check for `StringBuffer[4]`, need to compare it to `short` value, I think.

ayende (changes requested):

src/Raven.Client/Documents/Session/DocumentSessionRevisions.cs
ppekrol: I think we need to throw, because we will explicitly inform user how this works. If we will not throw then his intention might be to create N-number of revisions.

src/Raven.Client/Documents/Session/IRevisionsSessionOperations.cs
```diff
@@ -10,6 +10,19 @@
...
+    public enum ForceRevisionStrategy
+    {
+        // Don't force revision
+        None,
+        
+        // Create a forced revision from the document that is currently in store - BEFORE applying any changes made by the user 
+        // The only exception is: with current implementation the forced revision for *new* document only is 'after', not 'before'  
+        Before,
+        
+        // After is currently Not implemented on the server side ! TBD only if requested by customer
```
ppekrol: Can be removed

src/Raven.Client/Documents/Session/IRevisionsSessionOperations.cs
```diff
@@ -40,5 +53,23 @@ public interface IRevisionsSessionOperations
...
         T Get<T>(string id, DateTime date);
+        
+        /// <summary>
+        /// Make the session create a revision for the specified entity.
+        /// Can be used with tracked entities only.
+        /// Revision will be created Even If:
+        ///    1. Revisions configuration is Not set for the collection
+        ///    2. Document was Not modified
+        /// </summary>
+        void ForceRevisionCreationFor<T>(T entity, ForceRevisionStrategy forceRevisionCreationStrategy = ForceRevisionStrategy.Before);
```
ppekrol: Rename paramter name to 'strategy'

src/Raven.Client/Documents/Session/IRevisionsSessionOperations.cs
```diff
@@ -40,5 +53,23 @@ public interface IRevisionsSessionOperations
...
+        void ForceRevisionCreationFor<T>(T entity, ForceRevisionStrategy forceRevisionCreationStrategy = ForceRevisionStrategy.Before); 
+        
+        /// <summary>
+        /// Make the session create a revision for the specified document id.
+        /// Revision will be created Even If:
+        ///    1. Revisions configuration is Not set for the collection
+        ///    2. Document was Not modified
+        /// </summary>
+        /// <param name="id"></param>
+        void ForceRevisionCreationFor(string id, ForceRevisionStrategy forceRevisionCreationStrategy = ForceRevisionStrategy.Before);
```
ppekrol: Rename paramter name to 'strategy'

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -197,6 +197,8 @@ public async Task<ServerNode> GetCurrentSessionNode()
             new Dictionary<(string, CommandType, string), ICommandData>();
 
         public readonly bool NoTracking;
+       
+        public Dictionary<string, ForceRevisionStrategy> IDsForCreatingForcedRevisions = new Dictionary<string, ForceRevisionStrategy>(StringComparer.OrdinalIgnoreCase);
```
ppekrol: can be private or internal?

Also rename to 'IdsForCreatingForcedRevisions'

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -900,6 +901,17 @@ private static bool UpdateMetadataModifications(DocumentInfo documentInfo)
 
             return true;
         }
+        
+        private void PrepareForCreatingRevisionsFromIDs(SaveChangesData result)
```
ppekrol: ID -> Id

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -986,6 +1012,76 @@ protected override int ExecuteCmd(DocumentsOperationContext context)
...
+                            if (revisionCreated)
+                            {
+                                // Reply with new revision data
+                                forceRevisionReply = new DynamicJsonValue
+                                {
+                                    ["RevisionCreated"] = true,
+                                    ["Type"] = nameof(CommandType.ForceRevisionCreation), 
+                                    [Constants.Documents.Metadata.Id] = existingDoc.Id,
+                                    [Constants.Documents.Metadata.Collection] =  CollectionName.GetCollectionName(existingDoc.Data),
+                                    [Constants.Documents.Metadata.ChangeVector] = existingDoc.ChangeVector,
```
ppekrol, line 1066: I'm wondering if we need this actually. What we are going to do with this on client-side?

src/Raven.Client/Documents/Session/IRevisionsSessionOperations.cs
```diff
@@ -10,6 +10,19 @@
 
 namespace Raven.Client.Documents.Session
 {
+    
+    public enum ForceRevisionStrategy
```
ppekrol, line 14: Let's rename this maybe to 'RevisionCreationStrategy'

src/Raven.Client/Documents/Commands/Batches/PutCommandDataBase.cs
```diff
@@ -8,8 +8,8 @@ namespace Raven.Client.Documents.Commands.Batches
 {
     internal class PutCommandDataWithBlittableJson : PutCommandDataBase<BlittableJsonReaderObject>
     {
-        public PutCommandDataWithBlittableJson(string id, string changeVector, BlittableJsonReaderObject document)
-            : base(id, changeVector, document)
+        public PutCommandDataWithBlittableJson(string id, string changeVector, BlittableJsonReaderObject document, ForceRevisionStrategy forceRevisionCreationStrategy = ForceRevisionStrategy.None)
```
ppekrol: rename the argument to 'revisionCreationStrategy'

src/Raven.Client/Documents/Commands/Batches/RevisionsCommandData.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+using Raven.Client.Documents.Conventions;
+using Raven.Client.Documents.Session;
+using Sparrow.Json;
+using Sparrow.Json.Parsing;
+
+namespace Raven.Client.Documents.Commands.Batches
+{
+    public class RevisionsCommandData : ICommandData
```
ppekrol: Can we change the name? Not sure what would be suitable... `CreateRevisionCommandData` ?

src/Raven.Client/Documents/Commands/Batches/ICommandData.cs
```diff
@@ -32,16 +32,14 @@ public enum CommandType
         AttachmentCOPY,
         CompareExchangePUT,
         CompareExchangeDELETE,
-
+        ForceRevisionCreation,
```
ppekrol, line 35: Maybe just `CreateRevision` ? Thoughts?

src/Raven.Server/Documents/Handlers/BatchHandler.cs
ayende: I don't think we _can_ do anything with it. There is no record of revisions on the session, and no need to track those.
I guess we might want to save the change vector of the revision? But we aren't returning it

src/Raven.Client/Documents/Session/DocumentSessionRevisions.cs
ayende: I think that this is more likely that he might be trying to register the same thing twice.
Once he is doing that by id, once with the entity (can happen in complex cases).
That isn't an _error_, the end result is just the same.
And we don't even have a way for him to _check_ that.

src/Raven.Client/Documents/Session/IRevisionsSessionOperations.cs
```diff
@@ -10,6 +10,19 @@
...
+    {
+        // Don't force revision
+        None,
```
grisha-kotler, line 17: Why do we need None?

src/Raven.Client/Documents/Commands/Batches/RevisionsCommandData.cs
Danielle9897: imho - change to - **ForceRevisionCommandData**
because this command is created only when coming from the 'Force Revision' flow
What do you think?

src/Raven.Client/Documents/Session/IRevisionsSessionOperations.cs
Danielle9897: done

src/Raven.Client/Documents/Session/IRevisionsSessionOperations.cs
Danielle9897: done

src/Raven.Client/Documents/Session/IRevisionsSessionOperations.cs
Danielle9897: done

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
Danielle9897: done

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
Danielle9897: Done. Renamed and made internal.

src/Raven.Client/Documents/Commands/Batches/RevisionsCommandData.cs
ppekrol: fine

src/Raven.Client/Documents/Session/IRevisionsSessionOperations.cs
ayende: Forward compatability, we may have other values there, and `Before` is an action, so we need a neutral value

Danielle9897: Removed - used nullable param instead

Danielle9897: ok - leaving None

src/Raven.Server/Documents/Handlers/BatchRequestParser.cs
Danielle9897: Fixed. Used int & short.

src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
Danielle9897: ok - extracted common part

src/Raven.Server/Documents/Handlers/BatchHandler.cs
grisha-kotler: We're tracking the document in the session.
When the document doesn't have the "HasRevisions" flag we need to update it and return the new document change vector.

src/Raven.Client/Documents/Session/DocumentSessionRevisions.cs
Danielle9897: Done. Now throwing only if a different strategy is requested in the same session.

src/Raven.Client/Documents/Session/DocumentSessionRevisions.cs
Danielle9897: Done. Now throwing only if a different strategy is requested in the same session.

src/Raven.Server/Documents/Handlers/BatchHandler.cs
Danielle9897: 1. Comment here means: 
We cannot force/create a Revision if the document (for which a Forced Revision was requested) was Not yet saved to the server.

2. Comment in `HandleForceRevisionCreation` means:
The original document (for which you request a Forced Revision) may have changed & PUT to the server with this flow:

* Create document
* Save document to server (document is created without the HasRevisions Flag)
* Request a Forced Revision
   => A forced revision is created
   => Original document is modified and saved - since HasRevisions Flag is added to it 
         So for tracked entities - `HandleForceRevisionCreation` must update the changeVector in the session 

I will rephrase so its clearer

src/Raven.Client/Documents/Commands/Batches/ICommandData.cs
Danielle9897: imho - I prefer **ForceRevisionCreation** because this is not a regular revision - 
this command is created from the special flow of forcing a revision - 
as opposed to a revision created just because the Revisions Settings are set.

src/Raven.Client/Documents/Commands/Batches/PutCommandDataBase.cs
Danielle9897: or maybe rename the param to just 'strategy' ?

src/Raven.Client/Documents/Session/IRevisionsSessionOperations.cs
Danielle9897: But why ? 
ASFAIK This doesn't apply to 'regular' revisions, only to Forced Revisions

src/Raven.Client/Documents/Commands/Batches/PutCommandDataBase.cs
```diff
@@ -20,8 +20,8 @@ public override void OnBeforeSaveChanges(InMemoryDocumentSessionOperations sessi
 
     public class PutCommandData : PutCommandDataBase<DynamicJsonValue>
     {
-        public PutCommandData(string id, string changeVector, DynamicJsonValue document)
-            : base(id, changeVector, document)
+        public PutCommandData(string id, string changeVector, DynamicJsonValue document, ForceRevisionStrategy strategy = ForceRevisionStrategy.None)
```
ppekrol: We need to create a 2nd ctor to avoid issues with binary compatibility

ppekrol (approved):
LGTM, needs one small adjustment

src/Raven.Client/Documents/Commands/Batches/PutCommandDataBase.cs
Danielle9897: done
https://github.com/ravendb/ravendb/pull/9327/commits/a283f2ad28424f4f2bc547234d610be460e7ad19

PR #9332, opened 2019-05-30
RavenDB-13602 HardResetToNewCluster remove the irrelevant nodes from …

karmeli87:
…database groups

### DIFF
--- a/src/Raven.Server/ServerWide/ClusterStateMachine.cs
+++ b/src/Raven.Server/ServerWide/ClusterStateMachine.cs
@@ -18,6 +18,7 @@ using Raven.Client.Documents.Session;
 using Raven.Client.Exceptions.Database;
 using Raven.Client.Exceptions.Documents.Subscriptions;
 using Raven.Client.Exceptions.Security;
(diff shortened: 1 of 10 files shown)


grisha-kotler:
Can we have a test for this?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1419,59 +1419,74 @@ public override void Initialize(RachisConsensus parent, TransactionOperationCont
 
         private void OnStateChange(object sender, RachisConsensus.StateTransition transition)
         {
-            if (transition.From == RachisState.Passive && transition.To == RachisState.LeaderElect)
+            if (transition.To != RachisState.LeaderElect)
```
ayende: This happens as soon as we move to `Passive` state.
But previously, that was when we moved from `Passive` to something else.

How do you handle removal of node from cluster and then adding it again in this case?

ppekrol (approved):

ayende (approved):

PR #9334, opened 2019-05-30
RavenDB-13022

talweiss1982:
@ayende look at my comment at CryptoPager.cs
My changes do compute the memory correctly but if TransactionSizeLimit is not set we are not breaking the batch so I'm not sure if we want to modify this behavior and have some default value there or not.


src/Voron/Impl/LowLevelTransaction.cs
```diff
@@ -87,6 +87,26 @@ public void Reset()
 
 
         public event Action<IPagerLevelTransactionState> BeforeCommitFinalization;
+
+        public long TotalEncryptionBufferSize
+        {
+            get
+            {
+                var cryptoTransactionStates = ((IPagerLevelTransactionState)this).CryptoPagerTransactionState;
```
ayende, line 95: Why do you need this cast?

src/Voron/Impl/Paging/CryptoPager.cs
```diff
@@ -15,6 +15,23 @@ namespace Voron.Impl.Paging
     public class CryptoTransactionState
     {
         public Dictionary<long, EncryptionBuffer> LoadedBuffers = new Dictionary<long, EncryptionBuffer>();
+        /// <summary>
+        /// Used for computing the total memory used by the transaction crypto buffers
+        /// </summary>
+        //TODO: we can optimize this by hiding the dictionary and computing the total size when we enter/remove a buffer
```
ayende: That would probably be best, yes. This might be called a LOT

src/Voron/Impl/Paging/TempPagerTransaction.cs
```diff
@@ -30,6 +30,26 @@ public void Dispose()
...
+        {
+            get
+            {
+                var cryptoTransactionStates = CryptoPagerTransactionState;
+                if (cryptoTransactionStates == null)
+                {
+                    return 0L;
+                }
+
+                var total = 0L;
```
ayende, line 44: Can we remember this here as well?

src/Voron/Impl/Paging/CryptoPager.cs
arekpalinski: Maybe something similar to `IScratchSpaceMonitor`:

https://github.com/ravendb/ravendb/blob/v4.2/src/Voron/Impl/Scratch/IScratchSpaceMonitor.cs

src/Voron/Impl/LowLevelTransaction.cs
talweiss1982: because somebody decided to implicitly implement this method and this is the only way to access it

src/Voron/Impl/Paging/TempPagerTransaction.cs
talweiss1982: remember what?

ayende: Remember the value, I meant. So we don't have to re-compute it at all times

talweiss1982: You can't just remember it, it is been modified constantly, the majority of the computation is done per CryptoState and that is already been computed on the fly.
Computing the total for CryptoPagerTransactionState will mean hiding it too and I don't see the gain here it is already a tiny price to pay for a very expensive operation.

src/Voron/Impl/Paging/CryptoPager.cs
```diff
@@ -12,9 +14,47 @@
...
+        {
+            return _loadedBuffers.GetEnumerator();
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return GetEnumerator();
+        }
+
+        public void ReverseBuffers()
```
ayende: What is this? 
This has no meaning, dictionary has no defined order.

test/SlowTests/Voron/Issues/RavenDB_10825.cs
```diff
@@ -37,7 +37,7 @@ public unsafe void Encryption_buffer_of_freed_scratch_page_must_not_affect_anoth
                 var cryptoTxState = ((IPagerLevelTransactionState)tx.LowLevelTransaction).CryptoPagerTransactionState.Single().Value;
 
                 // explicitly change the order of items in dictionary so we'll apply the buffer of overflow page before the already freed one
-                cryptoTxState.LoadedBuffers = cryptoTxState.LoadedBuffers.Reverse().ToDictionary(x => x.Key, x => x.Value);
+                cryptoTxState.ReverseBuffers();
```
ayende: This is wrong. It _may_ work on a dictionary that is very small, but that is wrong. 
It certainly shouldn't be in the API

ayende (changes requested):

test/SlowTests/Voron/Issues/RavenDB_10825.cs
talweiss1982: This was the code that was written I have just hidden the dictionary

talweiss1982: I can delete the test if you want

src/Voron/Impl/Paging/CryptoPager.cs
talweiss1982: i'm the wrong guy to complain about this:
https://github.com/ravendb/ravendb/blame/v4.2/test/SlowTests/Voron/Issues/RavenDB_10825.cs#L40

test/SlowTests/Voron/Issues/RavenDB_10825.cs
arekpalinski: I don't think we want to delete a test. The test does this hack in order to confirm that something is working correctly but it doesn't need to use a dedicated method for that I think.

src/Voron/Impl/IPagerLevelTransactionState.cs
```diff
@@ -8,6 +8,8 @@ public interface IPagerLevelTransactionState : IDisposable
     {
         Dictionary<AbstractPager, TransactionState> PagerTransactionState32Bits { get; set; }
         Dictionary<AbstractPager, CryptoTransactionState> CryptoPagerTransactionState { get; set; }
+        long TotalEncryptionBufferSize { get; }
```
arekpalinski: Can we change its type to `Size`? The usage of `Size` make the usage of it a bit less error prone

test/SlowTests/Voron/Issues/RavenDB_10825.cs
talweiss1982: if I expose the dictionary than I didn't do anything by hiding it.
I could add an expensive setter that will re-calculate the size of the buffers and if it is used only by one test than it should be okay.

src/Voron/Impl/IPagerLevelTransactionState.cs
talweiss1982: I have changed it to long so we won't have to allocate size object every time we check the size. This is done this way everywhere else I have looked.

src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
```diff
@@ -33,7 +33,7 @@ public class SubscriptionDocumentsFetcher
         private readonly SubscriptionState _subscription;
         private readonly SubscriptionPatchDocument _patch;
         // make sure that we don't use too much memory for subscription batch
-        private readonly Size _maximumAllowedMemory;
+        private readonly long _maximumAllowedMemory;
```
arekpalinski: Why change it to `long`?

talweiss1982: so you won't have to do size -> long and long and then allocate new size.
I can't accumulate the size of the buffers I only got the total and I don't want to add the total to the size each time I check the size, those are too many allocations it is better to use long.

src/Voron/Impl/LowLevelTransaction.cs
```diff
@@ -87,6 +87,26 @@ public void Reset()
 
 
         public event Action<IPagerLevelTransactionState> BeforeCommitFinalization;
+
+        public long TotalEncryptionBufferSize
```
arekpalinski: Shouldn't `TotalEncryptionBufferSize` be part of `IPagerLevelTransactionState` (just next to `IPagerLevelTransactionState.CryptoPagerTransactionState`)?

talweiss1982: it is

src/Voron/Impl/IPagerLevelTransactionState.cs
arekpalinski: The reason we have `Size` is that it makes it more difficult to make a mistake like - oh, that's in bytes, I thought the value is in kilobytes etc. `Size` is a struct so it won't have expensive allocation impact.

test/SlowTests/Voron/Issues/RavenDB_10825.cs
arekpalinski: Then maybe let's rename the method `ReverseBuffers` to `ReverseBuffers_DebugOnly` and mark it as `internal` so it will be accessible in tests.

src/Voron/Impl/LowLevelTransaction.cs
arekpalinski: Right. I missed that.

src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
arekpalinski: As commented above - `Size` is a struct. I'm for using `Size` everywhere across the board (where possible) since it makes the code less error prone.

test/SlowTests/Voron/Issues/RavenDB_10825.cs
ayende: This behavior is fine with the _test_, but it isn't fine in the API.

src/Voron/Impl/IPagerLevelTransactionState.cs
maximburyak: @arekpalinski on the other hand, we could add 'InBytes' to the end of the property name, achieving the same readability, without impacting potential JIT optimizations, due to lack of ability to optimize structs to registers usage.

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -3395,6 +3395,8 @@ private int MinBatchSize
             var txAllocations = indexingContext.Transaction.InnerTransaction.LowLevelTransaction.NumberOfModifiedPages
                                 * Voron.Global.Constants.Storage.PageSize;
 
+            txAllocations += indexingContext.Transaction.InnerTransaction.LowLevelTransaction.TotalEncryptionBufferSize;
```
maximburyak: @ayende  Regarding txAllocations limit in index: By default the value for Configuration.TransactionSizeLimit is null, therefore, we will not engage the logic that uses this value. In that case, should we have a default value for Index.TransactionSizeLimit in the case that there is no such value in the config? If we do, should it be 64MB, as we have in Bulk Insert?

src/Voron/Impl/IPagerLevelTransactionState.cs
arekpalinski: We use `Size` pretty commonly in many places so I'm not so worried about its impact on JIT but we might ask @redknightlois about opinion.

Adding 'InBytes'  prefix is fine but the usage of `Size` is even better because it has dedicated methods, operators etc and given that we use `Size` in other places too then it's easier to deal with them IMO e.g. like here just using `>` operator:

https://github.com/ravendb/ravendb/blob/843cf60f5341a57838e439265a77f1319494fb81/src/Raven.Server/Documents/Indexes/Index.cs#L3336

What do you think?

maximburyak: Although it seems t's more of a style issue because both approaches are used throught the code, , while Size demands calling contructors and function calls even for increasing values (while in the bottom line it relies on an inner propery "_valueInBytes") using just a `long` type that represents value in bytes, when it's the only form of data that we are interested in is the simplest, most readable way to go. 
I think that Size starts to be significant when you are required to switch measurement units, therefore I would prefere a cleaner approach here. 
Generally speaking, I think that we should be careful not to make `Size` the default tool, because it might introduce coding standards that may impact overall performance.

ppekrol: Size is a struct, should not impact performance that much. It was introduced to avoid miscalculations in the code in places like that (which we had few times in the past).

ayende: 1. Let's use struct here.
2. @maximburyak The JIT is smart enough to inline such a process and eliminate the costs.

src/Raven.Server/Documents/Indexes/Index.cs
ayende: 64MB is _very_ small. In the bulk insert case, we use that to limit the amount of data we read from the network.
The problem here is that we need to balance the speed of indexing with the cost of the transaction.

src/Voron/Impl/IPagerLevelTransactionState.cs
redknightlois: Use `Size` whenever possible. The cost of `Size` in allocations is 0 and in parameter passing, it is 2 registers, which is negligible. The positive impact it brings in reliability is so big that it is a no-brainer.

maximburyak: @redknightlois I think that only in .net 3.0 structs will be optimized into registers. Anyway, you could look into a little comparison I made, it's actually interesting: https://gist.github.com/maximburyak/23dde736e6d5d4178cbcbbc5b6c7f126 bottom line is that run time of Size.Add is more than an order of magnitude slower than long +=, and new Size is two time slower than the first one

redknightlois: @maximburyak There are many things wrong on that microbenchmark setup, we cannot trust those results. For proper setup, use BenchmarkDotNet which will handle most of the nuisances (albeit not all of them) so the probability of getting a proper microbenchmark from the get go are far higher.

Aside from that, there is no reason why you would want to do something like this in a hot loop:
```csharp
 for (Size k = new Size(0, SizeUnit.Bytes); k < maxSize; k.Add(1, SizeUnit.Bytes));
``` 
You are obviously paying the cost of the constructor, and the `.Add()` which may not even get inlined depending on the context. Inside a hot loop, using the long is OK... but `Size` should be used for exchanging and cold checks like the one at the `Index`.

https://twitter.com/federicolois/status/1117762997072617472
This is one of the cases where uglification is not warranted. 

Struct promotion for 2 longs structs already exist since 1.1, it was limited back then but kinda good, On 3.0 it is just amazing.

maximburyak: In my(subjective) eyes, using `+=` is at least as readable and elegant and safe as `.Add` or `new Size(` in that particular case, where we are tracking byte count. 
I think that this discussion is more fundamental than the specific case it is initiated for.

Anyway, for completeness, here is a benchmark.net output:

```
// * Summary *

BenchmarkDotNet=v0.11.5, OS=Windows 10.0.17134.765 (1803/April2018Update/Redstone4)
Intel Core i7-8700K CPU 3.70GHz (Coffee Lake), 1 CPU, 12 logical and 6 physical cores
Frequency=3609377 Hz, Resolution=277.0561 ns, Timer=TSC
  [Host]     : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.7.3416.0
  DefaultJob : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.7.3416.0


|          Method |      Mean |     Error |    StdDev |
|---------------- |----------:|----------:|----------:|
|   LongIncrement | 0.0110 ns | 0.0016 ns | 0.0014 ns |
|   SizeIncrement | 1.5476 ns | 0.0085 ns | 0.0071 ns |
| SizeConstructor | 3.0374 ns | 0.0058 ns | 0.0054 ns |

// * Warnings *
ZeroMeasurement
  MyBenchmark.LongIncrement: Default -> The method duration is indistinguishable from the empty method duration

// * Hints *
Outliers
  MyBenchmark.LongIncrement: Default -> 1 outlier  was  removed (1.19 ns)
  MyBenchmark.SizeIncrement: Default -> 2 outliers were removed (2.75 ns, 2.77 ns)

// * Legends *
  Mean   : Arithmetic mean of all measurements
  Error  : Half of 99.9% confidence interval
  StdDev : Standard deviation of all measurements
  1 ns   : 1 Nanosecond (0.000000001 sec)

// ***** BenchmarkRunner: End *****
// ** Remained 0 benchmark(s) to run **
Run time: 00:01:13 (73.8 sec), executed benchmarks: 3

Global total time: 00:01:15 (75.17 sec), executed benchmarks: 3
// * Artifacts cleanup *
```

See code here: https://gist.github.com/maximburyak/3f83611c0e1a5dabd2da9943d2b53549

redknightlois: Not a big difference, and can be fixed if you aggressively inline the add and the constructor which by default is not applied. Either way, while agree about the `Add()` vs `+=`, the addition of the scale far outweigh the performance impact (in this case) and should be the default when such questions arise. Reliability and maintainability first, performance right after.

ayende:
Just to clear things, we will use `Size` here. Please make the changes to the PR

src/Raven.Server/Documents/Indexes/Index.cs
ayende: I didn't notice that this is about encrypted databases.
For those scenarios, we don't want to keep too much in plain text in memory. We also have to worry about how much memory we can lock. So let's got with 64MB by default for encrypted dbs here.

src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
ayende: I think you missed changing this one

src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
```diff
@@ -79,7 +80,7 @@ public class SubscriptionDocumentsFetcher
             long startEtag)
         {
             int numberOfDocs = 0;
-            Size size = new Size(0, SizeUnit.Megabytes);
+            long size = 0;
```
ayende, line 83: Here as well

src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
```diff
@@ -151,7 +152,7 @@ public class SubscriptionDocumentsFetcher
...
```
ayende: Here

ayende (commented):
A couple of minor long -> size still remains.
Looks good otherwise

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
```diff
@@ -150,6 +150,13 @@ public virtual bool RunInMemory
         [ConfigurationEntry("Indexing.TransactionSizeLimitInMb", ConfigurationEntryScope.ServerWideOrPerDatabase)]
         public Size? TransactionSizeLimit { get; protected set; }
 
+        [Description("Transaction size limit, for encrypted database only, after which an index will stop and complete the current batch")]
+        [DefaultValue(64)]
+        [SizeUnit(SizeUnit.Megabytes)]
+        [IndexUpdateType(IndexUpdateType.Refresh)]
+        [ConfigurationEntry("Indexing.TransactionSizeLimitInMb.Encrypted", ConfigurationEntryScope.ServerWideOrPerDatabase)]
```
ppekrol: Let's go with `Indexing.Encrypted.TransactionSizeLimitInMb`

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -3222,7 +3222,15 @@ private int MinBatchSize
 
                 if (Configuration.TransactionSizeLimit == null)
                 {
-                    _transactionSizeLimit = new Lazy<Size?>(() => null);
+                    _transactionSizeLimit = new Lazy<Size?>(() =>
+                    {
+                        if (DocumentDatabase.IsEncrypted)
```
ppekrol: I think we need to enhance the logic here. Notice that we are doing some adjustments for map-reduce when the value is not null.

maybe sth like to simply the code here:
```
var limit = DocumentDatabase.IsEncrypted 
    ? Configuration.EncryptedTransactionSizeLimit ?? Configuration.TransactionSizeLimit 
    : Configuration.TransactionSizeLimit;

if (limit != null && Type.IsMapReduce())
    limit = limit * 0.75;

_transactionSizeLimit = new Lazy<Size?>(() => limit);
```

ayende (approved):

ppekrol (approved):

PR #9342, opened 2019-06-03
RavenDB-13311 Cluster observer should be more informative for the rea…

karmeli87:
…son of why not promoting a node to become a member

### DIFF
--- a/src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
+++ b/src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
@@ -144,7 +144,8 @@ namespace Raven.Server.ServerWide.Maintenance
 
             if (_lastLogs.TryGetValue(message, out var last))
             {
-                if (last + 60 > _iteration) // 30 sec
+                if (last + 60 > _iteration) 
+                    // each iteration occur every 500 ms, so we update the log with the _same_ message every 30 sec (60 * 0.5s)
                     return;
             }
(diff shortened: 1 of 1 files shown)


ppekrol (approved):

src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
```diff
@@ -137,28 +137,28 @@ public void Run(CancellationToken token)
...
-            if (_logger.IsInfoEnabled || _logger.IsOperationsEnabled)
+            if (_lastLogs.TryGetValue(message, out var last))
             {
-                if (_lastLogs.TryGetValue(message, out var last))
-                {
-                    if (last + 10 > _iteration)
-                        return;
-                }
-                _lastLogs[message] = _iteration;
+                if (last + 60 > _iteration) // 30 sec
```
ayende: I don't follow this computation?

karmeli87: each iteration happened every 500 ms.
500ms * 60 = 30 s

ayende: Can you add this as a comment?

karmeli87: done

ayende (approved):

PR #9347, opened 2019-06-04
RavenDB-13659 Possible deadlock when changing rachis state and execut…

karmeli87:
…ing `SwitchToCandidateStateOnTimeout`

### DIFF
--- a/src/Raven.Server/Rachis/TimeoutEvent.cs
+++ b/src/Raven.Server/Rachis/TimeoutEvent.cs
@@ -117,27 +117,41 @@ namespace Raven.Server.Rachis
             }
         }
 
+        private int _inProgress;
+
         public void ExecuteTimeoutBehavior()
         {
-            lock (this)
+            var action = _timeoutHappened;
+
+            if (Interlocked.CompareExchange(ref _inProgress, 1, 0) == 1)
             {
-                if (Disable)
-                    return;
+                Defer(_currentLeader);
+                return;
+            }
 
-                if (_singleShot)
-                    _timer.Change(Timeout.Infinite, Timeout.Infinite);
-
-                try
+            try
+            {
+                lock (this)
                 {
-                    if (_timeoutHappened == null)
+                    if (Disable)
                         return;
 
-                    _timeoutHappened?.Invoke();
+                    if (_singleShot)
+                        _timer.Change(Timeout.Infinite, Timeout.Infinite);
                 }
-                catch (ConcurrencyException)
-                {
-                    // expected, ignoring
-                }
+
+                if (action == null)
+                    return;
+
+                action.Invoke();
+            }
+            catch (ConcurrencyException)
+            {
+                // expected, ignoring
+            }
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Rachis/TimeoutEvent.cs
```diff
@@ -117,27 +117,32 @@ public void DisableTimeout()
             }
         }
 
+        private int _inProgress;
+
         public void ExecuteTimeoutBehavior()
         {
-            lock (this)
-            {
-                if (Disable)
```
ppekrol, line 124: Shouldn't we take into account that it might be disabled?

karmeli87: yes, was too eager on removing the `lock` :)

src/Raven.Server/Rachis/TimeoutEvent.cs
```diff
@@ -117,27 +117,41 @@ public void DisableTimeout()
             }
         }
 
+        private int _inProgress;
+
         public void ExecuteTimeoutBehavior()
         {
+            var action = _timeoutHappened;
+
+            if (Interlocked.CompareExchange(ref _inProgress, 1, 0) == 1)
```
ayende, line 124: What is the dead lock that you are running into here?
And why not handle that using the already existing `lock (this)` by changing this to be `Monitor.TryEnter` ?

karmeli87: p1 is the timer callback (`SwitchToCandidateStateOnTimeout`)
p2 switching the rachis state (`SetNewState`)

p1 enters the lock -> p2 opens write tx -> p2 try to enter the lock but blocked -> p1 tries to open write tx.

karmeli87: Not sure how `Monitor.TryEnter` would help.
I need that the execution of the callback to be outside of the lock.

ppekrol (approved):

src/Raven.Server/Rachis/TimeoutEvent.cs
```diff
@@ -117,27 +117,41 @@ public void DisableTimeout()
...
 
+        private int _inProgress;
+
         public void ExecuteTimeoutBehavior()
         {
-            lock (this)
-            {
-                if (Disable)
-                    return;
+            var action = _timeoutHappened;
```
ayende: I don't like this.
We need to do this, but we have to capture the `_timouteHappened` value _inside_ the lock

ayende (approved):

src/Raven.Server/Rachis/TimeoutEvent.cs
karmeli87: done

ayende (approved):

ppekrol:
@karmeli87 we are getting a lot of timeouts in the tests results, can you check that?

PR #9348, opened 2019-06-05
RDBS-11 Dispose single patch result after each executing in bulk operation

haludi:

### DIFF
--- a/src/Raven.Server/Documents/Document.cs
+++ b/src/Raven.Server/Documents/Document.cs
@@ -11,6 +11,7 @@ namespace Raven.Server.Documents
     {
         public static readonly Document ExplicitNull = new Document();
 
+        private bool _disposed;
         private ulong? _hash;
         private bool _metadataEnsured;
 
@@ -84,9 +85,19 @@ namespace Raven.Server.Documents
 
         public void Dispose()
         {
+            if (_disposed)
+                return;
+
             Id?.Dispose();
+            Id = null;
+
             LowerId?.Dispose();
+            LowerId = null;
+
             Data?.Dispose();
+            Data = null;
+
+            _disposed = true;
         }
     }
 }
--- a/src/Raven.Server/Documents/Document.cs
+++ b/src/Raven.Server/Documents/Document.cs
@@ -7,7 +7,7 @@ using Sparrow.Json.Parsing;
 
 namespace Raven.Server.Documents
 {
-    public class Document
+    public class Document : IDisposable
     {
         public static readonly Document ExplicitNull = new Document();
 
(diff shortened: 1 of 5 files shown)


src/Raven.Server/Documents/Document.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Raven.Server.Documents
 {
-    public class Document
+    public class Document : IDisposable
```
arekpalinski, line 10: This would require to change how we deal with `Document` all over the places where we use that object because we needed to dispose it everywhere. We typically just dispose `BlittableJsonReaderObject`, `LazyStringValue` explicitly. So instead of introducing `IDisposable` here we can dispose what's needed inside the dispose method of `PatchResult`, right?

haludi: What can be the problem? that I will dispose an external value that we still need?

src/Raven.Server/Documents/Queries/AbstractQueryRunner.cs
```diff
@@ -131,17 +140,18 @@ protected Task<IOperationResult> ExecuteDelete(IndexQueryServerSide query, Index
             onProgress(progress);
 
             var result = new BulkOperationResult();
+            void RetrieveDetails(IBulkOperationDetails details) => result.Details.Add(details);
```
ppekrol, line 146:
```
foreach (var document in results.Results)
{
	token.Delay();
	resultIds.Enqueue(document.Id.ToString());
}
```

Can we dispose the document there also (~10 lines up)?

ppekrol: I've added this.

ppekrol (approved):

src/Raven.Server/Documents/Document.cs
karmeli87: Question: 
I know that we cache `LazyStringValue`s inside the context.
Might be that the `Id`/`LowerId` are cached and we are disposing them?

ppekrol: IIRC the JsonOperationContext contains fieldName cache, and this is not related here, because the Document.Id and LowerId comes from allocator. I might be wrong... @ayende do you remember?

haludi: I and @karmeli87 followed it and the pointer for the `LazyString` comes from the table itself (Memory Map File) and in this situation the `LazyString` object will not dispose the memory. It will dispose in a situation when the memory was  allocated specially for the `LazyString` but it is safe to dispose in this situation too. There is a reuse of the `LazyString` object but the context will renew the `LasyString` object before so no problem to dispose the `LasyString` object. The `FieldName` cache related to `DocumentBuilder` while it building a document.

ayende (approved):

src/Raven.Server/Documents/Document.cs
ayende: If you read from the mmap blittable, that gives you a lazy string instance on the mmap, nothing is allocated in this case.

PR #9359, opened 2019-06-10
Raven db 13503 Integer addition in projection while using "store files" with indexes

Or-Geva:
…" with indexes

### DIFF
--- a/src/Raven.Server/Documents/Patch/BlittableObjectInstance.cs
+++ b/src/Raven.Server/Documents/Patch/BlittableObjectInstance.cs
@@ -142,7 +142,19 @@ namespace Raven.Server.Documents.Patch
                                         value = string.Empty;
                                         break;
                                 }
-                                _value = value;
+                                
+                                if (long.TryParse(value, out var valueAsLong))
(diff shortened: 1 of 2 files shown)


Or-Geva:
https://issues.hibernatingrhinos.com/issue/RavenDB-13503

src/Raven.Server/Documents/Patch/BlittableObjectInstance.cs
```diff
@@ -142,7 +142,19 @@ public BlittableObjectProperty(BlittableObjectInstance parent, string property)
                                         value = string.Empty;
                                         break;
                                 }
-                                _value = value;
+                                
+                                if (long.TryParse(value, out var valueAsLong))
```
ayende: I think we should just first do double, then long.
This is because JS have doubles,not longs

Or-Geva: Removed long since JavaScript numbers are always stored as double precision floating point.

ayende (approved):

PR #9361, opened 2019-06-11
RDBS-11 Fixing for memory leaks in patch and tests

haludi:

### DIFF
--- a/src/Sparrow/ByteString.cs
+++ b/src/Sparrow/ByteString.cs
@@ -698,7 +698,7 @@ namespace Sparrow
 
 
         private const int ExternalFastPoolSize = 16;
-        private int _externalAlignedSize = 0;
+        private readonly int _externalAlignedSize = (sizeof(ByteStringStorage) + (sizeof(long) - sizeof(ByteStringStorage) % sizeof(long)));
         private int _externalCurrentLeft = 0;
         private int _externalFastPoolCount = 0;
         private readonly IntPtr[] _externalFastPool = new IntPtr[ExternalFastPoolSize];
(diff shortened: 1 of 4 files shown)


src/Sparrow/ByteString.cs
```diff
@@ -698,7 +698,7 @@ public int SizeLeft
 
 
         private const int ExternalFastPoolSize = 16;
-        private int _externalAlignedSize = 0;
+        private readonly int _externalAlignedSize = (sizeof(ByteStringStorage) + (sizeof(long) - sizeof(ByteStringStorage) % sizeof(long)));
```
ayende: why?

haludi: As I understand it, it shouldn't change so to be more explicit about it

ayende: The value doesn't make sense to me.
You `ByteStringStorage` size is 19, right?
So this translate to 16, but I don't know why

haludi: The size is calculated in run time (sizeof(struct) is not constant !?!?) 
and in my computer sizeof(ByteStringStorage) =  24
so: 24 + (8 - 24 % 8) = 32

haludi: I actually tried to make it `const` but I found I can't do it.
Is it make sense to make it static? One value for all the ByteStringContext objects of the same generic type? for better memory usage (less space and more likely to be on cash)?

ayende: This cannot change, no.
Make this a `const` with value of 32, yes. 
Then create `Debug.Assert` that will validate that this is the same value, and add a comment explaining why

ppekrol (approved):

ayende (approved):

PR #9365, opened 2019-06-11
RavenDB-13603: v4.2 on Windows 7 / Server 2008 ("Windows Ver < 6.2")

aviviadi:


src/Sparrow.Server/Platform/Pal.cs
```diff
@@ -34,7 +34,18 @@ static Pal()
             }
             else if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
-                fromFilename = Environment.Is64BitProcess ? $"{toFilename}.win.x64.dll" : $"{toFilename}.win.x86.dll";
+                var win7 = "";
+                var osdesc = RuntimeInformation.OSDescription.Split(new[] { "Windows " }, StringSplitOptions.RemoveEmptyEntries);
+                if (osdesc.Length > 1)
```
ppekrol: use `PlatformDetails.IsWindows8OrNewer` for this.

src/Raven.Pal/build-all-windows.bat
```diff
@@ -13,18 +13,38 @@ cmd.exe /c ""%vcbin%\amd64\vcvars64.bat" & "%vcbin%\amd64\cl" -Felibrvnpal.x64.d
 copy librvnpal.x64.dll artifacts\librvnpal.win.x64.dll
 del *.obj librvnpal.x*
 
-echo =================================================
+cmd.exe /c ""%vcbin%\vcvars32.bat" & "%vcbin%\cl" -Felibrvnpal.win7x86.dll -I inc /D RVN_WIN7 /D WINVER=0x0601 /D _WIN32_WINNT=0x0601 /D NTDDI_VERSION=0x6010000 /D WINNT=1 /O2 /analyze /sdl /LD src\*.c src\win\*.c /link"
```
ppekrol, line 16: I think it would be best to add those `WINVER` (and friends) to the other DLLs. Those need to match Win8.

ppekrol (changes requested):

src/Sparrow.Server/Platform/Pal.cs
```diff
@@ -34,7 +34,18 @@ static Pal()
...
+                {
```
ayende: Is there no other way to do this?

src/Sparrow.Server/Platform/Pal.cs
aviviadi: We said do not want to use that

ppekrol: Why we do not want to use that?

aviviadi: We decided not to use PlatformDetails (and eventually throw it away).
I can copy the methods to Pal.cs though

src/Raven.Pal/build-all-windows.bat
aviviadi: Another dll for Win8 or do you want to add WIN8 flags also for WIN10 ?

src/Sparrow.Server/Platform/Pal.cs
ppekrol: Let's use them and when we will be throwing them away, you can refactor.

src/Sparrow.Server/Platform/Pal.cs
aviviadi: OSDescription can give you Windows OS version in dotnet core.. We can parse it in many ways, but I am not familiar with another way to get the OS version in non-string corefx call

src/Sparrow.Server/Platform/Pal.cs
aviviadi: done

src/Raven.Pal/build-all-windows.bat
ppekrol: single dll, exactly the one that we have now, but with correct flags, so we will not add features that we do not have in the future.

PR #9366, opened 2019-06-11
RDBC-328: Moved initalization of JsonSerializer.NullValueHandling val…

maximburyak:
…ue to conventions, so it could be modified by Conventions.CustomizeJsonSerializer

### DIFF
--- a/src/Raven.Client/Documents/Conventions/DocumentConventions.cs
+++ b/src/Raven.Client/Documents/Conventions/DocumentConventions.cs
@@ -296,7 +296,8 @@ namespace Raven.Client.Documents.Conventions
 
 
         /// <summary>
-        ///     Register an action to customize the json serializer used by the <see cref="DocumentStore" /> for deserializations
+        ///     Register an action to customize the json serializer used by the <see cref="DocumentStore" /> for deserializations.
+        ///     When creating a JsonSerializer, the CustomizeJsonSerializer is always called before CustomizeJsonDeserializer
         /// </summary>
         public Action<JsonSerializer> CustomizeJsonDeserializer
         {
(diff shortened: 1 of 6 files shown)


ayende (approved):

ppekrol:
@maximburyak can you check the failing tests?

maximburyak:
Fixed those tests. It seems that we are breaking behavior (again?) in relation to pre https://github.com/karmeli87/ravendb/commit/95617a3787fcce2a4beca3f46fa7cb1997007f03 condition. @arekpalinski @ayende what do you think?

maximburyak:
@ayende for example there is a test that stores documents with no value in "Name" field and then asks to retrieve all documents in that collection ordered by "Name" field. If we won't send the "Name" field at all when we store that document, the query above won't retrieve that document at all because it is not indexed in the first place.

maximburyak:
hmm, some more failing tests

arekpalinski:
The change we made in https://github.com/karmeli87/ravendb/commit/95617a3787fcce2a4beca3f46fa7cb1997007f03 was a breaking change because it was set globally. 

As I understand we need this:

`_serializer.Value.NullValueHandling = NullValueHandling.Ignore;`

in projections, right? CC @aviv86 

Then maybe we should have a dedicated convention in addition to `DeserializeEntityFromBlittable` - let's call it `DeserializeProjectedEntityFromBlittable` and use it where necessary to handle the projections? (Would it make sense to have it configurable for projections? Maybe we just want to stick to `NullValueHandling.Ignore` then?)

Given that we'd have separate conventions we can have both: backward compatibility and properly handled projections. Thoughts?

maximburyak:
I ended up introducing a CustomizeJsonDeserializer function, that will allow to avoid using blittables for that purpose

arekpalinski (approved):

ppekrol:
@maximburyak I've spoke with @arekpalinski and we agree that your solution is correct. This will unfortunately break what @karmeli87 did in the past there, but we think that the behavior there was incorrect and we promised too much. Drawback of this is that @lahma will have to do small enhancement to his solution during next client update (like you did in the test).

ppekrol (approved):

arekpalinski (approved):

PR #9367, opened 2019-06-11
RavenDB-13682 - can get the distance from spatial queries

ayende:


src/Raven.Server/Documents/Document.cs
```diff
@@ -72,6 +73,8 @@ public void EnsureMetadata(IMetadataModifier modifier = null)
                 mutatedMetadata[Constants.Documents.Metadata.LastModified] = LastModified;
             if (IndexScore.HasValue)
                 mutatedMetadata[Constants.Documents.Metadata.IndexScore] = IndexScore;
+            if (IndexScore.HasValue)
```
ppekrol: Should be `Distance`

src/Raven.Server/Documents/Queries/Results/GraphQueryResultRetriever.cs
```diff
@@ -116,7 +116,7 @@ internal Document ProjectFromMatch(Match match, JsonOperationContext context)
                     key = fieldToFetch.ProjectedName ?? (fieldToFetch.ProjectedName ?? fieldToFetch.Name.Value);
                     fieldVal = GetFunctionValue(fieldToFetch, null, args);
 
-                    var immediateResult = AddProjectionToResult(item, 1f, FieldsToFetch, result, key, fieldVal);
```
ppekrol, line 118: where are we setting this now?

src/Raven.Server/Documents/Queries/Results/GraphQueryResultRetriever.cs
```diff
@@ -131,7 +131,7 @@ internal Document ProjectFromMatch(Match match, JsonOperationContext context)
                             if (TryGetValue(fieldToFetch, d, null, null, out key, out fieldVal) == false)
                                 continue;
                             d.EnsureMetadata();
-                            var immediateResult = AddProjectionToResult(d, 1f, FieldsToFetch, result, key, fieldVal);
```
ppekrol, line 133: where are we setting this now?

src/Raven.Server/Documents/Queries/Results/GraphQueryResultRetriever.cs
```diff
@@ -142,7 +142,7 @@ internal Document ProjectFromMatch(Match match, JsonOperationContext context)
                             if (TryGetValue(fieldToFetch, doc, null, null, out key, out fieldVal) == false)
                                 continue;
                             doc.EnsureMetadata();
-                            var immediateResult = AddProjectionToResult(doc, 1f, FieldsToFetch, result, key, fieldVal);
```
ppekrol, line 144: where are we setting this now?

src/Raven.Server/Documents/Queries/Results/GraphQueryResultRetriever.cs
```diff
@@ -33,10 +33,10 @@ public class GraphQueryResultRetriever : QueryResultRetrieverBase
 
         public Document Get(Document doc)
         {
-            return GetProjectionFromDocument(doc, null, 0, FieldsToFetch, _context, null);
```
ppekrol, line 36: where are we setting this now?

src/Raven.Server/Documents/Queries/CollectionQueryEnumerable.cs
```diff
@@ -169,7 +169,7 @@ public bool MoveNext()
                     _innerCount++;
 
                     var doc = _fieldsToFetch.IsProjection
-                        ? _resultsRetriever.GetProjectionFromDocument(_inner.Current, null, 0f, _fieldsToFetch, _context, null)
```
ppekrol, line 172: where are we setting this now?

src/Raven.Server/Documents/Queries/CollectionQueryEnumerable.cs
```diff
@@ -273,7 +273,7 @@ private int Initialize()
                         count++;
 
                         var doc = _fieldsToFetch.IsProjection
-                            ? _resultsRetriever.GetProjectionFromDocument(document, null, 0f, _fieldsToFetch, _context, null)
```
ppekrol, line 276: where are we setting this now?

src/Raven.Server/Documents/Queries/Results/QueryResultRetrieverBase.cs
```diff
@@ -55,9 +58,25 @@ protected QueryResultRetrieverBase(DocumentDatabase database, IndexQueryServerSi
...
+        {
+            if (doc == null)
+                return;
+
+            doc.IndexScore = score;
+            if (_query.Distances != null)
+            {
+                var distanceInRadians = _query.Distances.Get(index);
+
+                doc.Distance = distanceInRadians * DistanceUtils.RADIANS_TO_DEGREES * DistanceUtils.MILES_TO_KM;
```
ppekrol: is this valid?

ppekrol: BTW, the comparator has a spatial field with option where you have access to units. Maybe we can take leverage from that?

ppekrol (commented):
In general looks good, but I'm a bit worried about those changes around IndexScore, we need to be extra careful there.

src/Raven.Server/Documents/Queries/CollectionQueryEnumerable.cs
ayende: FinishDocumentSetup

src/Raven.Server/Documents/Queries/CollectionQueryEnumerable.cs
ayende: FinishDocumentSetup

src/Raven.Server/Documents/Queries/Results/GraphQueryResultRetriever.cs
ayende: FinishDocumentSetup

src/Raven.Server/Documents/Queries/Results/GraphQueryResultRetriever.cs
ayende: FinishDocumentSetup

src/Raven.Server/Documents/Queries/Results/GraphQueryResultRetriever.cs
ayende: FinishDocumentSetup

src/Raven.Server/Documents/Queries/Results/GraphQueryResultRetriever.cs
ayende: FinishDocumentSetup

src/Raven.Server/Documents/Queries/Results/QueryResultRetrieverBase.cs
ayende: No, it isn't. This was meant just to get me working on _something_ so we'll have a result.

src/Raven.Server/Documents/Indexes/Static/Spatial/SpatialDistanceFieldComparatorSource.cs
```diff
@@ -112,6 +123,11 @@ public override void SetNextReader(IndexReader reader, int docBase, IState state
             public override IComparable this[int slot] => _values[slot];
         }
 
+        internal float? GetDistance(int documentId)
```
arekpalinski: There is no usage of this method?

ayende:
That should be everything for this feature, I think

src/Raven.Client/Documents/Queries/Spatial/DynamicSpatialField.cs
```diff
@@ -37,5 +37,13 @@ public override string ToField(Func<string, bool, string> ensureValidFieldName)
     public abstract class DynamicSpatialField
     {
         public abstract string ToField(Func<string, bool, string> ensureValidFieldName);
+
+        public double RoundFactor;
+
+        public DynamicSpatialField RoundTo(double factor)
```
ppekrol, line 43: I think we need to return either WktField or PointField, so probably generic would have to be used

src/Raven.Client/Documents/Session/AbstractDocumentQuery.Spatial.cs
```diff
@@ -86,13 +86,19 @@ public void OrderByDistance(DynamicSpatialField field, double latitude, double l
 
             AssertIsDynamicQuery(field, nameof(OrderByDistance));
 
-            OrderByDistance($"'{field.ToField(EnsureValidFieldName)}'", latitude, longitude);
+            OrderByDistance($"'{field.ToField(EnsureValidFieldName)}'", latitude, longitude, field.RoundFactor);
         }
 
-        /// <inheritdoc />
```
ppekrol: We shouldn't remove this probably?

src/Raven.Client/Documents/Session/AbstractDocumentQuery.Spatial.cs
```diff
@@ -103,13 +109,19 @@ public void OrderByDistance(DynamicSpatialField field, string shapeWkt)
...
         }
 
         /// <inheritdoc />
         public void OrderByDistance(string fieldName, string shapeWkt)
         {
-            OrderByTokens.AddLast(OrderByToken.CreateDistanceAscending(fieldName, AddQueryParameter(shapeWkt)));
+            OrderByDistance(fieldName, shapeWkt);
+        }
+
+        public void OrderByDistance(string fieldName, string shapeWkt, double roundFactor)
```
ppekrol, line 123: can we add `inheritdoc` ?

src/Raven.Server/Documents/Indexes/Static/Spatial/SpatialDistanceFieldComparatorSource.cs
```diff
@@ -29,14 +40,28 @@ public class SpatialDistanceFieldComparator : FieldComparator
...
+            private double _roundFactor;
+
+            public SpatialUnits Units => _spatialField.Units;
 
             private IndexReader _currentIndexReader;
 
-            public SpatialDistanceFieldComparator(SpatialField spatialField, Point origin, int numHits)
+            public double Get(int doc)
+            {
+                _cache.TryGetValue(doc, out var cache);
```
arekpalinski: What if `TryGetValue` returns `false`? Is it fine to ignore it?

src/Raven.Server/Documents/Queries/Dynamic/DynamicQueryToIndexMatcher.cs
```diff
@@ -174,8 +174,6 @@ internal DynamicQueryMatchResult ConsiderUsageOfIndex(DynamicQueryMapping query,
 
                     if (field.Spatial != null)
                     {
-                        Debug.Assert(indexField.Spatial != null);
```
ppekrol, line 177: Not needed?

src/Raven.Server/Documents/Indexes/Static/Spatial/SpatialDistanceFieldComparatorSource.cs
```diff
@@ -101,12 +145,66 @@ private double CalculateDistance(int doc, IState state)
...
+                        break;
+                    case SpatialUnits.Miles:
+                    default:
+                        break;
+                }
+
+                return dist;
+            }
+
+            public static double HaverstineDistance(double lat1, double lng1, double lat2, double lng2)
```
arekpalinski: It returns value in miles, right? Let's rename the method to `HaverstineDistanceInMiles` to be explicit about that

src/Raven.Server/Documents/Indexes/Static/Spatial/SpatialDistanceFieldComparatorSource.cs
```diff
@@ -101,12 +145,66 @@ private double CalculateDistance(int doc, IState state)
...
+                    var radtheta = DistanceUtils.DEGREES_TO_RADIANS * theta;
+                    var dist = Math.Sin(radlat1) * Math.Sin(radlat2) + Math.Cos(radlat1) * Math.Cos(radlat2) 
+                        * Math.Cos(radtheta);
+                    if (dist > 1)
+                    {
+                        dist = 1;
+                    }
+                    dist = Math.Acos(dist);
+                    dist = dist * DistanceUtils.RADIANS_TO_DEGREES;
+                    dist = dist * 60 * 1.1515;
```
arekpalinski: What is `60 * 1.1515` ?

src/Raven.Server/Documents/Indexes/Static/Spatial/SpatialDistanceFieldComparatorSource.cs
```diff
@@ -101,12 +145,66 @@ private double CalculateDistance(int doc, IState state)
...
+            {
+                double result = 0;
+
+                double v = lat1 - lat2;
+                result += (v * v);
+
+                v = lng2 - lng2;
+                result += (v * v);
+
+                return result;
```
arekpalinski, line 217: Based on the above call of this method I assume it should return value in miles? Doesn't it mean we need to convert the longitude and latitude differences calculated here into miles?

ppekrol:
@ayende I think some of the changes are causing tests to hang. It looks like `QueriesWithCustomFunctions` are affected.

src/Raven.Server/Documents/Indexes/Static/Spatial/SpatialDistanceFieldComparatorSource.cs
ayende: Yes, we'll return `0` then

ayende: But `null` is probably better

src/Raven.Server/Documents/Indexes/Static/Spatial/SpatialDistanceFieldComparatorSource.cs
ayende: This is for Cartesian, so no fixed measurement, whatever that values says it is.
This is useful for spatial on flat locations (such as places in warehouse), no sense is miles/km, the unit is whatever the user say it is

src/Raven.Client/Documents/Queries/Spatial/DynamicSpatialField.cs
ayende: No, we can return this safely, we don't care what the actual type is.
There is a virtual method to handle this

src/Raven.Server/Documents/Queries/Dynamic/DynamicQueryToIndexMatcher.cs
ayende: And wrong, it is fine to have two fields that match but don't have the same spatial info

ayende:
Let's see how the next batch of test run, I'll look at it then

src/Raven.Server/Documents/Indexes/Static/Spatial/SpatialDistanceFieldComparatorSource.cs
arekpalinski: My comment didn't make sense. I just noticed we do this:
https://github.com/ravendb/ravendb/pull/9367/files#diff-e5a57e10a19c3b4fc6617d363b0188a1R154

only for Haverstine distance, not for Cartesian.
https://github.com/ravendb/ravendb/pull/9367/files#diff-e5a57e10a19c3b4fc6617d363b0188a1R150

arekpalinski: Right. Same as commented above:

My comment didn't make sense. I just noticed we do this:
https://github.com/ravendb/ravendb/pull/9367/files#diff-e5a57e10a19c3b4fc6617d363b0188a1R154

only for Haverstine distance, not for Cartesian.
https://github.com/ravendb/ravendb/pull/9367/files#diff-e5a57e10a19c3b4fc6617d363b0188a1R150

ayende: You were correct at the time, I changed that based on your comment

src/Raven.Client/Constants.cs
```diff
@@ -161,6 +161,8 @@ private Metadata()
 
                 public const string IndexScore = "@index-score";
 
+                public const string SpatialResult = "@spatial-result";
```
ppekrol: maybe just `@spatial` to make it simpler and shorter?

src/Raven.Server/Documents/Indexes/Static/Spatial/SpatialDistanceFieldComparatorSource.cs
```diff
@@ -96,17 +142,85 @@ private double CalculateDistance(int doc, IState state)
...
+                var result = new SpatialResult
+                {
+                    Latitude = pt.GetY(),
+                    Longitude = pt.GetX()
+                };
+
+                if (_isGeo == false)
+                {
+                    result.Distance = CartesianDistance(pt.GetY(), pt.GetX(), _originPt.GetY(), _originPt.GetX());
+                    return result;
```
ppekrol, line 160: what if we want KM?

src/Raven.Server/Documents/Queries/CollectionQueryEnumerable.cs
ppekrol: I do not think this is the case in this flow? But I see inconsistency here, because for projection we were setting 0 and `null` otherwise. I think it should be set to 1 regardless if this is projection or not.

src/Raven.Server/Documents/Queries/CollectionQueryEnumerable.cs
ppekrol: I do not think this is the case in this flow? But I see inconsistency here, because for projection we were setting 0 and `null` otherwise. I think it should be set to 1 regardless if this is projection or not.

src/Raven.Server/Documents/Queries/Results/GraphQueryResultRetriever.cs
ppekrol: I think we should set 1 here?

src/Raven.Server/Documents/Queries/Results/GraphQueryResultRetriever.cs
ppekrol: I do not think this is the case here?

src/Raven.Server/Documents/Queries/Results/GraphQueryResultRetriever.cs
ppekrol: I do not think this is the case here?

src/Raven.Server/Documents/Queries/Results/GraphQueryResultRetriever.cs
ppekrol: I do not think this is the case here?

src/Raven.Server/Documents/Indexes/Static/Spatial/SpatialDistanceFieldComparatorSource.cs
ayende: That isn't how it works for cartesian, that is arbitrary system

ppekrol:
Rebased: https://github.com/ravendb/ravendb/pull/9759

PR #9370, opened 2019-06-12
RavenDB-12800 Adding option to stop ETL batch if size limit is exceed…

arekpalinski:
…ed (default: 64MB)

### DIFF
--- a/src/Raven.Server/Config/Categories/EtlConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/EtlConfiguration.cs
@@ -1,6 +1,7 @@
 using System.ComponentModel;
 using Raven.Server.Config.Attributes;
 using Raven.Server.Config.Settings;
+using Sparrow;
 
 namespace Raven.Server.Config.Categories
 {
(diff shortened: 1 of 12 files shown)


ppekrol (approved):

src/Raven.Server/Config/Categories/EtlConfiguration.cs
```diff
@@ -34,5 +35,10 @@ public class EtlConfiguration : ConfigurationCategory
         [ConfigurationEntry("ETL.MaxFallbackTimeInSec", ConfigurationEntryScope.ServerWideOrPerDatabase)]
         public TimeSetting MaxFallbackTime { get; set; }
 
+        [Description("Maximum batch size of transformed data in megabytes that will be sent to the destination in a single batch")]
+        [DefaultValue(64)]
```
ayende, line 39: Isn't this too big?

arekpalinski: We have 3 "regular" stop conditions to complete the batch and send the data:
- 30 sec duration
- 8192 docs
- 64MB of data

I believe that most cases 8192 will cause batches to complete. The default I set here - 64MB would be in my opinion mostly applicable if someone plays with attachments so I didn't limit it too much. What do you think?

BTW I'll add the info that attachments are also accounted to the configuration description

ayende: Okay, that looks good

PR #9373, opened 2019-06-13
RavenDB-13693 RavenDB should understand resource limits for burstable…

ayende:
… instances

### DIFF
--- a/src/Raven.Server/RavenServer.cs
+++ b/src/Raven.Server/RavenServer.cs
@@ -60,6 +60,7 @@ using Sparrow;
 using Sparrow.Json;
 using Sparrow.Json.Parsing;
 using Sparrow.Logging;
+using Sparrow.Threading;
 
 namespace Raven.Server
 {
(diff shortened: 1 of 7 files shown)


src/Raven.Server/RavenServer.cs
```diff
@@ -284,7 +301,119 @@ void ConfigureKestrel(KestrelServerOptions options)
...
+                }
+
+                return new DynamicJsonValue
+                {
+                    [nameof(Used)] = Used,
+                    [nameof(BaseCredits)] = BaseCredits,
+                    [nameof(MaxCredits)] = MaxCredits,
+                    [nameof(Threshold)] = Threshold,
+                    [nameof(RemainingCpuCredits)] = RemainingCpuCredits,
+                    [nameof(ThresholdReleasValue)] = ThresholdReleasValue,
```
ppekrol: typo: `ThresholdReleaseValue`

src/Raven.Server/RavenServer.cs
```diff
@@ -284,7 +301,119 @@ void ConfigureKestrel(KestrelServerOptions options)
...
+                    [nameof(CreditsAccuredPerSecond)] = CreditsAccuredPerSecond,
```
ppekrol: typo: `CreditsAccuredPerSecond`... `CreditsOccuredPerSecond`? `CreditsAquiredPerSecond`? Not sure what you meant.

ppekrol (approved):

src/Raven.Server/Routing/RequestRouter.cs
```diff
@@ -122,6 +122,15 @@ public async ValueTask HandlePath(RequestHandlerContext reqCtx)
...
                 if (reqCtx.Database != null)
                 {
+                    if (_ravenServer.CpuCreditsAlertRaised.IsRaised())
+                    {
+                        if (_ravenServer.Configuration.Server.CpuCreditsExhaustionShouldTriggerFailover)
+                        {
+                            RejectRequestBecauseOfCpuThreshold(context);
+                            return;
+                        }
+                    }
```
arekpalinski, line 132: Maybe we can add an option to throttle requests then?

arekpalinski: I mean when 'CpuCreditsExhaustionShouldTriggerFailover' is `false` then having additional option that will throttle the requests and reduce the load?

ppekrol:
Couple of tests are failing here

src/Raven.Server/Routing/RequestRouter.cs
ayende: That would just defer it, no? 
We can drop every Nth request, but that would probably suck

arekpalinski: If we slow down the incoming load then we should be able to recover from current state earlier, don't you think? Once we manage to recover from it then slowly decrease the throttling (not immediately)

PR #9374, opened 2019-06-13
RavenDB-13694 [Testing Environment] Unloading locked database crashes…

ayende:
… server

### DIFF
--- a/src/Sparrow.Server/Utils/TaskExecutor.cs
+++ b/src/Sparrow.Server/Utils/TaskExecutor.cs
@@ -119,7 +119,16 @@ namespace Sparrow.Server.Utils
                 if(Interlocked.CompareExchange(ref _callback, null, callback) != callback)
                     return;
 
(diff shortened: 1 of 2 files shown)


src/Sparrow.Server/Utils/TaskExecutor.cs
```diff
@@ -116,10 +116,19 @@ public void Execute(object state)
...
                     return;
 
-                callback(state);
+                try
+                {
+                    callback(state);
+                }
+                catch (Exception e)
+                {
+                    var logger = Logging.LoggingSource.Instance.GetLogger<RunOnce>("TaskExecutor");
```
ppekrol, line 128: Do we want to create new logger instance per each exception?

ayende: We saw it once in 4 years, I don't think that this is going to be a hot spot

PR #9379, opened 2019-06-13
RavenDB-10520 IDocumentQuery and IRawDocumentQuery should not inherit from IEnumerable

ppekrol:


ayende (changes requested):
Pawel,
This is this way on purpose.

```
session.Advanced.DocumentQuery<User>().Where(x=>x.Name == "ayende")
```

Right now, this will generate complication error.
With your change, it will silently materialize the query and handle this via in memory filter.

ayende (commented):
Sorry, read it the wrong way, I think.
But isn't this a breaking change for us?

ayende (approved):
And it is for 5.0 and it is too early for me to notice stuff like "details"

PR #9381, opened 2019-06-14
RavenDB-13703 Allow disposing of EmbeddableDocumentStore even if it was not initialized

danielmarbach:
It should be possible to dispose an `EmbeddableDocumentStore` even if `Initialize` wasn't called. Without this patch `Dispose` throws an exception which in my opinion doesn't make sense and might masquerade other problems the client code has


ayende (approved):

danielmarbach:
What should I change in the commit message? I have no visibility into the error message

ppekrol:
@danielmarbach It requires Issue # from our BugTracker, but it is OK, I will handle that. Thanks for contribution!

PR #9383, opened 2019-06-14
RavenDB-7725

ayende:
- Update to a custom build of ClrMd that allows use to do passive attach.
- On Linux, start the Raven.Debug process, but ask it to wait. Then grant it debug permissions on ourselves, let it run, then remove the permissions.
- No need for any additional security configurations


src/Raven.Server/Documents/Handlers/Debugging/ThreadsHandler.cs
```diff
@@ -19,12 +21,6 @@ public class ThreadsHandler : RequestHandler
         [RavenAction("/admin/debug/threads/stack-trace", "GET", AuthorizationStatus.Operator)]
         public Task StackTrace()
         {
-            if (PlatformDetails.RunningOnPosix)
```
ppekrol, line 22: I think that we need to keep it, but for MacOS?

src/Raven.Server/Raven.Server.csproj
```diff
@@ -102,6 +102,7 @@
     <PackageReference Include="Microsoft.AspNetCore.WebSockets" Version="2.2.1" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.1.0" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.1.0" />
+    <PackageReference Include="Microsoft.Diagnostics.Runtime-ravendb" Version="1.0.7" />
```
ppekrol: I think we can remove this?

ppekrol (commented):
@ayende did you made any changes to clrmd or just compiled master branch?

tools/Raven.Debug/CommandLineApp.cs
```diff
@@ -139,10 +144,20 @@ private static int ExitWithError(string errMsg, CommandLineApplication cmd)
 
             var threadInfoList = new List<ThreadInfo>();
 
-            using (var dataTarget = DataTarget.AttachToProcess(processId, attachTimeout))
+            using (var dataTarget = DataTarget.AttachToProcess(processId, attachTimeout, AttachFlag.Passive))
```
ppekrol, line 147: Do we want always to attach passively? IIRC it is less accurate, on Windows we could stay with what we had?

ayende:
> @ayende did you made any changes to clrmd or just compiled master branch?

I had to remove a bit of debug code, this is available here: https://github.com/ayende/clrmd

src/Raven.Server/Documents/Handlers/Debugging/ThreadsHandler.cs
ayende: Yes

src/Raven.Server/Raven.Server.csproj
ayende: Yes

tools/Raven.Debug/CommandLineApp.cs
ayende: The speed we are seeing in generating stack traces is _huge_, even if it isn't as accurate, that is far more useful

ppekrol (approved):

arekpalinski (approved):

PR #9384, opened 2019-06-14
RavenDB-13699 

ppekrol:


src/Raven.Server/Documents/PeriodicBackup/BackupTask.cs
```diff
@@ -271,13 +272,19 @@ public async Task<IOperationResult> RunPeriodicBackup(Action<IOperationProgress>
             }
         }
 
-        private async Task UpdateRemoteConfigurationFromScript()
+        private async Task UpdateConfigurationFromScript()
         {
+            _configuration.LocalSettings = await GetBackupConfigurationFromScript(_configuration.LocalSettings, x => JsonDeserializationServer.LocalSettings(x));
```
grisha-kotler: This needs to be called separately from the cloud options.
The local backup can take a few hours and we request temp AWS credentials before it. (it can expire while we create the backup).

grisha-kotler (changes requested):

src/Raven.Server/Documents/PeriodicBackup/BackupTask.cs
ppekrol: Done

grisha-kotler (approved):

PR #9387, opened 2019-06-16
RavenDB-13697 - Allow to define separate thresholds for stopping back…

ayende:
…ground processes and rejection of database requests

### DIFF
--- a/src/Raven.Server/RavenServer.cs
+++ b/src/Raven.Server/RavenServer.cs
@@ -284,7 +284,7 @@ namespace Raven.Server
                 {
                     if (Configuration.Server.CpuCreditsBase == null ||
                         Configuration.Server.CpuCreditsMax == null)
-                        throw new InvalidOperationException("Both Server.CpuCreditsBase and Server.CpuCreditsMax must be specified");
+                        throw new InvalidOperationException($"Both {RavenConfiguration.GetKey(s=> s.Server.CpuCreditsBase)} and {RavenConfiguration.GetKey(s => s.Server.CpuCreditsMax)} must be specified");
 
                     CpuCreditsBalance.BaseCredits = Configuration.Server.CpuCreditsBase.Value;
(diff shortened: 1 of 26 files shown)


src/Raven.Server/RavenServer.cs
```diff
@@ -279,16 +279,21 @@ void ConfigureKestrel(KestrelServerOptions options)
...
-                    Configuration.Server.CpuCreditsThreshold != null)
+                    Configuration.Server.CpuCreditsExhaustionFailoverThreshold != null ||
+                    Configuration.Server.CpuCreditsExhaustionBackgroundTasksThreshold != null)
                 {
                     if (Configuration.Server.CpuCreditsBase == null ||
-                        Configuration.Server.CpuCreditsMax == null ||
-                        Configuration.Server.CpuCreditsThreshold == null)
-                        throw new InvalidOperationException("If Server.CpuCreditsBase, Server.CpuCreditsMax and Server.CpuCreditsThreshold must be specified together");
+                        Configuration.Server.CpuCreditsMax == null)
+                        throw new InvalidOperationException("Both Server.CpuCreditsBase and Server.CpuCreditsMax must be specified");
```
arekpalinski: Minor - we could use `RavenConfiguration.GetKey(x => .... )` instead of hard coded option names

src/Raven.Server/Routing/RouteScanner.cs
```diff
@@ -69,6 +69,7 @@ public class RouteScanner
                                 route.RequiredAuthorization, 
                                 route.SkipUsagesCount,
                                 route.CorsMode,
+                                route.IsDebugInformationEndpoint,
                                 route.IsDebugInformationEndpoint);
```
arekpalinski: Invalid argument - 2nd one should be `route.DisableOnCpuCreditsExhaustion` instead of `route.IsDebugInformationEndpoint`

arekpalinski (approved):

PR #9388, opened 2019-06-16
RavenDB-8451 DR support for encrypted database

talweiss1982:
A high-level explanation for this PR:
For encrypted database we need to decrypt pages there were 2 options:
1) Use crypto pager and avoid code duplication and code deviation (if a bug was found in the encryption pager), will require active transaction though.
2) Write a decryption object that holds encryption buffer pool, this can avoid having to deal with the transaction but still have to allocate the same buffers.

I decided to go with option 1.

The majority of the complexity is that the acquired pointer from the decrypted page is not sequential while the code was relying on that fact.
The solution was using a page pointer where it can either point to the same page if it was not decrypted or to a buffer if it is a decrypted page.

### DIFF
--- a/tools/Voron.Recovery/Recovery.cs
+++ b/tools/Voron.Recovery/Recovery.cs
@@ -45,7 +45,9 @@ namespace Voron.Recovery
             _pageSize = config.PageSizeInKB * Constants.Size.Kilobyte;
             _initialContextSize = config.InitialContextSizeInMB * Constants.Size.Megabyte;
             _initialContextLongLivedSize = config.InitialContextLongLivedSizeInKB * Constants.Size.Kilobyte;
-            
+
+            _masterKey = config.MasterKey;
+
             // by default CopyOnWriteMode will be true
             _copyOnWrite = !config.DisableCopyOnWriteMode;
             _config = config;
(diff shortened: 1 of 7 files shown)


ppekrol:
@talweiss1982 can we add a test? Check RavenDB_11904.cs and RavenDB_11905.cs for those.

tools/Voron.Recovery/Recovery.cs
```diff
@@ -832,13 +881,16 @@ private void WriteAttachment(BlittableJsonTextWriter writer, long totalSize, str
         private bool ValidateOverflowPage(PageHeader* pageHeader, byte* eof, long startOffset, ref byte* mem)
         {
             ulong checksum;
-            var endOfOverflow = (byte*)pageHeader + VirtualPagerLegacyExtensions.GetNumberOfOverflowPages(pageHeader->OverflowSize) * _pageSize;
+            //pageHeader might be a buffer address we need to verify we don't exceed the original memory boundary here
+            var numberOfPages = VirtualPagerLegacyExtensions.GetNumberOfOverflowPages(pageHeader->OverflowSize);
+            var sizeOfPages = numberOfPages * _pageSize;
+            var endOfOverflow = startOffset + sizeOfPages;
```
ayende: Another look here, please

ayende (approved):

talweiss1982:
@ppekrol I had to rebase this PR to use the code I added in my other PR.
Added test

arekpalinski (approved):

PR #9395, opened 2019-06-17
RavenDB-13714 - Add RemoteFolderName to Glacier Backup

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-13714

### DIFF
--- a/src/Raven.Server/Documents/PeriodicBackup/BackupTask.cs
+++ b/src/Raven.Server/Documents/PeriodicBackup/BackupTask.cs
@@ -950,7 +950,7 @@ namespace Raven.Server.Documents.PeriodicBackup
         {
             using (var client = new RavenAwsGlacierClient(settings, progress, TaskCancelToken.Token))
             {
-                var key = CombinePathAndKey(settings.RemoteFolderName, folderName, fileName);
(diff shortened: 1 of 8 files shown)


src/Raven.Server/Documents/PeriodicBackup/BackupTask.cs
```diff
@@ -950,7 +950,7 @@ private static string Pluralize(int number)
         {
             using (var client = new RavenAwsGlacierClient(settings, progress, TaskCancelToken.Token))
             {
-                var key = CombinePathAndKey(_database.Name, folderName, fileName);
+                var key = CombinePathAndKey(settings.RemoteFolderName, folderName, fileName);
```
ppekrol: I think this should be `settings.RemoteFolderName ?? _database.Name` to maintain backward compatibility

ppekrol:
This also requires a bit of Studio work (it does not compile and we need to show that field in the UI)

src/Raven.Server/Documents/PeriodicBackup/BackupTask.cs
grisha-kotler: done

grisha-kotler:
fixed and rebased

ppekrol (approved):

PR #9399, opened 2019-06-17
RavenDB-13719 Recognize SMB2 shares as not valid for dir sync

ayende:


src/Raven.Pal/src/posix/linuxonly.c
```diff
@@ -46,6 +48,7 @@ _sync_directory_allowed(int32_t dir_fd)
   case NFS_SUPER_MAGIC:
   case CIFS_MAGIC_NUMBER:
   case SMB_SUPER_MAGIC:
+  case SMB2_SUPER_MAGIC:
```
ppekrol: you have defined `SMB2_MAGIC_NUMBER` but here we are using `SMB2_SUPER_MAGIC` ? Does it compile? Maybe if it does then `SMB2_SUPER_MAGIC` comes from somewhere else?

ppekrol (commented):
I think you haven't compiled the linux files (let's revert to 42009 and recompile again)

src/Raven.Pal/src/posix/linuxonly.c
ayende: 😞 Sorry, PEBKAC. Thanks for finding it.

PR #9401, opened 2019-06-17
RavenDB-13472: Consider RetryOn when reconnecting failed replication

garayx:


src/Raven.Server/Config/Categories/ReplicationConfiguration.cs
```diff
@@ -25,6 +25,12 @@ public class ReplicationConfiguration : ConfigurationCategory
         [ConfigurationEntry("Replication.RetryReplicateAfterInSec", ConfigurationEntryScope.ServerWideOrPerDatabase)]
         public TimeSetting RetryReplicateAfter { get; set; }
 
+        [Description("Max retry timeout in seconds on replication failure.")]
+        [DefaultValue(300)]
+        [TimeUnit(TimeUnit.Seconds)]
+        [ConfigurationEntry("Replication.RetryReplicateMaxTimeoutInSec", ConfigurationEntryScope.ServerWideOrPerDatabase)]
```
ayende:
```suggestion
        [ConfigurationEntry("Replication.RetryReplicationMaxTimeoutInSec", ConfigurationEntryScope.ServerWideOrPerDatabase)]
```

ayende: Same on the prop name

ppekrol: Maybe just `Replication.RetryMaxTimeoutInSec` since we already know that this is replication

PR #9404, opened 2019-06-18
RDBCL-447 Monitoring cpu credits in a dedicated thread

karmeli87:


ayende (approved):

src/Raven.Server/RavenServer.cs
```diff
@@ -295,7 +295,18 @@ void ConfigureKestrel(KestrelServerOptions options)
                         // default to disabled
                         Configuration.Server.CpuCreditsExhaustionFailoverThreshold ?? -1;
 
-                    var _ = StartMonitoringCpuCredits();
+                    PoolOfThreads.GlobalRavenThreadPool.LongRunning(_ =>
```
grisha-kotler: Need to wait for it to finish on server dispose:
https://github.com/ravendb/ravendb/blob/0568c39f984877569deead635c74a9b35928e094/src/Raven.Server/ServerWide/ServerStore.cs#L1793

grisha-kotler (changes requested):

PR #9405, opened 2019-06-18
RavenDB-13729 - Compare exchange throws when saving null values

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-13729

### DIFF
--- a/src/Raven.Client/Documents/Session/EntityToBlittable.cs
+++ b/src/Raven.Client/Documents/Session/EntityToBlittable.cs
@@ -55,6 +55,9 @@ namespace Raven.Client.Documents.Session
             DocumentInfo documentInfo,
             bool removeIdentityProperty = true)
         {
+            if (value == null)
+                return null;
+
             if (value is ValueType ||
                 value is string ||
                 value is BlittableJsonReaderArray)
--- a/src/Raven.Client/Documents/Session/EntityToBlittable.cs
+++ b/src/Raven.Client/Documents/Session/EntityToBlittable.cs
@@ -46,8 +46,7 @@ namespace Raven.Client.Documents.Session
             return props;
         }
 
-
-        internal static object ConvertToBlittableIfNeeded(
+        internal static object ConvertToBlittableForCompareExchangeIfNeeded(
             object value,
             DocumentConventions conventions,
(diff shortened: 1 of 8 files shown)


src/Sparrow/Json/BlittableJsonReaderObject.cs
```diff
@@ -1303,7 +1303,7 @@ public static void AssertNoModifications(BlittableJsonReaderObject data, string
             }
         }
 
-        public bool Contains(LazyStringValue propertyName)
+        public bool Contains(string propertyName)
```
ppekrol: This will materialize a lot of LazyStringValues, can we avoid that. I do not see you using that anywhere?

src/Raven.Client/Documents/Session/EntityToBlittable.cs
```diff
@@ -55,6 +55,9 @@ public BlittableJsonReaderObject ConvertEntityToBlittable(object entity, Documen
             DocumentInfo documentInfo,
             bool removeIdentityProperty = true)
         {
+            if (value == null)
```
ppekrol, line 57: Is this required? If no, then we should remove this to avoid adding this

src/Sparrow/Json/BlittableJsonReaderObject.cs
ayende: Yes, this is to be avoided if possible.

src/Raven.Client/Documents/Session/EntityToBlittable.cs
ayende: I think it is, won't `ConvertEntityToBlittableInternal` throw on null?

grisha-kotler: It is required only for CompareExchange as its value can be null. Renamed the function name to ConvertToBlittableForCompareExchangeIfNeeded.

src/Sparrow/Json/BlittableJsonReaderObject.cs
grisha-kotler: done

ppekrol (approved):

PR #9406, opened 2019-06-18
RavenDB-13720 added cpu credits to SNMP

ppekrol:


src/Raven.Server/Monitoring/Snmp/Objects/Server/1.5/CpuCreditsCurrentConsumption.cs
```diff
@@ -0,0 +1,20 @@
...
+
+        public CpuCreditsCurrentConsumption(RavenServer.CpuCreditsState state) 
+            : base(SnmpOids.Server.CpuCreditsCurrentConsumption)
+        {
+            _state = state;
+        }
+
+        protected override Integer32 GetData()
+        {
+            return new Integer32((int)_state.CurrentConsumption);
```
ppekrol: @ayende is this acceptable? SNMP does not have notion of double, I can return string, but it is hard to create alerts then

ayende: See here: https://stackoverflow.com/a/2144106/6366
Is this something we can do?

ppekrol: Maybe we can use Opaque and encode Float there, looking into that.

ppekrol: Resolved by using OctetString, the conversion can be done easily on the SNMP software side (e.g. in Zabbix you can change expected value type to be numeric (float) and everything will happen automatically)

PR #9410, opened 2019-06-19
RDBS-24: Fix available memory without shared clean calculation in con…

aviviadi:
…tainers

### DIFF
--- a/src/Sparrow.Server/LowMemory/MemoryInformation.cs
+++ b/src/Sparrow.Server/LowMemory/MemoryInformation.cs
@@ -335,6 +335,7 @@ namespace Sparrow.LowMemory
                     commitedMemoryInBytes = cgroupMemoryUsage.Value;
                     fromProcMemInfo.Commited.Set(commitedMemoryInBytes, SizeUnit.Bytes);
                     fromProcMemInfo.MemAvailable.Set(maxMemoryUsage - cgroupMemoryUsage.Value, SizeUnit.Bytes);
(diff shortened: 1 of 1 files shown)


src/Sparrow.Server/LowMemory/MemoryInformation.cs
```diff
@@ -335,6 +335,7 @@ private static MemoryInfoResult GetMemoryInfoLinux(SmapsReader smapsReader)
                     commitedMemoryInBytes = cgroupMemoryUsage.Value;
                     fromProcMemInfo.Commited.Set(commitedMemoryInBytes, SizeUnit.Bytes);
                     fromProcMemInfo.MemAvailable.Set(maxMemoryUsage - cgroupMemoryUsage.Value, SizeUnit.Bytes);
+                    fromProcMemInfo.AvailableWithoutTotalCleanMemory.Set(maxMemoryUsage - cgroupMemoryUsage.Value + fromProcMemInfo.SharedCleanMemory.GetValue(SizeUnit.Bytes), SizeUnit.Bytes);
```
ayende: Can this ever be negative?

aviviadi: How? ..
It takes `maxMemoryUsage` (which is the containers max mem), decrement container's usage which always less then max usage, and adding zero or above clean mem.
Unless cgroup usage is extremely high (and there's an if condition for this block to prevent that and decide we are inside a container), then no negative number can be produced..

ayende: User define a limit that is higher than the mem? 
OS weirdness? 
Even if negative, would this break stuff?

ppekrol: @aviviadi ?

aviviadi: done

ayende (approved):

PR #9412, opened 2019-06-19
Memory "leak" Insights - for review

talweiss1982:
This PR is about memory utilization when using session.Store to bulk insert documents rather than bulkInsert.Store.
On the same workload, 32 threads each putting 8192 documents of the same size (two guid fields as strings) the bulk insert uses about 200MB steadily the session store will grow until it will throw OOM and crash.

The first place i saw that we leak some memory is the Context cached paths there is quite a large memory left there but it is not the main reason for the leak and cleaning those paths might hurt performance for some scenarios where we reset the context but keep using it for the same type of operation (e.g smuggler or bulk insert)

After cleaning the paths there still seems to be too many BlittableReaderObjects left even after calling the GC.Collect on all generations looking at them they looked like the save changes results and looking at the code we indeed never dispose of those blittable readers we even have a remark that we are doing it on purpose (will be collected when context will be reset)
There are two issues here:
1. The context (request executor from the document store) grows quite big
2. The reader objects seem to somehow survive the GC collect and we accumulate enough of them that they show on the dumpheap as a major memory leak ( i saw 200M of them and they are 114bytes big)

In conclusion, I'm not sure regarding my changes to the context field but I think it is safe to dispose of the save changes response as it is already a clone of the cached object (and I ran slow test which seems to be okay with it)

@ayende @ppekrol @arekpalinski @maximburyak @redknightlois please have a look.

### DIFF
--- a/src/Raven.Client/Documents/Session/AsyncDocumentSession.cs
+++ b/src/Raven.Client/Documents/Session/AsyncDocumentSession.cs
@@ -139,10 +139,16 @@ namespace Raven.Client.Documents.Session
 
                 if (NoTracking)
                     throw new InvalidOperationException($"Cannot execute '{nameof(SaveChangesAsync)}' when entity tracking is disabled in session.");
-
-                await RequestExecutor.ExecuteAsync(command, Context, SessionInfo, token).ConfigureAwait(false);
-                UpdateSessionAfterSaveChanges(command.Result);
-                saveChangesOperation.SetResult(command.Result);
+                try
+                {
+                    await RequestExecutor.ExecuteAsync(command, Context, SessionInfo, token).ConfigureAwait(false);
+                    UpdateSessionAfterSaveChanges(command.Result);
+                    saveChangesOperation.SetResult(command.Result);
+                }
+                finally
+                {
+                    command.Result?.Results?.Dispose();
+                }
             }
         }
     }
--- a/src/Sparrow/Json/JsonOperationContext.cs
+++ b/src/Sparrow/Json/JsonOperationContext.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Buffers;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
@@ -49,8 +50,8 @@ namespace Sparrow.Json
         }
 
         private int _numberOfAllocatedPathCaches = -1;
-        private readonly PathCacheHolder[] _allocatePathCaches = new PathCacheHolder[512];
-        private readonly Stack<MemoryStream> _cachedMemoryStreams = new Stack<MemoryStream>();
+        private PathCacheHolder[] _allocatePathCaches = new PathCacheHolder[512];
+        private Stack<MemoryStream> _cachedMemoryStreams = new Stack<MemoryStream>();
 
         private int _numberOfAllocatedStringsValues;
         private readonly FastList<LazyStringValue> _allocateStringValues = new FastList<LazyStringValue>(256);
@@ -1036,7 +1037,10 @@ namespace Sparrow.Json
 
                 _pooledArrays = null;
             }
-        }
+            //TODO:Don't PR this, just testing.
+            PathCacheHolder[] _allocatePathCaches = new PathCacheHolder[512];
+            Stack<MemoryStream> _cachedMemoryStreams = new Stack<MemoryStream>();
+    }
 
         public void Write(Stream stream, BlittableJsonReaderObject json)
         {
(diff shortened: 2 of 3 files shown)


src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -1036,7 +1037,10 @@ protected internal virtual unsafe void Reset(bool forceReleaseLongLivedAllocator
 
                 _pooledArrays = null;
             }
-        }
+            //TODO:Don't PR this, just testing.
```
ayende, line 1040: You forgot these?

ayende (commented):
I want to see more of your findings, but I think that you are on the right track.
Wonder if we are causing the data to reside in memory long enough for Gen2 ?

Why does the context grows big? Shouldn't it be reused between sessions?

src/Raven.Client/Documents/Session/AsyncDocumentSession.cs
```diff
@@ -139,10 +139,16 @@ public async Task SaveChangesAsync(CancellationToken token = default(Cancellatio
...
-                saveChangesOperation.SetResult(command.Result);
+                try
+                {
+                    await RequestExecutor.ExecuteAsync(command, Context, SessionInfo, token).ConfigureAwait(false);
+                    UpdateSessionAfterSaveChanges(command.Result);
+                    saveChangesOperation.SetResult(command.Result);
+                }
+                finally
+                {
+                    command.Result?.Results?.Dispose();
```
ayende, line 150: Can we move this to the `BatchCommand.Dispose` ? That is called anyway, no?

One thing that worries me here, though, is the `SetResult`. IIRC, we are holding the blittable in the session, and they are the same as the one we have here, no?
If we can `SaveChanges` twice, won't that break?

src/Sparrow/Json/JsonOperationContext.cs
talweiss1982: This PR is meant not for merging, it is more of a note to self what i have modified while playing with the code.
I'm not sure why i can't reply to the comment above, i'll reply here regarding Gen2.
What is going on is that memory grows pretty much linearly with the parallel level, on my machine each thread is able to peak 100MB per context. there are about 32 session threads so the 4GB of allocated memory is somewhat expected.
You're correct regarding the fact that we utilize the unmanaged memory once the session is disposed of the request executor with its context pool is returned to the document store, there is only one request executor in this case but the pool it owns has about 37 contexts in it each with 100MB.
So the blittable unmanaged memory isn't leaking!
The thing is we are generating _alot_ of reader objects and the stress on the system might be the cause of them to reach Gen2 but what i can't explain is why they are not all collected when i call GC.Collect, there are the ones held by the context in caches it is expected they won't be released but there seems to be a whole bunch of them without GC roots and they are the results of the save changes and indeed looking at the code they are not been refrenced, not sure why they are not collected every 32 sessions i invoked GC.Collect, i would expect none to survive...
I tested this on a new project, generated a bunch of unsafe object pointing to blobs of unmanaged memory and calling GC collect, one of them seems to always survive but the rest are been cleaned by the GC... I guess it is somehow been pointed by a temp variable on the stack generated by the JIT or something...

talweiss1982: One important thing I missed in my scroll above is the fact that we don't have low memory notification on the client so the ContextPool is never shrinking even after the "peak load" I think that introducing low memory machanizem on the client side even if as an optional thing could help us behave nicer on client machines.

src/Raven.Client/Documents/Session/AsyncDocumentSession.cs
talweiss1982: Good questions, 
I'm not sure if BatchCommand.Dispose will dispose of the results it seems to dispose of the commands as blittable (in the base class) while here the command is serialized to a C# object so I'm not sure those are the same instances, need to check.
I also wasn't sure about SetResult I ran slow tests and they passed so either we don't have this in our tests or the results are been cloned, either way, we need to test this scenario explicitly.
Regarding save changes twice we could clean the session commands in the first SaveChanges.

ayende:
Too complex to follow over github comments. Let's pair on this when you are at the office.

src/Raven.Client/Documents/Session/AsyncDocumentSession.cs
talweiss1982: 1. After testing this, save changes twice seems safe.
2. indeed the commands are results are not related as I expected, but we could probably move the dispose of the result into the dispose of the batch command.

talweiss1982:
I'll open a clean PR for this issue

PR #9413, opened 2019-06-19
RavenDB-13735 @refresh feature for doc update

ayende:
This allows you to define a `"@refresh: <date>` metadata property.
If expiration is enabled, at around that date, RavenDB will automatically remove the `@refresh` property.
That is all.

This is done for the sole purpose of triggering actions on document modifications.
In particular, this gives us a very nice way to defer actions when using subscriptions.

Things that we still need to decide:
* This runs using the expiration mechanism, do we enable this by default (just for refresh?)
* In a cluster, what should the behavior be? Normally, this will happen on all nodes at the same time, generating a conflict. With expiration,the document is deleted. No biggie, the conflict ends up being the same thing. 
* Should we make this use `WhoseTaskIsIt` to decide for each topology who owns the refreshes?
* When using external replication, how would that work?

### DIFF
--- a/src/Raven.Server/Documents/Expiration/ExpirationStorage.cs
+++ b/src/Raven.Server/Documents/Expiration/ExpirationStorage.cs
@@ -136,7 +136,7 @@ namespace Raven.Server.Documents.Expiration
                                     var document = _database.DocumentsStorage.Get(context, clonedId);
                                     if (document == null ||
                                         document.TryGetMetadata(out var metadata) == false || 
-                                        HasExpired(metadata, metadataPropertyToCheck, currentTime) == false)
+                                        HasPassed(metadata, metadataPropertyToCheck, currentTime) == false)
                                     {
                                         expiredDocs.Add((clonedId, null));
                                         continue;
@@ -186,7 +186,7 @@ namespace Raven.Server.Documents.Expiration
                 {
                     id = conflict.Id;
 
-                    if (HasExpired(conflict.Doc, currentTime))
+                    if (HasPassed(conflict.Doc, currentTime))
                         continue;
 
                     allExpired = false;
@@ -197,17 +197,17 @@ namespace Raven.Server.Documents.Expiration
             return (allExpired, id);
         }
 
-        public static bool HasExpired(BlittableJsonReaderObject data, DateTime currentTime)
+        public static bool HasPassed(BlittableJsonReaderObject data, DateTime currentTime)
         {
             // Validate that the expiration value in metadata is still the same.
             // We have to check this as the user can update this value.
             if (data.TryGet(Constants.Documents.Metadata.Key, out BlittableJsonReaderObject metadata) == false)
                 return false;
 
-            return HasExpired(metadata, Constants.Documents.Metadata.Expires, currentTime);
+            return HasPassed(metadata, Constants.Documents.Metadata.Expires, currentTime);
         }
 
-        private static bool HasExpired(BlittableJsonReaderObject metadata, string metadataPropertyName, DateTime currentTime)
+        private static bool HasPassed(BlittableJsonReaderObject metadata, string metadataPropertyName, DateTime currentTime)
         {
             if (metadata.TryGet(metadataPropertyName, out string expirationDate))
             {
@@ -238,7 +238,7 @@ namespace Raven.Server.Documents.Expiration
                             var doc = _database.DocumentsStorage.Get(context, ids.LowerId, throwOnConflict: true);
                             if (doc != null && doc.TryGetMetadata(out var metadata))
                             {
-                                if (HasExpired(metadata, Constants.Documents.Metadata.Expires, currentTime))
+                                if (HasPassed(metadata, Constants.Documents.Metadata.Expires, currentTime))
                                 {
                                     _database.DocumentsStorage.Delete(context, ids.LowerId, ids.Id, expectedChangeVector: null);
                                 }
@@ -275,7 +275,7 @@ namespace Raven.Server.Documents.Expiration
                         var doc = _database.DocumentsStorage.Get(context, ids.LowerId, throwOnConflict: false);
                         if (doc != null && doc.TryGetMetadata(out var metadata))
                         {
-                            if (HasExpired(metadata, Constants.Documents.Metadata.Refresh, currentTime))
+                            if (HasPassed(metadata, Constants.Documents.Metadata.Refresh, currentTime))
                             {
                                 // remove the @refresh tag
                                 metadata.Modifications = new Sparrow.Json.Parsing.DynamicJsonValue(metadata);
--- a/src/Raven.Server/Documents/Expiration/ExpirationStorage.cs
+++ b/src/Raven.Server/Documents/Expiration/ExpirationStorage.cs
@@ -18,6 +18,7 @@ namespace Raven.Server.Documents.Expiration
     public unsafe class ExpirationStorage
     {
         private const string DocumentsByExpiration = "DocumentsByExpiration";
+        private const string DocumentsByRefresh = "DocumentsByRefresh";
 
         private readonly DocumentDatabase _database;
         private readonly DocumentsStorage _documentsStorage;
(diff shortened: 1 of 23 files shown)


ayende:
@karmeli87 Please take a look as well

karmeli87:
- I would leave this as part of the expiration mechanism, it is basically very similar and fits good into it.
- Does the subscription works when the cluster is down? (if connection was already established) 
If it is, leaving this on the database level would make more sense to the user would be notified.
- How does expiration works in external replication? `refresh` is no different in that matter.
If you have a geo-distributed system, I guess you want to have this property and get notified.

ayende:
* Subscriptions don't really work if the cluster is down.
* Expiration will cause conflicts, but they will auto resolve automatically.
* I can see only doing `@refresh` on a particular node in the cluster. And having additional logic to resolve `@refresh` on conflicts. Thoughts?

ayende:
Ready for review

arekpalinski (commented):
1) I don't really like that the refresh is triggered by Expiration feature. I understand that it matches the implementation etc but I don't think it would be natural for a user. The requirement that Expiration needs to be enabled in order to make it work seems to be counter intuitive to me.
I'd prefer to have it configured separately (and separate timer under the covers actually) - note that expiration might have 60 sec delay in deletion, we might want to have it configured separately for "refresh" feature.

2) I'm a bit worried about this change: https://github.com/ravendb/ravendb/pull/9413/commits/7bb1c8345ab8fb2e8f1ac27859e60ff9af85af49 I don't consider it to be breaking change but because it changes the server side behavior let's think if that could affect someone's existing system. I mean that someone rely on the fact that expired docs are deleted on each node separately but this would be no longer the case.

talweiss1982:
@ayende I wouldn't be so worried about creating conflicts in the cluster since if the data is identical we don't create conflicts and I believe we already ignore some metadata fields when checking for conflicts so we could just add the '@refresh' field to the list of ignored fields

src/Raven.Server/Documents/Expiration/ExpirationStorage.cs
```diff
@@ -146,38 +159,90 @@ private void ThrowWrongExpirationDateFormat(Slice lowerId, string expirationDate
...
+                if (DateTime.TryParseExact(expirationDate, DefaultFormat.DateTimeFormatsToRead, CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind, out var date))
+                {
+                    if (currentTime >= date.ToUniversalTime())
+                        return true;
+                }
+            }
+            return false;
         }
 
         public int DeleteExpiredDocuments(DocumentsOperationContext context, Dictionary<Slice, List<(Slice LowerId, LazyStringValue Id)>> expired)
```
talweiss1982: The name here is now confusing since it is now 'Trigger' some actions and not just delete documents

src/Raven.Server/Documents/Expiration/ExpirationStorage.cs
```diff
@@ -146,38 +159,90 @@ private void ThrowWrongExpirationDateFormat(Slice lowerId, string expirationDate
...
+                                    {
+                                        _database.DocumentsStorage.Put(context, doc.Id, doc.ChangeVector, updated, flags: doc.Flags);
+                                    }
+                                }
+                            }
+                        }
+                        catch (DocumentConflictException)
+                        {
+                            if(GetConflictedExpiration(context, currentTime, ids.LowerId).AllExpired)
+                                _database.DocumentsStorage.Delete(context, ids.LowerId, ids.Id, expectedChangeVector: null);
```
talweiss1982, line 250: i'm not sure if this behavior is right if we are running on '@refresh' and get a conflict shouldn't we do a Put here rather than a delete?

src/Raven.Server/Documents/Expiration/ExpirationStorage.cs
arekpalinski: +1

src/Raven.Server/Documents/Expiration/ExpiredDocumentsCleaner.cs
```diff
@@ -90,9 +91,17 @@ internal async Task CleanupExpiredDocs()
 
                 using (_database.DocumentsStorage.ContextPool.AllocateOperationContext(out DocumentsOperationContext context))
                 {
+                    DatabaseTopology topology;
+                    using (_database.ServerStore.ContextPool.AllocateOperationContext(out TransactionOperationContext serverContext))
+                    {
+                        topology = _database.ServerStore.Cluster.ReadDatabaseTopology(serverContext, _database.Name);
+                    }
+
+                    var isFirstInTopology = string.Equals(topology.AllNodes.FirstOrDefault(), _database.ServerStore.NodeTag, StringComparison.OrdinalIgnoreCase);
```
ppekrol, line 137: I'm wondering if this should not run e.g. on last node. We are performing a lot of operations on first already, e.g. client API is connecting to first by default. 

Also I'm wondering if we should not check only Members?

ayende:
@arekpalinski 

1) We can add two separate flags / timers, one for expiration, one for refresh.
I don't think we want to have another implementation of this. 

2) I can't see how someone would even be aware if this is happening on all node or just some.
I _guess_ if they are working on disconnected system, but even so, that is both rare and not promised.

3) I don't want to promise specific time frame. Having it happen X amount of time after the refresh timer is the intent. It isn't meant to be on the spot accurate.

@talweiss1982 I'm actually concerned between conflicts from the node removing the flag and another node accepting updates to it.
The removal of the flag would be later, so we would cause a conflict there.

src/Raven.Server/Documents/Expiration/ExpiredDocumentsCleaner.cs
ayende: I'm using the first one intentionally, that is the one that we are likely to already be writing for, so the chance of creating a conflict with user writes is even lower.
And we need to take into account that we don't _have_ members, that all have a rehab, so we'll use that.

src/Raven.Server/Documents/Expiration/ExpirationStorage.cs
ayende: This can only happen if all the conflicts have expired, so we can auto resolve by delete.
Can't happen if there isn't such a scenario, so I don't think there is a problem here?

arekpalinski:
> 
>     1. We can add two separate flags / timers, one for expiration, one for refresh.
>        I don't think we want to have another implementation of this.

@ayende  I'm not saying we want to have separate implementation (server side). I'd like to make it a separate feature (from the end user point of view) not part of Expiration. So for example having a dedicated operations in client API, separate page in the studio etc.

ayende:
Oh, the UI should be  different, yes

arekpalinski:
Then don't you think it would make sense to have dedicated operations in Client API as well to manage / configure the @refresh feature?

BTW there are test failures related to Expiration

ayende:
Fixed tests and separated the client ops

src/Raven.Server/Documents/Expiration/ExpirationStorage.cs
```diff
@@ -188,5 +259,43 @@ public int DeleteExpiredDocuments(DocumentsOperationContext context, Dictionary<
...
+            foreach (var pair in expired)
+            {
+                foreach (var ids in pair.Value)
+                {
+                    if (ids.Id != null)
+                    {
+                        var doc = _database.DocumentsStorage.Get(context, ids.LowerId, throwOnConflict: false);
+                        if (doc != null && doc.TryGetMetadata(out var metadata))
+                        {
+                            if (HasExpired(metadata, Constants.Documents.Metadata.Refresh, currentTime))
```
arekpalinski: `HasExpired` ?

arekpalinski: It applies to the refresh feature here, right?

ayende: This checks the specified metadata property and check if it passed, yes.

arekpalinski: Then maybe rename to `HasPassed`? So it won't look it's specific for expiration feature

ppekrol (approved):
Pushed 1 small fix

PR #9414, opened 2019-06-20
RavenDB-13716 After server upgrade - version is not updated on Cluste…

karmeli87:
…r page

### DIFF
--- a/src/Raven.Studio/typescript/common/shell/clusterTopologyManager.ts
+++ b/src/Raven.Studio/typescript/common/shell/clusterTopologyManager.ts
@@ -51,7 +51,6 @@ class clusterTopologyManager {
         const serverWideClient = changesContext.default.serverNotifications();
 
         serverWideClient.watchClusterTopologyChanges(e => this.onTopologyUpdated(e));
-        serverWideClient.watchReconnect(() => this.fetchTopology());
     }
 
     private onTopologyUpdated(e: Raven.Server.NotificationCenter.Notifications.Server.ClusterTopologyChanged) {


src/Raven.Studio/typescript/common/shell/clusterTopologyManager.ts
```diff
@@ -51,7 +51,6 @@ class clusterTopologyManager {
         const serverWideClient = changesContext.default.serverNotifications();
 
         serverWideClient.watchClusterTopologyChanges(e => this.onTopologyUpdated(e));
-        serverWideClient.watchReconnect(() => this.fetchTopology());
```
grisha-kotler, line 54: Don't we need this when the studio reconnects to the WebSocket?

karmeli87: Tested this manually, seems to work. 
But @ml054 is the higher authority for this :-)

ml054: I wonder if it breaks something else. We refresh topology in few cases. 

Why can't we send valid information via web socket instead of not trusting web socket? 

@karmeli87 Could you test the following case:
1. studio is connected to node A and A is leader
2. kill A 
3. run A again
4. I wonder if data will be refreshed. 

Or any other scenario when topology is changed and node is not reachable for a moment (which cases WS to disconnect).

ml054: ok I talked with @karmeli87 and we don't need to fetch topology using endpoint, because we get current topology as first message after we connect to websocket. Moreover websocket gets information indirectly from leader, where endpoint does not.

ml054 (approved):

PR #9421, opened 2019-06-20
RavenDB-13741 - Missing Raven-Clr-Type in Export/Import CSV

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-13741

### DIFF
--- a/src/Raven.Server/Documents/Handlers/StreamCsvResultWriter.cs
+++ b/src/Raven.Server/Documents/Handlers/StreamCsvResultWriter.cs
@@ -21,6 +21,13 @@ namespace Raven.Server.Documents.Handlers
         private (string, string)[] _properties;
         private bool _writeHeader = true;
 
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Documents/Handlers/StreamCsvResultWriter.cs
```diff
@@ -21,6 +21,13 @@ public abstract class StreamCsvResultWriter<T> : IStreamQueryResultWriter<T>
         private (string, string)[] _properties;
         private bool _writeHeader = true;
 
+        private readonly HashSet<string> _metadataPropertiesToSkip = new HashSet<string>
```
ppekrol, line 24: @grisha-kotler can we create some sort of fail check e.g. a test that will help us to manage those in the future? Currently this is a very easy place to skip when new metadata properties are being added e.g. timeseries.

PR #9423, opened 2019-06-23
RavenDB-13744

talweiss1982:


talweiss1982:
Need to release acquired path cache items rather than clearing the lists

src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -1036,7 +1037,10 @@ protected internal virtual unsafe void Reset(bool forceReleaseLongLivedAllocator
 
                 _pooledArrays = null;
             }
-        }
+
+            PathCacheHolder[] _allocatePathCaches = new PathCacheHolder[512];
```
talweiss1982: this is some merge issue, will be modified anyway soon

ayende (approved):
@talweiss1982 This was a great catch.
Pretty sure that I was the one who both wrote the cache and decided to not mass Reset the context, and I wouldn't have caught it.

PR #9428, opened 2019-06-24
RavenDB-13744 prevent big memory streams to be held by the context ca…

talweiss1982:
…che also limit the size of the cache

### DIFF
--- a/src/Sparrow/Json/JsonOperationContext.cs
+++ b/src/Sparrow/Json/JsonOperationContext.cs
@@ -1226,15 +1226,20 @@ namespace Sparrow.Json
                 return new MemoryStream();
             }
 
-            return _cachedMemoryStreams.Pop();
+            var stream = _cachedMemoryStreams.Pop();
+            _sizeOfMemoryStreamCache -= stream.Capacity;
+
+            return stream;
         }
 
         private const long MemoryStreamCacheThreshold = Constants.Size.Megabyte;
-        private const int MemoryStreamCacheMaxCapacity = 512;
+        private const int MemoryStreamCacheMaxCapacityInBytes = 64 * Constants.Size.Megabyte;
+
+        private long _sizeOfMemoryStreamCache;
         public void ReturnMemoryStream(MemoryStream stream)
         {
             //We don't want to hold big streams in the cache or have too big of a cache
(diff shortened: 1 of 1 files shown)


src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -1229,9 +1229,18 @@ public MemoryStream CheckoutMemoryStream()
             return _cachedMemoryStreams.Pop();
         }
 
+        private const long MemoryStreamCacheThreshold = Constants.Size.Megabyte;
+        private const int MemoryStreamCacheMaxCapacity = 512;
         public void ReturnMemoryStream(MemoryStream stream)
         {
+            //We don't want to hold big streams in the cache or have too big of a cache
+            if (stream.Capacity > MemoryStreamCacheThreshold || _cachedMemoryStreams.Count >= MemoryStreamCacheMaxCapacity)
```
ayende: This still gives us a single context with 512MB of memory.
Let's limit this to 64 MB total?

talweiss1982: 64MB total and without capacity limit or 64MB instead of 512 items and still remove large streams?

src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -1226,14 +1226,29 @@ public MemoryStream CheckoutMemoryStream()
...
+        private const long MemoryStreamCacheThreshold = Constants.Size.Megabyte;
+        private const int MemoryStreamCacheMaxCapacityInBytes = 64 * Constants.Size.Megabyte;
+
+        private long _sizeOfMemoryStreamCache;
         public void ReturnMemoryStream(MemoryStream stream)
         {
+            //We don't want to hold big streams in the cache or have too big of a cache
+            if (stream.Capacity > MemoryStreamCacheThreshold || _sizeOfMemoryStreamCache >= MemoryStreamCacheMaxCapacityInBytes)
+            {
+                return;
```
ppekrol, line 1244: I think that we need to dispose the stream in this case?

maximburyak: Disposing a memory stream doesn't release any data, and we don't really care about flags like isOpen being closed

ppekrol: Ok

PR #9432, opened 2019-06-25
RavenDB-13746 Remove Replace License and Renew License when it's disa…

Danielle9897:
…bled in configuration

### DIFF
--- a/src/Raven.Server/Web/Studio/LicenseHandler.cs
+++ b/src/Raven.Server/Web/Studio/LicenseHandler.cs
@@ -1,11 +1,12 @@
 ﻿using System.Net;
 using System.Reflection;
 using System.Threading.Tasks;
-using Raven.Client.Documents.Operations.OngoingTasks;
 using Raven.Server.Commercial;
 using Raven.Server.Json;
 using Raven.Server.Routing;
+using Raven.Server.ServerWide.Context;
 using Sparrow.Json;
+using Sparrow.Json.Parsing;
 
 namespace Raven.Server.Web.Studio
 {
@@ -38,14 +39,32 @@ namespace Raven.Server.Web.Studio
             return NoContent();
         }
 
-
         [RavenAction("/license/status", "GET", AuthorizationStatus.ValidUser)]
         public Task Status()
         {
-            using (var context = JsonOperationContext.ShortTermSingleUse())
+            using (ServerStore.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
             using (var writer = new BlittableJsonTextWriter(context, ResponseBodyStream()))
             {
                 context.Write(writer, ServerStore.LicenseManager.GetLicenseStatus().ToJson());
+            }
+
+            return Task.CompletedTask;
+        }
+        
+        [RavenAction("/license/configuration", "GET", AuthorizationStatus.ValidUser)]
+        public Task GetLicenseConfigurationSettings()
+        {
+            var licenseConfiguration = new LicenseConfigurationSettings
+            {
+                RenewSettings = ServerStore.Configuration.Licensing.CanRenewLicense,
+                ActivateSettings = ServerStore.Configuration.Licensing.CanActivate,
+                ForceUpdateSettings = ServerStore.Configuration.Licensing.CanForceUpdate
+            };
+
+            using (ServerStore.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
+            using (var writer = new BlittableJsonTextWriter(context, ResponseBodyStream()))
+            {
+                context.Write(writer, licenseConfiguration.ToJson());
             }
 
             return Task.CompletedTask;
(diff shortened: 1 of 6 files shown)


src/Raven.Studio/typescript/viewmodels/shell/about.ts
```diff
@@ -183,11 +183,25 @@ class about extends viewModelBase {
            return "icon-cancel";
         });
     }
+    
+    renewLicenseConfigurationSettings = ko.observable<boolean>(false);
```
ml054: those names are misleading. It isn't settings, but flag if user can renew license, right?
I'd suggest name: canRenewLicense, canActivateLicense, etc.

src/Raven.Server/Web/Studio/LicenseHandler.cs
```diff
@@ -106,12 +125,29 @@ public async Task RenewLicense()
...
             {
                 var renewLicense = await ServerStore.LicenseManager.RenewLicense();
                 context.Write(writer, renewLicense.ToJson());
             }
         }
     }
+    
+    public class LicenseConfigurationSettings : IDynamicJson
+    {
+        public bool RenewSettings { get; set; }
```
ml054: names in LicenseConfigurationSettings are misleading,

src/Raven.Studio/typescript/viewmodels/shell/about.ts
```diff
@@ -183,11 +183,25 @@ class about extends viewModelBase {
...
+    activateLicenseConfigurationSettings = ko.observable<boolean>(false);
+    forceLicenseUpdateConfigurationSettings = ko.observable<boolean>(false);
 
     register() {
         registration.showRegistrationDialog(license.licenseStatus(), false, true);
     }
 
+    private getLicenseConfigurationSettings() {
+        new getLicenseConfigurationSettingsCommand()
```
ml054: you don't return this promise, but you await for that in $.when

src/Raven.Studio/wwwroot/App/views/shell/about.html
```diff
@@ -139,22 +139,33 @@ <h4>Security</h4>
             <div class="padding">
                 <div class="license-actions flex-horizontal">
                     <!-- ko if: !registered() -->
-                    <button class="btn btn-primary flex-grow" data-bind="click: register" title="Register a new license">REGISTER LICENSE</button>
+                    <button class="btn btn-primary flex-grow" 
+                            data-bind="click: register, 
+                                       enable: activateLicenseConfigurationSettings,
+                                       attr: { title: activateLicenseConfigurationSettings() ? 'Register a new license' : 'Registering new license is disabled in the server configuration' }">
```
ml054: please make sure title works when button is disabled

ml054 (changes requested):

src/Raven.Studio/typescript/viewmodels/shell/about.ts
Danielle9897: Well - no - it is not representing a can/can't flag 
The flags that are responsible for that are in:
https://github.com/ravendb/ravendb/blob/726ce180fc0a7f102d63ddab55f048dfb5e942f5/src/Raven.Studio/typescript/common/shell/accessManager.ts#L32

These new params represent the content of the settings.json file, 
or whatever is set with the env variables, or command line args
i.e.  "License.CanRenewLicense": false

src/Raven.Server/Web/Studio/LicenseHandler.cs
Danielle9897: same answer as above

src/Raven.Studio/wwwroot/App/views/shell/about.html
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/shell/about.ts
Danielle9897: Added 'return'

src/Raven.Server/Web/Studio/LicenseHandler.cs
ppekrol: I concur with Marcin, I think it would be better to match the configuration properties e.g. CanActivate, CanRenew (rename the property and add a 2nd key), etc

ml054: like I wrote: when I see property with name: SomethingCount, I expect number. When I see SomethingList I expect array/list. But when I see SomethingSettings then boolean type is unexpected.

Danielle9897: Changed the endpoint to use the original `LicenseConfiguration ` class

src/Raven.Studio/typescript/viewmodels/shell/about.ts
Danielle9897: renamed to:
```
isActivateLicenseEnabled
isRenewLicenseEnabled
isForceUpdateEnabled
```

ppekrol (approved):

PR #9351, opened 2019-06-05
RavenDB-13650: Add Select & OrderBy overloads for IDictionary, add ca…

garayx:
…sts to ToDictionary and Distinct in index expression, use DynamicInvocationExpressionsRewriter in reduce.

PR #9434, opened 2019-06-25
RavenDB-13650: Add DynamicDictionary

garayx:


src/Raven.Server/Documents/Indexes/Static/DynamicDictionary.cs
```diff
@@ -0,0 +1,593 @@
...
+        private readonly Dictionary<object, object> _dictionary;
+
+        public DynamicDictionary(Dictionary<object, object> dictionary)
+        {
+            _dictionary = dictionary;
+        }
+
+        public DynamicDictionary(IEnumerable<KeyValuePair<object, object>> dictionary)
+        {
+            _dictionary = (Dictionary<object, object>)dictionary;
```
ayende: You need to construct a new dictionary here, you can't assume that this is a Dictionary

garayx: done

src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/LuceneDocumentConverterBase.cs
```diff
@@ -315,6 +317,14 @@ int HandleArray(IEnumerable itemsToIndex)
                 return newFields;
             }
 
+            if (valueType == ValueType.DynamicDictionary)
+            {
+                instance.Add(GetOrCreateField(path + IsDictionaryFieldSuffix, TrueString, null, null, storage, Field.Index.NOT_ANALYZED_NO_NORMS, Field.TermVector.NO));
```
arekpalinski: I wonder if we want / have to introduce `_IsDictionary` suffix? What is the benefit?

I see we're using to set `FieldType.IsDictionary` but I don't see _any_ read usage of it

src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/LuceneDocumentConverterBase.cs
```diff
@@ -315,6 +317,14 @@ int HandleArray(IEnumerable itemsToIndex)
...
+                newFields++;
+
+                valueType = ValueType.ConvertToJson;
```
arekpalinski: So we're intentionally changing it to `ValueType.ConvertToJson` ?

arekpalinski: I really have some doubts if we need to introduce any new concept to LuceneDocumentCoverter? Since we're indexing it same as we did it - `ConvertToJson`

test/SlowTests/MailingList/IndexCompilation.cs
```diff
@@ -222,7 +224,7 @@ public void CanCompileIndexWithDistinct()
                             SelectSum = listOfUsers[i].LoginCountByDate.ToDictionary(y => y.Key, y => y.Value).OrderBy(x => x.Value).Select(x => x.Value).Sum(x => x),
                             OrderBySum = listOfUsers[i].LoginCountByDate.ToDictionary(y => y.Key, y => y.Value).Select(x => x.Value).Sum(x => x),
                             IdsWithDecimals = listOfUsers[i].ListOfDecimals.ToDictionary(k => k, k => 0),
-                            OrderByDescending = listOfUsers[i].LoginCountByDate.OrderByDescending(x => x.Value).ToDictionary(y => y.Key.ToString(CultureInfo.InvariantCulture), y => y.Value),
+                            OrderByDescending = listOfUsers[i].LoginCountByDate.OrderByDescending(x => x.Value).ToDictionary(y => y.Key.GetDefaultRavenFormat(isUtc: y.Key.Kind == DateTimeKind.Utc), y => y.Value),
```
arekpalinski, line 227: Why you need to change it? Would it be a breaking change?

src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/LuceneDocumentConverterBase.cs
garayx: Yes, it is the same as it was.
I introduced _ValueType.DynamicDictionary_ just to get the ability to add __IsDictionary_ suffix, I tried to follow the same logic we have with Array (we adding this field there also). Anyway I can easily remove it, if needed

src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/LuceneDocumentConverterBase.cs
garayx: followed the same logic as array.

src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/LuceneDocumentConverterBase.cs
arekpalinski: I'm think sure we don't need to change anything in `LuceneDocumentConverterBase` because from the Lucene point of view it doesn't matter. We're just storing a dictionary as JSON and that's it - so same behavior as we had.

@ppekrol thoughts?

arekpalinski: I've talked to Pawel. Given that we don't have any requirement to introduce _IsDictionary at this let's drop it.

test/SlowTests/MailingList/IndexCompilation.cs
garayx: this is how we store dicitonary key if it is DateTime

arekpalinski: Does it mean that if someone has the index definition as in this test (with the usage of `y.Key.ToString(CultureInfo.InvariantCulture)` then his index will stop working after merging this PR? That would be breaking change - that's my concern

garayx: If he has a DateTime dicitonary key then his index doesnt work now, because in document we store it using GetDefaultRavenFormat
and in dictionary we store it just as ToString
https://github.com/ravendb/ravendb/blob/v4.2/src/Raven.Server/Utils/TypeConverter.cs#L243

arekpalinski: Okay, I see this test was added by you in your previous PR related to the same issue (https://github.com/ravendb/ravendb/pull/9351/files#diff-2fb714770e61aff7a2f4afd3c71eb209R225)

That's fine then. Just wanted to double check that.

src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/LuceneDocumentConverterBase.cs
garayx: done

arekpalinski: "done"? - did you forget to push? :)

ayende (approved):

PR #9435, opened 2019-06-25
RavenDB-13738 Have a better error why 4.2 can't talk with 4.1.4

karmeli87:


src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -1112,6 +1116,15 @@ public void AcceptNewConnection(Stream stream, Action disconnect, EndPoint remot
             }
         }
 
+        private static void ValidateCompatibility(RachisHello initialMessage)
+        {
+            if (ServerVersion.IsNightlyOrDev(initialMessage.ServerBuildVersion) == false &&
+                initialMessage.ServerBuildVersion <= 41010)
```
ppekrol: Old versions will not send anything so you will get 0. What will happen then?
Also 4.2.0 will not send anything. Actually nodes <= 41010 are not sending anything so I'm not sure if there is a point of having this at the moment? Can be used in future, but not know?

karmeli87: you are correct, at all points :)

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -1112,6 +1116,13 @@ public void AcceptNewConnection(Stream stream, Action disconnect, EndPoint remot
             }
         }
 
+        private static void ValidateCompatibility(RachisHello initialMessage)
+        {
+            var version = initialMessage.ServerBuildVersion;
+            if (version == 0 || ServerVersion.IsNightlyOrDev(version))  
+                return;
+        }
```
ayende, line 1124: What is the point of this? Preparing for the future?

ppekrol: Yes, correct. In theory we can remove this, but maybe it is better to have this extension point prepared.

ayende (approved):

PR #9437, opened 2019-06-25
RavenDB-13437: Add search overload that receive array of searchTerm v…

garayx:
…alues

### DIFF
--- a/src/Raven.Client/Documents/LinqExtensions.cs
+++ b/src/Raven.Client/Documents/LinqExtensions.cs
@@ -873,6 +873,28 @@ namespace Raven.Client.Documents
             return (IRavenQueryable<T>)queryable;
         }
 
+        /// <summary>        
+        /// Perform a search for documents which fields match the searchTerms array.
+        /// Each array element will be checked independently.        
+        /// </summary>
+        /// <typeparam name="T">The type of element of self</typeparam>
+        /// <param name="self">The <see cref="IQueryable{T}"/> to search on</param>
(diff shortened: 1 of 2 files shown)


src/Raven.Client/Documents/LinqExtensions.cs
```diff
@@ -873,6 +873,28 @@ public static Task<TSource> SingleOrDefaultAsync<TSource>(this IQueryable<TSourc
...
+        /// Each array element will be checked independently.        
+        /// </summary>
+        /// <typeparam name="T">The type of element of self</typeparam>
+        /// <param name="self">The <see cref="IQueryable{T}"/> to search on</param>
+        /// <param name="fieldSelector">Function returning the field to search on</param>
+        /// <param name="searchTerms">Array of terms to search for.</param>
+        /// <param name="boost">Boost factor for sorting purposes</param>
+        /// <param name="options">Logical operator to use in relation to the previous filtering statement.</param>
+        /// <param name="operator">Determines the logical operator between all of the terms received in searchTerms parameter</param>
+        public static IRavenQueryable<T> Search<T>(this IQueryable<T> self, Expression<Func<T, object>> fieldSelector, string[] searchTerms,
```
ppekrol: I think it would be more flexible for the users if we will have `IEnumerable<string>` instead of an array

garayx: done

ppekrol (approved):

PR #9440, opened 2019-06-26
Raven db 13761 Subscriptions on delete revisions should return (current, null) tuple

maximburyak:


maximburyak:
Failing tests seems to be related to https://issues.hibernatingrhinos.com/issue/RavenDB-13754 I don't think that this should block merging

PR #9447, opened 2019-06-27
RavenDB-10956 support a command to remove CV entries from a specific database

karmeli87:


src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -357,9 +358,60 @@ public static string GetDatabaseChangeVector(Transaction tx)
...
+                return list;
+            }
+        }
+
+        public bool TryRemoveUnusedIds(ref string changeVector)
+        {
+            if (string.IsNullOrEmpty(changeVector))
+                return false;
+
+            var list = GetUnusedList();
```
talweiss1982: This seems to be really expensive opening a read transaction each time we want to update a change vector.
Maybe we can hold an in-memory list that we will modify in a volatile manner if we need to change it.
The persistent state will be written and read on startup.

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -1136,6 +1136,8 @@ private void NotifyFeaturesAboutStateChange(DatabaseRecord record, long index)
                     try
                     {
                         DatabaseGroupId = record.Topology.DatabaseTopologyIdBase64;
+                        Interlocked.Exchange(ref DocumentsStorage.UnusedDatabaseIds, record.UnusedDatabaseIds);
```
ayende: Let's check if the values are the same before replacing them.
I'm worried about this getting to Gen2, then being replaced, having the new val in Gen2, etc.

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -357,9 +358,41 @@ public static string GetDatabaseChangeVector(Transaction tx)
...
+            var list = UnusedDatabaseIds;
+            if (list == null || list.Count == 0)
+                return false;
+
+            var parsed = changeVector.ToChangeVectorList();
+            var removed = false;
+            for (var i = parsed.Count - 1; i >= 0; i--)
+            {
+                var entry = parsed[i];
+                if (list.Contains(entry.DbId))
```
ayende, line 377: If we have a value here, we'll look it up a LOT.
Let's make this into a `HashSet`?

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -357,9 +358,41 @@ public static string GetDatabaseChangeVector(Transaction tx)
...
+                {
+                    list.Remove(entry.DbId);
```
ayende: Why are you removing from the list? You are mutating the global `UnusedDatabaseIds` here, no?

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -357,9 +358,41 @@ public static string GetDatabaseChangeVector(Transaction tx)
...
+                    parsed.Remove(entry);
```
ayende: Use `RemoveAt` ? To avoid scanning through it

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -1165,6 +1165,24 @@ CrontabSchedule VerifyBackupFrequency(string backupFrequency)
             }
         }
 
+        [RavenAction("/admin/unused-databases", "POST", AuthorizationStatus.DatabaseAdmin)]
```
ayende: This doesn't look right.
This is a cluster wide operation, not a database operation.

ayende (changes requested):

src/Raven.Server/Documents/DocumentDatabase.cs
karmeli87: done

src/Raven.Server/Documents/DocumentsStorage.cs
karmeli87: done

src/Raven.Server/Documents/DocumentsStorage.cs
karmeli87: done

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
karmeli87: It is a cluster wide operation, but will affect a single database.

changed to `Operator`.

src/Raven.Server/ServerWide/Commands/UpdateUnusedDatabaseIdsCommand.cs
```diff
@@ -0,0 +1,33 @@
...
+            
+        }
+        public UpdateUnusedDatabaseIdsCommand(string database, HashSet<string> list, string raftId) : base(database, raftId)
+        {
+            UnusedDatabaseIds = list;
+        }
+
+        public override string UpdateDatabaseRecord(DatabaseRecord record, long etag)
+        {
+            record.UnusedDatabaseIds = UnusedDatabaseIds;
```
maximburyak, line 24: Maybe we should make it more atomic operation, like adding/removing database ids

ayende: We are very unlikely to have to do this very often, I don't think it matters

ayende (approved):

src/Raven.Client/ServerWide/Operations/UpdateUnusedDatabasesOperation.cs
```diff
@@ -0,0 +1,70 @@
...
+                    Content = new BlittableJsonContent(stream =>
+                    {
+                        ctx.Write(stream, EntityToBlittable.ConvertCommandToBlittable(_parameters, ctx));
+                    })
+                };
+            }
+
+            public string RaftUniqueRequestId => RaftIdGenerator.NewId();
+        }
+        public class Parameters
```
ppekrol: can be private, but need to put in inside the command

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -357,9 +358,40 @@ public static string GetDatabaseChangeVector(Transaction tx)
             return Encodings.Utf8.GetString(val.Reader.Base, val.Reader.Length);
         }
 
+        internal HashSet<string> UnusedDatabaseIds;
```
ppekrol, line 361: What kind of comparer is set here? Does it have to be case insensitive one?

src/Raven.Server/ServerWide/ClusterCommandsVersionManager.cs
```diff
@@ -105,7 +105,8 @@ public static class ClusterCommandsVersionManager
             [nameof(PutSubscriptionBatchCommand)] = Base42CommandsVersion,
             [nameof(EditDatabaseClientConfigurationCommand)] = Base42CommandsVersion,
             [nameof(PutServerWideBackupConfigurationCommand)] = 42_001,
-            [nameof(DeleteServerWideBackupConfigurationCommand)] = 42_001
+            [nameof(DeleteServerWideBackupConfigurationCommand)] = 42_001,
+            [nameof(UpdateUnusedDatabaseIdsCommand)] = 42_001
```
ppekrol: Shouldn't this be higher?

src/Raven.Server/ServerWide/Commands/RemoveNodeFromDatabaseCommand.cs
```diff
@@ -1,24 +1,31 @@
...
         }
 
         public override string UpdateDatabaseRecord(DatabaseRecord record, long etag)
         {
             record.Topology.RemoveFromTopology(NodeTag);
             record.DeletionInProgress?.Remove(NodeTag);
+            if (record.UnusedDatabaseIds == null)
+                record.UnusedDatabaseIds = new HashSet<string>();
+
+            record.UnusedDatabaseIds.Add(DatabaseId);
```
ppekrol, line 32: DatabaseId will be null in 'old' commands, right?

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -1165,6 +1165,24 @@ CrontabSchedule VerifyBackupFrequency(string backupFrequency)
             }
         }
 
+        [RavenAction("/admin/unused-databases", "POST", AuthorizationStatus.Operator)]
```
ppekrol: Can we change that to `/admin/databases/unused-ids`? Will match other endpoints here

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -1165,6 +1165,24 @@ CrontabSchedule VerifyBackupFrequency(string backupFrequency)
...
+        public async Task SetUnusedDatabaseIds()
+        {
+            var database = GetStringQueryString("database");
```
ppekrol: change to `name` to match other endpoints

src/Raven.Client/ServerWide/Operations/UpdateUnusedDatabasesOperation.cs
karmeli87: Can't be private. It is used also in `JsonDeserializationServer`

src/Raven.Server/Documents/DocumentsStorage.cs
karmeli87: it is base64, so it has to be case sensetive

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
karmeli87: done

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
karmeli87: done

src/Raven.Server/ServerWide/Commands/RemoveNodeFromDatabaseCommand.cs
karmeli87: right! done

src/Raven.Server/ServerWide/ClusterCommandsVersionManager.cs
karmeli87: done

src/Raven.Client/ServerWide/Operations/UpdateUnusedDatabasesOperation.cs
ppekrol: Then let's move it inside UpdateUnusedDatabasesCommand  and make it internal

ppekrol (approved):

PR #9453, opened 2019-06-29
RavenDB-12432

ppekrol:


src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1048,35 +1048,67 @@ public long GetLastTombstoneEtag(DocumentsOperationContext context, string colle
...
 #endif
             return document;
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private static Document ParseDocument(JsonOperationContext context, ref TableValueReader tvr)
+        private static Document ParseDocument(JsonOperationContext context, ref TableValueReader tvr, DocumentFields fields)
         {
-            var result = new Document
+            if (fields == DocumentFields.All)
```
ayende, line 1068: This is going to be the most common action, right?
The problem is that now this method is pretty big, and full of conditional.
Let's move all the checks after this if to a dedicated method.

ppekrol: Done

ayende (approved):

PR #9459, opened 2019-07-02
HRINT-1111 Sql test modified to work with remote sql server (mysql, mssql, oracle & npgsql)

haludi:
https://issues.hibernatingrhinos.com/issue/HRINT-1111


test/Tests.Infrastructure/ConnectionString/MssqlConnectionString.cs
```diff
@@ -0,0 +1,54 @@
...
+{
+    public class MssqlConnectionString: SqlConnectionString<SqlConnection>
+    {
+        public static readonly MssqlConnectionString Instance = new MssqlConnectionString();
+
+        private MssqlConnectionString(): base("RAVEN_MSSQL_CONNECTION_STRING")
+        {
+        }
+
+        protected override string VerifiedConnectionStringFactor()
```
haludi, line 16: Maybe the overriding is unnecessary? If we want to use just the mssql from the remote VM?

test/SlowTests/Server/Documents/SqlMigration/SqlAwareTestBase.cs
```diff
@@ -207,30 +208,37 @@ private DisposableAction WithMsSqlDatabase(out string connectionString, out stri
 
         protected DisposableAction WithMySqlDatabase(out string connectionString, out string databaseName, string dataSet, bool includeData = true)
         {
+            const int commandTimeout = 2 * 60;
```
haludi: I checked it and because of all the SQL database run on the same machine, the requests can take time and lead to timeout exception. I set it to two minutes to be sure it has its time

PR #9462, opened 2019-07-03
RavenDB-13776 Have a good error explaining that adding a *new* 4.2 node to 4.1 cluster isn't supported

karmeli87:


ayende (approved):

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -1116,11 +1117,18 @@ public void AcceptNewConnection(Stream stream, Action disconnect, EndPoint remot
...
+        private void ValidateCompatibility(RachisHello initialMessage, ClusterTopology clusterTopology)
         {
             var version = initialMessage.ServerBuildVersion;
-            if (version == 0 || ServerVersion.IsNightlyOrDev(version))  
+            if (ServerVersion.IsNightlyOrDev(version))
                 return;
+
+            if (clusterTopology.Members.ContainsKey(Tag) == false)
+            {
+                if (version == 0 || (version >= 40_000 && version < 42_000))
```
karmeli87: The problem with version 0 is that we can't distinguish between 4.1.x and 4.2.0 :/

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -1116,11 +1117,18 @@ public void AcceptNewConnection(Stream stream, Action disconnect, EndPoint remot
...
+        private void ValidateCompatibility(RachisHello initialMessage, ClusterTopology clusterTopology)
         {
             var version = initialMessage.ServerBuildVersion;
-            if (version == 0 || ServerVersion.IsNightlyOrDev(version))  
+            if (ServerVersion.IsNightlyOrDev(version))
                 return;
+
+            if (clusterTopology.Members.ContainsKey(Tag) == false)
+            {
+                if (version >= 40_000 && version < 42_000)
```
ppekrol, line 1128: Since we cannot distinguish between 4.1 and 4.2.0 we will allow then, but all new 4.1.x builds will be blocked.

ppekrol (approved):

PR #9466, opened 2019-07-04
RDBS-28 Old node tag should be kept as part of `GlobalStateSlice`

karmeli87:


src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -462,6 +464,14 @@ public string ReadNodeTag(TransactionOperationContext context)
             return readResult == null ? InitialTag : readResult.Reader.ToStringValue();
         }
 
+        public string ReadPreviousNodeTag(TransactionOperationContext context)
```
ayende, line 467: Can a server have _multiple_ old tags?

karmeli87: You can call `HardResetToNewCluster` multiple times with different tag every time.

ayende: Wouldn't that put us in the same situation?

karmeli87: Shouldn't, because on every tag update we keep also the old one.

ayende: ok

ayende (approved):

PR #9475, opened 2019-07-08
RavenDB-13760 Sodium & Pal modified to catch all exception on dll load and return detailed message

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-13760

### DIFF
--- a/src/Sparrow.Server/Platform/Pal.cs
+++ b/src/Sparrow.Server/Platform/Pal.cs
@@ -80,17 +80,17 @@ namespace Sparrow.Server.Platform
 
                 rc = rvn_get_system_information(out SysInfo, out errorCode);
             }
-            catch (IncorrectDllException)
-            {
-                throw;
-            }
             catch (Exception ex)
             {
-                var ErrString =
-                    "'Microsoft Visual C++ 2015 Redistributable Package' (or newer). It can be downloaded from https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads";
-                throw new IncorrectDllException(
-                    $"{LIBRVNPAL} version might be invalid or not usable on current platform. Initialization error could also be caused by missing {ErrString}",
-                    ex);
+                var platformStr = $"Arch:{RuntimeInformation.OSArchitecture}, OSDesc:{RuntimeInformation.OSDescription}";
+
+                var errString = $"{LIBRVNPAL} version might be invalid, missing or not usable on current platform '${platformStr}";
+
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                    errString += " Initialization error could also be caused by missing 'Microsoft Visual C++ 2015 Redistributable Package' (or newer). " +
+                                 "It can be downloaded from https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads";
+
+                throw new IncorrectDllException(errString, ex);
             }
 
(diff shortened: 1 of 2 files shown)


ayende (approved):

src/Sparrow.Server/Platform/Pal.cs
```diff
@@ -80,17 +80,17 @@ static Pal()
...
-                throw;
-            }
             catch (Exception ex)
             {
-                var ErrString =
-                    "'Microsoft Visual C++ 2015 Redistributable Package' (or newer). It can be downloaded from https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads";
-                throw new IncorrectDllException(
-                    $"{LIBRVNPAL} version might be invalid or not usable on current platform. Initialization error could also be caused by missing {ErrString}",
-                    ex);
+                var platformStr = $"Arch:{RuntimeInformation.OSArchitecture}, OSDesc:{RuntimeInformation.OSDescription}";
```
ppekrol: Missing space after `Arch:` and `OSDesc:`. Do we really need that? It produces longer exception method and the assumption here is what? That the user does not know what OS he is running on? Same info is available in welcome message.

ppekrol: I would suggest to remove it so the message would be more clear.

src/Sparrow.Server/Platform/Pal.cs
```diff
@@ -80,17 +80,17 @@ static Pal()
...
+
+                var errString = $"{LIBRVNPAL} version might be invalid, missing or not usable on current platform '${platformStr}";
```
ppekrol: Should be `current platform (${platformStr}).`

src/Sparrow.Server/Sodium.cs
```diff
@@ -67,32 +68,19 @@ static Sodium()
...
-            catch (DllNotFoundException dllNotFoundEx)
+            catch (Exception ex)
             {
-                // make sure the lib file is not there (this exception might pop when incorrect libsodium lib is does exists)
-                if (File.Exists(LIBSODIUM))
-                {
-                    throw new IncorrectDllException(
-                        $"{LIBSODIUM} probably contains the wrong version or not usable on the current platform. Try installing {LIBSODIUM} and re-run server.",
-                        dllNotFoundEx);
+                var platformStr = $"Arch:{RuntimeInformation.OSArchitecture}, OSDesc:{RuntimeInformation.OSDescription}";
```
ppekrol: Same comments as in Pal.

src/Sparrow.Server/Sodium.cs
```diff
@@ -67,32 +68,19 @@ static Sodium()
...
 
-                }
-            }
+                var errString = $"{LIBSODIUM} version might be invalid, missing or not usable on current platform '${platformStr}";
```
ppekrol: Same comments as in Pal.

src/Sparrow.Server/Platform/Pal.cs
haludi: This is for us so if we get this message from a client we don't need to do a ping pong with the client for this information

ppekrol: So it should be at the end of the message. It hides the main part.

haludi: O.k.

PR #9486, opened 2019-07-10
RavenDB-13655

talweiss1982:
* Refactored GetNewServer so it will be disposed of if not in a using statement (at the test dispose)
* fix to byte[] leak
* decoupling NativeMemoryCleaner from context and nullifying handlers in env options
* fixed javascript index leak from the index compilation cache
* more minor nullification fixes to cutdown gcroot issues

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Static/IndexCompilationCache.cs
+++ b/src/Raven.Server/Documents/Indexes/Static/IndexCompilationCache.cs
@@ -50,7 +50,12 @@ namespace Raven.Server.Documents.Indexes.Static
 
             try
             {
-                return result.Value;
+                var index = result.Value;
+                if (index is JavaScriptIndex javaScriptIndex)
+                {
+                    javaScriptIndex.TryRemoveFromCache = () => { IndexCache.TryRemove(key, out _); };
+                }
+                return index;
             }
             catch (Exception)
             {
--- a/src/Voron/Data/BTrees/Tree.Stream.cs
+++ b/src/Voron/Data/BTrees/Tree.Stream.cs
@@ -87,55 +87,59 @@ namespace Voron.Data.BTrees
             public void Write(Stream stream)
             {
                 _localBuffer = ArrayPool<byte>.Shared.Rent(512 * Constants.Size.Kilobyte);
-
-                AllocateNextPage();
-              
-                ((StreamPageHeader*)_currentPage.Pointer)->StreamPageFlags |= StreamPageFlags.First;
-                
-                fixed (byte* pBuffer = _localBuffer)
+                try
                 {
-                    while (true)
-                    {
-                        var read = stream.Read(_localBuffer, 0, _localBuffer.Length);
-                        if (read == 0)
-                            break;
+                    AllocateNextPage();
 
-                        var toWrite = 0L;
+                    ((StreamPageHeader*)_currentPage.Pointer)->StreamPageFlags |= StreamPageFlags.First;
+
+                    fixed (byte* pBuffer = _localBuffer)
+                    {
                         while (true)
                         {
-                            toWrite += WriteBufferToPage(pBuffer + toWrite, read - toWrite);
-                            if (toWrite == read)
+                            var read = stream.Read(_localBuffer, 0, _localBuffer.Length);
+                            if (read == 0)
                                 break;
 
-                            // run out of room, need to allocate more
-                            RecordChunkPage(_currentPage.PageNumber,  (int)(_writePos - _currentPage.DataPointer));
-                            AllocateNextPage();
+                            var toWrite = 0L;
+                            while (true)
+                            {
+                                toWrite += WriteBufferToPage(pBuffer + toWrite, read - toWrite);
+                                if (toWrite == read)
+                                    break;
+
+                                // run out of room, need to allocate more
+                                RecordChunkPage(_currentPage.PageNumber, (int)(_writePos - _currentPage.DataPointer));
+                                AllocateNextPage();
+                            }
                         }
-                    }
 
-                    var chunkSize = (int)(_writePos - _currentPage.DataPointer);
-                    RecordChunkPage(_currentPage.PageNumber, chunkSize);
-
-                    var remaining = _writePosEnd - _writePos;
-                    var infoSize = StreamInfo.SizeOf;
-
-                    if (_tag != null)
-                        infoSize += _tag.Value.Size;
-
-                    if (remaining < infoSize)
-                    {
-                        _numberOfPagesPerChunk = 1;
-                        AllocateNextPage();
-                        chunkSize = 0;
+                        var chunkSize = (int)(_writePos - _currentPage.DataPointer);
                         RecordChunkPage(_currentPage.PageNumber, chunkSize);
+
+                        var remaining = _writePosEnd - _writePos;
+                        var infoSize = StreamInfo.SizeOf;
+
+                        if (_tag != null)
+                            infoSize += _tag.Value.Size;
+
+                        if (remaining < infoSize)
+                        {
+                            _numberOfPagesPerChunk = 1;
+                            AllocateNextPage();
+                            chunkSize = 0;
+                            RecordChunkPage(_currentPage.PageNumber, chunkSize);
+                        }
+
+                        RecordStreamInfo();
+
+                        _parent._tx.LowLevelTransaction.ShrinkOverflowPage(_currentPage.PageNumber, chunkSize + infoSize, _parent.State);
                     }
-
-                    RecordStreamInfo();
-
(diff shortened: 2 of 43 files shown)


src/Raven.Server/Documents/Indexes/Static/IndexCompilationCache.cs
```diff
@@ -50,7 +50,12 @@ public static StaticIndexBase GetIndexInstance(IndexDefinition definition, Raven
 
             try
             {
-                return result.Value;
+                var index = result.Value;
```
maximburyak, line 53: I wonder if it would be better to just drop the js index caching altogether, if we are having such a hard time finding that leak.

src/Voron/Data/BTrees/Tree.Stream.cs
```diff
@@ -87,55 +87,59 @@ public void Init(Tree parent, Slice key, Slice? tag, int? initialNumberOfPagesPe
...
 
-                    _parent._tx.LowLevelTransaction.ShrinkOverflowPage(_currentPage.PageNumber, chunkSize + infoSize, _parent.State); 
+                        _parent._tx.LowLevelTransaction.ShrinkOverflowPage(_currentPage.PageNumber, chunkSize + infoSize, _parent.State);
+                    }
                 }
-
-                ArrayPool<byte>.Shared.Return(_localBuffer);
+                finally
+                {
+                    ArrayPool<byte>.Shared.Return(_localBuffer);
```
maximburyak, line 141: Nice

src/Voron/StorageEnvironment.cs
```diff
@@ -469,6 +469,8 @@ public void Dispose()
 
                 OnLogsApplied = null;
 
+                _options.NullifyHandlers();
```
maximburyak, line 472: Is this call critical here, or is it just a precaution? You did add it to the dispose as well. There are some cases (mostly tests, but also recovery) when we receive the options, and then temporarily change the OwnsPagers just to prevent disposal, but we always clear the option's events, so I'm not sure if something could get wrong then. 
Also, would you recommend a code reactoring task that will see ways to remove the OwnsPagers mechanism, I think it's more confusing the helping right now/

test/Tests.Infrastructure/TestBase.cs
```diff
@@ -341,53 +341,160 @@ public void UseNewLocalServer(IDictionary<string, string> customSettings = null,
...
+
+            public ServerCreationOptions(bool frozen = false)
+            {
+                _frozen = frozen;
+            }
+            private static readonly Lazy<ServerCreationOptions> _default = new Lazy<ServerCreationOptions>(() => new ServerCreationOptions(frozen:true));
+            public static ServerCreationOptions Default => _default.Value;
+        }
+
+        protected virtual RavenServer GetNewServer(ServerCreationOptions options = null)
```
maximburyak, line 446: Theoretically RegisterForDisposal, could have been yet another parameter in the function, but they just pile up, so I do like the refactoring of the way we pass those params, it ends up being more readable for the test reader.

src/Voron/StorageEnvironmentOptions.cs
```diff
@@ -1106,6 +1109,14 @@ public unsafe void Dispose()
             Disposing();
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
ayende: This isn't required. We use this on small methods that we need to be high perfomant. 
The first is true, but the second isn't.

ayende (approved):

src/Raven.Server/Documents/Indexes/Static/IndexCompilationCache.cs
```diff
@@ -50,7 +50,12 @@ public static StaticIndexBase GetIndexInstance(IndexDefinition definition, Raven
...
+                if (index is JavaScriptIndex javaScriptIndex)
+                {
+                    javaScriptIndex.TryRemoveFromCache = () => { IndexCache.TryRemove(key, out _); };
```
ppekrol, line 56: IndexCache is static, what if I will have 2 identical indexes e.g. in two separate databases with sample data?

talweiss1982: https://github.com/ravendb/ravendb/blob/v4.2/src/Raven.Server/Documents/Indexes/Static/IndexCompilationCache.cs#L33

src/Voron/StorageEnvironmentOptions.cs
talweiss1982: does it hurt us though?

src/Raven.Server/Documents/Indexes/Static/IndexCompilationCache.cs
ppekrol: Ok, didn't know that

src/Raven.Server/Documents/Indexes/Static/IndexCompilationCache.cs
talweiss1982: The cache is used for reusing the same instance of an index and this is why it is in the cache other than that the index is unique since it has a state unlike the c# ones

src/Voron/StorageEnvironment.cs
talweiss1982: To be honest I find this confusing as well. I haven't dug too deep into this just saw that those handlers are holding quite a bunch of resources with them and also notice we didn't dispose of the options even though they are disposables.
When and if you should dispose of them is unclear to me but for sure unregistering handlers is required otherwise you might be running handlers of a disposed of ENV in a context of a new ENV

src/Voron/StorageEnvironmentOptions.cs
ayende: Yes, because it gives the wrong impression.
Basically, the rule is that unless you have looked at the ASM / have benchmark that show it, avoid it.

src/Raven.Server/Documents/Indexes/Static/IndexCompilationCache.cs
ayende: Can we move this from static to the database level?

src/Voron/StorageEnvironment.cs
ayende: @maximburyak we are just interested in the pagers, not he handlers. The handlers are always to the storage env, so that is going to be invalid anyway

src/Raven.Server/Documents/Indexes/Static/IndexCompilationCache.cs
talweiss1982: no we can't and refactoring how indexing is cached is out of the scope of this PR lets focus on memory leaks and not further expend this PR

maximburyak: I just created one: https://issues.hibernatingrhinos.com/issue/RavenDB-13819 separating the work will allow us to divide and conquer.

maximburyak:
rebased and ready to merge

talweiss1982:
I'm investigating NRE around the thread pool area please don't merge yet:
Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at Sparrow.Json.JsonContextPoolBase`1.MaybeGetCurrentContextStack() in C:\work\4.2\ravendb\src\Sparrow\Json\JsonContextPoolBase.cs:line 173
   at Sparrow.Json.JsonContextPoolBase`1.CleanThreadLocalState() in C:\work\4.2\ravendb\src\Sparrow\Json\JsonContextPoolBase.cs:line 194
   at System.Action.Invoke()
   at Sparrow.Utils.ThreadLocalCleanup.Run() in C:\work\4.2\ravendb\src\Sparrow\Utils\ThreadLocalCleanup.cs:line 11
   at Raven.Server.Utils.PoolOfThreads.PooledThread.DoWork() in C:\work\4.2\ravendb\src\Raven.Server\Utils\PoolOfThreads.cs:line 239
   at Raven.Server.Utils.PoolOfThreads.PooledThread.Run() in C:\work\4.2\ravendb\src\Raven.Server\Utils\PoolOfThreads.cs:line 187
   at System.Threading.Thread.ThreadMain_ThreadStart()
   at System.Threading.ThreadHelper.ThreadStart_Context(Object state)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ThreadHelper.ThreadStart()

talweiss1982:
This doesn't seem to be related to our changes

ayende:
Create an issue for this please, then.

talweiss1982:
apparently, it is related, i'll open a new PR for the fix.

PR #9492, opened 2019-07-11
RavenDB-13581: Pass query elapsed time on LazyQueryOperation

garayx:


src/Raven.Client/Documents/Session/Operations/Lazy/LazyQueryOperation.cs
```diff
@@ -56,12 +56,12 @@ public void HandleResponse(GetResponse response)
                     queryResult.DurationInMs = -1; // taken from cache
             }
 
-            HandleResponse(queryResult);
+            HandleResponse(queryResult, response.Elapsed);
         }
 
-        private void HandleResponse(QueryResult queryResult)
+        private void HandleResponse(QueryResult queryResult, TimeSpan responseElapsed)
```
ppekrol: rename to `duration`

src/Raven.Client/Documents/Session/Operations/QueryOperation.cs
```diff
@@ -273,6 +273,21 @@ public void EnsureIsAcceptableAndSaveResult(QueryResult result)
 
             EnsureIsAcceptable(result, _indexQuery.WaitForNonStaleResults, _sp, _session);
 
+            SaveQueryResult(result);
+        }
+
+        public void EnsureIsAcceptableAndSaveResult(QueryResult result, TimeSpan responseElapsed)
```
ppekrol: rename parameter to `duration`

src/Raven.Client/Documents/Session/Operations/QueryOperation.cs
```diff
@@ -331,6 +346,16 @@ public static void EnsureIsAcceptable(QueryResult result, bool waitForNonStaleRe
             }
         }
 
+        public static void EnsureIsAcceptable(QueryResult result, bool waitForNonStaleResults, TimeSpan? elapsed, InMemoryDocumentSessionOperations session)
```
ppekrol: rename to `duration` and call this method from the other `EnsureIsAcceptable

src/Raven.Client/Documents/Session/Operations/QueryOperation.cs
ppekrol: Maybe change to `internal` and make `duration` nullable? Then from the other `EnsureIsAcceptableAndSaveResult` call this method, but pass `null` for duration

src/Raven.Client/Documents/Session/Operations/Lazy/LazyQueryOperation.cs
garayx: done

src/Raven.Client/Documents/Session/Operations/QueryOperation.cs
garayx: done

PR #9495, opened 2019-07-11
Make JavaScriptConversionExtensions public

jonstelly:
Simple change to expose Linq2Js extensions so clients can reuse them.

https://issues.hibernatingrhinos.com/issue/RavenDB-13823

### DIFF
--- a/src/Raven.Client/Util/JavascriptConversionExtensions.cs
+++ b/src/Raven.Client/Util/JavascriptConversionExtensions.cs
@@ -18,7 +18,7 @@ using Sparrow.Extensions;
 
 namespace Raven.Client.Util
 {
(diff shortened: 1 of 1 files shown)


ppekrol:
@jonstelly We would like to avoid exposing those. If needed you can use `Activator` to create instances of those. FYI, client library is under MIT license.

The issue with exposing this to public is that it becomes a part of public API which we need to maintain in a more organized manner e.g. we need to maintain source-level or binary-level of compatibility when changes are being applied. When it is internal then those rules are less strict and we can apply more drastic modifications if needed, including major refactoring and method signature changes.

jonstelly:
That makes sense, thanks for the quick reply.  I'll head in the Activator direction.

PR #9501, opened 2019-07-14
RavenDB-13655 skipping test that throws SO on 32-bit and avoiding AVE…

talweiss1982:
… on cluster tx one by one

### DIFF
--- a/test/SlowTests/Issues/RavenDB_10621.cs
+++ b/test/SlowTests/Issues/RavenDB_10621.cs
@@ -9,7 +9,7 @@ namespace SlowTests.Issues
 {
     public class RavenDB_10621 : RavenTestBase
     {
-        [Fact]
+        [Fact(Skip = "Waiting for RavenDB-13830 to be resolved")]
         public void ShouldNotErrorIndexOnInvalidProgramException()
         {
             // if this test fails it's very likely the following issue got fixed: https://github.com/dotnet/coreclr/issues/14672
@@ -17,7 +17,6 @@ namespace SlowTests.Issues
             using (var store = GetDocumentStore())
             {
                 new BigIndexOutput().Execute(store);
-
                 using (var session = store.OpenSession())
                 {
                     session.Store(new ErroringDocument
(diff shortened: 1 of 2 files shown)


test/SlowTests/Issues/RavenDB_10621.cs
```diff
@@ -9,15 +9,14 @@ namespace SlowTests.Issues
 {
     public class RavenDB_10621 : RavenTestBase
     {
-        [Fact]
+        [Fact(Skip = "Waiting for RavenDB-13830 to be resolved")]
```
ayende, line 12: How does this generate this error?

talweiss1982: It doesn't it's an old test Michael wrote now it throws SO

maximburyak: @ayende I think that this issue is related to an old issue that we had: https://github.com/dotnet/coreclr/issues/14672

arekpalinski: Yes, we wrote this test to make sure that we're correctly handling `InvalidProgramException` during indexing. Once https://github.com/dotnet/coreclr/issues/14672 got fixed we adjusted the test. See the following commit:

https://github.com/ravendb/ravendb/commit/de998b8b9d36b181dfc95f69b4733c3074a4bb23#diff-452a08ab6d945d7e51d8683da35507b3

Note that Pawel changed `[Fact64Bit]` to `[Fact]`. That's why we never saw any failures on 32bits before probably.

Investigation is reported in: https://issues.hibernatingrhinos.com/issue/RavenDB-13830

arekpalinski:
@talweiss1982 can we close this PR. the same changes we have in https://github.com/ravendb/ravendb/pull/9511 right?

talweiss1982:
Yes

arekpalinski:
Closing in favor of https://github.com/ravendb/ravendb/pull/9511

PR #9502, opened 2019-07-15
RavenDB-13153 OutOfMemory notification changed to show once for each environment

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-13153

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -1073,13 +1073,17 @@ namespace Raven.Server.Documents.Indexes
                                     if (_logger.IsOperationsEnabled)
                                         _logger.Operations($"Failed to open write transaction, indexing will be retried", te);
                                 }
+                                catch (InsufficientMemoryException ime)
+                                {
+                                    HandleOutOfMemoryException(ime, scope, $"{_environment}:{nameof(InsufficientMemoryException)}");
+                                }
                                 catch (OutOfMemoryException oome)
                                 {
-                                    HandleOutOfMemoryException(oome, scope);
+                                    HandleOutOfMemoryException(oome, scope, $"{_environment}:{nameof(OutOfMemoryException)}");
                                 }
                                 catch (EarlyOutOfMemoryException eoome)
                                 {
-                                    HandleOutOfMemoryException(eoome, scope);
+                                    HandleOutOfMemoryException(eoome, scope, $"{_environment}:{nameof(EarlyOutOfMemoryException)}");
                                 }
                                 catch (VoronUnrecoverableErrorException ide)
                                 {
@@ -1543,7 +1547,7 @@ namespace Raven.Server.Documents.Indexes
             SetState(IndexState.Error, ignoreWriteError: true);
         }
 
-        private void HandleOutOfMemoryException(Exception oome, IndexingStatsScope scope)
+        private void HandleOutOfMemoryException(Exception oome, IndexingStatsScope scope, string key)
         {
             try
             {
@@ -1555,20 +1559,8 @@ namespace Raven.Server.Documents.Indexes
                 if (_logger.IsInfoEnabled)
                     _logger.Info(title, oome);
 
-                var message = $"Error message: {oome.Message}";
-                var alert = AlertRaised.Create(
-                    null,
-                    title,
-                    message,
-                    AlertType.OutOfMemoryException,
-                    NotificationSeverity.Error,
-                    key: message,
-                    details: new MessageDetails
-                    {
-                        Message = OutOfMemoryDetails(oome)
-                    });
-
-                DocumentDatabase.NotificationCenter.Add(alert);
+//                DocumentDatabase.NotificationCenter.Dismiss(AlertRaised.GetKey(AlertType.OutOfMemoryException, key));
+                DocumentDatabase.NotificationCenter.OutOfMemory.Add(title,  key, oome);
             }
             catch (Exception e) when (e.IsOutOfMemory())
             {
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1066,13 +1066,17 @@ protected void ExecuteIndexing()
                                     if (_logger.IsOperationsEnabled)
                                         _logger.Operations($"Failed to open write transaction, indexing will be retried", te);
                                 }
+                                catch (InsufficientMemoryException ime)
+                                {
+                                    HandleOutOfMemoryException(ime, scope, $"{_environment}:{nameof(InsufficientMemoryException)}");
```
ayende: We are in low memory already, and you are allocating additional string.
Let's create this once

ayende (commented):
Also, fix the commit message "shoe"

src/Raven.Server/NotificationCenter/OutOfMemoryNotifications.cs
```diff
@@ -0,0 +1,127 @@
...
+        private readonly ConcurrentDictionary<string, NotificationTime> _notificationTimes = new ConcurrentDictionary<string, NotificationTime>();
+        private readonly ConcurrentDictionary<(StorageEnvironment, Type), NotificationTime> _notificationsMetadata = new ConcurrentDictionary<(StorageEnvironment, Type), NotificationTime>();
+        private readonly NotificationCenter _notificationsCenter;
+
+        public OutOfMemoryNotifications(NotificationCenter notificationsCenter)
+        {
+            _notificationsCenter = notificationsCenter;
+        }
+
+        public void Add(string title, string key, Exception exception)
```
arekpalinski: I don't see any usage of this method (overload with 3 parameters). Do we need it?

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1536,32 +1536,18 @@ private void SetErrorState(string reason)
...
-                    NotificationSeverity.Error,
-                    key: message,
-                    details: new MessageDetails
-                    {
-                        Message = OutOfMemoryDetails(oome)
-                    });
+                    _logger.Info($"Out of memory occurred for '{Name}'", exception);
 
-                DocumentDatabase.NotificationCenter.Add(alert);
+                DocumentDatabase.NotificationCenter.OutOfMemory.Add(_environment, exception);
```
maximburyak, line 1557: This looks a bit fishy to me. Can you please verify that we do not leak _environment here?

arekpalinski:
@haludi Please rebase your branch agains v4.2 where .net core has been updated so Jenkins will be able to run tests (currently it's failing doe to .net core mismatch)

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1547,7 +1548,11 @@ private void HandleOutOfMemoryException(Exception exception, IndexingStatsScope
                 if (_logger.IsInfoEnabled)
                     _logger.Info($"Out of memory occurred for '{Name}'", exception);
 
-                DocumentDatabase.NotificationCenter.OutOfMemory.Add(_environment, exception);
+                //We are in low of memory so we want to allocate the strings only once
+                if (_environmentStr == null)
```
ayende: You haven't changed much. You are still allocating _when you are in OOM condition_

src/Raven.Server/NotificationCenter/OutOfMemoryNotifications.cs
```diff
@@ -13,52 +13,15 @@ public class OutOfMemoryNotifications
     {
         private readonly TimeSpan _updateFrequency = TimeSpan.FromSeconds(15);
         private readonly ConcurrentDictionary<string, NotificationTime> _notificationTimes = new ConcurrentDictionary<string, NotificationTime>();
-        private readonly ConcurrentDictionary<(StorageEnvironment, Type), NotificationTime> _notificationsMetadata = new ConcurrentDictionary<(StorageEnvironment, Type), NotificationTime>();
+        private readonly ConcurrentDictionary<(string, Type), NotificationTime> _notificationsMetadata = new ConcurrentDictionary<(string, Type), NotificationTime>();
```
ayende: I don't like this much.
The problem is that this is never cleaned, and may end up being a problem area all on its own.

ayende (changes requested):

src/Raven.Server/Documents/Indexes/Index.cs
maximburyak: on the other hand, the for no InMemory databases, .ToString() eventually return a reference to the config path value, that is already materialized, so I don't think it's a big issue here.
But maybe we should return to using the Environment, but use a ConditionalTable instead of ConcurrentDictionary

ayende: I think that this is a much better option, yes.

src/Raven.Server/NotificationCenter/OutOfMemoryNotifications.cs
```diff
@@ -0,0 +1,92 @@
...
+            Volatile.Write(ref notificationMetadata.IsInProgress, 0);
+        }
+
+        private static MessageDetails OutOfMemoryDetails(Exception exception)
+        {
+            var memoryInfo = MemoryInformation.GetMemoryInformationUsingOneTimeSmapsReader();
+
+            return new MessageDetails
+            {
+                Message = $"Managed memory: {new Size(AbstractLowMemoryMonitor.GetManagedMemoryInBytes(), SizeUnit.Bytes)}, " +
```
maximburyak, line 73: In continuation to a conversation with @haludi, created this issue: https://issues.hibernatingrhinos.com/issue/RavenDB-13851 for now, we inherit this code from previous state, so I'm fine with it.

test/SlowTests/Server/NotificationCenter/OutOfMemoryNotificationTests.cs
```diff
@@ -0,0 +1,165 @@
...
+                var environment = database.GetAllStoragesEnvironment().First().Environment;
+
+                var actions = new AsyncQueue<DynamicJsonValue>();
+
+                var writer = new TestWebSocketWriter();
+                var taskList = new List<Task>();
+                using (database.NotificationCenter.TrackActions(actions, writer))
+                {
+
+                    taskList.Add(Task.Run( () => database.NotificationCenter.OutOfMemory.Add(environment, new OutOfMemoryException())));
```
maximburyak, line 59: I think that we should also test that the environment is removed from the OutOfMemory collection by forcing GC

maximburyak (changes requested):

test/SlowTests/Server/NotificationCenter/OutOfMemoryNotificationTests.cs
arekpalinski: @haludi ?

arekpalinski (approved):

PR #9507, opened 2019-07-15
RavenDB-13712: Fix possible gaps in AsyncHiLoIdGenerator

garayx:


src/Raven.Client/Documents/Identity/AsyncHiLoIdGenerator.cs
```diff
@@ -94,6 +94,10 @@ public async Task<long> NextIdAsync()
                 //local range is exhausted , need to get a new range
                 var maybeNextTask = new Lazy<Task>(GetNextRangeAsync);
 
+                // other thread got new range
+                if (id <= Range.Max)
+                    continue;
+
```
karmeli87, line 99: If the thread is blocked here, you still got the race :-)

garayx: done

garayx:
will reopen in 4.1

PR #9510, opened 2019-07-16
RavenDB-13712: Fix race in AsyncHiLoIdGenerator

garayx:


src/Raven.Client/Documents/Identity/AsyncHiLoIdGenerator.cs
```diff
@@ -83,25 +83,30 @@ public async Task<long> NextIdAsync()
...
+                var nextTask = Interlocked.CompareExchange(ref _nextRangeTask, maybeNextTask, current);
+                if (nextTask == current) // replace was successful
                 {
-                    Interlocked.CompareExchange(ref _nextRangeTask, null, nextTask);
+                    await maybeNextTask.Value.ConfigureAwait(false);
+                    continue;
                 }
+
+                // failed to replace, let's wait on the previous task
+                await nextTask.Value.ConfigureAwait(false);
```
arekpalinski: You dropped `try .. finally` and effectively `_nextRangeTask` is no longer set to `null` on exception

garayx: done

ayende (approved):

src/Raven.Client/Documents/Identity/AsyncHiLoIdGenerator.cs
```diff
@@ -83,25 +83,37 @@ public async Task<long> NextIdAsync()
...
                 {
-                    await nextTask.Value.ConfigureAwait(false);
+                    // let's try to call the existing task for next range
+                    await current.Value.ConfigureAwait(false);
+                    if (range != Range)
+                        continue;
+                }
+                catch
+                {
+                    // previous task was faulted, we will replace it
```
arekpalinski: How do you assure that we'll replace it? If someone replace `_nextRangeTask` meanwhile (possible, right?)  then below condition `if (nextTask == current)` will be false an we'll throw on:

```
// failed to replace, let's wait on the previous task
                await nextTask.Value.ConfigureAwait(false);
```

ayende: Good point, so we need to only throw an error higher IFF our own created task failed

src/Raven.Client/Documents/Identity/AsyncHiLoIdGenerator.cs
```diff
@@ -83,25 +83,37 @@ public async Task<long> NextIdAsync()
...
+                var nextTask = Interlocked.CompareExchange(ref _nextRangeTask, maybeNextTask, current);
+                if (nextTask == current) // replace was successful
                 {
-                    Interlocked.CompareExchange(ref _nextRangeTask, null, nextTask);
+                    await maybeNextTask.Value.ConfigureAwait(false);
+                    continue;
                 }
+
+                // failed to replace, let's wait on the previous task
+                await nextTask.Value.ConfigureAwait(false);
```
ayende: This should be wrapped in try/catch and ignored.
Someone else failed, but we'll retry until our own task failed

ayende (changes requested):

src/Raven.Client/Documents/Identity/AsyncHiLoIdGenerator.cs
garayx: done

arekpalinski (approved):

ayende (approved):

PR #9511, opened 2019-07-16
Raven db 13655

talweiss1982:


test/SlowTests/Issues/RavenDB_12151.cs
```diff
@@ -116,7 +116,7 @@ public void IndexingWhenGlobalScratchSpaceLimitExceeded32()
             }
         }
 
-        [Fact]
+        [Fact(Skip ="Transaction size changed below 1MB")]
```
ayende: Why this change?

ayende: What changed the TX size? We need to fix the test / figure out why it changed

talweiss1982: hard to track what caused it, @maximburyak and I decided to finish fixing the rest of the tests and try to figure this one on a later PR since it was taking too long and I couldn't track what caused it.
I know for sure this test was passing when I wrote it.
now the largest Tx is 334kb...

talweiss1982: It is likely the limitations I have put on the merger that are limiting the indexing transactions. I'll open a separate issue to investigate this but for the short term let's get the tests green again.

talweiss1982: It is likely the limitations I have put on the merger that are limiting the indexing transactions. I'll open a separate issue to investigate this but for the short term let's get the tests green again.

arekpalinski (approved):

test/SlowTests/Issues/RavenDB_12151.cs
maximburyak: That was resolved in the commit below

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -3278,6 +3276,9 @@ private int MinBatchSize
         {
             var txAllocationsInBytes = UpdateThreadAllocations(indexingContext, indexWriteOperation, stats, updateReduceStats: false);
 
+            //We need to take the read transaction encryption size into account as we might read alot of document and produce very little indexing output.
+            txAllocationsInBytes += documentsOperationContext.Transaction.InnerTransaction.LowLevelTransaction.TotalEncryptionBufferSize.GetValue(SizeUnit.Bytes);
```
maximburyak, line 3280: Nice

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -1217,7 +1217,9 @@ public void Dispose()
                 }
             });
 
-            ea.Execute(() => ConflictResolver?.ResolveConflictsTask.Wait());
+            ea.Execute(() => ConflictResolver?.ResolveConflictsTask.Wait(TimeSpan.FromSeconds(15)));
```
maximburyak: I'm not sure about the consequences of not waiting through here, and we do a similar unbounded wait few line above, although I agree that unbounded waits in dispose function is a bad practice in general.
```suggestion
            ea.Execute(() => ConflictResolver?.ResolveConflictsTask.Wait());
```

test/SlowTests/Client/Subscriptions/RavenDB-8404.cs
```diff
@@ -48,7 +48,11 @@ public async Task SubscriptionsRQLSupportStartsWith()
                 _ = subscription.Run(x =>
                 {
                     users.AddRange(x.Items.Select(i => i.Result));
-                    amre.Set();
+                    //in 32bit we seem to send batches of 1 User
+                    if (users.Count == 3)
```
maximburyak, line 52: Right, because we started limiting the batch size in subscriptions

maximburyak (changes requested):

maximburyak (approved):

ayende (approved):
Very nice work.

PR #9512, opened 2019-07-16
RavenDB-13835 Concurrent operations on Raft sometimes fail due to pre…

karmeli87:
…cision of DateTime.UtcNow.Ticks

### DIFF
--- a/src/Raven.Server/Rachis/RachisLogHistory.cs
+++ b/src/Raven.Server/Rachis/RachisLogHistory.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Text;
+using System.Threading;
 using Raven.Client.Util;
 using Raven.Server.Config;
 using Raven.Server.ServerWide;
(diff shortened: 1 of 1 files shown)


talweiss1982:
I think that this is a very expensive way to ensure uniqueness.
Maybe we can hash the ticks with the term and index (with a reversible hashing method) to generate a unique timestamp and when we read the TVR we can un-hash the value to get back the ticks.

talweiss1982:
Or use compound key

karmeli87:
1. This is an indexed field, if you will hash it, you mess up the search.
2. Right now this called under the write lock, so this shouldn't be really CPU consuming.

src/Raven.Server/Rachis/RachisLogHistory.cs
```diff
@@ -49,6 +50,27 @@ private enum HistoryStatus : byte
             Committed
         }
 
+        private static long _lastTicks;
+        private static long GetUniqueTicks()
```
ayende: Let's avoid using `static` here, I want to be able to run this concurrently on two Rachis instances without issue.
We don't need multi threading support, since we can assume that this is in a write transaction.
Let's accept the transaction as a parameter here and check that we have a write transaction to avoid future issues.

src/Raven.Server/Rachis/RachisLogHistory.cs
```diff
@@ -49,6 +50,27 @@ private enum HistoryStatus : byte
...
+            {
+                var lastTicks = _lastTicks;
+                var ticks = DateTime.UtcNow.Ticks;
+
+                if (ticks <= lastTicks)
+                {
+                    ticks = lastTicks + 1;
+                }
+
+                var oldValue = Interlocked.CompareExchange(ref _lastTicks, ticks, lastTicks);
```
ayende: No need for compare exchange here

src/Raven.Server/Rachis/RachisLogHistory.cs
```diff
@@ -49,6 +50,27 @@ private enum HistoryStatus : byte
             Committed
         }
 
+        private static long _lastTicks;
+        private static long GetUniqueTicks()
+        {
+            while (true)
```
ayende: No need for a while here.

ayende (changes requested):

ayende (approved):

arekpalinski (approved):

PR #9517, opened 2019-07-17
RavenDB-13836 background worker should return canceled token if they …

talweiss1982:
…are disposed

### DIFF
--- a/src/Raven.Server/Background/BackgroundWorkBase.cs
+++ b/src/Raven.Server/Background/BackgroundWorkBase.cs
@@ -31,11 +31,18 @@ namespace Raven.Server.Background
                 }
                 catch (ObjectDisposedException)
                 {
-                    //we are disposed, so return "null" token                   
-                    return CancellationToken.None;
+                    //we are disposed, so return a Canceled token ( CancellationToken.None isn't canceled and causes task to freeze)               
+                    return CanceledToken.Value;
                 }
             }
         }
+
+        private static Lazy<CancellationToken> CanceledToken => new Lazy<CancellationToken>(() =>
+           {
+               var cts = new CancellationTokenSource();
+               cts.Cancel();
+               return cts.Token;
+           });
 
         public void Start()
         {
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Background/BackgroundWorkBase.cs
```diff
@@ -84,6 +91,10 @@ protected async Task WaitOrThrowOperationCanceled(TimeSpan time)
 
                 // if cancellation requested then it will throw TaskCancelledException and we stop the work
                 await TimeoutManager.WaitFor(time, CancellationToken).ConfigureAwait(false); 
+                if(CancellationToken.IsCancellationRequested)
```
ayende, line 94: Shouldn't the `WaitFor` already throw that?

ayende (approved):

src/Raven.Server/Background/BackgroundWorkBase.cs
talweiss1982: It doesn't I have checked we will get stuck in an infinite loop

ayende: We also need to update the `WaitFor`, then.

talweiss1982: I think it would change its current behavior and it is risky, no?

ayende: I expected that to fail, and not hang, if the cancellation token is cancelled.

talweiss1982: what is hanging is the expired document cleaner logic, not the timing manager.
But if you pass a default cancelation token to the timing manager it will hang so it is a problem with the usage, not the underlining logic.
@maximburyak will open an issue to review usages

maximburyak: We'll address that in a different issue: https://issues.hibernatingrhinos.com/issue/RavenDB-13849

arekpalinski (approved):

PR #9519, opened 2019-07-19
RavenDB-13840 attempted to delete value but it wasn't in index

karmeli87:
seems to fix the repro.
this is draft to let the test suit run.

### DIFF
--- a/src/Raven.Server/Config/Categories/StorageConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/StorageConfiguration.cs
@@ -13,7 +13,7 @@ namespace Raven.Server.Config.Categories
         public PathSetting TempPath { get; set; }
 
         [Description("Use the 32 bits memory mapped pager, even when running in 64 bits")]
-        [DefaultValue(false)]
+        [DefaultValue(true)]
         [ConfigurationEntry("Storage.ForceUsing32BitsPager", ConfigurationEntryScope.ServerWideOnly)]
         public bool ForceUsing32BitsPager { get; set; }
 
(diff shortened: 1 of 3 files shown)


karmeli87:
I think that regardless to this fix we have AVE in 32 bit :/

src/Raven.Server/Config/Categories/StorageConfiguration.cs
```diff
@@ -13,7 +13,7 @@ public class StorageConfiguration : ConfigurationCategory
         public PathSetting TempPath { get; set; }
 
         [Description("Use the 32 bits memory mapped pager, even when running in 64 bits")]
-        [DefaultValue(false)]
+        [DefaultValue(true)]
```
ayende: Need to revert this

ayende (commented):
The issue is that we are deleting the active section and then allocating from it?

src/Raven.Server/Config/Categories/StorageConfiguration.cs
karmeli87: This is draft and I want it to run on 32 bits (64 is by default)

karmeli87:
I'm still not 100% sure, but... 
The way we are getting the `ActiveDataSmallSection` is as following:
https://github.com/ravendb/ravendb/blob/71ea31731fbd3ead4aa0d09ef2902d4ba0d5340c/src/Voron/Data/Tables/Table.cs#L44

And when we delete a section we forget to update this value, so in the next tx we loading the wrong page.

ayende (approved):

karmeli87:
Closing and reopen in 4.1 since is has to be done there also. (https://github.com/ravendb/ravendb/pull/9520)

PR #9522, opened 2019-07-21
RavenDB-13788 Add Retention Policy in edit backup configuration

Danielle9897:
Note: The changes in `cronEditor.html `were done in order for the text to appear straight, in the same line.
Without this, the text is slightly shifted vertically.


ml054:
LGTM

PR #9528, opened 2019-07-21
RavenDB-13853 fix

myarichuk:


src/Raven.Server/Documents/Queries/Dynamic/CollectionQueryRunner.cs
```diff
@@ -186,6 +186,11 @@ private void ExecuteCollectionQuery(QueryResultServerSide<Document> resultToFill
                             resultToFill.TotalResults - (query.Offset ?? 0)
                         );    
                     }
+
+                    if (query.Offset.HasValue)
+                    {
+                        resultToFill.SkippedResults = query.Offset.GetValueOrDefault();
```
ayende, line 192: Do we do the same when we are using non collection runner?

arekpalinski (approved):

arekpalinski:
@myarichuk so we don't need this? https://issues.hibernatingrhinos.com/issue/RavenDB-13853#focus=streamItem-67-39514.0-0

arekpalinski:
SkippedResults is meant to be used only when paging through tampered query results - https://ravendb.net/docs/article-page/4.2/Csharp/indexes/querying/paging#paging-through-tampered-results

PR #9536, opened 2019-07-24
RavenDB-13130 When freeing a page and doing validation we need to tak…

arekpalinski:
…e into account the actual number of allocated pages which is different on 32bits


ayende (approved):

arekpalinski:
Related comment explaining we didn't just set NumberOfPagesInSection accordingly in 32bits: https://github.com/ravendb/ravendb/pull/8790#discussion_r260174091

PR #9541, opened 2019-07-24
RavenDB-13815 Can't see the properties of a document when the documen…

Danielle9897:
…t ID is long


ml054:
Can't we use CSS for that?

Danielle9897:
1. Closing this pr and passing to @Rafal Kwiatkowski to solve with CSS only
2. Will open a different issue for the 'Copy Id' issue 
https://issues.hibernatingrhinos.com/issue/RavenDB-13871

PR #9542, opened 2019-07-24
RavenDB-13864 Demote node should be for cluster admins only

karmeli87:


ayende (approved):

karmeli87:
We need this for 4.1 as well

PR #9544, opened 2019-07-24
RavenDB-13869: Clear EntityToBlittable's _missingDictionary on evict …

maximburyak:
…and clear

### DIFF
--- a/test/FastTests/Issues/RavenDB_13869.cs
+++ b/test/FastTests/Issues/RavenDB_13869.cs
@@ -1,0 +1,81 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using FastTests.Server.Basic.Entities;
+using Raven.Client.Documents.Session;
+using Raven.Tests.Core.Utils.Entities;
+using Xunit;
+
+namespace FastTests.Issues
+{
+    public class RavenDB_13869:RavenTestBase
+    {
+        
+        [Fact]
+        public void MissingFieldsDataShouldBeCleared()
+        {
+            using (var store = GetDocumentStore())
+            {
+                using (var session = store.OpenSession())
+                {
+                    //person
+                    session.Store(new User
+                    {
+                        Name = "John",
+                        LastName = "Doh"
+                    },"users/1");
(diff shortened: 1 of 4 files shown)


ayende (approved):

ayende:
Can you also add a test?

test/FastTests/Issues/RavenDB_13869.cs
```diff
@@ -0,0 +1,81 @@
...
+                    session.Advanced.Clear();
+
+                    session.Store(person, "users/2");                    
+                    session.SaveChanges();                    
+                }
+
+                using (var session = store.OpenSession())
+                {
+                    var person = session.Load<User>("users/2");
+                    Assert.Null(person.LastName);
```
arekpalinski: No sue if I follow - why is it `null`?

maximburyak: It should be null, because we called session.clear, so we didn't send the not materialized fields back to the server

maximburyak: It's a not direct method to test the cleanup behavior..

arekpalinski: but we called `session.Store(person, "users/2");` and `session.Clear` clears a state of UoW, not properties of entities - am I missing something?

ayende: `User` vs `Person`, I think.
`Person` doesn't have `LastName` ?

maximburyak: That's right @ayende , so after clear, we should not send this value to the server. And if we do, that's an indication we didn't clear this removed fields data

ayende: Can you use types that are defined in the test? It is _very_ hard to figure out inline

arekpalinski: Yeah, I haven't noticed this - `User` vs `Person`

ayende (approved):

arekpalinski (approved):

PR #9547, opened 2019-07-25
RavenDB-13872 Respect the transaction boundaries in counters replication

karmeli87:


src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -1248,11 +1248,17 @@ protected override int ExecuteCmd(DocumentsOperationContext context)
...
                     var database = _replicationInfo.DocumentDatabase;
+                    var lastTransactionMarker = 0;
 
                     context.LastDatabaseChangeVector = context.LastDatabaseChangeVector ?? DocumentsStorage.GetDatabaseChangeVector(context);
                     foreach (var item in _replicationInfo.ReplicatedItems)
                     {
-                        context.TransactionMarkerOffset = item.TransactionMarker;
+                        if (lastTransactionMarker != item.TransactionMarker)
+                        {
+                            context.TransactionMarkerOffset++;
```
arekpalinski, line 1258: Why do we need that?

karmeli87: `TransactionMarkerOffset` is to distinguish between transactions when they get merged by the tx merger.
Incoming replication can have a batch that contains several transactions, and we need to distinguish between them as well.

The fix here is in case this command is part of merged tx, we don't want to override the existing `TransactionMarkerOffset`.

arekpalinski (approved):

ayende (approved):

PR #9557, opened 2019-07-28
Raven db 13831 2

maximburyak:


src/Raven.Server/Documents/Subscriptions/SubscriptionStorage.cs
```diff
@@ -460,5 +463,43 @@ public void ReleaseSubscriptionsSemaphore()
         {
             _concurrentConnectionsSemiSemaphore.Release();
         }
+
+        internal void CleanupSubscriptions()
+        {
+            var maxTaskLifeTime = (PlatformDetails.Is32Bits || this._db.Configuration.Storage.ForceUsing32BitsPager
+                        ? TimeSpan.FromHours(12)
```
ayende, line 469: I don't think it matters, to be frank. Either way is very long

src/Raven.Server/Documents/Subscriptions/SubscriptionStorage.cs
```diff
@@ -460,5 +463,43 @@ public void ReleaseSubscriptionsSemaphore()
...
+            foreach (var state in _subscriptionConnectionStates)
+            {
+                if (state.Value.Connection != null)
+                    continue;
+
+                var recentConnection = state.Value.MostRecentEndedConnection();
+                if (recentConnection == null)
+                    return;
+
+                if (recentConnection.Stats.LastMessageSentAt < oldestPossibleIdleSubscription)
```
ayende: You are removing just one here, what if we have a lot of them? Should't we put this in a while?

src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -325,6 +325,7 @@ public JsonOperationContext(int initialSize, int longLivedSize, SharedMultipleUs
                 TryDispose(_documentBuilder);
                 TryDispose(_arenaAllocator);
                 TryDispose(_arenaAllocatorForLongLivedValues);
+                TryExecute(_allocateStringValues.Clear);
```
ayende, line 328: shouldn't we also set it to null?

src/Raven.Server/Documents/Subscriptions/SubscriptionStorage.cs
maximburyak: Yeah, I didn't want to add a new configuration here, so I decided to have a small preemptive move

ayende (commented):
I assume you verified that in the long living contexts, we aren't hanging on to instances after the reset?

ayende (approved):

maximburyak:
@ayende that's why I didn't touch bulk insert and smuggler, it's much trickier there..

arekpalinski:
@maximburyak Could you check failing tests

ppekrol:
@maximburyak can you take a look at whitespace issue also?

maximburyak:
pushing a rebase

src/Raven.Server/Documents/Subscriptions/SubscriptionStorage.cs
maximburyak: done

src/Sparrow/Json/JsonOperationContext.cs
maximburyak: IMO, clear is enough here, we don't want to expose ourselves to potential NREs

PR #9561, opened 2019-07-29
RavenDB-13884 fixed changes API connection leak due to invalid hash code calculation in the comparer

ppekrol:


ayende (approved):
I assume the issue was with the hash code calc?

ppekrol:
Correct, the comparer was executing:

```
        public int GetHashCode(DatabaseChangesOptions n)
        {
            return n.GetHashCode();
        }
```

PR #9562, opened 2019-07-30
RavenDB-13098 Replication sending / incoming refactoring

karmeli87:


src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -232,9 +232,9 @@ private static ReplicationBatchItem CreateReplicationBatchItem(DocumentsOperatio
             var data = GetCounterValuesData(context, ref tvh.Reader);
             var docId = ExtractDocId(context, ref tvh.Reader);
 
-            return new ReplicationBatchItem
+            return new CounterReplicationItem
             {
-                Type = ReplicationItemType.CounterGroup,
+                Type = ReplicationBatchItem.ReplicationItemType.CounterGroup,
```
ayende, line 237: Why do we need this now?

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -723,228 +632,24 @@ private void SendHeartbeatStatusToSource(DocumentsOperationContext documentsCont
...
-        private unsafe ArraySegment<ReplicationItem> ReadItemsFromSource(int replicatedDocs, DocumentsOperationContext context,
-            IncomingReplicationAllocator incomingReplicationAllocator, IncomingReplicationStatsScope stats)
-        {
-            var items = ArrayPool<ReplicationItem>.Shared.Rent(replicatedDocs);
-
-            var documentRead = stats.For(ReplicationOperation.Incoming.DocumentRead, start: false);
-            var attachmentRead = stats.For(ReplicationOperation.Incoming.AttachmentRead, start: false);
-            var tombstoneRead = stats.For(ReplicationOperation.Incoming.TombstoneRead, start: false);
+            if (data.ReplicatedItems == null)
+                data.ReplicatedItems = new ReplicationBatchItem[replicatedDocs];
```
ayende, line 639: Is this cached? Can you get a smaller value from a previous call?

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -1259,54 +957,65 @@ protected override int ExecuteCmd(DocumentsOperationContext context)
...
-                                if (_replicationInfo.ReplicatedAttachmentStreams.TryGetValue(item.Base64Hash, out var attachmentStream))
+                                break;
+                            case AttachmentTombstoneReplicationItem attachmentTombstone:
+                                database.DocumentsStorage.AttachmentsStorage.DeleteAttachmentDirect(context, attachmentTombstone.Key, false, "$fromReplication", null,
+                                    rcvdChangeVector,
+                                    attachmentTombstone.LastModifiedTicks);
+                                break;
+
+                            case RevisionTombstoneReplicationItem revisionTombstone:
+                                item.ToDispose(Slice.From(context.Allocator, revisionTombstone.Id, out var id)); // TODO: From not required
```
ayende, line 989: I don't understand the `TODO`?

ppekrol (approved):

src/Raven.Server/Documents/CountersStorage.cs
karmeli87: We are reading the `Type` from the stream, so basically we need those for backward compatibility.

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
karmeli87: No it isn't. for now we create a new instance every time.

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
karmeli87: `Slice.From` will allocate and copy the value of `revisionTombstone.Id`. This is not required here because we (the current context) already own `revisionTombstone.Id`.

This is used only to convert from `LazyStringValue` to `Slice`.

PR #9566, opened 2019-07-31
RavenDB-13771 Storage summary on the dashboard doesn't include the sy…

Danielle9897:
…stem storage

### DIFF
--- a/src/Raven.Server/Dashboard/DatabasesInfoNotificationSender.cs
+++ b/src/Raven.Server/Dashboard/DatabasesInfoNotificationSender.cs
@@ -240,7 +240,7 @@ namespace Raven.Server.Dashboard
                     
                         // Get new data 
                         var systemEnv = new StorageEnvironmentWithType("<System>", StorageEnvironmentWithType.StorageEnvironmentType.System, serverStore._env);
-                        var systemMountPoints = StorageEnvironment.GetMountPointUsageDetailsFor(systemEnv);
+                        var systemMountPoints = ServerStore.GetMountPointUsageDetailsFor(systemEnv);
                     
                         foreach (var systemPoint in systemMountPoints)
                         {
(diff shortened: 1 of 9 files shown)


Danielle9897:
Closing - I need to fix tests....

src/Raven.Server/Dashboard/DatabasesInfoNotificationSender.cs
```diff
@@ -97,6 +98,15 @@ private class DatabaseInfoCache
             public DateTime NextDiskSpaceCheck;
         }
 
+        private static readonly SystemInfoCache CachedSystemInfo = new SystemInfoCache();
```
ayende: Why have this?
Note that in some cases, we might have these things come and go dynamically.
I'm also concerned about `MountPoints` being a static `List<T>`, which can cause concurrency issues

src/Raven.Server/Dashboard/DatabasesInfoNotificationSender.cs
```diff
@@ -211,6 +222,37 @@ public static IEnumerable<AbstractDashboardNotification> FetchDatabasesInfo(Serv
...
+                // 2. Add <system> info 
+                if (isValidFor == null) 
+                {
+                    var currentSystemHash = serverStore._env.CurrentReadTransactionId;
+                    if (currentSystemHash != CachedSystemInfo.Hash || CachedSystemInfo.NextDiskSpaceCheck <  SystemTime.UtcNow)
+                    {
+                        // Update cached value
+                        CachedSystemInfo.Hash = currentSystemHash;
+                        CachedSystemInfo.NextDiskSpaceCheck = SystemTime.UtcNow.AddSeconds(30);
+                        CachedSystemInfo.MountPoints.Clear();
```
ayende: what happens if there is a concurrent read of this at the same time?

ayende: At a minimum, you need to replace the whole object in one go

Danielle9897: Code refactored - See the second commit

src/Raven.Server/Dashboard/DatabasesInfoNotificationSender.cs
Danielle9897: Code refactored - See the second commit

PR #9568, opened 2019-08-01
RavenDB-13891: Patch modifies collection in cached document

garayx:


src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
```diff
@@ -328,7 +329,18 @@ private void HandlePatch(BlittableJsonReaderObject batchResult)
                     }
 
                     if (documentInfo.Entity != null)
-                        _session.EntityToBlittable.PopulateEntity(documentInfo.Entity, id, documentInfo.Document, _session.JsonSerializer);
+                    {
+                        var old = _session.JsonSerializer.ObjectCreationHandling;
```
ppekrol: This needs to be inside Populate

src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
```diff
@@ -328,7 +329,18 @@ private void HandlePatch(BlittableJsonReaderObject batchResult)
...
+                        _session.JsonSerializer.ObjectCreationHandling = ObjectCreationHandling.Replace;
```
ppekrol: What should happen if I have a patch like
`this.Address.City = 'abc'`
? 

The `Address` will be a new instance or old? What behavior do we want? Only the root entity instance cannot be changed?

test/FastTests/Issues/RavenDB_13891.cs
```diff
@@ -0,0 +1,74 @@
...
+                    await s.SaveChangesAsync();
+
+                    person = await s.LoadAsync<Person>(personId);
+                    Assert.Equal("aaa", person.Text);
+                    Assert.Equal(1, person.Modifications.Count);
+                }
+            }
+        }
+
+        public class Person
```
ppekrol: private

src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
ayende: Agreed.

src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
ayende: It will create a new instance, yes.
The root instance is the only one we care about here, yes.

test/FastTests/Issues/RavenDB_13891.cs
ayende: What is the reasoning here? This is inside a test class.

ppekrol: Good practice. E.g. to avoid using it from different tests (happens often). Tests should be isolated and use only their own types or common ones defined elsewhere if possible.

src/Raven.Client/Documents/Session/Operations/BatchOperation.cs
garayx: done

test/FastTests/Issues/RavenDB_13891.cs
garayx: done

PR #9570, opened 2019-08-05
RavenDB-13673 [NOT FOR MERGING]

talweiss1982:


src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexFacetedReadOperation.cs
```diff
@@ -279,9 +279,83 @@ private static void ApplyAggregation(Dictionary<string, FacetedQueryParser.Facet
             }
         }
 
+        public class NoWeightQuery : Query
```
ayende, line 282: what is this for?

src/Raven.Server/Documents/Queries/LuceneQueryHelper.cs
```diff
@@ -137,7 +138,7 @@ public static Query Between(string fieldName, LuceneTermType termType, double fr
             return CreateRange(fieldName, fromValue, true, toValue, true);
         }
 
-        public static Query Term(string fieldName, string term, LuceneTermType type, float? boost = null, float? similarity = null, bool exact = false)
+        public static Query Term(string fieldName, string term, LuceneTermType type, float? boost = null, float? similarity = null, bool exact = false, bool lightWeight = false)
```
ayende, line 141: I would rather make this into an instance, that has two static properties `Normal` and `Lightweight`

ayende: So: `LuceneQueryHelper.Lightwegiht.Term(...)`

ppekrol:
Handled here: https://github.com/ravendb/ravendb/pull/9584

PR #9573, opened 2019-08-06
Raven db 13831 3

maximburyak:


maximburyak:
This is the last one (promise)

src/Sparrow/Utils/NativeMemory.cs
```diff
@@ -38,11 +38,20 @@ public class ThreadStats
...
+            private WeakReference<Thread> _threadInstance = new WeakReference<Thread>(null);
             public readonly int ManagedThreadId;
             private string _lastName = "Unknown";
-            public string Name => _threadInstance?.Name ?? _lastName;
-
+            public string Name
+            {
+                get
+                {
+                    if (_threadInstance.TryGetTarget(out var thread))
```
ayende, line 48: I don't like this. You missed `_lastName`

src/Raven.Server/Documents/Subscriptions/SubscriptionStorage.cs
```diff
@@ -476,10 +476,9 @@ internal void CleanupSubscriptions()
 
                 var recentConnection = state.Value.MostRecentEndedConnection();
 
-                while (recentConnection != null && recentConnection.Stats.LastMessageSentAt < oldestPossibleIdleSubscription)
+                if (recentConnection != null && recentConnection.Stats.LastMessageSentAt < oldestPossibleIdleSubscription)
```
ayende, line 479: this will now only kill a single one? Why would this code cause a hang?

src/Sparrow/Utils/NativeMemory.cs
maximburyak: In addition to _lastName, is there anything else you want to be changed? Becuase this change _must_ happen, without it we have Thread objects leak.
Also, keep in mind that most probably this object won't be accessible if Thread is dead, because it also won't be in the ThreadLocal hodling it.

src/Raven.Server/Documents/Subscriptions/SubscriptionStorage.cs
maximburyak: @ayende we are trying to kill useless SubscriptionConnection objects, not "recent connections".  and recent connections list is limited to 10 anyway. In case of real low memory, a more through cleanup will be performed.
This code _caused_ a hang because recentConnection was never replaced and it didn't try to  remove items from the recent connections list.
The point in this cleanup is to make sure there are no unused subscription objects in the SubscriptionStorage.

ayende (approved):

PR #9582, opened 2019-08-07
RavenDB-13762 session in subscription should not track revisions

ppekrol:


ayende (approved):

src/Raven.Client/Documents/Subscriptions/SubscriptionBatch.cs
```diff
@@ -168,8 +169,11 @@ public SubscriptionBatch(RequestExecutor requestExecutor, IDocumentStore store,
         internal string Initialize(BatchFromServer batch)
         {
             _includes = batch.Includes;
+
             Items.Capacity = Math.Max(Items.Capacity, batch.Messages.Count);
             Items.Clear();
+
+            var revision = typeof(T).IsConstructedGenericType && typeof(T).GetGenericTypeDefinition() == typeof(Revision<>);
```
maximburyak, line 176: I think it won't work in case of T as `dynamic`

ppekrol: can you give an example?

PR #9583, opened 2019-08-08
RavenDB-13873 Exception thrown when querying list bigger than 1024

ayende:


ppekrol (approved):

src/Raven.Server/Documents/Queries/AST/QueryVisitor.cs
```diff
@@ -9,6 +10,8 @@ public abstract class QueryVisitor
 
         public void Visit(Query q)
         {
+            RuntimeHelpers.EnsureSufficientExecutionStack();
```
arekpalinski, line 13: I wonder if we could do better here by using `TryEnsureSufficientExecutionStack ` and throw own exception which would warn about huge number of clauses instead of a bit scary `InsufficientExecutionStackException`

ayende: From my perspective, you have to do something _really_ nasty here, and I'm not sure that we _can_ have proper work done here, given the limited stack space.

arekpalinski: Oh I see. That's fine by me then

PR #9584, opened 2019-08-08
RavenDB-13673-2 bumping Lucene.Net.3.0.20.nupkg => Lucene.Net.3.0.21.…

talweiss1982:
…nupkg + LightWeightSimilarity for FacetedQuery


ayende (approved):
What are the perf numbers?

talweiss1982:
for the LiveHire scenario, 25% of the execution time was idfExplantion but they had a mix of facet and highly nested queries.
When i ran on small input only on facets it reduced the idfExplanation from 68ms to 1ms but the total query time was around 10 sec so its not much in % 
I can run the numbers on the full input if you wish.

PR #9585, opened 2019-08-08
RavenDB-13892 added /admin/stats

ppekrol:


src/Raven.Server/Web/System/AdminStatsHandler.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Threading.Tasks;
+using Raven.Server.Routing;
+using Sparrow.Json;
+
+namespace Raven.Server.Web.System
+{
+    public class AdminStatsHandler : RequestHandler
+    {
+        [RavenAction("/admin/stats", "GET", AuthorizationStatus.Operator)]
```
ppekrol: Not sure if this is best name of the endpoint. Thoughts?

src/Raven.Server/ServerWide/ServerStatistics.cs
```diff
@@ -10,8 +12,31 @@ public ServerStatistics()
...
         public readonly DateTime StartUpTime;
-        
+
         public DateTime? LastRequestTime { get; set; }
+
+        public void WriteTo(BlittableJsonTextWriter writer)
+        {
+            writer.WriteStartObject();
+
+            writer.WritePropertyName(nameof(UpTime));
```
ppekrol, line 25: Do we want to expose this in different format? E.g. in seconds only (separate property)? Note that monitoring should be done via SNMP anyway IMO.

ayende (approved):

ppekrol:
Added 1 more commit. We cannot update LastRequestTime when accessing this endpoint, doesn't make a lot of sense.

PR #9586, opened 2019-08-08
RavenDB-13903: Prevent from query operations (delete and patch) from …

maximburyak:
…blocking on execution. We want to add the operation and let the task to start running without any waiting

### DIFF
--- a/src/Raven.Server/Documents/Handlers/QueriesHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/QueriesHandler.cs
@@ -234,7 +234,7 @@ namespace Raven.Server.Documents.Handlers
             }
         }
 
-        private  async Task Graph(DocumentsOperationContext context, RequestTimeTracker tracker, HttpMethod method)
+        private async Task Graph(DocumentsOperationContext context, RequestTimeTracker tracker, HttpMethod method)
         {
             var indexQuery = await GetIndexQuery(context, method, tracker);
             var queryRunner = Database.QueryRunner.GetRunner(indexQuery);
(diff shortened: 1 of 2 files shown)


ayende (approved):

src/Raven.Server/Documents/Handlers/QueriesHandler.cs
```diff
@@ -529,8 +529,11 @@ private void TrafficWatchQuery(IndexQueryServerSide indexQuery)
                 Query = query.Query
             };
 
-            var task = Database.Operations.AddOperation(Database, indexName, operationType,
-                onProgress => operation(Database.QueryRunner, options, onProgress, token), operationId, details, token);
+            var task =
```
ppekrol: I think that continuation few lines below will be done on a wrong task, so context will be disposed too early. Ping me when available, we have similar places to this in other parts of the system. I want to understand what is the cause and if it should be treat globally or not.

ayende: Pretty sure that `Task.Run(Func<Task>)` will return a task that only completes when the inner task task

ppekrol: Ok, but it does suggest that AddOperation should do that internally for all executions of that method. Not only here? There are 20 usages of this.

ppekrol: I've moved `Task.Run` inside the `AddOperation`. Waiting for CI.

ayende (approved):

PR #9588, opened 2019-08-12
RDBCL-727 CPU credits exec

iftahbe:


src/Raven.Server/Config/Categories/ServerConfiguration.cs
```diff
@@ -56,5 +56,27 @@ public class ServerConfiguration : ConfigurationCategory
...
+        [DefaultValue(null)]
+        [ConfigurationEntry("Server.CpuCredits.Exec", ConfigurationEntryScope.ServerWideOnly)]
+        public string CpuCreditsExec { get; set; }
+
+        [Description("EXPERT: The command line arguments for the Server.CpuCredits.Exec command or executable.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Server.CpuCredits.Exec.Arguments", ConfigurationEntryScope.ServerWideOnly)]
+        public string CpuCreditsExecArguments { get; set; }
+
+        [Description("EXPERT: The number of seconds between every invokation of the Cpu Credits executable. Default: 30 minutes.")]
```
ppekrol: typo:
- invokation  -> invocation 
- Cpu -> CPU

src/Raven.Server/Config/Categories/ServerConfiguration.cs
```diff
@@ -56,5 +56,27 @@ public class ServerConfiguration : ConfigurationCategory
...
+        [DefaultValue(30*60)]
+        [TimeUnit(TimeUnit.Seconds)]
```
ppekrol: Why not in Minutes?

Better to write in description then 1800 seconds to avoid unit change

src/Raven.Server/Config/Categories/ServerConfiguration.cs
```diff
@@ -56,5 +56,27 @@ public class ServerConfiguration : ConfigurationCategory
...
+        [ConfigurationEntry("Server.CpuCredits.Exec.SyncInterval", ConfigurationEntryScope.ServerWideOnly)]
```
ppekrol: SyncIntervalInSec

src/Raven.Server/Config/Categories/ServerConfiguration.cs
```diff
@@ -56,5 +56,27 @@ public class ServerConfiguration : ConfigurationCategory
...
+        public TimeSetting CpuCreditsExecSyncInterval { get; set; }
+
+        [Description("EXPERT: The number of seconds to wait for the Cpu Credits executable to exit. Default: 30 seconds")]
```
ppekrol: typo:
- Cpu -> CPU

src/Raven.Server/RavenServer.cs
```diff
@@ -280,12 +280,29 @@ void ConfigureKestrel(KestrelServerOptions options)
                 if (Configuration.Server.CpuCreditsBase != null ||
                     Configuration.Server.CpuCreditsMax != null ||
                     Configuration.Server.CpuCreditsExhaustionFailoverThreshold != null ||
-                    Configuration.Server.CpuCreditsExhaustionBackgroundTasksThreshold != null)
+                    Configuration.Server.CpuCreditsExhaustionBackgroundTasksThreshold != null ||
+                    Configuration.Server.CpuCreditsExec != null ||
+                    Configuration.Server.CpuCreditsExecArguments != null)
```
ppekrol: Do we have to check arguments?

src/Raven.Server/RavenServer.cs
```diff
@@ -280,12 +280,29 @@ void ConfigureKestrel(KestrelServerOptions options)
...
                     if (Configuration.Server.CpuCreditsBase == null ||
                         Configuration.Server.CpuCreditsMax == null)
                         throw new InvalidOperationException($"Both {RavenConfiguration.GetKey(s=> s.Server.CpuCreditsBase)} and {RavenConfiguration.GetKey(s => s.Server.CpuCreditsMax)} must be specified");
 
+                    if (Configuration.Server.CpuCreditsExec == null)
+                    {
+                        CpuCreditsBalance.RemainingCpuCredits = Configuration.Server.CpuCreditsBase.Value;
+
+                        if (Logger.IsInfoEnabled)
+                            Logger.Info($"Cpu credits were configured but missing the {RavenConfiguration.GetKey(s => s.Server.CpuCreditsExec)} key.");
```
ppekrol: typo:
- Cpu -> CPU

src/Raven.Server/RavenServer.cs
```diff
@@ -379,20 +397,43 @@ public DynamicJsonValue ToJson()
...
             CpuCreditsBalance.Used = true;
-            CpuCreditsBalance.RemainingCpuCredits = CpuCreditsBalance.BaseCredits;
             CpuCreditsBalance.BackgroundTasksThresholdReleaseValue = CpuCreditsBalance.BackgroundTasksThreshold * 1.25;
             CpuCreditsBalance.FailoverThresholdReleaseValue = CpuCreditsBalance.FailoverThreshold * 1.25;
             CpuCreditsBalance.CreditsGainedPerSecond = CpuCreditsBalance.BaseCredits / 3600;
 
             int remainingTimeToBackgroundAlert = 15, remainingTimeToFailvoerAlert = 5;
             AlertRaised backgroundTasksAlert = null, failoverAlert = null;
+
+            if (string.IsNullOrEmpty(Configuration.Server.CpuCreditsExec) == false)
```
ppekrol: Here you are checking string is null or empty, few lines up only for null. Can we check null or empty in both places?

src/Raven.Server/RavenServer.cs
```diff
@@ -379,20 +397,43 @@ public DynamicJsonValue ToJson()
...
+                CpuCreditsBalance.HistoryCurrentIndex = 0;
+            }
+            else
+            {
+                var print = RavenConfiguration.GetKey(x => x.Server.CpuCreditsExec) +
+                            (string.IsNullOrEmpty(Configuration.Server.CpuCreditsExecArguments)
+                                ? " key and the " + RavenConfiguration.GetKey(x => x.Server.CpuCreditsExecArguments)
+                                : "");
+                if (Logger.IsInfoEnabled)
+                    Logger.Operations($"Cpu credits were configured but missing the {print} key.");
```
ppekrol: typo:
- Cpu -> CPU

src/Raven.Server/Config/Categories/ServerConfiguration.cs
iftahbe: Because in the code it is used as seconds. Do you prefer I use minutes here... and in the code do a conversion?

src/Raven.Server/RavenServer.cs
iftahbe: We don't *have* to... but if one of them is defined and the others not, it will give us a change to throw an appropriate error.

src/Raven.Server/Config/Categories/ServerConfiguration.cs
iftahbe: Oh ok... I got what you mean now. You prefer that in the text I'll write "Default: 1800 seconds"

ppekrol (approved):

src/Raven.Server/RavenServer.cs
```diff
@@ -379,20 +396,43 @@ public DynamicJsonValue ToJson()
...
+
+                CpuCreditsBalance.RemainingCpuCredits = response.Remaining;
+                CpuCreditsBalance.LastSyncTime = response.Timestamp;
+
+                CpuCreditsBalance.History = new double[60 * 60];
+                CpuCreditsBalance.HistoryCurrentIndex = 0;
+            }
+            else
+            {
+                var print = RavenConfiguration.GetKey(x => x.Server.CpuCreditsExec) +
```
ayende: Is it mandatory now?

src/Raven.Server/RavenServer.cs
```diff
@@ -441,6 +481,22 @@ private void StartMonitoringCpuCredits()
...
+                        i = 0;
+
+                        if (string.IsNullOrEmpty(Configuration.Server.CpuCreditsExec) == false)
+                        {
+                            var response = GetCpuCreditsFromExec();
+
+                            CpuCreditsBalance.RemainingCpuCredits = response.Remaining;
+                            CpuCreditsBalance.LastSyncTime = response.Timestamp;
+
+                            CpuCreditsBalance.History = new double[60 * 60];
```
ayende: Why create it instead of clearing it?

src/Raven.Server/RavenServer.cs
```diff
@@ -441,6 +481,22 @@ private void StartMonitoringCpuCredits()
 
                 try
                 {
+                    if (i++ == (int)Configuration.Server.CpuCreditsExecSyncInterval.AsTimeSpan.TotalSeconds)
```
ayende: Use `Stopwatch` and check for time, rather than assume that the `Delay` would be seconds

src/Raven.Server/RavenServer.cs
```diff
@@ -486,6 +542,87 @@ private void StartMonitoringCpuCredits()
...
+
+            Process process;
+
+            try
+            {
+                process = Process.Start(startInfo);
+            }
+            catch (Exception e)
+            {
+                throw new InvalidOperationException($"Unable to get cpu credits by executing {command} {arguments}. Failed to start process.", e);
```
ayende, line 575: will this kill the server?

src/Raven.Server/RavenServer.cs
```diff
@@ -486,6 +542,87 @@ private void StartMonitoringCpuCredits()
...
+            }
+
+            using (var ms = new MemoryStream())
```
ayende, line 578: This is a lot of duplicated code.
We know that we want to change this already, so let's centralized process execution

ayende (changes requested):

src/Raven.Server/RavenServer.cs
iftahbe: I think it should be mandatory if you supply other CPU credits configurations.

src/Raven.Server/RavenServer.cs
iftahbe: I assumed since this is all heuristic-based, it doesn't matter if it's going to be 30 minutes or 32 minutes... Does the accuracy matter here?

src/Raven.Server/RavenServer.cs
ayende: so it should throw, not log

src/Raven.Server/RavenServer.cs
iftahbe: Will open an issue for that

src/Raven.Server/RavenServer.cs
ayende: I don't like it because it isn't clear what you are doing.
The accuracy doesn't matter, but to understand the code you have to jump around a bit, and it is better if we said what we meant, rather than inferred it.

src/Raven.Server/RavenServer.cs
iftahbe: Yes, I fixed it to be inside a try-catch with logging.

src/Raven.Server/RavenServer.cs
iftahbe: ok

src/Raven.Server/RavenServer.cs
iftahbe: yes

src/Raven.Server/RavenServer.cs
```diff
@@ -439,6 +451,29 @@ private void StartMonitoringCpuCredits()
...
+                        {
+                            var response = GetCpuCreditsFromExec();
+
+                            CpuCreditsBalance.RemainingCpuCredits = response.Remaining;
+                            CpuCreditsBalance.LastSyncTime = response.Timestamp;
+                            Array.Clear(CpuCreditsBalance.History, 0, CpuCreditsBalance.History.Length);
+                            CpuCreditsBalance.HistoryCurrentIndex = 0;
+                        }
+
+                        sw.Restart();
```
ppekrol, line 466: do this maybe in `finally` so if something goes wrong this will be reset and we will not execute checks too often?

src/Raven.Server/Config/Categories/ServerConfiguration.cs
ppekrol: Let's change this to Minutes

src/Raven.Server/RavenServer.cs
```diff
@@ -439,6 +451,29 @@ private void StartMonitoringCpuCredits()
                         Logger.Operations("Unhandled exception occured during cpu credit monitoring", e);
                 }
 
+                try
+                {
+                    if (sw.Elapsed.TotalSeconds >= (int)Configuration.Server.CpuCreditsExecSyncInterval.AsTimeSpan.TotalSeconds)
```
ppekrol, line 464: use TimeSpan comparison

iftahbe: Why? I don't like the compare functions where you have an integer result. 
Using the >= operator does the exact same thing: https://referencesource.microsoft.com/#mscorlib/system/timespan.cs,164

src/Raven.Server/RavenServer.cs
iftahbe: If it's in the finally we'll always restart it and not get a chance to reach 30 minutes.

Instead, I'm changing the order of execution so that restart will always come before the attempt to get credits so that if we have a failure the stopwatch will still be restarted.

PR #9589, opened 2019-08-12
RavenDB-13053 Sharding infrastructure

ayende:
This is the very first step toward sharding. 
The idea is that we'll define a `Shards` property in the database record, which will define the nodes shards are located on. 
We'll also have a `ShardAllocations` map that will let us know based on shard id on which shard a document is.

I changed routing so we can create mirrors to database routes that will run the sharding logic.
For example, `PUT /databases/MyDb/docs?id=users/1` will compute the shard key from the document id and forward the call to the right shard.

`MyDb` is the sharded db (which will orchestrate the forwarding to the real nodes / databases). The real databases are `MyDB$0`, `MyDB$1`, etc. Note that `$` is not a valid db name by default, so users can't conflict with these.

Document ids are used to shard the documents. We run `XXHash64` on the output and the mod that with 1048576 (1M). The idea is that assuming perfect distribution, we'll have 1MB in each shard key per TB in the total database.
That make is feasible to move shards between nodes.

You can also control what shard a document will be on:  `users/1` and  `orders/321$users/1` will be on the same shard, for example.

Note that we have a few concepts here:

* Shards - 0 ... 1048576 - we have a lot of them. And they can move between nodes
* Fragments - `MyDB$0, `MyDb$1`, etc. - These are the databases that contain the shards. They are normal databases (although they share the same database record). 
* Database - `MyDb` - the actual sharded database, which is a set of routes that hold the sharding logic.

Of particular interest:

* ShardedContext - which contains the logic of sharding / holds request executors for the relevant nodes.
* Security between the nodes is using full server cert, was authenticated when reached the origin. There is no way to specify different security for the shard routes.
* There is no way to have a sharding route that isn't also an existing database route.
* Clients are *not* aware of sharding (the only client side change is the addition of a few field to the database record).
* `DeletionInProgress` can now contain sharded work, such as `A$0`, which indicate that we want to remove the `MyDb$0` from node `A`.

This PR is mostly about all the infrastructure that is required to actually get us going. We'll need to do quite a bit of work around implementing the routes, the logic, etc.
However, the most important bits are there and should allow us to start implementing this without having to do  a big bang integration.

### DIFF
--- a/src/Raven.Client/ServerWide/DatabaseRecord.cs
+++ b/src/Raven.Client/ServerWide/DatabaseRecord.cs
@@ -48,6 +48,15 @@ namespace Raven.Client.ServerWide
 
         public DatabaseTopology Topology;
 
+        public DatabaseTopology[] Shards;
+
+        public List<ShardRangeAssignment> ShardAllocations = new List<ShardRangeAssignment>();
+        public class ShardRangeAssignment
+        {
+            public int RangeStart;
+            public int Shard;
+        }
+
         // public OnGoingTasks tasks;  tasks for this node..
         // list backup.. list sub .. list etl.. list repl(watchers).. list sql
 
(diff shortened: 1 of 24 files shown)


ppekrol:
Closing for now.

PR #9591, opened 2019-08-12
RavenDB-13918 Negation in index cause a double to become int

ayende:


src/Sparrow/Json/LazyNumberValue.cs
```diff
@@ -150,6 +150,15 @@ private static void ThrowInvalidNumberFormat(LazyNumberValue self, string type)
         {
             return x % (decimal)y;
         }
+        public static decimal operator +(LazyNumberValue y)
+        {
+            return -(decimal)y;
```
DejanMilicic: @ayende Shouldn't this be + ?

ayende: Yes

PR #9598, opened 2019-08-15
RavenDB-13098 TimeSeries replication

karmeli87:


src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
```diff
@@ -164,19 +166,32 @@ protected override int ExecuteCmd(DocumentsOperationContext context)
 
                 if (_batch.Appends != null)
                 {
-                    foreach (var append in _batch.Appends)
+                    // TODO: I guess we can do it more efficient
```
ayende: Yes, we can. We can create a dictionary to hold that and add to it, would be much faster

ayende: Remember that in 99% of cases, this is going to be a single one

src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
```diff
@@ -164,19 +166,32 @@ protected override int ExecuteCmd(DocumentsOperationContext context)
...
-                            append.Timestamp,
-                            new Span<double>(append.Values),
-                            append.Tag,
-                            fromReplication: false
+                            group.Name,
+                            group.Appendies
                         );
 
-                        changes++;
+                        changes += group.Appendies.Count();
```
ayende: You are iterating the iterator again here

src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
karmeli87: So would make a sense to have a different code path for each case to avoid allocations?

ayende: I think so

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -675,6 +676,7 @@ public class MergedBatchCommand : TransactionMergedCommand, IDisposable
             public ArraySegment<BatchRequestParser.CommandData> ParsedCommands;
             public Queue<AttachmentStream> AttachmentStreams;
             public StreamsTempFile AttachmentStreamsTempFile;
+            private readonly Dictionary<string, SortedList<long, AppendTimeSeriesOperation>> _appendDictionary;
```
ayende: This can be removed, no? It's always null

src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
```diff
@@ -161,22 +172,46 @@ protected override int ExecuteCmd(DocumentsOperationContext context)
...
                         LastChangeVector = tss.AppendTimestamp(context,
                             _batch.Id,
                             docCollection,
-                            append.Name,
-                            append.Timestamp,
-                            new Span<double>(append.Values),
-                            append.Tag,
-                            fromReplication: false
+                            kvp.Key,
+                            kvp.Value.Values.Select(x =>
```
ayende: For perf, we probably want to do this without the lambda.

src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
```diff
@@ -161,22 +172,46 @@ protected override int ExecuteCmd(DocumentsOperationContext context)
...
+                {
+                    LastChangeVector = tss.AppendTimestamp(context,
+                        _batch.Id,
+                        docCollection,
+                        _singleValue.Name,
+                        Single(_singleValue));
+
+                    changes++;
+
+                    IEnumerable<TimeSeriesStorage.Reader.SingleResult> Single(AppendTimeSeriesOperation element)
```
ayende: I don't really like this approach, can we accept `AppendTimeSeriesOperation` directly and convert internally?

src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
```diff
@@ -136,20 +140,27 @@ public async Task Batch()
...
             private readonly DocumentTimeSeriesOperation _batch;
             private readonly bool _fromEtl;
 
+            private readonly Dictionary<string, SortedList<long, AppendTimeSeriesOperation>> _appendDictionary;
+            private readonly AppendTimeSeriesOperation _singleValue;
+
             public string LastChangeVector;
 
-            public ExecuteTimeSeriesBatchCommand(DocumentDatabase database, DocumentTimeSeriesOperation batch, bool fromEtl)
+            public ExecuteTimeSeriesBatchCommand(DocumentDatabase database, DocumentTimeSeriesOperation batch, bool fromEtl, Dictionary<string, SortedList<long, AppendTimeSeriesOperation>> appendDictionary = null)
```
ayende: `appendDictionary` is always null, can be removed.

src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
```diff
@@ -196,6 +231,34 @@ protected override int ExecuteCmd(DocumentsOperationContext context)
...
+                if (appendDictionary == null)
+                    appendDictionary = new Dictionary<string, SortedList<long, AppendTimeSeriesOperation>>();
+
+                foreach (var item in batch.Appends)
+                {
+                    if (appendDictionary.TryGetValue(item.Name, out var sorted) == false)
+                        sorted = new SortedList<long, AppendTimeSeriesOperation>();
+
+                    sorted[item.Timestamp.Ticks] = item;
+                    appendDictionary[item.Name] = sorted;
```
ayende: Add this in the `if` statement, instead

ayende (changes requested):

ayende (approved):

PR #9612, opened 2019-08-20
RavenDB-13937: Fix SearchForPage skipping values

garayx:


ppekrol (approved):

ayende (approved):

arekpalinski:
@garayx Could you please put some comments explaining the fix? I see two changes which aren't obvious:

1) Instead of `nodePos = (ushort)(p.LastSearchPosition - 1);` we have `p.LastSearchPosition--;`

https://github.com/ravendb/ravendb/pull/9612/files#diff-7fb829b9a1744e81c63ac57d621fd9d3R710

2) We no longer check if `cursorPage.Search(_llt, key) != null`

https://github.com/ravendb/ravendb/pull/9612/files#diff-d4f93dbb7f53a0819794ef3011a2da17R68

It would be nice to have some reasoning here that explains why we changed that so if we get back to this PR in the future we'll understand the changes without the need to actually debug it.

garayx:
We still assign `p.LastSearchPosition - 1` to `nodePos` see:
https://github.com/ravendb/ravendb/pull/9612/files#diff-7fb829b9a1744e81c63ac57d621fd9d3R744
only difference that we assign it with _actually_ decreasing `p.LastSearchPosition` value, which should be decreased regardless the result of `cursorPage.Search()`, the issue that we didn't decrease the `p.LastSearchPosition` value which led the search skip returning the value if it was the last value in the tree.

arekpalinski:
Yes, I see that we still assign to `nodePos` but as you noted, now we decrease `LastSearchPosition` _while_ searching for a given key. I just wanted you to confirm that's really the case.

What about my second point, that we no longer check if `cursorPage.Search(_llt, key) != null` ?

ayende:
That was intentional.
The issue is if I'm making a search on a key that doesn't exist (for iteration), then the `Search` will return null, but we need to ensure that the location of the position is proper, according to whatever the current key is `>` than the one we checked

arekpalinski:
Makes sense. Thanks!

PR #9614, opened 2019-08-21
RavenDB-13921 upgraded Jint

ppekrol:


src/Raven.Server/Documents/Patch/BlittableObjectInstance.cs
```diff
@@ -25,33 +24,32 @@ public class BlittableObjectInstance : ObjectInstance
...
 
         private void MarkChanged()
         {
             Changed = true;
             _parent?.MarkChanged();
-        }        
+        }
 
-        public ObjectInstance GetOrCreate(string key)
+        public ObjectInstance GetOrCreate(Key key)
```
lahma: Could have `in Key key`here

src/Raven.Server/Documents/Patch/ScriptRunnerResult.cs
```diff
@@ -9,26 +9,20 @@ namespace Raven.Server.Documents.Patch
...
             _parent = parent;
-            _instance = instance;
+            Instance = instance;
         }
 
-        public JsValue Instance => _instance;
+        public readonly JsValue Instance;
 
-        public void Set(string name, string value)
+        public ObjectInstance GetOrCreate(Key name)
```
lahma: Could have `in Key name` here as well

src/Raven.Server/Documents/Patch/BlittableObjectInstance.cs
```diff
@@ -25,33 +24,32 @@ public class BlittableObjectInstance : ObjectInstance
         public readonly string ChangeVector;
         public readonly BlittableJsonReaderObject Blittable;
         public readonly string DocumentId;
-        public HashSet<string> Deletes;
-        public Dictionary<string, BlittableObjectProperty> OwnValues = 
-            new Dictionary<string, BlittableObjectProperty>();
+        public HashSet<Key> Deletes;
+        public Dictionary<Key, BlittableObjectProperty> OwnValues = new Dictionary<Key, BlittableObjectProperty>();
```
lahma, line 28: As you anyway have custom build of Jint, there's also possibility to make StringDictionarySlim public and use it. It's basically the [tuned version of Dictionary](https://github.com/dotnet/corefxlab/blob/master/src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/DictionarySlim.cs) ([some history here](https://github.com/dotnet/corefxlab/pull/2458)).

ppekrol: Or maybe it is better to include 'Microsoft.Experimental.Collections' and use `DictionarySlim<TK, TV>` instead of Jint's string version. What are benefits of `StringDictionarySlim<T>` vs `DictionarySlim<string, T>` or `DictionarySlim<Jint.Key, T>` ?

Last one seems like the most reasonable option, because we would have an opportunity to take leverage from `Key`. Thoughts?

lahma: Probably the most convenient path is to use the MS Experimental. Jint version just skips GetHashCode etc and accesses the struct field directly.

PR #9615, opened 2019-08-21
RavenDB-12480 when encountering socket exception we want to get rid of the HttpClient so new one will be created next time

ppekrol:


src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -67,7 +68,19 @@ public class RequestExecutor : IDisposable
...
+
+        public HttpClient HttpClient
+        {
+            get
+            {
+                var httpClient = _httpClient;
+                if (httpClient != null)
+                    return httpClient;
+
+                return _httpClient = GetHttpClient();
```
karmeli87, line 86: Maybe it is overkill, but does it have to be more thread-safe?

```
var newClient = GetHttpClient();
var lastKnownClient = Interlocked.CompareExchange(ref _httpClient, newClient , httpClient );
if(  lastKnownClient == httpClient )
    return newClient ;

return lastKnownClient;
```

ayende: That would make you do a cross thread operation each time, why?

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -777,22 +830,38 @@ protected virtual bool TryLoadFromCache(JsonOperationContext context)
                 }
                 catch (HttpRequestException e) // server down, network down
                 {
-                    if (shouldRetry == false)
-                        throw;
+                    using (GetContext(out var requestContext))
+                    {
+                        if (e.InnerException is SocketException se && ShouldRemoveHttpClient(se))
```
ayende, line 835: Also, probably for proxy errors, no?

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -777,22 +830,38 @@ protected virtual bool TryLoadFromCache(JsonOperationContext context)
...
+                        {
+                            if (requestContext.HttpClientRemoved == false)
+                            {
+                                RemoveHttpClient();
+                                requestContext.HttpClientRemoved = true;
+                            }
+                            else
+                            {
+                                requestContext.HttpClientRemoved = false;
```
ayende, line 844: why are you setting this to the value it is already set to?

src/Raven.Client/Http/RequestExecutor.cs
ppekrol: I was getting `SocketError.ConnectionRefused` in that case (handled). Do you know if there is anything else?

src/Raven.Client/Http/RequestExecutor.cs
ppekrol: I'm setting this from `true` to `false` because I've already tried that once and I now want `HandleServerDown` to behave correctly (return `false` if there is no other node to check and throw appropriate exception)

src/Raven.Client/Http/RequestExecutor.cs
ayende: Consider the case where the proxy is alive, but returning errors to us. 
Proxy auth failure, for example.

ppekrol: Not sure that I entirely follow. I should retry on 407?

ayende: Yes

ppekrol: I will handle 407 in a separate PR

PR #9619, opened 2019-08-22
RavenDB-10177 allow to enforce revision configuration

karmeli87:


ayende (approved):

src/Raven.Client/Documents/Operations/Revisions/ConfigureRevisionsOperation.cs
```diff
@@ -13,24 +13,28 @@ namespace Raven.Client.Documents.Operations.Revisions
     public class ConfigureRevisionsOperation : IMaintenanceOperation<ConfigureRevisionsOperationResult>
     {
         private readonly RevisionsConfiguration _configuration;
+        private readonly bool _enforce;
 
-        public ConfigureRevisionsOperation(RevisionsConfiguration configuration)
+        public ConfigureRevisionsOperation(RevisionsConfiguration configuration, bool enforce = false)
```
ppekrol: This breaks binary compatibility. You need a separate 'ctor'.

src/Raven.Server/Documents/Handlers/RevisionsHandler.cs
```diff
@@ -90,6 +92,15 @@ public async Task ConfigRevisions()
                         }
                     }
                 });
+
+            if (enforce)
```
ppekrol: I do not like this. I think it would be better to remove this and just have a separate endpoint to do that.

src/Raven.Server/Documents/Handlers/RevisionsHandler.cs
```diff
@@ -90,6 +92,15 @@ public async Task ConfigRevisions()
...
+                await EnforceConfigRevisions();
+        }
+
+        [RavenAction("/databases/*/admin/revisions/config/enforce", "POST", AuthorizationStatus.DatabaseAdmin)]
+        public async Task EnforceConfigRevisions()
+        {
+            await Database.DocumentsStorage.RevisionsStorage.EnforceConfiguration();
```
ppekrol: This can take a long time. We should wrap this in an operation, return OperationId and create new `EnforceRevisionsConfigurationOperation` in Client API

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -1072,10 +1072,133 @@ public Document GetRevisionBefore(DocumentsOperationContext context, string id,
...
+                        {
+                            var state = ShouldProcessNextRevisionId(ctx, ref tvr.Reader, parameters, result, out var id);
+                            if (state == NextRevisionIdResult.Break)
+                                break;
+                            if (state == NextRevisionIdResult.Continue)
+                                continue;
+
+                            ids.Add(id);
+
+                            if (ids.Count > 1024 || ctx.AllocatedMemory > SizeLimit)
```
ppekrol, line 1116: Shouldn't we use 2 contexts like in RevertRevisions?

ppekrol:
Probably we need to add button to do that to Studio.

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
karmeli87: Unlike the revert revisions we don't need to keep the entire revision, we need only the id.

src/Raven.Server/Documents/Handlers/RevisionsHandler.cs
karmeli87: removed

src/Raven.Client/Documents/Operations/Revisions/ConfigureRevisionsOperation.cs
karmeli87: done

ppekrol (approved):

ppekrol:
@ml054 can you check failing test?

PR #9620, opened 2019-08-22
Raven db 13902

talweiss1982:
I haven't finished this task and the branch I pull from @ayende had some modification that is unrelated to this PR, it seems some of them slipt through.

I left TODOs where I didn't complete 
We need to debug and test the spawn.c POSIX version, it is just in a compiled state.

RavenProcess doesn't support event raising events.

We push both STDOUT and STDERR into STDOUT


ppekrol:
2nd part will be handled by @aviviadi

PR #9623, opened 2019-08-23
RavenDB-13077 removed SNMP warning after license activation

ppekrol:


src/Raven.Server/NotificationCenter/Notifications/Details/LicenseLimitWarning.cs
```diff
@@ -38,9 +38,15 @@ public static void AddLicenseLimitNotification(NotificationCenter notificationCe
                 licenseLimit.Message,
                 AlertType.LicenseManager_LicenseLimit,
                 NotificationSeverity.Warning,
+                key: licenseLimit.Type.ToString(),
                 details: new LicenseLimitWarning(licenseLimit));
 
-            notificationCenter.Add(alert);
+            notificationCenter.Add(alert, updateExisting: false);
```
ppekrol: I've set `updateExisting` to `false` to avoid opening a write tx each time. The only 'problem' is that the 'created' date will not be changed. Not sure if this is something that we should worry about? @grisha-kotler

src/Raven.Server/NotificationCenter/Notifications/Details/LicenseLimitWarning.cs
```diff
@@ -38,9 +38,15 @@ public static void AddLicenseLimitNotification(NotificationCenter notificationCe
                 licenseLimit.Message,
                 AlertType.LicenseManager_LicenseLimit,
                 NotificationSeverity.Warning,
+                key: licenseLimit.Type.ToString(),
```
ppekrol, line 41: without this the `Id` is just a `LicenseManager_LicenseLimit` (@grisha-kotler please double check). Which means that all license limits alerts will be under the same Id, which is not what we want?

src/Raven.Server/NotificationCenter/Notifications/Details/LicenseLimitWarning.cs
grisha-kotler: It should be very rare (when you start your server)

grisha-kotler (approved):

PR #9626, opened 2019-08-25
RavenDB-13963 Specific document with attachment insertion pattern cau…

karmeli87:
…se replication of many items in a single batch

### DIFF
--- a/src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
+++ b/src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
@@ -226,12 +226,12 @@ namespace Raven.Server.Documents.Replication
                                     AssertNotClusterTransactionDocumentForLegacyReplication(item);
                                 }
 
-                                // Include the attachment's document which is right after its latest attachment.
-                                if ((item.Type == ReplicationBatchItem.ReplicationItemType.Document ||
-                                     item.Type == ReplicationBatchItem.ReplicationItemType.DocumentTombstone) &&
-                                    // We want to limit batch sizes to reasonable limits.
-                                    ((maxSizeToSend.HasValue && size + documentsContext.Transaction.InnerTransaction.LowLevelTransaction.TotalEncryptionBufferSize.GetValue(SizeUnit.Bytes) > maxSizeToSend.Value.GetValue(SizeUnit.Bytes)) ||
-                                     (batchSize.HasValue && numberOfItemsSent > batchSize.Value)))
+                                // We want to limit batch sizes to reasonable limits.
+                                var totalSize =
+                                    size + documentsContext.Transaction.InnerTransaction.LowLevelTransaction.TotalEncryptionBufferSize.GetValue(SizeUnit.Bytes);
+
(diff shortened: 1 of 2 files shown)


grisha-kotler (approved):

ppekrol (approved):

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
```diff
@@ -226,12 +226,12 @@ public bool ExecuteReplicationOnce(OutgoingReplicationStatsScope stats, ref long
...
 
-                                // Include the attachment's document which is right after its latest attachment.
-                                if ((item.Type == ReplicationBatchItem.ReplicationItemType.Document ||
-                                     item.Type == ReplicationBatchItem.ReplicationItemType.DocumentTombstone) &&
-                                    // We want to limit batch sizes to reasonable limits.
-                                    ((maxSizeToSend.HasValue && size + documentsContext.Transaction.InnerTransaction.LowLevelTransaction.TotalEncryptionBufferSize.GetValue(SizeUnit.Bytes) > maxSizeToSend.Value.GetValue(SizeUnit.Bytes)) ||
-                                     (batchSize.HasValue && numberOfItemsSent > batchSize.Value)))
+                                // We want to limit batch sizes to reasonable limits.
+                                var totalSize =
+                                    size + documentsContext.Transaction.InnerTransaction.LowLevelTransaction.TotalEncryptionBufferSize.GetValue(SizeUnit.Bytes);
```
ayende, line 231: `TotalEncryptionBufferSize` that applies only to encypted dbs, does that work properly on non encrypted ones?
Also, where is the actual change here? I'm not seeing it?

grisha-kotler: `maxSizeToSend.HasValue && totalSize > maxSizeToSend.Value.GetValue(SizeUnit.Bytes) ||
                                     (batchSize.HasValue && numberOfItemsSent > batchSize.Value)))	                                    batchSize.HasValue && numberOfItemsSent > batchSize.Value`
was removed

grisha-kotler: The reason for this condition was that we wanted to send all the attachments with its document.

ayende: We still do, otherwise, the other side may see things that aren't complete.

karmeli87: If not encrypted this will return 0.

The actual change is removing those lines:
```
item.Type == ReplicationBatchItem.ReplicationItemType.Document ||
item.Type == ReplicationBatchItem.ReplicationItemType.DocumentTombstone
```

If we had: 
tx 1 -> doc1 + attachment1
tx 2 -> doc2 + attachment2

and the first tx exceeded the size limit we didn't break the batch, because we fetch the attachment before the document and the condition above will be `false`.

ayende (approved):

PR #9627, opened 2019-08-26
Raven db 13796

maximburyak:
This PR was initially meant to investigate a failing test (RachisTests.SubscriptionsFailover.SetMentorToSubscriptionWithFailover), but finished up resolving a race condition between a request executor's topology update and a command's failover

### DIFF
--- a/src/Raven.Client/Http/NodeSelector.cs
+++ b/src/Raven.Client/Http/NodeSelector.cs
@@ -105,7 +105,7 @@ namespace Raven.Client.Http
             return UnlikelyEveryoneFaultedChoice(state);
         }
 
-        public (int Index, ServerNode Node, long TopologyEtag) GetPreferredNodeWithTopology()
+        internal (int Index, ServerNode Node, long TopologyEtag) GetPreferredNodeWithTopology()
         {
             var state = _state;
             var preferredNode = GetPrefferedNodeInternal(state);
--- a/src/Raven.Client/Http/NodeSelector.cs
+++ b/src/Raven.Client/Http/NodeSelector.cs
@@ -100,6 +100,23 @@ namespace Raven.Client.Http
             return UnlikelyEveryoneFaultedChoice(state);
         }
 
+        public (int Index, ServerNode Node, long TopologyEtag) GetPreferredNodeWithTopology()
+        {
+            var state = _state;
+            var stateFailures = state.Failures;
+            var serverNodes = state.Nodes;
+            var len = Math.Min(serverNodes.Count, stateFailures.Length);
+            for (int i = 0; i < len; i++)
+            {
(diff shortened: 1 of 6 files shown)


src/Raven.Client/Http/NodeSelector.cs
```diff
@@ -100,6 +100,23 @@ public bool OnUpdateTopology(Topology topology, bool forceUpdate = false)
             return UnlikelyEveryoneFaultedChoice(state);
         }
 
+        public (int Index, ServerNode Node, long TopologyEtag) GetPreferredNodeWithTopology()
```
ayende: Isn't this duplicating the code in `GetPreferredNode`? Can we merge these?

src/Raven.Client/Http/RavenCommand.cs
```diff
@@ -51,6 +51,8 @@ public abstract class RavenCommand<TResult>
         public bool CanCacheAggressively { get; protected set; }
         public string SelectedNodeTag { get; protected set; }
 
+        public long InitialTopologyEtag = -1;
```
ppekrol: internal

maximburyak:
*Looking into failed tests

src/Raven.Client/Http/NodeSelector.cs
maximburyak: done

src/Raven.Client/Http/RavenCommand.cs
```diff
@@ -51,6 +51,8 @@ public abstract class RavenCommand<TResult>
         public bool CanCacheAggressively { get; protected set; }
         public string SelectedNodeTag { get; protected set; }
 
+        internal long FailoverTopologyEtag = -2;
```
ppekrol, line 54: maybe make it nullable?

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -647,6 +647,10 @@ protected virtual bool TryLoadFromCache(JsonOperationContext context)
             SessionInfo sessionInfo = null,
             CancellationToken token = default)
         {
+            if (command.FailoverTopologyEtag == -2)
+            {                
+                command.FailoverTopologyEtag = _nodeSelector?.GetPreferredNodeWithTopology().TopologyEtag ?? -2;
```
ppekrol: isn't this the same as `RequestExecutor.TopologyEtag` property?

ayende: That is not the same, no. Because the topology may change *while* the command is running

ppekrol: The source of it is the same. `NodeSelector.Topology` which is taken from `State` inside the NodeSelector.

ppekrol (approved):

PR #9641, opened 2019-09-02
RavenDB-13929 Create server-wide-backup icon for Database-Group-Topol…

ml054:
…ogy Graph


src/Raven.Studio/typescript/models/database/dbGroup/databaseGroupGraph.ts
```diff
@@ -99,6 +99,8 @@ interface taskNodeWithCache extends taskNode {
 
 class databaseGroupGraph {
 
+    private static serverWideNamePrefixFromServer = "Server Wide Backup";
```
ppekrol: Let's unify this and put this in one place. IIRC we have it in 2 places now. Too easy to desync.

ml054: done

PR #9648, opened 2019-09-03
RavenDB-13961 setting RemainingCpuCredits to 0 on init, because we are getting real value immediately afterwards and if this values it is better to start with 0 than with base value

ppekrol:


ayende (commented):
I wonder if we shouldn't make this 1, or something else really small.
Maybe 1 over the threshold? To avoid immediately going into throttling?

ppekrol:
Did `CpuCreditsBalance.RemainingCpuCredits = Math.Max(CpuCreditsBalance.BackgroundTasksThreshold, CpuCreditsBalance.FailoverThreshold) + 1;`

ayende (approved):

PR #9650, opened 2019-09-04
RavenDB-13983: Disponse of bjra should not dispose parent

garayx:


src/Sparrow/Json/BlittableJsonReaderObject.cs
```diff
@@ -822,7 +822,10 @@ private object GetObjectUnlikely(BlittableJsonToken type, int position, Blittabl
                 case BlittableJsonToken.EmbeddedBlittable:
                     return ReadNestedObject(position);
                 case BlittableJsonToken.StartArray:
-                    return new BlittableJsonReaderArray(position, _parent ?? this, type)
+                    return new BlittableJsonReaderArray(position, _parent ?? this, type,
```
ppekrol: I do not think we want such a behavior? All child objects are working on the same memory fragment and it was always an assumption that clone needs to be done when you want to 'detach' child object from parent.

ayende (approved):

src/Sparrow/Json/BlittableJsonReaderArray.cs
```diff
@@ -104,7 +107,11 @@ public BlittableJsonReaderArray Clone(JsonOperationContext context, BlittableJso
         public void Dispose()
         {
             AssertContextNotDisposed();
-            _parent?.Dispose();
+
+            // this is required only in cases that we get a BlittableJsonReaderArray, which is an only child of an BlittableJsonReaderObject and we lose track of it's parent, 
+            // like in BlittableJsonDocumentBuilder.CreateArrayReader.
+            if (_disposeParent)
+                _parent?.Dispose();
```
arekpalinski, line 114: This introduces almost totally opposite behavior to the current one. Instead of always disposing the parent, we do it _only_ in one particular case. Maybe that's ok, I'm just a bit concerned about implication and potential memory issues.

Can you explain, please?

ayende: The reason this feature existed is to handle the `CreateArrayReader`, where we don't *have* access to the parent blittable.
Otherwise, there is no point in trying to dispose a blittable that is an inner value.

arekpalinski (approved):

PR #9654, opened 2019-09-05
RavenDB-13940 Adding a configuration option that will allow skipping …

arekpalinski:
…corrupted transactions in journal if they were already synced

### DIFF
--- a/src/Raven.Server/Config/Categories/StorageConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/StorageConfiguration.cs
@@ -113,5 +113,10 @@ namespace Raven.Server.Config.Categories
         [DefaultValue(false)]
         [ConfigurationEntry("Storage.Dangerous.SkipChecksumValidationOnDatabaseLoading", ConfigurationEntryScope.ServerWideOnly)]
         public bool SkipChecksumValidationOnDatabaseLoading { get; set; }
+
+        [Description("EXPERT: Allows to load a database regardless encountered data integrity errors of already synced transactions in journals during the recovery operation on startup")]
+        [DefaultValue(false)]
+        [ConfigurationEntry("Storage.IgnoreDataIntegrityErrorsOfAlreadySyncedTransactions", ConfigurationEntryScope.ServerWideOnly)]
+        public bool IgnoreDataIntegrityErrorsOfAlreadySyncedTransactions { get; set; }
     }
 }
--- a/src/Raven.Server/Config/Categories/StorageConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/StorageConfiguration.cs
@@ -115,7 +115,7 @@ namespace Raven.Server.Config.Categories
         public bool SkipChecksumValidationOnDatabaseLoading { get; set; }
 
         [Description("EXPERT: Allows to load a database regardless encountered data integrity errors of already synced transactions in journals during the recovery operation on startup")]
-        [DefaultValue(false)]
+        [DefaultValue(true)]
         [ConfigurationEntry("Storage.IgnoreDataIntegrityErrorsOfAlreadySyncedTransactions", ConfigurationEntryScope.ServerWideOnly)]
         public bool IgnoreDataIntegrityErrorsOfAlreadySyncedTransactions { get; set; }
     }
--- a/src/Voron/Impl/Journal/JournalReader.cs
+++ b/src/Voron/Impl/Journal/JournalReader.cs
@@ -396,7 +396,14 @@ namespace Voron.Impl.Journal
                 catch (InvalidOperationException ex)
                 {
                     if (CanIgnoreDataIntegrityErrorBecauseTxWasSynced(current, options))
+                    {
+                        options.InvokeIntegrityErrorOfAlreadySyncedData(this,
+                            $"Unable to decrypt data of transaction which has been already synced (tx id: {current->TransactionId}, last synced tx: {_journalInfo.LastSyncedTransactionId}, journal: {_journalInfo.CurrentJournal}). " +
+                            "Safely continuing the startup recovery process.",
+                            ex);
+
                         return true;
+                    }
 
                     RequireHeaderUpdate = true;
                     options.InvokeRecoveryError(this, "Transaction " + current->TransactionId + " was not committed", ex);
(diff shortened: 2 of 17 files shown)


src/Raven.Server/Config/Categories/StorageConfiguration.cs
```diff
@@ -113,5 +113,10 @@ public class StorageConfiguration : ConfigurationCategory
         [DefaultValue(false)]
         [ConfigurationEntry("Storage.Dangerous.SkipChecksumValidationOnDatabaseLoading", ConfigurationEntryScope.ServerWideOnly)]
         public bool SkipChecksumValidationOnDatabaseLoading { get; set; }
+
+        [Description("EXPERT: Allows to load a database regardless encountered data integrity errors of already synced transactions in journals during the recovery operation on startup")]
+        [DefaultValue(false)]
```
ayende: Shouldn't we always ignore this? Why make this optional?

src/Voron/Impl/Journal/JournalReader.cs
```diff
@@ -54,15 +54,21 @@ public bool ReadOneTransactionToDataFile(StorageEnvironmentOptions options)
...
+            if (TryReadAndValidateHeader(options, out TransactionHeader* current) == false)
             {
                 var lastValid4Kb = _readAt4Kb;
                 _readAt4Kb++;
+
                 while (_readAt4Kb < _journalPagerNumberOfAllocated4Kb)
                 {
                     if (TryReadAndValidateHeader(options, out current))
                     {
+                        if (CanIgnoreDataIntegrityErrorBecauseTxWasSynced(current, options))
```
ayende, line 68: Here we need to see if the _next_ transaction is valid, (including the id), and only then agree to jump to it.
Because if the current tx is bad, we can't trust its length either

src/Voron/Impl/Journal/JournalReader.cs
```diff
@@ -427,18 +471,34 @@ private bool TryReadAndValidateHeader(StorageEnvironmentOptions options, out Tra
...
 
             return true;
         }
 
+        private bool CanIgnoreDataIntegrityErrorBecauseTxWasSynced(TransactionHeader* currentTx, StorageEnvironmentOptions options)
+        {
+            // if we have a journal which contains transactions that has been synced and this is the case for current transaction 
+            // then we can continue the recovery regardless encountered errors
+
+            return IsAlreadySyncTransaction(currentTx) && options.IgnoreDataIntegrityErrorsOfAlreadySyncedTransactions;
```
ayende: Here, need to (carefully) check that the _next_ tx is valid (proper id and checksum)

src/Raven.Server/Config/Categories/StorageConfiguration.cs
arekpalinski: I was thinking about enabling it by default, but I don't like to silently ignore this kind of errors, even if that's safe. I prefer a user to make aware of such error. 

I propose that if we detect that a corrupted transaction was corrupted then we'll be thowing still, but we'll include to the exception message info it can be potentially fixed (without any data loss) by enabling this option. What do you think?

src/Voron/Impl/Journal/JournalReader.cs
arekpalinski: Note that here `TryReadAndValidateHeader` returned `true` so we already found a valid transaction. `CanIgnoreDataIntegrityErrorBecauseTxWasSynced` call was to check that we're _still_ in this special mode (corrupted but already synced) and can continue the recovery.

src/Voron/Impl/Journal/JournalReader.cs
arekpalinski: Why? Next transaction can also be corrupted and that's fine I think. Then we're looking for next tx valid header (where you already commented - see my answer there):

https://github.com/ravendb/ravendb/pull/9654/files/2dd60ebbde8ee0da60f0c05559e906da9bd280c6#diff-500338b5995fcc8a61f42b513e64f35eR64-R66

and if we find it and it's _still_ a transaction which was synced then we can skip multiple corrupted transactions.

arekpalinski: For example, we do the recovery, last sync tx is 5:

tx 1 - ok
tx 2 - corrupted
tx 3 - corrupted
tx 4 - valid
tx 5 - valid

We found tx 2 is corrupted then we start looking for next valid transaction. We found tx 4 is valid and was synced so we can safely skip both: tx 2 and tx 3

ppekrol (approved):
LGTM

src/Voron/Impl/Journal/JournalReader.cs
arekpalinski: Added TODO comment for making sure the following won't pass: https://issues.hibernatingrhinos.com/issue/RavenDB-13940#focus=streamItem-67-40594.0-0

src/Raven.Server/Config/Categories/StorageConfiguration.cs
ayende: We saw how that turned out with the missing journal. Even with the actual fix in the message , we got a lot of support calls and scared users.
Given that we *know* that this is an okay scenario, no data loss, why force it on the user?

src/Voron/Impl/Journal/JournalReader.cs
ayende: ok

src/Voron/Impl/Journal/JournalReader.cs
ayende: Okay, make sense

src/Raven.Server/Config/Categories/StorageConfiguration.cs
arekpalinski: I think this case is slightly different because I believe that in this scenario users might be running into this due to hardware problem. So we need to let them know about that even if it's safe to continue the recovery operation.

I don't want to do it completely silently. I think this will be very rare BTW

ayende: So let's make this an alert by default? I agree that we want the user to be aware of this.

arekpalinski: Okay. Let's alert them at least.

arekpalinski: Added in https://github.com/ravendb/ravendb/pull/9654/commits/6eb47e553813fbabcb6ccaadf0cdca75d266cbac

src/Voron/Impl/Journal/JournalReader.cs
arekpalinski: Test added in https://github.com/ravendb/ravendb/pull/9654/commits/6eb47e553813fbabcb6ccaadf0cdca75d266cbac

ppekrol (approved):

ayende (approved):

PR #9656, opened 2019-09-05
RavenDB-13982

maximburyak:


src/Voron/Impl/Paging/Windows32BitsMemoryMapPager.cs
```diff
@@ -241,12 +241,15 @@ public override bool ShouldLockMemoryAtPagerLevel()
...
-            {
+            {                
                 if (Sodium.sodium_mlock(address, (UIntPtr)sizeToLock) != 0)
                 {
-                    if (DoNotConsiderMemoryLockFailureAsCatastrophicError == false)
-                    {
-                        TryHandleFailureToLockMemory(address, sizeToLock, state.TotalLoadedSize * 2);
+                    lock (WorkingSetIncreaseLocker)
+                    { 
+                        if (Sodium.sodium_mlock(address, (UIntPtr)sizeToLock) != 0 && DoNotConsiderMemoryLockFailureAsCatastrophicError == false)
```
ayende, line 249: Why is the `DoNotConsiderMemoryLockFailureAsCatastrophicError ` check inside the lock?

src/Voron/Impl/Paging/AbstractPager.cs
```diff
@@ -54,6 +55,7 @@ protected AbstractPager()
             CanPrefetch = new Lazy<bool>(CanPrefetchQuery);
         }
 
+        public static object WorkingSetIncreaseLocker = new object();
```
ayende: does it have to be public?

src/Voron/Impl/Paging/Windows32BitsMemoryMapPager.cs
ppekrol: This is inside lock because it applies to the result of `sodium_mlock` from the `if` here

ppekrol: done

src/Voron/Impl/Paging/AbstractPager.cs
ppekrol: done

src/Voron/Impl/Paging/Windows32BitsMemoryMapPager.cs
ayende: okay

ayende (approved):

PR #9661, opened 2019-09-06
RavenDB-14008 need to use the same value as in Server Dashboard in Studio

ppekrol:


src/Raven.Server/Monitoring/Snmp/Objects/Server/1.6/ServerTotalMemory.cs
```diff
@@ -1,19 +1,33 @@
...
+using Sparrow;
+using Sparrow.LowMemory;
+using Sparrow.Platform;
+using Sparrow.Platform.Posix;
 
 namespace Raven.Server.Monitoring.Snmp.Objects.Server
 {
     public class ServerTotalMemory : ScalarObjectBase<Gauge32>
     {
+        private readonly Lazy<SmapsReader> _smapsReader;
```
ayende, line 12: Is this thread safe?

ppekrol: Very unlikely to have an issue here but... https://github.com/ravendb/ravendb/pull/9663

PR #9675, opened 2019-09-08
RavenDB-13902: linux vfork implementation, and initial RavenProcess c…

aviviadi:
…lass impl


aviviadi:
This is draft PR (RavenServer.cs going to have different impl)
The PAL code is for review though

src/Raven.Pal/src/posix/linuxonly.c
```diff
@@ -108,5 +108,14 @@ rvn_test_storage_durability(
     return rc;
 }
 
+PRIVATE int32_t
+rvn_pipe_for_child(int descriptors[])
```
ayende:
```suggestion
rvn_pipe_for_child(int descriptors[2])
```

Or `int * descriptors` ?

src/Raven.Pal/src/posix/maconly.c
```diff
@@ -68,4 +69,19 @@ rvn_test_storage_durability(
     return SUCCESS; /* windows and mac are always true */
 }
 
+PRIVATE int32_t
+rvn_pipe_for_child(int descriptors[])
```
ayende:
```suggestion
rvn_pipe_for_child(int descriptors[2])
```

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,291 @@
...
+parse_cmd_line(char* line)
+{
+    if (line == NULL || *line == '\0')
+        return NULL;
+
+    char** argv = malloc(strlen(line)*2);
+    char * pch;
+    int argc = 1;    
+    
+    pch = strtok (line," \t\n");
```
ayende: This mutates the `line`, is that expected?

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,291 @@
...
+#define MSEC_TO_SEC(x) (x/1000)
+#define MSEC_TO_NANOSEC(x) ((x % 1000) * 1000000)
+
+PRIVATE char** 
+parse_cmd_line(char* line)
+{
+    if (line == NULL || *line == '\0')
+        return NULL;
+
+    char** argv = malloc(strlen(line)*2);
```
ayende: We don't actually need this size, do we? Let's default to `malloc(sizeof(char**) * 8)` and `realloc` as needed?

src/Raven.Pal/src/posix/spawn.c
ayende: Can we use `strtok_r`, instead?

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,291 @@
...
+        else
+        {
+            return -1;
+        }
+    }
+    return rv;
+}
+
+PRIVATE void 
+CloseIfOpen(int fd)
```
ayende: naming convention

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,291 @@
...
+{
+    if (fd >= 0)    
+        close(fd);    
+}
+
+PRIVATE ssize_t WriteSize(int fd, const void* buffer, size_t count)
```
ayende: naming convention

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,291 @@
...
+{
+    ssize_t rv = 0;
+    while (count > 0)
+    {
+        ssize_t result = write(fd, buffer, count);
+        while ( result  < 0 && errno == EINTR)
```
ayende: can we make this one liner, like in the pipe open?

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,291 @@
...
+        else
+        {
+            return -1;
+        }
+    }
+    return rv;
+}
+
+PRIVATE void 
+ExitChild(int pipeToParent, int error)
```
ayende: naming convention

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,291 @@
...
+EXPORT int32_t
+rvn_spawn_process(const char* filename, char* cmdline, void** pid, void** standard_in, void** standard_out, int32_t* detailed_error_code) {    
+    int stdinFds[2] = {-1, -1}, stdoutFds[2] = {-1, -1}, waitForChildToExecPipe[2] = {-1, -1};
+    pid_t processId = -1;
+    int thread_cancel_state;
+    sigset_t signal_set;
+    sigset_t old_signal_set;
+
+    char **argv = parse_cmd_line(cmdline);
+    argv[0] = (char *)filename;
```
ayende: I don't like this, better to pass `file` to `parse_cmd_line` and have it set it there.

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,291 @@
...
+                void (*oldhandler)(int) = (sa_old.sa_flags & SA_SIGINFO) ? (void (*)(int))sa_old.sa_sigaction : sa_old.sa_handler;
+                if (oldhandler != SIG_IGN && oldhandler != SIG_DFL)
+                    sigaction(sig, &sa_default, NULL);
+            }
+        }
+        pthread_sigmask(SIG_SETMASK, &old_signal_set, &junk_signal_set); 
+        
+        /* redirect both stdout and stderr to the same stream, and redirect stdin: */
+        result = dup2(stdoutFds[1], STDOUT_FILENO);
+        while ( result < 0 && errno == EINTR)
```
ayende: can use one line here?

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,291 @@
...
+
+    if (rc != SUCCESS)
+    {
+        CloseIfOpen(stdinFds[1]);
+        CloseIfOpen(stdoutFds[0]);
+        
+        if (processId > 0)
+        {
+            int status;
+            waitpid(processId, &status, 0);
```
ayende, line 235: can this cause us to hang?

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,291 @@
...
+    sigemptyset(&child_mask);
+    sigaddset(&child_mask, SIGCHLD);
+    time_t timeout_sec = MSEC_TO_SEC(timeout_ms);
+    
+    int status;
+
+    time_t start = time(NULL);
+
+    struct timespec houndrad_ms;
+    houndrad_ms.tv_sec = 0;
```
ayende: why is this zero here?

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,291 @@
...
+    houndrad_ms.tv_nsec = MSEC_TO_NANOSEC(timeout_ms);
```
ayende: If the value is 1000 ms, then this would be zero too, no?

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,291 @@
...
+    sigset_t child_mask;
+    sigemptyset(&child_mask);
+    sigaddset(&child_mask, SIGCHLD);
+    time_t timeout_sec = MSEC_TO_SEC(timeout_ms);
+    
+    int status;
+
+    time_t start = time(NULL);
+
+    struct timespec houndrad_ms;
```
ayende: what is the meaning of the variable name?

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,291 @@
...
+    houndrad_ms.tv_sec = 0;
+    houndrad_ms.tv_nsec = MSEC_TO_NANOSEC(timeout_ms);
+
+    struct timespec remaining;
+    /*busy wait for proc to end, can't use 'sigtimedwait' on OS X*/
```
ayende: Why are we setting up signals above?

src/Raven.Pal/src/win/spawn.c
```diff
@@ -0,0 +1,145 @@
...
+#include "status_codes.h"
+#include <stdio.h>
+
+
+
+EXPORT int32_t
+rvn_kill_process(void* pid, int32_t* detailed_error_code) {
+    if (!TerminateProcess(pid, 9)) {
+        *detailed_error_code = GetLastError();
+        return FAIL_KILL;
```
ayende: leaking the handle here, no?

src/Raven.Pal/src/win/spawn.c
```diff
@@ -0,0 +1,145 @@
...
+    DWORD rc = WaitForSingleObject(pid, timeout_ms);
+    
+    if (rc == WAIT_TIMEOUT)
+    {
+        *detailed_error_code = GetLastError();
+        return FAIL_TIMEOUT;
+    }
+
+    if (!GetExitCodeProcess(pid, exit_code)) {
+        *detailed_error_code = GetLastError();
```
ayende, line 33: leaking the handle here, no?

src/Raven.Pal/src/win/spawn.c
```diff
@@ -0,0 +1,145 @@
...
+
+success:
+    if (stdin_read != INVALID_HANDLE_VALUE)
+        CloseHandle(stdin_read);
+    if (stdout_write != INVALID_HANDLE_VALUE)
+        CloseHandle(stdout_write);
+
+    if(line != NULL)
+        free(line);
+    if (pi.hThread != INVALID_HANDLE_VALUE)
```
ayende, line 143: If we had an error before `CreateProcess`, then `pi` is zeroed, and `hThread` would be zero, so we'll try to close handle 0?
Let's set it explicitly above, before any failure can happen

ayende (changes requested):
Didn't look at the C# code

src/Raven.Pal/src/posix/spawn.c
aviviadi: strtok_r is not thread safe. If someone passes the same string twice (tests?) then we should avoid the _r func.

ayende: That is the other way around, actually

aviviadi:
comments fixed in https://github.com/ravendb/ravendb/pull/9723

PR #9680, opened 2019-09-10
RavenDB-13978 - fix a race condition where we could end up repeating the same Raven-Replication-Version

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-13978

### DIFF
--- a/Raven.Database/Bundles/Replication/Impl/ReplicationHiLo.cs
+++ b/Raven.Database/Bundles/Replication/Impl/ReplicationHiLo.cs
@@ -16,52 +16,59 @@ namespace Raven.Bundles.Replication.Impl
 	internal class ReplicationHiLo
 	{
 		private readonly object generatorLock = new object();
-		private volatile Hodler currentMax = new Hodler(0);
+		private volatile ValueHolder valueHolder = new ValueHolder(0);
 		private long capacity = 256;
-		private long current;
 		private DateTime lastRequestedUtc;
 		public DocumentDatabase Database { get; set; }
 
-		private class Hodler
-		{
-			public readonly long Value;
+        public class ValueHolder
+        {
+            public readonly long MaxValue;
 
-			public Hodler(long value)
-			{
-				Value = value;
-			}
-		}
+            public long Current;
 
-		public long NextId()
-		{
-			long incrementedCurrent = Interlocked.Increment(ref current);
-			while (incrementedCurrent > currentMax.Value)
-			{
-				lock (generatorLock)
-				{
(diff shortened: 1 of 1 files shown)


Raven.Database/Bundles/Replication/Impl/ReplicationHiLo.cs
```diff
@@ -16,52 +16,59 @@ namespace Raven.Bundles.Replication.Impl
...
+        }
+
+        public long NextId()
+        {
+            var current = valueHolder;
+            var incrementedCurrent = Interlocked.Increment(ref current.Current);
+
+            while (incrementedCurrent > current.MaxValue)
+            {
+                lock (generatorLock)
```
ayende, line 43: We probably don't need this, given that we now hold all of this in the ValueHolder

Raven.Database/Bundles/Replication/Impl/ReplicationHiLo.cs
```diff
@@ -16,52 +16,59 @@ namespace Raven.Bundles.Replication.Impl
...
+                {
+                    current = valueHolder;
```
ayende, line 45: Need to use volatile read here.

Raven.Database/Bundles/Replication/Impl/ReplicationHiLo.cs
```diff
@@ -16,52 +16,59 @@ namespace Raven.Bundles.Replication.Impl
...
+                    if (current.Current > current.MaxValue)
+                    {
                         using (var locker = Database.DocumentLock.TryLock(250))
                         {
                             if (locker == null)
                                 continue;
-                            currentMax = new Hodler(GetNextMax());
+
+                            current = valueHolder = GetNextMax(current.MaxValue);
```
ayende, line 53: valueHolder *must* only be set using `Interlocked` operations

ayende (changes requested):

Raven.Database/Bundles/Replication/Impl/ReplicationHiLo.cs
grisha-kotler: This is running under the DocumentLock

Raven.Database/Bundles/Replication/Impl/ReplicationHiLo.cs
grisha-kotler: valueHolder is volatile

ppekrol (approved):

ayende (approved):

PR #9682, opened 2019-09-10
RavenDB-13099 : TimeSeries backups

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-13099

### DIFF
--- a/src/Raven.Client/Documents/Operations/DatabaseStatistics.cs
+++ b/src/Raven.Client/Documents/Operations/DatabaseStatistics.cs
@@ -77,7 +77,7 @@ namespace Raven.Client.Documents.Operations
         public long CountOfCounterEntries { get; set; }
 
         /// <summary>
-        /// Total number of time-series entries in database.
+        /// Total number of time-series segments in database.
         /// </summary>
         public long CountOfTimeSeriesSegments { get; set; }
 
--- a/src/Raven.Client/Documents/Operations/DatabaseStatistics.cs
+++ b/src/Raven.Client/Documents/Operations/DatabaseStatistics.cs
@@ -77,9 +77,9 @@ namespace Raven.Client.Documents.Operations
         public long CountOfCounterEntries { get; set; }
 
         /// <summary>
-        /// Total number of counter-group entries in database.
+        /// Total number of time-series entries in database.
         /// </summary>
-        public long CountOfTimeSeries { get; set; }
+        public long CountOfTimeSeriesSegments { get; set; }
 
         /// <summary>
         /// List of stale index names in database..
--- a/src/Raven.Client/Documents/Operations/DatabaseStatistics.cs
+++ b/src/Raven.Client/Documents/Operations/DatabaseStatistics.cs
@@ -77,6 +77,11 @@ namespace Raven.Client.Documents.Operations
         public long CountOfCounterEntries { get; set; }
 
         /// <summary>
+        /// Total number of counter-group entries in database.
+        /// </summary>
+        public long CountOfTimeSeries { get; set; }
+
+        /// <summary>
         /// List of stale index names in database..
         /// </summary>
         public string[] StaleIndexes => Indexes?.Where(x => x.IsStale).Select(x => x.Name).ToArray();
--- a/src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
@@ -356,17 +356,13 @@ namespace Raven.Server.Documents.Handlers
                 var tss = _database.DocumentsStorage.TimeSeriesStorage;
 
                 var changes = 0;
-                CollectionName collectionName = default;
 
                 foreach (var (docId, items) in _dictionary)
                 {
+                    var collectionName = _database.DocumentsStorage.ExtractCollectionName(context, items[0].Collection);
+
                     foreach (var item in items)
                     {
-                        if (collectionName == default)
-                        {
-                            collectionName = _database.DocumentsStorage.ExtractCollectionName(context, item.Collection);
-                        }
-
                         using (var slicer = new TimeSeriesStorage.TimeSeriesSlicer(context, docId, item.Name, item.Baseline))
                         {
                             if (tss.TryAppendEntireSegment(context, slicer.TimeSeriesKeySlice, collectionName, item.ChangeVector, item.Segment, item.Baseline))
@@ -381,7 +377,6 @@ namespace Raven.Server.Documents.Handlers
                         context.LastDatabaseChangeVector = ChangeVectorUtils.MergeVectors(changeVector, item.ChangeVector);
                     }
 
-                    collectionName = default;
                     changes += items.Count;
                 }
 
--- a/src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
@@ -14,7 +14,9 @@ using Raven.Server.Routing;
 using Raven.Server.ServerWide.Context;
 using Raven.Server.Smuggler.Documents;
 using Raven.Server.TrafficWatch;
+using Raven.Server.Utils;
 using Sparrow.Json;
+using Voron;
 
 namespace Raven.Server.Documents.Handlers
 {
@@ -175,8 +177,7 @@ namespace Raven.Server.Documents.Handlers
             }
         }
 
-        
-        
+              
         public class ExecuteTimeSeriesBatchCommand : TransactionOperationsMerger.MergedTransactionCommand
         {
             private readonly DocumentDatabase _database;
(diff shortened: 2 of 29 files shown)


src/Raven.Client/Documents/Operations/DatabaseStatistics.cs
```diff
@@ -76,6 +76,11 @@ public class DatabaseStatistics
         /// </summary>
         public long CountOfCounterEntries { get; set; }
 
+        /// <summary>
+        /// Total number of time-series entries in database.
```
ppekrol: entries -> segments

src/Raven.Server/Smuggler/Documents/Data/SmugglerMetadataModifier.cs
```diff
@@ -25,6 +25,9 @@ public void ModifyMetadata(BlittableJsonReaderObject metadata, ref DynamicJsonVa
 
             if (_operateOnTypes.HasFlag(DatabaseItemType.Attachments) == false && metadata.TryGet(Constants.Documents.Metadata.Attachments, out BlittableJsonReaderArray _))
                 modifications.Remove(Constants.Documents.Metadata.Attachments);
+
+            if (metadata.TryGet(Constants.Documents.Metadata.TimeSeries, out BlittableJsonReaderArray _))
```
ppekrol: Shouldn't we check OperateOnTypes here?

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
```diff
@@ -1545,5 +1551,82 @@ private void SendCommands()
...
+            private Task _prevCommandTask = Task.CompletedTask;
+            private int _segmentsSize;
+            private readonly int _maxBatchSize;
+
+            public TimeSeriesActions(DocumentDatabase database)
+            {
+                _database = database;
+                _cmd = new TimeSeriesHandler.SmugglerTimeSeriesBatchCommand(database);
+
+                _maxBatchSize = PlatformDetails.Is32Bits || database.Configuration.Storage.ForceUsing32BitsPager
```
ppekrol: Let's start using `Size` from Sparrow for this kind of things

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
```diff
@@ -1545,5 +1551,82 @@ private void SendCommands()
...
+                    ? 2000 * 1024
+                    : 10_000 * 1024;
+            }
+
+            private void AddToBatch(TimeSeriesItem item)
+            {
+                _cmd.AddToDictionary(item);
+                _segmentsSize += item.Segment.NumberOfBytes;
```
ppekrol: Use `Size` struct

src/Raven.Server/Smuggler/Documents/DatabaseSource.cs
```diff
@@ -336,6 +340,13 @@ public IEnumerable<SubscriptionState> GetSubscriptions()
             return _database.SubscriptionStorage.GetAllSubscriptions(_serverContext, false, 0, int.MaxValue);
         }
 
+        public IEnumerable<TimeSeriesItem> GetTimeSeries()
```
ppekrol: Should we pass `collectionsToExport` here?

src/Raven.Server/Smuggler/Documents/StreamDestination.cs
```diff
@@ -632,6 +643,52 @@ public Stream GetTempStream()
...
+                Writer.WritePropertyName(nameof(TimeSeriesItem.SegmentSize));
+                Writer.WriteInteger(item.Segment.NumberOfBytes);
+                Writer.WriteComma();
+
+                Writer.WritePropertyName(nameof(TimeSeriesItem.Baseline));
+                Writer.WriteDateTime(item.Baseline, true);
+
+                Writer.WriteEndObject();
+
+                Writer.WriteMemoryChunk(item.Segment.Ptr, item.Segment.NumberOfBytes);
```
ppekrol, line 688: Will this yield a proper JSON?

src/Raven.Server/Smuggler/Documents/StreamDestination.cs
```diff
@@ -632,6 +643,52 @@ public Stream GetTempStream()
...
+            public unsafe void WriteTimeSeries(TimeSeriesItem item)
+            {
+                if (First == false)
+                    Writer.WriteComma();
+                First = false;
+
+                Writer.WriteStartObject();
+
+                Writer.WritePropertyName(nameof(TimeSeriesItem.DocId));
+                Writer.WriteString(item.DocId);
```
ppekrol, line 664: In `StreamCounterActions` we are setting skipEscaping also here. What is the correct way?

src/Raven.Client/Documents/Operations/DatabaseStatistics.cs
ayende: The value here is actually the segments. I'm not sure if we are even *storing* the number of the entries.
@aviv86 do we have a cheap way to find how many _entries_ we have?

src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
```diff
@@ -327,6 +331,74 @@ public static void ThrowArtificialDocument(Document doc)
...
+                                continue;
+                            }
+                        }
+
+                        var changeVector = tss.AppendTimestamp(context, docId, item.Collection, item.Name, item.Segment.YieldAllValues(context, item.Baseline), item.ChangeVector);
+                        context.LastDatabaseChangeVector = ChangeVectorUtils.MergeVectors(changeVector, item.ChangeVector);
+                    }
+
+                    collectionName = default;
+                    changes += items.Count;
```
ayende, line 380: Shouldn't this also be increased for `TryAppendEntireSegment` too?

src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
```diff
@@ -327,6 +331,74 @@ public static void ThrowArtificialDocument(Document doc)
...
+                var tss = _database.DocumentsStorage.TimeSeriesStorage;
+
+                var changes = 0;
+                CollectionName collectionName = default;
+
+                foreach (var (docId, items) in _dictionary)
+                {
+                    foreach (var item in items)
+                    {
+                        if (collectionName == default)
```
ayende: Why is this inside the inner loop?
Shouldn't we make this unconditional operation before this loop?

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
```diff
@@ -1545,5 +1551,82 @@ private void SendCommands()
...
+            private int _segmentsSize;
+            private readonly int _maxBatchSize;
+
+            public TimeSeriesActions(DocumentDatabase database)
+            {
+                _database = database;
+                _cmd = new TimeSeriesHandler.SmugglerTimeSeriesBatchCommand(database);
+
+                _maxBatchSize = PlatformDetails.Is32Bits || database.Configuration.Storage.ForceUsing32BitsPager
+                    ? 2000 * 1024
```
ayende: For 32 bits, we need to use 2048, not 2048000

src/Raven.Server/Smuggler/Documents/DatabaseSource.cs
ayende: Yes, but take into account @arekpalinski's work around this

src/Raven.Server/Smuggler/Documents/StreamDestination.cs
ayende: Very unlikely, I would say. It just throw the raw bytes there.
I think we need to handle that like attachments, no?

src/Raven.Server/Smuggler/Documents/Data/SmugglerMetadataModifier.cs
aviv86: done

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
aviv86: done

src/Raven.Server/Smuggler/Documents/DatabaseSource.cs
aviv86: done

src/Raven.Server/Smuggler/Documents/StreamDestination.cs
aviv86: no this will not yield a valid JSON, just raw bytes
attachments are handled in a similar way

src/Raven.Server/Smuggler/Documents/StreamDestination.cs
aviv86: in `StreamCounterActions` the `DocumentId` is a `LazyStringValue`, here `DocId` is a string

src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
aviv86: we are counting the total number of segments here. so either way it's increased. 
should we count the total number of entries instead?

src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
aviv86: done

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
aviv86: so, for 32 bits use 2KB and for 64 bits  use 10MB ?
for documents import we do 2MB and 32MB

ayende: 1MB for 32 bits, 16 MB for 64 bits

src/Raven.Client/Documents/Operations/DatabaseStatistics.cs
aviv86: no we are not storing the number of entries.
not sure about a cheap way to get this. 

we also want to be able to get start / end dates & number of items cheaply 
(see https://issues.hibernatingrhinos.com/issue/RavenDB-13519)

maybe we should store all this information in a separate table ?

ayende: Separate issues / use case.
* Tracking the total number of time series entries is for the whole db.
* Tracking the start/end dates for a particular time series in a particular doc.

ayende: I think we can probably just get start / end cheaply enough by looking at the base value for the first segment and last value + most recent value (which we keep in the header)

aviv86: yes, separate issues. 
but if we'll track start/end & count for a particular time series in a particular doc, then finding out the total number of time series entries for the whole db is trivial - it's just the number of entries in this new table

ppekrol (approved):

src/Raven.Client/Documents/Operations/DatabaseStatistics.cs
ayende: I don't follow. _What_ new table?
I meant that we'll be looking at the segments, not keep additional data

ayende (approved):

PR #9694, opened 2019-09-15
RavenDB-13987: Server wide backups should not wakeup idle databases

garayx:


grisha-kotler (approved):

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -874,41 +895,49 @@ public void UnloadDirectly(StringSegment databaseName, DateTime? wakeup = null,
 
             LastRecentlyUsed.TryRemove(databaseName, out _);
             DatabasesCache.RemoveLockAndReturn(databaseName.Value, CompleteDatabaseUnloading, out _, caller).Dispose();
-            ScheduleDatabaseWakeup(databaseName.Value, dueTime);
+
+            if (dueTime == 0)
+                return;
+
+            _wakeupTimers.TryAdd(databaseName.Value, new Timer(_ => StartDatabaseOnTimer(databaseName.Value, wakeup), null, dueTime, Timeout.Infinite));
```
ppekrol, line 902: Why do not use `RescheduleDatabaseWakeup`? Is there a chance that we will have old timer that we need to dispose?

src/Raven.Server/Documents/PeriodicBackup/PeriodicBackupRunner.cs
```diff
@@ -134,8 +138,8 @@ public PeriodicBackupRunner(DocumentDatabase database, ServerStore serverStore)
             bool skipErrorLog = false)
         {
             var now = SystemTime.UtcNow;
-            var lastFullBackup = backupStatus.LastFullBackupInternal ?? now;
-            var lastIncrementalBackup = backupStatus.LastIncrementalBackupInternal ?? backupStatus.LastFullBackupInternal ?? now;
+            var lastFullBackup = backupStatus.LastFullBackupInternal ?? _databaseWakeUpTime ?? now;
```
ppekrol, line 141: Why do we need this?

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -1010,6 +1012,37 @@ private void OnValueChanged(object sender, (long Index, string Type) t)
...
+                {
+                    topology = rawRecord.GetTopology();
+                    backupConfig = rawRecord.GetPeriodicBackupConfiguration(taskId);
+                }
+
+                var tag = topology.WhoseTaskIsIt(Engine.CurrentState, backupConfig, null);
+                if (Engine.Tag == tag)
+                {
+                    var wakeup = CrontabSchedule.Parse(backupConfig.FullBackupFrequency).GetNextOccurrence(SystemTime.UtcNow);
+                    var dueTime = (int)(wakeup - DateTime.UtcNow).TotalMilliseconds;
```
ppekrol: use `SystemTime.UtcNow`

test/SlowTests/Issues/RavenDB-13987.cs
```diff
@@ -0,0 +1,154 @@
...
+
+            Assert.Equal(3, GetIdleCount());
+
+            // wait for the backup occurrence
+            var idleCount = WaitForCount(_reasonableWaitTime, 2, GetIdleCount);
+            Assert.Equal(2, idleCount);
+
+            // backup status should not wakeup dbs
+            var count = WaitForCount(_reasonableWaitTime, 3, CountOfBackupStatus);
+            Assert.Equal(3, count);
```
ppekrol: Can we add to this test that database will be waken up after all the modifications and backups will be performed?

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -185,6 +189,22 @@ private void HandleClusterDatabaseChanged(string databaseName, long index, strin
             }
         }
 
+        private bool PreventWakeupIdleDatabase(string databaseName, string type)
```
ppekrol: Wakeup -> WakeUp

src/Raven.Server/Documents/DatabasesLandlord.cs
garayx: UnloadDirectly() is called if and only if the DB become idle so _wakeupTimers cannot have the DB. RescheduleDatabaseWakeup called from putServerWideBackup only If the DB is idle, so it could have a timer already.

src/Raven.Server/Documents/PeriodicBackup/PeriodicBackupRunner.cs
garayx: to prevent a case:
DB is idle and have timer to make a backup _for first time,_ 
In this case we was passing now as lastFullBackup => nextBackupOccurrence calculated as the next occurrence (according to cron expression).
For example, db is idle and has timer for _first backup_:
backup each 5 mins -> woke up on 00:05 -> lastfullbackup = now (00:05) -> nextBackupOccurrence(lastfullbackup) = 00:10

src/Raven.Server/ServerWide/ServerStore.cs
garayx: done

src/Raven.Server/Documents/DatabasesLandlord.cs

PR #9699, opened 2019-09-16
RDBS-45: Add UseRssInsteadOfMemUsageInContainer for use together with…

aviviadi:
… RunningOnDocker=Y, to avoid too EOOM on docker instances, assuming RavenDB is the only process in the container

### DIFF
--- a/src/Raven.Server/Config/Categories/MemoryConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/MemoryConfiguration.cs
@@ -39,5 +39,10 @@ namespace Raven.Server.Config.Categories
         [SizeUnit(SizeUnit.Megabytes)]
         [ConfigurationEntry("Memory.MaxFreeCommittedMemoryToKeepInMb", ConfigurationEntryScope.ServerWideOnly)]
         public Size MaxFreeCommittedMemoryToKeepInMb { get; set; }
+
+        [Description("Use 'Rss'instead of 'memory.usage_in_bytes minus Shared Clean Memory' value to determine machine memory usage in docker instance. Default set to true. Applicable only with environment variable RAVEN_IN_DOCKER set to Y. Will use configuration option LowMemoryLimit with RavenDB process Rss value")]
+        [DefaultValue(true)]
+        [ConfigurationEntry("Memory.UseRssInsteadOfMemUsageInContainer", ConfigurationEntryScope.ServerWideOnly)]
+        public bool UseRssInsteadOfMemUsageInContainer { get; set; }
     }
 }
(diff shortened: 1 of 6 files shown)


ppekrol (approved):

src/Raven.Server/Config/Categories/MemoryConfiguration.cs
```diff
@@ -39,5 +39,10 @@ public MemoryConfiguration()
         [SizeUnit(SizeUnit.Megabytes)]
         [ConfigurationEntry("Memory.MaxFreeCommittedMemoryToKeepInMb", ConfigurationEntryScope.ServerWideOnly)]
         public Size MaxFreeCommittedMemoryToKeepInMb { get; set; }
+
+        [Description("EXPERT: Use 'RSS' instead of 'memory.usage_in_bytes minus Shared Clean Memory' value to determine machine memory usage in docker instance. Applicable only with environment variable RAVEN_IN_DOCKER is set to 'true' and only when running on Linux. Will use configuration option 'Memory.LowMemoryLimitInMb' with RavenDB process RSS value. Default: true.")]
+        [DefaultValue(true)]
+        [ConfigurationEntry("Memory.UseRssInsteadOfMemUsageInContainer", ConfigurationEntryScope.ServerWideOnly)]
+        public bool UseRssInsteadOfMemUsageInContainer { get; set; }
```
ayende, line 46: I don't like this option. Too tailored.
Instead, let's just make it possible for the admin to make the selection between the option. Regardless of docker or not

src/Raven.Server/Config/Categories/MemoryConfiguration.cs
```diff
@@ -39,5 +39,10 @@ public MemoryConfiguration()
         [SizeUnit(SizeUnit.Megabytes)]
         [ConfigurationEntry("Memory.MaxFreeCommittedMemoryToKeepInMb", ConfigurationEntryScope.ServerWideOnly)]
         public Size MaxFreeCommittedMemoryToKeepInMb { get; set; }
+
+        [Description("EXPERT: Use 'RSS' instead of 'memory.usage_in_bytes minus Shared Clean Memory' value to determine machine memory usage in docker instance. Applicable only with environment variable RAVEN_IN_DOCKER is set to 'true' and only when running on Linux. Will use configuration option 'Memory.LowMemoryLimitInMb' with RavenDB process RSS value. Default: true.")]
+        [DefaultValue(true)]
```
ayende, line 44: Change this to be set in the ctor, so the _default_ can be affected by whatever we are in docker or not

src/Raven.Server/Config/Categories/MemoryConfiguration.cs
ppekrol: https://github.com/ravendb/ravendb/pull/9700

PR #9709, opened 2019-09-24
RDBCL-772 Exposing last authorized non cluster admin request time in SNMP

arekpalinski:


ppekrol (approved):

src/Raven.Server/Routing/RequestRouter.cs
```diff
@@ -114,11 +110,26 @@ public async ValueTask HandlePath(RequestHandlerContext reqCtx)
...
+                    }
+                }
+                finally
+                {
+                    if (tryMatch.Value.SkipLastRequestTimeUpdate == false)
+                    {
+                        var now = SystemTime.UtcNow;
+                        _ravenServer.Statistics.LastRequestTime = now;
+                        if (skipAuthorization == false && authResult && status != RavenServer.AuthenticationStatus.ClusterAdmin)
+                            _ravenServer.Statistics.LastAuthorizedNonClusterAdminRequestTime = now;
```
ayende, line 131: I wonder if we shouldn't be setting these operations only once every 1 - 5 seconds.
The reason is that otherwise under load, we may create a hotspot.
Another issue is whatever in multi threaded code, we'll have tearing of the the value in some cases

arekpalinski: Hmm... can setting a property cause a hotpath? It's the same what we already have for any request:
`_ravenServer.Statistics.LastRequestTime = now;`

and never got problems with that.

You're right that value might be not accurate in some multi threading cases although discrepancy won't be big so I think we can afford it here

ayende: You are writing to the same memory location from multiple threads, may cause sync costs / cache fights between cores.
We can also do the `LastRequestCheck` every 15 sec

arekpalinski: Fixed in https://github.com/ravendb/ravendb/pull/9717

PR #9712, opened 2019-09-24
RavenDB-14049 - Map-Reduce isn't updated after documents deletion

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-14051
https://issues.hibernatingrhinos.com/issue/RavenDB-14058

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -114,6 +114,9 @@ namespace Raven.Server.Documents.Indexes
         private readonly ConcurrentDictionary<string, IndexProgress.CollectionStats> _inMemoryIndexProgress =
             new ConcurrentDictionary<string, IndexProgress.CollectionStats>();
 
+        private readonly ConcurrentDictionary<string, IndexProgress.CollectionStats> _inMemoryReferencesIndexProgress =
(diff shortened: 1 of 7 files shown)


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -114,6 +114,9 @@ public abstract class Index : ITombstoneAware, IDisposable, ILowMemoryHandler
         private readonly ConcurrentDictionary<string, IndexProgress.CollectionStats> _inMemoryIndexProgress =
             new ConcurrentDictionary<string, IndexProgress.CollectionStats>();
 
+        private readonly ConcurrentDictionary<string, IndexProgress.CollectionStats> _inMemoryReferencesIndexProgress =
```
ppekrol, line 117: Shouldn't we use `StringComparer.OridinalIgnoreCase` here (and in the other stats) ?

ayende (approved):

src/Raven.Server/Documents/Indexes/Index.cs
ppekrol: Nvm, this comes from compiled.ReferencedCollections which has that compiler.

ppekrol: Comparer*

arekpalinski:
LGTM

PR #9715, opened 2019-09-25
RavenDB-14061: RAVEN_IN_DOCKER instance with UseTotalDirtyMemInsteadOfMemUsage …

aviviadi:
…set (default=true) will calculate usage memory as Dirty sum of entire RavenDB process

### DIFF
--- a/src/Sparrow.Server/LowMemory/MemoryInformation.cs
+++ b/src/Sparrow.Server/LowMemory/MemoryInformation.cs
@@ -225,7 +225,8 @@ namespace Sparrow.LowMemory
             }
         }
 
-        internal static (Size MemAvailable, Size TotalMemory, Size Commited, Size CommitLimit, Size AvailableWithoutTotalCleanMemory, Size SharedCleanMemory) GetFromProcMemInfo(SmapsReader smapsReader)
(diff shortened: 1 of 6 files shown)


src/Sparrow.Server/LowMemory/MemoryInformation.cs
```diff
@@ -225,7 +225,8 @@ public static MemoryInfoResult GetMemoryInformationUsingOneTimeSmapsReader()
             }
         }
 
-        internal static (Size MemAvailable, Size TotalMemory, Size Commited, Size CommitLimit, Size AvailableWithoutTotalCleanMemory, Size SharedCleanMemory) GetFromProcMemInfo(SmapsReader smapsReader)
+        internal static (Size MemAvailable, Size TotalMemory, Size Commited, Size CommitLimit, Size AvailableWithoutTotalCleanMemory, Size SharedCleanMemory, Size TotalDirty)
```
ayende, line 228: Let's make a proper struct of this at this point and accept it as a `ref` parameter

aviviadi: https://github.com/ravendb/ravendb/pull/9721

PR #9723, opened 2019-09-26
RavenDB-13902: vfork

aviviadi:
Please review. PAL Libraries will be built later.


aviviadi:
Comments were published also at https://github.com/ravendb/ravendb/pull/9675. The above tries to solve them too.

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,297 @@
...
+_parse_cmd_line(char* line, char *filename)
+{
+    int argc = 2;
+    char** argv = malloc(argc);
+    argv[0] = filename;
+    argv[--argc] = NULL;    
+    if (line == NULL || *line == '\0')
+        return argv;
+    
+    char * pch = strtok (line," \t\n");
```
ayende: Use `strtok_r` here.

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,297 @@
...
+                    sigaction(sig, &sa_default, NULL);
+            }
+        }
+        pthread_sigmask(SIG_SETMASK, &old_signal_set, &junk_signal_set); 
+        
+        /* redirect both stdout and stderr to the same stream, and redirect stdin: */
+        if ( (result = _redirect_std(stdoutFds[1], STDOUT_FILENO)) == -1 ||
+             (result = _redirect_std(stdoutFds[1], STDERR_FILENO)) == -1 ||
+             (result = _redirect_std(stdinFds[1],  STDIN_FILENO)) == -1  ||
+             (result = _redirect_std(stdoutFds[1], STDOUT_FILENO)) == -1 )
```
ayende: This is duplicated, no?

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,297 @@
...
+        {
+            int status;
+            waitpid(processId, &status, 0);
+        }
+
+        *pid = (void*)(intptr_t)-1;
+    }
+
+    free(modified);
+
```
ayende, line 303: Where is `argv` freed?

src/Sparrow.Server/Platform/RavenProcess.cs
```diff
@@ -0,0 +1,227 @@
...
+            StandardOutAndErr = stdout;
+            StandardIn = stdin;
+        }
+
+        public static void Start(string filename, string arguments)
+        {
+            using (var ravenProcess = new RavenProcess {StartInfo = new ProcessStartInfo {FileName = filename, Arguments = arguments}})
+            {
+                ravenProcess.Start(CancellationToken.None);
+                ravenProcess.ReadTo(Console.Out);
```
ayende: Under what scenario do I want _this_ to happen?

src/Sparrow.Server/Platform/RavenProcess.cs
```diff
@@ -0,0 +1,227 @@
...
+{
+    public class ProcessExitedEventArgs : EventArgs
+    {
+        public int ExitCode { get; set; }
+        public IntPtr Pid { get; set; }
+    }
+
+    public class LineOutputEventArgs : EventArgs
+    {
+        public string line { get; set; }
```
ayende:
```suggestion
        public string Line { get; set; }
```

src/Sparrow.Server/Platform/RavenProcess.cs
```diff
@@ -0,0 +1,227 @@
...
+                        } while (read != null && ctk.IsCancellationRequested == false);
+
+                        if (process._hasExited)
+                            break;
+                    }
+                }
+            }
+        }
+
+        private Task<string> ReadLineAsync(FileStream fs, CancellationToken ctk)
```
ayende: Remove this method, use `StreamReader`

src/Sparrow.Server/Platform/RavenProcess.cs
```diff
@@ -0,0 +1,227 @@
...
+        }
+
+        public void Dispose()
+        {
+            if (_hasExited == false)
+            {
+                _hasExited = true;
+                try
+                {
+                    var _ = ReadToEndAsync().Result;
```
ayende: You need to start this (and not wait), then wait for close, then wait on the result (to flush everything)

ayende (changes requested):

src/Sparrow.Server/Platform/RavenProcess.cs
aviviadi: That what is actually happens when you start any process (i.e: "xdg-open" writes its stderr to the console). May be it is better to pass a TextWriter's stream from the caller to `Start` ?

src/Sparrow.Server/Platform/RavenProcess.cs
aviviadi: sr returns only when the child process exists..
```
using (var sr = new StreamReader(fs, Encoding.UTF8))
    read = process.ReadLineAsync(fs, ctk).Result;
```

src/Sparrow.Server/Platform/RavenProcess.cs
ayende: That goes to `Console.Out`, which is the _server_ output, which is usually not observed.
We never want to allow that. Have the user provide a text writer for this, or better yet, a callback, instead.

src/Sparrow.Server/Platform/RavenProcess.cs
ayende: That doesn't match the behavior I'm familiar with.
If you are doing things sync, use `ReadLine`.

aviviadi: Moreover,
I used StreamReader's ReadLine(), and it waits until process exists and just prints the first line and the throws `System.ArgumentException: Stream was not readable.`

Looking at : https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/IO/StreamReader.cs

They do (more or less) the same as I do. Reading from TW, however I do that char by char w/o throwing on null read. In corefx they throw if they "read" null.

aviviadi: ..and ReadLine blocks.. until new char is read. So without ctk we might face hung

src/Sparrow.Server/Platform/RavenProcess.cs
```diff
@@ -61,27 +63,28 @@ public void Start(CancellationToken ctk)
             StandardIn = stdin;
         }
 
-        public static void Start(string filename, string arguments)
+        public static void Start(string filename, string arguments, StreamWriteDelegate outputDel)
```
ayende, line 64: Why is it a `MemoryStream` ?

ayende: Pass a `Span<byte>` here

src/Sparrow.Server/Platform/RavenProcess.cs
```diff
@@ -145,7 +148,10 @@ public static void Execute(string command, string arguments, int pollingTimeoutI
                         string read = null;
                         do
                         {
-                            read = process.ReadLineAsync(fs, ctk).Result;
+                            read = process.ReadLineAsync(fs, ctk);
+//                            using (var sr = new StreamReader(fs, Encoding.UTF8))
```
ayende: Remove commented code

src/Sparrow.Server/Platform/RavenProcess.cs
```diff
@@ -165,7 +171,7 @@ public static void Execute(string command, string arguments, int pollingTimeoutI
             }
         }
 
-        private Task<string> ReadLineAsync(FileStream fs, CancellationToken ctk)
+        private string ReadLineAsync(FileStream fs, CancellationToken ctk)
```
ayende: This is no longer async method

src/Sparrow.Server/Platform/RavenProcess.cs
```diff
@@ -174,7 +180,6 @@ private Task<string> ReadLineAsync(FileStream fs, CancellationToken ctk)
             {
                 if (sb == null)
                     sb = new StringBuilder();
-
                 var c = Encoding.UTF8.GetString(buffer, 0, read);
 
                 if (buffer[0] == '\n')
```
ayende: This is wrong, the `\n` can happen anywhere in the buffer, not just the first one.

ayende (changes requested):

src/Sparrow.Server/Platform/RavenProcess.cs
aviviadi: This is 1 byte buffer.  However I am going to rewrite this part (read 4096 and store the reset after \n for next iteration)

src/Sparrow.Server/Platform/RavenProcess.cs
aviviadi: my bad.. fixed in https://github.com/ravendb/ravendb/pull/9723/commits/fcb52895280d49b357ed0b831fda17cc66352a15

src/Raven.Pal/src/posix/maconly.c
```diff
@@ -68,4 +69,19 @@ rvn_test_storage_durability(
...
+{
+    int32_t result = FAIL;
+    while ((result = pipe(descriptors)) < 0 && errno == EINTR);
+    if (result != 0)
+        return FAIL_CREATE_PIPE;
+    
+    while ((result = fcntl(descriptors[0], F_SETFD, FD_CLOEXEC)) < 0 && errno == EINTR);
+    if (result == 0)
+        while ((result = fcntl(descriptors[1], F_SETFD, FD_CLOEXEC)) < 0 && errno == EINTR);
+
```
ayende, line 83: Shouldn't we be closing the `descriptors` if there is a failure here?

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,326 @@
...
+#include "rvn.h"
+#include "status_codes.h"
+#include "internal_posix.h"
+
+PRIVATE char** 
+_parse_cmd_line(char* line, char *filename)
+{
+    int argc = 2;
+    char** argv = malloc(argc);
```
ayende: check if `argv` is `NULL` here

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,326 @@
...
+    argv[0] = filename;
+    argv[--argc] = NULL;    
+    if (line == NULL || *line == '\0')
+        return argv;
+    
+    char *rest = line;
+    char *token;
+
+    while((token = strtok_r(rest, " \t\n", &rest)))
```
ayende: I'm pretty sure that you need a _separate_ context variable here.

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,326 @@
...
+    {
+        argv[argc++] = token;
+        argv = realloc(argv, argc + 1);
```
ayende: Check if this failed

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,326 @@
...
+
+    _exit(error != 0 ? error : EXIT_FAILURE);
+}
+
+PRIVATE int32_t
+_redirect_std(int fd, int fileno)
+{
+    int32_t result = dup2(fd, fileno);
+    while ( result < 0 && errno == EINTR)
+        result = dup2(fd, STDOUT_FILENO);
```
ayende:
```suggestion
        result = dup2(fd, fileno);
```

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,326 @@
...
+
+EXPORT int32_t
+rvn_spawn_process(const char* filename, char* cmdline, void** pid, void** standard_in, void** standard_out, int32_t* detailed_error_code) {    
+    int stdinFds[2] = {-1, -1}, stdoutFds[2] = {-1, -1}, waitForChildToExecPipe[2] = {-1, -1};
+    pid_t processId = -1;
+    int thread_cancel_state;
+    sigset_t signal_set;
+    sigset_t old_signal_set;
+
+    char *modified = malloc(strlen(filename) + 3);
```
ayende, line 181: check if malloc failed here.

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,326 @@
...
+        if (processId > 0)
+        {
+            int status;
+            waitpid(processId, &status, 0);
+        }
+
+        *pid = (void*)(intptr_t)-1;
+    }
+
+    free(modified);
```
ayende, line 302: `if (modified)`

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,326 @@
...
+        {        
+            if (WIFEXITED(status))
+            {
+                *exit_code = WEXITSTATUS(status);
+                return SUCCESS;
+            }
+            
+            if (WIFSIGNALED(status) || WIFSTOPPED(status))
+            {
+                *exit_code = WIFSIGNALED(status) || WIFSTOPPED(status);
```
ayende: I think you meant `|` here, no?

src/Sparrow.Server/Platform/RavenProcess.cs
```diff
@@ -0,0 +1,209 @@
...
+        public int ExitCode { get; set; }
+        public IntPtr Pid { get; set; }
+    }
+
+    public class RavenProcess : IDisposable
+    {
+        public ProcessStartInfo StartInfo { get; set; }
+        private bool _hasExited;
+
+        public event EventHandler ProcessExited;
```
ayende:
```suggestion
        public event EventHandler<ProcessExitedEventArgs> ProcessExited;
```

src/Sparrow.Server/Platform/RavenProcess.cs
```diff
@@ -0,0 +1,209 @@
...
+                }
+            }
+        }
+
+        public Task<string> ReadToEndAsync()
+        {
+            using (var fs = new FileStream(StandardOutAndErr, FileAccess.Read))
+            using (var sr = new StreamReader(fs))
+            {
+                return sr.ReadToEndAsync();
```
ayende: This is wrong here.
You are returning a task, but immediately disposing the stream before you actually read anything.

src/Sparrow.Server/Platform/RavenProcess.cs
```diff
@@ -0,0 +1,209 @@
...
+                    process.ProcessExited += exitHandler;
+                if (lineOutputHandler != null)
+                    process.LineOutput += lineOutputHandler;
+                process.Start(ctk);
+                using (var fs = new FileStream(process.StandardOutAndErr, FileAccess.Read))
+                using (var sr = new StreamReader(fs, Encoding.UTF8))
+                {
+                    while (ctk.IsCancellationRequested == false)
+                    {
+                        var rc = Pal.rvn_wait_for_close_process(process._pid, pollingTimeoutInSeconds, out var exitCode, out var errorCode);
```
ayende: why is the first thing that you are doing is waiting for it to close?

src/Sparrow.Server/Platform/RavenProcess.cs
```diff
@@ -0,0 +1,209 @@
...
+                {
+                    if (_logger.IsInfoEnabled)
+                        _logger.Info($"Kill {StartInfo.FileName} failed", ex);
+                }
+
+                try
+                {
+                    if (task.IsFaulted == false)
+                    {
+                        Task.WaitAny(new [] {task}, TimeSpan.FromSeconds(15));
```
ayende: task.Wait(15)?

ayende (changes requested):

src/Sparrow.Server/Platform/RavenProcess.cs
aviviadi: waiting to close with timeout.  polling...

src/Raven.Pal/src/posix/maconly.c
aviviadi: https://github.com/ravendb/ravendb/pull/9723/files/3406eb2840bcaf009737f0a829f0da841e86e133#diff-26816b18cf670fbae58f7f07ce0927fdR258

src/Raven.Pal/src/posix/spawn.c
arekpalinski: Is it enough to allocate only 2 bytes here? It's an allocation for array of pointers, isn't it?

src/Sparrow.Server/Platform/RavenProcess.cs
aviviadi: ok. will be fixed so instead of polling, this is general totoal waitForExitTimeoutInSeconds

src/Raven.Server/Utils/Cpu/CpuUsageExtensionPoint.cs
```diff
@@ -43,124 +42,97 @@ public class CpuUsageExtensionPoint : IDisposable
...
             catch (Exception e)
             {
-                HandleError(e.Message, e);
+                NotifyWarning("Could not start cpu usage extension point process", e);
+                // ignore
             }
         }
 
-        private bool HandleInfoReceived(string data)
+        private string HandleInfoReceived(string data)
```
arekpalinski: Maybe `bool TryHandleInfoReceived(string data, out string error)` instead to be more explicit if the method succeeds or not. Returning `null` on success isn't obvious and can be a surprise for someone in the future

src/Sparrow.Server/Platform/RavenProcess.cs
```diff
@@ -0,0 +1,209 @@
...
+
+        private readonly Logger _logger = LoggingSource.Instance.GetLogger<RavenProcess>("RavenProcess");
+        private IntPtr _pid = IntPtr.Zero;
+        private SafeFileHandle StandardOutAndErr { get; set; }
+        private SafeFileHandle StandardIn { get; set; }
+
+        public void Start(CancellationToken ctk)
+        {
+            if (StartInfo?.FileName == null)
+                throw new InvalidOperationException("RavenProcess Start() must be supplied with valid startInfo object and set Filename");
```
arekpalinski: Minor: using `nameof()` for `Start` and `Filename`

src/Sparrow.Server/Platform/RavenProcess.cs
```diff
@@ -0,0 +1,209 @@
...
+        }
+
+        public delegate void StreamWriteDelegate(Span<byte> bytes);
+
+        private readonly Logger _logger = LoggingSource.Instance.GetLogger<RavenProcess>("RavenProcess");
+        private IntPtr _pid = IntPtr.Zero;
+        private SafeFileHandle StandardOutAndErr { get; set; }
+        private SafeFileHandle StandardIn { get; set; }
+
+        public void Start(CancellationToken ctk)
```
arekpalinski: Can we drop `ctk` parameter? I don't see any usage of it

src/Sparrow.Server/Platform/RavenProcess.cs
```diff
@@ -0,0 +1,209 @@
...
+                    _logger.Info($"Waited 5 seconds for {StartInfo.FileName} to close, but it didn't, trying to kill");
+                try
+                {
+                    Kill();
+                }
+                catch (Exception ex)
+                {
+                    if (_logger.IsInfoEnabled)
+                        _logger.Info($"Kill {StartInfo.FileName} failed", ex);
+                }
```
arekpalinski, line 210: I think it should be logged at `Operations` level

src/Raven.Server/Utils/Cpu/CpuUsageCalculator.cs
```diff
@@ -256,19 +257,21 @@ protected override unsafe MacInfo GetProcessInfo()
...
+            NotificationCenter.NotificationCenter notificationCenter,
+            CancellationToken serverStoreServerShutdown)
         {
             _inspector = new CpuUsageExtensionPoint(
                 contextPool,
                 exec,
                 args,
                 notificationCenter
             );
+            _serverShutdown = CancellationToken.None;
```
ppekrol: shouldn't you assign 'serverStoreServerShutdown' here?

src/Raven.Server/Utils/Cpu/CpuUsageCalculator.cs
```diff
@@ -256,19 +257,21 @@ protected override unsafe MacInfo GetProcessInfo()
...
+        private readonly CancellationToken _serverShutdown;
 
-        public ExtensionPointCpuUsageCalculator(
-            JsonContextPool contextPool,
+        public ExtensionPointCpuUsageCalculator(JsonContextPool contextPool,
             string exec,
             string args,
-            NotificationCenter.NotificationCenter notificationCenter)
+            NotificationCenter.NotificationCenter notificationCenter,
+            CancellationToken serverStoreServerShutdown)
```
ppekrol: rename to 'serverShutdown'

src/Sparrow.Server/Platform/RavenProcess.cs
```diff
@@ -0,0 +1,209 @@
...
+                    PalHelper.ThrowLastError(rc, errorCode, $"Failed to kill proc id={_pid.ToInt64()}. Command: '{StartInfo.FileName} {StartInfo.Arguments}'");
+            }
+        }
+
+        private void ReadLines(StreamReader sr, Action<object, string> lineOutputHandler, CancellationToken ctk)
+        {
+            while (ctk.IsCancellationRequested == false)
+            {
+                var lineTask = sr.ReadLineAsync();
+                if (lineTask.Wait(-1, ctk) == false)
```
arekpalinski: Minor: you can use `Timeout.Infinite` instead of `-1` to make it a bit more clear

src/Raven.Pal/src/posix/spawn.c
ayende: It is _not_ valid, you are correct. **Good** catch!

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
```diff
@@ -178,6 +178,12 @@ public virtual bool RunInMemory
         [ConfigurationEntry("Indexing.MaxTimeToWaitAfterFlushAndSyncWhenExceedingScratchSpaceLimit", ConfigurationEntryScope.ServerWideOnly)]
         public TimeSetting MaxTimeToWaitAfterFlushAndSyncWhenExceedingScratchSpaceLimit { get; protected set; }
 
+        [Description("Indicates if missing fields should be indexed same as 'null' values or not. Default: false")]
+        [DefaultValue(false)]
+        [IndexUpdateType(IndexUpdateType.Reset)]
+        [ConfigurationEntry("Indexing.IndexMissingFieldsAsNull", ConfigurationEntryScope.ServerWideOrPerDatabase)]
+        public bool IndexMissingFieldsAsNull { get; set; }
```
ppekrol: Do not worry about this. Just tell me when PR is ready and I will rebase.

src/Raven.Server/Utils/Cpu/CpuUsageExtensionPoint.cs
```diff
@@ -43,124 +42,97 @@ public class CpuUsageExtensionPoint : IDisposable
...
+                var retry = 2;
+                const int maxMinutesBetweenIssues = 15;
+                var lastRestart = DateTime.UtcNow;
+                var lastReceivedLine = DateTime.UtcNow;
+                while (retry-- > 0)
                 {
-                    var nextLine = _process.StandardOutput.ReadLineAsync();
-
-                    if (await nextLine.WaitWithTimeout(_timeout) == false)
+                    var cts = new CancellationTokenSource();
```
ppekrol: should we dispose this?

src/Sparrow.Server/Platform/RavenProcess.cs
aviviadi: But future usage possible and also makes sense.

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -28,21 +28,40 @@ _parse_cmd_line(char* line, char *filename)
...
         return argv;
+    }
     
     char *rest = line;
     char *token;
 
-    while((token = strtok_r(rest, " \t\n", &rest)))
-    {
+    const char delimiters[] = " \t\n";
+    while((token = strtok_r(rest, delimiters, &rest)))
```
ayende: `rest` is shown here twice, pretty sure you need to specify different variable here.

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
```diff
@@ -178,6 +178,12 @@ public virtual bool RunInMemory
         [ConfigurationEntry("Indexing.MaxTimeToWaitAfterFlushAndSyncWhenExceedingScratchSpaceLimit", ConfigurationEntryScope.ServerWideOnly)]
         public TimeSetting MaxTimeToWaitAfterFlushAndSyncWhenExceedingScratchSpaceLimit { get; protected set; }
 
+        [Description("Indicates if missing fields should be indexed same as 'null' values or not. Default: false")]
```
ayende: this is a bad merge?

ayende (changes requested):

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
ppekrol: I will handle that and the end, we can ignore this for now

src/Raven.Pal/src/posix/spawn.c
arekpalinski: Then we have the same issue below `argv = realloc(argv, argc + 1);`

src/Sparrow.Server/Platform/RavenProcess.cs
```diff
@@ -0,0 +1,214 @@
...
+            using (var ravenProcess = new RavenProcess {StartInfo = new ProcessStartInfo {FileName = filename, Arguments = arguments}})
+            {
+                ravenProcess.Start(CancellationToken.None);
+                ravenProcess.ReadTo(outputDel);
+            }
+        }
+
+        private void ReadTo(StreamWriteDelegate outputDel)
+        {
+            var bytes = new byte[4096];
```
ppekrol: use ArrayPool here?

ppekrol: Add a const with 4096

src/Sparrow.Server/Platform/RavenProcess.cs
```diff
@@ -0,0 +1,214 @@
...
+            using (var fs = new FileStream(StandardOutAndErr, FileAccess.Read))
+            {
+                var read = fs.Read(bytes, 0, 4096);
```
ppekrol: Use the defined const here

src/Raven.Pal/src/posix/spawn.c
arekpalinski: argv = realloc(argv, argc + 1); - see comment above

src/Sparrow.Server/Platform/RavenProcess.cs
```diff
@@ -0,0 +1,214 @@
...
+                while (read != 0)
+                {
+                    outputDel?.Invoke(new Span<byte>(bytes, 0, read));
+                    read = fs.Read(bytes, 0, 4096);
```
ppekrol: Use the defined const here

src/Raven.Pal/src/posix/spawn.c
aviviadi: already fixed... failed while I tried more then 3 arguments ;) ..

src/Raven.Pal/src/posix/spawn.c
aviviadi: nop. I meant it to be this way.

src/Sparrow.Server/Platform/RavenProcess.cs
ayende: This is internal API, we can change it at will

aviviadi: https://github.com/ravendb/ravendb/pull/9723/commits/35cf7d7b67f312e093d3a1b6ea29fd493c82c10f

src/Raven.Pal/src/posix/spawn.c
aviviadi: Although it works, you are right, the correct usage is:
https://github.com/ravendb/ravendb/pull/9723/commits/34cebfa76f1074867f2e96a9f7d089719cf23ddb

src/Raven.Pal/src/posix/spawn.c
```diff
@@ -0,0 +1,358 @@
...
+{
+    int argc = 2;
+    har** argv = malloc(sizeof(long) * argc);
+    if (argv == NULL)
+        return NULL;
+    argv[0] = filename;
+    argv[--argc] = NULL;    
+    if (line == NULL || *line == '\0')
+    {
+        char **tmpargv = realloc(argv, sizeof(long)*(argc + 2));
```
ayende, line 37:
```suggestion
        char **tmpargv = realloc(argv, sizeof(long)*3); /* filename, empty, NULL */
```

src/Raven.Pal/src/posix/spawn.c
ayende: Still need the `if` here, and let's place it need the `argv`

src/Raven.Server/Utils/Cpu/CpuUsageExtensionPoint.cs
```diff
@@ -43,124 +42,97 @@ public class CpuUsageExtensionPoint : IDisposable
...
-
-                    var line = await nextLine;
-                    if (HandleInfoReceived(line) == false)
-                    {
-                        break;
+                        var lineOutHandler = new Action<object, string>((p, l) =>
+                        {
+                            Interlocked.Exchange(ref _isDataValid, 1);
+
+                            if (l == null)
```
ayende, line 75: You won't be _called_ here, so this wouldn't work

src/Raven.Server/Utils/Cpu/CpuUsageExtensionPoint.cs
```diff
@@ -43,124 +42,97 @@ public class CpuUsageExtensionPoint : IDisposable
...
-                        break;
-                    }
-
-                    var line = await nextLine;
-                    if (HandleInfoReceived(line) == false)
-                    {
-                        break;
+                        var lineOutHandler = new Action<object, string>((p, l) =>
+                        {
+                            Interlocked.Exchange(ref _isDataValid, 1);
```
ayende, line 73: what is this for?

src/Sparrow.Server/Platform/RavenProcess.cs
```diff
@@ -0,0 +1,224 @@
...
+        {
+            if (_pid != IntPtr.Zero)
+            {
+                var rc = Pal.rvn_kill_process(_pid, out var errorCode);
+                if (rc != PalFlags.FailCodes.Success)
+                    PalHelper.ThrowLastError(rc, errorCode, $"Failed to kill proc id={_pid.ToInt64()}. Command: '{StartInfo.FileName} {StartInfo.Arguments}'");
+            }
+        }
+
+        private void ReadLines(StreamReader sr, Action<object, string> lineOutputHandler, CancellationToken ctk)
```
ayende, line 113: `lineOutputHandler` is unused here?

src/Sparrow.Server/Platform/RavenProcess.cs
```diff
@@ -0,0 +1,224 @@
...
+        {
+            while (ctk.IsCancellationRequested == false)
+            {
+                var lineTask = sr.ReadLineAsync();
+                if (lineTask.Wait(Timeout.Infinite, ctk) == false)
+                    break;
+                var line = lineTask.Result;
+                OnLineOutput(line, ctk);
```
ayende, line 121: Did you mean to call this?

ayende (changes requested):

ayende:
Please create a new, clean, PR

PR #9724, opened 2019-09-26
RavenDB 13962 - Force Revisions Configuration

Danielle9897:


src/Raven.Studio/typescript/commands/database/settings/enforceRevisionsConfigurationCommand.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    execute(): JQueryPromise<operationIdDto> {
+        const url = endpoints.databases.revisions.adminRevisionsConfigEnforce;       
+
+        return this.post<void>(url, null, this.db)
+            .fail((response: JQueryXHR) => {
+                this.reportError("Failed to enforce revisions configuration", response.responseText, response.statusText); 
+            })
+            .done(() => {
+                this.reportSuccess("Revisions configuration enforced successfully");
```
ml054, line 19: please check if this is sync or async operation (in handler) - if async then we should write that config was enforced.

ml054 (changes requested):
please check my single comment in code

src/Raven.Studio/typescript/commands/database/settings/enforceRevisionsConfigurationCommand.ts
Danielle9897: `EnforceConfigRevisions` operation in handler is **sync** - so I assume it is ok like this.

ml054: ok great - I was just double checking.

ml054 (approved):

PR #9734, opened 2019-10-02
RavenDB-13932 fix err msg when creating ongoing-regular-backup + fix …

Danielle9897:
…server-wide-backup link in ongoing tasks list view

### DIFF
--- a/src/Raven.Server/ServerWide/Commands/PeriodicBackup/UpdatePeriodicBackupCommand.cs
+++ b/src/Raven.Server/ServerWide/Commands/PeriodicBackup/UpdatePeriodicBackupCommand.cs
@@ -24,9 +24,11 @@ namespace Raven.Server.ServerWide.Commands.PeriodicBackup
 
         public override string UpdateDatabaseRecord(DatabaseRecord record, long etag)
         {
+            bool newTask = false;
             if (Configuration.TaskId == 0)
             {
                 // this is a new backup configuration
+                newTask = true;
                 Configuration.TaskId = etag;
             }
             else
@@ -41,7 +43,7 @@ namespace Raven.Server.ServerWide.Commands.PeriodicBackup
             }
             else if (Configuration.Name.StartsWith(ServerWideBackupConfiguration.NamePrefix, StringComparison.OrdinalIgnoreCase))
             {
-                throw new InvalidOperationException($"Can't update task name '{Configuration.Name}', because it is a server-wide backup task");
+                throw new InvalidOperationException($"Can't {(newTask ? "create" : "update")} task: '{Configuration.Name}'. A regular (non server-wide) backup task name can't start with prefix 'server wide backup'");
             }
 
             EnsureTaskNameIsNotUsed(record, Configuration.Name);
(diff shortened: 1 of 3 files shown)


src/Raven.Studio/wwwroot/App/views/database/tasks/ongoingTasks.html
```diff
@@ -244,12 +244,9 @@ <h5 class="tasks-list-item no-text-transform" title="Navigate to the Server-Wide
...
                                     <div class="panel-name flex-grow">
-                                        <h3>
-                                            <span>
-                                                <a href="#" data-bind="click: editTask, attr: { 'title': 'Task name: ' + taskName() }">
-                                                    <span data-bind="text: taskName"></span>
-                                                </a>
-                                            </span>
+                                        <h3 data-bind="attr: { 'title': 'Task name: ' + taskName() }">
+                                            <a href="#" data-bind="if: !isServerWide(), click: editTask"><span data-bind="text: taskName"></span></a>
+                                            <a data-bind="if: isServerWide()"><span data-bind="text: taskName"></span></a>
```
ml054: can we avoid using `a` tag when it doesn't have href?

ml054 (changes requested):
please have a look at my comment

src/Raven.Studio/wwwroot/App/views/database/tasks/ongoingTasks.html
Danielle9897: If I use `span `instead of `a `then text will be 'gray', not 'red',
and it will look _different_ from the other tasks.
If that is ok - then I will change to `span`.

src/Raven.Server/ServerWide/Commands/PeriodicBackup/UpdatePeriodicBackupCommand.cs
```diff
@@ -41,7 +43,7 @@ public override string UpdateDatabaseRecord(DatabaseRecord record, long etag)
             }
             else if (Configuration.Name.StartsWith(ServerWideBackupConfiguration.NamePrefix, StringComparison.OrdinalIgnoreCase))
             {
-                throw new InvalidOperationException($"Can't update task name '{Configuration.Name}', because it is a server-wide backup task");
+                throw new InvalidOperationException($"Can't {(newTask ? "create" : "update")} task: '{Configuration.Name}'. A regular (non server-wide) backup task name can't start with prefix 'server wide backup'");
```
ppekrol: use `ServerWideBackupConfiguration.NamePrefix` instead 'server wide backup' `with prefix 'server wide backup'`

ppekrol (commented):
please also check failing tests

src/Raven.Server/ServerWide/Commands/PeriodicBackup/UpdatePeriodicBackupCommand.cs
Danielle9897: done + fixed tests

ppekrol (approved):

PR #9740, opened 2019-10-03

karmeli87:


ppekrol (approved):

ppekrol:
We will handle that in a different way. Karmel will open a new PR.

PR #9744, opened 2019-10-03

karmeli87:
- Merge `ValueChanged` and `StateChange` to have a single code path, because notifying about raft index must mean that all the preceding indexes were executed as well.
This might not be the case when we notifiy seperatly on value and recrods changes.

- Track the raft commands per database and act accordingly.

### DIFF
--- a/src/Raven.Server/Documents/DocumentDatabase.cs
+++ b/src/Raven.Server/Documents/DocumentDatabase.cs
@@ -239,12 +239,12 @@ namespace Raven.Server.Documents
 
         public StudioConfiguration StudioConfiguration { get; private set; }
 
-        private long _lastRaftIndexForStateChange;
+        private long _lastDatabaseRecordIndex = -1;
 
-        public long LastRaftIndexForStateChange
+        public long LastDatabaseRecordIndex
         {
-            get => Volatile.Read(ref _lastRaftIndexForStateChange);
-            private set => _lastRaftIndexForStateChange = value; // we write this always under lock
+            get => Volatile.Read(ref _lastDatabaseRecordIndex);
+            private set => _lastDatabaseRecordIndex = value; // we write this always under lock
         }
 
         public bool CanUnload => Interlocked.Read(ref _preventUnloadCounter) == 0;
@@ -319,7 +319,7 @@ namespace Raven.Server.Documents
                 {
                     try
                     {
-                        NotifyFeaturesAboutStateChange(record, index, ClusterStateMachine.All);
+                        NotifyFeaturesAboutStateChange(record, index, ClusterStateMachine.DatabaseRecordChange.All);
                     }
                     catch
                     {
@@ -1066,7 +1066,7 @@ namespace Raven.Server.Documents
 
                 StudioConfiguration = record.Studio;
 
-                NotifyFeaturesAboutStateChange(record, index);
+                NotifyFeaturesAboutStateChange(record, index, ClusterStateMachine.DatabaseRecordChange.None);
 
                 RachisLogIndexNotifications.NotifyListenersAbout(index, null);
             }
@@ -1089,9 +1089,9 @@ namespace Raven.Server.Documents
             }
         }
         
-        private void NotifyFeaturesAboutStateChange(DatabaseRecord record, long index , ClusterStateMachine.DatabaseRecordChange changes = ClusterStateMachine.DatabaseRecordChange.None)
+        private void NotifyFeaturesAboutStateChange(DatabaseRecord record, long index , ClusterStateMachine.DatabaseRecordChange changes)
         {
-            if (CanSkipStateChange(record.DatabaseName, index))
+            if (CanSkip(record.DatabaseName, index))
                 return;
 
             var taken = false;
@@ -1100,7 +1100,7 @@ namespace Raven.Server.Documents
                 Monitor.TryEnter(_clusterLocker, TimeSpan.FromSeconds(5), ref taken);
                 try
                 {
-                    if (CanSkipStateChange(record.DatabaseName, index))
+                    if (CanSkip(record.DatabaseName, index))
                         return;
 
                     if (taken == false)
@@ -1112,14 +1112,13 @@ namespace Raven.Server.Documents
                         $"{Name} != {record.DatabaseName}");
 
                     if (_logger.IsInfoEnabled)
-                        _logger.Info($"Starting to process record {index} (last {LastRaftIndexForStateChange}) for {record.DatabaseName}.");
+                        _logger.Info($"Starting to process record {index} (last {LastDatabaseRecordIndex}) for {record.DatabaseName}.");
 
                     try
                     {
                         DatabaseGroupId = record.Topology.DatabaseTopologyIdBase64;
 
-                        if (changes == ClusterStateMachine.DatabaseRecordChange.None && 
-                            ServerStore.Cluster.CommandsHolder.TryGetValue(Name, out var queue))
+                        if (ServerStore.Cluster.ChangesHolder.TryGetValue(Name, out var queue))
                         {
                             while (queue.TryPeek(out var result) && result.Index <= index)
                             {
@@ -1150,7 +1149,7 @@ namespace Raven.Server.Documents
                             InitializeFromDatabaseRecord(record);
                         }
 
-                        if (changes.HasFlag(ClusterStateMachine.DatabaseRecordChange.ETL))
+                        if (changes.HasFlag(ClusterStateMachine.DatabaseRecordChange.Etl))
                         {
                             EtlLoader?.HandleDatabaseRecordChange(record);
                         }
@@ -1162,7 +1161,7 @@ namespace Raven.Server.Documents
 
                         OnDatabaseRecordChanged(record);
 
-                        LastRaftIndexForStateChange = index;
+                        LastDatabaseRecordIndex = index;
 
                         if (_logger.IsInfoEnabled)
                             _logger.Info($"Finish to process record {index} for {record.DatabaseName}.");
@@ -1199,13 +1198,13 @@ namespace Raven.Server.Documents
             }
         }
 
-        private bool CanSkipStateChange(string database, long index)
+        private bool CanSkip(string database, long index)
         {
-            if (LastRaftIndexForStateChange > index)
+            if (LastDatabaseRecordIndex > index)
             {
-                // index and LastRaftIndexForStateChange could have equal values when we transit from/to passive and want to update the tasks. 
+                // index and LastDatabaseRecordIndex could have equal values when we transit from/to passive and want to update the tasks. 
                 if (_logger.IsInfoEnabled)
-                    _logger.Info($"Skipping record state change in index {index} (current {LastRaftIndexForStateChange}) for {database} because it was already precessed.");
+                    _logger.Info($"Skipping record state change in index {index} (current {LastDatabaseRecordIndex}) for {database} because it was already precessed.");
                 return true;
             }
 
@@ -1224,7 +1223,7 @@ namespace Raven.Server.Documents
             {
                 record = _serverStore.Cluster.ReadDatabase(context, Name, out index);
             }
-            NotifyFeaturesAboutStateChange(record, index, ClusterStateMachine.All);
+            NotifyFeaturesAboutStateChange(record, index, ClusterStateMachine.DatabaseRecordChange.All);
         }
 
         private void InitializeFromDatabaseRecord(DatabaseRecord record)
--- a/src/Raven.Server/Documents/DocumentDatabase.cs
+++ b/src/Raven.Server/Documents/DocumentDatabase.cs
@@ -239,12 +239,12 @@ namespace Raven.Server.Documents
 
         public StudioConfiguration StudioConfiguration { get; private set; }
 
-        private long _lastDatabaseRecordIndex;
+        private long _lastRaftIndexForStateChange;
 
-        public long LastDatabaseRecordIndex
+        public long LastRaftIndexForStateChange
         {
-            get => Volatile.Read(ref _lastDatabaseRecordIndex);
-            private set => _lastDatabaseRecordIndex = value; // we write this always under lock
+            get => Volatile.Read(ref _lastRaftIndexForStateChange);
+            private set => _lastRaftIndexForStateChange = value; // we write this always under lock
         }
 
         public bool CanUnload => Interlocked.Read(ref _preventUnloadCounter) == 0;
@@ -319,7 +319,7 @@ namespace Raven.Server.Documents
                 {
                     try
                     {
-                        NotifyFeaturesAboutStateChange(record, index);
+                        NotifyFeaturesAboutStateChange(record, index, ClusterStateMachine.All);
                     }
                     catch
                     {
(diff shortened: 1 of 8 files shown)


src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -1080,10 +1051,11 @@ public void StateChanged(long index)
                     ThrowDatabaseShutdown();
 
                 DatabaseRecord record;
+                long realIndex;
                 using (_serverStore.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
                 using (context.OpenReadTransaction())
                 {
-                    record = _serverStore.Cluster.ReadDatabase(context, Name);
+                    record = _serverStore.Cluster.ReadDatabase(context, Name, out realIndex);
```
grisha-kotler: Are you using the realIndex?

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -1115,10 +1088,10 @@ public void StateChanged(long index)
                 throw;
             }
         }
-
-        private void NotifyFeaturesAboutStateChange(DatabaseRecord record, long index)
+        
+        private void NotifyFeaturesAboutStateChange(DatabaseRecord record, long index , ClusterStateMachine.DatabaseRecordChange changes = ClusterStateMachine.DatabaseRecordChange.None)
```
ppekrol: not sure if having default value here is a good idea, someone will forget about it in the future at we will end up with 'none'

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -1139,20 +1112,57 @@ private void NotifyFeaturesAboutStateChange(DatabaseRecord record, long index)
...
+                        _logger.Info($"Starting to process record {index} (last {LastRaftIndexForStateChange}) for {record.DatabaseName}.");
 
                     try
                     {
                         DatabaseGroupId = record.Topology.DatabaseTopologyIdBase64;
 
+                        if (changes == ClusterStateMachine.DatabaseRecordChange.None && 
+                            ServerStore.Cluster.CommandsHolder.TryGetValue(Name, out var queue))
+                        {
+                            while (queue.TryPeek(out var result) && result.Index <= index)
```
ppekrol, line 1123: why not `TryDequeue`?

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -1139,20 +1112,57 @@ private void NotifyFeaturesAboutStateChange(DatabaseRecord record, long index)
...
+                        {
+                            PeriodicBackupRunner.UpdateConfigurations(record);
+                        }
+
+                        if (changes.HasFlag(ClusterStateMachine.DatabaseRecordChange.Configuration))
+                        {
+                            InitializeFromDatabaseRecord(record);
+                        }
+
+                        if (changes.HasFlag(ClusterStateMachine.DatabaseRecordChange.ETL))
```
ppekrol: `Etl`

src/Raven.Server/Documents/ETL/EtlLoader.cs
```diff
@@ -474,9 +474,14 @@ public void HandleDatabaseRecordChange(DatabaseRecord record)
                     }
                 }
             });
+
+            if (_processes.Length > 0)
```
ppekrol, line 478: @arekpalinski can you review this? LGTM

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1696,17 +1703,142 @@ private void NotifyValueChanged(TransactionOperationContext context, string type
             };
         }
 
+        [Flags]
+        public enum DatabaseRecordChange
+        {
+            None = 0,
+            Index = 1,
```
ppekrol, line 1713: use `<<` check e.g. `DatabaseItemType` enum

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1696,17 +1703,142 @@ private void NotifyValueChanged(TransactionOperationContext context, string type
...
+            Backup = 2,
+            Replication = 4,
+            ETL = 8,
+            Subscription = 16,
+            Configuration = 32,
+        }
+
+        public static DatabaseRecordChange All = DatabaseRecordChange.Index |
```
ppekrol: move it inside `enum`

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1696,17 +1703,142 @@ private void NotifyValueChanged(TransactionOperationContext context, string type
...
+        }
+
+        public static DatabaseRecordChange All = DatabaseRecordChange.Index | 
+                                                 DatabaseRecordChange.Backup | 
+                                                 DatabaseRecordChange.Replication | 
+                                                 DatabaseRecordChange.ETL | 
+                                                 DatabaseRecordChange.Subscription | 
+                                                 DatabaseRecordChange.Configuration;
+
+        public static DatabaseRecordChange TopologyAffected = DatabaseRecordChange.Backup |
```
ppekrol: move it inside enum

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1696,17 +1703,142 @@ private void NotifyValueChanged(TransactionOperationContext context, string type
...
+        public ConcurrentDictionary<string, ConcurrentQueue<(long Index, DatabaseRecordChange Change)>> CommandsHolder =
+            new ConcurrentDictionary<string, ConcurrentQueue<(long Index, DatabaseRecordChange Change)>>();
+
         private void NotifyDatabaseAboutChanged(TransactionOperationContext context, string databaseName, long index, string type, DatabasesLandlord.ClusterDatabaseChangeType change)
         {
             context.Transaction.InnerTransaction.LowLevelTransaction.OnDispose += transaction =>
             {
                 if (transaction is LowLevelTransaction llt && llt.Committed)
                 {
+                    var queue = CommandsHolder.GetOrAdd(databaseName, new ConcurrentQueue<(long Index, DatabaseRecordChange Type)>());
```
ppekrol: use action `() => new ConcurrentQueue(...)`

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1696,17 +1703,142 @@ private void NotifyValueChanged(TransactionOperationContext context, string type
...
+                case nameof(UpdatePeriodicBackupStatusCommand):
+                    return DatabaseRecordChange.None;
+
+                case nameof(RemoveNodeFromDatabaseCommand):
+                case nameof(RemoveNodeFromClusterCommand):
+                case nameof(AddDatabaseCommand):
+                case nameof(UpdateTopologyCommand):
+                case nameof(DeleteDatabaseCommand):
+                case nameof(PromoteDatabaseNodeCommand):
+                case "ClusterTopologyChanged":
```
ppekrol: no `nameof` ?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -979,12 +980,18 @@ private void RemoveNodeFromCluster(TransactionOperationContext context, Blittabl
...
+        private void ExecuteManyOnDispose(TransactionOperationContext context, long index, string type, List<Action> actions, List<string> affectedDatabases)
         {
             context.Transaction.InnerTransaction.LowLevelTransaction.OnDispose += transaction =>
             {
                 if (transaction is LowLevelTransaction llt && llt.Committed)
                 {
+                    foreach (var database in affectedDatabases)
+                    {
+                        var queue = CommandsHolder.GetOrAdd(database, new ConcurrentQueue<(long Index, DatabaseRecordChange Type)>());
+                        queue.Enqueue((index, GetChangeType(type)));
```
ppekrol: Is there a chance that those 'indexes' will not be ordered?

src/Raven.Server/Documents/DocumentDatabase.cs
ppekrol: Nvm here, it was late :)

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1696,17 +1703,142 @@ private void NotifyValueChanged(TransactionOperationContext context, string type
...
+
+                case nameof(ModifyConflictSolverCommand):
+                case nameof(UpdateExternalReplicationCommand):
+                case nameof(UpdatePullReplicationAsHubCommand):
+                case nameof(UpdatePullReplicationAsSinkCommand):
+                    return DatabaseRecordChange.Replication;
+
+                case nameof(ToggleTaskStateCommand):
+                case nameof(DeleteOngoingTaskCommand):
+                    return DatabaseRecordChange.Backup |
```
ppekrol: Maybe introduce `OngoingTasks` in enum?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1696,17 +1703,142 @@ private void NotifyValueChanged(TransactionOperationContext context, string type
...
+                case nameof(IncrementClusterIdentityCommand):
+                case nameof(IncrementClusterIdentitiesBatchCommand):
+                case nameof(UpdateClusterIdentityCommand):
+                case nameof(AddOrUpdateCompareExchangeCommand):
+                case nameof(RemoveCompareExchangeCommand):
+                case nameof(UpdateSnmpDatabaseIndexesMappingCommand):
+
+                case nameof(UpdateExternalReplicationStateCommand):
+                case nameof(AcknowledgeSubscriptionBatchCommand):
+                case nameof(UpdateEtlProcessStateCommand):
```
arekpalinski, line 1774: Why `UpdateEtlProcessStateCommand` is handled as `DatabaseRecordChange.None;` ? I think we should check if we need to reset ETL so I think  `EtlLoader.Reset` should be called then?

https://github.com/ravendb/ravendb/blob/832bdb96a6f29ae4159b270925ca74b23eb29df0/src/Raven.Server/Documents/ETL/EtlLoader.cs#L486-L491

src/Raven.Server/Documents/ETL/EtlLoader.cs
arekpalinski: Just to be sure - @karmeli87 this is because we no longer have a distinction between `StateChange` and `ValueChange`, right? 

Although, wouldn't be better off by having dedicated `DatabaseRecordChange.EtlStateChange` and then separate flow for that. Please check my comment below. If you agree that introducing dedicated change for that it's good idea then we should have the same for Subsctiptions - a subscription state is handled the same way

src/Raven.Server/Documents/DocumentDatabase.cs
karmeli87: done

src/Raven.Server/Documents/DocumentDatabase.cs
karmeli87: no, done.

src/Raven.Server/Documents/DocumentDatabase.cs
karmeli87: done

src/Raven.Server/ServerWide/ClusterStateMachine.cs
karmeli87: You are correct. `OnDispose` is called after we release the write lock!

src/Raven.Server/ServerWide/ClusterStateMachine.cs
karmeli87: done

src/Raven.Server/ServerWide/ClusterStateMachine.cs
karmeli87: done

src/Raven.Server/ServerWide/ClusterStateMachine.cs
karmeli87: done

src/Raven.Server/ServerWide/ClusterStateMachine.cs
karmeli87: done

src/Raven.Server/ServerWide/ClusterStateMachine.cs
karmeli87: Resetting an ETL is done by `RemoveEtlProcessStateCommand`.

src/Raven.Server/ServerWide/ClusterStateMachine.cs
karmeli87: done

src/Raven.Server/ServerWide/ClusterStateMachine.cs
karmeli87: it is now under `TopologyAffected`

src/Raven.Server/Documents/ETL/EtlLoader.cs
karmeli87: If this change is (logically) correct and doesn't worsen performance, I think we should do a finer granularity later.

src/Raven.Server/ServerWide/ClusterStateMachine.cs
arekpalinski: You're right

src/Raven.Server/Documents/ETL/EtlLoader.cs
arekpalinski: Current implementation is correct. Introducing finer granularity is an enhancement that can be implemented later.

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1696,8 +1706,41 @@ private void NotifyValueChanged(TransactionOperationContext context, string type
...
+        {
+            None = 0,
+            Index = 1,
+            Backup = 1 << 1,
+            Replication = 1 << 2,
+            Etl = 1 << 3,
+            Subscription = 1 << 4,
+            Configuration = 1 << 5,
+
+            All = Index |
```
ayende, line 1720: We need test / debug assert to verify that we aren't adding a new type here and not updating the `All`

ayende (approved):

ppekrol (approved):
@karmeli87 let's create an issue to do a separation in the future

karmeli87:
This is large and complex change that touch many places in a very fundamental component of RavenDB, my main concern is that this will break the current behavior of features and the interaction between them.

For now reverting to this commit https://github.com/ravendb/ravendb/commit/e28daed194af8723c9c611cfcd68fe9fa0258a92
 will be in my opinion the best option, since it fixes the original issue and doesn't change current behavior. 

As for the issue of "notifying about raft index must mean that all the preceding indexes were executed as well", this is relevante for `RachisLogIndexNotifications` on a _database_ level. 
Either we should completely drop it, or use the same technique as we did for the cluster state machine.

Performance benefits of the above PR can go later into v5.0

PR #9753, opened 2019-10-06
RavenDB-14043 & RavenDB-14053

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-14043
https://issues.hibernatingrhinos.com/issue/RavenDB-14053

### DIFF
--- a/src/Raven.Server/Utils/Cpu/CpuHelper.cs
+++ b/src/Raven.Server/Utils/Cpu/CpuHelper.cs
@@ -92,9 +92,11 @@ namespace Raven.Server.Utils.Cpu
             }
         }
 
+        private static bool _timesSyscallSupported = true;
+
         private static long GetTotalProcessorTimeInTicks(Process process)
         {
-            if (PlatformDetails.RunningOnLinux == false)
+            if (PlatformDetails.RunningOnLinux == false || _timesSyscallSupported == false)
                 return process.TotalProcessorTime.Ticks;
 
             // Linux: TotalProcessorTime is calling to: proc/{processId}/stat and parsing the whole result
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Utils/Cpu/CpuHelper.cs
```diff
@@ -92,48 +92,47 @@ public static (long TotalProcessorTimeTicks, long TimeTicks) GetProcessTimes(Pro
...
+                    stime = timeSample.tms_stime;
+                    utime = timeSample.tms_utime;
                 }
 
-                return (TotalProcessorTimeTicks: tmsUtime + tmsStime, TimeTicks: timeTicks);
+                return utime + stime;
             }
             catch (Exception e)
             {
+                // some linux distros don't support this call
```
ayende, line 129: what would be the exception in this case?

src/Raven.Server/Utils/ThreadsUsage.cs
```diff
@@ -98,6 +100,10 @@ public ThreadsInfo Calculate()
                     {
                         // thread has exited
                     }
+                    catch (Win32Exception e) when (e.Message.Contains("Access is denied", StringComparison.OrdinalIgnoreCase))
```
ayende: That is a bad idea. Windows messages are localized.
What is the error code here?

ayende (changes requested):

src/Raven.Server/Utils/ThreadsUsage.cs
grisha-kotler: done

ppekrol (approved):

ayende (approved):

PR #9761, opened 2019-10-10
RavenDB-13432 & RavenDB-14034

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-13432
https://issues.hibernatingrhinos.com/issue/RavenDB-14034

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -3087,7 +3087,7 @@ namespace Raven.Server.Documents.Indexes
 
             CalculateIndexEtagInternal(indexEtagBytes, isStale, State, documentsContext, indexContext);
 
-            UseAllDocumentsCounterAndCmpXngEtags(documentsContext, q, length, indexEtagBytes);
+            UseAllDocumentsCounterAndCmpXchgEtags(documentsContext, q, length, indexEtagBytes);
 
             unchecked
             {
@@ -3095,7 +3095,7 @@ namespace Raven.Server.Documents.Indexes
             }
         }
 
-        protected static unsafe void UseAllDocumentsCounterAndCmpXngEtags(DocumentsOperationContext documentsContext, 
+        protected static unsafe void UseAllDocumentsCounterAndCmpXchgEtags(DocumentsOperationContext documentsContext, 
             QueryMetadata q, int length, byte* indexEtagBytes)
         {
             if (q == null)
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -3087,7 +3087,7 @@ namespace Raven.Server.Documents.Indexes
 
             CalculateIndexEtagInternal(indexEtagBytes, isStale, State, documentsContext, indexContext);
 
-            UseAllDocumentsAndCounterEtag(documentsContext, q, length, indexEtagBytes);
+            UseAllDocumentsCounterAndCmpXngEtags(documentsContext, q, length, indexEtagBytes);
 
             unchecked
             {
@@ -3095,7 +3095,8 @@ namespace Raven.Server.Documents.Indexes
             }
         }
 
-        protected static unsafe void UseAllDocumentsAndCounterEtag(DocumentsOperationContext documentsContext, QueryMetadata q, int length, byte* indexEtagBytes)
+        protected static unsafe void UseAllDocumentsCounterAndCmpXngEtags(DocumentsOperationContext documentsContext, 
+            QueryMetadata q, int length, byte* indexEtagBytes)
         {
             if (q?.HasIncludeOrLoad == true)
             {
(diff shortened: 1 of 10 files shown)


src/Raven.Server/ServerWide/Commands/CompareExchangeCommands.cs
```diff
@@ -103,6 +103,37 @@ protected CompareExchangeCommandBase(string database, string key, long index, Js
...
         }
 
+        public static unsafe void GetDbPrefixAndLastSlices(
+            ByteStringContext allocator, string db,
+            out (Slice Slice, ByteStringContext.ExternalScope Scope) prefix,
+            out (Slice Slice, ByteStringContext.ExternalScope Scope) last)
+        {
+            var reservedSpace = Encoding.UTF8.GetMaxByteCount(db.Length) + 1; 
+
+            using (allocator.Allocate(reservedSpace, out var prefixBuffer)) // db + '/'
```
ayende: You are returning the memory to the pool when you exit the `using` statement

ayende (changes requested):

grisha-kotler (approved):

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -3107,14 +3111,47 @@ protected static unsafe void UseAllDocumentsEtag(DocumentsOperationContext docum
                 //buffer[2] - last processed doc etag
                 //buffer[3] - last process tombstone etag
             }
-        }
 
-        protected int MinimumSizeForCalculateIndexEtagLength()
+            if (q.CounterIncludes != null || q.HasCounterSelect)
+            {
+                Debug.Assert(length > sizeof(long) * 5);
```
ppekrol: Add a debug message here, if not we will not know during test execution where it failed.

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -3107,14 +3111,47 @@ protected static unsafe void UseAllDocumentsEtag(DocumentsOperationContext docum
...
+
+                var offset = length - (sizeof(long) * (q.HasCmpXchg || q.HasCmpXchgSelect ? 2 : 1)) ;
+
+                *(long*)(indexEtagBytes + offset) = DocumentsStorage.ReadLastCountersEtag(documentsContext.Transaction.InnerTransaction);
+            }
+
+            if (q.HasCmpXchg || q.HasCmpXchgSelect)
+            {
+                Debug.Assert(length > sizeof(long) * 5);
```
ppekrol: Same here

ppekrol (approved):

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -3081,23 +3081,27 @@ internal static bool WillResultBeAcceptable(bool isStale, IndexQueryBase<Blittab
...
         {
-            var length = MinimumSizeForCalculateIndexEtagLength();
+            var length = MinimumSizeForCalculateIndexEtagLength(q);
 
             var indexEtagBytes = stackalloc byte[length];
 
             CalculateIndexEtagInternal(indexEtagBytes, isStale, State, documentsContext, indexContext);
 
-            UseAllDocumentsEtag(documentsContext, q, length, indexEtagBytes);
+            UseAllDocumentsCounterAndCmpXngEtags(documentsContext, q, length, indexEtagBytes);
```
arekpalinski:
```suggestion
            UseAllDocumentsCounterAndCmpXchgEtags(documentsContext, q, length, indexEtagBytes);
```

src/Raven.Server/ServerWide/Commands/CompareExchangeCommands.cs
```diff
@@ -103,6 +103,36 @@ protected CompareExchangeCommandBase(string database, string key, long index, Js
...
+                var len = Encoding.UTF8.GetBytes(pDb, db.Length, prefixBuffer.Ptr, prefixBuffer.Length);
+
+                prefixBuffer.Ptr[len] = (byte)'/';
+                prefixBuffer.Truncate(len + 1);
+
+                allocator.ToLowerCase(ref prefixBuffer);
+
+                prefixBuffer.CopyTo(lastBuffer.Ptr);
+
+                lastBuffer.Ptr[len++] = '/' + 1;
```
arekpalinski, line 127: We pass it to `SeekOneBackwardFrom` method. In order to make sure it works as expected we need tests which have more than one database and multiple compare exchange values for each. This way we'll know that we get correct values as expected.

test/SlowTests/Issues/RavenDB_14034.cs
```diff
@@ -0,0 +1,604 @@
...
+                            var lastCmpXchgIndex = documentDatabase.ServerStore.Cluster
+                                .GetLastCompareExchangeIndexForDatabase(transactionContext, db);
+
+                            Assert.Equal(expectedIndex, lastCmpXchgIndex);
+                        }
+                    }
+                }
+            }
+        }
+
```
arekpalinski, line 578: We need a test with multiple databases - see my comment above

aviv86: This test case has 2 databases and 10 compare exchange values on each db.
Should I test with more than 2 databases?

arekpalinski: I didn't notice it. That's great then.

Small note - instead of using `CreateDatabaseOperation` you can just call `using (var store = GetDocumentStore())` one more time. This way the database would be automatically deleted after the test. Right now the database remains on the server (most tests uses shared server instance)

arekpalinski (approved):

PR #9762, opened 2019-10-10
RavenDB-14114 - Reduce the number of log messages in ReduceMemoryUsage

grisha-kotler:


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1418,8 +1415,13 @@ private void ReduceMemoryUsage(StorageEnvironment environment)
                 _currentMaximumAllowedMemory = DefaultMaximumMemoryAllocation;
 
                 var afterFree = NativeMemory.CurrentThreadStats.TotalAllocated;
-                if (_logger.IsInfoEnabled)
-                    _logger.Info($"After cleanup, using {new Size(afterFree, SizeUnit.Bytes)} by '{Name}'.");
+
```
karmeli87, line 1418: Do we want to check the previous `afterFree` with the current `beforeFree`, to skip all of those `Clean` methods?

ppekrol: Not sure that I follow? You can get `afterFree` only you execute them.

karmeli87: From the previous run

ppekrol: cc @grisha-kotler ?

grisha-kotler: @karmeli87 good idea.
https://github.com/ravendb/ravendb/pull/9781

PR #9767, opened 2019-10-11
 RavenDB-11488 Side by side indexing for map-reduce indexes with OutputReduceToCollection.

arekpalinski:
There is no longer a need to delete artificial documents before changing index definition.


src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -378,6 +418,29 @@ private void HandleStaticIndexChange(string name, IndexDefinition definition)
             }
         }
 
+        private static void CollectPrefixesOfDocumentsToDelete(MapReduceIndex mapReduceIndex, ref HashSet<string> prefixesOfDocumentsToDelete)
```
ayende, line 421: Why ref HashSet here? It is a class

ayende:
Looks okay, otherwise

src/Raven.Server/Documents/Indexes/IndexStore.cs
arekpalinski: I like to mark it explicitly that it's modified inside the method this way

ayende (approved):

PR #9773, opened 2019-10-15

karmeli87:


src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -3101,6 +3103,33 @@ public async Task WaitForIndexNotification(long index, CancellationToken token)
...
+
+            var task = tcs.Task;
+
+            if (task.IsCompleted)
+                return;
+
+            var result = await Task.WhenAny(task, waitAsync);
+
+            if (result.IsFaulted)
+                throw result.Exception;
```
ayende: Wrap this in another exception. You are modifying the exception stack trace here.

test/SlowTests/Cluster/ClusterIndexNotificationsTest.cs
```diff
@@ -93,6 +95,22 @@ public async Task ShouldThrowTimeoutException()
...
+            {
+                var indexes = new List<Task>();
+                for (int i = 0; i < 100; i++)
+                {
+                    var index = new Index();
+                    indexes.Add(index.ExecuteAsync(store));
+                }
+
+                await Task.WhenAll(indexes);
+            }
```
ayende, line 114: Don't you need to verify that they all exist here?

ayende (changes requested):

src/Raven.Server/ServerWide/ClusterStateMachine.cs
karmeli87: Right... maybe `await result` is a cleaner solution?

test/SlowTests/Cluster/ClusterIndexNotificationsTest.cs
karmeli87: done

src/Raven.Server/ServerWide/ClusterStateMachine.cs
arekpalinski: +1 for `await result`

ayende (approved):

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -1219,16 +1242,17 @@ private void NotifyFeaturesAboutValueChange(DatabaseRecord record, long index)
...
 
                     if (taken == false)
                         continue;
                 
-                    LastDatabaseRecordIndex = index;
                     DatabaseShutdown.ThrowIfCancellationRequested();
                     SubscriptionStorage?.HandleDatabaseRecordChange(record);
                     EtlLoader?.HandleDatabaseValueChanged(record);
+
+                    LastDatabaseRecordChangeIndex = index;
```
ppekrol: Shouldn't this be `LastValueChangeIndex` ?

karmeli87: You are absolutely correct! I will add a test for this.

ppekrol (approved):

PR #9778, opened 2019-10-16
RavenDB-14129 added /monitoring/snmp/bulk endpoint

ppekrol:


src/Raven.Server/Monitoring/Snmp/SnmpHandler.cs
```diff
@@ -44,6 +41,41 @@ public Task Get()
...
+
+        [RavenAction("/monitoring/snmp/bulk", "POST", AuthorizationStatus.Operator)]
+        public async Task PostBulk()
+        {
+            AssertSnmp();
+
+            using (ServerStore.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
+            {
+                var json = await context.ReadForMemoryAsync(RequestBodyStream(), "oids");
+                if (json.TryGet("Oids", out BlittableJsonReaderArray array) == false)
```
ayende: `Oids` here

src/Raven.Server/Monitoring/Snmp/SnmpHandler.cs
```diff
@@ -67,5 +99,68 @@ public Task GetOids()
...
+                foreach (var result in results)
+                {
+                    if (first == false)
+                        writer.WriteComma();
+
+                    first = false;
+
+                    writer.WriteStartObject();
+
+                    writer.WritePropertyName("OID");
```
ayende, line 139: But `OID` here, the casing is not consistent

ayende (changes requested):

src/Raven.Server/Monitoring/Snmp/SnmpHandler.cs
ppekrol: Done

PR #9784, opened 2019-10-17
RavenDB-14146 - Ignore TaskCanceledException on database shutdown

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-14146

### DIFF
--- a/src/Raven.Server/Documents/PeriodicBackup/BackupUploader.cs
+++ b/src/Raven.Server/Documents/PeriodicBackup/BackupUploader.cs
@@ -233,16 +233,16 @@ namespace Raven.Server.Documents.PeriodicBackup
                 catch (Exception e)
                 {
                     var extracted = e.ExtractSingleInnerException();
+                    var error = $"Failed to upload the backup file to {targetName}.";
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Documents/PeriodicBackup/BackupUploader.cs
```diff
@@ -226,14 +230,17 @@ private string CombinePathAndKey(string path)
...
-                catch (OperationCanceledException e)
-                {
-                    // shutting down
-                    localUploadStatus.Exception = e.ToString();
-                    _exceptions.Add(e);
-                }
                 catch (Exception e)
                 {
+                    var extracted = e.ExtractSingleInnerException();
+                    if (extracted is OperationCanceledException)
```
ppekrol, line 238: aren't we hiding potential exception from HttpClient here? Shouldn't we check that databaseShutdownToken is cancelled?

grisha-kotler: done

ppekrol (approved):

PR #9786, opened 2019-10-17
RavenDB-14127 Unify database and server rachis notifications [WIP]

karmeli87:


karmeli87:
Eventually we will unify database and server rachis notifications.

The problem is that upon failure we don't know which change is the problematic one, since we merge and scan for changes in all features for a database record change.

So we need to implement something similar as started here: https://github.com/ravendb/ravendb/pull/9744

PR #9795, opened 2019-10-21
RDBC-392 remove whitespace from nodeUrl and nodeTag when adding a node

ppekrol:


src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
```diff
@@ -313,6 +313,7 @@ public async Task AddNode()
             if (assignedCores <= 0)
                 throw new ArgumentException("Assigned cores must be greater than 0!");
 
+            nodeUrl = nodeUrl.Trim();
```
trevhunter, line 316: @ppekrol I would suggest applying a full validation of the URL using something like `Uri.IsWellFormedUriString()` to see if there is anything else wrong with it other than just white space.

ppekrol: Done here: https://github.com/ravendb/ravendb/pull/9801

PR #9811, opened 2019-10-23
RavenDB-12062 ACME V2 breaking changes

iftahbe:


src/Raven.Server/Commercial/LetsEncryptClient.cs
```diff
@@ -761,9 +783,15 @@ public JwsMessage Encode<TPayload>(TPayload payload, JwsHeader protectedHeader)
...
+                // From: https://tools.ietf.org/html/rfc8555#section-7.5.1
+                var encodedPayload = "{}".Equals(payload)
+                    ? Base64UrlEncoded("{}")
+                    : Base64UrlEncoded(JsonConvert.SerializeObject(payload));
+
                 var message = new JwsMessage
                 {
-                    Payload = Base64UrlEncoded(JsonConvert.SerializeObject(payload)),
+				    // From: https://tools.ietf.org/html/rfc8555#section-6.3
+                    Payload =  string.Empty.Equals(payload) ? string.Empty : encodedPayload,
```
ayende: Put this logic with the encodedPayload, so we'll have a `case` for the various options

ayende (changes requested):

src/Raven.Server/Commercial/LetsEncryptClient.cs
```diff
@@ -295,14 +315,13 @@ public async Task CompleteChallenges(CancellationToken token = default(Cancellat
...
-                        KeyAuthorization = _jws.GetKeyAuthorization(challenge.Token)
-                    }, token);
+				    // From: https://tools.ietf.org/html/rfc8555#section-7.5.1
+                    (result, responseText) = await SendAsync<AuthorizationChallengeResponse>(HttpMethod.Post, challenge.Url, "{}", token);
 
                     while (result.Status == "pending")
                     {
-                        (result, responseText) = await SendAsync<AuthorizationChallengeResponse>(HttpMethod.Get, challenge.Url, null, token);
+					    // post-as-get (https://community.letsencrypt.org/t/acme-v2-scheduled-deprecation-of-unauthenticated-resource-gets/74380)
+                        (result, responseText) = await SendAsync<AuthorizationChallengeResponse>(HttpMethod.Post, challenge.Url, string.Empty, token);
```
ppekrol, line 326: few lines up you have `{}` as content, not here. Is this correct?

src/Raven.Server/Commercial/LetsEncryptClient.cs
```diff
@@ -313,7 +332,8 @@ public async Task CompleteChallenges(CancellationToken token = default(Cancellat
                     string errorText = null;
                     try
                     {
-                        (err, errorText) = await SendAsync<AuthorizationChallengeResponse>(HttpMethod.Get, challenge.Url, null, token);
+					    // post-as-get (https://community.letsencrypt.org/t/acme-v2-scheduled-deprecation-of-unauthenticated-resource-gets/74380)
+                        (err, errorText) = await SendAsync<AuthorizationChallengeResponse>(HttpMethod.Post, challenge.Url, string.Empty, token);
```
ppekrol, line 338: same here. `{}` vs `string.Empty`

src/Raven.Server/Commercial/LetsEncryptClient.cs
iftahbe: yes that's correct. The first request of AuthorizationChallengeResponse should be a post with {} body.
Then all the subsequent request to that should be post with empty body.

ppekrol: Ok, thanks for explanation.

src/Raven.Server/Commercial/LetsEncryptClient.cs
```diff
@@ -783,15 +785,33 @@ public JwsMessage Encode<TPayload>(TPayload payload, JwsHeader protectedHeader)
                     protectedHeader.Key = _jwk;
                 }
 
-                // From: https://tools.ietf.org/html/rfc8555#section-7.5.1
-                var encodedPayload = "{}".Equals(payload)
-                    ? Base64UrlEncoded("{}")
-                    : Base64UrlEncoded(JsonConvert.SerializeObject(payload));
+                string encodedPayload;
+
+                if (payload is string p)
```
iftahbe: Please don't merge, this is not good. Fixing.

PR #9813, opened 2019-10-23
RavenDB-14169 - add support for types in Metadata

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-14169

### DIFF
--- a/src/Raven.Client/Json/BlittableJsonWriter.cs
+++ b/src/Raven.Client/Json/BlittableJsonWriter.cs
@@ -206,6 +206,9 @@ namespace Raven.Client.Json
                 case int i:
                     _manualBlittableJsonDocumentBuilder.WriteValue(i);
                     break;
+                case short s:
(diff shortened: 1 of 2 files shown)


src/Raven.Client/Json/BlittableJsonWriter.cs
```diff
@@ -203,6 +203,12 @@ private void WritePropertyValue(string propName, object value)
                 case long l:
                     _manualBlittableJsonDocumentBuilder.WriteValue(l);
                     break;
+                case int i:
```
ppekrol, line 212: There are couple of others that are missing here that I can think of e.g. `uint`, `ushort` etc

Is there a chance that they were not added by purpose? cc @ayende do you recall maybe if there was such a decision?

ayende: No such reason, we just didn't add them.

grisha-kotler: Added more types

PR #9829, opened 2019-10-27
RavenDB-14127 properly notify listeners about database record apply

karmeli87:
- Don't use `ContinueWith`, but wait for the the database to be loaded before notifying success.
- Rethrow exceptions, so they will be properly notified.

### DIFF
--- a/src/Raven.Server/Documents/DatabasesLandlord.cs
+++ b/src/Raven.Server/Documents/DatabasesLandlord.cs
@@ -5,6 +5,7 @@ using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
+using Nito.AsyncEx;
 using Raven.Client.Exceptions.Database;
 using Raven.Client.Extensions;
 using Raven.Client.ServerWide;
@@ -30,7 +31,7 @@ namespace Raven.Server.Documents
     public class DatabasesLandlord : IDisposable
     {
         public const string DoNotRemove = "DoNotRemove";
-        private readonly ReaderWriterLockSlim _disposing = new ReaderWriterLockSlim(LockRecursionPolicy.SupportsRecursion);
+        private readonly AsyncReaderWriterLock _disposing = new AsyncReaderWriterLock();
         public readonly ConcurrentDictionary<StringSegment, DateTime> LastRecentlyUsed =
             new ConcurrentDictionary<StringSegment, DateTime>(StringSegmentComparer.OrdinalIgnoreCase);
         private readonly ConcurrentDictionary<string, Timer> _wakeupTimers = new ConcurrentDictionary<string, Timer>();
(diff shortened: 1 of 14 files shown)


src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -108,54 +108,28 @@ private void HandleClusterDatabaseChanged(string databaseName, long index, strin
                 if (task.IsCanceled || task.IsFaulted)
                     return;
 
+                var database = task.Result;
```
ayende: This is probably a bad idea. What if we have a long database load time?
You cannot force a wait here.

ayende (changes requested):

src/Raven.Server/Documents/DatabasesLandlord.cs
karmeli87: `HandleClusterDatabaseChanged` method is executed asynchronously, so it is not blocking any other operation, but it might eat-up threads.. would `AsyncHelpers.RunSync(() => task)` help here?

ayende: No, would also hold us the thread.
Instead, change the execute to be async, so we can await there.

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -52,140 +53,113 @@ public DatabasesLandlord(ServerStore serverStore)
...
+                        // it. This is important so things like replication will start pumping, and that 
+                        // configuration changes such as running periodic backup will get a chance to run, which
+                        // they wouldn't unless the database is loaded / will have a request on it.          
+                        task = TryGetOrCreateResourceStore(databaseName);
+                    }
+
+                    if (task.IsCanceled || task.IsFaulted)
+                        return;
+
+                    var database = await task;
```
ayende, line 113: If this fails, the user is going to get an error, but does it impact anything else?

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -565,8 +490,22 @@ public Task<DocumentDatabase> TryGetOrCreateResourceStore(StringSegment database
...
+        }
+
+        private IDisposable EnterReadLockImmediately(StringSegment databaseName)
+        {
+            using (var cts = new CancellationTokenSource())
+            {
+                var awaiter = _disposing.ReaderLockAsync(cts.Token).GetAwaiter();
+                if (awaiter.IsCompleted == false)
+                {
+                    cts.Cancel();
```
ayende, line 504: I don't like this.
what happened if between the `IsCompleted` check and the `Cancel()`, you got the lock? Then you have a hanging lock here.

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -979,29 +1028,29 @@ private void RemoveNodeFromCluster(TransactionOperationContext context, Blittabl
...
+                context.Transaction.InnerTransaction.LowLevelTransaction.OnDispose += tx =>
+                {
                     var exceptionAggregator =
                         new ExceptionAggregator(_parent.Log, $"the raft index {index} is committed, but an error occured during executing the {type} command.");
 
-                    foreach (var action in actions)
+                    foreach (var task in tasks)
                     {
-                        TaskExecutor.Execute(_ =>
+                        Task.Run(async () =>
```
ayende, line 1051: This is going to block until the task is done running?

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -190,7 +190,16 @@ public async Task AddDatabaseNode()
...
-                await WaitForExecutionOnSpecificNode(context, clusterTopology, node, newIndex);
+                try
+                {
+                    await WaitForExecutionOnSpecificNode(context, clusterTopology, node, newIndex);
+                }
+                catch (DatabaseLoadFailureException e)
+                {
+                    // the node was added successfully, but failed to start 
+                    // in this case we don't want the request executor of the client to fail-over (so we wouldn't create an additional node)
+                    throw new InvalidOperationException(e.Message);
```
ayende:
```suggestion
                    throw new InvalidOperationException(e.Message, e);
```

ayende (changes requested):

src/Raven.Server/ServerWide/ClusterStateMachine.cs
karmeli87: No, it should execute the action inside the `Task.Run` asynchronously.

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
karmeli87: done

src/Raven.Server/Documents/DatabasesLandlord.cs
karmeli87: It might only impact on the server start-up, I guess it is better to create a task for each database.

src/Raven.Server/Documents/DatabasesLandlord.cs
karmeli87: done

ayende (approved):

PR #9830, opened 2019-10-28
RavenDB-14193

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-14193

### DIFF
--- a/Raven.Database/Indexing/ReducingExecuter.cs
+++ b/Raven.Database/Indexing/ReducingExecuter.cs
@@ -44,7 +44,10 @@ namespace Raven.Database.Indexing
             {
                 var mappedResultsInfo = actions.MapReduce.GetReduceTypesPerKeys(indexToWorkOn.IndexId,
                                                                     context.CurrentNumberOfItemsToReduceInSingleBatch, 
-                                                                    context.NumberOfItemsToExecuteReduceInSingleStep, token);
+                                                                    context.NumberOfItemsToExecuteReduceInSingleStep, message =>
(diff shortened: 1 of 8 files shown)


arekpalinski (approved):

ayende (approved):

Raven.Database/Indexing/ReducingExecuter.cs
```diff
@@ -645,15 +648,15 @@ protected override bool ShouldSkipIndex(Index index)
         protected override bool IsIndexStale(IndexStats indexesStat, IStorageActionsAccessor actions, bool isIdle, Reference<bool> onlyFoundIdleWork)
         {
             onlyFoundIdleWork.Value = false;
-            var isReduceStale = actions.Staleness.IsReduceStale(indexesStat.Id);
 
-            if (isReduceStale == false)
+            if (indexesStat.Priority.HasFlag(IndexingPriority.Disabled))
```
ppekrol, line 652: Just to double-check... This is used only to detect if reducing executer needs to do some work on that index or not? It is not used anywhere alse?

Also, do we have similar logic for IndexingExecuter?

grisha-kotler: It is used only in AbstractIndexingExecuter.
IndexingExecuter has the same logic:
https://github.com/ravendb/ravendb/blob/v3.5/Raven.Database/Indexing/IndexingExecuter.cs#L72

ppekrol (approved):

PR #9832, opened 2019-10-28
RavenDB-13990: Pass to Pal strings as UTF8 on linux and Unicode on Windows

aviviadi:


ayende (approved):

ppekrol:
@aviviadi Can we add test for that?

aviviadi:
We have validation in client which prevents us from entering non english characters.. Studio seems non enforcing these rules.

This validation, obviously, prevent unicode test to succeed.

PR #9834, opened 2019-10-29
RavenDB-14185 forcing newer Microsoft.Win32.Registry package

ppekrol:


tools/Raven.Debug/Raven.Debug.csproj
```diff
@@ -15,6 +15,7 @@
   <ItemGroup>
     <PackageReference Include="McMaster.Extensions.CommandLineUtils" Version="2.4.2" />
     <PackageReference Include="Microsoft.Diagnostics.Runtime-ravendb" Version="1.0.8" />
+    <PackageReference Include="Microsoft.Win32.Registry" Version="4.6.0" />
```
ayende, line 18: where do we even use this?

ppekrol: in Debug we do not, it is a part of runtime so when we are deploying standalone app the file is being included anyway. And since the Server was using newer version and Debug had the version from runtime, we encountered a DLL mismatch situation in our deployment package.

PR #9838, opened 2019-10-30
Raven db 14192 WIP, please review

maximburyak:
New tests don't pass yet, but I'm pretty sure I understand the problem now : subscriptions handling nodes becoming passive in the middle of a worker's lifecycle.

### DIFF
--- a/src/Raven.Server/Documents/Subscriptions/SubscriptionStorage.cs
+++ b/src/Raven.Server/Documents/Subscriptions/SubscriptionStorage.cs
@@ -140,18 +140,83 @@ namespace Raven.Server.Documents.Subscriptions
             {
                 var subscription = GetSubscriptionFromServerStore(serverStoreContext, name);
                 var topology = _serverStore.Cluster.ReadDatabaseTopology(serverStoreContext, _db.Name);
+                
                 var whoseTaskIsIt = _db.WhoseTaskIsIt(topology, subscription, subscription);
+                
                 if (whoseTaskIsIt != _serverStore.NodeTag)
                 {
+                    var databaseTopologyAvailabilityExplanation = new Dictionary<string, string>();
+
+                    string generalState = string.Empty;
+                    if (_serverStore.Engine.CurrentState == RachisState.Candidate || _serverStore.Engine.CurrentState == RachisState.Passive)
+                    {
+                        generalState = $"Current node ({_serverStore.NodeTag}) is in {_serverStore.Engine.CurrentState.ToString()} state therefore, we can't answer who's task is it and returning null";
+                    }
+                    else
+                    {
+                        generalState = _serverStore.Engine.CurrentState.ToString();
+                    }
+                    databaseTopologyAvailabilityExplanation["NodeState"] = generalState;                    
+
+                    FillNodesAvailabilityReportForState(subscription, topology, databaseTopologyAvailabilityExplanation, stateGroup:topology.Rehabs, stateName:"rehab");
+                    FillNodesAvailabilityReportForState(subscription, topology, databaseTopologyAvailabilityExplanation, stateGroup: topology.Promotables, stateName: "promotable");
+
+                    //whoseTaskIsIt!= null && whoseTaskIsIt == subscription.MentorNode 
+                    foreach (var member in topology.Members)
+                    {
+                        if (whoseTaskIsIt != null)
+                        {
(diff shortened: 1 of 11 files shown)


src/Raven.Server/Documents/Subscriptions/SubscriptionStorage.cs
```diff
@@ -140,19 +140,84 @@ public async Task<SubscriptionState> AssertSubscriptionConnectionDetails(long id
...
+                    FillNodesAvailabilityReportForState(subscription, topology, databaseTopologyAvailabilityExplanation, stateGroup: topology.Promotables, stateName: "promotable");
+
+                    //whoseTaskIsIt!= null && whoseTaskIsIt == subscription.MentorNode 
+                    foreach (var member in topology.Members)
+                    {
+                        if (whoseTaskIsIt != null)
+                        {
+                            if (whoseTaskIsIt == subscription.MentorNode && member == subscription.MentorNode)
+                            {
+                                databaseTopologyAvailabilityExplanation[member] = "Is the mentor node and a valid member of the topology, it should be the mento node";
```
ayende, line 171:
```suggestion
                                databaseTopologyAvailabilityExplanation[member] = "Is the mentor node and a valid member of the topology, it should be the mentor node";
```

Maybe should use: 'designated' instead of mentor, though?

src/Raven.Server/Documents/Subscriptions/SubscriptionStorage.cs
```diff
@@ -140,19 +140,84 @@ public async Task<SubscriptionState> AssertSubscriptionConnectionDetails(long id
...
+                                databaseTopologyAvailabilityExplanation[member] = "Is a valid member of the topology, but not chosen to be the node running the subscription";
+                            }
+                            else if (whoseTaskIsIt == member)
+                            {
+                                databaseTopologyAvailabilityExplanation[member] = "Is a valid member of the topology and is chosen to be running the subscription";
+                            }
+                        }                        
+                        else
+                        {
+                            databaseTopologyAvailabilityExplanation[member] = "Is a valid member of the topology but was not chosen to run the subscription, we didn't find any other match either";
```
ayende, line 184: Add here if we have a mentor node?

ayende: When can this happen?

src/Raven.Server/Routing/RouteInformation.cs
```diff
@@ -81,6 +81,8 @@ public void Build(MethodInfo action)
         public Task CreateDatabase(RequestHandlerContext context)
         {
             var databaseName = context.RouteMatch.GetCapture();
+
+            // todo: think if we need to pass this check to the landlord
```
ayende, line 85: Agreed, `AssertIsNotPassive`? Note that this should still happen here.

src/Raven.Server/Web/System/TcpConnectionInfoHandler.cs
```diff
@@ -77,6 +79,12 @@ public async Task GetRemoteTaskTcp()
             var remoteTask = GetStringQueryString("remote-task");
             var database = GetStringQueryString("database");
             var verifyDatabase = GetBoolValueQueryString("verify-database", false);
+
+            if (ServerStore.IsPassive())
```
ayende, line 83: This is not a good idea, note that this will mean that if the node is passive, we can't connect to it at all.
Including to be able to add to it to a cluster.
We need to check it only if this is for a particular db

ayende (changes requested):

src/Raven.Server/Documents/Subscriptions/SubscriptionStorage.cs
maximburyak: Mentor is the terminology in subscriptions

src/Raven.Server/Documents/Subscriptions/SubscriptionStorage.cs
maximburyak: When the node becomes passive in the middle of the subscription connection

src/Raven.Server/Web/System/TcpConnectionInfoHandler.cs
maximburyak: This endpoint is only for specific database, otherwise the endpoint /info/tcp will be used.
Note that in this endpoint, database is mandatory

PR #9840, opened 2019-10-30
RavenDB-12932 Allow to define a pattern for documents which reference IDs of documents from map-reduce output collection

arekpalinski:


src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
```diff
@@ -0,0 +1,365 @@
...
+            var djv = new DynamicJsonValue();
+
+            if (_index.OutputReduceToCollectionPropertyAccessor == null)
+                _index.OutputReduceToCollectionPropertyAccessor = PropertyAccessor.Create(reduceObject.GetType(), reduceObject);
+
+            OutputReferencesPattern.DocumentIdBuilder referenceDocIdBuilder = null;
+
+            string referenceDocumentId = null;
+
+            using (_patternForReduceOutputReferences?.BuildReferenceDocumentId(out referenceDocIdBuilder))
```
ayende, line 152: Wrap the whole thing in `if(_patternForReduceOutputReferences != null) ` ?

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
```diff
@@ -0,0 +1,365 @@
...
+                foreach (var item in outputs)
+                {
+                    item.Json.Dispose();
+                }
+            }
+        }
+
+        public static bool IsOutputDocumentPrefix(string prefix)
+        {
+            var parts = prefix.Split("/", StringSplitOptions.RemoveEmptyEntries);
```
ayende, line 244: Can we try this without allocations?

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
```diff
@@ -0,0 +1,365 @@
...
+                _outputReduceToCollection = outputReduceToCollection;
+                _indexWriteTransaction = indexWriteTransaction;
+                _database = index.DocumentDatabase;
+            }
+
+            public void Add(string referenceDocumentId, string reduceOutputDocumentId)
+            {
+                if (_referencesOfReduceOutputs.TryGetValue(referenceDocumentId, out var values) == false)
+                {
+                    values = new HashSet<string>(1);
```
ayende: case insensitive match?

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
```diff
@@ -0,0 +1,365 @@
...
+                    
+                    if (referenceDocument == null)
+                        continue;
+
+                    if (referenceDocument.Data.TryGet(nameof(ReduceOutputIdsReference.ReduceOutputs), out BlittableJsonReaderArray ids) == false)
+                        ThrowIdsPropertyNotFound(referenceDocument.Id);
+
+                    var idsToRemove = reduceReferenceIdToReduceOutputIds.Value;
+
+                    if (idsToRemove.Count >= ids.Length)
```
ayende: Is it possible that we have ids in the `idsToRemove` that aren't in the `ids` ? 
I think we need to do the work to remove it always

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
```diff
@@ -0,0 +1,365 @@
...
+                    {
+                        _database.DocumentsStorage.Delete(context, referenceDocument.Id, null);
+                        continue;
+                    }
+
+                    if (ids.Modifications == null)
+                        ids.Modifications = new DynamicJsonArray();
+
+                    var indexesToRemove = new List<int>();
```
ayende: Wouldn't it be easier, instead of removing it, to just build the new array?

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReferencesPattern.cs
```diff
@@ -0,0 +1,132 @@
...
+                        // {OrderedAt}
+
+                        return $"{{{numberOfFields}}}";
+                    }
+
+                    if (groups.Count == 3)
+                    {
+                        // {OrderedAt:yyyy-MM-dd}
+
+                        return $"{{{numberOfFields}:{groups[2]}}}";
```
ayende, line 45: what happens when the pattern we have here is not valid?

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/ReduceOutputDocumentActions.cs
```diff
@@ -51,6 +65,34 @@ public void Initialize(RavenTransaction tx)
             }
         }
 
+        public void AddPatternGeneratedIdForReduceOutput(Transaction tx, string reduceResultId, string referenceDocumentId)
+        {
+            var tree = tx.CreateTree(ReduceOutputsIdsToPatternReferenceIdsTreeName);
```
ayende: shouldn't we cache that?

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/ReduceOutputDocumentActions.cs
```diff
@@ -51,6 +65,34 @@ public void Initialize(RavenTransaction tx)
...
+
+            tree.Add(reduceResultId, referenceDocumentId);
+        }
+
+        public string GetPatternGeneratedIdForReduceOutput(Transaction tx, string reduceResultId)
+        {
+            var tree = tx.CreateTree(ReduceOutputsIdsToPatternReferenceIdsTreeName);
+
+            var result = tree.Read(reduceResultId);
+
```
ayende: can this be null?

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
arekpalinski: +1

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
arekpalinski: We call it very occasionally - once per indexing batch. For validation purposes only:

https://github.com/ravendb/ravendb/blob/9ce10e6847103dafa0f59c75ab5097a4b741b28d/src/Raven.Server/Documents/Indexes/Persistence/DeleteReduceOutputDocumentsCommand.cs#L16-L17

I don't think it's worth to optimize that

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
arekpalinski: I used `StringComparer.OrdinalIgnoreCase` in all placed where I use HashSet or Dictionary

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
arekpalinski: Not possible. I'll add `Debug.Assert` for that.

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
arekpalinski: I believe that most cases we won't have to many deletions here so just removing it should be a bit lighter I think

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/ReduceOutputDocumentActions.cs
arekpalinski: `Transaction` has it cached already. I just changed `string ReduceOutputsIdsToPatternReferenceIdsTreeName` to `Slice` so we won't cause additional allocations

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/ReduceOutputDocumentActions.cs
arekpalinski: Should never happen. I'll add explicit exception for that

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
ayende: But not here ,though?

arekpalinski: Here as well. Why not? Document IDs are case insensitive - am I missing something?

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReferencesPattern.cs
arekpalinski: it will be handled as regular indexing error

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
ayende: `new HashSet<string>(1)` will do case sensitive comparisons

arekpalinski: Oh, I meant that I changed it locally to `new HashSet<string>(1, StringComparer.OrdinalIgnoreCase);`

src/Raven.Client/Documents/Indexes/AbstractGenericIndexCreationTask.cs
```diff
@@ -95,6 +95,11 @@ protected AbstractGenericIndexCreationTask()
         /// If not null than each reduce result will be created as a document in the specified collection name.
         /// </summary>
         protected string OutputReduceToCollection { get; set; }
+        
+        /// <summary>
+        /// Defines pattern for identifiers of documents which reference IDs of reduce outputs documents
+        /// </summary>
+        protected Expression<Func<TReduceResult, string>> PatternOfReduceOutputReferences { get; set; }
```
ppekrol: Shouldn't this be 'PatternFor...` ?

ppekrol: Also the main property is `OutputReduceToCollection` maybe we should build upon that? e.g. `PatternForOutputReduceToCollectionReferences` ?

src/Raven.Client/Documents/Indexes/MapReduce/ReduceOutputIdsReference.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+
+namespace Raven.Client.Documents.Indexes.MapReduce
+{
+    public class ReduceOutputIdsReference
```
ppekrol: Similar here, maybe we should consider renaming it to something more similar to the property name in index definition?

ppekrol: Are we using this client side?

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
```diff
@@ -0,0 +1,365 @@
...
+                    _database.DocumentsStorage.DeleteDocumentsStartingWith(context, prefix);
+                }
+            }
+
+            foreach (var output in _reduceDocuments)
+            {
+                for (var i = 0; i < output.Value.Count; i++) // we have hash collision so there might be multiple outputs for the same reduce key hash
+                {
+                    var id = output.Value.Count == 1 ? GetOutputDocumentKey(output.Key) : GetOutputDocumentKeyForSameReduceKeyHash(output.Key, i);
+                    var doc = output.Value[i].Json;
```
ppekrol: we can dispose this right after `.Put` probably?

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
```diff
@@ -0,0 +1,365 @@
...
+                        if (idsToRemove.Count == indexesToRemove.Count)
+                            break;
+                    }
+
+                    foreach (int toRemove in indexesToRemove)
+                    {
+                        ids.Modifications.RemoveAt(toRemove);
+                    }
+
+                    var doc = context.ReadObject(referenceDocument.Data, referenceDocument.Id);
```
ppekrol: We can probably dispose this after Put?

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
```diff
@@ -0,0 +1,365 @@
...
+                    var referenceDoc = new DynamicJsonValue
+                    {
+                        [nameof(ReduceOutputIdsReference.ReduceOutputs)] = new DynamicJsonArray(referencesOfReduceOutput.Value),
+                        [Constants.Documents.Metadata.Key] = new DynamicJsonValue
+                        {
+                            [Constants.Documents.Metadata.Collection] = $"{_outputReduceToCollection}/References"
+                        }
+                    };
+
+                    var referenceJson = context.ReadObject(referenceDoc, "reference-of-reduce-output", BlittableJsonDocumentBuilder.UsageMode.ToDisk);
```
ppekrol: We can probably dispose this after put?

src/Raven.Client/Documents/Indexes/AbstractGenericIndexCreationTask.cs
arekpalinski: It's very long but yes it will be consistent. I'll made that change.

src/Raven.Client/Documents/Indexes/MapReduce/ReduceOutputIdsReference.cs
arekpalinski: - Renamed to `OutputReduceToCollectionReference`
- Yes, this is meant to be used client side. Eg.

```
var doc = session.Load<ReduceOutputIdsReference>("reports/daily/2019-10-26", x => x.IncludeDocuments(y => y.ReduceOutputs));

foreach (var docReduceOutput in doc.ReduceOutputs)
{
    output = session.Load<Result>(docReduceOutput);
}
```

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
arekpalinski: Yes, it was done by @grisha-kotler in the following PR https://github.com/ravendb/ravendb/pull/9852. I need to rebase

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
arekpalinski: Yes

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
arekpalinski: Yes

ayende (approved):

arekpalinski:
Please hold on with merging this PR until I'll figure out why Debug.Assert is occasionally failing: 
https://issues.hibernatingrhinos.com/issue/RavenDB-12932#focus=streamItem-67-42158.0-0

src/Raven.Studio/wwwroot/App/views/database/indexes/editIndex.html
```diff
@@ -102,6 +102,17 @@ <h2 id="reduce-title" class="margin-bottom margin-bottom-xs on-base-background">
                         <input type="text" class="form-control flex-grow has-error" placeholder="Enter collection name" data-bind="textInput: editedIndex().reduceToCollectionName, visible: editedIndex().outputReduceToCollection">
                     </div>
                 </div>
+                <div class="flex-horizontal margin-bottom" data-bind="visible: editedIndex().outputReduceToCollection">
```
ml054, line 105: is it only visible when output reduce to collection is set to true?

src/Raven.Studio/typescript/models/database/index/indexDefinition.ts
```diff
@@ -196,6 +208,7 @@ class indexDefinition {
             Configuration: this.configurationToDto(),
             Fields: this.fieldToDto(),
             OutputReduceToCollection: this.outputReduceToCollection() ? this.reduceToCollectionName() : null,
+            PatternForOutputReduceToCollectionReferences: this.patternForOutput() ? this.patternForOutputReduceToCollectionReferences() : null,
```
ml054: should we check for this.outputReduceToCollection() as well?

arekpalinski: yes, done

src/Raven.Studio/wwwroot/App/views/database/indexes/editIndex.html
arekpalinski: yes, that's correct

PR #9842, opened 2019-10-30
RavenDB-14201: Add Test for restoring legacy backups via MigrateDatabaseOffline

garayx:


grisha-kotler (approved):

src/Raven.Server/Web/ServerRequestHandler.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    public abstract class ServerRequestHandler : RequestHandler
+    {
+        public override void Init(RequestHandlerContext context)
+        {
+            base.Init(context);
+
+            var topologyEtag = GetLongFromHeaders(Constants.Headers.TopologyEtag);
+            if (topologyEtag.HasValue && Server.ServerStore.HasTopologyChanged(topologyEtag.Value))
+                context.HttpContext.Response.Headers[Constants.Headers.RefreshTopology] = "true";
```
ayende, line 13: why is this required?

grisha-kotler: It's for the `ClusterRequestExecutor` in the client. So we can update its topology if needed.

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -705,7 +705,6 @@ public async Task RestoreDatabase()
                     throw new ArgumentException($"No enough free space to restore a backup. Required space {desiredFreeSpace}, available space: {destinationDriveInfo.TotalFreeSpace}");
             }
 
-            HttpContext.Response.Headers[Constants.Headers.RefreshTopology] = "true";
```
ppekrol, line 708: Do we know why it was here?

src/Raven.Server/Web/ServerRequestHandler.cs
ppekrol: I think it would be better to store this in memory in OnTopologyChangeInternal. This would avoid us opening transaction for each request and we are already doing that for DatabaseRequestHandler.

garayx: done

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -945,6 +950,9 @@ public void NotifyAboutClusterTopologyAndConnectivityChanges()
 
         private void OnTopologyChangeInternal(ClusterTopology topology, Dictionary<string, NodeStatus> status = null)
         {
+            if (_lastClusterTopologyIndex < topology.Etag)
```
ppekrol, line 953: @garayx do we need to read it during ServerStore initialization?

garayx: done

ppekrol (approved):

PR #9849, opened 2019-11-01
RavenDB-14220 - Take into account the output collection for a map-reduce index budget

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-14220

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -70,8 +70,8 @@ namespace Raven.Server.Documents.Indexes
     {
         public new TIndexDefinition Definition => (TIndexDefinition)base.Definition;
 
-        protected Index(IndexType type, TIndexDefinition definition)
-            : base(type, definition)
+        protected Index(IndexType type, TIndexDefinition definition, bool hasOutputReduceToCollection)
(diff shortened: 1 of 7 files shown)


ayende (approved):

ppekrol (approved):

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -206,7 +206,9 @@ public abstract class Index : ITombstoneAware, IDisposable, ILowMemoryHandler
 
         internal bool IsLowMemory => _lowMemoryFlag.IsRaised();
 
-        protected Index(IndexType type, IndexDefinitionBase definition)
+        private readonly double _txAllocationsRatio;
+
+        protected Index(IndexType type, IndexDefinitionBase definition, bool hasOutputReduceToCollection)
```
arekpalinski, line 211: Instead of adding additional parameter I would prefer to use `definition` to detect if an index has `OutputReduceToCollection` defined. It would just require to cast it to `MapReduceIndexDefinition` when you're inside  `if (type.IsMapReduce())`

grisha-kotler: https://github.com/ravendb/ravendb/pull/9868

PR #9854, opened 2019-11-03
RavenDB-13080 & RavenDB-13081 : Time Series Querying

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-13080
https://issues.hibernatingrhinos.com/issue/RavenDB-13081

### DIFF
--- a/src/Raven.Server/Documents/Queries/Parser/QueryScanner.cs
+++ b/src/Raven.Server/Documents/Queries/Parser/QueryScanner.cs
@@ -531,6 +531,78 @@ namespace Raven.Server.Documents.Queries.Parser
 
         }
 
+        public bool TimeSeriesFunctionBody()
+        {
+            var original = _pos;
+            if (TryScan('(') == false)
+                return false;
+
+            // now we need to find the matching )
+
+            int nested = 1;
+            for (; _pos < _q.Length; _pos++)
+            {
+                switch (_q[_pos])
+                {
+                    case '"':
+                    case '\'':
+                        if (String(out _) == false)
+                            goto Failed;
+                        // we are now positioned at the _next_character, but we'll increment it
+                        // need to go back to stay in the same place :-)
+                        _pos--;
+                        break;
+                    case '/':
+                        // Detect // from the second /, we know that there is at least the ( before us,
+                        // so no need to do range check
+                        if (_q[_pos - 1] == '/')
+                        {
+                            for (; _pos < _q.Length; _pos++)
+                            {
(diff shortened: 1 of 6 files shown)


src/Raven.Server/Documents/Queries/Parser/QueryScanner.cs
```diff
@@ -531,6 +531,78 @@ public bool FunctionBody()
 
         }
 
+        public bool TimeSeriesFunctionBody()
```
ayende, line 536: This is duplicated, no?
Merge this with the JS body but accept the `{` vs `(` as param?

test/SlowTests/Client/TimeSeries/Query/TimeSeriesQueryTests.cs
```diff
@@ -118,5 +163,861 @@ select out(u)
...
+                new PeopleIndex().Execute(store);
+
+                WaitForIndexing(store);
+
+                using (var session = store.OpenSession())
+                {
+                    var query = session.Advanced.RawQuery<RawQueryResult>(@"
+from People as p
+where p.Age > 49
+select timeseries(
```
ayende, line 1089: Needs to think how we can do this on _JS_ calls as well.

aviv86: will be handled separately in https://issues.hibernatingrhinos.com/issue/RavenDB-13082

src/Raven.Server/Documents/Queries/Parser/QueryScanner.cs
aviv86: If I'll accept open parentheses and closing parentheses as parameters, then I won't be able to use them as switch cases anymore (non constant values).

I can handle that inside the default case, i.e :
```
default:
	if (_q[_pos] == open)
	{
		nested++;
		break;
	}
	if (_q[_pos] == close && --nested == 0)
	{
		_pos += 1;
		_tokenStart = original;
		_tokenLength = _pos - original;
		return true;
	}
	break;
```
but that will affect performance, wouldn't it ?

ayende: Then add a `isCurly` and do case for `{` and `(` and jump out if not? 
Honestly, though, I doubt that it would matter.

aviv86: done

ppekrol (approved):

ayende (approved):

PR #9859, opened 2019-11-04
RavenDB-14234: TrySimplifyJson should assign the right type to derive…

garayx:
…d properties

### DIFF
--- a/src/Raven.Client/Documents/Session/EntityToBlittable.cs
+++ b/src/Raven.Client/Documents/Session/EntityToBlittable.cs
@@ -452,12 +452,34 @@ namespace Raven.Client.Documents.Session
             return typeof(IEnumerable).IsAssignableFrom(type.GetGenericTypeDefinition());
         }
 
-        private static Type GetPropertyType(string propertyName, Type rootType)
+        internal static Type GetPropertyType(string propertyName, Type rootType)
         {
             if (rootType == null)
                 return null;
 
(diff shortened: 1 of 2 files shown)


src/Raven.Client/Documents/Session/EntityToBlittable.cs
```diff
@@ -452,12 +452,34 @@ private static bool ShouldSimplifyJsonBasedOnType(string typeValue)
...
+            }
+            catch (AmbiguousMatchException)
+            {
+                var memberInfos = ReflectionUtil.GetPropertiesAndFieldsFor(rootType, BindingFlags.Public | BindingFlags.Instance | BindingFlags.NonPublic)
+                    .Where(x => x.Name == propertyName)
+                    .ToList();
+
+                while (typeof(object) != rootType)
+                {
+                    memberInfo = memberInfos.FirstOrDefault(x => x.DeclaringType == rootType);
```
ayende, line 473: IsAssignableFrom, instead?

ppekrol: we want to get 'closest' decendant and IsAssignableFrom returns any, or I'm wrong?

ayende: But you are missing interfaces

garayx: Added test with classes that inherit from interface

ppekrol (approved):

ayende (approved):

PR #9860, opened 2019-11-04
RavenDB-13650 - Fixing needless usage of ToDictionary() for simple ca…

ayende:
…lls.

RavenDB-14196 - Fixing missing int cast on method call arguments


arekpalinski (commented):
It's hard to spot an issue with that change so it looks good to me. Although some tests has failed.

PR #9861, opened 2019-11-05
RavenDB-13990: Use UTF8 for linux and Unicode for windows on Pal, and…

aviviadi:
… use CreateFileW in windows where Pal still not in use (and zero end of buffer passed to PAL with at least sizeof(char) length). Libs included.


src/Sparrow.Server/Platform/Pal.cs
```diff
@@ -235,15 +323,55 @@ static Pal()
...
+            public byte* Pointer => (byte*)PinnedHandle.AddrOfPinnedObject();
+            private static readonly Encoding CurrentEncoding = PlatformDetails.RunningOnPosix ? Encoding.UTF8 : Encoding.Unicode;
+            private GCHandle PinnedHandle;
+
+            public Converter(string s)
+            {
+                var size = CurrentEncoding.GetMaxByteCount(s.Length) + sizeof(char);
+                _buffer = ArrayPool<byte>.Shared.Rent(size);
+                int length = CurrentEncoding.GetBytes(s, 0, s.Length, _buffer, 0);
+                for (int i = length; i < size; i++)
```
ayende:
```suggestion
                for (int i = length; i < sizeof(char); i++)
```

src/Voron/Platform/Win32/Win32NativeFileMethods.cs
```diff
@@ -87,9 +87,8 @@ public struct FileSegmentElement
             out int pNumberOfBytesRead,
             NativeOverlapped* lpOverlapped
         );
-
-        [DllImport("kernel32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
-
+        
+        [DllImport("kernel32.dll", EntryPoint = "CreateFileW", SetLastError = true, CharSet = CharSet.Unicode, BestFitMapping = false)]
```
ayende, line 91: Why do we need this?

src/Voron/Platform/Win32/Win32NativeFileMethods.cs
```diff
@@ -423,8 +422,8 @@ public struct SYSTEM_INFO
             int OutBufferSize,
             out int BytesReturned,
             IntPtr Overlapped);
-
-        [DllImport("Kernel32.dll", SetLastError = true)]
+        
+        [DllImport("kernel32.dll", EntryPoint = "CreateFileW", SetLastError = true, CharSet = CharSet.Unicode, BestFitMapping = false)]
```
ayende, line 426: Why?

ayende (changes requested):

src/Sparrow.Server/Platform/Pal.cs
ppekrol: Notice that `int i = length` so I think this is correct?

src/Voron/Platform/Win32/Win32NativeFileMethods.cs
aviviadi: This in order to match what we do in Pal also in Win MMap Pager (until it will go into PAL as well)

src/Sparrow.Server/Platform/Pal.cs
ayende: We have a string of 200 chars, meaning UTF8 max size is 1KB.
We wrote 200 bytes, but now we are manually zeroing 1KB slowly.

aviviadi: https://github.com/ravendb/ravendb/pull/9861/commits/20df27533b971d1bad662f48f62a59d4052b9e0b

PR #9863, opened 2019-11-05
RDBCL-1007 We must not rely on TryAutorize result but the actual Auth…

arekpalinski:
…enticationStatus when updating LastAuthorizedNonClusterAdminRequestTime

### DIFF
--- a/src/Raven.Server/Routing/RequestRouter.cs
+++ b/src/Raven.Server/Routing/RequestRouter.cs
@@ -115,13 +115,11 @@ namespace Raven.Server.Routing
                 }
 
                 var status = RavenServer.AuthenticationStatus.ClusterAdmin;
-                bool authResult = true;
                 try
                 {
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Routing/RequestRouter.cs
```diff
@@ -138,12 +136,29 @@ public async ValueTask HandlePath(RequestHandlerContext reqCtx)
...
+                            switch (status)
+                            {
+                                case RavenServer.AuthenticationStatus.Allowed:
+                                case RavenServer.AuthenticationStatus.Operator:
+                                {
+                                    _ravenServer.Statistics.LastAuthorizedNonClusterAdminRequestTime = now;
+                                    _lastAuthorizedNonClusterAdminRequestTime = now;
+                                    break;
+                                }
+                                case RavenServer.AuthenticationStatus.None:
```
ppekrol, line 150: Do we need those and the exception in default path?

PR #9865, opened 2019-11-05
RavenDB-13902: vfork, re-PRing

aviviadi:


src/Raven.Server/Utils/Cpu/CpuUsageExtensionPoint.cs
```diff
@@ -43,124 +42,97 @@ public class CpuUsageExtensionPoint : IDisposable
...
-                        break;
-                    }
-
-                    var line = await nextLine;
-                    if (HandleInfoReceived(line) == false)
-                    {
-                        break;
+                        var lineOutHandler = new Action<object, string>((p, l) =>
+                        {
+                            Interlocked.Exchange(ref _isDataValid, 1);
```
ayende: I can't follow what you are doing with this here, but that is suspicious and likely racy behavior

ayende (changes requested):

ayende (approved):

ayende:
Please clean up this PR, so we'll have only your changes here

aviviadi:
It is cleaned

ayende:
Oh, I saw a lot of additional commits there that weren't related.
Thanks, looking into it again now

src/Raven.Server/Utils/Cpu/CpuUsageExtensionPoint.cs
```diff
@@ -43,124 +38,93 @@ public class CpuUsageExtensionPoint : IDisposable
...
-
-                    var line = await nextLine;
-                    if (HandleInfoReceived(line) == false)
-                    {
-                        break;
+                        var lineOutHandler = new Action<object, string>((p, l) =>
+                        {
+                            if (l == null)
+                            {
+                                if (DateTime.UtcNow - lastReceivedLine > TimeSpan.FromSeconds(60))
```
ayende: Am I reading this wrong, what can cause us to get a `null` here except at the end of the input?

aviviadi: Yes. `l` here contains the result of `sr.ReadLineAsync().Result`. It is `null` when nothing is being read (not the really end the process though). So as long as something is being sent to the standard stream within 60 seconds, it is ok.
(This is at least what I understand from the previous implementation)

ayende: `ReadLineAsync` will only return null on end of input, no other scenario

aviviadi: True.. (https://docs.microsoft.com/en-us/dotnet/api/system.io.streamreader.readlineasync?view=netframework-4.8#returns)
So we read line by line until no output.. I do not see where the problem is... (?)

ayende: Your code here does: `if (DateTime.UtcNow - lastReceivedLine`
It is attempting to handle the case where we have no input from the process.
But the current behavior means that it will never happen

aviviadi: Ok, you mean ReadLineAsync will block if process hangs and will not return null. 
iirc I've checked that but I will check again.

ayende: Yes, there is no input there, so it will not return anything.

ayende (approved):

PR #9870, opened 2019-11-07
RavenDB-14235: fix memory overlap on id slice creation, when we have …

garayx:
…two or more control characters in a row.

### DIFF
--- a/src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
+++ b/src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
@@ -40,6 +40,8 @@ namespace Raven.Client.Documents.Session
         [ThreadStatic]
         private static int _clientSessionIdCounter;
 
+        internal long _asyncTasksCounter;
+
         protected readonly int _clientSessionId = ++_clientSessionIdCounter;
 
         protected readonly RequestExecutor _requestExecutor;
@@ -91,23 +93,26 @@ namespace Raven.Client.Documents.Session
 
         public async Task<ServerNode> GetCurrentSessionNode()
         {
-            (int Index, ServerNode Node) result;
-            switch (RequestExecutor.Conventions.ReadBalanceBehavior)
+            using (AsyncTaskHolder())
             {
-                case ReadBalanceBehavior.None:
-                    result = await _requestExecutor.GetPreferredNode().ConfigureAwait(false);
-                    break;
-                case ReadBalanceBehavior.RoundRobin:
-                    result = await _requestExecutor.GetNodeBySessionId(_clientSessionId).ConfigureAwait(false);
-                    break;
-                case ReadBalanceBehavior.FastestNode:
-                    result = await _requestExecutor.GetFastestNode().ConfigureAwait(false);
-                    break;
-                default:
-                    throw new ArgumentOutOfRangeException(RequestExecutor.Conventions.ReadBalanceBehavior.ToString());
-            }
+                (int Index, ServerNode Node) result;
+                switch (RequestExecutor.Conventions.ReadBalanceBehavior)
+                {
+                    case ReadBalanceBehavior.None:
+                        result = await _requestExecutor.GetPreferredNode().ConfigureAwait(false);
+                        break;
+                    case ReadBalanceBehavior.RoundRobin:
+                        result = await _requestExecutor.GetNodeBySessionId(_clientSessionId).ConfigureAwait(false);
+                        break;
+                    case ReadBalanceBehavior.FastestNode:
+                        result = await _requestExecutor.GetFastestNode().ConfigureAwait(false);
+                        break;
+                    default:
+                        throw new ArgumentOutOfRangeException(RequestExecutor.Conventions.ReadBalanceBehavior.ToString());
+                }
 
-            return result.Node;
+                return result.Node;
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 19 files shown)


src/Raven.Server/Documents/DocumentIdWorker.cs
```diff
@@ -178,11 +177,16 @@ private static ByteStringContext.InternalScope UnlikelyGetLowerUnicode(ByteStrin
...
-
             int idSize = JsonParserState.VariableSizeIntSize(strLength);
 
-            int escapePositionsSize = JsonParserState.FindEscapePositionsMaxSize(str);
+            int escapePositionsSize = JsonParserState.FindEscapePositionsMaxSize(str, out var escapedCount);
+
+            /*
+             *  add the size of all control characters
+             *  this is to treat case when we have 2+ control character in a row
+             *  GetMaxByteCount returns smaller size than the actual size with escaped control characters
```
ayende, line 186: Add actual data: string such as `"\0\0"` will be translated to `\u0000\u0000`, etc

test/SlowTests/Issues/RavenDB_14235.cs
```diff
@@ -0,0 +1,112 @@
...
+        [InlineData(64)]
+        [InlineData(128)]
+        [InlineData(256)]
+        public async Task PassingOnlyEscapedCharactersAsId(int size)
+        {
+            char[] chars = new char[size];
+
+            for (int c = 0; c < 32; c++)
+            {
+                if (c.In(_listOfNumsThatAreNulls))
```
ayende, line 39: I don't understand what this is for.
Did you mean, known special chars here?

garayx: those are whitespaces, so we save a guid as an ID in place of those

src/Raven.Server/Documents/DocumentIdWorker.cs
garayx: done

arekpalinski:
What about handling those characters in collection name? We saw issues like:  https://issues.hibernatingrhinos.com/issue/RDBS-47
where document ID was fine but `@collection` metadata was corrupted (client side probably) and that resulted in throwing NRE on database startup because we couldn't find a relevant tree for given collection name.

Should it be also supported?

ayende:
We certainly need to at least test such collection names, yes

garayx:
@arekpalinski the `@collection` creation going trough same piece of code
https://github.com/ravendb/ravendb/blob/v4.2/src/Raven.Server/Documents/DocumentsStorage.cs#L1846

test/SlowTests/Issues/RavenDB_14235.cs
```diff
@@ -0,0 +1,256 @@
...
+                str += new string(chars);
+                str += new string(Enumerable.Repeat(abc, partialSize).Select(s => s[random.Next(s.Length)]).ToArray());
+
+                using var store = GetDocumentStore();
+                store.Commands().Put(str, null, new { WeirdName = str }, new Dictionary<string, object>
+                {
+                    { "@collection", str }
+                });
+
+                using var session = store.OpenAsyncSession();
```
ppekrol, line 154: I love C# :)

test/SlowTests/Issues/RavenDB_14235.cs
```diff
@@ -0,0 +1,256 @@
...
+                var lazy = session.Advanced.Lazily.LoadAsync<User>(new[] { name, name, name, name });
+                lazy.Value.Start();
+            }
+            catch (Exception e)
+            {
+                Assert.Equal(typeof(InvalidOperationException), e.GetType());
+                exceptionsCounter++;
+            }
+
+            Assert.Equal(tryCounter, exceptionsCounter);
```
ppekrol: This will be racy, high chance IMO that some of those will finish and not throw.

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -1291,6 +1305,10 @@ private void Dispose(bool isDisposing)
             if (_isDisposed)
                 return;
 
+            var asyncTasksCounter = Interlocked.Read(ref _asyncTasksCounter);
+            if (asyncTasksCounter != 0)
+                throw new InvalidOperationException($"Disposing session with active async task is forbidden. Number of active async tasks: {asyncTasksCounter}");
```
arekpalinski: Can we improve the error message a bit and add:

```
Please make sure that all asynchronous session methods returning Task are awaited.
```

to be explicit about what needs to be checked.

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -1893,4 +1911,20 @@ internal void OnBeforeQueryInvoke(BeforeQueryEventArgs args)
...
     }
+
+    internal struct AsyncTaskHolder : IDisposable
+    {
+        private readonly InMemoryDocumentSessionOperations _session;
+
+        public AsyncTaskHolder(InMemoryDocumentSessionOperations session)
+        {
+            _session = session;
+            Interlocked.Increment(ref _session._asyncTasksCounter);
```
ayende: Here, we also want to error if we have a non zero old value.
We don't allow concurrent async tasks in the async session.
Only a single such task is allowed.

src/Raven.Server/Documents/DocumentIdWorker.cs
```diff
@@ -230,11 +236,19 @@ private static ByteStringContext.InternalScope UnlikelyGetLowerUnicode(ByteStrin
 
             var writePos = ptr + maxStrSize;
 
+            var sizeDifference = idSize - JsonParserState.VariableSizeIntSize(strLength);
+            if (sizeDifference > 0)
```
ayende: Can this be negative?

garayx: The strLength should be smaller/equal to the initial value that was used to calculate idSize, so the sizeDifference value should be positive/zero.

test/SlowTests/Issues/RavenDB_14235.cs
garayx: done

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
garayx: done

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
garayx: done

src/Raven.Server/Documents/DocumentIdWorker.cs
garayx: added debug assert there

ayende: So we can just have an assert here, and avoid the `if` entirely?

garayx: no, we should move the pointer for idSize if there is difference, this is the same code like here :
https://github.com/ravendb/ravendb/blob/v4.2/src/Raven.Server/Documents/DocumentIdWorker.cs#L285

ppekrol (approved):

PR #9882, opened 2019-11-14
RavenDB-14094: Fix(2) JsonContextPool leak (originally identified in RunLowMemoryHandlers)

aviviadi:


aviviadi:
Lets see if passes tests first

src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -10,129 +9,53 @@
...
+        [ThreadStatic]
+        private static ContextStackThreadReleaser _releaser;
 
         private class ContextStackThreadReleaser
         {
-            private HashSet<JsonContextPoolBase<T>> _parents = new HashSet<JsonContextPoolBase<T>>();
-            int _threadId;
-            public List<ThreadIdHolder> ThreadIdHolders = new List<ThreadIdHolder>();            
+            private readonly int _ownerThread;
+            private readonly StrongReference<ConcurrentDictionary<int, ContextStack>> _contextStacksByThreadId;
```
ayende: We need a _list_ of these

src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -10,129 +9,53 @@
...
         }
 
-        [ThreadStatic]
-        private static ContextStackThreadReleaser _releaser;
-        
-
         private void EnsureCurrentThreadContextWillBeReleased(int currentThreadId)
-        {            
+        {
             if (_releaser == null)
```
ayende, line 58: If null, create it.
If not null, add yourself to it. But may already exists, so probably use a set.

ayende (changes requested):

src/Sparrow/Json/JsonContextPoolBase.cs
aviviadi: I preferred to add the strong ref instead of "myself" (this == pool).  Because the ref set to null at pool's Dispose, and then I do not care about this parent anyhow

ayende (approved):

arekpalinski (approved):

PR #9888, opened 2019-11-17
RavenDB-14225 Allow to define separate conflict revision configuration

karmeli87:


src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -185,6 +185,9 @@ public void InitializeFromDatabaseRecord(DatabaseRecord dbRecord)
         {
             try
             {
+                if (dbRecord.RevisionsForConflicts != null)
+                    ConflictConfiguration.Default = dbRecord.RevisionsForConflicts;
```
ayende, line 189: This is a _static_ instance, no?

ayende (changes requested):

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
karmeli87: yes! done.

karmeli87:
Same as previously, first time in the `ctor` of `RevisionsStorage` and then updated by `InitializeFromDatabaseRecord`.

src/Raven.Client/ServerWide/Operations/EditRevisionsForConflictsOperation.cs
```diff
@@ -0,0 +1,76 @@
...
+using Raven.Client.Documents.Session;
+using Raven.Client.Http;
+using Raven.Client.Json;
+using Raven.Client.Json.Converters;
+using Raven.Client.Util;
+using Sparrow.Json;
+
+namespace Raven.Client.ServerWide.Operations
+{
+    public class EditRevisionsForConflictsOperation : IServerOperation<EditRevisionsForConflictsOperationResult>
```
arekpalinski: Shouldn't be renamed to `ConfigureRevisionsForConflictsOperation`? For consistency with `ConfigureRevisionsOperation`

src/Raven.Client/ServerWide/Operations/EditRevisionsForConflictsOperationResult.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Raven.Client.ServerWide.Operations
+{
+    public class EditRevisionsForConflictsOperationResult
```
arekpalinski: Rename to `ConfigureRevisionsForConflictsOperationResult`?

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -343,6 +346,10 @@ public RevisionsCollectionConfiguration GetRevisionsConfiguration(string collect
             if (configuration == null)
                 configuration = GetRevisionsConfiguration(collectionName.Name);
 
+            if (configuration.Disabled && 
+                nonPersistentFlags.Contain(NonPersistentDocumentFlags.FromReplication) == false)
+                return false;
```
arekpalinski, line 351: How is that change related to introduced option of defining separate conflict revision configuration? Or this is another fix?

karmeli87: I guess it both.

On conflict we were passing the configuration directly, and if we don't check for `Disabled` we will keep conflicted revision even if they were disabled.
In general if the configuration is disabled we should simply leave the state as is.

src/Raven.Client/ServerWide/Operations/EditRevisionsForConflictsOperationResult.cs
karmeli87: done

src/Raven.Client/ServerWide/Operations/EditRevisionsForConflictsOperation.cs
karmeli87: done

PR #9889, opened 2019-11-17
RavenDB-14121: NoDisposalNeeded should inherit from LinuxRaceConditio…

garayx:
…nWorkAround


arekpalinski:
@garayx Why do we need this? Some test has failed

garayx:
> @garayx Why do we need this? Some test has failed

@arekpalinski if NoDisposalNeeded doesnt derive from LinuxRaceConditionWorkAround, I am getting excpetion and cannot get the output from test. 

I will check the tests.

arekpalinski:
CC @topalach

PR #9890, opened 2019-11-17
RavenDB-14273 - Backup task should delete the .in-progress file when it fails

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-14273

### DIFF
--- a/src/Raven.Server/Documents/PeriodicBackup/BackupTask.cs
+++ b/src/Raven.Server/Documents/PeriodicBackup/BackupTask.cs
@@ -676,19 +676,8 @@ namespace Raven.Server.Documents.PeriodicBackup
             }
 
             var directoryPath = Path.GetDirectoryName(filePath);
-            var destinationDriveInfo = DiskSpaceChecker.GetDiskSpaceInfo(directoryPath);
-            if (destinationDriveInfo == null)
-            {
-                if (_logger.IsInfoEnabled)
-                    _logger.Info($"Couldn't find the disk space info for path: {filePath}");
 
-                return;
-            }
-
-            var desiredFreeSpace = Size.Min(new Size(512, SizeUnit.Megabytes), destinationDriveInfo.TotalSize * 0.01) + new Size(totalUsedSpace, SizeUnit.Bytes);
-            if (destinationDriveInfo.TotalFreeSpace < desiredFreeSpace)
-                throw new DiskFullException($"Not enough free space to create a snapshot. " +
-                                            $"Required space {desiredFreeSpace}, available space: {destinationDriveInfo.TotalFreeSpace}");
+            BackupHelper.AssertFreeSpaceForSnapshot(directoryPath, totalUsedSpace, "create a snapshot", _logger);
         }
 
         private void BackupTypeValidation()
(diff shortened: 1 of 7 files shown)


arekpalinski:
I've scheduled test on Jenkins once again - it looks there was some glitch

src/Raven.Server/Documents/PeriodicBackup/BackupTask.cs
```diff
@@ -631,6 +638,9 @@ private class InternalBackupResult
                 catch (Exception e)
                 {
                     status.LocalBackup.Exception = e.ToString();
+
+                    // deleting the temp backup file if the backup failed
+                    IOExtensions.DeleteFile(tempBackupFilePath);
```
ayende: Maybe another try/catch here.
We explicitly don't care if we failed to delete the file, no?

src/Raven.Server/Documents/PeriodicBackup/BackupTask.cs
```diff
@@ -644,6 +654,30 @@ private class InternalBackupResult
             return internalBackupResult;
         }
 
+        private void ValidateFreeSpaceForSnapshot(string filePath)
+        {
+            long totalUsedSpace = 0;
+            foreach (var mountPointUsage in _database.GetMountPointsUsage(includeTempBuffers: false))
```
ayende, line 673: These checks can fail, no?

grisha-kotler: No. All the info is in memory.

src/Raven.Server/Documents/PeriodicBackup/BackupTask.cs
grisha-kotler: done

arekpalinski (approved):

src/Raven.Server/Documents/PeriodicBackup/BackupTask.cs
```diff
@@ -644,6 +654,43 @@ private class InternalBackupResult
...
+            var destinationDriveInfo = DiskSpaceChecker.GetDiskSpaceInfo(directoryPath);
+            if (destinationDriveInfo == null)
+            {
+                if (_logger.IsInfoEnabled)
+                    _logger.Info($"Couldn't find the disk space info for path: {filePath}");
+
+                return;
+            }
+
+            var desiredFreeSpace = Size.Min(new Size(512, SizeUnit.Megabytes), destinationDriveInfo.TotalSize * 0.01) + new Size(totalUsedSpace, SizeUnit.Bytes);
```
arekpalinski: This actually requires to have 512 MB of free space regardless the backup size. Why?

grisha-kotler: This is the same logic that we use for database restore

arekpalinski: Okay. That might make sense. We won't consume free space completely this way. Do you think we should extract this logic to some helper method and used it here and during restore, so it will be consistent in both?

grisha-kotler: done

arekpalinski (approved):

arekpalinski:
Can you please resolve conflicts?

grisha-kotler:
Rebased

PR #9892, opened 2019-11-18
RavenDB-14250 Allow to fix database topology

karmeli87:
With this change the order of the members is fixed.
Note that as a side-effect moving from member to rehab is disabled.

### DIFF
--- a/src/Raven.Client/ServerWide/DatabaseTopology.cs
+++ b/src/Raven.Client/ServerWide/DatabaseTopology.cs
@@ -153,6 +153,40 @@ $"NodeTag of 'InternalReplication' can't be modified after 'GetHashCode' was inv
         public LeaderStamp Stamp;
         public bool DynamicNodesDistribution;
         public int ReplicationFactor = 1;
+        public List<string> PriorityOrder;
+
+        private readonly Random _random = new Random();
+        public int Shuffle(string _, string __)
+        {
+            return _random.Next(-100, 100);
+        }
+
+        public void ReorderMembers()
+        {
+            if (PriorityOrder == null || PriorityOrder.Count == 0)
+            {
+                Members.Sort(Shuffle);
+                return;
+            }
+
+            var members = new List<string>();
+            for (int i = 0; i < PriorityOrder.Count; i++)
+            {
+                var priorityNode = PriorityOrder[i];
+                if (Members.Contains(priorityNode))
+                    members.Add(priorityNode);
+            }
+
+            for (int i = 0; i < Members.Count; i++)
+            {
+                var member = Members[i];
+                if (members.Contains(member) == false)
+                    members.Add(member);
+            }
+
+            Members = members;
+        }
+
 
         public bool RelevantFor(string nodeTag)
         {
@@ -329,7 +363,8 @@ $"NodeTag of 'InternalReplication' can't be modified after 'GetHashCode' was inv
                 [nameof(DemotionReasons)] = DynamicJsonValue.Convert(DemotionReasons),
                 [nameof(DynamicNodesDistribution)] = DynamicNodesDistribution,
                 [nameof(ReplicationFactor)] = ReplicationFactor,
-                [nameof(DatabaseTopologyIdBase64)] = DatabaseTopologyIdBase64
+                [nameof(DatabaseTopologyIdBase64)] = DatabaseTopologyIdBase64,
+                [nameof(PriorityOrder)] = new DynamicJsonArray(PriorityOrder)
             };
         }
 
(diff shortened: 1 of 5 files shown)


ayende (commented):
I don't think that this is a complete situation.
Consider the case of A,B,C topology, where A is the primary.
It goes down, and then we have B,C as members, and A as rehab.
But once A is back online, we want it to be the primary again, not sit there in the end.

So the `Fixed` shouldn't be just a boolean, but rather be something like `DesiredTopology` or something like that, so users can say how they want things to show up.

grisha-kotler:
Do we need to change anything in the studio?

karmeli87:
Yes, when re-ordering database nodes we need to add an option to "fix" this order.

arekpalinski:
Will you need @ml054 to take care of it?

src/Raven.Client/ServerWide/DatabaseTopology.cs
```diff
@@ -153,6 +153,40 @@ public class DatabaseTopology
         public LeaderStamp Stamp;
         public bool DynamicNodesDistribution;
         public int ReplicationFactor = 1;
+        public List<string> PriorityOrder;
+
+        private readonly Random _random = new Random();
```
ayende: We only need it for a specific scenario, this shouldn't be here.

src/Raven.Client/ServerWide/DatabaseTopology.cs
```diff
@@ -153,6 +153,40 @@ public class DatabaseTopology
...
+        public int Shuffle(string _, string __)
```
ayende: Same, shouldn't be on this class, certainly not as a public method.
Create a separate class that can use this

src/Raven.Client/ServerWide/DatabaseTopology.cs
```diff
@@ -153,6 +153,40 @@ public class DatabaseTopology
...
+
+            var members = new List<string>();
+            for (int i = 0; i < PriorityOrder.Count; i++)
+            {
+                var priorityNode = PriorityOrder[i];
+                if (Members.Contains(priorityNode))
+                    members.Add(priorityNode);
+            }
+
+            for (int i = 0; i < Members.Count; i++)
```
ayende, line 184: Should we shuffle the other stuff, that isn't in the priority order?

karmeli87: okay, so I shuffle the members any way

src/Raven.Client/ServerWide/DatabaseTopology.cs
karmeli87: done

src/Raven.Client/ServerWide/DatabaseTopology.cs
karmeli87: done

src/Raven.Client/ServerWide/DatabaseTopology.cs
```diff
@@ -140,6 +140,17 @@ public override int GetHashCode()
         }
     }
 
+    public static class ThreadSafeRandom
```
arekpalinski: Can we change it to `private` or `internal`? I don't want it to be considered public API

src/Raven.Client/ServerWide/DatabaseTopology.cs
```diff
@@ -153,6 +164,33 @@ public class DatabaseTopology
         public LeaderStamp Stamp;
         public bool DynamicNodesDistribution;
         public int ReplicationFactor = 1;
+        public List<string> PriorityOrder;
+
+        public void ReorderMembers()
```
arekpalinski: Can we change it to `internal`?

src/Raven.Client/ServerWide/Operations/ReorderDatabaseMembersOperation.cs
```diff
@@ -15,20 +15,22 @@ public class ReorderDatabaseMembersOperation : IServerOperation
...
         {
             public List<string> MembersOrder;
+            public bool Fixed;
         }
 
         private readonly string _database;
         private readonly Parameters _parameters;
 
-        public ReorderDatabaseMembersOperation(string database, List<string> order)
+        public ReorderDatabaseMembersOperation(string database, List<string> order, bool fixedTopology = false)
```
arekpalinski: This breaks binary compatibility. Please add a new constructor for that overload.

src/Raven.Client/ServerWide/DatabaseTopology.cs
karmeli87: done

src/Raven.Client/ServerWide/DatabaseTopology.cs
karmeli87: done

src/Raven.Client/ServerWide/Operations/ReorderDatabaseMembersOperation.cs
karmeli87: done

PR #9899, opened 2019-11-20
RavenDB-14284 Don't drop existing connection if we get a new one from…

karmeli87:
… the same source

### DIFF
--- a/src/Raven.Server/Documents/Replication/ReplicationLoader.cs
+++ b/src/Raven.Server/Documents/Replication/ReplicationLoader.cs
@@ -268,31 +268,36 @@ namespace Raven.Server.Documents.Replication
             if (Interlocked.CompareExchange(ref _reconnectInProgress, 1, 0) == 1)
                 return;
 
-            foreach (var failure in _reconnectQueue)
+            try
             {
-                try
+                foreach (var failure in _reconnectQueue)
                 {
-                    if (_reconnectQueue.TryRemove(failure) == false)
-                        continue;
-
-                    if (failure.RetryOn > DateTime.UtcNow)
+                    try
                     {
-                        _reconnectQueue.Add(failure);
-                        continue;
+                        if (_reconnectQueue.TryRemove(failure) == false)
+                            continue;
+
+                        if (failure.RetryOn > DateTime.UtcNow)
+                        {
+                            _reconnectQueue.Add(failure);
+                            continue;
+                        }
+
+                        AddAndStartOutgoingReplication(failure.Node, failure.External);
                     }
-
-                    AddAndStartOutgoingReplication(failure.Node, failure.External);
-                }
-                catch (Exception e)
-                {
-                    if (_log.IsInfoEnabled)
+                    catch (Exception e)
                     {
-                        _log.Info($"Failed to start outgoing replication to {failure.Node}", e);
+                        if (_log.IsInfoEnabled)
+                        {
+                            _log.Info($"Failed to start outgoing replication to {failure.Node}", e);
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -312,17 +317,10 @@ private void AssertValidConnection(IncomingConnectionInfo connectionInfo)
...
-                {
-                    _log.Info(
-                        $"Disconnecting existing connection from {value.FromToString} because we got a new connection from the same source db");
-                }
-
-                IncomingReplicationRemoved?.Invoke(value);
-
-                value.Dispose();
+                throw new InvalidOperationException(
+                    $"connection for this database already exists from {value.ConnectionInfo.SourceUrl}.");
```
arekpalinski: Minor: connection -> Connection

arekpalinski (approved):

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -288,6 +291,8 @@ private void ForceTryReconnectAll()
                     }
                 }
             }
+
+            Interlocked.Exchange(ref _reconnectInProgress, 0);
```
ayende: Put it inside a finally?

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
ayende: I think we want to make this slightly smarter. Do we have any state around the connection?
Can we see when was the last time we got something from it?

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
karmeli87: done

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
karmeli87: We can have the last heartbeat time and last etag from source (and maybe other stuff as well).
Do we want here a better exception? or we want to decide which connection should be dropped?

ayende: We always want better errors :-)
But yes, if the old connection hasn't had communication recently, we want to drop that and let the new one in

ayende (approved):

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -312,13 +322,15 @@ private void AssertValidConnection(IncomingConnectionInfo connectionInfo)
                     $"Cannot accept the incoming replication connection from {connectionInfo.SourceUrl}, because this node is in passive state.");
             }
 
-            if (_incoming.TryRemove(connectionInfo.SourceDatabaseId, out IncomingReplicationHandler value))
+            if (_incoming.TryGetValue(connectionInfo.SourceDatabaseId, out var value))
             {
+                if (value.LastHeartbeatTicks + TimeSpan.FromMinutes(1).Ticks > Database.Time.GetUtcNow().Ticks)
```
arekpalinski: Do we really need to calculate that with the usage of `Ticks`? :) 

- Can we use `new DateTime(value.LastHeartbeatTicks)` ?
- `TimeSpan.FromMinutes(1)` - please move to static readonly field

karmeli87: done

PR #9902, opened 2019-11-21
RavenDB-13948 - Creating duplicate backups from the same server

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-13948

### DIFF
--- a/src/Raven.Client/Documents/Operations/OngoingTasks/GetTaskInfoResult.cs
+++ b/src/Raven.Client/Documents/Operations/OngoingTasks/GetTaskInfoResult.cs
@@ -299,23 +299,19 @@ namespace Raven.Client.Documents.Operations.OngoingTasks
     {
         public TimeSpan TimeSpan { get; set; }
 
-        public DateTime StartTimeLocal { get; set; }
-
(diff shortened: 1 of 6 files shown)


ayende (approved):

src/Raven.Client/Documents/Operations/OngoingTasks/GetTaskInfoResult.cs
```diff
@@ -299,17 +299,23 @@ public class NextBackup : IDynamicJson
     {
         public TimeSpan TimeSpan { get; set; }
 
-        public DateTime DateTime { get; set; }
+        public DateTime StartTimeLocal { get; set; }
```
arekpalinski: I don't think we can change it. Operations API is out public API so that would be breaking change. You can add additional property and mark `public DateTime DateTime { get; set; }` as obsolete.

Then both `DateTime ` and `StartTimeLocal ` needs to be provided in `ToJson()`

grisha-kotler: The `DateTime` property name can stay (I changed it for convenience).

PR #9907, opened 2019-11-22
RavenDB-13078 TS Indexing

arekpalinski:


src/Raven.Server/Documents/Indexes/Static/TimeSeries/MapTimeSeriesIndex.cs
```diff
@@ -71,7 +72,7 @@ protected override IIndexingWork[] CreateIndexWorkExecutors()
         {
             var workers = new List<IIndexingWork>
             {
-                //new CleanupDocuments(this, DocumentDatabase.DocumentsStorage, _indexStorage, Configuration, null)
+                new CleanupDocuments(this, DocumentDatabase.DocumentsStorage, _indexStorage, Configuration, null)
```
ppekrol, line 79: This means that we are processing Document Tombstones, right?

ppekrol: We probably need always to subscribe to document deletes because of this.

ppekrol (approved):

src/Raven.Server/Documents/Indexes/Static/TimeSeries/MapTimeSeriesIndex.cs
arekpalinski: Yes, that's correct. It was working in the test because we forced to stop / start index

ppekrol (approved):

src/Raven.Server/Utils/TypeConverter.cs
```diff
@@ -391,6 +392,9 @@ public static dynamic ToDynamicType(object value)
             if (value is DynamicNullObject)
                 return value;
 
+            if (value is DynamicObject)
```
ayende, line 395: Maybe `IDynamicMetaObject` ?

arekpalinski: Yes. Changed to `IDynamicMetaObjectProvider`

PR #9908, opened 2019-11-24
RavenDB-14290 - support load, store and delete in OnBeforeStore and OnBeforeDelete

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-14290

### DIFF
--- a/src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
+++ b/src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
@@ -128,9 +128,7 @@ namespace Raven.Client.Documents.Session
         /// <summary>
         /// hold the data required to manage the data for RavenDB's Unit of Work
         /// </summary>
-        protected internal readonly Dictionary<object, DocumentInfo> DocumentsByEntity =
-            new Dictionary<object, DocumentInfo>(ObjectReferenceEqualityComparer<object>.Default);
-
+        internal readonly DocumentsByEntityHolder DocumentsByEntity = new DocumentsByEntityHolder();
         /// <summary>
         /// hold the data required to manage Counters tracking for RavenDB's Unit of Work
         /// </summary>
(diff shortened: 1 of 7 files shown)


arekpalinski (approved):

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -1931,4 +1940,215 @@ public void Dispose()
...
+            _deletedEntities.Add(entity);
+        }
+
+        public void Remove(object entity)
+        {
+            _deletedEntities.Remove(entity);
+            _onBeforeDeletedEntities?.Remove(entity);
+        }
+
+        public void Evict(object entity)
```
ayende, line 2091: Why does evict goes here?

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -1931,4 +1940,215 @@ public void Dispose()
...
+                {
+                    Key = doc.Key,
+                    Value = doc.Value,
+                    ExecuteOnBeforeStore = true
+                };
+            }
+
+            if (_onBeforeStoreDocumentsByEntity != null)
+            {
+                foreach (var doc in _onBeforeStoreDocumentsByEntity)
```
ayende, line 2032: Create a new instance, and null the old one.
That way, we can get call on our own changes and safely handle them

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -1931,4 +1940,215 @@ public void Dispose()
...
+                yield return new DeletedEntitiesEnumeratorResult
+                {
+                    Entity = entity,
+                    ExecuteOnBeforeDelete = true
+                };
+            }
+
+            if (_onBeforeDeletedEntities != null)
+            {
+                foreach (var entity in _onBeforeDeletedEntities)
```
ayende, line 2129: New instance here as well

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
grisha-kotler: We need the old one since we track those entities

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
grisha-kotler: Because we cannot Evict during `OnBeforeStore`

PR #9915, opened 2019-11-26
RavenDB-13575 : Baseline timeseries queries

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-13575


src/Raven.Server/Documents/Queries/AST/Query.cs
```diff
@@ -327,123 +328,159 @@ public enum Type
...
+
+        private byte _numberOfValues;
+
+        public TimeSeriesAggregation(TimeSeriesAggregation.Type type)
         {
-            _valIndex = valIndex;
             Aggregation = type;
-            _val = 0;
-            Count = 0;
+            _count = new object[TimeSeriesValuesSegment.MaxNumberOfValues];
```
ppekrol: I'm wondering if we should use some kind of an array pool for this. From the other side this is cached in query cache?

src/Raven.Server/Documents/Queries/Results/QueryResultRetrieverBase.cs
```diff
@@ -734,130 +740,714 @@ public object InvokeFunction(string methodName, Query query, string documentId,
             }
         }
 
-        private BlittableJsonReaderObject InvokeTimeSeriesFunction(TimeSeriesFunction func, string documentId, object[] args)
+        private BlittableJsonReaderObject InvokeTimeSeriesFunction(DeclaredFunction declaredFunction, string documentId, object[] args)
```
ppekrol: can we make this class partial and move all TS stuff to `QueryResultRetrieverBase..TimeSeries.cs` ? This becomes monolith.

ppekrol: Or maybe we can extract that to separate class and introduce field e.g. `TimeSeries` in this retriever?

src/Raven.Server/Documents/Queries/AST/Query.cs
ayende: This is fairly constant, I don't think we are going to need more. 
Especially because it will be complex to return them.
That said, why not use a List here?

src/Raven.Server/Documents/Queries/AST/TimeSeriesBetweenExpression.cs
```diff
@@ -0,0 +1,50 @@
...
+        public TimeSeriesBetweenExpression(QueryExpression source, QueryExpression min, QueryExpression max) : base(source, null, null)
+        {
+            if (min is ValueExpression ve)
+            {
+                Min = ve;
+            }
+            else
+            {
+                _minExpression = min;
+            }
```
ayende: This looks really awkward, why do we have to this it this way?

src/Raven.Server/Documents/Queries/Results/QueryResultRetrieverBase.cs
ayende: I like this idea, yes

src/Raven.Server/Documents/Queries/Results/QueryResultRetrieverBase.cs
```diff
@@ -734,130 +740,714 @@ public object InvokeFunction(string methodName, Query query, string documentId,
...
+                            return singleResult.Tag?.ToString();
+                        case "values":
+                            if (fe.Compound.Count == 1)
+                                return singleResult.Values;
+
+                            if (fe.Compound.Count > 2)
+                                throw new InvalidQueryException($"Failed to evaluate expression '{fe}'");
+
+                            if (int.TryParse(fe.Compound[1].Value, out var index) == false)
+                                throw new InvalidQueryException($"Failed to evaluate expression '{fe}'");
```
ayende: Can this be a parameter?

src/Raven.Server/Documents/Queries/Results/QueryResultRetrieverBase.cs
```diff
@@ -734,130 +740,714 @@ public object InvokeFunction(string methodName, Query query, string documentId,
...
+
+            object GetValue(QueryExpression expression, SingleResult singleResult)
             {
-                foreach (var cur in items)
+                if (expression is FieldExpression fe)
                 {
-                    MaybeMoveToNextRange(cur.TimeStamp);
-                    count++;
-                    for (int i = 0; i < aggStates.Length; i++)
+                    switch (fe.Compound[0].Value.ToLower())
```
ayende: Note, this does allocation per call. Can we avoid this?

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
```diff
@@ -387,8 +386,15 @@ internal SeriesSummary GetSummary()
...
                         }
 
-                        result.Min[i] = Math.Min(result.Min[i], _currentSegment.SegmentValues.Span[i].Min);
-                        result.Max[i] = Math.Max(result.Max[i], _currentSegment.SegmentValues.Span[i].Max);
+                        if (double.IsNaN(_currentSegment.SegmentValues.Span[i].Min) == false)
+                        {
+                            result.Min[i] = Math.Min(result.Min[i], _currentSegment.SegmentValues.Span[i].Min);
+                        }
+
+                        if (double.IsNaN(_currentSegment.SegmentValues.Span[i].Max) == false)
```
ayende, line 408: Why did you add these?

test/SlowTests/Client/TimeSeries/Query/TimeSeriesQueryTests.cs
```diff
@@ -1189,17 +1243,3913 @@ as Stocks
...
+
+                        Assert.Equal(141.5, val.Min[1]);
+                        Assert.Equal(142.82, val.Max[1]);
+                        Assert.Equal(2, val.Count[1]);
+                        Assert.Equal((141.5 + 142.82) / 2, val.Avg[1]);
+
+                        var expectedFrom = new DateTime(baseline.Year, baseline.Month, 1, 0, 0, 0);
+                        var expectedTo = expectedFrom.AddMonths(1);
+
+
```
ayende, line 4702: Can I also write:

```
declare timeseries out(x) 
{
    from x.HeartRate between $start and $end
        load $config as src
        where (Values[0] between src.Min and src.Max) OR 
              (src.IsoCompliant = true AND Values[0] > $val)  
    group by '1 month' 
    select min(), max(), avg()
}
```

Where the `$config` is some arbitrary document?

ayende (approved):
Nice work!

src/Raven.Server/Documents/Queries/Results/QueryResultRetrieverBase.cs
aviv86: what do you mean? 
something like that? 
```
declare timeseries out(x, index) 
{
    from x.HeartRate between $start and $end
	where Values[index] > 100
}
```

src/Raven.Server/Documents/Queries/AST/Query.cs
aviv86: no good reason.
changed to List

src/Raven.Server/Documents/Queries/Results/QueryResultRetrieverBase.cs
aviv86: done

src/Raven.Server/Documents/Queries/AST/TimeSeriesBetweenExpression.cs
aviv86: removed this

src/Raven.Server/Documents/Queries/Results/QueryResultRetrieverBase.cs
aviv86: done

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
aviv86: otherwise, if the segment has an item with missing values (NaN), then min and max will both be double.NaN

test/SlowTests/Client/TimeSeries/Query/TimeSeriesQueryTests.cs
aviv86: no, this is not supported, we can only load by Tag
do we want to support that?

we can achieve the same by loading the document in root RQL and 
pass it as an argument to the timeseries function, no? 
e.g.
```
declare timeseries out(x, src) 
{
    from x.HeartRate between $start and $end
        where (Values[0] between src.Min and src.Max) OR 
              (src.IsoCompliant = true AND Values[0] > $val)  
    group by '1 month' 
    select min(), max(), avg()
}
from users as u
load $config as src
select out(u, src)
```

src/Raven.Server/Documents/Queries/Parser/QueryParser.cs
```diff
@@ -1308,69 +1312,84 @@ private DeclaredFunction DeclaredFunction()
...
-                ThrowParseException($"Unable to parse timeseries query for {name}, missing FROM");
+                }
 
-            if (Field(out var source) == false)
-                ThrowParseException($"Unable to parse timeseries query for {name}, missing FROM");
+                if (fe.FieldValue != "Tag")
+                {
+                    ThrowInvalidQueryException($"Expected to find 'Tag' after LOAD in time series function '{name}', but got '{fe.FieldValue}'." +
+                                               "In time series functions you can only use LOAD in order to load by the 'Tag' property of a time series entry. " +
+                                               "You can use the root RQL in order to load any other documents, and can pass them as arguments to the time series function");
```
ayende, line 1379: Great error message

src/Raven.Server/Documents/Queries/Results/TimeSeriesRetriever.cs
```diff
@@ -0,0 +1,815 @@
...
+            if (_loadedDocuments == null)
+                _loadedDocuments = new Dictionary<string, Document>();
+
+            var tag = singleResult.Tag?.ToString();
+            if (tag == null)
+                throw new InvalidQueryException("Unable to load document from 'Tag' property of time series entry. 'Tag' is null. " +
+                                                $"Timestamp: '{singleResult.TimeStamp.GetDefaultRavenFormat()}'");
+
+            if (_loadedDocuments.TryGetValue(tag, out var document) == false)
+                _loadedDocuments[tag] = document = _database.DocumentsStorage.Get(_context, tag);
```
ayende, line 726: Do we need to check for null here? Before you put it in the `_loadedDocuments` or return it below

ayende (approved):

src/Raven.Server/Documents/Queries/Results/TimeSeriesRetriever.cs
aviv86: right, I missed that.

but now that I'm thinking about it, I'm not sure if we should throw on null document/null tag. maybe just return null?

It might be useful to support things like this:

```
from x.HeartRate between $start and $end
	load Tag as src
	where src != null and src.Accuracy > 2
```

ayende: Good point, you are right

PR #9916, opened 2019-11-27
RavenDB-14254 Handling a conflict should take into account the unused…

karmeli87:
… databases

### DIFF
--- a/src/Raven.Server/Documents/ConflictsStorage.cs
+++ b/src/Raven.Server/Documents/ConflictsStorage.cs
@@ -872,6 +872,8 @@ namespace Raven.Server.Documents
             else
                 return ConflictStatus.Update; //document with 'id' doesn't exist locally, so just do PUT
 
+            context.SkipChangeVectorValidation = context.DocumentDatabase.DocumentsStorage.TryRemoveUnusedIds(ref local);
+
             status = ChangeVectorUtils.GetConflictStatus(changeVector, local);
             if (status == ConflictStatus.Conflict)
             {
--- a/src/Raven.Server/ServerWide/Context/DocumentsOperationContext.cs
+++ b/src/Raven.Server/ServerWide/Context/DocumentsOperationContext.cs
@@ -12,7 +12,18 @@ namespace Raven.Server.ServerWide.Context
 
         internal string LastDatabaseChangeVector;
         internal Dictionary<string, long> LastReplicationEtagFrom;
-        internal bool SkipChangeVectorValidation;
+        private bool _skipChangeVectorValidation;
+
+        internal bool SkipChangeVectorValidation
+        {
+            get => _skipChangeVectorValidation;
+            set
+            {
+                if (_skipChangeVectorValidation == false)
+                    _skipChangeVectorValidation = value;
+            }
+        }
+
         protected internal override void Reset(bool forceResetLongLivedAllocator = false)
         {
             base.Reset(forceResetLongLivedAllocator);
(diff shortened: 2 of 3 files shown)


src/Raven.Server/ServerWide/Context/DocumentsOperationContext.cs
```diff
@@ -12,7 +12,18 @@ public class DocumentsOperationContext : TransactionOperationContext<DocumentsTr
...
         internal Dictionary<string, long> LastReplicationEtagFrom;
-        internal bool SkipChangeVectorValidation;
+        private bool _skipChangeVectorValidation;
+
+        internal bool SkipChangeVectorValidation
+        {
+            get => _skipChangeVectorValidation;
+            set
+            {
+                if (_skipChangeVectorValidation == false)
```
arekpalinski, line 22: So we _never_ switch from `true` to `false` ?

src/Raven.Server/Documents/ConflictsStorage.cs
```diff
@@ -872,6 +872,8 @@ public static ConflictStatus GetConflictStatusForDocument(DocumentsOperationCont
             else
                 return ConflictStatus.Update; //document with 'id' doesn't exist locally, so just do PUT
 
+            context.SkipChangeVectorValidation = context.DocumentDatabase.DocumentsStorage.TryRemoveUnusedIds(ref local);
```
ayende, line 875: Why do we do this here?

src/Raven.Server/ServerWide/Context/DocumentsOperationContext.cs
```diff
@@ -12,7 +12,18 @@ public class DocumentsOperationContext : TransactionOperationContext<DocumentsTr
 
         internal string LastDatabaseChangeVector;
         internal Dictionary<string, long> LastReplicationEtagFrom;
-        internal bool SkipChangeVectorValidation;
+        private bool _skipChangeVectorValidation;
+
+        internal bool SkipChangeVectorValidation
```
ayende, line 17: I'm not seeing usage of this?

src/Raven.Server/ServerWide/Context/DocumentsOperationContext.cs
karmeli87: Not in the same transaction.

src/Raven.Server/ServerWide/Context/DocumentsOperationContext.cs
karmeli87: This is for debug purpose to assert that we don't move the global change vector back in time.

src/Raven.Server/ServerWide/Context/DocumentsOperationContext.cs
arekpalinski: I don't follow. Can you elaborate? What's purpose of it then?

karmeli87: This property is for debug assertion to validate that we do not moving the global change vector back in time (or conflicted). 

For example if it was `A:10 B:5` it shouldn't be set back to let's say `A:11` or `B:5`.
But there are some exceptions to this, for instance when we remove unused change vector entries. 
In this case we know that we will move the change vector back in time, so we need to skip the assertion.

We set the global change vector at the end of each transaction.
So once we set it to `true` we don't want to set it back to `false` until the transaction is over.

arekpalinski: Thanks. Makes sense

src/Raven.Server/Documents/ConflictsStorage.cs
arekpalinski: @karmeli87 Can you answer this comment?

karmeli87: We might put the the `local` one which after the removal of the entries is conflicted with the current global change vector, so we want to skip the assertion.

arekpalinski (approved):

arekpalinski:
@ayende Could you check @karmeli87's answers and please approve if it's fine

ayende (approved):

PR #9920, opened 2019-11-28
RavenDB-12648 - disable the MaxServicePointIdleTime

grisha-kotler:


src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1583,6 +1583,7 @@ private static void UpdateConnectionLimit(IEnumerable<string> urls)
 
                 try
                 {
+                    ServicePointManager.MaxServicePointIdleTime = -1;
```
ayende: This seems like a global change that we are doing.
We can use: https://docs.microsoft.com/en-us/dotnet/api/system.net.servicepoint.maxidletime?view=netframework-4.8#System_Net_ServicePoint_MaxIdleTime

Instead, no?

ayende (approved):

ppekrol:
Explanation of this change for future:
When MaxIdleTime is hit, GC can collect the service point and new one will get default values

PR #9921, opened 2019-12-01
RavenDB-14292: _databaseWakeUpTime should be UTC

garayx:


src/Raven.Server/Documents/PeriodicBackup/PeriodicBackupRunner.cs
```diff
@@ -360,7 +359,7 @@ public long StartBackupTask(long taskId, bool isFullBackup)
             return CreateBackupTask(periodicBackup, isFullBackup, SystemTime.UtcNow);
         }
 
-        public DateTime GetWakeDatabaseTime()
+        public DateTime GetWakeDatabaseTimeUtc()
         {
             var wakeupDatabase = DateTime.MaxValue;
```
grisha-kotler:
```suggestion
            DateTime? wakeupDatabase = null;
```

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -927,7 +927,7 @@ private bool ShouldContinueDispose(string name, DateTime? wakeup, out int dueTim
                 timer.Dispose();
             }
 
-            if (wakeup.HasValue == false || wakeup.Value == DateTime.MaxValue)
+            if (wakeup.HasValue == false || wakeup.Value == DateTime.MaxValue.ToUniversalTime())
```
grisha-kotler: We can remove this:
`wakeup.Value == DateTime.MaxValue.ToUniversalTime()`

grisha-kotler (changes requested):

src/Raven.Server/Documents/DatabasesLandlord.cs
ayende: This can actually throw, I think.
If you have a local time where the UTC is beyond the limits, no?

src/Raven.Server/Documents/PeriodicBackup/PeriodicBackupRunner.cs
garayx: done

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -42,6 +42,9 @@ public class DatabasesLandlord : IDisposable
         private readonly SemaphoreSlim _databaseSemaphore;
         private readonly ServerStore _serverStore;
 
+        // used in ServerWideBackupStress
+        internal static bool SkipShouldContinueDisposeCheck = false;
```
ppekrol: I would prefer this to not be 'static'. You can get instance of RavenServer in tests and access DatabasesLandlord there.

src/Raven.Server/Documents/PeriodicBackup/PeriodicBackupRunner.cs
```diff
@@ -360,26 +359,36 @@ public long StartBackupTask(long taskId, bool isFullBackup)
...
-        public DateTime GetWakeDatabaseTime()
+        public DateTime? GetWakeDatabaseTimeUtc()
         {
-            var wakeupDatabase = DateTime.MaxValue;
             long lastEtag;
 
             using (_database.DocumentsStorage.ContextPool.AllocateOperationContext(out DocumentsOperationContext context))
-            using (var tx = context.OpenReadTransaction())
             {
+                using var tx = context.OpenReadTransaction();
```
ppekrol: Let's avoid changes like this. Readability here is -10

src/Raven.Server/Documents/DatabasesLandlord.cs

PR #9927, opened 2019-12-02
RavenDB-14319: Add timeout when there is compare exchange subscriptio…

maximburyak:
…n substitution concurrency to prevent a partial live lock

### DIFF
--- a/src/Raven.Server/Documents/Subscriptions/SubscriptionConnectionState.cs
+++ b/src/Raven.Server/Documents/Subscriptions/SubscriptionConnectionState.cs
@@ -9,6 +9,7 @@ using Raven.Client.Exceptions.Documents.Subscriptions;
 using Raven.Client.Util;
 using Raven.Server.Documents.TcpHandlers;
 using Sparrow.Server;
+using Sparrow.Threading;
 
 namespace Raven.Server.Documents.Subscriptions
 {
@@ -44,7 +45,7 @@ namespace Raven.Server.Documents.Subscriptions
 
         // we should have two locks: one lock for a connection and one lock for operations
         // remember to catch ArgumentOutOfRangeException for timeout problems
-        public async Task<IDisposable> RegisterSubscriptionConnection(
+        public async Task<DisposeOnce<SingleAttempt>> RegisterSubscriptionConnection(
             SubscriptionConnection incomingConnection,
             TimeSpan timeToWait)
         {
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/TcpHandlers/SubscriptionConnection.cs
```diff
@@ -162,7 +164,8 @@ public async Task InitAsync()
                         _logger.Info(
                             $"Subscription Id {SubscriptionId} from IP {TcpConnection.TcpClient.Client.RemoteEndPoint} starts to wait until previous connection from {_connectionState.Connection?.TcpConnection.TcpClient.Client.RemoteEndPoint} is released");
                     }
-                    timeout = TimeSpan.FromMilliseconds(Math.Max(250, (long)_options.TimeToWaitBeforeConnectionRetry.TotalMilliseconds / 2));
+                  
+                    timeout = TimeSpan.FromMilliseconds(Math.Max(250, (long)_options.TimeToWaitBeforeConnectionRetry.TotalMilliseconds / 2) + random.Next(15,50));
```
ayende, line 169: Is this enough jitter? Can we increase it?

src/Raven.Server/Documents/Subscriptions/SubscriptionConnectionState.cs
```diff
@@ -88,7 +88,10 @@ public override string ToString()
 
             var subscriptionConnection = Interlocked.CompareExchange(ref _currentConnection, incomingConnection, null);
             if (subscriptionConnection != null && subscriptionConnection != incomingConnection)
+            {
+                await Task.Delay(timeToWait);
```
ayende, line 93: Didn't we already wait here?  `await ConnectionInUse.WaitAsync(timeToWait) `

maximburyak: The problem occures when ConnectionInUse is not turned on, and we go straight to this line

src/Raven.Server/Documents/TcpHandlers/SubscriptionConnection.cs
maximburyak: It's enough IMO, even without it, the problem was resolved, because Task.Delay is not too accurate

ppekrol (approved):

ayende (approved):

PR #9929, opened 2019-12-03
RavenDB-14323 - No index error when there is an error in the referenced document

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-14323

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Workers/HandleReferences.cs
+++ b/src/Raven.Server/Documents/Indexes/Workers/HandleReferences.cs
@@ -7,6 +7,7 @@ using System.Threading;
 using Raven.Server.Config.Categories;
 using Raven.Server.Documents.Indexes.Persistence.Lucene;
 using Raven.Server.ServerWide.Context;
+using Raven.Server.Utils;
 using Sparrow.Json;
 using Sparrow.Logging;
 using Voron;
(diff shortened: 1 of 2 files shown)


src/Raven.Server/Documents/Indexes/Workers/HandleReferences.cs
```diff
@@ -210,10 +200,17 @@ private unsafe bool HandleDocuments(ActionType actionType, DocumentsOperationCon
 
                                                 _index.MapsPerSec.MarkSingleThreaded(numberOfResults);
                                             }
-                                            catch (Exception e)
```
ayende, line 213: how are normal errors handled now?

ppekrol: This is fine. Copy from `MapDocuments`

src/Raven.Server/Documents/Indexes/Workers/HandleReferences.cs
```diff
@@ -261,6 +258,26 @@ private unsafe bool HandleDocuments(ActionType actionType, DocumentsOperationCon
             return moreWorkFound;
         }
 
+        private IEnumerable<Document> GetDocumentFromCollectionThatReference(DocumentsOperationContext databaseContext, TransactionOperationContext indexContext, string collection, Reference referencedDocument, long lastIndexedEtag)
```
ppekrol, line 261: Can we avoid this? This will conflict in 5.0 and I do not see any benefit at the moment.

grisha-kotler: This is useful when we have a document that is referenced by many documents

ppekrol: Ok, I haven't noticed that you have removed the List.

src/Raven.Server/Documents/Indexes/Workers/HandleReferences.cs
```diff
@@ -261,6 +258,26 @@ private unsafe bool HandleDocuments(ActionType actionType, DocumentsOperationCon
...
+        {
+            foreach (var key in _indexStorage.GetDocumentKeysFromCollectionThatReference(collection, referencedDocument.Key, indexContext.Transaction))
+            {
+                using (GetLower(out Slice loweredKey))
+                {
+                    // when there is conflict, we need to apply same behavior as if the document would not exist
+                    var doc = _documentsStorage.Get(databaseContext, loweredKey, throwOnConflict: false);
+
+                    if (doc != null && doc.Etag <= lastIndexedEtag)
```
ppekrol: Let's dispose the doc otherwise.

grisha-kotler: done

PR #9933, opened 2019-12-05
RavenDB-14175 Broadcast cluster command to all nodes

karmeli87:


src/Raven.Client/Documents/Operations/Counters/CounterBatchOperation.cs
```diff
@@ -38,12 +38,13 @@ public CounterBatchCommand(CounterBatch counterBatch, DocumentConventions conven
             public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, ServerNode node, out string url)
             {
                 url = $"{node.Url}/databases/{node.Database}/counters";
+                // TODO: to avoid putting twice on retry it should have some kind of a guid
```
ppekrol, line 41: Create a separate issue for this?

src/Sparrow/Json/BlittableJsonReaderObject.cs
```diff
@@ -1024,6 +1024,10 @@ public BlittableJsonReaderObject Clone(JsonOperationContext context)
 
             if (Modifications == null)
             {
+                if (forceCloning == false &&
```
ppekrol, line 1027: I do not like that this is only for one case here. Shouldn't this be just after Assert in this method?

src/Sparrow/Json/BlittableJsonReaderObject.cs
```diff
@@ -1015,7 +1015,7 @@ public BlittableJsonReaderObject CloneOnTheSameContext()
             return Clone(_context);
         }
 
-        public BlittableJsonReaderObject Clone(JsonOperationContext context)
+        public BlittableJsonReaderObject Clone(JsonOperationContext context, bool forceCloning = true)
```
ppekrol, line 1018: This breaks binary compatibility. Can you create a separate overload for this?

test/Tests.Infrastructure/DocumentStoreExtensions.cs
```diff
@@ -374,14 +374,15 @@ public JsonCommandWithPayload(string url, HttpMethod method, BlittableJsonReader
                 public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, ServerNode node, out string url)
                 {
                     url = $"{node.Url}/databases/{node.Database}{_url}";
+                    var clone = _payload?.Clone(ctx);
```
ppekrol, line 377: Do we support concurrent cloning of blittable?

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1342,6 +1386,64 @@ public static async Task<Stream> ReadAsStreamUncompressedAsync(HttpResponseMessa
...
         }
 
+        private async Task Broadcast<TResult>(RavenCommand<TResult> command, SessionInfo sessionInfo, CancellationToken token)
+        {
+            if (command is IRaftCommand == false)
+                throw new InvalidOperationException("You can broadcast only raft commands.");
+
+            var tasks = new Dictionary<Task, CancellationTokenSource>();
+
+            command.SetTimeout(_defaultBroadcastTimeout ?? TimeSpan.FromSeconds(30));
```
ppekrol, line 1396: Maybe it would be better to set default broadcast timeout in conventions to this value instead of introducing nullable?

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1191,6 +1222,12 @@ internal HttpRequestMessage CreateRequest<TResult>(JsonOperationContext ctx, Ser
 
                 var raftRequestString = "raft-request-id=" + raftCommand.RaftUniqueRequestId;
                 builder.Query = builder.Query?.Length > 1 ? $"{builder.Query.Substring(1)}&{raftRequestString}" : raftRequestString;
+
+                if (command.Timeout.HasValue == false && 
+                    TopologyNodes?.Count > 1)
+                {
+                    command.SetTimeout(InitialRaftRequestTimeout ?? TimeSpan.FromSeconds(5));
```
ppekrol, line 1229: Maybe set default in conventions

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -139,6 +139,34 @@ public string Url
             }
         }
 
+        private TimeSpan? _defaultBroadcastTimeout;
+
+        public TimeSpan? DefaultBroadcastTimeout
```
ppekrol, line 144: DefaultRaftBroadcastTimeout?

src/Raven.Client/Http/RavenCommand.cs
```diff
@@ -37,7 +38,8 @@ public interface IRaftCommand
     public abstract class RavenCommand<TResult>
     {
         public CancellationToken CancellationToken = CancellationToken.None;
-        public Dictionary<ServerNode, Exception> FailedNodes;
+
+        public ConcurrentDictionary<ServerNode, Exception> FailedNodes;
```
ppekrol, line 42: This breaks backward compatibility

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1082,7 +1112,8 @@ private bool ShouldExecuteOnAll<TResult>(ServerNode chosenNode, RavenCommand<TRe
                    selector.Topology?.Nodes?.Count > 1 &&
                    command.IsReadRequest &&
                    command.ResponseType == RavenCommandResponseType.Object &&
-                   chosenNode != null;
+                   chosenNode != null && 
+                   typeof(IRaftCommand) != command.GetType();
```
ppekrol, line 1116: This doesn't look correct? `command is IRaftCommand == false` ?

src/Raven.Client/Http/RavenCommand.cs
```diff
@@ -116,21 +119,67 @@ public virtual async Task<ResponseDisposeHandling> ProcessResponse(JsonOperation
...
         }
 
+        private TaskCompletionSource<bool> _successfulResponseWaiter;
+
+        internal void SetTimeout(TimeSpan timeout)
+        {
+            Timeout = timeout;
+        }
+
+        private async Task SetFirstSuccessfulResult(Action action)
```
ppekrol, line 154: Isn't this method a bit of an overkill? SemaphoreSlim should be enough? It would make this method more readable.

src/Raven.Client/Documents/Operations/Counters/CounterBatchOperation.cs
karmeli87: https://issues.hibernatingrhinos.com/issue/RavenDB-14349

src/Sparrow/Json/BlittableJsonReaderObject.cs
karmeli87: It make sense to me that cloning the same object on the same context should return `this`.
But was afraid to make this globally (without this parameter), since someone might rely on the fact that cloning gives a completely new instance.

ppekrol: I'm not saying that we should not have it, just worried that we are only using this variable when we do not have modifications.

test/Tests.Infrastructure/DocumentStoreExtensions.cs
karmeli87: you are correct, I will move the `CreateRequest` to be synchronous.

src/Raven.Client/Http/RequestExecutor.cs
karmeli87: done

src/Raven.Client/Http/RequestExecutor.cs
karmeli87: done

src/Raven.Client/Http/RequestExecutor.cs
karmeli87: done

src/Raven.Client/Http/RequestExecutor.cs
karmeli87: done

src/Raven.Client/Http/RavenCommand.cs
karmeli87: done

src/Sparrow/Json/BlittableJsonReaderObject.cs
ayende: This is likely going to cause issues and should be removed.
Scenario, we have a blittable on raw data storage
We clone it explicitly, because we can't modify it and write at the same time.
You have now broken the write, because we read from where we write.

Other issues are about modifying the instance, disposing the old one and then using the new one, etc.

Let's roll back this change.

src/Raven.Client/Http/RavenCommand.cs
ayende: Why is this concurrent, at all? The command is single threaded, after all.
when we do stuff like read faster, we clone the _command_.

ayende (changes requested):

karmeli87:
After all of the comments I realize that it would be better to implement a "copy constructor" to clone the command, instead of trying to make the command thread-safe.

PR #9939, opened 2019-12-06
RavenDB-14334 changed timeout

ppekrol:


src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -1739,7 +1739,7 @@ public async Task SaveAsync()
 
                     var indexCount = _command.Static.Count + _command.Auto.Count;
                     var operationTimeout = _store._serverStore.Engine.OperationTimeout;
-                    var timeout = TimeSpan.FromSeconds((indexCount / 50.0) * operationTimeout.TotalSeconds);
+                    var timeout = TimeSpan.FromSeconds((indexCount / 10.0) * operationTimeout.TotalSeconds);
```
grisha-kotler: It needs to take into account the number of cores on that machine

ppekrol: There is no good value here. My machine can have 12 cores but different one in the cluster can have only 1.

grisha-kotler (approved):

PR #9942, opened 2019-12-08
RavenDB-14334 - start the indexes concurrently

grisha-kotler:


src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -76,21 +73,54 @@ public IndexStore(DocumentDatabase documentDatabase, ServerStore serverStore)
...
+
+            HandleSorters(record, raftIndex);
+            HandleDeletes(record, raftIndex);
+
+            try
+            {
+                HandleChangesForStaticIndexes(record, raftIndex, indexesToStart);
+                HandleChangesForAutoIndexes(record, raftIndex, indexesToStart);
+            }
+            finally
```
ayende: Why is this in a finally clause?

grisha-kotler: Done

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -76,21 +73,69 @@ public IndexStore(DocumentDatabase documentDatabase, ServerStore serverStore)
...
+                }
+                catch (OperationCanceledException e)
+                {
+                    _documentDatabase.RachisLogIndexNotifications.NotifyListenersAbout(raftIndex, e);
+                    return;
+                }
+
+                try
+                {
+                    if (_documentDatabase.DatabaseShutdown.IsCancellationRequested)
```
karmeli87: Should we NotifyListeners about this?

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -76,21 +73,69 @@ public IndexStore(DocumentDatabase documentDatabase, ServerStore serverStore)
...
+            if (indexesToStart.Count <= 0)
+                return;
+
+            var sp = Stopwatch.StartNew();
+
+            if (Logger.IsInfoEnabled)
+                Logger.Info($"Starting {indexesToStart.Count} indexes");
+
+            Parallel.ForEach(indexesToStart, index =>
+            {
```
karmeli87, line 107: It is CPU bounded, no? If it is we might what to set `MaxDegreeOfParallelism` to the number of cores.

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -57,7 +54,7 @@ public class IndexStore : IDisposable
 
         public readonly IndexIdentities Identities = new IndexIdentities();
 
-        public Logger Logger => _logger;
+        public Logger Logger { get; }
```
ppekrol: Change to readonly field

src/Raven.Server/Documents/Indexes/IndexStore.cs
ppekrol: Not sure if we should not take 50% of the cores or 75%. Thoughts?

karmeli87: +1 for 50%

ayende: Remember to handle a single core scenario as well.
But let's not use the cores on the machine, use the _licensed_ cores.

ppekrol: But all of licensed ones or 50%?

ayende: Math.Max(1, LicensedCores / 2)

src/Raven.Server/Documents/Indexes/IndexStore.cs
grisha-kotler: done

src/Raven.Server/Documents/Indexes/IndexStore.cs
grisha-kotler: done

src/Raven.Server/Documents/Indexes/IndexStore.cs
grisha-kotler: Actually Math.Max(1, UtilizedCores / 2)

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -57,40 +56,103 @@ public class IndexStore : IDisposable
...
+                    _documentDatabase.RachisLogIndexNotifications.NotifyListenersAbout(raftIndex, e);
+                    return;
+                }
+
+                try
+                {
+                    StartIndex(index);
+                }
+                catch (Exception e)
+                {
```
karmeli87, line 125: Should we continue to try and create the next index? 
Since we will notify the client about a failure, but still will try to create the next index,

ayende: This is running in parallel, the next index might have already run

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -57,40 +56,103 @@ public class IndexStore : IDisposable
...
+
+            if (Logger.IsInfoEnabled)
+                Logger.Info($"Started {indexesToStart.Count} indexes, took: {sp.ElapsedMilliseconds}ms");
+        }
+
+        private void ExecuteForIndexes(IEnumerable<Index> indexes, Action<Index> action)
+        {
+            int numberOfUtilizedCores;
+            using (_documentDatabase.ServerStore.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
+            using (context.OpenReadTransaction())
```
ppekrol: I do not think that this context is needed?

grisha-kotler: removed

PR #9945, opened 2019-12-09
RavenDB-14325 Tombstone cleaner should take into account backups and ETLs cluster-wide

karmeli87:


src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -867,6 +867,15 @@ public long ReadLastEtag()
             }
         }
 
+        public (long Etag, string ChangeVector) ReadLastEtagAndChangeVector()
+        {
+            using (DocumentsStorage.ContextPool.AllocateOperationContext(out DocumentsOperationContext documentsContext))
+            using (var tx = documentsContext.OpenReadTransaction())
+            {
+                return (DocumentsStorage.ReadLastEtag(tx.InnerTransaction), DocumentsStorage.GetDatabaseChangeVector(tx.InnerTransaction));
```
ayende, line 875: Is there a match between the etag and the change vector?
_should_ there be?

src/Raven.Server/Documents/ETL/EtlProcess.cs
```diff
@@ -814,7 +810,7 @@ protected void EnsureThreadAllocationStats()
             _threadAllocations = NativeMemory.CurrentThreadStats;
         }
 
-        public override Dictionary<string, long> GetLastProcessedTombstonesPerCollection()
+        /*public override Dictionary<string, long> GetLastProcessedTombstonesPerCollection()
```
ayende: why leave commented code?

src/Raven.Server/Documents/DocumentDatabase.cs
karmeli87: I think there should be a match.
But, does it really matter?

src/Raven.Server/Documents/ETL/EtlProcess.cs
karmeli87: removed

ppekrol (approved):

ppekrol:
@grisha-kotler please check Backup part
@arekpalinski please check ETL part

src/Raven.Server/Documents/ETL/EtlLoader.cs
```diff
@@ -19,7 +22,7 @@
 
 namespace Raven.Server.Documents.ETL
 {
-    public class EtlLoader : IDisposable
+    public class EtlLoader : IDisposable, ITombstoneAware
```
arekpalinski, line 25: Why do we need to make the loaded aware of tomstones? Why was it not enough that each process implemented `ITombstoneAware`?

ayende (approved):

src/Raven.Server/Documents/ETL/EtlLoader.cs
```diff
@@ -492,5 +491,60 @@ public void HandleDatabaseValueChanged(DatabaseRecord record)
...
+
+            return lastProcessedTombstones;
+        }
+
+        private void MarkTombstonesForDeletion<T>(EtlConfiguration<T> config, Dictionary<string, long> lastProcessedTombstones) where T : ConnectionString
+        {
+            foreach (var transform in config.Transforms)
+            {
+                var state = EtlProcess.GetProcessState(_database, config.Name, transform.Name);
+                var etag = ChangeVectorUtils.GetEtagById(state.ChangeVector, _database.DbBase64Id);
```
arekpalinski, line 518: Why you no longer use `EtlProcessState.GetLastProcessedEtagForNode` ?

src/Raven.Server/Documents/ETL/EtlLoader.cs
karmeli87: Because even node that isn't responsible for the ETL should take it into account.
It can't purge the tombstone until the responsible  node sent it successfully.

src/Raven.Server/Documents/ETL/EtlLoader.cs
karmeli87: This dictionary is updated on the responsible node.
How the other nodes should know if it is okay to delete their tombstones?

src/Raven.Server/Documents/ETL/EtlLoader.cs
arekpalinski: Oh, I see it now. Thanks

src/Raven.Server/Documents/ETL/EtlLoader.cs
arekpalinski: Makes sense.

arekpalinski (approved):

src/Raven.Server/Documents/PeriodicBackup/PeriodicBackupRunner.cs
```diff
@@ -634,13 +635,37 @@ private PeriodicBackupStatus GetBackupStatusFromCluster(long taskId)
...
+        private long GetMinLastEtag()
+        {
+            var min = long.MaxValue;
 
-                var periodicBackupStatusJson = JsonDeserializationClient.PeriodicBackupStatus(statusBlittable);
-                return periodicBackupStatusJson;
+            using (_serverStore.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
+            using (context.OpenReadTransaction())
+            {
+                var record = _serverStore.Cluster.ReadRawDatabaseRecord(context, _database.Name);
```
grisha-kotler, line 660: When we have a backup with `Full Backup/Snapshot` without an incremental, we don't need to keep the tombstones at all.

grisha-kotler (changes requested):

src/Raven.Server/Documents/PeriodicBackup/PeriodicBackupRunner.cs
karmeli87: I think we do need to keep the tombstones... consider we have an ETL or external replication task that didn't sent the tombstones yet.

grisha-kotler: We don't backup tombstones during a full backup

karmeli87: done

grisha-kotler (approved):

PR #9950, opened 2019-12-11
RavenDB-13078 Making sure TS indexes property cleanup references if a…

arekpalinski:
… source document is deleted. Proper handling of references in TS map-reduce indexes

### DIFF
--- a/src/Raven.Server/Documents/Indexes/IndexStorage.cs
+++ b/src/Raven.Server/Documents/Indexes/IndexStorage.cs
@@ -589,6 +589,35 @@ namespace Raven.Server.Documents.Indexes
             }
         }
 
+        public void RemoveReferencesByPrefix(Slice prefixKey, string collection, HashSet<Slice> referenceKeysToSkip, RavenTransaction tx)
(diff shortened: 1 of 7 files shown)


src/Raven.Server/Documents/Indexes/IndexStorage.cs
```diff
@@ -589,6 +589,35 @@ public void RemoveReferences(Slice key, string collection, HashSet<Slice> refere
...
+                if (it.Seek(prefixKey) == false)
+                    return;
+
+                do
+                {
+                    var key = it.CurrentKey.Clone(tx.InnerTransaction.Allocator);
+
+                    try
+                    {
+                        RemoveReferences(key, collection, referenceKeysToSkip, tx);
```
ppekrol, line 609: I cannot recall details, but removing from the same tree that we are iterating could cause a corruption.

ayende: This is fine, we iterate once and re-create the iteration afterward

PR #9958, opened 2019-12-13
RavenDB-14368 Fixing incorrect usage of table's PK index iterator. Le…

arekpalinski:
…t's evaluate results immediately so we'll be protected from running into such issues in the future.

### DIFF
--- a/src/Raven.Server/ServerWide/ClusterStateMachine.cs
+++ b/src/Raven.Server/ServerWide/ClusterStateMachine.cs
@@ -977,7 +977,7 @@ namespace Raven.Server.ServerWide
 
                 var tasks = new List<Func<Task>>();
 
-                foreach (var record in ReadAllDatabases(context))
+                foreach (var record in GetAllDatabases(context))
                 {
                     using (Slice.From(context.Allocator, "db/" + record.DatabaseName.ToLowerInvariant(), out Slice lowerKey))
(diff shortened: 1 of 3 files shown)


ppekrol (approved):

ayende (approved):

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -2425,59 +2425,71 @@ private static unsafe long ReadIdentitiesIndex(TableValueReader reader)
...
                     if (take-- <= 0)
-                        yield break;
+                        break;
 
                     var doc = Read(context, GetCurrentItemKey(result.Value));
                     if (doc == null)
                         continue;
 
-                    yield return JsonDeserializationCluster.DatabaseRecord(doc);
+                    records.Add(JsonDeserializationCluster.DatabaseRecord(doc));
```
grisha-kotler, line 2488: Can we avoid the deserialization and return the blittable?

ayende: Likely not. You are modifying the data, after all.
And the blittable just point to the data.

ppekrol: This is not used in any hot-paths at the moment. IIRC this is used in RemoveNodeFromCluster, which is not that common.

PR #9959, opened 2019-12-13
RavenDB-13078 querying

ppekrol:


ppekrol:
TODO:
- Map-Reduce (I suspect it should work out of the box)
- Session should not track those results
- Session.Query - check isMapReduce usage

src/Raven.Client/Constants.cs
```diff
@@ -241,6 +241,8 @@ private Fields()
 
                     public const string ReduceKeyValueFieldName = "key()";
 
+                    public const string ValueFieldName = "value()";
```
ayende, line 244: I'm not sure that I'm following, what are you storing under this field name?

ppekrol: Result of Map function for Map TS indexes.

ayende: Why is that needed? Aren't we outputting the fields as usual here?

ppekrol: This is how we are handling that in Map-Reduce and since Map TS is behaving similar to Map-Reduce (qyerying-wise). I;m storing outcome same way as there. The returned query results from Map TS are exactly the values from the index, not the TS entries.

ayende: I see, make sense, then.

PR #9961, opened 2019-12-15
RavenDB-14175 Infrastructure to broadcast commands

karmeli87:


src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -772,235 +803,295 @@ protected virtual bool TryLoadFromCache(JsonOperationContext context)
...
+        {
+            try
+            {
+                if (sessionInfo != null)
+                {
+                    sessionInfo.AsyncCommandRunning = true;
                 }
-                catch (HttpRequestException e) // server down, network down
+
+                Console.WriteLine(command.GetType());
```
grisha-kotler: Can be removed

grisha-kotler: Can be removed

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1342,6 +1448,64 @@ public static async Task<Stream> ReadAsStreamUncompressedAsync(HttpResponseMessa
...
+            if (TopologyNodes == null || 
+                TopologyNodes.Count < 2)
+                return false;
+
+            return true;
+        }
+
+        private async Task<TResult> Broadcast<TResult>(RavenCommand<TResult> command, SessionInfo sessionInfo, CancellationToken token)
+        {
+            if (command is IBroadcast == false)
```
grisha-kotler:
```suggestion
            var iBroadcast = command as IBroadcast 
            if (iBroadcast == null)
```

and use it later

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1342,6 +1448,64 @@ public static async Task<Stream> ReadAsStreamUncompressedAsync(HttpResponseMessa
...
+                    var failed = tasks[completed];
+                    var node = _nodeSelector.Topology.Nodes[failed.Index];
+                    command.FailedNodes[node] = completed.Exception.ExtractSingleInnerException();
+                    _nodeSelector.OnFailedRequest(failed.Index);
+
+                    tasks.Remove(completed);
+                    continue;
+                }
+                _nodeSelector.RestoreNodeIndex(tasks[completed].Index);
+                return tasks[completed].Command.Result;
```
grisha-kotler, line 1520: What happens to the other tasks?
Maybe we can pass `LinkedToken` and cancel them.

grisha-kotler: Or better, record it as a failed request.
If we have the same command, will it succeed on all of the nodes? (if they are online)

grisha-kotler (changes requested):

src/Raven.Client/Http/RequestExecutor.cs
ayende:
```suggestion
            if (command is IBroadcast broadcast == false)
```

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1342,6 +1448,64 @@ public static async Task<Stream> ReadAsStreamUncompressedAsync(HttpResponseMessa
...
+            var tasks = new Dictionary<Task, (RavenCommand<TResult> Command, int Index)>();
+            command.FailedNodes = new Dictionary<ServerNode, Exception>(); // clear the current failures
+
+            for (var index = 0; index < _nodeSelector.Topology.Nodes.Count; index++)
+            {
+                var node = _nodeSelector.Topology.Nodes[index];
+
+                var returnContext = ContextPool.AllocateOperationContext(out JsonOperationContext ctx);
+
+                var clone = (RavenCommand<TResult>)((IBroadcast)command).PrepareToBroadcast(ctx, Conventions);
```
ayende:
```suggestion
                var clone = (RavenCommand<TResult>)(broadcast .PrepareToBroadcast(ctx, Conventions);
```

src/Raven.Client/Http/RequestExecutor.cs
ayende: We need to cancel the others, I think.
Important, we have to be sure that we dispose the ctx properly.

src/Raven.Client/Http/RequestExecutor.cs
karmeli87: funny, compilation failed because "Use of unassigned local variable 'broadcast".
But `!(command is IBroadcast broadcast)` works.
Will use girsha's solution..

grisha-kotler (approved):

src/Raven.Client/Http/RavenCommand.cs
```diff
@@ -173,6 +193,11 @@ public virtual void OnResponseFailure(HttpResponseMessage response)
         {
             
         }
+
+        public void SetTimeout(TimeSpan timeout)
```
ppekrol: internal?

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
```diff
@@ -229,6 +231,38 @@ public DocumentConventions()
             }
         }
 
+        /// <summary>
+        /// Set the timeout for the second broadcast attempt.
+        /// Default: 30 Seconds.
+        ///
+        /// Upon failure of the first attempt the request executor will resend the command to all nodes simultaneously. 
+        /// </summary>
+        public TimeSpan? SecondBroadcastAttemptTimeout
```
ppekrol: Set defaults in `DocumentConventions` ctor and remove nullable

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
```diff
@@ -229,6 +231,38 @@ public DocumentConventions()
...
+            }
+        }
+
+        /// <summary>
+        /// Set the timeout for the first broadcast attempt.
+        /// Default: 5 Seconds.
+        ///
+        /// First attempt will send a single request to a selected node.
+        /// </summary>
+        public TimeSpan? FirstBroadcastAttemptTimeout
```
ppekrol: Set defaults in `DocumentConventions` ctor and remove nullable

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1342,6 +1447,83 @@ public static async Task<Stream> ReadAsStreamUncompressedAsync(HttpResponseMessa
...
+        private async Task<TResult> WaitForResult<TResult>(RavenCommand<TResult> command, Dictionary<Task, (RavenCommand<TResult> Command, int Index)> tasks, CancellationTokenSource broadcastCts)
+        {
+            while (tasks.Count > 0)
+            {
+                var completed = await Task.WhenAny(tasks.Keys).ConfigureAwait(false);
+                if (completed.IsCanceled || completed.IsFaulted)
+                {
+                    var failed = tasks[completed];
+                    var node = _nodeSelector.Topology.Nodes[failed.Index];
+                    command.FailedNodes[node] = completed.Exception.ExtractSingleInnerException();
```
ppekrol: If task is canceled I'm not 100% sure if Exception is set?

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1342,6 +1447,83 @@ public static async Task<Stream> ReadAsStreamUncompressedAsync(HttpResponseMessa
...
+                    _nodeSelector.OnFailedRequest(failed.Index);
+
+                    tasks.Remove(completed);
+                    continue;
+                }
+
+                broadcastCts.Cancel(throwOnFirstException: false);
```
ppekrol, line 1517: Don't we need to wait here for other tasks to 'complete' (become canceled) ?

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1342,6 +1447,83 @@ public static async Task<Stream> ReadAsStreamUncompressedAsync(HttpResponseMessa
...
+
+            using (var broadcastCts = CancellationTokenSource.CreateLinkedTokenSource(token))
+            {
+                var tasks = SendToAllNodes<TResult>(sessionInfo, broadcastCommand, broadcastCts.Token);
+
+                return await WaitForResult(command, tasks, broadcastCts).ConfigureAwait(false);
+            }
+        }
+
+        private async Task<TResult> WaitForResult<TResult>(RavenCommand<TResult> command, Dictionary<Task, (RavenCommand<TResult> Command, int Index)> tasks, CancellationTokenSource broadcastCts)
```
ppekrol: WaitForBroadcastResult

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1342,6 +1447,83 @@ public static async Task<Stream> ReadAsStreamUncompressedAsync(HttpResponseMessa
...
+                _nodeSelector.RestoreNodeIndex(tasks[completed].Index);
+                return tasks[completed].Command.Result;
+            }
+
+            var ae = new AggregateException(command.FailedNodes.Select(x => new UnsuccessfulRequestException(x.Key.Url, x.Value)));
+            throw new AllTopologyNodesDownException($"Broadcasting {command.GetType()} failed.", ae);
+        }
+
+        private Dictionary<Task, (RavenCommand<TResult> Command, int Index)> SendToAllNodes<TResult>(SessionInfo sessionInfo, IBroadcast command, CancellationToken token)
+        {
```
ppekrol, line 1528: if this method will throw we will leak the context, maybe we should have a try/catch clause here and in case of an exception we should cancel everything and return all contexts?

src/Raven.Client/Http/RavenCommand.cs
karmeli87: done

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
karmeli87: done

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
karmeli87: done

src/Raven.Client/Http/RequestExecutor.cs
karmeli87: done

src/Raven.Client/Http/RequestExecutor.cs
karmeli87: done

src/Raven.Client/Http/RequestExecutor.cs
karmeli87: done

src/Raven.Client/Http/RequestExecutor.cs
karmeli87: I think we want to return as quickly as possible.

ppekrol (approved):

PR #9962, opened 2019-12-15
RavenDB-14317 Add NGram Analyzer & all indexing configuration options

Danielle9897:


grisha-kotler (approved):

src/Raven.Studio/typescript/models/database/index/configurationItem.ts
```diff
@@ -3,13 +3,16 @@ import jsonUtil = require("common/jsonUtil");
...
-        configuration.indexing.mapTimeout,
-        configuration.indexing.mapTimeoutAfterEtagReached
+
+    static readonly ServerWideIndexingConfigurationOptions = [        
+        configuration.indexing.size,
+        configuration.indexing.cleanupInterval,
+        configuration.indexing.maxTimeToWaitAfterFlushAndSyncWhenExceedingScratchSpaceLimit
     ];
 
+    static readonly PerDatabaseIndexingConfigurationOptions: Array<string> = _.sortBy(_.pullAll(_.values<string>(configuration.indexing),
```
ml054, line 13: do you know that pullAll mutates input array? Is it desired here?

Danielle9897: 1. Here it doesn't matter that the input array is mutated -
    because it is the array resulting from `_.values()`, Not the original list.
2. I can use `_.difference` if you prefer

ml054: ok you are right we have values there as well, so it is fine.

ml054 (approved):

src/Raven.Studio/typescript/models/database/index/configurationItem.ts
```diff
@@ -3,13 +3,16 @@ import jsonUtil = require("common/jsonUtil");
...
 
 class configurationItem {
-    static readonly ConfigurationOptions = [
-        configuration.indexing.maxTimeForDocumentTransactionToRemainOpen,
-        configuration.indexing.minNumberOfMapAttemptsAfterWhichBatchWillBeCanceledIfRunningLowOnMemory,
-        configuration.indexing.mapTimeout,
-        configuration.indexing.mapTimeoutAfterEtagReached
+
+    static readonly ServerWideIndexingConfigurationOptions = [        
+        configuration.indexing.size,
```
ppekrol, line 8: This list does not contain all ServerWide (only) configurations. Also the 'size' suggests that there is something clearly wrong with parsing those. I've created a separate issue to handle that. This is RegEx issue probably: https://issues.hibernatingrhinos.com/issue/RavenDB-14374

@Danielle9897 Can you fill the list with all the options that are parsed by the current implementation, then after RavenDB-14374 is done, we will update the list?

Danielle9897: Looking at _IndexingConfiguration.cs_ I see only 3 server-wide options :

* CleanupInterval
* GlobalScratchSpaceLimit
* MaxTimeToWaitAfterFlushAndSyncWhenExceedingScratchSpaceLimit

=> What are the other configurations that you mention ?

ppekrol: I was missing 'GlobalScratchSpaceLimit' but I guess this is 'size' ?

Danielle9897: Yes
Taken from _configuration.ts:_

```
size: "Indexing.GlobalScratchSpaceLimitInMb",
cleanupInterval: "Indexing.CleanupIntervalInMin",
maxTimeToWaitAfterFlushAndSyncWhenExceedingScratchSpaceLimit: "Indexing.MaxTimeToWaitAfterFlushAndSyncWhenExceedingScratchSpaceLimit",
```

PR #9963, opened 2019-12-15
RavenDB-14364: Use blittable writer instead of modifying metadata

garayx:


src/Raven.Server/Json/BlittableJsonTextWriterExtensions.cs
```diff
@@ -1497,7 +1497,7 @@ private static async Task WriteCountersForDocumentAsync(this AsyncBlittableJsonT
             writer.WriteEndObject();
         }
 
-        public static void WriteMetadata(this AbstractBlittableJsonTextWriter writer, Document document, BlittableJsonReaderObject metadata)
+        public static void WriteMetadata(this AbstractBlittableJsonTextWriter writer, Document document, BlittableJsonReaderObject metadata, Func<string, bool> method = null)
```
ayende:
```suggestion
        public static void WriteMetadata(this AbstractBlittableJsonTextWriter writer, Document document, BlittableJsonReaderObject metadata, Func<string, bool> filterMetadataProperty= null)
```

ayende: Naming is important

src/Raven.Server/Smuggler/Documents/StreamDestination.cs
```diff
@@ -683,13 +681,23 @@ public void WriteDocument(DocumentItem item, SmugglerProgressBase.CountsWithLast
...
+                    Writer.WriteDocumentPropertiesWithoutMetadata(_context, document, out BlittableJsonReaderObject metadata);
 
-                    using (var old = document.Data)
-                    using (var data = _context.ReadObject(document.Data, document.Id))
+                    if (metadata != null)
                     {
-                        _context.Write(Writer, data);
+                        Writer.WriteComma();
+                        Writer.WriteMetadata(document, metadata, s =>
+                        {
```
ayende: I think it actually might be better to do this outside. That is, create 4 methods to do this filter.
Select between them based on the `_OperateOnTypes` in the ctor, and then pass it in.
That would save some indirection in a hot code path

ayende (changes requested):

src/Raven.Server/Json/BlittableJsonTextWriterExtensions.cs
garayx: done

src/Raven.Server/Smuggler/Documents/StreamDestination.cs
garayx: done, please check if I understood correctly.

src/Raven.Server/Smuggler/Documents/StreamDestination.cs
```diff
@@ -54,6 +55,23 @@ public IDisposable Initialize(DatabaseSmugglerOptionsServerSide options, Smuggle
             _writer = new BlittableJsonTextWriter(_context, _gzipStream);
             _options = options;
 
+            var skipCountersMetadata = _options.OperateOnTypes.HasFlag(DatabaseItemType.CounterGroups) == false;
+            var skipAttachmentsMetadata = _options.OperateOnTypes.HasFlag(DatabaseItemType.Attachments) == false;
+
+            _filterMetadataProperty = metadataProperty =>
+            {
+                switch (metadataProperty)
```
ppekrol: add 

```
if (skipCountersMetadata == false && skipAttachmentsMetadata  == false)
    return false;
```

This way you will avoid doing string comparisons for each metadata property if both of them are `false`

src/Raven.Server/Smuggler/Documents/StreamDestination.cs
```diff
@@ -40,6 +40,7 @@ public class StreamDestination : ISmugglerDestination
         private readonly DatabaseSource _source;
         private BlittableJsonTextWriter _writer;
         private static DatabaseSmugglerOptionsServerSide _options;
+        private static Func<string, bool> _filterMetadataProperty;
```
ppekrol: It would be better to have `Func<LazyStringValue, bool>` to avoid converting each metadata property to string

src/Raven.Server/Smuggler/Documents/StreamDestination.cs
garayx: done

src/Raven.Server/Smuggler/Documents/StreamDestination.cs
garayx: done

ppekrol (approved):

PR #9964, opened 2019-12-16
Revert "RavenDB-13902: vfork, re-PRing"

aviviadi:
Reverts ravendb/ravendb#9865


ayende:
why do this?

ppekrol:
Merged by accident, before review. We have handled that.

PR #9968, opened 2019-12-16
RavenDB-14325 Tombstone cleaner should take into account external rep…

karmeli87:
…lication


ppekrol:
@karmeli87 can you check failing test?

ayende (approved):

karmeli87:
should be fixed now

PR #9977, opened 2019-12-18
RavenDB-14325 Allow multiple tombstone in different collections

karmeli87:
So far we can't have two tombstones in different collections and we need for the cleaner to claim the old one before we can generate a new tombstone.
This is a problem because if you have an etl, external replication to a non-existing/broken destination the old tombstone will be never claimed and breaks replication for ever.

The idea here is to add a suffix to the tombstone ID if one is already exists (maybe we should always added the suffix?).
This way we don't have conflict. Indexing, replication and all other feature shouldn't behave differently, since strip the suffix when materializing the tombstone.

### DIFF
--- a/src/Raven.Server/Documents/DocumentPutAction.cs
+++ b/src/Raven.Server/Documents/DocumentPutAction.cs
@@ -599,12 +599,16 @@ namespace Raven.Server.Documents
             var tombstoneTable = context.Transaction.InnerTransaction.OpenTable(TombstonesSchema, collectionName.GetTableName(CollectionTableType.Tombstones));
             using (Slice.External(context.Allocator, lowerId, lowerSize, out Slice id))
             {
-                if (tombstoneTable.ReadByKey(id, out var reader) == false)
-                    return;
-
-                if (tombstoneTable.IsOwned(reader.Id))
+                foreach (var (tombstoneKey, tvh) in tombstoneTable.SeekByPrimaryKeyPrefix(id, Slices.Empty, 0))
                 {
-                    tombstoneTable.Delete(reader.Id);
+                    if (IsTombstoneOfId(tombstoneKey, id) == false)
+                        return;
+
+                    if (tombstoneTable.IsOwned(tvh.Reader.Id))
+                    {
+                        tombstoneTable.Delete(tvh.Reader.Id);
+                        return; // there could be only one tombstone per collection
+                    }
                 }
             }
         }
--- a/src/Raven.Server/Documents/DocumentsStorage.cs
+++ b/src/Raven.Server/Documents/DocumentsStorage.cs
@@ -831,13 +831,14 @@ namespace Raven.Server.Documents
 
             var tombstoneTable = new Table(TombstonesSchema, context.Transaction.InnerTransaction);
 
-            foreach (var item in tombstoneTable.SeekByPrimaryKeyPrefix(lowerId, Slices.Empty, 0))
+            // return tombstone in any collection with the requested id
+            foreach (var (tombstoneKey, tvh) in tombstoneTable.SeekByPrimaryKeyPrefix(lowerId, Slices.Empty, 0))
             {
-                if (item.Value != null && IsTombstoneOfId(item.Key, lowerId))
+                if (IsTombstoneOfId(tombstoneKey, lowerId))
                 {
                     return new DocumentOrTombstone
                     {
-                        Tombstone = TableValueToTombstone(context, ref item.Value.Reader)
+                        Tombstone = TableValueToTombstone(context, ref tvh.Reader)
                     };
                 }
                 break;
@@ -1259,12 +1260,18 @@ namespace Raven.Server.Documents
                 if (expectedChangeVector != null)
                     throw new ConcurrencyException($"Document {local.Tombstone.LowerId} does not exist, but delete was called with change vector '{expectedChangeVector}'. " +
                                                    "Optimistic concurrency violation, transaction will be aborted.");
-
-                collectionName = ExtractCollectionName(context, local.Tombstone.Collection);
+                if (collectionName == null)
+                {
+                    collectionName = ExtractCollectionName(context, local.Tombstone.Collection);
+                }
 
                 var tombstoneTable = context.Transaction.InnerTransaction.OpenTable(TombstonesSchema,
                     collectionName.GetTableName(CollectionTableType.Tombstones));
-                tombstoneTable.Delete(local.Tombstone.StorageId);
+
+                if (tombstoneTable.IsOwned(local.Tombstone.StorageId))
+                {
+                    tombstoneTable.Delete(local.Tombstone.StorageId);
+                }
 
                 var localFlags = local.Tombstone.Flags.Strip(DocumentFlags.FromClusterTransaction);
                 var flags = localFlags | documentFlags;
@@ -1608,7 +1615,14 @@ namespace Raven.Server.Documents
 
             if (lowerId[lowerId.Size - ConflictedTombstoneOverhead] == SpecialChars.RecordSeparator)
             {
-                return new LazyStringValue(null, lowerId.Buffer, lowerId.Size - ConflictedTombstoneOverhead, context);
+                var size = lowerId.Size - ConflictedTombstoneOverhead;
+                var allocated = context.GetMemory(size + 1); // we need this extra byte to mark that there is no escaping
+                allocated.Address[size] = 0;
+
+                Memory.Copy(allocated.Address, lowerId.Buffer, size);
+                var lsv = context.AllocateStringValue(null, allocated.Address, size);
+                lsv.AllocatedMemoryData = allocated;
+                return lsv;
             }
 
             return lowerId;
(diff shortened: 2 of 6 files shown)


src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1207,6 +1208,7 @@ private static Tombstone TableValueToTombstone(JsonOperationContext context, ref
                 result.Collection = TableValueToId(context, (int)TombstoneTable.Collection, ref tvr);
                 result.Flags = TableValueToFlags((int)TombstoneTable.Flags, ref tvr);
                 result.LastModified = TableValueToDateTime((int)TombstoneTable.LastModified, ref tvr);
+                result.LowerId = UnwrapLowerIdIfNeeded(context, result.LowerId);
```
arekpalinski, line 1224: `result.LowerId` is `LazyStringValue`. Although here we force to materialize it as `string` because that's the argument of `UnwrapLowerIdIfNeeded`.

Please also see my comment below

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1556,14 +1560,44 @@ public void EnsureLastEtagIsPersisted(DocumentsOperationContext context, long do
...
+            
+            return newLowerId;
+        }
+
+        private const long ConflictedTombstoneIdMarkerLong = 20L;
+        private static readonly string ConflictedTombstoneIdMarkerString = Encoding.UTF8.GetString(BitConverter.GetBytes(ConflictedTombstoneIdMarkerLong));
+
+        private static LazyStringValue UnwrapLowerIdIfNeeded(JsonOperationContext context, string lowerId)
+        {
+            var index = lowerId.IndexOf(ConflictedTombstoneIdMarkerString, StringComparison.OrdinalIgnoreCase);
```
arekpalinski: We `lowerId` be `LSV` instead of string? 

Then instead of calling `IndexOf` we could check if at position `lowerId.Size - 2*sizeof(long)` we have `ConflictedTombstoneIdMarkerString`?

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1556,14 +1560,44 @@ public void EnsureLastEtagIsPersisted(DocumentsOperationContext context, long do
...
+            var length = lowerId.Content.Length;
+            Slice.From(context.Allocator, lowerId.Content.Ptr, length + sizeof(long) * 2, out var newLowerId); // TODO: dispose?
+
+            *(long*)(newLowerId.Content.Ptr + length) = ConflictedTombstoneIdMarkerLong;
+            *(long*)(newLowerId.Content.Ptr + length + sizeof(long)) = Bits.SwapBytes(GenerateNextEtag()); // now the id will be unique
+            
+            return newLowerId;
+        }
+
+        private const long ConflictedTombstoneIdMarkerLong = 20L;
```
ayende: Why do you take an 8 bytes value for this?

ayende: We typically use `30`, a byte, the record separator in ASCII

src/Raven.Server/Documents/DocumentsStorage.cs
ayende: Here you would check if the value is > 9, and if so, if lowerId[len-9] == 30

ayende: However, you seems to be checking doing string comparisons here on value that isn't actually a string.

ayende (changes requested):

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1556,14 +1561,45 @@ public void EnsureLastEtagIsPersisted(DocumentsOperationContext context, long do
...
+            var length = lowerId.Content.Length;
+            Slice.From(context.Allocator, lowerId.Content.Ptr, length + ConflictedTombstoneOverhead, out var newLowerId); // TODO: dispose?
+
+            *(newLowerId.Content.Ptr + length) = SpecialChars.RecordSeparator;
+            *(long*)(newLowerId.Content.Ptr + length + sizeof(byte)) = Bits.SwapBytes(GenerateNextEtag()); // now the id will be unique
+            
+            return newLowerId;
+        }
+
+        private const int ConflictedTombstoneOverhead = sizeof(long) + sizeof(byte);
```
arekpalinski, line 1609: Please add a comment that long is etag and byte is separator char

arekpalinski (approved):

arekpalinski:
@karmeli87 One more thing came up do my mind. How `DeleteTombstoneIfNeeded` will behave then?

Will a tombstone with suffix be deleted on document put?

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1556,14 +1561,45 @@ public void EnsureLastEtagIsPersisted(DocumentsOperationContext context, long do
...
 
             return (newEtag, changeVector);
         }
+        private Slice ModifyLowerIdIfNeeded(DocumentsOperationContext context, Table table, Slice lowerId)
+        {
+            if (table.ReadByKey(lowerId, out _) == false)
+                return lowerId;
+
+            var length = lowerId.Content.Length;
+            Slice.From(context.Allocator, lowerId.Content.Ptr, length + ConflictedTombstoneOverhead, out var newLowerId); // TODO: dispose?
```
arekpalinski: TODO ?

karmeli87: fixed

karmeli87:
@arekpalinski You are correct.
`DeleteTombstoneIfNeeded` will delete only the tombstone of the relevant collection

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -828,11 +830,22 @@ public DocumentOrTombstone GetDocumentOrTombstone(DocumentsOperationContext cont
...
+                        Tombstone = TableValueToTombstone(context, ref item.Value.Reader)
+                    };
+                }
+                break;
+            }
 
             return new DocumentOrTombstone
             {
-                Tombstone = TableValueToTombstone(context, ref tvr)
+                Tombstone = null
```
ppekrol, line 849: null?

src/Raven.Server/Documents/DocumentsStorage.cs
ppekrol: How this affects all GetTombstonesFrom methods? E.g. What are we expecting to export via Smuggler? Let's talk about this on Sunday/Monday.

src/Raven.Server/Documents/DocumentPutAction.cs
```diff
@@ -599,12 +599,16 @@ private static void DeleteTombstoneIfNeeded(DocumentsOperationContext context, C
             var tombstoneTable = context.Transaction.InnerTransaction.OpenTable(TombstonesSchema, collectionName.GetTableName(CollectionTableType.Tombstones));
             using (Slice.External(context.Allocator, lowerId, lowerSize, out Slice id))
             {
-                if (tombstoneTable.ReadByKey(id, out var reader) == false)
-                    return;
-
-                if (tombstoneTable.IsOwned(reader.Id))
+                foreach (var (tombstoneKey, tvh) in tombstoneTable.SeekByPrimaryKeyPrefix(id, Slices.Empty, 0))
```
ayende, line 602: This will delete the existing values in the same collection, right?
But what about removing the postfix for the value when we delete the oldest value?

karmeli87: Yes, this will delete the tombstone only if it in the given collection.

We don't care about removing the suffix, it is transparent to the other components since we always unwrap it (for replication, indexing, etl, etc.).

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1608,7 +1609,7 @@ private static LazyStringValue UnwrapLowerIdIfNeeded(JsonOperationContext contex
 
             if (lowerId[lowerId.Size - ConflictedTombstoneOverhead] == SpecialChars.RecordSeparator)
             {
-                return new LazyStringValue(null, lowerId.Buffer, lowerId.Size - ConflictedTombstoneOverhead, context);
+                return context.GetLazyStringValue(lowerId.Buffer);
```
karmeli87: @arekpalinski I do need to parse it for the export/backup, so I ended up doing it like this.

ppekrol: Are we unwrapping here anything now? It looks like it takes the whole buffer so it will be the same as `lowerId`?

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -114,7 +114,7 @@ public long GetMinimalEtagForReplication()
                     foreach (var external in externals)
                     {
                         var state = GetExternalReplicationState(_server, Database.Name, external.TaskId, ctx);
-                        var myEtag = ChangeVectorUtils.GetEtagById(state.SourceChangeVector, Database.DbBase64Id);
+                        var myEtag = ChangeVectorUtils.GetEtagById(state.DestinationChangeVector, Database.DbBase64Id);
```
ayende, line 117: why this change?

ayende (approved):

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
karmeli87: The "source" change vector is actually `_lastSentChangeVectorDuringHeartbeat`, which normally will be delayed in comparison to the destination change vector, which is the `LastAcceptedChangeVector` that gets updated every replication batch.

ppekrol (approved):

PR #9987, opened 2019-12-20
RavenDB-14381 Allow to modify the database record from the studio

ml054:


src/Raven.Studio/typescript/commands/resources/saveDatabaseSettingsCommand.ts
```diff
@@ -0,0 +1,23 @@
+import commandBase = require("commands/commandBase");
+import database = require("models/resources/database");
+import endpoints = require("endpoints");
+
+class saveDatabaseSettingsCommand extends commandBase {
```
ppekrol, line 5: maybe `createDatabaseCommand` ? It is called like that in the Client API

ml054: (talked with @ppekrol about that personally) - solved

PR #9989, opened 2019-12-20
RavenDB-13100

ml054:


src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
```diff
@@ -411,6 +409,28 @@ internal SeriesSummary GetSummary()
...
+
+                var result = 0L;
+
+                do
+                {
+                    
+                    if (_currentSegment.NumberOfEntries == 0)
+                        continue;
+                    
+                    result += _currentSegment.SegmentValues.Span[0].Count;
```
aviv86, line 435: summing up `_currentSegment.SegmentValues.Span[0].Count` is incorrect
we should sum up `_currentSegment.NumberOfEntries`
however, `_currentSegment.NumberOfEntries` does not exclude the "dead" entries.. so we might also need to keep a "dead count" in the `SegmentHeader`

ayende: why is this incorrect?

aviv86: for example, if we have series (item1, item2) :
item1.Values = [1, 2, 3]
item2.Values = [NaN, 2, 3]
then `Span[0].Count` is 1
but I expect `GetNumberOfEntries()` to return 2

ayende: I'm pretty sure that this is not a valid option.
The `NaN` can only appear at the end, not at start / middle

ml054: IIRC I copy that from other place (in time series handler)? Could you push correct version to this PR? I'm not familiar with implementation, so I was writing this by mimic other endpoints.

ml054: ok I took it from here: https://github.com/ravendb/ravendb/blob/85e1eb64b043c29ad724e57748852062ca304ea6/src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs#L406

I'm not sure if it is correct. @aviv86 Could you verify?

ml054: reported as separate issue:
https://issues.hibernatingrhinos.com/issue/RavenDB-14429

ayende (approved):

PR #9995, opened 2019-12-23
RavenDB-13079 AbstractTimeSeriesIndexCreationTask

ppekrol:


src/Raven.Client/Documents/Indexes/TimeSeries/AbstractTimeSeriesIndexCreationTask.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using Raven.Client.Documents.Conventions;
+
+namespace Raven.Client.Documents.Indexes.TimeSeries
+{
+    public abstract class AbstractTimeSeriesIndexCreationTask : AbstractIndexCreationTaskBase<TimeSeriesIndexDefinition>
+    {
+        public class TimeSeriesSegment
```
ayende, line 11: I wonder if these should be higher level classes.
The issue is that users may want to use them in queries, no?

ppekrol: I will sync with @aviv86 after both of our PRs are merged.

PR #9996, opened 2019-12-23
RavenDB-14348 

garayx:


grisha-kotler (approved):

test/RachisTests/RavenDB_14348.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace RachisTests
+{
+    public class RavenDB_14348 : RavenTestBase
+    {
+        public RavenDB_14348(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        [Fact]
+        public async Task ShouldWork()
```
ayende, line 15: What are you actually testing here?

test/RachisTests/RavenDB_14348.cs
```diff
@@ -0,0 +1,28 @@
...
+        {
+            using (var store = GetDocumentStore())
+            {
+                var task = Task.Run(async () =>
```
ayende, line 19: This task "leaks"

test/Tests.Infrastructure/RavenTestBase.cs
```diff
@@ -237,6 +237,19 @@ protected virtual DocumentStore GetDocumentStore(Options options = null, [Caller
                                 await WaitForRaftIndexToBeAppliedInCluster(result.RaftCommandIndex, timeout);
                             });
                         }
+
+                        // skip 'wait for requests' on DocumentDatabase dispose
+                        Servers.ForEach(x =>
```
ayende: Why?

ayende (changes requested):

test/Tests.Infrastructure/RavenTestBase.cs
garayx: On store dispose we send command to delete the db, and then WaitForIndexNotification of that command, which has a timeout of 15 sec, but the DB dispose has '[drain requests code](https://github.com/ravendb/ravendb/blob/v4.2/src/Raven.Server/Documents/DocumentDatabase.cs#L633)' which can take up to 60 secs, and make the WaitForIndexNotification throw timeout exception and make the test fail.

test/RachisTests/RavenDB_14348.cs
garayx: 1. On store dispose, the WaitForIndexNotification doesn't gets timeout exception, because of  '[drain requests code](https://github.com/ravendb/ravendb/blob/v4.2/src/Raven.Server/Documents/DocumentDatabase.cs#L633)'
2. In DeleteDatabaseCommand we assign '[new topology](https://github.com/ravendb/ravendb/blob/v4.2/src/Raven.Server/ServerWide/Commands/DeleteDatabaseCommand.cs#L74)' _without_ databaseId, if we have ClusterTransactions kicking in for that DB, they would throw on '[wrong serialization](https://github.com/ravendb/ravendb/blob/v4.2/src/Raven.Server/ServerWide/ClusterStateMachine.cs#L764)'
3. After db deleted and ClusterTransactions are still kicking in, the db record would be null, and we get NRE.

test/RachisTests/RavenDB_14348.cs
garayx: it is stopped when store gets disposed
https://github.com/ravendb/ravendb/blob/v4.2/test/RachisTests/SubscriptionFailoverWIthWaitingChains.cs#L310

ayende (approved):

ppekrol (approved):

PR #10003, opened 2019-12-25
RavenDB-14406 Pass cancellation token to `ReadForMemoryAsync`

karmeli87:


src/Raven.Client/Http/RavenCommand.cs
```diff
@@ -145,8 +149,7 @@ public virtual async Task<ResponseDisposeHandling> ProcessResponse(JsonOperation
                 }
 
                 // We do not cache the stream response.
-                using (var uncompressedStream = await RequestExecutor.ReadAsStreamUncompressedAsync(response).ConfigureAwait(false))
```
ayende, line 148: Where is the call to `ReadAsStreamUncompressedAsync` now?

ayende (changes requested):

src/Raven.Client/Http/RavenCommand.cs
karmeli87: This is draft PR so I temperately removed it, since I noticed that `await response.Content.ReadAsStreamAsync()` return `GZipStream`, so it make me think that maybe we don't need that function any longer (but I need to dig into it).

karmeli87: Maybe I wrong, but to me it seems that `ReadAsStreamUncompressedAsync` was relevant only if the server return zipped response without the client supporting it. 

Is it even possible now? and if yes, why it was relevant only for `Raw` and not for `Object` response type as well.

But let's tackle this in a different issue I guess.

ayende: No, you are correct that the http handler should be in charge of this

ayende (changes requested):
Need to change how we do this. And probably need to cosndier why we even need a stream instead instead of using timed cancellation token

src/Raven.Client/Util/StreamWithTimeout.cs
```diff
@@ -0,0 +1,141 @@
...
+        public override bool CanSeek => _stream.CanSeek;
+        public override bool CanWrite => _stream.CanWrite;
+        public override long Length => _stream.Length;
+        public override long Position
+        {
+            get => _stream.Position;
+            set => _stream.Position = value;
+        }
+
+        public new void Dispose()
```
ppekrol: You need to override the Dispose(bool)

src/Sparrow/Utils/TimeoutManager.cs
```diff
@@ -111,6 +112,32 @@ private static TimerTaskHolder GetHolderForDuration(uint duration)
             return value;
         }
 
+        public static async Task<T> WaitForTaskCompletion<T>(this Task<T> task, TimeSpan duration, CancellationTokenSource source = null)
```
ppekrol: internal

src/Sparrow/Utils/TimeoutManager.cs
```diff
@@ -111,6 +112,32 @@ private static TimerTaskHolder GetHolderForDuration(uint duration)
...
+            if (t == timeoutTask)
+            {
+                source?.Cancel();
+                throw new TimeoutException();
+            }
+
+            return await task.ConfigureAwait(false);
+        }
+
+        public static async Task WaitForTaskCompletion(this Task task, TimeSpan duration, CancellationTokenSource source = null)
```
ppekrol: internal

src/Sparrow/Utils/TimeoutManager.cs
```diff
@@ -111,6 +112,32 @@ private static TimerTaskHolder GetHolderForDuration(uint duration)
...
         }
 
+        public static async Task<T> WaitForTaskCompletion<T>(this Task<T> task, TimeSpan duration, CancellationTokenSource source = null)
+        {
+            var timeoutTask = WaitFor(duration);
+            var t = await Task.WhenAny(timeoutTask, task).ConfigureAwait(false);
+            if (t == timeoutTask)
+            {
+                source?.Cancel();
+                throw new TimeoutException();
```
ppekrol: Can we put a message here

src/Sparrow/Utils/TimeoutManager.cs
```diff
@@ -111,6 +112,32 @@ private static TimerTaskHolder GetHolderForDuration(uint duration)
...
+        }
+
+        public static async Task WaitForTaskCompletion(this Task task, TimeSpan duration, CancellationTokenSource source = null)
+        {
+            var timeoutTask = WaitFor(duration);
+            var t = await Task.WhenAny(timeoutTask, task).ConfigureAwait(false);
+            if (t == timeoutTask)
+            {
+                source?.Cancel();
+                throw new TimeoutException();
```
ppekrol: Can we put a message here

ppekrol (approved):

src/Raven.Client/Util/StreamWithTimeout.cs
karmeli87: done

src/Sparrow/Utils/TimeoutManager.cs
karmeli87: removed those 2 methods.

ppekrol:
@ayende can you do final review?

src/Raven.Client/Util/StreamWithTimeout.cs
```diff
@@ -0,0 +1,141 @@
...
+            else
+                _readTimeout = (int)(readTimeout ?? DefaultReadTimeout).TotalMilliseconds;
+        }
+
+        private void SetWriteTimeoutIfNeeded(TimeSpan? writeTimeout)
+        {
+            _canBaseStreamTimeoutOnWrite = _stream.CanTimeout && _stream.WriteTimeout < int.MaxValue;
+
+            if (_canBaseStreamTimeoutOnWrite)
+                _writeTimeout = (int?)writeTimeout?.TotalMilliseconds ?? _stream.WriteTimeout;
```
ayende, line 58: Shouldn't we set the `WriteTimeout` here? If there is a smaller timeout already defined?

src/Raven.Client/Util/StreamWithTimeout.cs
```diff
@@ -0,0 +1,141 @@
...
+            SetWriteTimeoutIfNeeded(writeTimeout);
+            SetReadTimeoutIfNeeded(readTimeout);
+        }
+
+        private void SetReadTimeoutIfNeeded(TimeSpan? readTimeout)
+        {
+            _canBaseStreamTimeoutOnRead = _stream.CanTimeout && _stream.ReadTimeout < int.MaxValue;
+
+            if (_canBaseStreamTimeoutOnRead)
+                _readTimeout = (int?)readTimeout?.TotalMilliseconds ?? _stream.ReadTimeout;
```
ayende, line 33: Set `ReadTimeout` ?

src/Raven.Client/Util/StreamWithTimeout.cs
```diff
@@ -0,0 +1,141 @@
...
+                return _stream.WriteAsync(buffer, offset, count, cancellationToken);
+            }
+
+            return WriteAsyncWithTimeout(buffer, offset, count, cancellationToken);
+        }
+
+        private Task WriteAsyncWithTimeout(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            _cts?.Dispose();
+            _cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
```
ayende: It is fine (and fairly common) to do both read & write operations at the same time.
But you are using the same `_cts` on both

ayende (changes requested):

src/Raven.Client/Util/StreamWithTimeout.cs
karmeli87: Done.

src/Raven.Client/Util/StreamWithTimeout.cs
karmeli87: You right, we need to set the inner stream timeouts as well.

About the smaller value, I think that if we specify a timeout is should take precedence over the existing one,

ayende (approved):

PR #10004, opened 2019-12-29
RavenDB-14414: add failover when copying pal file

garayx:


src/Sparrow.Server/Platform/Pal.cs
```diff
@@ -99,6 +98,33 @@ static Pal()
                 PalHelper.ThrowLastError(rc, errorCode, "Cannot get system information");
         }
 
+        private static bool CopyPalFile(string toFilename, string fromFilename, out Exception exception)
```
ayende:
```suggestion
        private static Exception CopyPalFile(string toFilename, string fromFilename, out Exception exception)
```

src/Sparrow.Server/Platform/Pal.cs
```diff
@@ -45,28 +46,26 @@ static Pal()
...
 
-                if (copy)
-                    File.Copy(fromFilename, toFilename, overwrite: true);
-            }
-            catch (IOException e)
-            {
-                throw new IOException(
-                    $"Cannot copy {fromFilename} to {toFilename}, make sure appropriate {toFilename} to your platform architecture exists in Raven.Server executable folder",
-                    e);
+                throw e;
```
ayende: Here you lost the exception trace, `throw new InvalidOperationException('msg', e);`

ayende (changes requested):

src/Sparrow.Server/Platform/Pal.cs
garayx: done

src/Sparrow.Server/Platform/Pal.cs
garayx: done

ppekrol (approved):

PR #10013, opened 2020-01-02
RavenDB-14332 Allow to filter index that has errors

Danielle9897:


grisha-kotler (approved):

ml054 (changes requested):
The one thing that bothers me is we use word error in multiple meaning so it might be confusing for user. 
Index might have error state (so indexing is paused) or .. index can have *indexing errors*.

So instead of splitting normal status, I would put subsection under normal with extra toggle. It should control we should be show all indexes with normal status or with normal status but *indexing errors* count > 0

ml054:
```
- Normal     (toggle - on by default) 
  - with indexing errors only  (toggle - off by default)
```

Danielle9897:
https://issues.hibernatingrhinos.com/issue/RavenDB-14332#focus=streamItem-67-43581.0-0

ml054 (approved):

PR #10020, opened 2020-01-05
RavenDB-13562 Adding flushed journal to reports

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-13562#comment=67-43490


ayende (approved):
This still needs to be updated in the UI, right?

haludi:
> This still needs to be updated in the UI, right?

Actually, I added the flushed journal report to the rest of the journals report so the stats view is fine now and no need for further steps in the UI

arekpalinski:
It would be nice to expose the newly added property `Flushed` on the UI

PR #10023, opened 2020-01-07
RavenDB-14439 Ensure that @change-vector is exposed to scripts properly

ayende:


grisha-kotler:
Can we have a test for this?

PR #10030, opened 2020-01-08
RavenDB-14442 FollowerAmbassador and CandidateAmassador leak sockets

karmeli87:


ppekrol (approved):

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -2130,6 +2128,12 @@ public static bool IsExpectedException(Exception e)
             return e is OperationCanceledException ||
                    e is ObjectDisposedException;
         }
+
+
+        internal static void DisconnectAction(Task<RachisConnection> connectionTask)
+        {
```
ayende, line 2134: Maybe assert here that the `connectionTask` is completed?

karmeli87: done.
But it can be faulted, so I only assert for `IsCompleted`.

ppekrol (approved):

ayende (approved):

PR #10031, opened 2020-01-08
RavenDB-14347 be more explicit to what address family we want to bind when creating TcpClient

ppekrol:


src/Raven.Client/Util/TcpUtils.cs
```diff
@@ -132,7 +132,7 @@ private static TcpClient NewTcpClient(TimeSpan? timeout, bool useIPv6)
             }
             else
             {
-                tcpClient = new TcpClient();
+                tcpClient = new TcpClient(AddressFamily.InterNetwork);
```
ayende, line 135: That limits us to IPv4 only, no?
I'm not sure that we want this. In tests, sure, but we do want to be able to run on IPv6 as well

ayende (changes requested):

src/Raven.Client/Util/TcpUtils.cs
ppekrol: We do support IPv6 (check few lines up).
In .NET Core 2.2 `new TcpClient()` and `new TcpClient(AddressFamily.InterNetwork)` are doing the same thing.

2.2: 
https://github.com/dotnet/corefx/blob/release/2.2/src/System.Net.Sockets/src/System/Net/Sockets/TCPClient.cs#L23

3.1: 
https://github.com/dotnet/corefx/blob/release/3.1/src/System.Net.Sockets/src/System/Net/Sockets/TCPClient.cs#L23
https://github.com/dotnet/corefx/blob/release/3.1/src/System.Net.Sockets/src/System/Net/Sockets/TCPClient.cs#L478-L486
https://github.com/dotnet/corefx/blob/release/3.1/src/System.Net.Sockets/src/System/Net/Sockets/Socket.cs#L76

ayende (approved):

PR #10038, opened 2020-01-09
RavenDB-14435 Don't stops replication after double dispose

karmeli87:


src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -1096,8 +1097,13 @@ private void AddReplicationPerformance(IncomingReplicationStatsAggregator stats)
                 _lastReplicationStats.TryDequeue(out stats);
         }
 
+        private readonly MultipleUseFlag _disposed = new MultipleUseFlag();
+
         public void Dispose()
         {
+            if (_disposed.Raise() == false)
+                return; // already disposed
```
ppekrol: Wouldn't it be better to use DisposeOnce? It was created for this purpose.

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
```diff
@@ -357,7 +357,7 @@ private void Replicate()
                     if (_parent.DebugWaitAndRunReplicationOnce != null)
                     {
                         _parent.DebugWaitAndRunReplicationOnce.Wait(_cts.Token);
-                        _parent.DebugWaitAndRunReplicationOnce.Reset();
+                        _parent.DebugWaitAndRunReplicationOnce?.Reset();
```
ppekrol, line 360: We checked that this is not null 2 lines up?

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
karmeli87: done

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
karmeli87: in the test I do
```
database.ReplicationLoader.DebugWaitAndRunReplicationOnce = null;
mre.Set();
```

ppekrol (approved):

PR #10042, opened 2020-01-10
RavenDB-14457 added the ability to configure HttpVersion via Conventions

ppekrol:
`null` -> framework default (.Net Core 2.1 and 2.2 -> HTTP 2.0, others 1.1)
`default` -> 1.1


arekpalinski (approved):

ayende:
Why is this needed? I don't believe that we are supporting HTTP 2.0 at the Kestrel level, do we?

PR #10049, opened 2020-01-12
RavenDB-14464: On operation initialization exchange the _result task …

garayx:
…with new if it is already completed

### DIFF
--- a/src/Raven.Client/Documents/Operations/Operation.cs
+++ b/src/Raven.Client/Documents/Operations/Operation.cs
@@ -21,7 +21,7 @@ namespace Raven.Client.Documents.Operations
         private readonly DocumentConventions _conventions;
         private readonly Task _additionalTask;
         private readonly long _id;
-        private readonly TaskCompletionSource<IOperationResult> _result = new TaskCompletionSource<IOperationResult>(TaskCreationOptions.RunContinuationsAsynchronously);
+        private TaskCompletionSource<IOperationResult> _result = new TaskCompletionSource<IOperationResult>(TaskCreationOptions.RunContinuationsAsynchronously);
         private readonly SemaphoreSlim _lock = new SemaphoreSlim(1, 1);
 
         public Action<IOperationProgress> OnProgressChanged;
@@ -55,6 +55,8 @@ namespace Raven.Client.Documents.Operations
 
         private async Task Initialize()
         {
+            _result = _result.Task.IsCompleted ? new TaskCompletionSource<IOperationResult>(TaskCreationOptions.RunContinuationsAsynchronously) : _result;
+
             try
             {
                 await Process().ConfigureAwait(false);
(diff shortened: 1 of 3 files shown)


test/SlowTests/Issues/RavenDB_14464.cs
```diff
@@ -0,0 +1,148 @@
...
+                                await operation.WaitForCompletionAsync(TimeSpan.FromSeconds(1));
+                            }
+                            break;
+                        }
+                        catch (Exception e)
+                        {
+                            if (serverToggled == false)
+                            {
+                                Assert.True(e.GetType() == typeof(TimeoutException));
+                                Assert.True(e.Message.Contains("did not get a reply for operation"));
```
ayende:
```suggestion
                                Assert.Contains("did not get a reply for operation", e.Message);
```

ayende (approved):

test/SlowTests/Issues/RavenDB_14464.cs
garayx: added messages to asserts

test/SlowTests/Issues/RavenDB_14464.cs
```diff
@@ -0,0 +1,148 @@
...
+                            Assert.True(e.GetType() == typeof(InvalidOperationException), "e.GetType() == typeof(InvalidOperationException)");
+                            Assert.True(e.Message.StartsWith("Could not fetch state of operation"), "e.Message.StartsWith('Could not fetch state of operation')");
+                            await Task.Delay(1000);
+                        }
+                    }
+                });
+
+                await Task.Delay(1000);
+                server = await ToggleServer(server);
+                await Task.Delay(5000);
```
ppekrol: If there is no other choice than doing that, then tests like that should be moved to Stress.

src/Raven.Client/Documents/Operations/Operation.cs
```diff
@@ -55,6 +55,8 @@ internal Operation(RequestExecutor requestExecutor, Func<IDatabaseChanges> chang
 
         private async Task Initialize()
         {
+            _result = _result.Task.IsCompleted ? new TaskCompletionSource<IOperationResult>(TaskCreationOptions.RunContinuationsAsynchronously) : _result;
```
ppekrol: There is a race condition between executing 'WaitForCompletionAsync' and 'Initialize'. The 'Initialize' is executed in the background and it can swap the '_result'.

I would suggest creating a separate synchronous method e.g. 'InitializeResult' that will just do this and execute it in the 'WaitForCompletionAsync'.

Check also the behavior of 'OnError'

garayx: done

test/SlowTests/Issues/RavenDB_14464.cs
garayx: done

PR #10053, opened 2020-01-12
Raven db 14462

talweiss1982:


tools/Voron.Recovery/Recovery.cs
```diff
@@ -496,6 +530,24 @@ void PrintRecoveryProgress(long startOffset, byte* mem, byte* eof, DateTime now)
             }
         }
 
+        private void CheckInvalidPagesInARaw(PageHeader* pageHeader, byte* mem)
+        {
+            if(_shouldIgnoreInvalidPagesInARaw)
+                return;
+
+            if (*(short*)pageHeader->Mac != 0)
```
ayende: That isn't correct. the `Mac` is an array of 16 bytes, you need to check that the full array is set to zero

tools/Voron.Recovery/Recovery.cs
```diff
@@ -496,6 +530,24 @@ void PrintRecoveryProgress(long startOffset, byte* mem, byte* eof, DateTime now)
...
+            {
+                if (MaxNumberOfInvalidChecksumWithNoneZeroMac <= _InvalidChecksumWithNoneZeroMac++)
+                {
+                    var error =
+                        $"Found {_InvalidChecksumWithNoneZeroMac} pages in a raw that had invalid checksum and none zero MAC " +
+                        $"this is a strong indication that you're recovering an encrypted database and didn't provide the 'Master Key'";
```
ayende:
```suggestion
                        $"this is a strong indication that you're recovering an encrypted database and didn't provide the encryption key using the  '--MasterKey=<KEY>' command line flag";
```

tools/Voron.Recovery/Recovery.cs
```diff
@@ -910,11 +962,14 @@ private bool ValidateOverflowPage(PageHeader* pageHeader, byte* eof, long startO
...
                 if (checksum != pageHeader->Checksum)
                 {
+                    CheckInvalidPagesInARaw(pageHeader, mem);
                     var message =
                         $"Invalid checksum for overflow page {pageHeader->PageNumber}, expected hash to be {pageHeader->Checksum} but was {checksum}";
                     mem = PrintErrorAndAdvanceMem(message, mem);
                     return false;
                 }
+
+                _InvalidChecksumWithNoneZeroMac = 0;
```
ayende: If we were able to get a _single_ valid page, we don't need to deal with this any longer.

ayende (changes requested):

talweiss1982:
done

talweiss1982:
test is failing

ayende (approved):

tools/Voron.Recovery/CommandLineApp.cs
```diff
@@ -164,6 +164,15 @@ private static void ConfigureRecoveryCommand()
 
                         config.IgnoreDataIntegrityErrorsOfAlreadySyncedTransactions = ignoreDataIntegrityErrorsOfAlreadySyncedTransactions;
                     }
+                    
+                    if (ignoreInvalidPagesInARawArg.HasValue())
+                    {
+                        var value = ignoreInvalidPagesInARawArg.Value();
+                        if (bool.TryParse(value, out var ignoreInvalidPagesInARaw) == false)
+                            return ExitWithError($"{nameof(config.IgnoreDataIntegrityErrorsOfAlreadySyncedTransactions)} argument value ({value}) is invalid", cmd);
```
arekpalinski: Invalid argument name in the message - `{nameof(config.IgnoreDataIntegrityErrorsOfAlreadySyncedTransactions)}`

test/SlowTests/Issues/RavenDB-8451.cs
```diff
@@ -24,8 +24,29 @@ public RavenDB_8451(ITestOutputHelper output) : base(output)
...
+        [Fact64Bit]
+        public async Task RecoveryOfEncryptedDatabaseWithoutMasterKeyShouldThrow()
+        {
+            try
+            {
+                await CanRecoverEncryptedDatabaseInternal(true);
+            }
+            catch (InvalidOperationException ioe)
+            {
+            }
```
arekpalinski: Can we use `Assert.Throws<>()` and check the exception message to be sure that it throws what we need

tools/Voron.Recovery/Recovery.cs
```diff
@@ -442,8 +474,13 @@ void PrintRecoveryProgress(long startOffset, byte* mem, byte* eof, DateTime now)
                                         conflictsWriter, countersWriter, context, startOffset, ((RawDataSmallPageHeader*)page)->TableType) == false)
                                     break;
                             }
+
                             mem += _pageSize;
                         }
+                        catch (InvalidOperationException ioe) when (ioe.Message.Contains("this is a strong indication that you're"))
```
arekpalinski: Can we extract error message to const and use it here?

tools/Voron.Recovery/Recovery.cs
```diff
@@ -496,6 +533,31 @@ void PrintRecoveryProgress(long startOffset, byte* mem, byte* eof, DateTime now)
             }
         }
 
+        private void CheckInvalidPagesInARaw(PageHeader* pageHeader, byte* mem)
```
arekpalinski, line 536: Just to make sure.. what did you mean by "in a raw" ? I see you count it so maybe you meant "in a row" but from the other side I don't see you reset `_InvalidChecksumWithNoneZeroMac` anywhere

talweiss1982: It is only the first pages in a row, once a valid page checksum was found we disable this check by using the flag: _shouldIgnoreInvalidPagesInARaw

arekpalinski (approved):

PR #10056, opened 2020-01-13
RavenDB-14468

ayende:
- We must always return the document flags, we assert on them.
- A time series batch will first process removals, and then appends

### DIFF
--- a/src/Raven.Server/Documents/DocumentsStorage.cs
+++ b/src/Raven.Server/Documents/DocumentsStorage.cs
@@ -1196,17 +1196,11 @@ namespace Raven.Server.Documents
         {
             var result = new Document();
 
-            if (fields.Contain(DocumentFields.StorageId))
-                result.StorageId = tvr.Id;
-
             if (fields.Contain(DocumentFields.LowerId))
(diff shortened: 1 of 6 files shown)


src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1217,8 +1217,7 @@ private static Document ParseDocumentPartial(JsonOperationContext context, ref T
             if (fields.Contain(DocumentFields.LastModified))
                 result.LastModified = TableValueToDateTime((int)DocumentsTable.LastModified, ref tvr);
 
-            if (fields.Contain(DocumentFields.Flags))
-                result.Flags = TableValueToFlags((int)DocumentsTable.Flags, ref tvr);
+            result.Flags = TableValueToFlags((int)DocumentsTable.Flags, ref tvr);
```
ppekrol, line 1213: I think that the issue is that we did not request 'Flags' in one of the flows that specified flags (by default we are retrieving 'All').

ayende: There is no scenarios where we _can_ avoid using `Flags`, we assert on it immediately afterward

ppekrol (approved):

PR #10068, opened 2020-01-15
PAL 32-bit pager, working for Linux WIP

myarichuk:
FastTests and SlowTests pass, this PR is for review


myarichuk:
Merge issues, I will fix then reopen

myarichuk:
The merge with current v4.2 is still bad, it causes AVEs in tests. Will reopen when the merge will properly work.

PR #10072, opened 2020-01-17
RavenDB-14448 added ForNode support to ServerOperationExecutor

ppekrol:


src/Raven.Client/ServerWide/Operations/ServerOperationExecutor.cs
```diff
@@ -10,22 +12,74 @@ namespace Raven.Client.ServerWide.Operations
...
-            store.AfterDispose += (sender, args) => _requestExecutor.Dispose();
+            store.AfterDispose += (sender, args) => Dispose();
         }
 
         public void Dispose()
         {
-            _requestExecutor.Dispose();
+            _requestExecutor?.Dispose();
+
+            if (_nodeTag != null)
```
ayende: Why would we want to do this?
Isn't it better to have a cache that will live as long as the parent live?

ppekrol: We are holding main server operation executor for the whole lifetime of the DocumentStore. I'm assuming that vast amount of usages will be without a dispose, but when someone will dispose then I'm removing that from the cache to not hold request executor.

ayende: The problem is that we may have two threads, one that dispose, one that doesn't.
You give them both the same instance, and it gets dispose under you.

We need to either:
* Give a different instance (and require disposal)
* Give same instance and forbid disposal

ppekrol: Done

PR #10075, opened 2020-01-19
RavenDB-14509 Let user supply external certificate validation callbac…

iftahbe:
…k for the *server*

### DIFF
--- a/src/Raven.Server/Config/Categories/SecurityConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/SecurityConfiguration.cs
@@ -142,6 +142,24 @@ namespace Raven.Server.Config.Categories
         [ConfigurationEntry("Security.WellKnownIssuerHashes.Admin", ConfigurationEntryScope.ServerWideOnly)]
         public string[] WellKnownIssuerHashes { get; set; }
 
+        [Description("EXPERT: A command or executable to validate a server authentication request. " +
+                     "RavenDB will execute: command [user-arg-1] ... [user-arg-n] <sender-url> <base64-certificate> <errors>. " +
+                     "The executable will return a case-insensitive boolean string through the standard output (e.g. true, false) indicating whether to approve the connection.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.CertificateValidation.Exec", ConfigurationEntryScope.ServerWideOnly)]
+        public string CertificateValidationExec { get; set; }
+
+        [Description("EXPERT: The optional user arguments for the 'Security.CertificateValidation.Exec' command or executable. The arguments must be escaped for the command line.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.CertificateValidation.Exec.Arguments", ConfigurationEntryScope.ServerWideOnly)]
+        public string CertificateValidationExecArguments { get; set; }
+
+        [Description("The number of seconds to wait for the Certificate Validation executable to exit. Default: 30 seconds")]
+        [DefaultValue(30)]
+        [TimeUnit(TimeUnit.Seconds)]
+        [ConfigurationEntry("Security.CertificateValidation.Exec.TimeoutInSec", ConfigurationEntryScope.ServerWideOnly)]
+        public TimeSetting CertificateValidationExecTimeout { get; set; }
+
         internal bool? IsUnsecureAccessSetupValid { get; private set; }
 
         internal string UnsecureAccessWarningMessage { get; private set; }
--- a/src/Raven.Server/Config/Categories/SecurityConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/SecurityConfiguration.cs
@@ -146,18 +146,18 @@ namespace Raven.Server.Config.Categories
                      "RavenDB will execute: command [user-arg-1] ... [user-arg-n] <sender-url> <base64-certificate> <errors>. " +
                      "The executable will return a case-insensitive boolean string through the standard output (e.g. true, false) indicating whether to approve the connection.")]
         [DefaultValue(null)]
-        [ConfigurationEntry("Security.CertificateValidation.Exec", ConfigurationEntryScope.ServerWideOnly)]
+        [ConfigurationEntry("Security.Certificate.Validation.Exec", ConfigurationEntryScope.ServerWideOnly)]
         public string CertificateValidationExec { get; set; }
 
-        [Description("EXPERT: The optional user arguments for the 'Security.CertificateValidation.Exec' command or executable. The arguments must be escaped for the command line.")]
+        [Description("EXPERT: The optional user arguments for the 'Security.Certificate.Validation.Exec' command or executable. The arguments must be escaped for the command line.")]
         [DefaultValue(null)]
-        [ConfigurationEntry("Security.CertificateValidation.Exec.Arguments", ConfigurationEntryScope.ServerWideOnly)]
+        [ConfigurationEntry("Security.Certificate.Validation.Exec.Arguments", ConfigurationEntryScope.ServerWideOnly)]
         public string CertificateValidationExecArguments { get; set; }
 
-        [Description("The number of seconds to wait for the Certificate Validation executable to exit. Default: 30 seconds")]
-        [DefaultValue(30)]
+        [Description("The number of seconds to wait for the 'Security.Certificate.Validation.Exec' executable to exit. Default: 5 seconds")]
+        [DefaultValue(5)]
         [TimeUnit(TimeUnit.Seconds)]
-        [ConfigurationEntry("Security.CertificateValidation.Exec.TimeoutInSec", ConfigurationEntryScope.ServerWideOnly)]
+        [ConfigurationEntry("Security.Certificate.Validation.Exec.TimeoutInSec", ConfigurationEntryScope.ServerWideOnly)]
         public TimeSetting CertificateValidationExecTimeout { get; set; }
 
         internal bool? IsUnsecureAccessSetupValid { get; private set; }
--- a/src/Raven.Server/Https/ExternalCertificateValidator.cs
+++ b/src/Raven.Server/Https/ExternalCertificateValidator.cs
@@ -20,7 +20,6 @@ namespace Raven.Server.Https
     {
         private readonly RavenServer _server;
         private readonly Logger _logger;
-        private List<string> _userArgs; 
 
         private ConcurrentDictionary<Key, Task<CachedValue>> _externalCertificateValidationCallbackCache;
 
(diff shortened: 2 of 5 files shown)


iftahbe:
@ppekrol, there is no test because of this:
https://github.com/ravendb/ravendb/blob/v4.2/test/SlowTests/Authentication/AuthenticationClusterTests.cs#L1

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -142,6 +142,24 @@ public class SecurityConfiguration : ConfigurationCategory
...
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.CertificateValidation.Exec", ConfigurationEntryScope.ServerWideOnly)]
+        public string CertificateValidationExec { get; set; }
+
+        [Description("EXPERT: The optional user arguments for the 'Security.CertificateValidation.Exec' command or executable. The arguments must be escaped for the command line.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.CertificateValidation.Exec.Arguments", ConfigurationEntryScope.ServerWideOnly)]
+        public string CertificateValidationExecArguments { get; set; }
+
+        [Description("The number of seconds to wait for the Certificate Validation executable to exit. Default: 30 seconds")]
```
ayende: 30 seconds is a _lot_ here, no? We have a pending client connection.
Would it make sense to use 5 sec and diverge from our normal?

src/Raven.Server/RavenServer.cs
```diff
@@ -330,6 +339,116 @@ void ConfigureKestrel(KestrelServerOptions options)
...
+            try
+            {
+                var timeout = Configuration.Security.CertificateValidationExecTimeout.AsTimeSpan;
+
+                var userArgs = Configuration.Security.CertificateValidationExecArguments ?? string.Empty;
+                var senderHostname = RequestExecutor.ConvertSenderObjectToHostname(sender);
+                var base64Cert = Convert.ToBase64String(certificate.Export(X509ContentType.Cert));
+
+                var args = $"{userArgs} " +
+                           $"{CommandLineArgumentEscaper.EscapeSingleArg(senderHostname)} " +
```
ayende: Don't we need to escape it here? What if this is empty?

src/Raven.Server/RavenServer.cs
```diff
@@ -330,6 +339,116 @@ void ConfigureKestrel(KestrelServerOptions options)
...
+                {
+                    readStdOut.Wait(timeout);
+                    readErrors.Wait(timeout);
+                }
+                catch (Exception e)
+                {
+                    throw new InvalidOperationException($"Unable to read redirected stderr and stdout when executing '{Configuration.Security.CertificateValidationExec} {args}'", e);
+                }
+
+                // Can have exit code o (success) but still get errors. We log the errors anyway.
```
ayende:
```suggestion
                // Can have exit code 0 (success) but still get errors. We log the errors anyway.
```

src/Raven.Server/RavenServer.cs
```diff
@@ -330,6 +339,116 @@ void ConfigureKestrel(KestrelServerOptions options)
...
+                        $"Command or executable '{Configuration.Security.CertificateValidationExec} {args}' failed. Exit code: {process.ExitCode}{Environment.NewLine}Output: {GetStdOut()}{Environment.NewLine}Errors: {GetStdError()}{Environment.NewLine}");
+                }
+
+                string output = GetStdOut();
+                bool result;
+                try
+                {
+                    bool.TryParse(output, out result);
+                }
+                catch (Exception e)
```
ayende: This will NOT throw, so the error will never be raised.

src/Raven.Server/RavenServer.cs
```diff
@@ -330,6 +339,116 @@ void ConfigureKestrel(KestrelServerOptions options)
...
+                {
+                    throw new InvalidOperationException(
+                        $"Cannot parse to boolean the result of Command or executable '{Configuration.Security.CertificateValidationExec} {args}'. Exit code: {process.ExitCode}{Environment.NewLine}Output: {GetStdOut()}{Environment.NewLine}Errors: {GetStdError()}{Environment.NewLine}", e);
```
ayende: Is it safe to call `GetStdOUt` and `GetStdErr` multiple times?

ayende (changes requested):

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -142,6 +142,24 @@ public class SecurityConfiguration : ConfigurationCategory
         [ConfigurationEntry("Security.WellKnownIssuerHashes.Admin", ConfigurationEntryScope.ServerWideOnly)]
         public string[] WellKnownIssuerHashes { get; set; }
 
+        [Description("EXPERT: A command or executable to validate a server authentication request. " +
+                     "RavenDB will execute: command [user-arg-1] ... [user-arg-n] <sender-url> <base64-certificate> <errors>. " +
+                     "The executable will return a case-insensitive boolean string through the standard output (e.g. true, false) indicating whether to approve the connection.")]
+        [DefaultValue(null)]
+        [ConfigurationEntry("Security.CertificateValidation.Exec", ConfigurationEntryScope.ServerWideOnly)]
```
ppekrol: Security.CertificateValidation.Exec -> `Security.Certificate.Validation.Exec` (same convention for others)

src/Raven.Server/Https/ExternalCertificateValidator.cs
```diff
@@ -0,0 +1,288 @@
...
+            _server = server;
+            _logger = logger;
+        }
+
+        public void Initialize()
+        {
+            if (string.IsNullOrEmpty(_server.Configuration.Security.CertificateValidationExec))
+                return;
+
+            _externalCertificateValidationCallbackCache = new ConcurrentDictionary<Key, Task<CachedValue>>();
```
ppekrol, line 37: We should validate here that the file exists probably?

src/Raven.Server/Https/ExternalCertificateValidator.cs
```diff
@@ -0,0 +1,288 @@
...
+            if (!m.Success)
+                throw new ArgumentException("Malformed input.");
+
+            return from Capture capture in m.Groups["r"].Captures select capture.Value;
+        }
+
+        private long GetScriptLastWriteHash()
+        {
+            long hash = File.GetLastWriteTimeUtc(_server.Configuration.Security.CertificateValidationExec).Ticks;
+            var args = RegexSplit(_server.Configuration.Security.CertificateValidationExecArguments);
```
ppekrol: Do we need args here? You cannot set them more than once, during server startup?

src/Raven.Server/Https/ExternalCertificateValidator.cs
```diff
@@ -0,0 +1,288 @@
...
+
+                // not done yet? return the cached value
+                return cachedValue.Valid;
+            }
+        }
+
+        // https://stackoverflow.com/questions/14241479/how-to-split-a-space-delimited-list-of-paths-where-paths-can-include-spaces-in
+        private static readonly Regex re = new Regex(@"^([ ]*((?<r>[^ ""]+)|[""](?<r>[^""]*)[""]))*[ ]*$");
+
+        private static IEnumerable<string> RegexSplit(string input)
```
ppekrol: Why do we need it now?

src/Raven.Server/Https/ExternalCertificateValidator.cs
```diff
@@ -0,0 +1,288 @@
...
+            Task<CachedValue> task;
+            if (_externalCertificateValidationCallbackCache.TryGetValue(cacheKey, out var existingTask) == false)
+            {
+                task = new Task<CachedValue>(() => CheckExternalCertificateValidation(senderHostname, certificate, chain, sslPolicyErrors, log));
+                existingTask = _externalCertificateValidationCallbackCache.GetOrAdd(cacheKey, task);
+                if (existingTask == task)
+                {
+                    task.Start();
+
+                    if (_externalCertificateValidationCallbackCache.Count > 50)
```
ppekrol, line 158: Shouldn't we clear that not only when we do not have given key in cache?

src/Raven.Server/Https/ExternalCertificateValidator.cs
iftahbe: We need to iterate over all the arguments (GetScriptLastWriteHash) and find their last write time because we don't know which argument is the actual script file. We do this so that if the script changes, we invalidate the cache value.

The special split function is used to split all the args even if they have a space in the path.
For example:
"Security.Certificate.Validation.Exec.Arguments": "-File \"c:\\work\\folder with spaces\\exec.ps1\" \"another argument\"",

src/Raven.Server/Https/ExternalCertificateValidator.cs
iftahbe: It's a little tricky because the file can be the exec (if it's a .exe for example) or it can be any of the arguments, for example, if we run a script via Powershell or bash.

Instead, if the command won't succeed it will throw and log the error.

src/Raven.Server/Https/ExternalCertificateValidator.cs
iftahbe: No. When we don't have a key in the cache is when we add a new item to the cache which causes it to grow.

When a key exists we just use the cached value and return quickly

src/Raven.Server/Https/ExternalCertificateValidator.cs
iftahbe: Fixing

src/Raven.Server/Https/ExternalCertificateValidator.cs
ppekrol: I do not like it. We should either expose an endpoint to clear this cache or just let user either wait few minutes or restart the server. This is a very rare situation. We should not do any magic around that.

ayende: @ppekrol This is done so if you develop the script, we can give you a good experience of fixing issues as you test it.

ppekrol: I understand the cause, but I'm afraid that this regex could not bulletproof. Especially if you will take into account all the platforms and all shells. It would be better IMO to add an endpoint to clear the cache.

src/Raven.Server/Https/ExternalCertificateValidator.cs
```diff
@@ -0,0 +1,290 @@
...
+            RequestExecutor.RemoteCertificateValidationCallback += (sender, cert, chain, errors) => ExternalCertificateValidationCallback(sender, cert, chain, errors, _logger);
+        }
+
+        private CachedValue CheckExternalCertificateValidation(string senderHostname, X509Certificate certificate, X509Chain chain, SslPolicyErrors sslPolicyErrors, Logger log)
+        {
+            var base64Cert = Convert.ToBase64String(certificate.Export(X509ContentType.Cert));
+
+            var timeout = _server.Configuration.Security.CertificateValidationExecTimeout.AsTimeSpan;
+
+            var args = $"{CommandLineArgumentEscaper.EscapeAndConcatenate(_userArgs)} " +
```
ppekrol: Use original arguments not splitted ones?

src/Raven.Server/Https/ExternalCertificateValidator.cs
```diff
@@ -0,0 +1,290 @@
...
+            return from Capture capture in m.Groups["r"].Captures select capture.Value;
+        }
+
+        private long GetScriptLastWriteHash()
+        {
+            long hash = File.GetLastWriteTimeUtc(_server.Configuration.Security.CertificateValidationExec).Ticks;
+
+            foreach (string arg in _userArgs)
+            {
+                hash = Hashing.Combine(hash, File.GetLastWriteTimeUtc(arg).Ticks);
```
ppekrol: Can File.GetLastWriteTimeUtc throw? Especially when this is not a file?

iftahbe: If it's not a file or the file doesn't exist it returns a date 1.1.1601 and that is hashed.

ppekrol: I'm concerned about the exceptions listed here: https://docs.microsoft.com/en-us/dotnet/api/system.io.file.getlastwritetimeutc?view=netcore-3.1#exceptions

src/Raven.Server/Https/ExternalCertificateValidator.cs
ayende: I would say that it would be better to add a method and be able to clear it from the JS Console, instead.

src/Raven.Server/Https/ExternalCertificateValidator.cs
ayende: Hm... agree. We need to catch exception there, then

src/Raven.Server/Https/ExternalCertificateValidator.cs
```diff
@@ -0,0 +1,262 @@
...
+                return;
+            
+            _externalCertificateValidationCallbackCache = new ConcurrentDictionary<Key, Task<CachedValue>>();
+
+            RequestExecutor.RemoteCertificateValidationCallback += (sender, cert, chain, errors) => ExternalCertificateValidationCallback(sender, cert, chain, errors, _logger);
+        }
+
+        public void ClearCache()
+        {
+            _externalCertificateValidationCallbackCache = new ConcurrentDictionary<Key, Task<CachedValue>>();
```
ppekrol: `_externalCertificateValidationCallbackCache?.Clear()` wouldn't be better?

iftahbe: yes

ppekrol (approved):

PR #10076, opened 2020-01-19
RavenDB-14506 Handle reference - Optimization for reducing multiple necessary mapping + test

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-14506

### DIFF
--- a/src/Raven.Server/Documents/Indexes/IndexStorage.cs
+++ b/src/Raven.Server/Documents/Indexes/IndexStorage.cs
@@ -546,42 +546,6 @@ namespace Raven.Server.Documents.Indexes
                     }
                 }
             }
-
-            if (indexingScope.ReferenceEtagsByCollection != null)
-            {
-                foreach (var kvp in indexingScope.ReferenceEtagsByCollection)
-                {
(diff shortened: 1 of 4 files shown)


grisha-kotler (approved):

src/Raven.Server/Documents/Indexes/IndexStorage.cs
```diff
@@ -547,41 +547,42 @@ public unsafe void WriteReferences(CurrentIndexingScope indexingScope, RavenTran
                 }
             }
 
-            if (indexingScope.ReferenceEtagsByCollection != null)
-            {
-                foreach (var kvp in indexingScope.ReferenceEtagsByCollection)
-                {
-                    var lastIndexedEtag = ReadLastIndexedEtag(tx, kvp.Key);
```
ppekrol, line 554: Shouldn't we read here following:
```
var lastEtagInCollection = ...
var isLastEtagInCollection = lastIndexedEtag  == lastEtagInCollection;
```

and in `if (etag > lastIndexedEtag)` change to:

```
if (etag > lastIndexedEtag && isLastEtagInCollection == false)
```

?

PR #10077, opened 2020-01-19
RavenDB-13296 Expose task count on Manage Ongoing tasks view

Danielle9897:


src/Raven.Studio/typescript/viewmodels/database/tasks/ongoingTasks.ts
```diff
@@ -73,6 +75,20 @@ class ongoingTasks extends viewModelBase {
     private initObservables() {
         this.myNodeTag(this.clusterManager.localNodeTag());
         this.serverWideBackupUrl = appUrl.forServerWideBackupList();
+        
+        this.taskNameToCount = ko.pureComputed<dictionary<number>>(() => {
+            var tasksNameToCount: { [id: string] : number; } = {};
```
ml054: never use var in typescript. Const or let.

src/Raven.Studio/typescript/viewmodels/database/tasks/ongoingTasks.ts
```diff
@@ -73,6 +75,20 @@ class ongoingTasks extends viewModelBase {
...
+          
+            tasksNameToCount["External Replication"] = this.replicationTasks().length;
```
ml054: and return this as object w/o assigning props line by line. I know you spent last week in java, but hopefully this is TS and we have nice syntax for doing this. ;-)

ml054: so it goes like this:

return {
   "External replication" : this.replicationTasks().length,
....
}

src/Raven.Studio/typescript/viewmodels/database/tasks/ongoingTasks.ts
Danielle9897: right - can't believe I overlooked that  :)

src/Raven.Studio/typescript/viewmodels/database/tasks/ongoingTasks.ts
Danielle9897: done - thanks

src/Raven.Studio/typescript/viewmodels/database/tasks/ongoingTasks.ts
```diff
@@ -369,7 +374,8 @@ class ongoingTasks extends viewModelBase {
                     default:
                         return taskType;
                 }
-            }));
+            })
+            .sort((a, b) => { return this.tasksOrderForUI[a] - this.tasksOrderForUI[b] }));
```
ml054: does this sort stable? https://www.geeksforgeeks.org/stability-in-sorting-algorithms/

ml054 (approved):

PR #10078, opened 2020-01-19
Fix for the issue of mapping too much in 32-bit PAL pager (NOT for merge, just review)

myarichuk:
* There is still an issue - when running SlowTests, test process sometimes randomly crashes and a couple of tests fail
* The relevant commit for review is https://github.com/ravendb/ravendb/commit/42179f9747ee460d9a463acbe98342d45fc9054f
* before doing a final commit I will do a proper merge with 4.2, for now I haven't done it


src/Raven.Pal/inc/rvn.h
```diff
@@ -100,6 +103,18 @@ rvn_test_storage_durability(const char *temp_file_name, int32_t *detailed_error_
 EXPORT int32_t
 rvn_get_path_disk_space(const char * path, uint64_t* total_free_bytes, uint64_t* total_size_bytes, int32_t* detailed_error_code);
 
+EXPORT int32_t
+rvn_remap(void *base_address, void **new_address, void *handle, int64_t size, int32_t flags, int64_t offset, int32_t *detailed_error_code);
+
+EXPORT int32_t
+rvn_mmap_file(int64_t sz, int64_t flags, void *handle, int64_t offset, void **addr, int32_t *detailed_error_code);
```
ayende, line 110:
```suggestion
rvn_mmap_file(int64_t size, int64_t flags, void *handle, int64_t offset, void **address, int32_t *detailed_error_code);
```

ayende: I don't like the order of the parameters, in these declarations
It should be:
* Handle - what we operate on
* size/offset - together, they are related
* output variables - the result of the call
* detailed error code

ayende: Also on `rvn_allocate_more_space`

src/Raven.Pal/src/posix/mapping.c
```diff
@@ -77,21 +86,25 @@ rvn_create_and_mmap64_file(
     if (mfh->path == NULL)
     {
         rc = FAIL_CALLOC;
-        goto error_cleanup;
+        goto error_clean_with_error;
     }
     mfh->flags = flags;
-    mfh->fd = open(path, O_RDWR | O_CREAT, S_IWUSR | S_IRUSR);
+    int32_t largefile = 0;
+     if (sizeof(int) == 4) /* 32 bits */
```
ayende: This isn't correct. `sizeof(int)` is _always_ 4 bytes on Linux and Windows.
You need to check `sizeof(void*)` if you want point size.

src/Raven.Pal/src/posix/mapping.c
```diff
@@ -100,14 +113,14 @@ rvn_create_and_mmap64_file(
     if (sys_page_size == -1)
     {
         rc = FAIL_SYSCONF;
-        goto error_cleanup;
+        goto error_clean_with_error;
     }
 
     int32_t allocation_granularity = ALLOCATION_GRANULARITY;
```
ayende: Why not use the constant?

src/Raven.Pal/src/posix/mapping.c
```diff
@@ -267,3 +277,44 @@ rvn_discard_virtual_memory(void* address, int64_t size, int32_t* detailed_error_
         *detailed_error_code = errno;
     return rc;
 }
+
+EXPORT int32_t
+rvn_remap(void *base_address, void **new_address, void *handle, int64_t size, int32_t flags, int64_t offset, int32_t *detailed_error_code)
+{
+    int32_t rc = rvn_unmap(MMOPTIONS_NONE, base_address, ALLOCATION_GRANULARITY, detailed_error_code);
```
ayende, line 286: This is *not* safe to do.
When we remap, we create a _new_ mapping. The old one is still there and valid.
We will be disposing that _separately_. If you don't handle this, you'll have crashes when read transactions that hold a reference to the old mapping are trying to read it.

src/Raven.Pal/src/posix/mapping.c
```diff
@@ -180,25 +194,21 @@ rvn_allocate_more_space(
...
-    else
-        mmap_flags |= MAP_SHARED;
-
-    void *address = rvn_mmap(NULL, new_length_after_adjustment, PROT_READ | PROT_WRITE, mmap_flags, mfh->fd, 0L);
-    if (address == MAP_FAILED)
+    if (map_after_allocation_flag != 0)
     {
-        rc = FAIL_MMAP64;
-        goto error_cleanup;
+        int32_t mmap_flags = 0;
```
ayende, line 198: What is this for?

src/Voron/Impl/Paging/Rvn32BitsMemoryMapPager.cs
```diff
@@ -3,83 +3,87 @@
...
-
-            _totalAllocationSize = GetFileSize();
 
-            if (_totalAllocationSize == 0 && initialFileSize.HasValue)
-            {
-                _totalAllocationSize = NearestSizeToAllocationGranularity(initialFileSize.Value);
-            }
-            if (_totalAllocationSize == 0 || _totalAllocationSize % AllocationGranularity != 0 ||
-                _totalAllocationSize != GetFileSize())
+//            _totalAllocationSize = GetFileSize();
```
ayende: Nope

ayende (changes requested):

src/Raven.Pal/src/posix/mapping.c
myarichuk: Done.

src/Raven.Pal/src/posix/mapping.c
myarichuk: Done.

PR #10080, opened 2020-01-20
RavenDB-13053 fixing race condition when adding or getting ShardedCon…

talweiss1982:
…text

### DIFF
--- a/src/Raven.Server/Documents/DatabasesLandlord.cs
+++ b/src/Raven.Server/Documents/DatabasesLandlord.cs
@@ -569,7 +569,7 @@ namespace Raven.Server.Documents
                 if (databaseRecord.Shards?.Length > 0)
                 {
                     var shardedContext = new ShardedContext(_serverStore, databaseRecord);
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -569,7 +569,7 @@ public DatabaseSearchResult TryGetOrCreateDatabase(StringSegment databaseName)
                 if (databaseRecord.Shards?.Length > 0)
                 {
                     var shardedContext = new ShardedContext(_serverStore, databaseRecord);
-                    _shardedDatabases.GetOrAdd(databaseName, Task.FromResult(shardedContext));
+                    shardedContext = _shardedDatabases.GetOrAdd(databaseName, Task.FromResult(shardedContext)).Result;
```
ayende, line 572: Is is possible that the value here would be an async task?

talweiss1982: No since the only way we add a result to this dictionary is using Task.FromResult

PR #10086, opened 2020-01-21
RavenDB-13871 Copy Document ID

Danielle9897:


src/Raven.Studio/wwwroot/Content/css/pages/documents-edit.less
```diff
@@ -218,6 +218,22 @@
             max-width: 100%;
         }
     }
+
+    #copy-document-id-icon {
+        visibility: hidden;
+
+        &:hover {
```
ml054, line 226: please move this color outside hover.

src/Raven.Studio/typescript/viewmodels/database/documents/editDocument.ts
```diff
@@ -569,8 +569,9 @@ class editDocument extends viewModelBase {
         return collectionForNewDocument + "/";
     }
 
-    toClipboard() {
-        copyToClipboard.copy(this.documentText(), "Document has been copied to clipboard");
+    toClipboard(copyId: boolean = false) {
```
ml054: maybe we should split this into 2 methods:
copyDocumentIdToClipboard
copyDocumentBodyToClipboard?

src/Raven.Studio/wwwroot/App/views/database/documents/editDocument.html
```diff
@@ -20,7 +20,7 @@
                                 <!--<span data-bind="attr: { class: connectedDocuments.currentDocumentIsStarred()?'icon-star-filled favorite':'icon-star' }"></span>-->
                             <!--</a>-->
                             <span class="obj-name on-base-background" data-bind="text: userSpecifiedId, attr: { title: userSpecifiedId }"></span>
-                            
+                            <span id="copy-document-id-icon" title="Copy document Id" data-bind="click: _.partial(toClipboard, true)"><i class="icon-copy"></i></span>
```
ml054: let's avoid convey information about type , and rename to: copy-document-id

ml054: + probably we should rename from span to a to have proper cursor.

src/Raven.Studio/typescript/viewmodels/database/documents/editDocument.ts
Danielle9897: That was exactly my first thought - and then I changed to this because I thought you would prefer it united :)

So I will split

src/Raven.Studio/wwwroot/Content/css/pages/documents-edit.less
Danielle9897: leaving this - as discussed

src/Raven.Studio/typescript/viewmodels/database/documents/editDocument.ts
Danielle9897: done

src/Raven.Studio/wwwroot/App/views/database/documents/editDocument.html
Danielle9897: done

src/Raven.Studio/wwwroot/Content/css/pages/documents-edit.less
```diff
@@ -219,8 +219,9 @@
         }
     }
 
-    #copy-document-id-icon {
+    #copy-document-id {
         visibility: hidden;
+        color: @gray-light;
```
ml054, line 224: I think there is no purpose of having that ?

Danielle9897: This was added because when **'a'** is used instead of **'span**' then the default color is **red**

ml054: ok, thanks

ml054 (approved):

PR #10090, opened 2020-01-21
RavenDB-13053  WIP for review

talweiss1982:


src/Raven.Server/Documents/Handlers/ShardedDocumentHandler.cs
```diff
@@ -25,29 +35,128 @@ public async Task Get()
...
+                        //TODO: Add performance hints
+                    }
+                }
+            }
+            finally
+            {
+                Task.Clear();
+                Cmds.Clear();
+
+                Parallel.ForEach(CmdsContext, c =>
```
ayende: That is not a good idea, we want to return this to the same thread as we pull this from

src/Raven.Server/Documents/Handlers/ShardedDocumentHandler.cs
```diff
@@ -1,20 +1,30 @@
...
+using Sparrow.Json;
 
 namespace Raven.Server.Documents.Handlers
 {
     public class ShardedDocumentHandler : ShardedRequestHandler
     {
+        protected readonly StringBuilder UrlStringBuilder = new StringBuilder();
+        protected readonly List<Task> Task = new List<Task>();
+        protected readonly Dictionary<int, ShardedCommand> Cmds = new Dictionary<int, ShardedCommand>();
+        protected readonly List<TransactionOperationContext> CmdsContext = new List<TransactionOperationContext>();
```
ayende: Why do we want these to be instance fields?

src/Raven.Server/Documents/Handlers/ShardedDocumentHandler.cs
```diff
@@ -25,29 +35,128 @@ public async Task Get()
...
+                        var task = ShardedContext.RequestExecutors[shard.Key].ExecuteAsync(cmd, ctx);
+                        Task.Add(task);
+                        Cmds.Add(shard.Key, cmd);
+                    }
+
+                    await System.Threading.Tasks.Task.WhenAll(Task);
+
+                    HttpContext.Response.StatusCode = (int)ShardedStatusCodeChooser.GetStatusCode(Cmds.Values);
+
+                    if (HttpContext.Response.IsSuccessStatusCode() == false)
```
ayende: We need to check if we have errors from any of the other nodes. And if so, we _have_ to give that error to the caller.

src/Raven.Server/Documents/ComputeHttpEtags.cs
```diff
@@ -96,6 +96,39 @@ public static unsafe string ComputeEtagForRevisions(List<Document> revisions)
...
+
+            var size = Sodium.crypto_generichash_bytes();
+            Debug.Assert((int)size == 32);
+            var cryptoGenerichashStatebytes = (int)Sodium.crypto_generichash_statebytes();
+            byte* state = stackalloc byte[cryptoGenerichashStatebytes];
+            if (Sodium.crypto_generichash_init(state, null, UIntPtr.Zero, size) != 0)
+                ThrowFailToInitHash();
+
+            foreach (var etag in etags)
+            {
```
ayende, line 110: You are passing potential nulls here, that is likely going to kill the server

src/Raven.Server/Documents/Handlers/ShardedDocumentHandler.cs
```diff
@@ -25,29 +35,128 @@ public async Task Get()
...
-                cmd.Result?.WriteJsonTo(ResponseBodyStream());
+        private IEnumerable<string> EnumerateEtags(Dictionary<int, ShardedCommand> cmds)
+        {
+            foreach (var cmd in cmds.Values)
+            {
+                yield return cmd.Response?.Headers?.ETag?.Tag;
+            }
+        }
+
+        private int WriteArray(AsyncBlittableJsonTextWriter writer, string arrayName, IEnumerable<string> ids,
```
ayende: We should be able to simplify this

src/Raven.Server/Documents/Sharding/ShardedRequestHandler.cs
```diff
@@ -1,16 +1,16 @@
...
 
-namespace Raven.Server.Documents
+namespace Raven.Server.Documents.Sharding
 {
     public class ShardedRequestHandler : RequestHandler
     {
         public ShardedContext ShardedContext;
         public TransactionContextPool ContextPool;
 
+        protected static readonly ObjectPool<StringBuilder> StringBuilderObjectPool = new ObjectPool<StringBuilder>(()=> new StringBuilder());
```
ayende: unused

src/Raven.Server/Documents/Sharding/ShardedStatusCodeChooser.cs
```diff
@@ -0,0 +1,38 @@
...
+                    first = false;
+                }
+                else
+                {
+                    if (statusCode != command.StatusCode)
+                    {
+                        //If both statuses are successful we should try to return a successful status code
+                        if (statusCode.IsSuccessStatusCode() == false || command.StatusCode.IsSuccessStatusCode() == false)
+                        {
+                            return HttpStatusCode.Ambiguous;
```
ayende: That is going to cause us to hide errors from the user, which is something we want to avoid

src/Raven.Server/Documents/Handlers/ShardedDocumentHandler.cs
talweiss1982: The handlers are long-living aren't they?
I want to "object-pool " those since they are managed objects are going to be generated a lot

src/Raven.Server/Documents/Handlers/ShardedDocumentHandler.cs
talweiss1982: The first error?

talweiss1982: The same goes for success status, this is unclear what the behavior should be.

src/Raven.Server/Documents/Handlers/ShardedDocumentHandler.cs
ayende: They aren't long lived. They are for a single request.
And if they were, you would have thread safety issues

src/Raven.Server/Documents/ComputeHttpEtags.cs
talweiss1982: how is it different from what we do with documents:
https://github.com/ravendb/ravendb/blob/feature/sharding/src/Raven.Server/Documents/ComputeHttpEtags.cs#L103

src/Raven.Server/Documents/Handlers/ShardedDocumentHandler.cs
ayende: Gather all the errors you got.
If there is a single one, throw it.
If multiple, `AggregateException`

All the nodes should return the same status code

src/Raven.Server/Documents/ComputeHttpEtags.cs
ayende: There it is known you can't get null.
In you EnumerateEtags, you are expecting nulls

src/Raven.Server/Documents/Handlers/ShardedDocumentHandler.cs
talweiss1982: If we need to keep the load order how can we simplify this?
@ppekrol do we need to keep load order in LoadOperation results?
Breaking down commands and aggregate their results as blittables is going to be complex for sure and error-prone.

src/Raven.Server/Documents/Sharding/ShardedRequestHandler.cs
talweiss1982: yes I thought of using object pools for the managed objects I need but decided that a private member will be better

src/Raven.Server/Documents/Handlers/ShardedDocumentHandler.cs
ayende: Got some ideas on this, coming over

PR #10092, opened 2020-01-21
RavenDB-14461

ppekrol:
- NoTracking session should behave same as regular one when loading empty array
- removed redundant operation when loading by array of ids

### DIFF
--- a/src/Raven.Client/Documents/Session/Operations/LoadOperation.cs
+++ b/src/Raven.Client/Documents/Session/Operations/LoadOperation.cs
@@ -2,7 +2,6 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using Raven.Client.Documents.Commands;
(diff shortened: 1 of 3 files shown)


src/Raven.Client/Documents/Session/Operations/LoadOperation.cs
```diff
@@ -75,12 +74,10 @@ public LoadOperation WithAllCounters()
...
         public LoadOperation ByIds(IEnumerable<string> ids)
         {
-            _ids = ids.Where(x => x != null).ToArray();
-
-            foreach (var id in _ids.Distinct(StringComparer.OrdinalIgnoreCase))
-            {
-                ById(id);
-            }
+            _ids = ids
+                .Where(x => x != null)
```
ayende: Also check empty?

PR #10093, opened 2020-01-22
RavenDB-14502 Output which fields are time series fields, so Studio c…

ayende:
…an show them nicely

### DIFF
--- a/src/Raven.Client/Documents/Queries/QueryResultBase.cs
+++ b/src/Raven.Client/Documents/Queries/QueryResultBase.cs
@@ -77,6 +77,12 @@ namespace Raven.Client.Documents.Queries
         public string NodeTag { get; set; }
 
         /// <summary>
+        /// If the query returned time series results, this field will contain
+        /// the names of the fields with time series data
+        /// </summary>
+        public List<string> TimeSeriesFields { get; set; }
+
+        /// <summary>
         /// Detailed timings for various parts of a query (Lucene search, loading documents, transforming results) - if requested.
         /// </summary>
         public QueryTimings Timings { get; set; }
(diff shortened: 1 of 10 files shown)


src/Raven.Client/Documents/Queries/QueryResultBase.cs
```diff
@@ -76,6 +76,12 @@ public abstract class QueryResultBase<TResult, TInclude>
         /// </summary>
         public string NodeTag { get; set; }
 
+        /// <summary>
+        /// If the query returned time series results, this field will contain
+        /// the names of the fields with time series data
+        /// </summary>
+        public List<string> TimeSeriesFields { get; set; }
```
ppekrol: This is metadata needed by the Studio. It should not be part of Client API IMO. We probably should move this to QueryResultServerSide?

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -2688,6 +2688,7 @@ public IndexDefinition GetIndexDefinition()
 
                                 if (includeTimeSeriesCommand != null)
                                     resultToFill.AddTimeSeriesIncludes(includeTimeSeriesCommand);
+                                resultToFill.RegisterTimeSeriesFields(fieldsToFetch);
```
ppekrol: Fill it only when query has 'fromStudio' ?

src/Raven.Client/Documents/Queries/QueryResultBase.cs
ppekrol: Or even to DocumentQueryResult?

PR #10094, opened 2020-01-22
RavenDB-14484 database should not be placed as preferred node if it i…

karmeli87:
…s still loading

### DIFF
--- a/src/Raven.Client/ServerWide/DatabaseTopology.cs
+++ b/src/Raven.Client/ServerWide/DatabaseTopology.cs
@@ -192,6 +192,34 @@ $"NodeTag of 'InternalReplication' can't be modified after 'GetHashCode' was inv
         }
 
 
+        internal bool TryUpdateByPriorityOrder()
+        {
+            if (IsReorderNeeded() == false)
+                return false;
+
+            var originalOrder = new List<string>(Members);
+            ReorderMembers();
+
+            if (originalOrder.SequenceEqual(Members))
+                return false; // members hasn't changed after the reorder
+
+            return true;
+        }
+        private bool IsReorderNeeded()
+        {
+            if (PriorityOrder == null || PriorityOrder.Count == 0)
+                return false;
+
+            for (var index = 0; index < Math.Min(Members.Count, PriorityOrder.Count); index++)
+            {
+                var member = Members[index];
+                if (PriorityOrder[index] == member)
+                    return true;
+            }
+
+            return false;
+        }
+
         public bool RelevantFor(string nodeTag)
         {
             return Members.Contains(nodeTag) ||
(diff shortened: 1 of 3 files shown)


test/RachisTests/DatabaseCluster/ClusterDatabaseMaintenance.cs
```diff
@@ -273,6 +276,72 @@ public async Task ReshuffleAfterPromotion()
             }
         }
 
+        [Fact]
+        public async Task MoveLoadingNodeToLast()
+        {
+            DebuggerAttachedTimeout.DisableLongTimespan = true;
```
ppekrol: should be removed?

src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
```diff
@@ -652,6 +665,10 @@ private string UpdateDatabaseTopology(DatabaseObservationState state, ref List<D
                 RaiseNoLivingNodesAlert($"None of '{dbName}' database nodes are responding to the supervisor, the database is unreachable.", dbName);
             }
 
+            if (someNodesRequireMoreTime == false &&
+                databaseTopology.TryUpdateByPriorityOrder())
+                return "Reorder the member nodes to ensure the priority order.";
```
ppekrol: Reordering? Reordered?

karmeli87: done

test/RachisTests/DatabaseCluster/ClusterDatabaseMaintenance.cs
karmeli87: done

ppekrol (approved):

src/Raven.Client/ServerWide/DatabaseTopology.cs
```diff
@@ -192,6 +192,36 @@ internal void ReorderMembers()
...
+
+            ReorderMembers();
+            if (IsReorderNeeded() == false)
+                return true;
+
+            // we fail to ensure the order, so let's revert to the original and continue.
+            Members = originalOrder;
+            return false;
+        }
+        private bool IsReorderNeeded()
```
ayende, line 208: I don't understand the logic here. I think that you are checking to see if the members list match the priority order, but don't you need the inverse check in the for ?

karmeli87: Yes, I check if reordering of the members is needed.
It might happened that we shift the preferred node to be the last until he is loaded, and when he does we need to shift it back.
What do you mean by inverse check?

ayende (approved):

PR #10099, opened 2020-01-23
RavenDB-14002 move to long

ppekrol:


src/Voron/Data/BTrees/TreePageIterator.cs
```diff
@@ -134,9 +134,12 @@ public bool MovePrev()
 
         }
 
-        public bool Skip(int count)
+        public bool Skip(long count)
         {
-            _page.LastSearchPosition += count;
+            if (count > int.MaxValue)
+                throw new InvalidOperationException($"Cannot skip by '{count}' because it is more than int.MaxValue");
```
arekpalinski: Maybe move exception throwing to a dedicated method?

src/Voron/Data/Fixed/FixedSizeIterators.cs
```diff
@@ -113,7 +113,7 @@ public long CurrentKey
                 {
                     if (_pos >= _header->NumberOfEntries)
                         throw new InvalidOperationException("Invalid position, cannot read past end of tree");
-                    return FixedSizeTreePage.GetEntry(_dataStart, _pos, _fst._entrySize)->Key;
+                    return FixedSizeTreePage.GetEntry(_dataStart, (int)_pos, _fst._entrySize)->Key;
```
arekpalinski, line 116: Shouldn't we check if `_pos > int.MaxValue` and throw then? Just in case - same as you did above

ppekrol: We have a `if (_pos >= _header->NumberOfEntries)` check. Since the `NumberOfEntries` is ushort then I assume that this is not needed.

ayende (approved):

PR #10104, opened 2020-01-26
RavenDB-14236: Flushing occurs only when have something to write, so …

aviviadi:
…force flush if High Dirty Mem happens as no actual write in commit will occur

### DIFF
--- a/src/Raven.Server/Documents/TransactionOperationsMerger.cs
+++ b/src/Raven.Server/Documents/TransactionOperationsMerger.cs
@@ -22,6 +22,7 @@ using Sparrow.Platform;
 using Sparrow.Server.Meters;
 using Sparrow.Server.Utils;
 using Sparrow.Utils;
+using Voron;
 using Voron.Debugging;
 using Voron.Global;
 using Voron.Impl;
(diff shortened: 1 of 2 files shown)


ppekrol (approved):

src/Raven.Server/Documents/TransactionOperationsMerger.cs
```diff
@@ -849,6 +850,7 @@ private enum PendingOperations
                                 rejectedBuffer.Add(operationToReject);
                             }
                             NotifyOnThreadPool(rejectedBuffer);
+                            GlobalFlushingBehavior.GlobalFlusher.Value?.MaybeFlushEnvironment(context.Environment); // commit stage 2 won't trigger so we need to ask for flush here
```
arekpalinski, line 853: I wonder if we shouldn't wait a bit after we schedule the flush / sync. Maybe same behavior as we have for indexes:

https://github.com/ravendb/ravendb/blob/3ec9c0e4cd21c945b48e0ba58c7b204233fa518e/src/Raven.Server/Documents/Indexes/Index.cs#L1252-L1258

?

grisha-kotler: This flushes only the current environment while the high dirty memory check is for all environments

PR #10108, opened 2020-01-26
fix for rvn32 pager segfault on Linux (RavenDB-12877) - for code review only

myarichuk:
* Proper nearest size calculation in PAL.
* Make sure to dispose the PAL handle properly
* Fix alignment issue at p/invoke of rvn_mmap_file()
* Make sure "Validate" compilation mode works
* Some precautions and some 'printf' debugging (conditional)

Commit of interest: 33440eba8504e3db37d768d41c51f120c3c4493b 

Note: still need to run tests and validate that nothing else is broken. RavenDB server seems to properly run on arm32, so I don't expect many issues (if any).


aviviadi:
I've tested the above code, it doesn't flush scratches (which are all created at size of 32MB and not resizing more then that).  
Also on idle after importing 5M Docs we have 1G mapped area in the pager.  We should have 0.

myarichuk:
> 
> 
> I've tested the above code, it doesn't flush scratches (which are all created at size of 32MB and not resizing more then that).
> Also on idle after importing 5M Docs we have 1G mapped area in the pager. We should have 0.

Working on it now.

PR #10112, opened 2020-01-27
RavenDB-14542 Support quoting for complex collection names

ppekrol:


ayende (approved):

src/Raven.Client/Documents/Session/Tokens/FromToken.cs
```diff
@@ -63,9 +55,34 @@ public override void WriteTo(StringBuilder writer)
             }
         }
 
-        private void ThrowInvalidCollectionName()
+        private static bool RequiresQuotes(string collectionName, out string escapedCollectionName)
```
montyclt, line 58: Why a method that check if collection name requires quotes instead of wrapping it with quotes ever? Reduce complexity and avoid possible errors in the algorithm that check if it require quotes.

ppekrol: https://github.com/ravendb/ravendb/pull/10119

@montyclt I've opened a PR that addresses your comment. We were a bit defensive with the approach, but did additional testing and there is no impact of the performance of the RQL parser on the server.

PR #10116, opened 2020-01-28
RavenDB-14236 - Handle I/O starvation

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-14236
https://issues.hibernatingrhinos.com/issue/RavenDB-14485

### DIFF
--- a/src/Raven.Server/Documents/TransactionOperationsMerger.cs
+++ b/src/Raven.Server/Documents/TransactionOperationsMerger.cs
@@ -60,8 +60,6 @@ namespace Raven.Server.Documents
             _maxTimeToWaitForPreviousTxInMs = _parent.Configuration.TransactionMergerConfiguration.MaxTimeToWaitForPreviousTx.AsTimeSpan.TotalMilliseconds;
             _maxTxSizeInBytes = _parent.Configuration.TransactionMergerConfiguration.MaxTxSize.GetValue(SizeUnit.Bytes);
             _maxTimeToWaitForPreviousTxBeforeRejectingInMs = _parent.Configuration.TransactionMergerConfiguration.MaxTimeToWaitForPreviousTxBeforeRejecting.AsTimeSpan.TotalMilliseconds;
-            _timeToCheckHighDirtyMemory = _parent.Configuration.Memory.TemporaryDirtyMemoryChecksPeriod;
-            _lastHighDirtyMemCheck = _parent.Time.GetUtcNow();
         }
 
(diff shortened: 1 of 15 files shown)


src/Raven.Server/Documents/TransactionOperationsMerger.cs
```diff
@@ -580,6 +593,29 @@ private void MergeTransactionsOnce()
             }
         }
 
+        private void NotifyHighDirtyMemoryFailure(List<MergedTransactionCommand> pendingOps, string details, long transactionId)
+        {
+            var highDirtyMemory = new HighDirtyMemoryException(
```
ayende: The exception is already there in the caller, why not just pass it in?

grisha-kotler: done

ppekrol (approved):

src/Raven.Server/Documents/TransactionOperationsMerger.cs
```diff
@@ -827,34 +874,23 @@ private enum PendingOperations
...
-                            NotifyOnThreadPool(rejectedBuffer);
-                            GlobalFlushingBehavior.GlobalFlusher.Value?.MaybeFlushEnvironment(context.Environment); // commit stage 2 won't trigger so we need to ask for flush here
-                            break;
-                        }
-                        _lastHighDirtyMemCheck = now; // reset timer for next check only if no errors (otherwise check every single write until back to normal)
+                        // we need to ask for a flush here
+                        GlobalFlushingBehavior.GlobalFlusher.Value?.MaybeFlushEnvironment(context.Environment);
+                        _lastHighDirtyMemCheck = now;
                     }
+
```
aviviadi, line 887: Suggestion: Maybe not throw immediately here, because we request a flush and it might be fast enough. So we can wait longer and just then throw.

karmeli87 (approved):
LGTM

src/Raven.Server/Documents/TransactionOperationsMerger.cs
ayende: Flush isn't actually going to release any memory, the OS still need to do some work around that.
Would rather trigger the flush and wait for the _next_ operation

ayende (approved):

src/Raven.Server/Documents/TransactionOperationsMerger.cs
aviviadi: By "wait longer" I meant: raise flag and if IsHighDirty is still on after X seconds, start rejecting.
And it makes more sense to start rejecting after period of time and not after "operation" as it is disk write speeds issue.

aviviadi (approved):

PR #10121, opened 2020-01-30
RavenDB-12605 Check for duplicate fields when saving index

Danielle9897:


src/Raven.Studio/typescript/models/database/index/indexDefinition.ts
```diff
@@ -88,6 +91,16 @@ class indexDefinition {
         this.configuration(this.parseConfiguration(dto.Configuration));
 
         this.additionalSources(_.map(dto.AdditionalSources, (code, name) => additionalSource.create(name, code)));
+
+        this.hasDuplicateFieldsNames = ko.pureComputed(() => {
+            let hasDuplicates = false;
+
+            if (_.uniqBy(this.fields(), field => field.name()).length !== this.fields().length) {
```
ml054: we can simply return this value, no need to have if, let, etc.

Danielle9897: ok - thanks - done

ml054 (approved):

PR #10124, opened 2020-01-31
RavenDB-13765: Fix attachments recovery

aviviadi:


src/Raven.Server/Smuggler/Documents/StreamSource.cs
```diff
@@ -1092,15 +1095,21 @@ private IEnumerable<DocumentItem> ReadDocuments(INewDocumentActions actions = nu
...
+
                         ProcessAttachmentStream(context, data, ref attachment);
-                        attachments.Add(attachment);
+
+                        if (TryMatchAttachmentToDocumentOrStoreForLaterMatching(
+                                storeDocsForLaterMatchWithAttachments,
+                                storeAttachmentsForLaterMatchWithDocs,
+                                attachment) == false)
+                        {
+                            // TODO: CreateOrphanAttachment(attachment, "Failed to decode attachment's tag");
```
aviviadi, line 1110: Should we create or fail?

src/Raven.Server/Smuggler/Documents/StreamSource.cs
```diff
@@ -1125,20 +1134,58 @@ private IEnumerable<DocumentItem> ReadDocuments(INewDocumentActions actions = nu
...
+                        true);
+
+                    Debug.Assert(successInIdExtraction, "Already executed once so can't fail here");
+                }
+
+                foreach (var docItem in storeDocsForLaterMatchWithAttachments)
+                {
+                    if (docItem.Value.Attachments == null)
+                    {
+                        // TODO: what to do with docs reported to have attachments but attachments weren't found?
```
aviviadi, line 1186: What should we do here? We can reset the doc's "has attachment" flag but it feels wrong

src/Raven.Server/Smuggler/Documents/StreamSource.cs
```diff
@@ -1148,6 +1195,45 @@ private IEnumerable<DocumentItem> ReadDocuments(INewDocumentActions actions = nu
...
+                        }
+                        else
+                        {
+                            if (reportOrphan == false)
+                            {
+                                storeAttachmentsForLaterMatchWithDocs.Add(relatedDocId, attachment);
+                            }
+                            else
+                            {
+                                // TODO: CreateOrphanAttachment(attachment, $"Attachment's tag contain not exists document {relatedDocId}");
```
aviviadi, line 1226: Should we create or fail?

src/Raven.Server/Smuggler/Documents/StreamSource.cs
```diff
@@ -1148,6 +1195,45 @@ private IEnumerable<DocumentItem> ReadDocuments(INewDocumentActions actions = nu
             }
         }
 
+        private bool TryMatchAttachmentToDocumentOrStoreForLaterMatching(
+            Dictionary<string, DocumentItem> storeDocsForLaterMatchWithAttachments,
+            Dictionary<string, DocumentItem.AttachmentStream> storeAttachmentsForLaterMatchWithDocs,
+            DocumentItem.AttachmentStream attachment,
+            bool reportOrphan = false)
+        {
+            if (attachment.Data.TryGetMember("Tag", out var tag))
```
ayende, line 1204: You cannot use this for anything. This is information only, debug, not operational

ayende (commented):
What we need to do is:
* when we run into an attachment, save it immediately and remember it in `OrphanAttachments`
* when we run into a document that has attachments that we haven't seen, record it in `MissingAttachments`
* when we run into a document with attachments, remove them from the `OrphanAttachments` list.
* when we find an attachment, check the `MissingAttachments` list and add it to the previously seen docs.

aviviadi:
Does document has a record of pointed attachments ? (hash / other id)

aviviadi:
Decision: To change the Recovery process from 'exporting recovered database into a smuggler dump file' to a 'new recovered database' where instead of importing the recovered data, we actually use the db api and create a complete new db (named in example: MyDB_Recovered_Date). It really simplifies recovery process and make it a lot more accurate and not protocols bounded.

PR #10125, opened 2020-01-31
RavenDB-14555 - Removing thread local usages

ayende:


karmeli87:
I think we also want to remove the usage of `ThreadLocal` in the client.
Two can be found here:
https://github.com/ravendb/ravendb/blob/af3b7efe4e16d961ed146b386da0569867da9c12/src/Raven.Client/Documents/Conventions/JsonNetBlittableEntitySerializer.cs#L13

karmeli87 (approved):

src/Sparrow/Utils/LightThreadLocal.cs
```diff
@@ -0,0 +1,97 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+
+namespace Sparrow.Utils
```
ppekrol: can we move it to Sparrow.Threading? Looks more appropriate?

src/Voron/Util/ThreadHoppingReaderWriterLock.cs
```diff
@@ -36,7 +32,7 @@ public void EnterWriteLock()
             var currentWaiters = Interlocked.Add(ref _waiters, WriterMarker);
 
             // try take ownership on lock
-            var currentLock = Interlocked.CompareExchange(ref _writeLockOwnerThreadId, _localThreadIdCache.Value, 0);
+            var currentLock = Interlocked.CompareExchange(ref _writeLockOwnerThreadId, Thread.CurrentThread.ManagedThreadId, 0);
```
ppekrol: Wasn't the cache crated because accessing 'Thread.CurrentThread.ManagedThreadId' is expensive? Maybe it is better to access 'NativeMemory.CurrentThreadStats.ManagedThreadId' instead?

Anyway, at least we can create a variable in this method to avoid accessing it twice.

ayende:
> I think we also want to remove the usage of `ThreadLocal` in the client.
> Two can be found here:
> https://github.com/ravendb/ravendb/blob/af3b7efe4e16d961ed146b386da0569867da9c12/src/Raven.Client/Documents/Conventions/JsonNetBlittableEntitySerializer.cs#L13

Those are probably going to be only twice per DocumentStore, which means that they are likely _very_ small. 
I want to first see that everything is fine with the actual impl

src/Voron/Util/ThreadHoppingReaderWriterLock.cs
ayende: It used to be expensive, but now, I don't think so.
See:

https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Threading/Thread.cs#L19

https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Threading/Thread.cs#L156

https://github.com/dotnet/runtime/blob/master/src/coreclr/src/System.Private.CoreLib/src/System/Threading/Thread.CoreCLR.cs#L161

https://github.com/dotnet/runtime/blob/fcd862e06413a000f9cafa9d2f359226c60b9b42/src/coreclr/src/vm/object.h#L1426

So this ends up being as fast as our own impl

ppekrol:
Rebased and used newest Lucene

ppekrol (approved):

PR #10127, opened 2020-01-31
RavenDB-14558 set PooledConnectionLifetime in SocketsHttpHandler to 10 minutes

ppekrol:


src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1732,6 +1733,13 @@ public static HttpClientHandler CreateHttpMessageHandler(X509Certificate2 certif
                 ValidateClientKeyUsages(certificate);
             }
 
+#if NETCOREAPP
+            if (_getSocketsHttpHandler?.GetValue(httpMessageHandler) is SocketsHttpHandler socketsHttpHandler && socketsHttpHandler != null)
+            {
+                socketsHttpHandler.PooledConnectionLifetime = TimeSpan.FromMinutes(10);
```
ppekrol, line 1739: Should this be configurable via Conventions (NETCOREAPP only)? What should be the default value?

karmeli87:
If I upload/download a backup that takes more than 10 min, the connection will be closed?

ppekrol:
If I understood this https://github.com/dotnet/runtime/issues/18348#issuecomment-415845645 correctly then it should not happen.

PR #10129, opened 2020-02-01
RavenDB-13456 ability to change identity parts separator on the server-side

ppekrol:


src/Raven.Server/Utils/IncludeUtil.cs
```diff
@@ -183,7 +183,7 @@ private static string HandlePrefixValue(object val, StringSegment prefixSegment)
         private static string ValueWithPrefix(StringSegment prefixSegment, object val)
         {
             var prefix = prefixSegment.Subsegment(0, prefixSegment.Length - 1);
-            return (prefix.Length > 0) && (prefix[prefix.Length - 1] != '/') ? null : $"{prefix}{val}";
+            return (prefix.Length > 0) && (prefix[prefix.Length - 1] != '/') ? null : $"{prefix}{val}"; // TODO [ppekrol]
```
ppekrol: Still need to look into this

src/Raven.Server/Documents/Handlers/HiLoHandler.cs
```diff
@@ -53,7 +53,7 @@ public async Task GetNextHiLo()
                 var tag = GetQueryStringValueAndAssertIfSingleAndNotEmpty("tag");
                 var lastSize = GetLongQueryString("lastBatchSize", false) ?? 0;
                 var lastRangeAt = GetStringQueryString("lastRangeAt", false);
-                var identityPartsSeparator = GetStringQueryString("identityPartsSeparator", false) ?? "/";
+                var identityPartsSeparator = GetCharQueryString("identityPartsSeparator", false) ?? Database.IdentityPartsSeparator;
```
ppekrol, line 56: Need to think if this will not cause issues when 4.2 client will talk with 5.x server

ayende: How could it? it will just default to the usual one, no?

ayende: Maybe a problem if you are using a multi char string, you mean? Very unlikely

ayende (approved):

PR #10130, opened 2020-02-01
RavenDB-14487 tombstone cleaner should delete in batches

ppekrol:


src/Raven.Server/Documents/TombstoneCleaner.cs
```diff
@@ -69,7 +73,14 @@ internal async Task ExecuteCleanup()
                 if (state.Tombstones.Count == 0)
                     return;
 
-                await _documentDatabase.TxMerger.Enqueue(new DeleteTombstonesCommand(state.Tombstones, state.MinAllDocsEtag, _documentDatabase, Logger));
+                while (CancellationToken.IsCancellationRequested == false)
+                {
+                    var command = new DeleteTombstonesCommand(state.Tombstones, state.MinAllDocsEtag, numberOfTombstonesToDeleteInBatch ?? _numberOfTombstonesToDeleteInBatch, _documentDatabase, Logger);
+                    await _documentDatabase.TxMerger.Enqueue(command);
+
+                    if (command.NumberOfTombstonesDeleted <= 0)
```
karmeli87: This mean you will always issue at least 2 delete commands, one will delete and the other one will confirm that there is no further deletions needed.

Would something like 
```
command.NumberOfTombstonesDeleted <= numberOfTombstonesToDeleteInBatch ?? _numberOfTombstonesToDeleteInBatch
```
 
be better?

ppekrol: True. Pushed.

ayende (approved):

PR #10137, opened 2020-02-03
RavenDB-14304 fix issues when appeneding values with different dimens…

karmeli87:
…ions

### DIFF
--- a/src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
+++ b/src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
@@ -1248,7 +1248,7 @@ namespace Raven.Server.Documents.TimeSeries
                                 break;
                             }
 
-                            EnsureNumberOfValues(context, segmentHolder.ReadOnlySegment.NumberOfValues, current);
+                            EnsureNumberOfValues(segmentHolder.ReadOnlySegment.NumberOfValues, current);
 
                             if (TryAppendToCurrentSegment(context, segmentHolder, appendEnumerator, out var newValueFetched))
                                 break;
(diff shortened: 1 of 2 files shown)


src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
```diff
@@ -1113,6 +1095,32 @@ Reader.SingleResult ToResult(TimeSeriesOperation.AppendOperation element)
             return context.LastDatabaseChangeVector;
         }
 
+        private bool ValueTooFar(TimeSeriesSegmentHolder segmentHolder, Reader.SingleResult current)
+        {
+            var deltaInMs = (current.Timestamp.Ticks / 10_000) - segmentHolder.BaselineMilliseconds;
```
ppekrol, line 1296: Can we wrap this in something similar to `Size`
I'm afraid that without that it is a potential place for easy bug introduction. Let's talk about this tomorrow.

ppekrol (approved):

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
ayende: Seems reasonable to do so, yes

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
```diff
@@ -1113,6 +1095,32 @@ Reader.SingleResult ToResult(TimeSeriesOperation.AppendOperation element)
...
+                {
+                    var fullValues = new Span<double>(fullValuesBuffer.Ptr, segmentNumberOfValues);
+                    current.Values.Span.CopyTo(fullValues);
+                    for (int i = current.Values.Length; i < fullValues.Length; i++)
+                    {
+                        fullValues[i] = double.NaN;
+                    }
+
+                    current.Values = new Memory<double>(fullValues.ToArray());
+                }
```
ayende: You are returning the allocation here, but still using it.
This is _not_ valid

ayende (changes requested):

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
karmeli87: `fullValues.ToArray()` copy them to managed memory?

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
karmeli87: Let's do in a separate issue: https://issues.hibernatingrhinos.com/issue/RavenDB-14577

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
ayende: Oh, correct, but then why bother will all the song and ance? Just allocate them once directly and save all the complexity.

karmeli87: right.

ayende (approved):

PR #10152, opened 2020-02-05
RavenDB-14576: fix NRE & Wrong results in js index

garayx:


src/Raven.Server/Documents/Indexes/MapReduce/ReduceMapResultsBase.cs
```diff
@@ -335,7 +335,7 @@ private void WriteLastEtags(TransactionOperationContext indexContext)
 
                         _index.ErrorIndexIfCriticalException(e);
 
-                        HandleReductionError(e, reduceKeyHash, writer, stats, updateStats: parentPage == -1, page: leafPage);
+                        HandleReductionError(e, reduceKeyHash, writer, stats, updateStats: true, page: leafPage);
```
ayende: What is the reasoning here?

arekpalinski (approved):

src/Raven.Server/Documents/Indexes/MapReduce/ReduceMapResultsBase.cs
arekpalinski: I wouldn't touch this. The original idea was to update stats only for the highest level reduce errors in the reduce tree. Those are supposed to be propagated from the bottom to the top anyway.

arekpalinski: Note that we might be getting an error because of the same reduce output multiple times when going over reduce tree - that's why we wanted stats to be updated just ones for a tree.

garayx: The problem is a case that you get a _serious_ error which is not in the root. Than the index is not marked as _error_ and the index stats doesn't have any errors, which makes few methods (waitforindexing, getindexerrorsoperation, getindexstatsoperation) have wrong results, also its doesn't displayed as faulty in the studio. 

In the case that reduce has errors the index definition going to changed, so I don't see any problems here.

garayx: another option could be updateStats when parentPage == -1 || typeof(e) == NRE (or other reasonable errors to log them)

arekpalinski: What test reproduces that? I'd like to see it with debugger

garayx: JavascriptIndexShouldPassContextToTypeConverter will throw NRE in StoreAggregationResult(leafPage, table, result); above, when parentPage != -1

arekpalinski: 1. Did you mean `JavascriptIndexShouldThrowOnExceptionInReduceStage` test? (`JavascriptIndexShouldPassContextToTypeConverter ` doesn't enter in that piece of code).


Anyway, I have reverted that change and all tests `RavenDB_14576` are passing. Am I missing something?

2. You said `"when parentPage != -1"` - the idea was that if an error is thrown at leaf level and that leaf isn't a root then an error is supposed to be thrown at branch level as well. Then we always update stats:

https://github.com/ravendb/ravendb/blob/ec89c9988d3882204cd82fabd07188fb0eeb0e9d/src/Raven.Server/Documents/Indexes/MapReduce/ReduceMapResultsBase.cs#L393

Did you see a case where it didn't happen?

3. I'm not saying "no" for that change but just want to be sure we really need this.

garayx: Did you remove the context to get the NRE ? https://github.com/ravendb/ravendb/pull/10152/files#diff-fb0e7df433238a2900c2270687216d68R40

Thats what I get:
image

arekpalinski: No, I didn't. So, now I repeated it with removing the context to force NRE and I still got indexing error where NRE was included as the inner exception (what is fine):

image

arekpalinski: And `JavascriptIndexShouldPassContextToTypeConverter` test was failing as expected then

garayx: > 1. You said `"when parentPage != -1"` - the idea was that if an error is thrown at leaf level and that leaf isn't a root then an error is supposed to be thrown at branch level as well. Then we always update stats:
> 
> https://github.com/ravendb/ravendb/blob/ec89c9988d3882204cd82fabd07188fb0eeb0e9d/src/Raven.Server/Documents/Indexes/MapReduce/ReduceMapResultsBase.cs#L393

The index error do get thrown but it doesn't mark the index as faulty because we didn't get the 15% of results as error, and this is because we only throw it on the root level.

arekpalinski: That's the point - threshold limit for marking an index as error. And that's why we updated the stats only if `parentPage == -1`.

Note that in your case the error was triggered by _bug_ in RavenDB. It's not "regular" indexing error - like invalid index definition, invalid documents to index etc. 

After your change we're going to mark index as faulty _much faster_ because we'll error multiple times because of the same error - at leafs level (which is going to be a lot) and the same at branch levels. We shouldn't behave this way for "regular" indexing errors.

The current behavior is fine IMO - we had a bug in RavenDB, we put it as index error so user can get back to us and let us know about it (no error was hidden). Marking index as error would be good in this particular case, but your change affects "regular" indexing errors and how fast we mark indexes as faulty.

HibernatingRhinosServiceAccount:
When the index has errors (doesn't matter which type regular or not), don't
we want to actually mark it as failed as fast as possible?
Because from my perspective the index should get errors only during the
index build stage, in this case we want to show the user that the index
faulty as fast as possible so he can rework it.
What is the chance that an index that is working in application
suddenly will start to fail ? I dont think they are big, and if it does ? I
think we still want to mark the index as faulty.

Tell me your decision, so I can make a commit to change it back.

arekpalinski:
I don't want to change that behavior. I'm afraid this could be a breaking change  - for large indexes we can have thousands of leafs so any errors detected there could make index as faulty very early while currently this index index is able to return some results - regardless the errors (which are always bad and user should fix them asap)

arekpalinski:
So please let's revert this change

arekpalinski (approved):

PR #10153, opened 2020-02-05
RavenDB-14503 Handle stats for time-series

karmeli87:
- store the total count in a dedicated tree. So we can retrive it fast.
- lookup for the start/end date time

### DIFF
--- a/src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
+++ b/src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
@@ -141,48 +141,29 @@ namespace Raven.Server.Documents.TimeSeries
                 var start = DateTime.MinValue;
                 var end = DateTime.MaxValue;
 
-                // find the first good segment for this ts
-                foreach (var (key, value) in table.SeekByPrimaryKeyPrefix(slice, Slices.Empty, 0))
+                foreach (var (key, _) in table.SeekByPrimaryKeyPrefix(slice, Slices.Empty, 0))
                 {
-                    var ptr = value.Reader.Read((int)TimeSeriesTable.Segment, out var size);
-                    var segment = new TimeSeriesValuesSegment(ptr, size);
-                    if (segment.NumberOfLiveEntries > 0)
-                    {
-                        var baseline = ExtractDateTimeFromKey(key);
-                        foreach (var result in segment.YieldAllValues(context, context.Allocator, baseline))
-                        {
-                            if (result.Status == TimeSeriesValuesSegment.Live)
-                            {
-                                start = result.Timestamp;
-                                break;
-                            }
-                        }
-                        break;
-                    }
+                    start = ExtractDateTimeFromKey(key);
+                    break;
                 }
 
                 // to find the last value we set to highest possible segment value (long.MaxValue) and seeking backward from there.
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
```diff
@@ -106,13 +110,111 @@ static TimeSeriesStorage()
...
+                {
+                    var segmentPtr = tvr.Read((int)TimeSeriesTable.Segment, out var size);
+                    var segment = new TimeSeriesValuesSegment(segmentPtr, size);
+
+                    var keyPtr = tvr.Read((int)TimeSeriesTable.TimeSeriesKey, out var keySize);
+                    
+                    dispose.Dispose();
+                    dispose = Slice.External(context.Allocator, keyPtr, keySize, out var keySlice);
+                    lastKey = keySlice;
+                    if (segment.NumberOfLiveEntries > 0)
```
ayende: What if I have 100000 segments with no live values?

karmeli87: It depends..
If they are deleted as a segment, we store only a single dead entry so each page will probably have several hundreds of those.. Which you still have to iterate.
If those segments are full with dead values, it is indeed a problem.

But in both cases after the cleaner did his job this shouldn't matter. 

I see 2 options:
1. We can do binary search (not sure this is a good idea.. the cost might be even higher)
2. Somehow track the first value

I don't see any nice/better way here.
I will test it on the 300M points data and see if it an issue.
If it isn't we leave this optimization for later. I think.

ayende: Alternatively, ignore dead values and report start /end as they are.
Then let the cleaner do its job?

karmeli87: Also an option

ayende: Let's go with this

karmeli87: done

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
```diff
@@ -106,13 +110,92 @@ static TimeSeriesStorage()
...
         {
             _documentDatabase = documentDatabase;
             _documentsStorage = documentDatabase.DocumentsStorage;
 
             tx.CreateTree(TimeSeriesKeysSlice);
             tx.CreateTree(DeletedRangesKey);
+            tx.CreateTree(TimeSeriesStats);
+        }
+
+        private void UpdateTotalCountStats(DocumentsOperationContext context, string docId, string name, long count)
```
ayende, line 124: I think that I'm missing calls to this from where we _delete_ values, no?

karmeli87: Upon every segment update, I reduce the old count and add the new count. This valid also for deletes. Even when we remove entire segment, we replace it with an empty one.. so the same logic apply as well.

ayende (approved):

ppekrol:
We still need tests that cover Start/End dates.

PR #10157, opened 2020-02-05
RDBCL-1248 disable downloading the crl file

iftahbe:


ayende (commented):
Missing the same behavior we have in the TCP endpoint

iftahbe:
It's already set to false. (The last variable is checkCertificateRevocation)
https://github.com/ravendb/ravendb/blob/a544f15faa10713a187b1479bf048911c1c5218b/src/Raven.Server/RavenServer.cs#L2211

ayende:
Let's use the SslServerOptions overload, instead.
There are more options that we turn off there.

ml054:
Does it have some impact on security? I know that on the other side, timeout on check kills us.

ayende:
No impact on security, we aren't relying on the CRL and we actually get better security by NOT doing this.

ml054:
May I ask why is it better?

ayende:
@ml054 
a) Questions are always welcome, especially in such a situation where it is not always obvious what is going on.
b) The security issue is that if we leave our own system, we may be subject to DoS attack. You can do that without needing any credentials, by connecting to us with a bad cert that has a CRL URL that won't respond. That can hold up the thread.
Do that 100 times, and there are no threads available, we are down.

There is no security impact because if the CRL is not available, we fail open and accept the cert. We have documented that we don't rely on PKI (and hence, CRLs) for client auth, so that is known.

ml054:
oh right, I forgot that we have second fence to force before breaking in, as having valid cert isn't enought to access system. Thanks for answer. :)

src/Raven.Server/RavenServer.cs
```diff
@@ -2207,8 +2207,15 @@ private async Task<(Stream Stream, X509Certificate2 Certificate)> AuthenticateAs
                         // SSL, because that generate a nicer error for the user to read then just aborted
                         // connection because SSL negotiation failed.
                         true);
-
-                await sslStream.AuthenticateAsServerAsync(Certificate.Certificate, true, TcpUtils.SupportedSslProtocols, false);
+                await sslStream.AuthenticateAsServerAsync(new SslServerAuthenticationOptions
+                {
+                    ServerCertificate = Certificate.Certificate,
+                    ClientCertificateRequired = true,
+                    AllowRenegotiation = false,
```
ppekrol: This by default is true. Do we want to change this? https://github.com/dotnet/runtime/blob/516956a01fee9aa732b199012a6ea0bcc2bd789f/src/libraries/System.Net.Security/src/System/Net/Security/SslServerAuthenticationOptions.cs#L16-L22

iftahbe: Probably don't want to change this, you're right. I don't think we want to interfere with their logic.
https://github.com/dotnet/runtime/blob/6f6c94dbb2259deb083eefb755698cc8e0f0a4ce/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpConnectionPool.cs#L216

ayende (approved):

PR #10166, opened 2020-02-06
RavenDB-14247 : Allow to specify offset in timeseries queries

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-14247


ayende (approved):

ayende:
There is one thing that still needs to be done here. We need to _convert_ the dates from the UTC to the specified offset. 
Note that this needs to be done _before_ we do the group by, because it may change what *bucket* something will go on.
For example, if I do `group by 1d offset '02:00' `  , we'll want to see days in Israel's time

ppekrol (approved):
We need to look into the case that @ayende described.

PR #10171, opened 2020-02-08
RavenDB-14336 apply TLS Cipher Suites to SslStream

ppekrol:


src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
```diff
@@ -211,7 +211,14 @@ await requestExecutor.ExecuteAsync(_redirectNode, null, context, command, should
                 _tcpClient.SendBufferSize = _options?.SendBufferSizeInBytes ?? SubscriptionWorkerOptions.DefaultSendBufferSizeInBytes;
                 _tcpClient.ReceiveBufferSize = _options?.ReceiveBufferSizeInBytes ?? SubscriptionWorkerOptions.DefaultReceiveBufferSizeInBytes;
                 _stream = _tcpClient.GetStream();
-                _stream = await TcpUtils.WrapStreamWithSslAsync(_tcpClient, tcpInfo, _store.Certificate, requestExecutor.DefaultTimeout).ConfigureAwait(false);
+                _stream = await TcpUtils.WrapStreamWithSslAsync(
+                    _tcpClient,
+                    tcpInfo,
+                    _store.Certificate,
+#if !(NETSTANDARD2_0 || NETCOREAPP2_1)
+                    null,
```
ayende, line 219: Why pass null here? Because it is on the client?

src/Raven.Client/Util/TcpUtils.cs
```diff
@@ -106,7 +107,13 @@ public static async Task<TcpClient> ConnectAsync(string url, TimeSpan? timeout =
             return tcpClient;
         }
 
-        internal static async Task<Stream> WrapStreamWithSslAsync(TcpClient tcpClient, TcpConnectionInfo info, X509Certificate2 storeCertificate,
+        internal static async Task<Stream> WrapStreamWithSslAsync(
+            TcpClient tcpClient,
+            TcpConnectionInfo info,
+            X509Certificate2 storeCertificate,
+#if !(NETSTANDARD2_0 || NETCOREAPP2_1)
+            CipherSuitesPolicy cipherSuitesPolicy,
```
ayende, line 115: Am I missing something? I don't see where this is being used.

src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
ppekrol: Correct.

src/Raven.Client/Util/TcpUtils.cs
ppekrol: Thanks

PR #10173, opened 2020-02-08
RavenDB-7070 detect processor affinity changes in tests

ppekrol:


src/Raven.Server/Commercial/LicenseManager.cs
```diff
@@ -1064,6 +1067,12 @@ private void SetAffinity(Process process, int cores, bool addPerformanceHint)
             try
             {
                 var currentlyAssignedCores = Bits.NumberOfSetBits(process.ProcessorAffinity.ToInt64());
+                if (IgnoreProcessorAffinityChanges && currentlyAssignedCores != cores)
+                {
+                    Console.WriteLine($"Processor affinity change detected. Current cores: {currentlyAssignedCores}. Requested cores: {cores}.");
```
ayende, line 1072: Why do we want to run `Console.WriteLine` here?

ppekrol: It is only used in tests. I can easily check the test output because of that.

ayende: It is only used in tests right now, I worry about this changing down the line.

ppekrol: After https://issues.hibernatingrhinos.com/issue/RavenDB-14605 is done I will switch this to throw during test execution.

ayende: Great!

PR #10176, opened 2020-02-09
RavenDB-14612 - Critical index error (out of disk space)

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-14612


ayende (commented):
Needs to do recompliation as well
@aviviadi

PR #10179, opened 2020-02-09
RavenDB-14606 Switching ContextPool to per core model.  Removed some …

ayende:
…behaviours that assume per thread model. We are now more consistent and conservative about how we retain memory and when we release it.

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -991,8 +991,6 @@ namespace Raven.Server.Documents.Indexes
                     return; // can be null if we disposed immediately
                 try
                 {
-                    _contextPool.SetMostWorkInGoingToHappenOnThisThread();
-
                     DocumentDatabase.Changes.OnDocumentChange += HandleDocumentChange;
                     storageEnvironment.OnLogsApplied += HandleLogsApplied;
 
--- a/src/Raven.Server/ServerWide/Context/TransactionContextPool.cs
+++ b/src/Raven.Server/ServerWide/Context/TransactionContextPool.cs
@@ -15,16 +15,6 @@ namespace Raven.Server.ServerWide.Context
     {
         private StorageEnvironment _storageEnvironment;
 
-        // this is safe to do across instances, because all the pools in a thread are going to share
-        // the same optimizations
-        [ThreadStatic]
-        private static bool _mostlyThreadDedicatedWork;
-
-        public void SetMostWorkInGoingToHappenOnThisThread()
-        {
-            _mostlyThreadDedicatedWork = true;
-        }
-
         public TransactionContextPool(StorageEnvironment storageEnvironment)
         {
             _storageEnvironment = storageEnvironment;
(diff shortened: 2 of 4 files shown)


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -991,8 +991,6 @@ protected void ExecuteIndexing()
                     return; // can be null if we disposed immediately
                 try
                 {
-                    _contextPool.SetMostWorkInGoingToHappenOnThisThread();
```
ayende, line 994: Assumes that we are using per thread values, no longer relevant

src/Raven.Server/ServerWide/Context/TransactionContextPool.cs
```diff
@@ -15,16 +15,6 @@ public class TransactionContextPool : JsonContextPoolBase<TransactionOperationCo
...
-        // this is safe to do across instances, because all the pools in a thread are going to share
-        // the same optimizations
-        [ThreadStatic]
-        private static bool _mostlyThreadDedicatedWork;
-
-        public void SetMostWorkInGoingToHappenOnThisThread()
-        {
-            _mostlyThreadDedicatedWork = true;
-        }
-
```
ayende, line 27: Assumes that we are using per thread values, no longer relevant

src/Raven.Server/ServerWide/Context/TransactionContextPool.cs
```diff
@@ -40,14 +30,6 @@ protected override TransactionOperationContext CreateContext()
...
             {
                 initialSize = 32*1024;
-
-                if (_mostlyThreadDedicatedWork)
-                {
-                    // if this is a context pool dedicated for a thread (like for indexes), we probably won't do a lot of 
-                    // work on that outside of its thread, so let not allocate a lot of memory for that. We just need enough
-                    // there process simple stuff like IsStale, etc, so let us start small
-                    initialSize = 16 * 1024 * 1024;  // the initial budget is 32 MB, so let us now blow through that all at once
-                }
```
ayende, line 50: Assumes that we are using per thread values, no longer relevant

src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -1,240 +1,34 @@
...
-                    if (ctx == null)
-                        continue;
-                    if (!ctx.InUse.Raise())
-                        continue;
-                    ctx.Dispose();
-                }
-            }
-        }
+        private readonly T[][] _perCoreContexts;
+        private readonly ConcurrentQueue<T> _globalQueue = new ConcurrentQueue<T>();
```
ayende, line 20: We have up to 64 slots per core, and then we fail over to a global queue. 
The idea is that we have an intentionally small per core value and then a shared system.
There is no work stealing here because we assume that the cores are going to be scheduled by the OS significantly.

src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -426,10 +278,30 @@ public void LowMemoryOver()
...
+        private static extern int sched_getcpu();
+
+        [DllImport("kernel32.dll", SetLastError = true)]
+        private static extern int GetCurrentProcessorNumber();
+
+        public static int CurrentProcessNumber
+        {
+            get
+            {
+                if (PlatformDetails.RunningOnPosix)
```
ayende, line 295: Need to measure whatever we have to cache that, see the ProcessorIdCache here:
https://github.com/dotnet/runtime/blob/master/src/coreclr/src/System.Private.CoreLib/src/System/Threading/ProcessorIdCache.cs

PR #10183, opened 2020-02-10
RavenDB-14441 Indexes refresh when Auto-Refresh is Off

Danielle9897:


src/Raven.Studio/typescript/viewmodels/database/indexes/deleteIndexesConfirm.ts
```diff
@@ -35,7 +35,7 @@ class deleteIndexesConfirm extends dialogViewModelBase {
             throw new Error("Indexes must not be null or empty.");
         }
         
-        this.indexesInfoForDelete = (indexes as index[]).map(x => new indexInfoForDelete(x.name, x.reduceOutputCollectionName(), x.hasPatternForReduceOutputCollection()));
+        this.indexesInfoForDelete = (indexes as indexDefinition[]).map(x => new indexInfoForDelete(x.name(), x.reduceOutputCollectionName(), x.hasPatternForReduceOutputCollection()));
```
ml054: does it work from indexes page now? maybe we can use ko.unwrap function which unwrap value if this is observable or returns input value when it isn;t

Danielle9897: used ko.unwrap - thanks!

ml054 (approved):

PR #10186, opened 2020-02-11
RavenDB-14570 tighten ReAddMemberNode test

karmeli87:


src/Raven.Server/Commercial/LicenseStatus.cs
```diff
@@ -70,9 +70,11 @@ public bool Expired
             }
         }
 
-        public int MaxCores => GetValue<int?>("cores") ?? 3;
+        internal static bool UnderTest;
 
-        public int MaxMemory => GetValue<int?>("memory") ?? 6;
+        public int MaxCores => GetValue<int?>("cores") ?? (UnderTest ? 3 : 256);
```
ppekrol: What is this?

karmeli87: Should be the other way around.. (UnderTest ? 256 : 3)
If we unable to set the license we should use 256 cores as default and not 3

src/Raven.Server/Rachis/Leader.cs
```diff
@@ -57,6 +57,8 @@ public class Leader : IDisposable
         private readonly ConcurrentDictionary<string, FollowerAmbassador> _nonVoters =
             new ConcurrentDictionary<string, FollowerAmbassador>(StringComparer.OrdinalIgnoreCase);
 
+        public Dictionary<string, FollowerAmbassador> CurrentPeers => new  Dictionary<string, FollowerAmbassador>(_voters.Concat(_nonVoters).Concat(_promotables));
```
ayende: What updates this? What happens when we add a new node? Remove one?

karmeli87: `RefreshAmbassadors` will update this and handle add/remove.

ayende: This is only called from the tests, right?

ayende (approved):

src/Raven.Server/Rachis/Leader.cs
ppekrol: Yes, but maybe we should add a comment 'DEBUG ONLY', just in case

PR #10190, opened 2020-02-12
RavenDB-14510 Better controls over embedded server

ayende:


src/Raven.Embedded/EmbeddedServer.cs
```diff
@@ -29,26 +31,48 @@ internal EmbeddedServer()
...
+            if (server == null)
+                throw new InvalidOperationException($"Please run {nameof(StartServer)}() before trying to use the server");
+            return (await server.Value.WithCancellation(token).ConfigureAwait(false)).ServerProcess.Id;
+        }
 
-            _gracefulShutdownTimeout = options.GracefulShutdownTimeout;
-            var startServer = new Lazy<Task<(Uri ServerUrl, Process ServerProcess)>>(() => RunServer(options));
-            if (Interlocked.CompareExchange(ref _serverTask, startServer, null) != null)
-                throw new InvalidOperationException("The server was already started");
+        public void Restart()
```
ppekrol: Maybe make this method async?

ppekrol: It should be called `RestartServer` to match the `StartServer` one

src/Raven.Embedded/EmbeddedServer.cs
```diff
@@ -193,13 +229,20 @@ private void ShutdownServerProcess(Process process)
             }
         }
 
-        private async Task<(Uri ServerUrl, Process ServerProcess)> RunServer(ServerOptions options)
+        public event EventHandler? ServerProcessExited;
```
ppekrol: We should add `ServerProcessExitedEventArgs` here. They can be empty for now on, but it is easier to expand later on.

src/Raven.Embedded/EmbeddedServer.cs
ayende: I made is sync and run in background because StartServer is also that.
But it make sense that you'll want to get any errors immediately.

ppekrol (approved):

PR #10198, opened 2020-02-13
RavenDB-14247 Allow to specify offset in timeseries queries

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-14247

### DIFF
--- a/src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
+++ b/src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
@@ -542,9 +542,9 @@ namespace Raven.Server.Documents.TimeSeries
             }
         }
 
-        public Reader GetReader(DocumentsOperationContext context, string documentId, string name, DateTime from, DateTime to)
+        public Reader GetReader(DocumentsOperationContext context, string documentId, string name, DateTime from, DateTime to, TimeSpan? offset = null)
         {
-            return new Reader(context, documentId, name, from, to);
+            return new Reader(context, documentId, name, from, to, offset);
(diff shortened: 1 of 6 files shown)


src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
```diff
@@ -565,17 +565,26 @@ public class Reader
...
 
                 _from = EnsureMillisecondsPrecision(from);
                 _to = EnsureMillisecondsPrecision(to);
+
+                if (_offset.HasValue)
+                {
+                    _from = _from.Add(_offset.Value);
+                    _to = _to.Add(_offset.Value);
+                }
+
```
ayende: It would be easier to still require min & max to be in UTC, no?

ppekrol: So what is the scope of the offset?

ayende: The entire query

ppekrol: I do not understand your comment here then.

ayende: Instead of add flags to `GetNextRangeStart`, just always require it to be UTC
And convert before calling this code

ppekrol:
@aviv86 can you rebase?

ppekrol (approved):

ayende (approved):

PR #10201, opened 2020-02-14
RDBC-301 Possible missing escaping of URL data

ml054:


ml054:
we have few failing tests, I didn't realize escaping throw for null value.

src/Raven.Client/Documents/Operations/Indexes/GetTermsOperation.cs
```diff
@@ -45,7 +45,7 @@ public GetTermsCommand(string indexName, string field, string fromValue, int? pa
 
             public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, ServerNode node, out string url)
             {
-                url = $"{node.Url}/databases/{node.Database}/indexes/terms?name={Uri.EscapeDataString(_indexName)}&field={Uri.EscapeDataString(_field)}&fromValue={_fromValue}&pageSize={_pageSize}";
+                url = $"{node.Url}/databases/{node.Database}/indexes/terms?name={Uri.EscapeDataString(_indexName)}&field={Uri.EscapeDataString(_field)}&fromValue={Uri.EscapeDataString(_fromValue)}&pageSize={_pageSize}";
```
ml054: from value might be null!

src/Raven.Server/Web/Studio/DataDirectoryInfo.cs
```diff
@@ -230,8 +230,8 @@ public GetDataDirectoryInfoCommand(string path, string name, bool isBackup)
             public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, ServerNode node, out string url)
             {
                 var path = _isBackup
-                    ? $"databases/{_name}/admin/backup-data-directory?path={_path}"
-                    : $"admin/studio-tasks/full-data-directory?path={_path}&name={_name}";
+                    ? $"databases/{_name}/admin/backup-data-directory?path={Uri.EscapeDataString(_path)}"
```
ml054: check for null values!
in /admin/studio-tasks/full-data-directory path isn't required.

src/Raven.Client/Documents/Operations/Indexes/GetTermsOperation.cs
ml054: done

PR #10206, opened 2020-02-16
RavenDB-14555 Reducing memory pressure and fixing thread local

ayende:


src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -39,11 +40,13 @@ static LoggingSource()
         private readonly ManualResetEventSlim _readyToCompress = new ManualResetEventSlim(false);
         private readonly CancellationTokenSource _tokenSource = new CancellationTokenSource();
         private readonly LightWeightThreadLocal<LocalThreadWriterState> _localState;
+
+        private readonly SingleProducerSingleConsumerCircularQueue<WebSocketMessageEntry>[] _pooledMessageEntries;
+        private readonly SingleProducerSingleConsumerCircularQueue<WebSocketMessageEntry>[] _activeMessages;
```
arekpalinski, line 45: Maybe better names like:

- `_freePooledMessageEntries`
- `_activePoolMessageEntries`

?

src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -471,14 +485,38 @@ public int Compare(string x, string y)
...
+        {
+            internal const string LIBC_6 = "libc";
+
+            [DllImport(LIBC_6, EntryPoint = "sched_getcpu", CallingConvention = CallingConvention.Cdecl, SetLastError = false)]
+            private static extern int sched_getcpu();
+
+            [DllImport("kernel32.dll", SetLastError = true)]
+            private static extern int GetCurrentProcessorNumber();
+
+            public static int CurrentProcessNumber
```
arekpalinski, line 506: `CurrentProcessorNumber` or `CurrentProcessorIndex` ?

src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -489,7 +527,10 @@ public void Log(ref LogEntry entry, TaskCompletionSource<object> tcs = null)
                 state = _localState.Value = GenerateThreadWriterState();
             }
 
-            if (state.Free.Dequeue(out var item))
+            int currentProcessNumber = JsonContextPoolHelpers.CurrentProcessNumber;
+            var pool = _pooledMessageEntries[currentProcessNumber% _pooledMessageEntries.Length];
```
arekpalinski, line 531: Why do we need `% _pooledMessageEntries.Length` ? Shouldn't property `CurrentProcessNumber` already ensure that value is adjusted to `Environment.ProcessorCount`?

src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -514,7 +554,7 @@ public void Log(ref LogEntry entry, TaskCompletionSource<object> tcs = null)
             WriteEntryToWriter(state.Writer, ref entry);
             item.Data = state.ForwardingStream.Destination;
 
-            state.Full.Enqueue(item, timeout: 128);
+            _activeMessages[currentProcessNumber%_activeMessages.Length].Enqueue(item, timeout: 128);
```
arekpalinski, line 557: Same as above - why `%_activeMessages.Length`?

src/Sparrow/Threading/LightWeightThreadLocal.cs
```diff
@@ -11,13 +11,27 @@ public sealed class LightWeightThreadLocal<T> : IDisposable
...
         private readonly Func<T> _generator;
         private readonly GlobalState _globalState = new GlobalState();
 
         public LightWeightThreadLocal(Func<T> generator = null)
         {
             _generator = generator;
+            SelfReference = new WeakReferenceCompareValue<LightWeightThreadLocal<T>>(this);
+        }
+
+        public List<object> GetTargets()
```
arekpalinski, line 25: Is it for debugging? Can we change accessibility to `internal`?

src/Sparrow/Threading/LightWeightThreadLocal.cs
```diff
@@ -184,9 +207,18 @@ public override bool Equals(object obj)
                     return false;
                 if (ReferenceEquals(this, obj))
                     return true;
+                if (obj.GetType() == typeof(TK))
+                {
+                    int hashCode = obj.GetHashCode();
+                    if (hashCode != _hashCode)
+                        return false;
+                    if (_weak.TryGetTarget(out var other) == false)
```
arekpalinski, line 215: Just to make sure - now we don't reference in `LightWeightThreadLocal._values` collection directly but via `WeakReferenceCompareValue` so it can return `false` more often, right? And that makes we we'll cleanup local thread resources - is it correct?

src/Sparrow/Logging/LoggingSource.cs
ayende: We technically don't need it, but I want to be on the safe side here.

src/Sparrow/Threading/LightWeightThreadLocal.cs
ayende: Yes, that is correct.

src/Sparrow/Logging/LoggingSource.cs
ppekrol: Maybe it is better to do it inside the property then?

src/Sparrow/Threading/LightWeightThreadLocal.cs
ayende: Removed

src/Sparrow/Logging/LoggingSource.cs
ayende: Moved it so it is only called once.

PR #10208, opened 2020-02-16
RavenDB-13083 TimeSeries patching

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-13083

### DIFF
--- a/src/Raven.Server/Documents/Patch/ScriptRunner.cs
+++ b/src/Raven.Server/Documents/Patch/ScriptRunner.cs
@@ -237,14 +237,14 @@ namespace Raven.Server.Documents.Patch
                 }
             }
 
-            private (string Id, BlittableJsonReaderObject Doc) GetIdAndDocFromArg(JsValue firstArgs, string signature)
+            private (string Id, BlittableJsonReaderObject Doc) GetIdAndDocFromArg(JsValue docArg, string signature)
             {
-                if (firstArgs.IsObject() && firstArgs.AsObject() is BlittableObjectInstance doc)
+                if (docArg.IsObject() && docArg.AsObject() is BlittableObjectInstance doc)
                     return (doc.DocumentId, doc.Blittable);
 
-                if (firstArgs.IsString())
+                if (docArg.IsString())
                 {
-                    var id = firstArgs.AsString();
+                    var id = docArg.AsString();
                     var document = _database.DocumentsStorage.Get(_docsCtx, id);
                     if (document == null)
                         throw new DocumentDoesNotExistException(id, "Cannot operate on a missing document.");
@@ -252,27 +252,29 @@ namespace Raven.Server.Documents.Patch
                     return (id, document.Data);
                 }
 
-                throw new InvalidOperationException($"{signature}: 'doc' must be a string argument (the document id) or the actual document instance itself");
+                throw new InvalidOperationException($"{signature}: 'doc' must be a string argument (the document id) or the actual document instance itself. {GetTypes(docArg)}");
             }
+
+            private static string GetTypes(JsValue value) => $"JintType({value.Type}) .NETType({value.GetType().Name})";
             
-            private string GetIdFromArg(JsValue firstArgs, string signature)
+            private string GetIdFromArg(JsValue docArg, string signature)
             {
-                if (firstArgs.IsObject() && firstArgs.AsObject() is BlittableObjectInstance doc)
+                if (docArg.IsObject() && docArg.AsObject() is BlittableObjectInstance doc)
                     return doc.DocumentId;
 
-                if (firstArgs.IsString())
+                if (docArg.IsString())
                 {
-                    var id = firstArgs.AsString();
+                    var id = docArg.AsString();
                     return id;
                 }
 
-                throw new InvalidOperationException($"{signature}: 'doc' must be a string argument (the document id) or the actual document instance itself");
+                throw new InvalidOperationException($"{signature}: 'doc' must be a string argument (the document id) or the actual document instance itself. {GetTypes(docArg)}");
             }
             
             private static string GetStringArg(JsValue jsArg, string signature, string argName)
             {
                 if (jsArg.IsString() == false)
-                    throw new ArgumentException($"{signature}: The {argName} argument should be a string, but got {jsArg.GetType().Name}");
+                    throw new ArgumentException($"{signature}: The {argName} argument should be a string, but got {GetTypes(jsArg)}");
                 return jsArg.AsString();
             }
             
@@ -284,7 +286,7 @@ namespace Raven.Server.Documents.Patch
                     if (key == "length")
                         continue;
                     if (value.Value.IsNumber() == false)
-                        throw new ArgumentException($"{signature}: The values argument must be an array of numbers, but got {value.Value.GetType().Name} key({key}) value({value})");
+                        throw new ArgumentException($"{signature}: The values argument must be an array of numbers, but got {GetTypes(value.Value)} key({key}) value({value})");
                     array[i] = value.Value.AsNumber();
                     ++i;
                 }
@@ -337,7 +339,7 @@ namespace Raven.Server.Documents.Patch
                     }
                     else
                     {
-                        throw new ArgumentException($"{signature}: The values should be an array but got {valuesArg.Type.GetType().Name}");
+                        throw new ArgumentException($"{signature}: The values should be an array but got {GetTypes(valuesArg)}");
                     }
 
                     var toAppend = new TimeSeriesStorage.Reader.SingleResult
@@ -1243,7 +1245,7 @@ namespace Raven.Server.Documents.Patch
                 }
                 
                 void ThrowInvalidDateArgument() =>
-                    throw new ArgumentException($"{signature} : {argName} must be of type 'DateInstance' or a DateTime string");
+                    throw new ArgumentException($"{signature} : {argName} must be of type 'DateInstance' or a DateTime string. {GetTypes(arg)}");
             }
 
             private static unsafe JsValue ToStringWithFormat(JsValue self, JsValue[] args)
--- a/src/Raven.Server/Documents/Patch/ScriptRunner.cs
+++ b/src/Raven.Server/Documents/Patch/ScriptRunner.cs
@@ -15,6 +15,7 @@ using Jint.Native.Function;
 using Jint.Native.Object;
 using Jint.Runtime.Interop;
 using Raven.Client.Documents.Indexes.Spatial;
+using Raven.Client.Documents.Session.TimeSeries;
 using Raven.Client.Exceptions.Documents;
 using Raven.Client.Exceptions.Documents.Patching;
 using Raven.Server.Config;
@@ -236,7 +237,7 @@ namespace Raven.Server.Documents.Patch
                 }
             }
 
-            private (string Id, BlittableJsonReaderObject Doc) GetIdAndDocFromFirstArg(JsValue firstArgs, string signature)
+            private (string Id, BlittableJsonReaderObject Doc) GetIdAndDocFromArg(JsValue firstArgs, string signature)
             {
                 if (firstArgs.IsObject() && firstArgs.AsObject() is BlittableObjectInstance doc)
                     return (doc.DocumentId, doc.Blittable);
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -208,6 +211,10 @@ public SingleRun(DocumentDatabase database, RavenConfiguration configuration, Sc
 
                 ScriptEngine.SetValue("scalarToRawString", new ClrFunctionInstance(ScriptEngine, "scalarToRawString", ScalarToRawString));
 
+                //TimeSeries
+                ScriptEngine.SetValue("appendTs", new ClrFunctionInstance(ScriptEngine, "appendTs", AppendTimeSeries));
```
ppekrol: appendTimeSeries

src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -208,6 +211,10 @@ public SingleRun(DocumentDatabase database, RavenConfiguration configuration, Sc
...
+                ScriptEngine.SetValue("deleteRangeTs", new ClrFunctionInstance(ScriptEngine, "deleteRangeTs", DeleteRangeTimeSeries));
```
ppekrol: deleteTimeSeries

src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -208,6 +211,10 @@ public SingleRun(DocumentDatabase database, RavenConfiguration configuration, Sc
...
+                ScriptEngine.SetValue("getRangeTs", new ClrFunctionInstance(ScriptEngine, "getRangeTs", GetRangeTimeSeries));
```
ppekrol: to match counters it should be called `timeSeries`. Maybe we can support both `timeSeries` and `timeseries`.

src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -230,6 +237,174 @@ public SingleRun(DocumentDatabase database, RavenConfiguration configuration, Sc
...
+                const int requiredArgs = 5;
+                
+                if (args.Length != requiredArgs)
+                    throw new ArgumentException($"{signature}: This method requires {requiredArgs} arguments but was called with {args.Length}");
+                
+                var (id, doc) = GetIdAndDocFromArg(args[0], signature);
+
+                string timeseries = GetStringArg(args[1], signature, "timeseries");
+                var timestamp = GetDateArg(args[2], signature, "timestamp");
+                var tag = GetStringArg(args[3], signature, "tag");
```
ppekrol: This is optional. @karmeli87 @aviv86 please work together on signatures. We have similar issue in the client API.

src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -230,6 +237,174 @@ public SingleRun(DocumentDatabase database, RavenConfiguration configuration, Sc
...
+                        CollectionName.GetCollectionName(doc),
+                        timeseries,
+                        new[] { toAppend});
+                }
+                finally                
+                {
+                    ArrayPool<double>.Shared.Return(values);
+                }
+                
+                return Undefined.Instance;
```
ppekrol, line 367: Why are we returning this? What should be returned here? Do we need to return anything?

src/Raven.Server/Documents/Patch/ScriptRunner.cs
ayende: `timeseriesAppend` - same for the rest

src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -230,6 +237,174 @@ public SingleRun(DocumentDatabase database, RavenConfiguration configuration, Sc
...
+                    return id;
+                }
+
+                throw new InvalidOperationException($"{signature}: 'doc' must be a string argument (the document id) or the actual document instance itself");
+            }
+            
+            private static string GetStringArg(JsValue jsArg, string signature, string argName)
+            {
+                if (jsArg.IsString() == false)
+                    throw new ArgumentException($"{signature}: The {argName} argument should be a string, but got {jsArg.GetType().Name}");
```
ayende: You are getting the .NET type of the `JsValue`, which will be `JsValue`, you need to get the Jint type here.

src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -230,6 +237,174 @@ public SingleRun(DocumentDatabase database, RavenConfiguration configuration, Sc
...
+                {
+                    DocumentId = id,
+                    Collection = CollectionName.GetCollectionName(doc),
+                    Name = timeseries,
+                    From = from,
+                    To = to,
+                };
+                _database.DocumentsStorage.TimeSeriesStorage.RemoveTimestampRange(_docsCtx, deletionRangeRequest);
+                
+                return JsValue.Undefined;
```
ppekrol, line 397: Why are we returning this? What should be returned here? Do we need to return anything?

src/Raven.Server/Documents/Patch/ScriptRunner.cs
ayende: This can be null, but `GetStringArg` doesn't allow it

src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -230,6 +237,174 @@ public SingleRun(DocumentDatabase database, RavenConfiguration configuration, Sc
...
+                if (args.Length != requiredArgs)
+                    throw new ArgumentException($"{signature}: This method requires {requiredArgs} arguments but was called with {args.Length}");
+                
+                var (id, doc) = GetIdAndDocFromArg(args[0], signature);
+
+                string timeseries = GetStringArg(args[1], signature, "timeseries");
+                var timestamp = GetDateArg(args[2], signature, "timestamp");
+                var tag = GetStringArg(args[3], signature, "tag");
+                
+                if(args[4].IsArray() == false)
```
ayende: We can probably allow to pass either `[numbers]` or `number`, so not just array

src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -230,6 +237,174 @@ public SingleRun(DocumentDatabase database, RavenConfiguration configuration, Sc
...
+                
+                if (args.Length != requiredArgs)
+                    throw new ArgumentException($"{signature}: This method requires {requiredArgs} arguments but was called with {args.Length}");
+                
+                var (id, doc) = GetIdAndDocFromArg(args[0], signature);
+
+                string timeseries = GetStringArg(args[1], signature, "timeseries");
+
+                var from = GetDateArg(args[2], signature, "from");
+                var to = GetDateArg(args[3], signature, "to");
```
ayende, line 385: Do we need to normalize the values to `/10_000` here?

src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -230,6 +237,174 @@ public SingleRun(DocumentDatabase database, RavenConfiguration configuration, Sc
...
+                if (args.Length != requiredArgs)
+                    throw new ArgumentException($"{signature}: This method requires {requiredArgs} arguments but was called with {args.Length}");
+                
+                var id = GetIdFromArg(args[0], signature);
+
+                string timeseries = GetStringArg(args[1], signature, "timeseries");
+
+                var from = GetDateArg(args[2], signature, "from");
+                var to = GetDateArg(args[3], signature, "to");
+
```
ayende, line 416: Do we need to normalize the values?

src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -230,6 +237,174 @@ public SingleRun(DocumentDatabase database, RavenConfiguration configuration, Sc
...
+                var reader = _database.DocumentsStorage.TimeSeriesStorage.GetReader(_docsCtx, id, timeseries, from, to);
+
+                var entries = new DynamicJsonArray();
```
ayende: Why no create the Jint objects directly?

ayende (changes requested):

src/Raven.Server/Documents/Patch/ScriptRunner.cs
ppekrol: This will not match e.g. counter methods where you have `incrementCounter`

src/Raven.Server/Documents/Patch/ScriptRunner.cs
haludi: As I see it this is like `void`. The `TimeSeriesStorage.AppendTimestamp` and `RemoveTimestampRange` functions return `LastDatabaseChangeVector` maybe it will be better?

ppekrol: Talk with @karmeli87 and @aviv86 what is valuable to return there. JS methods by default are returning undefined so maybe this is enough what you did.

src/Raven.Server/Documents/Patch/ScriptRunner.cs
haludi: It is because I found that both the `objectInstance` and `arrayInstance` return Object type.
So Just not lose the information I will return Jint type and .NET type, OK?

ayende: ok

src/Raven.Server/Documents/Patch/ScriptRunner.cs
karmeli87: Yes, we will do that on a separate issue: https://issues.hibernatingrhinos.com/issue/RavenDB-14651

src/Raven.Server/Documents/Patch/ScriptRunner.cs
karmeli87: I can't think of anything valuable to return here. 
So returning `void` is fine and this also something we can change in the future without breaking any changes.

src/Raven.Server/Documents/Patch/ScriptRunner.cs
karmeli87: This happened internally https://github.com/ravendb/ravendb/blob/3680f63b61405ea221dff2fe5f589fb4d94f15be/src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs#L1328

src/Raven.Server/Documents/Patch/ScriptRunner.cs
haludi: @ayende @ppekrol What should it be?

ayende: Go with @ppekrol suggestion

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
```diff
@@ -1854,16 +1844,6 @@ public long GetNumberOfTimeSeriesSegments(DocumentsOperationContext context)
             return fst.NumberOfEntries;
         }
 
-        private static void AssertNoNanValue(Reader.SingleResult toAppend)
```
ppekrol: Needs to be reverted

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
```diff
@@ -1599,17 +1594,12 @@ private static bool ShouldAddLocal(DateTime localTime, Span<double> localValues,
...
 
-        private static Reader.SingleResult GetNext(IEnumerator<Reader.SingleResult> reader, bool fromReplication)
+        private Reader.SingleResult GetNext(IEnumerator<Reader.SingleResult> reader)
         {
             Reader.SingleResult next = null;
             if (reader.MoveNext())
             {
                 next = reader.Current;
-
-                if (fromReplication == false)
```
ppekrol: Needs to be reverted

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
```diff
@@ -1316,12 +1317,6 @@ Reader.SingleResult ToResult(TimeSeriesOperation.AppendOperation element)
                         var current = appendEnumerator.Current;
                         Debug.Assert(current != null);
 
-                        if (changeVectorFromReplication == null)
```
ppekrol: Needs to be reverted

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
```diff
@@ -12,7 +12,6 @@
 using Raven.Server.Utils;
```
ppekrol: All changes in this file need to be reverted.

ppekrol (approved):

PR #10210, opened 2020-02-16
RavenDB-14590 : Include timeseries should modify the document etag

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-14590

### DIFF
--- a/src/Raven.Server/Documents/Includes/IncludeTimeSeriesCommand.cs
+++ b/src/Raven.Server/Documents/Includes/IncludeTimeSeriesCommand.cs
@@ -1,8 +1,10 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using Raven.Client.Documents.Operations.TimeSeries;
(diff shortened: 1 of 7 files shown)


src/Raven.Server/Documents/Includes/IncludeTimeSeriesCommand.cs
```diff
@@ -63,23 +63,32 @@ private TimeSeriesRangeResult GetTimeSeries(string docId, string name, DateTime
         {
             var values = new List<TimeSeriesEntry>();
             var reader = _database.DocumentsStorage.TimeSeriesStorage.GetReader(_context, docId, name, from, to);
+            var hc = new HashCode();
```
ayende: I wonder if we shouldn't also use `ComputeHttpEtags` here, instead.

ayende (changes requested):

ppekrol (approved):

ayende (approved):

PR #10213, opened 2020-02-17
RavenDB-14555 Moving to per core behavior

ayende:


src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -388,32 +152,90 @@ public void Dispose()
...
+                if (Interlocked.CompareExchange(ref core[i], context, null) == null)
                     return;
             }
+
+            if (core.Length < 64)
+            {
+                var newCore = new T[core.Length * 2];
+                Array.Copy(core, 0, newCore, 0, core.Length);
+                newCore[core.Length] = context;
+                if (Interlocked.CompareExchange(ref _perCoreContexts[currentProcessorId], newCore, core) == core)
```
karmeli87: I think you might leak contexts here. 
If someone put the context back, after you copied it, but before the exchange.

src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -426,27 +248,49 @@ public virtual void Dispose()
...
                 _disposed = true;
-                CleanThreadLocalState();
-                ThreadLocalCleanup.ReleaseThreadLocalState -= CleanThreadLocalState;
-                _nativeMemoryCleaner.Dispose();
-                _nativeMemoryCleaner = null;
+                _idleTimer.Dispose();
+
+                while (_globalQueue.TryDequeue(out var result))
+                {
+                    result.Dispose();
```
karmeli87, line 246: Do we need to check if in use?

src/Sparrow/Json/JsonContextPoolBase.cs
ayende: Good point, I think this is going to be really rare thing, and will be covered by the finalizer. But will think on how to resolve this.

src/Sparrow/Json/JsonContextPoolBase.cs
ayende: This is in the `Dispose`, caller is responsible for ensuring this.

ayende:
Looks good to me

PR #10222, opened 2020-02-18
RavenDB-14030 Patching View: Allow to define query options

Danielle9897:


src/Raven.Studio/typescript/commands/database/patch/patchCommand.ts
```diff
@@ -9,17 +9,25 @@ interface patchCommandOptions {
 
 class patchCommand extends commandBase {
 
-    constructor(private queryStr: string, private db: database, private options: patchCommandOptions = null) {
-        super();
+    constructor(private queryStr: string, 
+                private db: database, 
+                private queryOptions?: Raven.Client.Documents.Queries.QueryOperationOptions,
```
ml054: hmm I don't quite like that we are using query operation options here, then we have something different in model, then we pass through dto with upper case and finally we lower case keys since we send this via request paramters. So let's use patchCommandOptions

src/Raven.Studio/typescript/viewmodels/database/patch/patch.ts
```diff
@@ -121,6 +121,12 @@ class patchList {
 
 class patch extends viewModelBase {
 
+    staleIndexBehvior = ko.observable("patchStale");
```
ml054: typo

src/Raven.Studio/typescript/viewmodels/database/patch/patch.ts
```diff
@@ -121,6 +121,12 @@ class patchList {
 
 class patch extends viewModelBase {
 
+    staleIndexBehvior = ko.observable("patchStale"); 
+    staleTimeout = ko.observable<number>(60); // one minute
```
ml054: no need to comment - it tents to remain not in sync after some period in time

src/Raven.Studio/wwwroot/App/views/database/patch/patch.html
```diff
@@ -262,3 +272,36 @@ <h2>Select patch</h2>
         </div>
     </div>
 </script>
+
+<script type="text/html" id="settings-template">
+    <div class="padding">
```
ml054: as I wrote in issue, please get rid of that padding

src/Raven.Studio/wwwroot/App/views/database/patch/patch.html
```diff
@@ -262,3 +272,36 @@ <h2>Select patch</h2>
...
+        <div class="panel padding padding-xs">
+            <div class="btn-group btn-group-toggle" data-toggle="buttons">
```
ml054: Is it button group?

src/Raven.Studio/wwwroot/App/views/database/patch/patch.html
```diff
@@ -262,3 +272,36 @@ <h2>Select patch</h2>
...
+                <h4>Patching behavior for stale indexes:</h4>
```
ml054: for stale index (singular form)

src/Raven.Studio/wwwroot/Content/css/pages/patch.less
```diff
@@ -362,6 +362,13 @@
     }
 }
 
+.margin-toggle {
```
ml054: let's avoid having this - try to use standard classes like: margin, padding, margin-lg, etc. Also you created global styles in view specific style. We should put such styles nested.

ml054 (changes requested):

src/Raven.Studio/typescript/commands/database/patch/patchCommand.ts
arekpalinski: @ml054 in c# api we use `QueryOperationOptions` for that

https://github.com/ravendb/ravendb/blob/v4.2/src/Raven.Client/Documents/Queries/QueryOperationOptions.cs

src/Raven.Studio/typescript/viewmodels/database/patch/patch.ts
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/database/patch/patch.ts
Danielle9897: done

src/Raven.Studio/typescript/commands/database/patch/patchCommand.ts
Danielle9897: right - done - using patchCommandOptions

src/Raven.Studio/wwwroot/App/views/database/patch/patch.html
Danielle9897: Leaving current title - as discussed w/ @morhilai

src/Raven.Studio/wwwroot/App/views/database/patch/patch.html
Danielle9897: removed

src/Raven.Studio/wwwroot/Content/css/pages/patch.less
Danielle9897: done

src/Raven.Studio/wwwroot/App/views/database/patch/patch.html
Danielle9897: done

src/Raven.Studio/wwwroot/App/views/database/patch/patch.html
```diff
@@ -262,3 +272,42 @@ <h2>Select patch</h2>
...
+                    <input type="text" class="form-control" data-bind="durationPicker: staleTimeout, durationPickerOptions: { showDays: false, showSeconds: true }" />
+                </div>
+                <div class="margin-left margin-left-lg margin-top margin-top-sm" data-bind="visible: staleIndexBehavior() === 'timeoutDefined'">
+                    <small><i class="icon-warning"></i>  An exception is thrown when timeout has passed</small>
+                </div>
+            </div>
+        </div>
+    </div>
+    <hr class="margin-left margin-left-lg margin-right margin-right-lg">
+    <div class="padding padding-xs">
```
ml054: can we use padding here only, remove padding-xs, and then remove inner div with margin-left margin-left-lg to avid so many padding inside margin inside padding inside margin ;-)

src/Raven.Studio/wwwroot/App/views/database/query/query.html
```diff
@@ -260,9 +260,9 @@ <h2 class="on-base-background">
             <div class="control-label">Cache enabled</div>
         </div>
         <div class="col-xs-5">
-            <div class="toggle">
+            <div class="toggle flex-row">
```
ml054, line 263: I can see you started using flex. Great. I hope flexbox froggy was helpful. 

image

src/Raven.Studio/wwwroot/Content/css/pages/patch.less
```diff
@@ -362,11 +362,18 @@
...
 }
 
-
 .patch {
     .patch-table-container {
         margin-top: -30px;
     }
+    
+    .input-border {
+            border: 1px solid @gray-darker;
```
ml054: indent here is incorrect.

Danielle9897: done

src/Raven.Studio/wwwroot/App/views/database/patch/patch.html
Danielle9897: Played w/ this some more and removed some excess classes.
Now looking like this:

image
image

ml054 (approved):

PR #10227, opened 2020-02-19
RavenDB-14481: Core.RunInMemory configuration in CustomSettings wasn't respected

garayx:


ayende (approved):

test/Tests.Infrastructure/TestBase.cs
```diff
@@ -567,15 +567,18 @@ protected virtual RavenServer GetNewServer(ServerCreationOptions options = null,
...
                 configuration.Initialize();
 
                 configuration.Logs.Mode = LogMode.None;
                 configuration.Server.Name = ServerName;
-                configuration.Core.RunInMemory = options.RunInMemory;
                 configuration.Server.MaxTimeForTaskToWaitForDatabaseToLoad = new TimeSetting(60, TimeUnit.Seconds);
                 configuration.Licensing.EulaAccepted = true;
 
+                if (hasRunInMemory == false)
+                    configuration.Core.RunInMemory = options.RunInMemory;
```
ppekrol: We should forbid setting `RunInMemory` via custom settings then. Throw when this is set and point user that `ServerCreationOptions.RunInMemory` should be used to do that.

It will be better, because now someone can change CustomSettings and then do the opposite in the ServerCreationOptions.

ppekrol: Other option (easier?) is to change `ServerCreationOptions.RunInMemory` to nullable and detect that only one of them can be set.

garayx: What about DataDirectory ? it is also can be set via both ServerCreationOptions & CustomSettings ?

ppekrol: https://github.com/ravendb/ravendb/blob/93d8b29d565b2d1245a321d119ee49d9efb2ca39/test/Tests.Infrastructure/TestBase.cs#L582-L598

It is handled, I think?

garayx: But you mentioned that someone can change CustomSettings and then do the opposite in the ServerCreationOptions, it also apply to DataDirectory, so my question is do we also want to disallow to pass DataDirectory via CustomSettings  ?

ppekrol: If you will mark ServerCreationOptions.RunInMemory as nullable. You can allow to set it either via CustomSettings or options, but not both.

PR #10237, opened 2020-02-20
RavenDB-14635 Adding support for document / revisions compression

ayende:
Ready

### DIFF
--- a/src/Raven.Client/Exceptions/Commercial/LimitType.cs
+++ b/src/Raven.Client/Exceptions/Commercial/LimitType.cs
@@ -28,8 +28,8 @@ namespace Raven.Client.Exceptions.Commercial
         [Description("Encryption")]
         Encryption,
         
-        [Description("Document Compression")]
-        DocumentCompression,
+        [Description("Documents Compression")]
+        DocumentsCompression,
 
         [Description("External Replication")]
         ExternalReplication,
--- a/src/Raven.Client/Exceptions/Commercial/LimitType.cs
+++ b/src/Raven.Client/Exceptions/Commercial/LimitType.cs
@@ -27,6 +27,9 @@ namespace Raven.Client.Exceptions.Commercial
 
         [Description("Encryption")]
         Encryption,
+        
+        [Description("Document Compression")]
+        DocumentCompression,
 
         [Description("External Replication")]
         ExternalReplication,
--- a/src/Raven.Client/ServerWide/DatabaseRecord.cs
+++ b/src/Raven.Client/ServerWide/DatabaseRecord.cs
@@ -54,7 +54,7 @@ namespace Raven.Client.ServerWide
 
         public ConflictSolver ConflictSolverConfig;
 
-        public DocumentsCompressionConfiguration DocumentsCompression = new DocumentsCompressionConfiguration(compressRevisions: true);
+        public DocumentsCompressionConfiguration DocumentsCompression;
 
         public Dictionary<string, SorterDefinition> Sorters = new Dictionary<string, SorterDefinition>();
 
--- a/src/Raven.Client/ServerWide/DatabaseRecord.cs
+++ b/src/Raven.Client/ServerWide/DatabaseRecord.cs
@@ -54,6 +54,8 @@ namespace Raven.Client.ServerWide
 
         public ConflictSolver ConflictSolverConfig;
 
+        public HashSet<string> CompressedCollections = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
         public Dictionary<string, SorterDefinition> Sorters = new Dictionary<string, SorterDefinition>();
 
         public Dictionary<string, IndexDefinition> Indexes;
--- a/src/Raven.Client/ServerWide/DatabaseRecord.cs
+++ b/src/Raven.Client/ServerWide/DatabaseRecord.cs
@@ -54,7 +54,7 @@ namespace Raven.Client.ServerWide
 
         public ConflictSolver ConflictSolverConfig;
 
-        public HashSet<string> CompressedCollections = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+        public CompressionConfiguration Compression = new CompressionConfiguration(compressRevisions: true);
 
         public Dictionary<string, SorterDefinition> Sorters = new Dictionary<string, SorterDefinition>();
 
@@ -308,5 +308,49 @@ namespace Raven.Client.ServerWide
         No,
         SoftDelete,
         HardDelete
+    }
+
+    public class CompressionConfiguration
+    {
+        public string[] Collections;
+        public bool CompressRevisions;
+
+        public CompressionConfiguration()
+        {
+            
+        }
+
+        public CompressionConfiguration(bool compressRevisions, params string[] collections)
+        {
+            Collections = collections ?? throw new ArgumentNullException(nameof(collections));
+            CompressRevisions = compressRevisions;
+        }
+
+        protected bool Equals(CompressionConfiguration other)
+        {
+            return CompressRevisions == other.CompressRevisions && 
+                   Collections.SequenceEqual(other.Collections);
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (ReferenceEquals(null, obj)) return false;
+            if (ReferenceEquals(this, obj)) return true;
+            if (obj.GetType() != GetType()) return false;
+            return Equals((CompressionConfiguration)obj);
+        }
+
+        public override int GetHashCode()
+        {
+            int hash = Collections.Length;
+
+            foreach (string collection in Collections)
+            {
+                hash = 31 * hash + collection.GetHashCode();
+            }
+
+            hash = 31 * hash + CompressRevisions.GetHashCode();
+            return hash;
+        }
     }
 }
--- a/src/Raven.Client/ServerWide/DatabaseRecord.cs
+++ b/src/Raven.Client/ServerWide/DatabaseRecord.cs
@@ -328,8 +328,10 @@ namespace Raven.Client.ServerWide
 
         protected bool Equals(CompressionConfiguration other)
         {
+            var mine = new HashSet<string>(Collections,StringComparer.OrdinalIgnoreCase);
+            var them = new HashSet<string>(other.Collections, StringComparer.OrdinalIgnoreCase);
             return CompressRevisions == other.CompressRevisions && 
-                   Collections.SequenceEqual(other.Collections);
+                   mine.SetEquals(them);
         }
 
         public override bool Equals(object obj)
--- a/src/Raven.Client/ServerWide/DatabaseRecord.cs
+++ b/src/Raven.Client/ServerWide/DatabaseRecord.cs
@@ -54,7 +54,7 @@ namespace Raven.Client.ServerWide
 
         public ConflictSolver ConflictSolverConfig;
 
-        public CompressionConfiguration Compression = new CompressionConfiguration(compressRevisions: true);
+        public DocumentsCompressionConfiguration DocumentsCompression = new DocumentsCompressionConfiguration(compressRevisions: true);
 
         public Dictionary<string, SorterDefinition> Sorters = new Dictionary<string, SorterDefinition>();
 
@@ -310,23 +310,23 @@ namespace Raven.Client.ServerWide
         HardDelete
     }
 
-    public class CompressionConfiguration
+    public class DocumentsCompressionConfiguration
     {
         public string[] Collections;
         public bool CompressRevisions;
 
-        public CompressionConfiguration()
+        public DocumentsCompressionConfiguration()
         {
             
         }
 
-        public CompressionConfiguration(bool compressRevisions, params string[] collections)
+        public DocumentsCompressionConfiguration(bool compressRevisions, params string[] collections)
         {
             Collections = collections ?? throw new ArgumentNullException(nameof(collections));
             CompressRevisions = compressRevisions;
         }
 
-        protected bool Equals(CompressionConfiguration other)
+        protected bool Equals(DocumentsCompressionConfiguration other)
         {
             var mine = new HashSet<string>(Collections,StringComparer.OrdinalIgnoreCase);
             var them = new HashSet<string>(other.Collections, StringComparer.OrdinalIgnoreCase);
@@ -339,7 +339,7 @@ namespace Raven.Client.ServerWide
             if (ReferenceEquals(null, obj)) return false;
             if (ReferenceEquals(this, obj)) return true;
             if (obj.GetType() != GetType()) return false;
-            return Equals((CompressionConfiguration)obj);
+            return Equals((DocumentsCompressionConfiguration)obj);
         }
 
         public override int GetHashCode()
--- a/src/Raven.Server/Commercial/LicenseManager.cs
+++ b/src/Raven.Server/Commercial/LicenseManager.cs
@@ -1327,7 +1327,7 @@ namespace Raven.Server.Commercial
                 throw GenerateLicenseLimit(LimitType.DynamicNodeDistribution, message);
             }
 
-            if(!newLicenseStatus.HasCompression && documentCompressionCount > 0)
+            if(!newLicenseStatus.HasDocumentCompression && documentCompressionCount > 0)
             {
                 var message = GenerateDetails(documentCompressionCount, "document compression");
                 throw GenerateLicenseLimit(LimitType.DocumentCompression, message);
@@ -1458,7 +1458,7 @@ namespace Raven.Server.Commercial
             if (IsValid(out var licenseLimit) == false)
                 throw licenseLimit;
 
-            if (_licenseStatus.HasCompression)
+            if (_licenseStatus.HasDocumentCompression)
                 return;
 
             var details = $"Your current license ({_licenseStatus.Type}) does not allow document compression";
--- a/src/Raven.Server/Commercial/LicenseManager.cs
+++ b/src/Raven.Server/Commercial/LicenseManager.cs
@@ -1329,7 +1329,7 @@ namespace Raven.Server.Commercial
 
             if(!newLicenseStatus.HasDocumentsCompression && documentCompressionCount > 0)
             {
-                var message = GenerateDetails(documentCompressionCount, "document compression");
+                var message = GenerateDetails(documentCompressionCount, "documents compression");
                 throw GenerateLicenseLimit(LimitType.DocumentsCompression, message);
             }
         }
@@ -1461,7 +1461,7 @@ namespace Raven.Server.Commercial
             if (_licenseStatus.HasDocumentsCompression)
                 return;
 
-            var details = $"Your current license ({_licenseStatus.Type}) does not allow document compression";
+            var details = $"Your current license ({_licenseStatus.Type}) does not allow documents compression";
             throw GenerateLicenseLimit(LimitType.DocumentsCompression, details);
         }
 
--- a/src/Raven.Server/Commercial/LicenseManager.cs
+++ b/src/Raven.Server/Commercial/LicenseManager.cs
@@ -1191,6 +1191,7 @@ namespace Raven.Server.Commercial
             var externalReplicationCount = 0;
             var delayedExternalReplicationCount = 0;
             var pullReplicationAsHubCount = 0;
+            var documentCompressionCount = 0;
             var pullReplicationAsSinkCount = 0;
             var ravenEtlCount = 0;
             var sqlEtlCount = 0;
@@ -1221,6 +1222,10 @@ namespace Raven.Server.Commercial
                     if (databaseRecord.HubPullReplications != null &&
                         databaseRecord.HubPullReplications.Count > 0)
                         pullReplicationAsHubCount++;
+
+                    if (databaseRecord.Compression?.CompressRevisions == true ||
+                        databaseRecord.Compression?.Collections?.Length > 0)
+                        documentCompressionCount++;
 
                     if (databaseRecord.SinkPullReplications != null &&
                         databaseRecord.SinkPullReplications.Count > 0)
@@ -1320,6 +1325,12 @@ namespace Raven.Server.Commercial
             {
                 var message = GenerateDetails(dynamicNodesDistributionCount, "dynamic database distribution");
                 throw GenerateLicenseLimit(LimitType.DynamicNodeDistribution, message);
+            }
+
+            if(!newLicenseStatus.HasCompression && documentCompressionCount > 0)
+            {
+                var message = GenerateDetails(documentCompressionCount, "document compression");
+                throw GenerateLicenseLimit(LimitType.DocumentCompression, message);
             }
         }
 
@@ -1439,6 +1450,19 @@ namespace Raven.Server.Commercial
 
             var details = $"Your current license ({_licenseStatus.Type}) does not allow adding external replication";
             throw GenerateLicenseLimit(LimitType.ExternalReplication, details);
+        }
+
+
+        public void AssertCanUseDocumentCompression()
+        {
+            if (IsValid(out var licenseLimit) == false)
+                throw licenseLimit;
+
+            if (_licenseStatus.HasCompression)
+                return;
+
+            var details = $"Your current license ({_licenseStatus.Type}) does not allow document compression";
+            throw GenerateLicenseLimit(LimitType.DocumentCompression, details);
         }
 
         public void AssertCanDelayReplication()
--- a/src/Raven.Server/Commercial/LicenseManager.cs
+++ b/src/Raven.Server/Commercial/LicenseManager.cs
@@ -1327,10 +1327,10 @@ namespace Raven.Server.Commercial
                 throw GenerateLicenseLimit(LimitType.DynamicNodeDistribution, message);
             }
 
-            if(!newLicenseStatus.HasDocumentCompression && documentCompressionCount > 0)
+            if(!newLicenseStatus.HasDocumentsCompression && documentCompressionCount > 0)
             {
                 var message = GenerateDetails(documentCompressionCount, "document compression");
-                throw GenerateLicenseLimit(LimitType.DocumentCompression, message);
+                throw GenerateLicenseLimit(LimitType.DocumentsCompression, message);
             }
         }
 
@@ -1453,16 +1453,16 @@ namespace Raven.Server.Commercial
         }
 
 
-        public void AssertCanUseDocumentCompression()
+        public void AssertCanUseDocumentsCompression()
         {
             if (IsValid(out var licenseLimit) == false)
                 throw licenseLimit;
 
-            if (_licenseStatus.HasDocumentCompression)
+            if (_licenseStatus.HasDocumentsCompression)
                 return;
 
             var details = $"Your current license ({_licenseStatus.Type}) does not allow document compression";
-            throw GenerateLicenseLimit(LimitType.DocumentCompression, details);
+            throw GenerateLicenseLimit(LimitType.DocumentsCompression, details);
         }
 
         public void AssertCanDelayReplication()
--- a/src/Raven.Server/Commercial/LicenseManager.cs
+++ b/src/Raven.Server/Commercial/LicenseManager.cs
@@ -1223,8 +1223,8 @@ namespace Raven.Server.Commercial
                         databaseRecord.HubPullReplications.Count > 0)
                         pullReplicationAsHubCount++;
 
-                    if (databaseRecord.Compression?.CompressRevisions == true ||
-                        databaseRecord.Compression?.Collections?.Length > 0)
+                    if (databaseRecord.DocumentsCompression?.CompressRevisions == true ||
+                        databaseRecord.DocumentsCompression?.Collections?.Length > 0)
                         documentCompressionCount++;
 
                     if (databaseRecord.SinkPullReplications != null &&
--- a/src/Raven.Server/Commercial/LicenseStatus.cs
+++ b/src/Raven.Server/Commercial/LicenseStatus.cs
@@ -105,9 +105,9 @@ namespace Raven.Server.Commercial
 
         public bool HasEncryption => GetValue<bool>("encryption");
 
-        public bool HasCompression =>
+        public bool HasDocumentCompression =>
             // TODO: we don't have compression in any license yet
-            GetValue<bool?>("compression") ??
+            GetValue<bool?>("documentsCompression") ??
             GetValue<bool>("encryption");
 
         public bool HasExternalReplication => GetValue<bool>("externalReplication");
--- a/src/Raven.Server/Commercial/LicenseStatus.cs
+++ b/src/Raven.Server/Commercial/LicenseStatus.cs
@@ -105,7 +105,7 @@ namespace Raven.Server.Commercial
 
         public bool HasEncryption => GetValue<bool>("encryption");
 
-        public bool HasDocumentCompression =>
+        public bool HasDocumentsCompression =>
             // TODO: we don't have compression in any license yet
             GetValue<bool?>("documentsCompression") ??
             GetValue<bool>("encryption");
--- a/src/Raven.Server/Commercial/LicenseStatus.cs
+++ b/src/Raven.Server/Commercial/LicenseStatus.cs
@@ -105,6 +105,11 @@ namespace Raven.Server.Commercial
 
         public bool HasEncryption => GetValue<bool>("encryption");
 
+        public bool HasCompression =>
+            // TODO: we don't have compression in any license yet
+            GetValue<bool?>("compression") ??
+            GetValue<bool>("encryption");
+
         public bool HasExternalReplication => GetValue<bool>("externalReplication");
 
         public bool HasDelayedExternalReplication => GetValue<bool>("delayedExternalReplication");
--- a/src/Raven.Server/Commercial/LicenseValidator.cs
+++ b/src/Raven.Server/Commercial/LicenseValidator.cs
@@ -17,7 +17,7 @@ namespace Raven.Server.Commercial
             "distributedCluster", "maxClusterSize", "snapshotBackup", "cloudBackup", "dynamicNodesDistribution",
             "externalReplication", "delayedExternalReplication", "ravenEtl", "sqlEtl", "highlyAvailableTasks",
             "pullReplicationAsHub", "pullReplicationAsSink", "encryptedBackup", "letsEncryptAutoRenewal", "cloud",
-            "compression"
+            "documentsCompression"
         };
 
         private enum ValueType : byte
--- a/src/Raven.Server/Commercial/LicenseValidator.cs
+++ b/src/Raven.Server/Commercial/LicenseValidator.cs
@@ -16,7 +16,8 @@ namespace Raven.Server.Commercial
             "type", "version", "expiration", "memory", "cores", "redist", "encryption", "snmp",
             "distributedCluster", "maxClusterSize", "snapshotBackup", "cloudBackup", "dynamicNodesDistribution",
             "externalReplication", "delayedExternalReplication", "ravenEtl", "sqlEtl", "highlyAvailableTasks",
-            "pullReplicationAsHub", "pullReplicationAsSink", "encryptedBackup", "letsEncryptAutoRenewal", "cloud"
+            "pullReplicationAsHub", "pullReplicationAsSink", "encryptedBackup", "letsEncryptAutoRenewal", "cloud",
+            "compression"
         };
 
         private enum ValueType : byte
--- a/src/Raven.Server/Config/Categories/DatabaseConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/DatabaseConfiguration.cs
@@ -45,6 +45,16 @@ namespace Raven.Server.Config.Categories
         public TimeSetting CollectionOperationTimeout { get; set; }
 
         /// <summary>
+        /// Whatever revisions compression should be on by default or not. This is useful to turn off if you are expected
+        /// to run on very low end hardware.
+        /// This does not prevent you from enabling this after the fact, only sets the default
+        /// </summary>
+        [Description("Whatever revisions compression should be on by default or not on new databases")]
+        [DefaultValue(true)]
+        [ConfigurationEntry("Databases.DefaultRevisionsCompression", ConfigurationEntryScope.ServerWideOnly)]
+        public bool DefaultRevisionsCompression { get; set; }
+
+        /// <summary>
         /// This much time has to wait for the database to become available when too much
         /// different databases get loaded at the same time
         /// </summary>
--- a/src/Raven.Server/Config/Categories/DatabaseConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/DatabaseConfiguration.cs
@@ -51,8 +51,8 @@ namespace Raven.Server.Config.Categories
         /// </summary>
         [Description("Whatever revisions compression should be on by default or not on new databases")]
         [DefaultValue(true)]
-        [ConfigurationEntry("Databases.DefaultRevisionsCompression", ConfigurationEntryScope.ServerWideOnly)]
-        public bool DefaultRevisionsCompression { get; set; }
+        [ConfigurationEntry("Databases.Compression.CompressRevisionsDefault", ConfigurationEntryScope.ServerWideOnly)]
+        public bool CompressRevisionsDefault { get; set; }
 
         /// <summary>
         /// This much time has to wait for the database to become available when too much
--- a/src/Raven.Server/Documents/CompactDatabaseTask.cs
+++ b/src/Raven.Server/Documents/CompactDatabaseTask.cs
@@ -91,6 +91,7 @@ namespace Raven.Server.Documents
                         configuration.Storage.TempPath = new PathSetting(compactTempDirectory);
                     }
 
+                    var revisionsPrefix = CollectionName.GetTablePrefix(CollectionTableType.Revisions);
                     var compressedCollectionsTableNames = databaseRecord.CompressedCollections
                         .Select(name => new CollectionName(name).GetTableName(CollectionTableType.Documents))
                         .ToHashSet(StringComparer.OrdinalIgnoreCase);
@@ -113,7 +114,9 @@ namespace Raven.Server.Documents
                             onProgress?.Invoke(result.Progress);
                         }, (name, schema) =>
                         {
-                            schema.Compressed = compressedCollectionsTableNames.Contains(name);
+                            schema.Compressed = 
+                                name.StartsWith(revisionsPrefix,StringComparison.OrdinalIgnoreCase) ||
+                                compressedCollectionsTableNames.Contains(name);
                         },_token);
                     }
 
--- a/src/Raven.Server/Documents/CompactDatabaseTask.cs
+++ b/src/Raven.Server/Documents/CompactDatabaseTask.cs
@@ -92,7 +92,7 @@ namespace Raven.Server.Documents
                     }
 
                     var revisionsPrefix = CollectionName.GetTablePrefix(CollectionTableType.Revisions);
-                    var compressedCollectionsTableNames = databaseRecord.CompressedCollections
+                    var compressedCollectionsTableNames = databaseRecord.Compression.Collections
                         .Select(name => new CollectionName(name).GetTableName(CollectionTableType.Documents))
                         .ToHashSet(StringComparer.OrdinalIgnoreCase);
 
--- a/src/Raven.Server/Documents/CompactDatabaseTask.cs
+++ b/src/Raven.Server/Documents/CompactDatabaseTask.cs
@@ -114,8 +114,9 @@ namespace Raven.Server.Documents
                             onProgress?.Invoke(result.Progress);
                         }, (name, schema) =>
                         {
+                            bool isRevision = name.StartsWith(revisionsPrefix,StringComparison.OrdinalIgnoreCase);
                             schema.Compressed = 
-                                name.StartsWith(revisionsPrefix,StringComparison.OrdinalIgnoreCase) ||
+                                (isRevision && databaseRecord.Compression?.CompressRevisions == true) ||
                                 compressedCollectionsTableNames.Contains(name);
                         },_token);
                     }
--- a/src/Raven.Server/Documents/CompactDatabaseTask.cs
+++ b/src/Raven.Server/Documents/CompactDatabaseTask.cs
@@ -92,7 +92,7 @@ namespace Raven.Server.Documents
                     }
 
                     var revisionsPrefix = CollectionName.GetTablePrefix(CollectionTableType.Revisions);
-                    var compressedCollectionsTableNames = databaseRecord.Compression.Collections
+                    var compressedCollectionsTableNames = databaseRecord.DocumentsCompression.Collections
                         .Select(name => new CollectionName(name).GetTableName(CollectionTableType.Documents))
                         .ToHashSet(StringComparer.OrdinalIgnoreCase);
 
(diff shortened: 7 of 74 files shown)


ayende:
Initial results, here is the sample data revisions without compression:

image

As you can see, we have 3.65 MB of table data. 

Here is the results with compression:

image

So total of 1.67 MB table data, so we are only taking 45% of the previous size.

As we have more revisions, we seems to get better rate, too.

ayende:
With compression:

image


Without compression:

image

We got it down by 20% !

ayende:
This is a big change, and it isn't completely done. Still need to handle recovery here.
But it is ready for review.

topalach:
retest this please

topalach:
The previous commit was added to trigger jenkins PR tests.

src/Raven.Server/Commercial/LicenseStatus.cs
```diff
@@ -105,6 +105,11 @@ public int MaxClusterSize
 
         public bool HasEncryption => GetValue<bool>("encryption");
 
+        public bool HasCompression =>
```
grisha-kotler: We'll add the compression to all the licenses.
You also need to update `LicenseValidator`

ayende: done

src/Raven.Client/ServerWide/DatabaseRecord.cs
```diff
@@ -307,4 +309,48 @@ public enum DeletionInProgressStatus
         SoftDelete,
         HardDelete
     }
+
+    public class CompressionConfiguration
+    {
+        public string[] Collections;
+        public bool CompressRevisions;
```
arekpalinski, line 316: Given that we have an option to control compression of documents per collection then why not to let user control revisions per specific collection as well (with the global option for all revisions of course)?

src/Raven.Client/ServerWide/DatabaseRecord.cs
```diff
@@ -307,4 +309,48 @@ public enum DeletionInProgressStatus
...
+        public CompressionConfiguration(bool compressRevisions, params string[] collections)
+        {
+            Collections = collections ?? throw new ArgumentNullException(nameof(collections));
+            CompressRevisions = compressRevisions;
+        }
+
+        protected bool Equals(CompressionConfiguration other)
+        {
+            return CompressRevisions == other.CompressRevisions && 
+                   Collections.SequenceEqual(other.Collections);
```
arekpalinski: The order of elements shouldn't matter I think

src/Raven.Server/Commercial/LicenseStatus.cs
```diff
@@ -105,6 +105,11 @@ public int MaxClusterSize
...
+            // TODO: we don't have compression in any license yet
+            GetValue<bool?>("compression") ??
+            GetValue<bool>("encryption");
```
arekpalinski, line 111: Why do we default to "encryption"?

src/Raven.Server/Documents/CompactDatabaseTask.cs
```diff
@@ -103,7 +112,12 @@ public async Task Execute(Action<IOperationProgress> onProgress, CompactionResul
                             result.Progress.GlobalTotal = progressReport.GlobalTotal;
                             result.AddMessage(progressReport.Message);
                             onProgress?.Invoke(result.Progress);
-                        }, _token);
+                        }, (name, schema) =>
+                        {
+                            schema.Compressed = 
+                                name.StartsWith(revisionsPrefix,StringComparison.OrdinalIgnoreCase) ||
```
arekpalinski: This sets `Compressed` to `true` for _any_ revision table - regardless `public bool CompressRevisions;` configuration option

src/Raven.Server/Documents/DocumentPutAction.cs
```diff
@@ -219,6 +221,36 @@ public DocumentPutAction(DocumentsStorage documentsStorage, DocumentDatabase doc
...
+        private CompressionConfiguration _compression = new CompressionConfiguration(false, Array.Empty<string>());
+        private HashSet<string> _compressedCollections = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
+        public TableSchema GetDocsSchemaForCollection(CollectionName collection)
+        {
+            if(_compressedCollections.Contains(collection.Name))
+            {
+                return CompressedDocsSchema;
+            } 
+            return DocsSchema;
```
arekpalinski, line 251: Given that the schema is different - do we need to prevent from changing the configuration once the compression was turned on for a given collection?

src/Raven.Server/Raven.Server.csproj
```diff
@@ -77,7 +77,25 @@
...
+    <None Include="..\..\libs\libzstd\libzstd.arm.32.so" Link="libzstd.arm.32.so">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Include="..\..\libs\libzstd\libzstd.arm.64.so" Link="libzstd.arm.64.so">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <!--
+    <None Include="..\..\libs\libzstd\libzstd.mac.x64.dylib" Link="libzstd.mac.x64.dylib">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>-->
```
arekpalinski: Why commented?

src/Raven.Client/Exceptions/Commercial/LimitType.cs
```diff
@@ -27,6 +27,9 @@ public enum LimitType
 
         [Description("Encryption")]
         Encryption,
+        
+        [Description("Document Compression")]
+        DocumentCompression,
```
ppekrol: What is the official name for this? Compression or Document Compression?

cc @Danielle9897

src/Raven.Server/Commercial/LicenseManager.cs
```diff
@@ -1320,6 +1325,12 @@ private void ThrowIfCannotActivateLicense(LicenseStatus newLicenseStatus)
                 var message = GenerateDetails(dynamicNodesDistributionCount, "dynamic database distribution");
                 throw GenerateLicenseLimit(LimitType.DynamicNodeDistribution, message);
             }
+
+            if(!newLicenseStatus.HasCompression && documentCompressionCount > 0)
```
ppekrol: Let's stick either to Compression (too generic?) or Document Compression.

If we will go with 'DocumentCompression' then let's change this to `newLicenseStatus.HasDocumentCompression`

src/Raven.Server/Commercial/LicenseStatus.cs
```diff
@@ -105,6 +105,11 @@ public int MaxClusterSize
 
         public bool HasEncryption => GetValue<bool>("encryption");
 
+        public bool HasCompression =>
+            // TODO: we don't have compression in any license yet
+            GetValue<bool?>("compression") ??
```
ppekrol: "documentCompression" ?

src/Raven.Server/Commercial/LicenseValidator.cs
```diff
@@ -16,7 +16,8 @@ public static class LicenseValidator
             "type", "version", "expiration", "memory", "cores", "redist", "encryption", "snmp",
             "distributedCluster", "maxClusterSize", "snapshotBackup", "cloudBackup", "dynamicNodesDistribution",
             "externalReplication", "delayedExternalReplication", "ravenEtl", "sqlEtl", "highlyAvailableTasks",
-            "pullReplicationAsHub", "pullReplicationAsSink", "encryptedBackup", "letsEncryptAutoRenewal", "cloud"
+            "pullReplicationAsHub", "pullReplicationAsSink", "encryptedBackup", "letsEncryptAutoRenewal", "cloud",
+            "compression"
```
ppekrol: "documentCompression" ?

src/Raven.Server/Config/Categories/DatabaseConfiguration.cs
```diff
@@ -43,6 +43,16 @@ public class DatabaseConfiguration : ConfigurationCategory
...
 
+        /// <summary>
+        /// Whatever revisions compression should be on by default or not. This is useful to turn off if you are expected
+        /// to run on very low end hardware.
+        /// This does not prevent you from enabling this after the fact, only sets the default
+        /// </summary>
+        [Description("Whatever revisions compression should be on by default or not on new databases")]
+        [DefaultValue(true)]
+        [ConfigurationEntry("Databases.DefaultRevisionsCompression", ConfigurationEntryScope.ServerWideOnly)]
+        public bool DefaultRevisionsCompression { get; set; }
```
ppekrol: I think this feature deserves dedicated configuration section. CompressionConfiguration with `Compression.ABC` keys.

src/Raven.Server/Documents/DocumentPutAction.cs
```diff
@@ -219,6 +221,36 @@ public DocumentPutAction(DocumentsStorage documentsStorage, DocumentDatabase doc
             }
         }
 
+        public void InitializeFromDatabaseRecord(DatabaseRecord record)
+        {
+            if (_compression.Equals(record.Compression))
+                return;
+
+            if (record.Compression == null) // legacy configurations
```
ppekrol: Will this be `null`? It is initialized in `DatabaseRecord` by default.

src/Raven.Server/Raven.Server.csproj
```diff
@@ -77,7 +77,25 @@
...
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Include="..\..\libs\libzstd\libzstd.arm.32.so" Link="libzstd.arm.32.so">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Include="..\..\libs\libzstd\libzstd.arm.64.so" Link="libzstd.arm.64.so">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <!--
+    <None Include="..\..\libs\libzstd\libzstd.mac.x64.dylib" Link="libzstd.mac.x64.dylib">
```
ppekrol: Do we have a issue for this?

src/Raven.Server/Storage/Schema/SchemaUpgrader.cs
```diff
@@ -16,7 +16,7 @@ internal class CurrentVersion
 
             public const int ConfigurationVersion = 11;
 
-            public const int DocumentsVersion = 17;
+            public const int DocumentsVersion = 18;
```
ppekrol: We need to apply https://github.com/ravendb/ravendb/pull/9300

src/Raven.Server/Utils/DebuggerAttachedTimeout.cs
```diff
@@ -5,7 +5,7 @@ namespace Raven.Server.Utils
 {
     public static class DebuggerAttachedTimeout
     {
-        public static bool DisableLongTimespan;
+        public static bool DisableLongTimespan = true;
```
ppekrol, line 8: Remove?

src/Raven.Server/Web/Studio/SampleDataHandler.cs
```diff
@@ -23,6 +23,7 @@ public async Task PostCreateSampleData()
         {
             using (ContextPool.AllocateOperationContext(out DocumentsOperationContext context))
             {
+                var skipIndexes = GetBoolValueQueryString("skipIndexes", false) ?? false;
```
ppekrol, line 26: Why this is needed?

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -132,6 +134,12 @@ public async Task AddDatabaseNode()
                 if (databaseRecord.Encrypted)
                     ServerStore.LicenseManager.AssertCanCreateEncryptedDatabase();
 
+                if (databaseRecord.Compression?.CompressRevisions == true ||
```
ppekrol: Aren't we setting this value in ctor by default to true?

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -241,9 +249,27 @@ public async Task Put()
...
                 if (databaseRecord.Encrypted)
                     ServerStore.LicenseManager.AssertCanCreateEncryptedDatabase();
-                
+
+                if (databaseRecord.Compression?.CompressRevisions == true ||
+                    databaseRecord.Compression?.Collections?.Length > 0)
+                {
+                    ServerStore.LicenseManager.AssertCanUseDocumentCompression();
+                    if (Server.Configuration.Core.FeaturesAvailability != FeaturesAvailability.Experimental)
+                        FeaturesAvailabilityException.Throw("Compression");
```
ppekrol: Compression vs Document Compression

src/Raven.Server/Config/Categories/DatabaseConfiguration.cs
ppekrol: Maybe we should go with Compression name, similar to Revisions?

src/Voron/Data/Tables/Table.cs
```diff
@@ -174,21 +193,72 @@ private bool TryFindIdFromPrimaryKey(Slice key, out long id)
...
+        {
+            _cachedDecompressedBuffersByStorageId ??= new Dictionary<long, ByteString>();
+
+            if (_cachedDecompressedBuffersByStorageId.TryGetValue(id, out var t))
+            {
+                size = t.Length;
+                return t.Ptr;
+            }
+
+            //TODO: For encrypted databases, we need to allocate on locked memory
```
ppekrol: TODO. Do we have an issue for this?

src/Voron/Data/Tables/Table.cs
```diff
@@ -333,13 +417,25 @@ public void Delete(long id)
         {
             AssertWritableTable();
 
-            var ptr = DirectRead(id, out int size);
-            if (ptr == null)
-                return;
+            var ptr = DirectReadRaw(id, out int size, out bool compressed);
```
ppekrol, line 418: Can this be null?

test/FastTests/Server/Documents/Indexing/Static/BasicStaticMapIndexing.cs
```diff
@@ -500,10 +500,10 @@ public void NumberOfDocumentsAndTombstonesToProcessShouldBeCalculatedCorrectly()
         }
 
         [Theory]
-        [InlineData(200, 1000)]
-        [InlineData(1000, 2000)]
+      //  [InlineData(200, 1000)]
+      //  [InlineData(1000, 2000)]
         [InlineData(128, 2048)]
-        [InlineData(null, 1000)]
+    //    [InlineData(null, 1000)]
```
arekpalinski: Why is it commented?

src/Voron/Data/Tables/Table.cs
```diff
@@ -704,59 +921,56 @@ internal FixedSizeTree GetFixedSizeTree(Tree parent, Slice name, ushort valSize,
             return tree;
         }
 
-        private long AllocateFromSmallActiveSection(TableValueBuilder builder, int size)
+        private void CreateNewActiveSection()
         {
-            if (ActiveDataSmallSection.TryAllocate(size, out long id) == false)
+            ushort maxSectionSizeInPages =
```
ppekrol, line 935: Something that came to my mind. We need to check if moving between 32 and 64 bits work.

arekpalinski (commented):
This is not my final review. I'll go over the compression code once again.

src/Voron/Data/Tables/TableSchema.cs
```diff
@@ -27,8 +27,13 @@ public unsafe class TableSchema
         public static readonly Slice StatsSlice;
         public static readonly Slice SchemasSlice;
         public static readonly Slice PkSlice;
+        public static readonly Slice DictionariesSlice;
```
ppekrol: CompressionDictionariesSlice

src/Voron/Data/Tables/TableSchema.cs
```diff
@@ -27,8 +27,13 @@ public unsafe class TableSchema
...
+        public static readonly Slice CompressionDictionaryIdSlice;
```
ppekrol: CurrentCompressionDictionaryIdSlice?

src/Voron/Data/Tables/TableValueCompressor.cs
```diff
@@ -0,0 +1,401 @@
...
+            return count;
+        }
+
+
+        private void MaybeTrainCompressionDictionary(Table table, FixedSizeTree etagsTree)
+        {
+            // the idea is that we'll get better results by including the most recently modified documents
+            // by iterating over the tag index, which is guaranteed to be always increasing
+            var dataIds = new List<long>(256);
+            var sizes = new UIntPtr[256];
```
ppekrol: Should we use ArrayPool here?

src/Voron/Impl/Backup/FullBackup.cs
```diff
@@ -167,6 +167,32 @@ public class StorageEnvironmentInformation
 
                 try
                 {
+                    if (env.Options is StorageEnvironmentOptions.DirectoryStorageEnvironmentOptions)
+                    {
+                        foreach (var recoveryFile in Directory.GetFiles(env.Options.BasePath.FullPath, "*.Recovery"))
```
ppekrol: Why are we taking those files now?

src/Voron/ValueReader.cs
```diff
@@ -79,9 +79,6 @@ public int ReadLittleEndianInt32()
 
             _pos += sizeof (int);
 
-            if (!BitConverter.IsLittleEndian)
```
ppekrol, line 82: Why it was removed?

src/Raven.Client/ServerWide/DatabaseRecord.cs
ayende: The default for revisions is that they are all included, because they are rarely touched.
Can you think of a reason why to have this for only specific collections?

src/Raven.Server/Commercial/LicenseStatus.cs
ayende: There are a lot of licenses here without this flag, so we alias the encryption to allow this as well.
If the compression is set, it takes precedence.  So you can have a compression false, but with encryption

src/Raven.Server/Documents/DocumentPutAction.cs
ayende: What do you mean? The on disk schema is identical on both ends.

src/Raven.Server/Raven.Server.csproj
ayende: I don't have access to a mac to build and test it yet

test/FastTests/Server/Documents/Indexing/Static/BasicStaticMapIndexing.cs
ayende: Should restore it

src/Raven.Client/Exceptions/Commercial/LimitType.cs
ayende: I think we want to call it document compression

src/Raven.Server/Documents/DocumentPutAction.cs
ayende: Existing database records don't have this.

src/Raven.Server/Raven.Server.csproj
ayende: https://issues.hibernatingrhinos.com/issue/RavenDB-14840

src/Raven.Server/Utils/DebuggerAttachedTimeout.cs
ayende: I think we want to keep it.
We need it enabled only in rare cases, but it hurts usability on debugging

src/Raven.Server/Web/Studio/SampleDataHandler.cs
ayende: I want to be able to create sample database without creating indexes.
Useful for tests where we test compression only

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
ayende: When reading from old records, it it still null

src/Voron/Data/Tables/Table.cs
ayende: No, we need to remove it. This is handled by wiping the allocator memory

src/Voron/Data/Tables/Table.cs
ayende: No, this must return a value.

src/Voron/Data/Tables/Table.cs
ayende: It *shouldn't* matter, but we are taking more memory in this case, I think

src/Voron/Impl/Backup/FullBackup.cs
ayende: So recovery of database snapshot could work

src/Voron/ValueReader.cs
ayende: Gave wrong results

src/Raven.Client/ServerWide/DatabaseRecord.cs
arekpalinski: Actually I was rather thinking about excluding some revisions from compression. For example a user might have heavy subscription on revisions from a particular collection, so faster access time and lower CPU usage can be preferred in that case.

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -122,25 +128,10 @@ full name
...
-            });
-            DocsSchema.DefineFixedSizeIndex(new TableSchema.FixedSizeSchemaIndexDef
-            {
-                StartIndex = (int)DocumentsTable.Etag,
-                IsGlobal = true,
-                Name = AllDocsEtagsSlice
-            });
+            DefineIndexesForDocsSchema(DocsSchema);
+            DefineIndexesForDocsSchema(CompressedDocsSchema);
+            DocsSchema.CompressValues(CompressedDocsSchema.FixedSizeIndexes[CollectionEtagsSlice], compress: false);
```
arekpalinski:
```suggestion
            DocsSchema.CompressValues(DocsSchema.FixedSizeIndexes[CollectionEtagsSlice], compress: false);
```

Minor but shouldn't be this way?

src/Raven.Server/Documents/DocumentPutAction.cs
arekpalinski: Okay, I see. Anyway, it is valid to turn off compression if it was already enabled?

src/Voron/Data/RawData/ActiveRawDataSmallSection.cs
```diff
@@ -22,9 +22,22 @@ namespace Voron.Data.RawData
...
-            : base(tx, pageNumber)
+        private readonly Transaction _transaction;
+
+        public ActiveRawDataSmallSection(Transaction tx, long pageNumber)
+            : base(tx.LowLevelTransaction, pageNumber)
         {
+            _transaction = tx;
+        }
+
+        public static int CompressionDictionaryHashFor(LowLevelTransaction tx,long id)
```
arekpalinski: Where do we use it? (I can't find any usage on github)

src/Voron/Data/Tables/Table.cs
```diff
@@ -174,21 +193,72 @@ private bool TryFindIdFromPrimaryKey(Slice key, out long id)
...
+            var dicId = BlittableJsonReaderBase.ReadVariableSizeIntInReverse(ptr, size-1, out var offset);
+            var data = new ReadOnlySpan<byte>(ptr, size - offset);
+            var dictionary = tx.LowLevelTransaction.Environment.CompressionDictionariesHolder
+                .GetCompressionDictionaryFor(tx, dicId);
+    
+            int decompressedSize = ZstdLib.GetDecompressedSize(data);
+            var internalScope = tx.Allocator.Allocate(decompressedSize, out buffer);
+            var actualSize = ZstdLib.Decompress(data, buffer.ToSpan(), dictionary);
+            if (actualSize != decompressedSize)
+                throw new InvalidDataException($"Got decompressed size {actualSize} but expected {decompressedSize}");
```
arekpalinski: Can we add more details here?
- transaction id
- dictionary id
- can we include the entire dictionary? would it make sense?

I'm trying to collect as much details as possible in case we'll need that for troubleshooting

src/Voron/Data/RawData/RawDataSection.cs
```diff
@@ -24,16 +27,47 @@ public unsafe class RawDataSection : IDisposable
...
+            }
+
+            public short UsedSize
+            {
+                get
+                {
+                    Debug.Assert(IsFreed == false, "Should not try to get size of freed instance");
+                    return (short)(UsedSize_Buffer & ValueOnlyMask);
+                }
+                set => UsedSize_Buffer = (short)(value & ValueOnlyMask);
```
arekpalinski, line 61: It's meant to be called only for non-compressed data, right?

src/Voron/Data/Tables/Table.cs
```diff
@@ -174,21 +193,72 @@ private bool TryFindIdFromPrimaryKey(Slice key, out long id)
...
+
+            if (_cachedDecompressedBuffersByStorageId.TryGetValue(id, out var t))
+            {
+                size = t.Length;
+                return t.Ptr;
+            }
+
+            //TODO: For encrypted databases, we need to allocate on locked memory
+
+            // we explicitly do *not* dispose the buffer, it lives as long as the tx
```
arekpalinski, line 235: This means that for long running transactions we might have a lot of cached values even if we don't need them anymore. Maybe we should introduce LRU cache and clear old values if the cache grows too much?

src/Voron/Data/Tables/Table.cs
```diff
@@ -389,15 +490,32 @@ public void Delete(long id)
...
 
-                if (ActiveDataSmallSection.TryWriteDirect(newId, itemSize, out byte* writePos) == false)
+                if (ActiveDataSmallSection.TryWriteDirect(newId, itemSize, compressed, out byte* writePos) == false)
                     throw new VoronErrorException($"Cannot write to newly allocated size in {Name} during delete");
 
                 Memory.Copy(writePos, pos, itemSize);
+
+                OnDataMoved(idToMove, newId, dataPtr, dataSize);
+
+                decompressedScope.Dispose();
```
arekpalinski: Shouldn't the dispose be in inside `finally`?

src/Voron/Data/Tables/Table.cs
```diff
@@ -333,13 +417,25 @@ public void Delete(long id)
...
+            {
+                decompressValue = DecompressValue(_tx, ptr, size, out var buffer);
+                ptr = buffer.Ptr;
+                size = buffer.Length;
+            }
 
             var tvr = new TableValueReader(ptr, size);
             DeleteValueFromIndex(id, ref tvr);
 
+            decompressValue.Dispose();
```
arekpalinski, line 435: Shouldn't be inside `finally` or just wrap by `using`?

src/Voron/Data/Tables/Table.cs
```diff
@@ -518,7 +629,102 @@ public long Insert(TableValueBuilder builder)
...
+            if (ActiveDataSmallSection.TryAllocate(itemSize, out id) == false)
+                ThrowBadAllocation(itemSize);
+
+            return id;
+        }
+
+        private void ThrowBadWriter(int size)
+        {
+            throw new VoronErrorException(
+                $"After successfully allocating {size:#,#;;0} bytes, failed to write them on {Name}");
```
arekpalinski: Can we add more details here? Basically all parameters that we pass to `TryWriteDirect`:
- id
- size
- compressed

src/Voron/Data/Tables/Table.cs
arekpalinski: This look _very_ familiar to the issue we had with `NewPageAllocator` and we ended up with persisting the actual allocation size - see here:

https://github.com/ravendb/ravendb/pull/9610

src/Voron/Data/Tables/Table.cs
```diff
@@ -704,59 +921,56 @@ internal FixedSizeTree GetFixedSizeTree(Tree parent, Slice name, ushort valSize,
...
-                _activeDataSmallSection = ActiveRawDataSmallSection.Create(_tx.LowLevelTransaction, Name, _tableType, newNumberOfPages);
-                _activeDataSmallSection.DataMoved += OnDataMoved;
-                var val = _activeDataSmallSection.PageNumber;
-                using (Slice.External(_tx.Allocator, (byte*)&val, sizeof(long), out Slice pageNumber))
-                {
-                    _tableTree.Add(TableSchema.ActiveSectionSlice, pageNumber);
+                            ActiveCandidateSection.Delete(sectionPageNumber);
+                            {
+                                return true;
+                            }
```
arekpalinski: Wrapped by `{ }` ?

src/Voron/Data/Tables/TableValueBuilder.cs
```diff
@@ -175,5 +199,41 @@ public void CopyTo(byte* ptr)
...
+
+        public bool TryCompression(Table table, TableSchema schema)
+        {
+            var tx = table._tx;
+
+            Compression.RawScope = tx.Allocator.Allocate(RawSize, out Compression.RawBuffer);
+            CopyTo(Compression.RawBuffer.Ptr);
+
+            return Compression.TryCompression(table, schema);
+        }
```
arekpalinski, line 211: Do/should we return memory allocated here of `TryCompression` returns `false`?

src/Voron/Data/Tables/TableValueCompressor.cs
```diff
@@ -0,0 +1,401 @@
...
+        private byte GetCompressionRatio(int compressed, int raw)
+        {
+            return (byte)(compressed / (float)raw * 100);
+        }
+
+        public int WriteVariableSizeIntInReverse(byte* buffer, int value)
+        {
+            var count = 0;
+            var v = (uint)value;
+            while (v >= 0x80)
```
arekpalinski, line 130: Can we create const for `0x80`?

src/Voron/Data/Tables/TableValueCompressor.cs
```diff
@@ -0,0 +1,401 @@
...
+            {
+                buffer[count++] = (byte)(v | 0x80);
+                v >>= 7;
```
arekpalinski, line 133: Let's add comment explaining `>>= 7` ?

src/Voron/Data/Tables/TableValueCompressor.cs
```diff
@@ -0,0 +1,401 @@
...
+                        // we don't want to skip documents that are too big, they will compress
+                        // well on their own, and likely be *too* unique to add meaningfully to the
+                        // dictionary
+                        continue; 
+                    }
+
+                    sizes[dataIds.Count] = (UIntPtr)size;
+                    dataIds.Add(id);
+                    totalSize += size;
+                } while (dataIds.Count < 256 && it.MovePrev() && totalSize < 1024 * 1024);
```
arekpalinski: What is the reasoning that we iterate from the last element?

src/Voron/Data/Tables/TableValueCompressor.cs
```diff
@@ -0,0 +1,401 @@
...
+            {
+                return;
+            }
+
+            table.CurrentCompressionDictionaryId = newId;
+            compressionDictionary.ExpectedCompressionRatio = GetCompressionRatio(CompressedBuffer.Length, RawBuffer.Length);
+
+            var rev = Bits.SwapBytes(newId);
+            using var _____ = Slice.External(tx.Allocator, (byte*)&rev, sizeof(int), out var slice);
+            using var ____ = dictionariesTree.DirectAdd(slice, sizeof(CompressionDictionaryInfo) + dictionaryBufferSpan.Length, out var dest);
```
arekpalinski: This is very crucial piece of code - I'd prefer to see explicitly the scopes of `usings` in this method. It's easier to read it IMO.

src/Voron/Data/Tables/TableValueCompressor.cs
```diff
@@ -0,0 +1,401 @@
...
+                return; // no need for in mem mode
+
+            lock (obj.Environment.CompressionDictionariesHolder)
+            {
+                using var tx = obj.Environment.ReadTransaction();
+
+                var dictionaries = tx.ReadTree(TableSchema.DictionariesSlice);
+
+                if (dictionaries == null)
+                    return; // should never happen
```
arekpalinski, line 264: Debug.Assert ?

src/Voron/Data/Tables/TableValueCompressor.cs
```diff
@@ -0,0 +1,401 @@
...
+
+                // for reliability's sake, we keep two copies of the compression
+                // dictionaries
+                for (int i = 0; i < 2; i++)
+                {
+                    var newPath = obj.Environment.Options.BasePath
+                        .Combine(path: $"Compression{(i == 0 ? "A" : "B")}.Recovery")
+                        .FullPath;
+
+                    using var fs = File.Open(newPath, FileMode.OpenOrCreate, FileAccess.ReadWrite, FileShare.Read);
```
arekpalinski, line 282: Don't we need to call `fs.Flush(flushToDisk: true);` at the end?

src/Voron/Data/Tables/ZstdLib.cs
```diff
@@ -0,0 +1,279 @@
...
+                return "Dic #" + Id 
+#if DEBUG
+					 + " - " + DictionaryHash
+#endif
+
+;
+            }
+
+            ~CompressionDictionary()
+            {
```
arekpalinski, line 241: Maybe let's add some logging or just console write line in DEBUG when the finalizer is called

src/Raven.Client/ServerWide/DatabaseRecord.cs
ayende: I think this is a rather niche scenario, and even so, we aren't likely to spend too much time of this. Subscriptions are no time sensitive, I think. 
The issue I have is that we would need `CompressRevisions` and `ExcludeRevisionCompression` and that seems like very rare scenario to need it.

src/Raven.Server/Documents/DocumentPutAction.cs
ayende: Yes

src/Voron/Data/RawData/ActiveRawDataSmallSection.cs
ayende: dead code

src/Voron/Data/RawData/RawDataSection.cs
ayende: Yes, this is called, and then the is compressed is set.

src/Voron/Data/Tables/Table.cs
ayende: We can't, someone may be looking at this value.
We are accounting for that in the tx sizes.

src/Voron/Data/Tables/Table.cs
ayende: The actual dictionary is 8KB in size, and is binary.
Doesn't make sense.

It is there is we messed up the numbers, mostly.

src/Voron/Data/Tables/Table.cs
ayende: I want to avoid either options for perf costs. 
If we leak, this will be disposed by the tx closing anyway

src/Voron/Data/Tables/Table.cs
ayende: Same, this is perf sensitive, and I want to avoid the try/finally costs.
It isn't an issue if we leak, because the tx will dispose the values anway

src/Voron/Data/Tables/Table.cs
ayende: The number of pages is persisted, see inside the ActiveRawDataSection

ayende: Note that this part is just moved code, I didn't change functionality from prev behavior

src/Voron/Data/Tables/TableValueBuilder.cs
ayende: No, this is going to be cleanup anyway shortly when the operation is done.

src/Voron/Data/Tables/TableValueCompressor.cs
ayende: No, this is actually clearer when you see the constant.
Because it follows the structure of varint.

src/Voron/Data/Tables/TableValueCompressor.cs
ayende: That is explained elsewhere in the code, this is the same varint we use everywhere

src/Voron/Data/Tables/TableValueCompressor.cs
ayende: We want to get the most recently update values, we assume that those are the most similar to the currently inserted dta.

src/Voron/Data/Tables/TableValueCompressor.cs
ayende: I intentionally didn't do that. 
We have two separate files here, and I want to avoid flushing the entire fs cache
That can kill our perf

src/Raven.Client/ServerWide/DatabaseRecord.cs
arekpalinski: I'm fine with that

src/Voron/Data/Tables/Table.cs
arekpalinski: Okay, but transaction id and dictionary id might be helpful when investigating some corruption issues

src/Voron/Data/Tables/Table.cs
arekpalinski: Can we have the concept of tracking the actual usage and count it?

src/Voron/Data/Tables/Table.cs
arekpalinski: OK. Worth to comments that this is intentional because of performance

src/Voron/Data/Tables/TableValueCompressor.cs
arekpalinski: 1. But if non of the files is flushed then how can we be guaranteed that at least one have everything synced? 
2. This is called once per write transaction, right?

ayende: This is called per value saved.

ayende: Yes, sorry, once per write tx that created new dictionary
And we don't flush because:
* We have the data saved in the data file
* We re-create it on new dictionarys

src/Raven.Client/ServerWide/DatabaseRecord.cs
```diff
@@ -54,6 +54,8 @@ public DatabaseRecord(string databaseName)
 
         public ConflictSolver ConflictSolverConfig;
 
+        public CompressionConfiguration Compression = new CompressionConfiguration(compressRevisions: true);
```
ppekrol: Compression => DocumentCompression

src/Raven.Client/ServerWide/DatabaseRecord.cs
```diff
@@ -307,4 +309,50 @@ public enum DeletionInProgressStatus
         SoftDelete,
         HardDelete
     }
+
+    public class CompressionConfiguration
```
ppekrol: DocumentCompressionConfiguration

src/Raven.Server/Commercial/LicenseStatus.cs
```diff
@@ -105,6 +105,11 @@ public int MaxClusterSize
 
         public bool HasEncryption => GetValue<bool>("encryption");
 
+        public bool HasDocumentCompression =>
+            // TODO: we don't have compression in any license yet
+            GetValue<bool?>("documentsCompression") ??
```
ppekrol, line 110: documentCompression

src/Raven.Server/Commercial/LicenseValidator.cs
```diff
@@ -16,7 +16,8 @@ public static class LicenseValidator
             "type", "version", "expiration", "memory", "cores", "redist", "encryption", "snmp",
             "distributedCluster", "maxClusterSize", "snapshotBackup", "cloudBackup", "dynamicNodesDistribution",
             "externalReplication", "delayedExternalReplication", "ravenEtl", "sqlEtl", "highlyAvailableTasks",
-            "pullReplicationAsHub", "pullReplicationAsSink", "encryptedBackup", "letsEncryptAutoRenewal", "cloud"
+            "pullReplicationAsHub", "pullReplicationAsSink", "encryptedBackup", "letsEncryptAutoRenewal", "cloud",
+            "documentsCompression"
```
ppekrol, line 20: documentCompression

src/Raven.Server/Config/Categories/DatabaseConfiguration.cs
```diff
@@ -43,6 +43,16 @@ public class DatabaseConfiguration : ConfigurationCategory
...
 
+        /// <summary>
+        /// Whatever revisions compression should be on by default or not. This is useful to turn off if you are expected
+        /// to run on very low end hardware.
+        /// This does not prevent you from enabling this after the fact, only sets the default
+        /// </summary>
+        [Description("Whatever revisions compression should be on by default or not on new databases")]
+        [DefaultValue(true)]
+        [ConfigurationEntry("Databases.Compression.CompressRevisionsDefault", ConfigurationEntryScope.ServerWideOnly)]
+        public bool CompressRevisionsDefault { get; set; }
```
ppekrol, line 55: Wouldn't it be better to move this to DocumentCompressionConfiguration?

src/Raven.Server/Storage/Schema/SchemaUpgrader.cs
ppekrol: Can you rebase? It should be fixed now.

src/Raven.Server/Storage/Schema/Updates/Documents/From17.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Raven.Server.Storage.Schema.Updates.Documents
+{
+    public class From17 : ISchemaUpdate
```
ppekrol: This can be removed.

src/Raven.Server/Web/Studio/SampleDataHandler.cs
ppekrol: Do we want to expose this in the UI?

src/Voron/Slice.cs
```diff
@@ -234,6 +234,13 @@ public Span<byte> AsSpan()
         {
             return new Span<byte>(Content.Ptr, Size);
         }
+
+        public static bool Equals(Slice a, Slice b)
+        {
+            if (a.Size != b.Size)
+                return false;
+            return Memory.Compare(a.Content.Ptr, b.Content.Ptr, a.Size) == 0;
```
ppekrol: Shouldn't we use SliceComparer here?

src/Voron/Data/Tables/TableValueCompressor.cs
```diff
@@ -0,0 +1,419 @@
...
+                var nonceBuffer = stackalloc byte[nonceSize];
+                var macBuffer = stackalloc byte[macSize];
+                var subKey = stackalloc byte[subKeyLen];
+
+                // for reliability's sake, we keep two copies of the compression
+                // dictionaries
+                for (int i = 0; i < 2; i++)
+                {
+                    var newPath = obj.Environment.Options.BasePath
+                        .Combine(path: $"Compression{(i == 0 ? "A" : "B")}.Recovery")
```
ppekrol: Can we move `.Recovery` to some const and use it here and in the Recovery.cs ? Maybe also consider changing the extension to `.compression-recovery`? It is easier to recognize the files via extension than the name. Some day we might add more 'recovery' files.

ppekrol (approved):

src/Raven.Server/Web/Studio/SampleDataHandler.cs
ayende: I don't think so, it is very rarely going to be useful for users.

src/Voron/Data/Tables/Table.cs
ayende: Done

src/Voron/Data/Tables/Table.cs
ayende: done

src/Voron/Data/Tables/Table.cs
ayende: Yes, this is something that we have to do. 
https://issues.hibernatingrhinos.com/issue/RavenDB-14870

src/Raven.Client/ServerWide/DatabaseRecord.cs
ayende: Done

src/Raven.Client/ServerWide/DatabaseRecord.cs
ayende: Done

src/Raven.Server/Commercial/LicenseStatus.cs
ayende: Why singular here?

src/Raven.Server/Config/Categories/DatabaseConfiguration.cs
ayende: We have just a single value, and it is already possible to change the code structure without the user seeing it. I don't think so.

src/Raven.Server/Storage/Schema/Updates/Documents/From17.cs
ayende: done

src/Voron/Slice.cs
ayende: Removed it entirely.

src/Voron/Data/Tables/TableValueCompressor.cs
ayende: done

src/Raven.Server/Commercial/LicenseManager.cs
```diff
@@ -1321,6 +1326,12 @@ private void ThrowIfCannotActivateLicense(LicenseStatus newLicenseStatus)
                 var message = GenerateDetails(dynamicNodesDistributionCount, "dynamic database distribution");
                 throw GenerateLicenseLimit(LimitType.DynamicNodeDistribution, message);
             }
+
+            if(!newLicenseStatus.HasDocumentsCompression && documentCompressionCount > 0)
+            {
+                var message = GenerateDetails(documentCompressionCount, "document compression");
```
ppekrol: "documents compression"

src/Raven.Server/Commercial/LicenseManager.cs
```diff
@@ -1441,6 +1452,19 @@ public void AssertCanAddExternalReplication()
...
+
+        public void AssertCanUseDocumentsCompression()
+        {
+            if (IsValid(out var licenseLimit) == false)
+                throw licenseLimit;
+
+            if (_licenseStatus.HasDocumentsCompression)
+                return;
+
+            var details = $"Your current license ({_licenseStatus.Type}) does not allow document compression";
```
ppekrol: documents compression

ppekrol (approved):

arekpalinski (approved):
LGTM. Please just rebase to use zsdt libs build by @gregolsky. Also after rebase we should see multiplatform FastTest results so it will test if zsdt dlls are compiled correctly.

PR #10246, opened 2020-02-22
RavenDB-14671 - Support bulk insert for timeseries

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-14456

### DIFF
--- a/src/Raven.Client/Documents/BulkInsert/BulkInsertOperation.cs
+++ b/src/Raven.Client/Documents/BulkInsert/BulkInsertOperation.cs
@@ -9,6 +9,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
 using Raven.Client.Documents.Commands;
+using Raven.Client.Documents.Commands.Batches;
 using Raven.Client.Documents.Conventions;
 using Raven.Client.Documents.Identity;
 using Raven.Client.Documents.Operations;
@@ -141,6 +142,8 @@ namespace Raven.Client.Documents.BulkInsert
         private Stream _stream;
         private readonly StreamExposerContent _streamExposerContent;
         private bool _first = true;
+        private CommandType _previousCommandType = CommandType.None;
+        private string _previousId = null;
         private long _operationId = -1;
 
         public CompressionLevel CompressionLevel = CompressionLevel.NoCompression;
@@ -160,6 +163,8 @@ namespace Raven.Client.Documents.BulkInsert
                     {
                         try
                         {
+                            EndPreviousCommandIfNeeded();
+
                             _currentWriter.Write(']');
                             _currentWriter.Flush();
                             await _asyncWrite.ConfigureAwait(false);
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Documents/Handlers/BulkInsertHandler.cs
```diff
@@ -172,43 +201,68 @@ public class MergedInsertBulkCommand : TransactionOperationsMerger.MergedTransac
...
+                    else if (cmd.Type == CommandType.TimeSeries)
+                    {
+                        foreach (var appendOperation in cmd.TimeSeries.Appends)
+                        {
+                            var docCollection = TimeSeriesHandler.ExecuteTimeSeriesBatchCommand.GetDocumentCollection(Database, context, cmd.Id, fromEtl: false);
+                            Database.DocumentsStorage.TimeSeriesStorage.AppendTimestamp(context,
+                                cmd.Id,
+                                docCollection,
+                                appendOperation.Name,
+                                new []{ appendOperation }
```
ayende, line 255: Can we avoid this allocation? Is there a a reason we run it one at a time?

ayende (commented):
And can you add support for counters as well?

src/Raven.Client/Documents/BulkInsert/BulkInsertOperation.cs
```diff
@@ -305,76 +295,184 @@ public async Task StoreAsync(object entity, string id, IMetadataDictionary metad
...
-                        _currentWriter.BaseStream.SetLength(0);
-                        ((MemoryStream)tmp.BaseStream).TryGetBuffer(out var buffer);
-                        _asyncWrite = _requestBodyStream.WriteAsync(buffer.Array, buffer.Offset, buffer.Count, _token);
-                    }
+        public void AppendTimeSeries(string id, string timeseries, DateTime timestamp, string tag, IEnumerable<double> values)
+        {
+            AsyncHelpers.RunSync(() => AppendTimeSeriesAsync(id, timeseries, timestamp, tag, values));
+        }
+
+        public async Task AppendTimeSeriesAsync(string id, string timeseries, DateTime timestamp, string tag, IEnumerable<double> values)
```
ppekrol, line 324: Wondering if it would not be better to have 

```csharp
bulkInsert.TimeSeries.Append
bulkInsert.Counters.Increment
```

We are doing similar separation in session.

ayende: I like that. Especially true since we need to have better API so we can have:

`bulk.TimeSeriesFor(docId, name)` And then you call `Append` multiple times here.
That matches the underlying format.

ppekrol: We can either do it now or as a part of TimeSeries API revamp. @karmeli87 @aviv86 ?

karmeli87: I like Oren's suggestion and would like to have it in the `session`, `bulkInsert`, `patching` and adjust how we send stuff on the wire accordingly.
 
We will handle this in a separate issue as part of the API unification.

grisha-kotler:
Will reopen once we refactor the Time Series BatchCommand

PR #10247, opened 2020-02-22
RavenDB-14418 Enhancements for PatternForOutputReduceToCollectionReferences

arekpalinski:
https://issues.hibernatingrhinos.com/issue/RavenDB-14418

### DIFF
--- a/src/Raven.Client/Documents/Indexes/ExpressionStringBuilder.cs
+++ b/src/Raven.Client/Documents/Indexes/ExpressionStringBuilder.cs
@@ -1760,8 +1760,21 @@ namespace Raven.Client.Documents.Indexes
             {
                 var type = node.Method.GetGenericArguments()[0];
                 _loadDocumentTypes.Add(type);
-                var collection = _conventions.GetCollectionName(type);
-                Out($", \"{collection}\"");
+
+                var parameters = node.Method.GetParameters();
+                if (parameters.Length == 1)
+                {
+                    var collection = _conventions.GetCollectionName(type);
+                    Out($", \"{collection}\"");
+                }
+                else if (parameters.Length == 2)
+                {
+                    // nothing do to - collection name was provided - LoadDocument<T>("id", "CollectionName")
+                }
+                else
+                {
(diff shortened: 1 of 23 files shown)


ppekrol (approved):
LGTM, but requires Studio work.

src/Raven.Client/Documents/Indexes/ExpressionStringBuilder.cs
```diff
@@ -1760,8 +1760,21 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
             {
                 var type = node.Method.GetGenericArguments()[0];
                 _loadDocumentTypes.Add(type);
-                var collection = _conventions.GetCollectionName(type);
-                Out($", \"{collection}\"");
+
+                var parameters = node.Method.GetParameters();
+                if (parameters.Length == 1)
+                {
+                    var collection = _conventions.GetCollectionName(type);
```
ayende, line 1767: Do we need to do escaping here?

ayende (approved):

src/Raven.Client/Documents/Indexes/ExpressionStringBuilder.cs
arekpalinski: I'm not sure I follow. Did you mean one line below:
`Out($", \"{collection}\"");`
? 
We have `\"` to pass collection name as string parameter - `LoadDocument("users/1", "Users")`

arekpalinski:
@Danielle9897 will take care of the studio part in this PR

src/Raven.Client/Documents/Indexes/ExpressionStringBuilder.cs
ayende: Yes, what happens if the collection contains `"` ?

arekpalinski: If `FindCollectionName` convention was overwritten and returned collection name would include special characters then I think yes we would need to escape it here.

I opened issue to verify all usages: https://issues.hibernatingrhinos.com/issue/RavenDB-14681

PR #10253, opened 2020-02-23
RDBS-68 - fix delay in deleting expired documents

grisha-kotler:


ppekrol:
Can we have a test for this? You can return number of removed/refreshed documents from `CleanupDocs` and use it in the test.

grisha-kotler:
done

test/SlowTests/Issues/RDBS-68.cs
```diff
@@ -0,0 +1,122 @@
...
+                    DeleteFrequencyInSec = 1000
+                };
+
+                await ExpirationHelper.SetupExpiration(store, Server.ServerStore, config);
+
+                var expiry = SystemTime.UtcNow.AddMinutes(-5);
+
+                var deleteList = new List<string>();
+
+                var totalInBatch = ExpiredDocumentsCleaner.BatchSize;
```
ppekrol: Either use smaller batch size or move this to stress

grisha-kotler: that's 4096 documents in total

ppekrol: Fixed.

PR #10255, opened 2020-02-24
RDBS-68 Put revision tombstone only once

karmeli87:
otherwise we put it with a new etag every time.
normally we know to skip it by comparing with the destination change vector, but under specific conditions (like removed database) we might end up in a replication loop.


ayende:
Really interesting, might also be related to yesterday's support call

ayende (approved):

ppekrol (approved):

PR #10260, opened 2020-02-25
RavenDB-14687 Making sure that index specific Indexing.MaxStepsForScr…

arekpalinski:
…ipt setting is respected

### DIFF
--- a/test/SlowTests/Issues/RavenDB_14687.cs
+++ b/test/SlowTests/Issues/RavenDB_14687.cs
@@ -1,9 +1,10 @@
 ﻿using System.Collections.Generic;
+using System.Threading.Tasks;
 using FastTests;
 using Orders;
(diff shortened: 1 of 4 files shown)


test/SlowTests/Issues/RavenDB_14687.cs
```diff
@@ -0,0 +1,69 @@
...
+
+            public MyJSIndex()
+            {
+                Maps = new HashSet<string>()
+                {
+                    @"
+map('Companies', (company) => {
+    var x = [];
+    for (var i = 0; i < 30000; i++) {
+        x.push(i);
```
ppekrol, line 73: Let's make it simpler. Either take the instance of the index from the database directly and assert number of steps or make this JS function simpler.

PR #10269, opened 2020-02-26
RavenDB-14633

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-14633

### DIFF
--- a/src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Net;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Primitives;
 using Raven.Client;
 using Raven.Client.Documents.Changes;
 using Raven.Client.Documents.Operations.TimeSeries;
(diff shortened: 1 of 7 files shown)


src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
```diff
@@ -109,10 +112,15 @@ public Task Stats()
...
-        
-        
+
         [RavenAction("/databases/*/timeseries", "GET", AuthorizationStatus.ValidUser)]
         public Task Read()
+        {
+            return GetRanges();
+        }
+
+        [RavenAction("/databases/*/timeseries/get-ranges", "GET", AuthorizationStatus.ValidUser)]
```
ppekrol: => `/databases/*/timeseries/ranges`

Would be better?

ayende: Yes, the `GET` is in the HTTP method

src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
```diff
@@ -109,10 +112,15 @@ public Task Stats()
             
             return Task.CompletedTask;
         }
-        
-        
+
         [RavenAction("/databases/*/timeseries", "GET", AuthorizationStatus.ValidUser)]
         public Task Read()
+        {
+            return GetRanges();
```
ayende: Why have both this and the `/ranges` endpoints?

ppekrol (approved):

ayende (approved):

PR #10271, opened 2020-02-27
RavenDB-14689 incoming replication should properly use the context

karmeli87:
The issue is that we were allocating slices in one context and releasing after the context was reset.


src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -953,6 +953,9 @@ public MergedDocumentReplicationCommand(DataForReplicationCommand replicationInf
 
             protected override long ExecuteCmd(DocumentsOperationContext context)
             {
+
+                var disposables = new List<IDisposable>();
```
ppekrol: we are not disposing anything?

PR #10274, opened 2020-02-27
RavenDB-14610 fix assertion @counters is missing from metadata

karmeli87:


src/Raven.Server/Documents/Handlers/CountersHandler.cs
```diff
@@ -440,7 +440,7 @@ void LoadDocument()
                     try
                     {
                         doc = _database.DocumentsStorage.Get(context, counterGroupDetail.DocumentId,
-                            throwOnConflict: true);
+                            throwOnConflict: true, skipValidationInDebug: true);
```
ppekrol: Can you add a comment why `skipValidationInDebug` is set to 'true' in this case?

karmeli87: I guess it is a bug in the legacy backup file that had document without the correct metadata.
But it okay to set it to `true` here, since we are putting a counter and going to update the metadata.

PR #10280, opened 2020-02-27
RDBS-68 - use the same date for the DeleteDocumentsExpiration command as we do for collecting the document to delete

grisha-kotler:


src/Raven.Server/Documents/Expiration/ExpiredDocumentsCleaner.cs
```diff
@@ -234,12 +237,14 @@ public ExpiredDocumentsCleaner.DeleteExpiredDocumentsCommand ToCommand(Documents
...
+            var command = new ExpiredDocumentsCleaner.DeleteExpiredDocumentsCommand(expired, database, ForExpiration, CurrentTime);
             return command;
         }
 
+        public KeyValuePair<Slice, List<(Slice LowerId, string Id)>>[] Expired { get; set; }
+
         public bool ForExpiration { get; set; }
 
-        public KeyValuePair<Slice, List<(Slice LowerId, string Id)>>[] Expired { get; set; }
+        public DateTime CurrentTime { get; set; }
```
ayende, line 248: Who sets this value?

ayende (changes requested):

src/Raven.Server/Documents/Expiration/ExpiredDocumentsCleaner.cs
ppekrol: It is set during deserialization in ReplayCommandTxHelper

PR #10281, opened 2020-02-27
RavenDB-13545 provide custom S3 hostname for backups

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-13545

### DIFF
--- a/src/Raven.Client/Documents/Operations/Backups/BackupSettings.cs
+++ b/src/Raven.Client/Documents/Operations/Backups/BackupSettings.cs
@@ -125,10 +125,15 @@ namespace Raven.Client.Documents.Operations.Backups
     public class S3Settings : AmazonSettings
     {
         /// <summary>
-        /// Amazon S3 Bucket name.
+        /// S3 Bucket name.
         /// </summary>
         public string BucketName { get; set; }
 
+        /// <summary>
+        /// S3 server Url when using custom server
+        /// </summary>
+        public string CustomS3ServerUrl { get; set; }
+        
         public override bool HasSettings()
         {
             if (base.HasSettings())
@@ -154,6 +159,9 @@ namespace Raven.Client.Documents.Operations.Backups
             if (other.RemoteFolderName != RemoteFolderName)
                 return false;
 
+            if (other.CustomS3ServerUrl != CustomS3ServerUrl)
+                return false;
+            
             return true;
         }
 
@@ -161,6 +169,7 @@ namespace Raven.Client.Documents.Operations.Backups
         {
             var djv = base.ToJson();
             djv[nameof(BucketName)] = BucketName;
+            djv[nameof(CustomS3ServerUrl)] = CustomS3ServerUrl;
             return djv;
         }
     }
--- a/src/Raven.Client/Documents/Operations/Backups/BackupSettings.cs
+++ b/src/Raven.Client/Documents/Operations/Backups/BackupSettings.cs
@@ -132,7 +132,7 @@ namespace Raven.Client.Documents.Operations.Backups
         /// <summary>
         /// S3 server Url when using custom server
         /// </summary>
-        public string CustomS3ServerUrl { get; set; }
+        public string CustomServerUrl { get; set; }
         
         public override bool HasSettings()
         {
@@ -159,7 +159,7 @@ namespace Raven.Client.Documents.Operations.Backups
             if (other.RemoteFolderName != RemoteFolderName)
                 return false;
 
-            if (other.CustomS3ServerUrl != CustomS3ServerUrl)
+            if (other.CustomServerUrl != CustomServerUrl)
                 return false;
             
             return true;
@@ -169,7 +169,7 @@ namespace Raven.Client.Documents.Operations.Backups
         {
             var djv = base.ToJson();
             djv[nameof(BucketName)] = BucketName;
-            djv[nameof(CustomS3ServerUrl)] = CustomS3ServerUrl;
+            djv[nameof(CustomServerUrl)] = CustomServerUrl;
             return djv;
         }
     }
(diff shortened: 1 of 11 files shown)


src/Raven.Server/Documents/PeriodicBackup/Aws/RavenAwsS3Client.cs
```diff
@@ -43,9 +47,19 @@ public RavenAwsS3Client(S3Settings s3Settings, Progress progress = null, Logger
...
+                if (string.IsNullOrWhiteSpace(s3Settings.AwsRegionName))
+                    throw new ArgumentException("AWS region Name cannot be null or empty");
+                AwsRegion = s3Settings.AwsRegionName.ToLower();
+            }
+            else
+            {
+                if (string.IsNullOrWhiteSpace(s3Settings.AwsRegionName) == false)
+                    throw new ArgumentException($"There is no need for region in custom S3 server: AwsRegionName({s3Settings.AwsRegionName})");
+                _customS3ServerUrl = new Uri(s3Settings.CustomS3ServerUrl);
+            }
```
ayende: This seems to indicate that the region name isn't required for custom S3, but is that the case?
I don't think that this is true, see: https://developers.digitalocean.com/documentation/spaces/

src/Raven.Client/Documents/Operations/Backups/BackupSettings.cs
```diff
@@ -125,10 +125,15 @@ public override DynamicJsonValue ToJson()
...
         /// <summary>
-        /// Amazon S3 Bucket name.
+        /// S3 Bucket name.
         /// </summary>
         public string BucketName { get; set; }
 
+        /// <summary>
+        /// S3 server Url when using custom server
+        /// </summary>
+        public string CustomS3ServerUrl { get; set; }
```
ppekrol: We are in S3 Settings, so CustomServerUrl is enough.

src/Raven.Server/Documents/PeriodicBackup/Aws/RavenAwsS3Client.cs
```diff
@@ -33,6 +33,10 @@ public class RavenAwsS3Client : RavenAwsClient
 
         private readonly string _bucketName;
         private readonly Logger _logger;
+        private readonly Uri _customS3ServerUrl;
+
+
+        private bool IsCustomS3Server => _customS3ServerUrl != null;
```
ppekrol: HasCustomServerUrl?

src/Raven.Server/Documents/PeriodicBackup/Aws/RavenAwsS3Client.cs
```diff
@@ -927,12 +945,16 @@ public string CalculateMD5Hash(string input)
...
-            return $"{baseUrl}/{_bucketName}";
+            return IsCustomS3Server 
+                ? $"{_customS3ServerUrl}{_bucketName}" 
+                : $"{base.GetUrl()}/{_bucketName}";
         }
 
         public override string GetHost()
         {
+            if (IsCustomS3Server)
+                return _customS3ServerUrl.GetComponents(UriComponents.Host | UriComponents.Port, UriFormat.UriEscaped);
```
ppekrol, line 953: IIRC everything needs to be on HTTPS, but is this the case with custom server? Maybe someone will want a local instance without HTTPS? If we do want to handle that, then we need to return the protocol here also.

test/SlowTests/Server/Documents/PeriodicBackup/Restore/RestoreFromAwsS3.cs
```diff
@@ -0,0 +1,92 @@
...
+        protected async Task incremental_and_full_check_last_file_for_backup() => await incremental_and_full_check_last_file_for_backup_internal();
+        [AmazonS3Fact, Trait("Category", "Smuggler")]
+        protected async Task incremental_and_full_backup_encrypted_db_and_restore_to_encrypted_DB_with_provided_key() => 
+            await incremental_and_full_backup_encrypted_db_and_restore_to_encrypted_DB_with_provided_key_internal();
+        [AmazonS3Fact, Trait("Category", "Smuggler")]
+        protected async Task snapshot_encrypted_db_and_restore_to_encrypted_DB() => await snapshot_encrypted_db_and_restore_to_encrypted_DB_internal();
+
+        protected override S3Settings GetS3Settings(string subPath = null)
+        {
+            var s3Settings = CustomS3FactAttribute.S3Settings;
```
ppekrol: AmazonS3FactAttribute

test/Tests.Infrastructure/CustomS3FactAttribute.cs
```diff
@@ -0,0 +1,46 @@
...
+            {
+                ParsingError = e.ToString();
+            }
+        }
+
+        public CustomS3FactAttribute([CallerMemberName] string memberName = "")
+        {
+            if (string.IsNullOrEmpty(ParsingError) == false)
+            {
+                Skip = $"Failed to parse S3 settings, error: {ParsingError}";
```
ppekrol: custom S3 settings

test/Tests.Infrastructure/CustomS3FactAttribute.cs
```diff
@@ -0,0 +1,46 @@
...
+                return;
+            }
+
+            if (S3Settings == null)
+            {
+                Skip = $"S3 {memberName} tests missing S3 settings.";
```
ppekrol, line 42: custom S3 settings

src/Raven.Server/Documents/PeriodicBackup/Aws/RavenAwsS3Client.cs
haludi: Actually, I check it so far just on HTTP. The `GetHost` function is used just for the header and it does not accept the schema.

PR #10284, opened 2020-02-28
RavenDB-14689 fix byte string release issue on a different context

ppekrol:


src/Raven.Server/Documents/Queries/CollectionQueryEnumerable.cs
```diff
@@ -435,7 +435,12 @@ private void AddId(string id)
                         _allocator.ToLowerCase(ref key.Content);
                     }
                     else
-                        key = Slices.Empty;
```
arekpalinski, line 438: Just a note - `Slices.Empty` is allocated on static context so we must not return it to another one.

https://github.com/ravendb/ravendb/blob/9803a5af2434339b6f1caf766cdf0a778fc2ce41/src/Voron/Slice.cs#L211

PR #10285, opened 2020-02-28
RavenDB-14482 Prevent from putting reference documents if generated pattern ID is invalid (if it ends with / or | or contains null field values)

arekpalinski:
https://issues.hibernatingrhinos.com/issue/RavenDB-14482

### DIFF
--- a/src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
+++ b/src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
@@ -145,16 +145,30 @@ namespace Raven.Server.Documents.Indexes.MapReduce.OutputToCollection
 
             using (_patternForReduceOutputReferences?.BuildReferenceDocumentId(out referenceDocIdBuilder))
             {
+                var patternPropertiesAddedSuccessfully = true;
+
                 foreach (var property in _index.OutputReduceToCollectionPropertyAccessor.GetPropertiesInOrder(reduceObject))
                 {
                     var value = property.Value;
                     djv[property.Key] = TypeConverter.ToBlittableSupportedType(value, context: _indexContext);
 
                     if (referenceDocIdBuilder?.ContainsField(property.Key) == true)
-                        referenceDocIdBuilder.Add(property.Key, property.Value);
+                    {
+                        try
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReferencesPattern.cs
```diff
@@ -105,7 +107,20 @@ public string GetId()
...
+                var id = _id.AppendFormat(_formattedPattern, _values).ToString();
+                
+                ValidateId(id);
+
+                return id;
+            }
+
+            public static void ValidateId(string id)
+            {
+                if (id.EndsWith('/'))
```
ppekrol, line 119: After merging this to 5.0 we need to use DocumentDatabase.IndentityPartsSeparator here

ppekrol (approved):

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
```diff
@@ -145,16 +145,30 @@ private BlittableJsonReaderObject GenerateReduceOutput(object reduceObject, Inde
...
+                        try
+                        {
+                            referenceDocIdBuilder.Add(property.Key, property.Value);
+                        }
+                        catch (Exception e)
+                        {
+                            patternPropertiesAddedSuccessfully = false;
+
+                            if (stats != null) // should never be null
+                                stats.AddReduceError($"Failed to build document ID based on provided pattern for output to collection references. {e.Message}");
```
ayende: Can we add the full exception details?

arekpalinski: Fixed in https://github.com/ravendb/ravendb/pull/10285/commits/14b715f1b78381ca8b839ca027e8fd7c57e4ee2e

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReferencesPattern.cs
arekpalinski: Okay. Added TODO in issue comment

PR #10288, opened 2020-02-28
RavenDB-14685 exposing events in the IDocumentStore interface

ppekrol:


src/Raven.Client/Documents/IDocumentStore.cs
```diff
@@ -31,9 +31,13 @@ public interface IDocumentStore : IDisposalNotification
         X509Certificate2 Certificate { get; }
 
         /// <summary>
-        /// Store events
+        /// Store event
         /// </summary>
         event EventHandler<BeforeStoreEventArgs> OnBeforeStore;
+
+        /// <summary>
+        /// Session event
```
ayende: What does this comment mean?

src/Raven.Client/Documents/IDocumentStore.cs
```diff
@@ -31,9 +31,13 @@ public interface IDocumentStore : IDisposalNotification
         X509Certificate2 Certificate { get; }
 
         /// <summary>
-        /// Store events
+        /// Store event
```
ayende: What is the usefulness of this doc comment?

ppekrol: Removed.

src/Raven.Client/Documents/IDocumentStore.cs
ppekrol: Removed.

PR #10299, opened 2020-03-01
RavenDB-14223 time-series rollups & retention 

karmeli87:


src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
```diff
@@ -0,0 +1,181 @@
...
+    {
+        /// <summary>
+        /// Name of the time series policy, defined by the convention "Per{Amplitude}{DownSampleFrequency} (e.g. Per1Hour)"
+        /// </summary>
+        public string Name; // defined by convention
+
+        /// <summary>
+        /// How far from UTC.Now this policy should be applied
+        /// </summary>
+        public TimeSpan TimeFromNow;
```
ayende: I don't understand this.
You have two sides here: What the time frame that we work on? And what is the frequency that we do the roll up.
But the API seems really confusing. Also, why not simple have two time spans here?

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
```diff
@@ -0,0 +1,181 @@
...
+
+namespace Raven.Client.Documents.Operations.TimeSeries
+{
+    public class TimeSeriesCollectionConfiguration : IDynamicJson
+    {
+        public TimeSpan? DeleteAfter;
+
+        public bool Disabled;
+
+        public List<TimeSeriesDownSamplePolicy> DownSamplePolicies;
```
ayende: Better to call it roll up, not down sampling

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
```diff
@@ -741,6 +741,24 @@ public void WriteDatabaseRecord(DatabaseRecord databaseRecord, SmugglerProgressB
                     progress.SqlEtlsUpdated = true;
                 }
 
+                if (databaseRecord?.TimeSeries != null)
+                {
+                    if (currentDatabaseRecord?.TimeSeries != null)
+                    {
+                        foreach (var collection in currentDatabaseRecord.TimeSeries.Collections)
+                        {
+                            if ((databaseRecord.TimeSeries.Collections.ContainsKey(collection.Key)) == false)
```
ayende, line 750: Why this check?

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
```diff
@@ -0,0 +1,181 @@
...
+
+        public void Validate()
+        {
+            if (DownSamplePolicies.Count == 0)
+                return;
+
+            DownSamplePolicies.Sort(TimeSeriesDownSamplePolicyComparer.Instance);
```
ayende: This needs to run server side as well

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
karmeli87: This draft PR is exactly for this purpose, to decide on a good API.
 
I need here some ASCII art to explain what I've tried to do:
`+---- raw -----|t1|----- rollup1 ------|t2|------ rollup2 -----| delete >`

Each rollup policy have `t`, which defines how far from `now` is should be applied.
Also, it defines the "aggregation factor" which define the down sampling..

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
```diff
@@ -40,12 +41,18 @@ internal RollupPolicy GetPolicyByTimeSeries(string name)
 
         internal RollupPolicy GetNextPolicy(RollupPolicy policy)
```
ayende: This looks rather awkward API.
Why not simply do enumeration?

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
```diff
@@ -0,0 +1,202 @@
...
+                return null;
+            }
+
+            if (current == RollupPolicies.Count - 1)
+                return RollupPolicy.AfterAllPolices;
+
+            return RollupPolicies[current + 1];
+        }
+
+        internal RollupPolicy GetPreviousPolicy(RollupPolicy policy)
```
ayende: I don't like this kind of API

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,655 @@
...
+                    if (Equals(policyRunner.Configuration, dbRecord.TimeSeries))
+                        return policyRunner;
+                }
+                policyRunner?.Dispose();
+                var runner = new TimeSeriesPolicyRunner(database, dbRecord.TimeSeries);
+                runner.Start();
+                return runner;
+            }
+            catch (Exception e)
+            {
```
ayende, line 54: Make sure that you dispose the old `policyRunner` here

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,655 @@
...
+                StartIndex = (int)RollUpColumn.NextRollUp, 
+                Count = 1,
+                Name = NextRollUpIndex
+            });
+        }
+
+        public unsafe void MarkForPolicy(DocumentsOperationContext context, Slice key, string collection, string name, long etag, DateTime baseline, string changeVector)
+        {
+            var config = Configuration.Collections[collection];
```
ayende: What if this does not exists?

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
```diff
@@ -0,0 +1,202 @@
...
+
+        internal RollupPolicy GetPolicyByName(string policy)
+        {
+            if (policy == RollupPolicy.RawPolicyString)
+                return RollupPolicy.RawPolicy;
+
+            return RollupPolicies.SingleOrDefault(p => string.Compare(p.Name,policy, StringComparison.InvariantCultureIgnoreCase) == 0);
+        }
+
+        internal RollupPolicy GetPolicyByTimeSeries(string name)
```
ayende: This is very strange way to find a policy by name.
Why are you searching for a greater than here in an unordered policies?

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
karmeli87: This is that check we do for revision configuration as well. 
I guess we don't want to override an existing configuration? But I agree that it might result in an unexpected behavior without any warning/message. 
What do we want to do here? overwrite or notify that we skip this config, because one is already exists?

src/Raven.Client/Documents/Operations/TimeSeries/ConfigureTimeSeriesOperation.cs
```diff
@@ -0,0 +1,70 @@
...
+
+namespace Raven.Client.Documents.Operations.TimeSeries
+{
+    public class ConfigureTimeSeriesOperation : IMaintenanceOperation<ConfigureTimeSeriesOperationResult>
+    {
+        private readonly TimeSeriesConfiguration _configuration;
+
+        public ConfigureTimeSeriesOperation(TimeSeriesConfiguration configuration)
+        {
+            _configuration = configuration;
```
ppekrol: add `null` check

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
```diff
@@ -0,0 +1,229 @@
...
+using System.Linq;
+using Raven.Client.Documents.Queries.TimeSeries;
+using Sparrow.Json.Parsing;
+
+namespace Raven.Client.Documents.Operations.TimeSeries
+{
+    public class TimeSeriesCollectionConfiguration : IDynamicJson
+    {
+
+        public bool Disabled;
```
ppekrol: Let's use properties here. We do not care about performance here and you can easily override properties in the future, and fields are not that easy to work with.

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
```diff
@@ -0,0 +1,229 @@
...
+
+            return Policies[current - 1];
+        }
+
+        public DynamicJsonValue ToJson()
+        {
+            return new DynamicJsonValue
+            {
+                [nameof(Policies)] = new DynamicJsonArray(Policies.Select(p=>p.ToJson())),
+                [nameof(RawPolicy)] = RawPolicy.ToJson(),
```
ppekrol, line 126: Can this be null?

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
```diff
@@ -0,0 +1,229 @@
...
+                return RawPolicy;
+
+            return Policies[current - 1];
+        }
+
+        public DynamicJsonValue ToJson()
+        {
+            return new DynamicJsonValue
+            {
+                [nameof(Policies)] = new DynamicJsonArray(Policies.Select(p=>p.ToJson())),
```
ppekrol, line 125: Policies can be null?

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
```diff
@@ -0,0 +1,229 @@
...
+                [nameof(RawPolicy)] = RawPolicy.ToJson(),
+                [nameof(Disabled)] = Disabled
+            };
+        }
+    }
+
+    public class RawTimeSeriesPolicy : TimeSeriesPolicy
+    {
+        internal const string PolicyString = "rawpolicy"; // must be lower case
+        public RawTimeSeriesPolicy()
```
ppekrol, line 138: IIRC you can make it internal?

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
```diff
@@ -0,0 +1,229 @@
...
+        {
+            // for de-serializer
+        }
+
+        public RawTimeSeriesPolicy(TimeSpan? retentionTime)
```
ppekrol: Sanity-check. Null is expected value here?

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
```diff
@@ -0,0 +1,229 @@
...
+        /// <summary>
+        /// Define the aggregation of this policy
+        /// </summary>
+        public TimeSpan AggregationTime { get; private set; }
+
+        /// <summary>
+        /// Define the aggregation type
+        /// </summary>
+        public AggregationType Type { get; private set; }
+        // TODO: consider Continuous Query approach
```
ppekrol: TODO

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
```diff
@@ -0,0 +1,229 @@
...
+
+            if (retentionTime <= TimeSpan.Zero)
+                throw new ArgumentException("Must be greater than zero", nameof(retentionTime));
+
+            RetentionTime = retentionTime;
+            AggregationTime = aggregationTime;
+            Type = type;
+
+            var retentionStr = retentionTime == TimeSpan.MaxValue ? "" : $"KeepFor{retentionTime}";
+            Name = $"Every{AggregationTime}By{Type}{retentionStr}";
```
ppekrol: Do we want to 'split' it using `_` like in index names?

src/Raven.Client/Documents/Queries/TimeSeries/TimeSeriesAggregation.cs
```diff
@@ -25,4 +25,15 @@ public class TimeSeriesRangeAggregation
         public DateTime To, From;
     }
 
+    public enum AggregationType
+    {
+        Min,
+        Max,
+        Mean,
+        Avg,
```
ppekrol: rename to `Average` it will match what we have in Facets

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
```diff
@@ -0,0 +1,229 @@
...
+
+            Policies.Sort(TimeSeriesDownSamplePolicyComparer.Instance);
+        }
+
+        internal TimeSeriesPolicy GetPolicyByName(string policy)
+        {
+            if (policy == RawTimeSeriesPolicy.PolicyString)
+                return RawPolicy;
+
+            return Policies.SingleOrDefault(p => string.Compare(p.Name,policy, StringComparison.InvariantCultureIgnoreCase) == 0);
```
ayende: I don't understand this. Why are you doing a query like that, and why is it `O(n)` ?

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
```diff
@@ -0,0 +1,229 @@
...
+        }
+
+        internal TimeSeriesPolicy GetPolicyByTimeSeries(string name)
+        {
+            if (name.Contains(TimeSeriesConfiguration.TimeSeriesRollupSeparator) == false)
+                return RawPolicy;
+
+            return Policies.SingleOrDefault(p => name.IndexOf(p.Name, StringComparison.InvariantCultureIgnoreCase) > 0);
```
ayende: Why comparing by greater than?

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
```diff
@@ -0,0 +1,229 @@
...
+
+        internal TimeSeriesPolicy GetNextPolicy(TimeSeriesPolicy policy)
+        {
+            if (Policies.Count == 0)
+                return null;
+
+            if (policy == RawPolicy)
+                return Policies[0];
+
+            var current = Policies.FindIndex(p => p == policy);
```
ayende: This is pretty ugly. Can't we hold the index of the policy, so we don't have to do this?

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
```diff
@@ -0,0 +1,229 @@
...
+
+            return Policies[current + 1];
+        }
+
+        internal TimeSeriesPolicy GetPreviousPolicy(TimeSeriesPolicy policy)
+        {
+            if (policy == RawPolicy)
+                return TimeSeriesPolicy.BeforeAllPolices;
+
+            var current = Policies.FindIndex(p => p == policy);
```
ayende: Same, enumerator, maybe?

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
```diff
@@ -0,0 +1,229 @@
...
+
+        /// <summary>
+        /// Define the aggregation of this policy
+        /// </summary>
+        public TimeSpan AggregationTime { get; private set; }
+
+        /// <summary>
+        /// Define the aggregation type
+        /// </summary>
+        public AggregationType Type { get; private set; }
```
ayende: If we do this, then we have to provide the ability to have multiple rollups. One for min, one for max, etc.
Can't we do all the aggregations, instead?

src/Raven.Server/Documents/TimeSeries/BitsBuffer.cs
```diff
@@ -123,6 +123,7 @@ public BitsBuffer(byte* buffer, int size)
             Header = (BitsBufferHeader*)buffer;
             _buffer = buffer + sizeof(BitsBufferHeader) + Header->CompressedSize;
             Size = size;
+            Debug.Assert(size > 0);
```
ppekrol: Write a message here. It can be as simple as `size > 0`. It is easier to trace failing assert then

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,683 @@
...
+        private static readonly TableSchema RollUpSchema;
+        public static readonly Slice TimeSeriesRollUps;
+        private static readonly Slice RollUpKey;
+        private static readonly Slice NextRollUpIndex;
+        private enum RollUpColumn
+        {
+            // documentId/Name
+            Key,
+            Collection,
+            NextRollUp,
```
ppekrol: Isn't the word just `Rollup` ?

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,683 @@
...
+            public string RollUpPolicy;
+            public string ChangeVector;
+        }
+
+        static TimeSeriesPolicyRunner()
+        {
+            using (StorageEnvironment.GetStaticContext(out var ctx))
+            {
+                Slice.From(ctx, "TimeSeriesRollUps", ByteStringType.Immutable, out TimeSeriesRollUps);
+                Slice.From(ctx, "RollUpKey", ByteStringType.Immutable, out RollUpKey);
```
ppekrol: nameof?

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,683 @@
...
+            {
+                StartIndex = (int)RollUpColumn.NextRollUp, 
+                Count = 1,
+                Name = NextRollUpIndex
+            });
+        }
+
+        public unsafe void MarkForPolicy(DocumentsOperationContext context, Slice key, string collection, string name, long etag, DateTime baseline, string changeVector)
+        {
+            var config = Configuration.Collections[collection];
```
ppekrol: Are we certain that this MUST be in that dictionary? Shouldn't we use TryGet?

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,683 @@
...
+
+            var integerPart = baseline.Ticks / nextPolicy.AggregationTime.Ticks;
+            var nextRollup = nextPolicy.AggregationTime.Ticks * (integerPart + 1);
+
+            // mark for rollup
+            RollUpSchema.Create(context.Transaction.InnerTransaction, TimeSeriesRollUps, 16);
+            var table = context.Transaction.InnerTransaction.OpenTable(RollUpSchema, TimeSeriesRollUps);
+            using (table.Allocate(out var tvb))
+            using (Slice.From(context.Allocator, key.Content.Ptr, key.Size - sizeof(long) - sizeof(byte), ByteStringType.Immutable, out var timeSeriesKey))
+            using (Slice.From(context.Allocator,collection, ByteStringType.Immutable,out var collectionSlice))
```
ppekrol: This should be lowered?

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,683 @@
...
+            using (Slice.From(context.Allocator, nextPolicy.Name, ByteStringType.Immutable, out var policyToApply))
```
ppekrol: This should be lowered?

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,683 @@
...
+                if (Logger.IsOperationsEnabled)
+                        Logger.Operations($"Failed to roll-up time series for '{_database.Name}' which are older than {now}", e);
+            }
+        }
+
+        internal async Task DoRetention()
+        {
+            var topology = _database.ServerStore.LoadDatabaseTopology(_database.Name);
+            var isFirstInTopology = string.Equals(topology.Members.FirstOrDefault(), _database.ServerStore.NodeTag, StringComparison.OrdinalIgnoreCase);
+            if (isFirstInTopology == false)
```
ppekrol, line 229: Do we want to go with first? Isn't that the most busiest one? It probably it the most up to date, but do we care about that?

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,683 @@
...
+                    Etag = DocumentsStorage.TableValueToLong((int)RollUpColumn.Etag, ref item.Result.Reader),
+                    ChangeVector = DocumentsStorage.TableValueToChangeVector(context, (int)RollUpColumn.ChangeVector, ref item.Result.Reader)
+                };
+
+                states.Add(state);
+                take--;
+            }
+        }
+
+        public static void SplitKey(Slice key, out string docId, out string name)
```
ppekrol: Sanity-check. Shouldn't be an issue that we are allocating here instead of 'slicing' Slices?

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,683 @@
...
+        {
+            var separatorIndex = key.Content.IndexOf(SpecialChars.RecordSeparator);
+            docId = key.Content.Substring(separatorIndex);
+            name = key.Content.Substring(separatorIndex + 1, key.Content.Length - separatorIndex - 1);
+        }
+
+        internal class TimeSeriesRetentionCommand : TransactionOperationsMerger.MergedTransactionCommand
+        {
+            public const int BatchSize = 1024;
```
ppekrol: What is this used for?

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,683 @@
...
+                RollUpSchema.Create(context.Transaction.InnerTransaction, TimeSeriesRollUps, 16);
+                var table = context.Transaction.InnerTransaction.OpenTable(RollUpSchema, TimeSeriesRollUps);
+                foreach (var key in tss.Stats.GetTimeSeriesNameByPolicy(context, _collection, _from.Name, _skip, BatchSize))
+                {
+                    using (table.Allocate(out var tvb))
+                    using (Slice.From(context.Allocator,_collection.Name, ByteStringType.Immutable,out var collectionSlice))
+                    using (Slice.From(context.Allocator, _to.Name, ByteStringType.Immutable, out var policyToApply))
+                    using (Slice.From(context.Allocator, string.Empty, ByteStringType.Immutable, out var changeVectorSlice))
+                    {
+                        tvb.Add(key);
```
ppekrol: Are we sure we want code like this here instead of e.g. `TimeSeriesStorage.Rollups.MethodName()` ?

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,683 @@
...
+                {
+                    table.DeleteByKey(item);
+                }
+
+                return RolledUp;
+            }
+
+            public override TransactionOperationsMerger.IReplayableCommandDto<TransactionOperationsMerger.MergedTransactionCommand> ToDto(JsonOperationContext context)
+            {
+                throw new NotImplementedException();
```
ppekrol: Remember that we need to fix this later on

src/Raven.Server/Documents/TimeSeries/TimeSeriesStats.cs
```diff
@@ -0,0 +1,290 @@
...
+            End = 4,
+            Count = 5,
+            
+        }
+
+        static TimeSeriesStats()
+        {
+            using (StorageEnvironment.GetStaticContext(out ByteStringContext ctx))
+            {
+                Slice.From(ctx, "TimeSeriesStats", ByteStringType.Immutable, out TimeSeriesStatsKey);
```
ppekrol: nameof?

src/Raven.Server/ServerWide/Commands/EditTimeSeriesConfigurationCommand.cs
```diff
@@ -0,0 +1,39 @@
...
+            Configuration = configuration;
+        }
+
+        public override string UpdateDatabaseRecord(DatabaseRecord record, long etag)
+        {
+            if (Configuration?.Collections != null)
+            {
+                foreach (var config in Configuration.Collections.Values)
+                {
+                    config.Validate();
```
ppekrol: can value be null?

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
ayende: This will generate names such as: "Every00:30ByAvgKeepFor1.00.00", no?
That seems pretty ugly

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,683 @@
...
+
+                if (policyRunner != null)
+                {
+                    // no changes
+                    if (Equals(policyRunner.Configuration, dbRecord.TimeSeries))
+                        return policyRunner;
+                }
+                policyRunner?.Dispose();
+                var runner = new TimeSeriesPolicyRunner(database, dbRecord.TimeSeries);
+                runner.Start();
```
ayende, line 50: Do we have to worry about concurrency here?

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,683 @@
...
+                return null;
+            }
+        }
+        public static readonly Slice RawPolicySlice;
+
+        private static readonly TableSchema RollUpSchema;
+        public static readonly Slice TimeSeriesRollUps;
+        private static readonly Slice RollUpKey;
+        private static readonly Slice NextRollUpIndex;
+        private enum RollUpColumn
```
ayende:
```suggestion
        private enum RollUpColumns
```

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,683 @@
...
+        public unsafe void MarkForPolicy(DocumentsOperationContext context, Slice key, string collection, string name, long etag, DateTime baseline, string changeVector)
+        {
+            var config = Configuration.Collections[collection];
+
+            if (config.Disabled)
+                return;
+
+            var current = config.GetPolicyByTimeSeries(name);
+            if (current == null) // policy not found
+                // TODO: delete this time-series if not the raw one?
```
ayende: No, changing the policy doesn't delete data

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,683 @@
...
+                return;
+
+            var nextPolicy = config.GetNextPolicy(current);
```
ayende, line 107: I really don't like the next / prev thing, what purpose does it serve?

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
ayende: Yes, collection names are case insensitive

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,683 @@
...
+
+                        var topology = _database.ServerStore.LoadDatabaseTopology(_database.Name);
+                        var isFirstInTopology = string.Equals(topology.Members.FirstOrDefault(), _database.ServerStore.NodeTag, StringComparison.OrdinalIgnoreCase);
+
+                        var command = new RollupTimeSeriesCommand(Configuration, states, isFirstInTopology);
+                        await _database.TxMerger.Enqueue(command);
+                        if (command.RolledUp == 0)
+                            break;
+
+                        states.Clear();
```
ayende: Shouldn't this be at the top ? Just in case we change the structure, it would be easier to manage

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
ayende: This is what we do for expiration and refreshing. That give us the most consistency

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,683 @@
...
+                        using (context.OpenReadTransaction())
+                        {
+                            collectionName = _database.DocumentsStorage.ExtractCollectionName(context, collection);
+                        }
+
+                        var request = new TimeSeriesStorage.DeletionRangeRequest
+                        {
+                            Collection = collection,
+                            From = DateTime.MinValue
+                        };
```
ayende, line 257: Shouldn't you have a `To` here?

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
ayende: I think that this uses implicit conversions. But should it?

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
ayende: Used elsewhere to limit how many items we scan per batch

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,683 @@
...
+            protected override long ExecuteCmd(DocumentsOperationContext context)
+            {
+                foreach (var key in _keys)
+                {
+                    SplitKey(key, out var docId, out var name);
+                         
+                    _request.Name = name;
+                    _request.DocumentId = docId;
+
+                    context.DocumentDatabase.DocumentsStorage.TimeSeriesStorage.RemoveTimestampRange(context, _request);
```
ayende: I think this will delete everything, because `To` isn't specified.

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,683 @@
...
+                _collection = collection;
+                _policy = policy;
+            }
+            protected override long ExecuteCmd(DocumentsOperationContext context)
+            {
+                var tss = context.DocumentDatabase.DocumentsStorage.TimeSeriesStorage;
+                RollUpSchema.Create(context.Transaction.InnerTransaction, TimeSeriesRollUps, 16);
+                var table = context.Transaction.InnerTransaction.OpenTable(RollUpSchema, TimeSeriesRollUps);
+
+                var toDelete = new List<Slice>(tss.Stats.GetTimeSeriesNameByPolicy(context, _collection, _policy, 0, BatchSize));
```
ayende: What if there are more than `BatchSize`?

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,683 @@
...
+                RollUpSchema.Create(context.Transaction.InnerTransaction, TimeSeriesRollUps, 16);
+                var table = context.Transaction.InnerTransaction.OpenTable(RollUpSchema, TimeSeriesRollUps);
+                var toRemove = new List<Slice>();
+
+                foreach (var item in _states)
+                {
+                    if (_configuration == null)
+                        return RolledUp;
+
+                    var config = _configuration.Collections[item.Collection];
```
ayende: TryGet, no?

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,683 @@
...
+
+                    if (item.Etag != DocumentsStorage.TableValueToLong((int)RollUpColumn.Etag, ref current))
+                        continue; // concurrency check
+
+                    var startFrom = item.NextRollUp.Add(-policy.AggregationTime);
+                    var rawTimeSeries = item.Name.Split(TimeSeriesConfiguration.TimeSeriesRollupSeparator)[0];
+                    var intoTimeSeries = policy.GetTimeSeriesName(rawTimeSeries);
+                    
+                    var intoReader = tss.GetReader(context, item.DocId, intoTimeSeries, startFrom, startFrom);
+                    var previouslyAggregated = intoReader.AllValues().Any();
```
ayende: Doesn't this consume the reader?

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,683 @@
...
+                        var changeVector = intoReader.GetCurrentSegmentChangeVector();
+                        if (ChangeVectorUtils.GetConflictStatus(item.ChangeVector, changeVector) == ConflictStatus.AlreadyMerged)
+                        {
+                            // this rollup is already done
+                            toRemove.Add(item.Key);
+                            continue;
+                        }
+                    }
+
+                    if (_isFirstInTopology == false)
```
ayende: Why do we have to get to this point to bail out? To remove the policies markers?

src/Raven.Server/Documents/TimeSeries/TimeSeriesStats.cs
```diff
@@ -0,0 +1,290 @@
...
+        private static readonly Slice TimeSeriesStatsKey;
+        private static readonly Slice PolicyIndex;
+        private static readonly Slice StartTimeIndex;
+        private static readonly TableSchema TimeSeriesStatsSchema = new TableSchema();
+
+        private enum StatsColumn
+        {
+            Key = 0, // documentId, separator, name
+            PolicyName = 1,
+            Separator = 2,
```
ayende: Why is this its own dedicated column?

src/Raven.Server/Documents/TimeSeries/TimeSeriesStats.cs
```diff
@@ -0,0 +1,290 @@
...
+namespace Raven.Server.Documents.TimeSeries
+{
+    public class TimeSeriesStats
+    {
+        private static readonly Slice TimeSeriesStatsKey;
+        private static readonly Slice PolicyIndex;
+        private static readonly Slice StartTimeIndex;
+        private static readonly TableSchema TimeSeriesStatsSchema = new TableSchema();
+
+        private enum StatsColumn
```
ayende:
```suggestion
        private enum StatsColumns
```

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
ayende: I think this was originally meant for import, so if you update an existing value?

src/Sparrow.Server/ByteString.cs
```diff
@@ -332,6 +332,17 @@ public string Substring(int length)
...
 
+        public string Substring(int offset, int length)
+        {
+            if (!HasValue)
+                return string.Empty;
+
+            EnsureIsNotBadPointer();
+
+            var encoding = Encodings.Utf8;
+            return encoding.GetString(_pointer->Ptr + offset, length);
```
ayende: This is not a valid way to handle things. `offset` might fall inside of UTF8 mulit byte char

src/Raven.Client/Documents/Operations/TimeSeries/ConfigureTimeSeriesOperation.cs
karmeli87: `null` should remove the configuration and close the runner. 
will verify and add a test

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
karmeli87: no, test added

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
karmeli87: no, test added

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
karmeli87: removed the `nullable` and the ambiguity between `null` and `TimeSpan.MaxValue`.

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
karmeli87: But there is a point that we are putting more and more load on the primary.. for now I leave it as is.

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
karmeli87: Not sure I'm following.

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
karmeli87: We are indeed allocating here. 
But let's optimize it a bit later, since changing this will require to change `string` into `slice` in various places

src/Raven.Server/ServerWide/Commands/EditTimeSeriesConfigurationCommand.cs
karmeli87: maybe. will safe-guard.

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
karmeli87: This actually an interesting issue.
Let's say we have a single value time-series. Stock price. 

Right now each aggregation result is represented by a point on the policy time-series. So if I aggregate daily by `Average`, each day I'll create a single point from the raw stock price.

But if I want to aggregate daily by `First`, `Last`, `High`, `Low`.
Right now it is a problem, since I can define only a single type.
To allow this kind of thing we can either create a complex policy tree or create a single 4-dim time-series (which I like more).

But do we want to tackle this now?

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
karmeli87: Don't believe so, since eventually all goes through the tx merger.

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
karmeli87: So changing the configuration should keep left-overs?
The user should manually delete them?

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
karmeli87: I populate this in line `394` .
Each policy has a different to.

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
karmeli87: Here we are in the tx merger. We will release it and loop in again from the caller.

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
karmeli87: Yes, but I need it only for this purpose.

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
karmeli87: Yes, remove rollups that were already done and replicated.

src/Raven.Server/Documents/TimeSeries/TimeSeriesStats.cs
karmeli87: Easier to work with, instead of adding/removing the separator each time.
Is it a problem?

src/Sparrow.Server/ByteString.cs
karmeli87: Okay, I'll remove this entirely, since we want to optimize the caller function anyway (`SplitKey`)

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
karmeli87: Now I actually using it.

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
karmeli87: Yes. Sorry. 
Should we let the user define it? 

Note that I persist the name of the policy which is unique for it's definition.

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
karmeli87: The policies sorted by `AggregationTime`, I can extract this time from the name. 
But should we care about it? `O(n)` where `n` is expected to be low.
What do you suggest? I really don't want to persist the index of the policy.

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
karmeli87: If it contains or not? and `Contains` doesn't work with `StringComparison`.

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
karmeli87: Okay, let's hold the index

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
karmeli87: So we have a bigger problem here, we keep it with the original casing
e.g. https://github.com/ravendb/ravendb/blob/23c596b3b0f84bca9f7c19ec02e64a89850189a2/src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs#L906
Also, why do we store the collection at all? for replication only? 

Anyway, why do we care about it here? we need only to ensure that when we read from the configuration we do it in case-insensitive matter.

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
ppekrol: What I've meant is that we are using xxxStorage classes for storage interaction and here you are inserting directly into table without using TimeSeriesStorage.Rollups for example

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
ppekrol: Can we convert those timespan values to some human readable value and use them in the name or do we prefer user to specify the name?

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -0,0 +1,706 @@
...
+        }
+
+        internal async Task DoRetention()
+        {
+            var topology = _database.ServerStore.LoadDatabaseTopology(_database.Name);
+            var isFirstInTopology = string.Equals(topology.Members.FirstOrDefault(), _database.ServerStore.NodeTag, StringComparison.OrdinalIgnoreCase);
+            if (isFirstInTopology == false)
+                return;
+
+            var now = DateTime.UtcNow;
```
ppekrol: database.Time.UtcNow()

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
karmeli87: Okay, I will move it to dedicated class.

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
karmeli87: We will let the user to define the policy name. It should be unique and we will do only best effort validation.

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
karmeli87: Removing the policy will keep the data by default and it will be user's responsibility to delete them.
 
Changing the parameters for an existing policy will take effect from the time of the change and will keep previous data intact. (unless is should re-aggregate due to the parameters change)

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
karmeli87: The raw time series will be always aggregated into (`First`,`Last`,`Min`,`Max`,`Sum`,`Count`).
In that order.
Other policies will aggregate each of those values further independently.

ayende (approved):

src/Raven.Client/Documents/Operations/TimeSeries/ConfigureTimeSeriesOperation.cs
```diff
@@ -0,0 +1,70 @@
...
+
+namespace Raven.Client.Documents.Operations.TimeSeries
+{
+    public class ConfigureTimeSeriesOperation : IMaintenanceOperation<ConfigureTimeSeriesOperationResult>
+    {
+        private readonly TimeSeriesConfiguration _configuration;
+
+        public ConfigureTimeSeriesOperation(TimeSeriesConfiguration configuration)
+        {
+            _configuration = configuration ?? new TimeSeriesConfiguration();
```
ppekrol, line 18: Haven't you wrote that setting null here will remove the configuration? `ConfigureExpirationOperation` allows that.

src/Raven.Client/Documents/Queries/TimeSeries/TimeSeriesAggregation.cs
```diff
@@ -21,8 +21,19 @@ public class TimeSeriesAggregationResult : TimeSeriesQueryResult
...
-        public double?[] Max, Min, Last, First, Avg;
+        public double?[] Max, Min, Last, First, Average;
         public DateTime To, From;
     }
 
+    public enum AggregationType
+    {
+        Min,
+        Max,
+        Mean,
```
ppekrol, line 32: What is the difference between `Mean` and `Average` ?

ppekrol: Arithmetic Mean IIRC is average, is this another type of mean?

ppekrol (approved):

src/Raven.Client/Documents/Queries/TimeSeries/TimeSeriesAggregation.cs
karmeli87: Indeed Arithmetic Mean is Average. We don't support anything else.

ppekrol: I think that we can remove 'Mean' ?

src/Raven.Client/Documents/Operations/TimeSeries/ConfigureTimeSeriesOperation.cs
karmeli87: Actually passing `null` to `ConfigureExpirationOperation` will throw 
```
System.InvalidOperationException : Cannot perform ReadValue when encountered the ReadObjectDocument state
   at Sparrow.Json.ManualBlittableJsonDocumentBuilder`1.ThrowIllegalStateException(ContinuationState state, String realOperation) in C:\Work\RavenDB-5.0\src\Sparrow\Json\ManualBlittableJsonDocumentBuilder.cs:line 582
   at Sparrow.Json.ManualBlittableJsonDocumentBuilder`1.FinishWritingScalarValue(BuildingState currentState) in C:\Work\RavenDB-5.0\src\Sparrow\Json\ManualBlittableJsonDocumentBuilder.cs:line 566
   at Sparrow.Json.ManualBlittableJsonDocumentBuilder`1.WriteValueNull() in C:\Work\RavenDB-5.0\src\Sparrow\Json\ManualBlittableJsonDocumentBuilder.cs:line 284
   at Raven.Client.Json.BlittableJsonWriter.WriteNull() in C:\Work\RavenDB-5.0\src\Raven.Client\Json\BlittableJsonWriter.cs:line 338
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeValue(JsonWriter writer, Object value, JsonContract valueContract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.Serialize(JsonWriter jsonWriter, Object value, Type objectType)
   at Newtonsoft.Json.JsonSerializer.SerializeInternal(JsonWriter jsonWriter, Object value, Type objectType)
   at Newtonsoft.Json.JsonSerializer.Serialize(JsonWriter jsonWriter, Object value)
   ....
```

PR #10307, opened 2020-03-03
RavenDB-14610 on import strip counters/attachments/time-series flags

karmeli87:


ppekrol (approved):

ppekrol:
Do we want this in 4.2?

karmeli87:
Probably we do

PR #10323, opened 2020-03-05
RavenDB-13343 Separate log messages

Danielle9897:


src/Raven.Studio/typescript/viewmodels/manage/adminLogs.ts
```diff
@@ -54,6 +54,8 @@ class heightCalculator {
         this.lineHeight = doubleLinesHeight - initialHeight;
         this.padding = doubleLinesHeight - 2 * this.lineHeight;
         this.charactersPerLine = charactersInline;
+        
+        row.populate("", 0, 0, undefined);
```
ml054: I think it should be "", 0, -200 so this row will be offscreen.

ml054 (approved):

PR #10327, opened 2020-03-08
HRINT-1536 promotable nodes shouldn't save documents from cluster transaction

karmeli87:


src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -446,6 +446,13 @@ private async Task ExecuteClusterTransactionTask()
         {
             while (DatabaseShutdown.IsCancellationRequested == false)
             {
+                var topology = ServerStore.LoadDatabaseTopology(Name);
+                if (topology.Promotables.Contains(ServerStore.NodeTag))
+                {
+                    await Task.Delay(1000);
```
ayende, line 452: Missing cancellation token for database shutdown.
And why can't we do this?

karmeli87: node A has change vector of `RAFT:1000` (and truncated log), now we add new node B.
Before A sends any document we do a cluster tx, so A & B execute it, so the change vector of A & B is now `RAFT:1001` and so A will skip sending the previous documents.

ayende: Isn't this the same scenario if you are running on normal members?
What if there is a small lag?

karmeli87: Because for cluster tx we aren't rely on the replication.
If we are a member we will read it from the state machine

ayende: But won't that stop the replication anyway?

In the promotable case, you are also reading it from the state machine.

PR #10329, opened 2020-03-08
RavenDB-14734: Disable bloom filters on 32 bits for indexes on large …

garayx:
…collection

### DIFF
--- a/src/Raven.Server/Documents/Indexes/CollectionOfBloomFilters.cs
+++ b/src/Raven.Server/Documents/Indexes/CollectionOfBloomFilters.cs
@@ -210,7 +210,7 @@ namespace Raven.Server.Documents.Indexes
             if (_currentFilter.Count < _currentFilter.Capacity)
                 return;
 
-            if (_mode == Mode.X86 && _filters.Length >= 8)
+            if (_mode == Mode.X86 && _filters.Length >= 20)
             {
                 Consumed = true;
                 return;
--- a/src/Raven.Server/Documents/Indexes/CollectionOfBloomFilters.cs
+++ b/src/Raven.Server/Documents/Indexes/CollectionOfBloomFilters.cs
@@ -37,7 +37,8 @@ namespace Raven.Server.Documents.Indexes
         private BloomFilter _currentFilter;
         private Mode _mode;
         private readonly Tree _tree;
-        public bool Consumed { get; private set; }
+        public bool Consumed;
+        private static readonly long _consumed = -1L;
 
         private CollectionOfBloomFilters(Mode mode, Tree tree, TransactionOperationContext context)
         {
@@ -63,6 +64,18 @@ namespace Raven.Server.Documents.Indexes
         {
             var tree = indexContext.Transaction.InnerTransaction.CreateTree("BloomFilters");
             var count = GetCount(tree, ref mode);
+            if (count == _consumed)
+            {
+                Debug.Assert(mode == Mode.X86, "BloomFilters in x64 mode got consumed, should not happen and likely a bug!");
+
+                var consumedCollection = new CollectionOfBloomFilters(mode, tree: null, context: null)
+                {
+                    Consumed = true
+                };
+
+                return consumedCollection;
+            }
+
             var collection = new CollectionOfBloomFilters(mode, tree, indexContext);
 
             for (var i = 0; i < count; i++)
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/Indexes/CollectionOfBloomFilters.cs
```diff
@@ -37,6 +37,7 @@ public enum Mode
         private BloomFilter _currentFilter;
         private Mode _mode;
         private readonly Tree _tree;
+        public bool Consumed { get; private set; }
```
ppekrol: switch to field

src/Raven.Server/Documents/Indexes/CollectionOfBloomFilters.cs
```diff
@@ -201,6 +211,16 @@ public void Flush()
             if (_filters == null || _filters.Length == 0)
                 return;
 
+            if (Consumed)
```
ppekrol, line 227: How does it behave after reloading the index?

src/Raven.Server/Documents/Indexes/MapIndexBase.cs
```diff
@@ -53,9 +54,16 @@ public override int HandleMap(LazyStringValue lowerId, LazyStringValue id, IEnum
             EnsureValidStats(stats);
 
             bool mustDelete;
-            using (_stats.BloomStats.Start())
+            if (_filters.Consumed == false)
             {
-                mustDelete = _filters.Add(lowerId) == false;
+                using (_stats.BloomStats.Start())
+                {
+                    mustDelete = _filters.Add(lowerId) == false;
```
ppekrol: I would rather prefer to handle everything inside the filters instead of adding this logging outside

src/Raven.Server/Documents/Indexes/CollectionOfBloomFilters.cs
garayx: after index reset the CollectionOfBloomFilters is recreated, and repopulated with BloomFilters

ppekrol: I'm talking here more about shutting down the database, not resetting the index.

garayx: If it was consumed then it will be marked as consumed from the storage when DB reset

ppekrol: Great, can you write a test for this?

garayx: Sure !

ppekrol (approved):

src/Raven.Server/Documents/Indexes/CollectionOfBloomFilters.cs
```diff
@@ -193,6 +210,12 @@ private void ExpandFiltersIfNecessary()
             if (_currentFilter.Count < _currentFilter.Capacity)
                 return;
 
+            if (_mode == Mode.X86 && _filters.Length >= 8)
```
ppekrol: How many docs +/- this is?

garayx: ~120k docs

ppekrol: Seems low?

garayx: increased to up to 500k docs (25k * 20)

src/Raven.Server/Documents/Indexes/CollectionOfBloomFilters.cs
```diff
@@ -441,6 +477,14 @@ public void Flush()
                 _tree.Increment(_keySlice, Count - _initialCount);
             }
 
+            internal void Delete()
+            {
+                if (_tree.Llt.Environment.Options.IsCatastrophicFailureSet)
```
ayende, line 482: Under what scenario this will happen?

ayende (approved):

src/Raven.Server/Documents/Indexes/CollectionOfBloomFilters.cs
garayx: I copied it from Flush() method,
The scenario of Delete() is similar to Flush() both called from same place.
https://github.com/ravendb/ravendb/pull/10329/files#diff-bbed84f00f70fe07cd5deb3915a02d4aR474

PR #10330, opened 2020-03-09
RavenDB-14620 Support string equals in all arguments order

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-14620

### DIFF
--- a/src/Raven.Client/Documents/Linq/RavenQueryProviderProcessor.cs
+++ b/src/Raven.Client/Documents/Linq/RavenQueryProviderProcessor.cs
@@ -635,66 +635,37 @@ namespace Raven.Client.Documents.Linq
 
         private void VisitEquals(MethodCallExpression expression)
         {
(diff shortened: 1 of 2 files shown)


src/Raven.Client/Documents/Linq/RavenQueryProviderProcessor.cs
```diff
@@ -718,6 +689,25 @@ private void VisitEquals(MethodCallExpression expression)
             });
         }
 
+        private static void EqualsGetArgs(MethodCallExpression expression, out Expression firstArg, out Expression secondArg, out Expression comparisonArg)
```
grisha-kotler:
```suggestion
        private static void GetComparisonExpressions(MethodCallExpression expression, out Expression firstArg, out Expression secondArg, out Expression comparisonArg)
```

grisha-kotler (changes requested):

grisha-kotler (approved):

PR #10337, opened 2020-03-11
RavenDB-14610 on import remove @timeseries from metadata

karmeli87:


src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
```diff
@@ -889,9 +909,13 @@ void AppendEntireSegment()
                     tvb.Add(segment.Ptr, segment.NumberOfBytes);
                     tvb.Add(collectionSlice);
                     tvb.Add(context.GetTransactionMarker());
+                    tvb.Add(tsNameSlice);
```
ppekrol: We explicitly didn't want to have it IIRC. The whole indexing was even designed in a way that the name is not available at the storage level?

This will have severe impact on the storage size.

karmeli87: Good point. Will try to avoid this.

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
```diff
@@ -470,7 +478,7 @@ void RemoveTimeSeriesNameFromMetadata(DocumentsOperationContext ctx, string docI
                         tsNamesList.Add(val);
                     }
 
-                    var location = tsNames.BinarySearch(tsName, StringComparison.Ordinal);
+                    var location = tsNames.BinarySearch(tsName, StringComparison.OrdinalIgnoreCase);
```
ayende, line 481: Here, are we *sure* that the names are sorted using case insensitive manner? I would expect that they are case sensitive here.

src/Raven.Server/Smuggler/Documents/DatabaseSource.cs
```diff
@@ -499,6 +500,61 @@ public IEnumerable<TimeSeriesItem> GetTimeSeries(List<string> collectionsToExpor
...
+                    }
+                }
+            }
+            catch
+            {
+#if DEBUG
+                throw;
+#endif
+            }
+
```
ayende: What is this for?

src/Raven.Server/Smuggler/Documents/DatabaseSource.cs
```diff
@@ -499,6 +500,61 @@ public IEnumerable<TimeSeriesItem> GetTimeSeries(List<string> collectionsToExpor
...
+            Debug.Assert(false, "Are you exporting an orphaned time-series?");
+            return lowerName;
```
ayende: If this is the case, need to actually throw real error here.
Otherwise, in release, we'll do horrible things here.

ayende (changes requested):

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
karmeli87: The time-series `Heartrate` and `HeartRate` must be the same. 
Why would you expect it to be case sensitive? 

anyway it should be consistent with:
https://github.com/ravendb/ravendb/blob/bad6628a9e11862a5158da0eaef9c82818e4a16b/src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs#L1656

ayende: Okay, that ensures the right behavior

src/Raven.Server/Smuggler/Documents/DatabaseSource.cs
karmeli87: It is only a precaution, but in case it did happened, do we want to fail the export because of this? if this happened in production, we need a way to fix it.
Worst case we will export time-series that doesn't have any document (and with the lowered id name)

src/Raven.Server/Smuggler/Documents/DatabaseSource.cs
karmeli87: related to the other comment

src/Raven.Server/Smuggler/Documents/DatabaseSource.cs
ayende: If this is fine, then don't throw, just log a note about this.
I really don't like debug / release different behaviors. They lead us to miss important issues.

ayende (approved):

ppekrol (approved):

PR #10338, opened 2020-03-11
RavenDB-14745 HTTP Client is cached even if the server callback is ch…

ayende:
…anged

### DIFF
--- a/src/Raven.Client/Http/RequestExecutor.cs
+++ b/src/Raven.Client/Http/RequestExecutor.cs
@@ -1873,6 +1873,8 @@ namespace Raven.Client.Http
                 lock (_locker)
                 {
                     _serverCertificateCustomValidationCallback = _serverCertificateCustomValidationCallback.Concat(new[] { value }).ToArray();
+                    GlobalHttpClientWithCompression.Clear();
+                    GlobalHttpClientWithoutCompression.Clear();
                 }
             }
 
@@ -1884,6 +1886,8 @@ namespace Raven.Client.Http
                 lock (_locker)
                 {
                     _serverCertificateCustomValidationCallback = _serverCertificateCustomValidationCallback.Except(new[] { value }).ToArray();
+                    GlobalHttpClientWithCompression.Clear();
+                    GlobalHttpClientWithoutCompression.Clear();
                 }
             }
         }


src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1873,6 +1873,8 @@ public bool Translate(object sender, X509Certificate cert, X509Chain chain, SslP
                 lock (_locker)
                 {
                     _serverCertificateCustomValidationCallback = _serverCertificateCustomValidationCallback.Concat(new[] { value }).ToArray();
+                    GlobalHttpClientWithCompression.Clear();
```
ppekrol, line 1876: I do not see how does it help? Handler is attaching to `OnServerCertificateCustomValidationCallback` method, which is using the `_serverCertificateCustomValidationCallback`. Can you explain?

ayende: If you already have an `HttpClient` for that thumbprint, you'll keep using this.
And will never re-create the `HttpHandler`, even if you create new `RequestExectuers`

ppekrol: But why you need to recreate it?

ayende: We are adding the callback only when we create the http handler, and only when creating the request executer.

ayende:
Let's hold on merging this, I want to see if this solve the actual problem on the customer's end

georgiosd:
@ayende @ppekrol do you remember what this was about?

I am seeing a problem where:
- I replace the callback to validate a self-signed cert
- Some requests seems to work fine
- I then try to create a database, with encryption, if it matters
- My callback is NOT executed, I get the mismatch error thrown

It seems that under some circumstances the callback is not registered?

ayende:
You have likely *already* created that http client, in that case.

At which point do you register this callback?
It should happen as the very first thing in the system.

georgiosd:
I even tried registering the callback at the beginning of `Program.cs` so something seems off.

I wanted to try and send a repro but I need to set up a cluster to do it (though not 100% sure it's related to encryption at all).

PR #10340, opened 2020-03-11
RavenDB-13239 Admin Logs - Add 'copy log message' button

Danielle9897:


src/Raven.Studio/typescript/viewmodels/manage/adminLogs.ts
```diff
@@ -210,6 +213,12 @@ class adminLogs extends viewModelBase {
             
             this.duringManualScrollEvent = false;
         });
+      
+        $( ".list-view" ).on( "click", "a", function(event) {
```
ml054: remove space after (

ml054: 2x

src/Raven.Studio/typescript/viewmodels/manage/adminLogs.ts
```diff
@@ -210,6 +213,12 @@ class adminLogs extends viewModelBase {
...
+            event.preventDefault();
+            event.stopImmediatePropagation();
+            copyToClipboard.copy($(this).parent().children().last().text(), "Log message has been copied to clipboard");
```
ml054: I think we should use better selector  - this is really fragile

src/Raven.Studio/typescript/viewmodels/manage/adminLogs.ts
```diff
@@ -194,9 +194,12 @@ class adminLogs extends viewModelBase {
     
     // noinspection JSMethodCanBeStatic
     itemHtmlProvider(item: string) {
+        const errorClass = this.hasError(item) ? "class='bg-danger'" : "";
+
         return $("<pre class='item'></pre>")
-            .toggleClass("bg-danger", this.hasError(item))
-            .text(item);
+            .addClass("flex-horizontal")
+            .prepend(`<span ${errorClass}>${item}</span>`)
```
ml054: do we need to escaping something here?

ml054 (changes requested):

ml054:
+ as we discuss:
- bring back red colors with shades when errors
- remove double scrolls when error items

src/Raven.Studio/typescript/viewmodels/manage/adminLogs.ts
Danielle9897: done - and also:
now using `$(".list-view").on("click", ".copy-item-button", function(event) { ....`
(changed class name in other places where needed)

src/Raven.Studio/typescript/viewmodels/manage/adminLogs.ts
Danielle9897: now using: `$(this).next().text()`

src/Raven.Studio/typescript/viewmodels/manage/adminLogs.ts
Danielle9897: no - I don't think we need to

src/Raven.Studio/typescript/viewmodels/manage/adminLogs.ts
ml054: can we use $("span", this).text() ?

src/Raven.Studio/typescript/viewmodels/manage/adminLogs.ts

src/Raven.Studio/typescript/viewmodels/manage/adminLogs.ts
Danielle9897: The dom structure per item is:

```
<div>
    <pre>
        <a>the copy button</a>
        <span>the text to copy</span>
    </pre>
</div>
```

`$(this)` is the `<a>` element that was clicked on - it has no text on it.
so I go to `$(this).next()` - which is the `<span>` that has the text.

ml054: oh ok, so `$(this).next().text()` should be good

src/Raven.Studio/typescript/viewmodels/manage/adminLogs.ts
ml054: yes - this is nice and less fragile

ml054 (approved):

PR #10344, opened 2020-03-13
RavenDB-14757 Update PAL / Sodium to use new form of finding the dll,…

ayende:
… instead of copying


src/Sparrow.Server/Sparrow.Server.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.1</TargetFrameworks>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
ayende, line 3: @ppekrol I'm pretty sure that this is fine to do. Can you verify?

src/Sparrow.Server/DynamicNativeLibraryResolver.cs
```diff
@@ -0,0 +1,62 @@
...
+                }
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                suffix = Environment.Is64BitProcess ? ".mac.x64.dylib" : ".mac.x86.dylib";
+                // in mac we are not : `toFilename += ".so";` as DllImport doesn't assume .so nor .dylib by default
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                suffix = Environment.Is64BitProcess ? ".win.x64.dll" : ".win.x86.dll";
```
ayende, line 51: Missing handling of `win7` here, which is now out of support.  Not sure if we want to remove the `win7` support or keep it. Thoughts?

ayende, line 51: Missing handling of `win7` here, which is now out of support.  Not sure if we want to remove the `win7` support or keep it. Thoughts?

src/Sparrow.Server/DynamicNativeLibraryResolver.cs
```diff
@@ -0,0 +1,62 @@
...
+            }
+            else
+            {
+                throw new NotSupportedException("Non supported platform - no Linux/OSX/Windows is detected ");
+            }
+
+            if (NativeLibrary.TryLoad(libraryName + suffix, assembly, dllImportSearchPath, out var ptr) == false)
+            {
```
ppekrol: Don't we want to throw here?

src/Sparrow.Server/DynamicNativeLibraryResolver.cs
ppekrol: From one side yes, from the other I have a feeling that it might still be popular in come circles. There is extended support option available? Or Windows Server equivalent that has still support?

src/Sparrow.Server/Sparrow.Server.csproj
ppekrol: Should be fine. It is used only in Raven.Server which is a netcoreapp3.1 project.

src/Sparrow.Server/Platform/Pal.cs
```diff
@@ -16,67 +16,17 @@ public static unsafe class Pal
...
-            }
-            else
-            {
-                throw new NotSupportedException("Not supported platform - no Windows/OSX/Linux is detected ");
-            }
-
-            var retries = 0;
-            while (true)
-            {
-                var e = CopyPalFile(toFilename, fromFilename);
```
ppekrol, line 55: If I understand this correctly, by using new method we do not need to do any copying. Double checking.

src/Sparrow.Server/DynamicNativeLibraryResolver.cs
ayende: As long as .NET core supports it, it isn't hard to support it too

src/Sparrow.Server/DynamicNativeLibraryResolver.cs
ayende: Yes

src/Sparrow.Server/Platform/Pal.cs
ayende: Yes, this is no longer needed

ayende:
Ready for final review now

src/Sparrow.Server/Platform/Pal.cs
```diff
@@ -16,67 +16,18 @@ public static unsafe class Pal
...
-            }
-
-            PalFlags.FailCodes rc = PalFlags.FailCodes.None;
+            PalFlags.FailCodes rc;
             int errorCode;
             try
             {
-                var palver = rvn_get_pal_ver();
-                if (palver != 0 && palver != PAL_VER)
+                var mutator = PlatformDetails.IsWindows8OrNewer == false ? (Func<string,string>)ToWin7DllName : default;
```
ppekrol, line 23: instead of using mutator we can use PlatformDetails inside the DynamicNativeLibraryResolver, we will not need the mutator at all then (maybe rename to SodiumNativeLibraryResolver?)

ayende: The problem is that we don't have `win7` for Sodium (and won't for zstd)

ppekrol (approved):

PR #10345, opened 2020-03-15
RavenDB-14743: Add TransferEncodingChunked = true to smuggler HttpReq…

garayx:
…uest, propagate smuggler task to operation

### DIFF
--- a/src/Raven.Client/Documents/Operations/Operation.cs
+++ b/src/Raven.Client/Documents/Operations/Operation.cs
@@ -248,7 +248,7 @@ namespace Raven.Client.Documents.Operations
                         StopProcessing();
                         if (_additionalTask.IsFaulted)
                         {
-                            _result.TrySetException(_additionalTask.Exception.ExtractSingleInnerException());
+                            _result.TrySetException(_additionalTask.Exception);
                             break;
                         }
 
--- a/src/Raven.Client/Documents/Operations/Operation.cs
+++ b/src/Raven.Client/Documents/Operations/Operation.cs
@@ -61,21 +61,7 @@ namespace Raven.Client.Documents.Operations
             }
             catch (Exception e)
             {
-                await _lock.WaitAsync().ConfigureAwait(false);
-
-                try
-                {
-                    StopProcessing();
-                }
-                catch
-                {
-                    // ignoring
-                }
-                finally
-                {
-                    _result.TrySetException(e);
-                    _lock.Release();
-                }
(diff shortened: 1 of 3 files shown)


src/Raven.Client/Documents/Operations/Operation.cs
```diff
@@ -246,6 +246,12 @@ public void OnNext(OperationStatusChange change)
                         break;
                     case OperationStatus.Faulted:
                         StopProcessing();
+                        if (_additionalTask.IsFaulted)
+                        {
+                            _result.TrySetException(_additionalTask.Exception.ExtractSingleInnerException());
```
ayende:
```suggestion
                            _result.TrySetException(_additionalTask.Exception);
```

`.ExtractSingleInnerException()` should be on the caller, not here. We don't want to possibly lose information for the caller

test/EmbeddedTests/SmugglerTests.cs
```diff
@@ -0,0 +1,197 @@
...
+                    using (store.SetRequestTimeout(TimeSpan.FromMilliseconds(1000), databaseName))
+                    {
+                        try
+                        {
+                            var operation = await store.Smuggler.ForDatabase(databaseName).ImportAsync(new DatabaseSmugglerImportOptions
+                            {
+                                OperateOnTypes = DatabaseItemType.Documents | DatabaseItemType.Identities | DatabaseItemType.CompareExchange,
+                            }, "dump.cmp");
+
+                            await operation.WaitForCompletionAsync();
```
ppekrol, line 62: You are not failing the test here. After this line you should add an assert or throw that it should not happen

test/EmbeddedTests/SmugglerTests.cs
```diff
@@ -0,0 +1,197 @@
...
+            using (var embedded = new EmbeddedServer())
+            {
+                embedded.StartServer(new ServerOptions
+                {
+                    ServerDirectory = paths.ServerDirectory,
+                    DataDirectory = paths.DataDirectory,
+                });
+
+                const string databaseName = "test";
+                var dummyDump = CreateDummyDump(100000);
```
ppekrol: Is there a smaller version of this test available?

src/Raven.Client/Documents/Operations/Operation.cs
ppekrol: Have you pushed this?

garayx: yes

ppekrol (approved):

PR #10350, opened 2020-03-16
RavenDB-14459 Allow rerun batch command with store attachments

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-14459

### DIFF
--- a/src/Raven.Server/Documents/Handlers/BatchHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/BatchHandler.cs
@@ -263,7 +263,7 @@ namespace Raven.Server.Documents.Handlers
 
                 if (command.AttachmentStreams == null)
                 {
-                    command.AttachmentStreams = new Queue<MergedBatchCommand.AttachmentStream>();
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -742,9 +742,11 @@ protected override long ExecuteCmd(DocumentsOperationContext context)
 
                 DocumentsStorage.PutOperationResults? lastPutResult = null;
 
+                using IEnumerator<AttachmentStream> attachmentIterator = AttachmentStreams?.GetEnumerator();
+                
                 for (int i = ParsedCommands.Offset; i < ParsedCommands.Count; i++)
                 {
-                    var cmd = ParsedCommands.Array[ParsedCommands.Offset + i];
```
haludi, line 747: It is not related to the main issue but I think it is wrong (It probably worked just because the offset was always zero)

ayende: Yes, that is fine, because we start from `Offset`

ayende (approved):

PR #10356, opened 2020-03-18
RavenDB-14543 Allow define FindTypeNameForDynamic delegate for store dynamic object

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-14543


src/Raven.Client/Documents/Conventions/DocumentConventions.cs
```diff
@@ -491,6 +492,19 @@ public IContractResolver JsonContractResolver
                 _findCollectionNameForDynamic = value;
             }
         }
+        
+        /// <summary>
+        ///     Gets or sets the function to find the collection name for dynamic type.
+        /// </summary>
+        public Func<dynamic, string> FindTypeNameForDynamic
```
ppekrol: Shouldn't this be `FindClrTypeNameForDynamic` to match the other convention `FindClrTypeName` ?

PR #10358, opened 2020-03-19
RavenDB-13478 allow to include counters in subscriptions

ppekrol:


src/Raven.Server/Documents/TcpHandlers/SubscriptionConnection.cs
```diff
@@ -875,7 +900,7 @@ private SubscriptionPatchDocument SetupFilterAndProjectionScript()
 
         public void Dispose()
         {
-            this.AddToStatusDescription($"Disposed at {Environment.StackTrace}"); //todo: remove that
+            AddToStatusDescription($"Disposed at {Environment.StackTrace}"); //todo: remove that
```
ayende: Need to remove this, no?

ayende (changes requested):

src/Raven.Server/Documents/TcpHandlers/SubscriptionConnection.cs
ppekrol: I will remove that in 4.2

PR #10361, opened 2020-03-20
RavenDB-14779 - allow to provide a certificate for one way replication to a v4.x server

grisha-kotler:


Raven.Database/Config/InMemoryRavenConfiguration.cs
```diff
@@ -1588,6 +1590,10 @@ public class ReplicationConfiguration
             /// Indicates how many seconds replication task will wait before propagating replication documents
             /// </summary>
             public int ReplicationPropagationDelayInSeconds { get; set; }
+
+            public string ReplicationToV4CertificatePath { get; set; }
```
ppekrol: Can we find a better name? What if we will have V5?

grisha-kotler: done

PR #10367, opened 2020-03-22
RavenDB-14781 Ensure certificate serial number is exactly 20 bytes in…

iftahbe:
… length

### DIFF
--- a/src/Raven.Server/Utils/CertificateUtils.cs
+++ b/src/Raven.Server/Utils/CertificateUtils.cs
@@ -186,8 +186,21 @@ namespace Raven.Server.Utils
                     new X509KeyUsage(X509KeyUsage.KeyCertSign | X509KeyUsage.CrlSign));
             }
 
-            // Serial Number
-            var serialNumber = new BigInteger(20 * BitsPerByte, random);
+            // Serial Number - we must generate a positive number, so retrying if it's not
+            BigInteger serialNumber;
+            var retries = 10;
+
+            while (true)
+            {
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Utils/CertificateUtils.cs
```diff
@@ -186,8 +186,21 @@ public static X509Certificate2 CreateSelfSignedExpiredClientCertificate(string c
...
+            // Serial Number - we must generate a positive number, so retrying if it's not
+            BigInteger serialNumber;
+            var retries = 10;
+
+            while (true)
+            {
+                var serialNumberBytes = new byte[20];
+                random.NextBytes(serialNumberBytes);
+                serialNumber = new BigInteger(serialNumberBytes);
+                if (serialNumber.SignValue > 0)
```
ayende: Huh? Why do we need this like that?

iftahbe: If the generated bytes form a negative number, certificateGenerator.SetSerialNumber() will throw, so I'm making sure random gives us a positive number.

ayende: Then why regenerate the number? flip the sign

https://github.com/bcgit/bc-csharp/blob/master/crypto/src/math/BigInteger.cs#L721

iftahbe: Ooh that's nice, I missed this method.

ayende (approved):

PR #10371, opened 2020-03-22
RavenDB-14769 Querying Search as Subclause return wrong results

ayende:


src/Raven.Server/Documents/Queries/QueryBuilder.cs
```diff
@@ -292,6 +292,16 @@ private static MethodExpression FindMoreLikeThisExpression(QueryExpression expre
             }
             if (expression is NegatedExpression ne)
             {
+                // 'not foo and bar' should be parsed as:
+                // (not foo) and bar, instead of not (foo and bar)
+                if (ne.Expression is BinaryExpression nbe)
+                {
```
ppekrol, line 301: I think this should only be applied when `nbe.Operator` is And or Or.

Also this changes:
`Not ( <Field>: SomeProperty NotEqual p0 (Parameter) And <Field>: SomeProperty NotEqual p1 (Parameter))`
into
`Not ( <Field>: SomeProperty NotEqual p0 (Parameter)) And <Field>: SomeProperty NotEqual p1 (Parameter)`

which I do not think is correct? (running SlowTests.Tests.Linq.IsNullOrEmpty tests)

ayende: You are correct, fixed now

ppekrol (approved):

arekpalinski (approved):

PR #10372, opened 2020-03-23
RavenDB-14760 Allow add socket or attach pipe without changing log file mode

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-14760

### DIFF
--- a/src/Sparrow/Logging/LogMessageEntry.cs
+++ b/src/Sparrow/Logging/LogMessageEntry.cs
@@ -7,6 +7,7 @@ namespace Sparrow.Logging
 {
     internal class LogMessageEntry
     {
+        public LogMode Type { get; set; }
         public MemoryStream Data;
         public TaskCompletionSource<object> Task;
 
--- a/src/Sparrow/Logging/LogMessageEntry.cs
+++ b/src/Sparrow/Logging/LogMessageEntry.cs
@@ -7,7 +7,7 @@ namespace Sparrow.Logging
 {
     internal class LogMessageEntry
     {
-        public LogMode Type { get; set; }
+        public LogMode Type;
         public MemoryStream Data;
         public TaskCompletionSource<object> Task;
 
--- a/src/Sparrow/Logging/LoggingSource.cs
+++ b/src/Sparrow/Logging/LoggingSource.cs
@@ -55,8 +55,11 @@ namespace Sparrow.Logging
         private readonly MultipleUseFlag _keepLogging = new MultipleUseFlag(true);
         private int _logNumber;
         private DateTime _today;
-        public bool IsInfoEnabled;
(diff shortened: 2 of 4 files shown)


src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -85,21 +88,24 @@ public class WebSocketContext
         public long RetentionSize { get; private set; }
         public bool Compressing => _compressLoggingThread != null;
 
-        private LogMode _oldLogMode;
 
+        private (bool Info, bool Operation) CalculateIsLogEnabled(LogMode? logMode = null)
+        {
+            logMode ??= LogMode;
```
ayende, line 97: `if (_listeners.IsEmpty == false || _pipeSink != null) return (true, true);`

src/Sparrow/Logging/LogMessageEntry.cs
```diff
@@ -7,6 +7,7 @@ namespace Sparrow.Logging
 {
     internal class LogMessageEntry
     {
+        public LogMode Type { get; set; }
```
ppekrol: Let's be consistent and convert this to a field

src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -809,20 +814,39 @@ private void BackgroundLoggerCompress()
...
 
-            file.Write(bytes.Array, bytes.Offset, bytes.Count);
-            _additionalOutput?.Write(bytes.Array, bytes.Offset, bytes.Count);
+            if (item.Type == LogMode.Operations && LogMode != LogMode.None || (LogMode & LogMode.Information) == LogMode.Information)
+            {
+                file.Write(bytes.Array, bytes.Offset, bytes.Count);
+                _additionalOutput?.Write(bytes.Array, bytes.Offset, bytes.Count);
+            }
 
             if (item.Task != null)
```
ppekrol: Shouldn't we move this inside the `if` now?

src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -221,43 +229,41 @@ void Restart()
...
-
-            _keepLogging.Raise();
-            _loggingThread = new Thread(BackgroundLogger)
-            {
-                IsBackground = true,
-                Name = _name + " Thread"
-            };
-            _loggingThread.Start();
-            if (compress)
+            lock (this)
```
haludi, line 232: While testing I ran to a situation when the logging thread run `SetupLogMode` it will run the restart on another thread.
So the `StartNewLoggingThreads` running without lock so someone else can  run `SetupLogMode` simultaneously

ppekrol (approved):

ayende (approved):

PR #10373, opened 2020-03-23
RavenDB-14713 Add alert on many calls from process for topology in a …

ayende:
…short time range


ppekrol:
This introduces a lot of binary-level compatibility issues. What I would suggest is:
1) Move GlobalApplicationIdentifier to RequestExecutor
2) Create a property, similar to `_disableClientConfigurationUpdates` and set it accordingly for relevant code paths or pass it via DocumentConvention (can be internal)

This way we will do the same thing, but handling will be easier, because it will be based on one value internally

ayende:
Updated

ppekrol (approved):

PR #10374, opened 2020-03-23
RavenDB-14594 Allow to change the 5 years expiration on client certificates

iftahbe:


src/Raven.Client/ServerWide/Operations/Certificates/CertificateDefinition.cs
```diff
@@ -12,7 +12,7 @@ public class CertificateDefinition
         public SecurityClearance SecurityClearance;
         public string Thumbprint;
         public string PublicKeyPinningHash;
-        public DateTime NotAfter;
+        public DateTime? NotAfter;
```
iftahbe, line 15: @ml054 I added this for the certificate view in the Studio, if NotAfter is null the default expiration is still 5 years so nothing is broken and you can add the UI later.

src/Raven.Server/Commercial/SetupInfo.cs
```diff
@@ -16,6 +16,7 @@ public class SetupInfo
         public bool EnableExperimentalFeatures { get; set; }
         public StudioConfiguration.StudioEnvironment Environment { get; set; }
         public bool RegisterClientCert { get; set; }
+        public DateTime? ClientCertNotAfter { get; set; }
```
iftahbe, line 19: @ml054 and this is for the setup wizard, if left null default is 5 years

src/Raven.Client/ServerWide/Operations/Certificates/CertificateDefinition.cs
ppekrol: @iftahbe when we are putting that into storage (in PutCertificateCommand for example - where else?) can this be null? If not, then maybe let's add a validation there?

iftahbe: There's also PutCertificateWithSamePinningHashCommand.
Yeah I'll add validation.
Also, I just realized this needs to be added to the cli as well.

iftahbe: There's also PutLocalState which is for when we are passive.

ppekrol (approved):

PR #10376, opened 2020-03-23
RavenDB-14596: Handle Exceptions when DB gets deleted/disabled while loading

garayx:


grisha-kotler (approved):

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -312,6 +312,17 @@ public void DeleteDatabase(string dbName, DeletionInProgressStatus deletionInPro
                     // this is already in the process of being deleted, we can just exit and let another thread handle it
                     return;
                 }
+                catch (DatabaseConcurrentLoadTimeoutException e)
+                {
+                    if (e.Caller == nameof(DeleteDatabase))
```
ppekrol: Use e.Data for this.

https://github.com/ravendb/ravendb/blob/69acc524c4bfc80cd9aa1bd6a45b361b091d49a7/src/Raven.Server/Documents/DatabasesLandlord.cs#L643

PR #10378, opened 2020-03-23
RavenDB-14782 Provide a good way to export a lucene index to disk so …

ayende:
…we can analyze it using standard tools


ppekrol:
What is the difference between this and https://github.com/ravendb/ravendb/pull/9881 ?
Also the 5.0 one requires DatabaseAdmin only

ayende:
Already have this feature. https://github.com/ravendb/ravendb/pull/9881

ppekrol:
Please note that the https://github.com/ravendb/ravendb/pull/9881 is a 5.0 PR. Do we want to backport it?

PR #10382, opened 2020-03-24
RavenDB-13286 - Support bulk insert for counters

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-13286

### DIFF
--- a/src/Raven.Client/Documents/BulkInsert/BulkInsertOperation.cs
+++ b/src/Raven.Client/Documents/BulkInsert/BulkInsertOperation.cs
@@ -8,6 +8,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
 using Raven.Client.Documents.Commands;
+using Raven.Client.Documents.Commands.Batches;
 using Raven.Client.Documents.Conventions;
 using Raven.Client.Documents.Identity;
 using Raven.Client.Documents.Operations;
@@ -139,6 +140,7 @@ namespace Raven.Client.Documents.BulkInsert
         private Stream _stream;
         private readonly StreamExposerContent _streamExposerContent;
         private bool _first = true;
+        private CommandType _continousCommandType;
         private long _operationId = -1;
 
         public CompressionLevel CompressionLevel = CompressionLevel.NoCompression;
@@ -152,6 +154,9 @@ namespace Raven.Client.Documents.BulkInsert
             {
                 try
                 {
+                    if (_continousCommandType != CommandType.None)
+                        ThrowOnMissingOperationDispose();
+
                     Exception flushEx = null;
 
                     if (_stream != null)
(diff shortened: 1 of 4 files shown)


src/Raven.Client/Documents/BulkInsert/BulkInsertOperation.cs
```diff
@@ -488,5 +519,106 @@ private string GetId(object entity)
             _generateEntityIdOnTheClient.TrySetIdentity(entity, id); //set Id property if it was null
             return id;
         }
+
+        public CountersBulkInsert CountersFor(string id)
```
ayende, line 522: I don't like this API. In this case, we want to prioritize performance.
The common case is for us to use many counters for many documents, so we'll be creating a lot of those instances.
Better to use API such as: `IncrementAsync(string docId, string name, long delta)`

src/Raven.Server/Documents/Handlers/BulkInsertHandler.cs
```diff
@@ -143,6 +147,46 @@ private async Task<IOperationResult> DoBulkInsert(Action<IOperationProgress> onP
...
+            int GetChangeVectorSize()
+            {
+                if (_changeVectorSize.HasValue)
+                    return _changeVectorSize.Value;
+
+                using (documentDatabase.DocumentsStorage.ContextPool.AllocateOperationContext(out DocumentsOperationContext ctx))
+                using (ctx.OpenReadTransaction())
+                {
+                    var databaseChangeVector = DocumentsStorage.GetDatabaseChangeVector(ctx);
+                    _changeVectorSize = Encoding.UTF8.GetBytes(databaseChangeVector).Length;
```
ayende, line 184: This is *really* expense. And you are allocating here.
Why do you need it? It would be sufficent to just call `changeVector.Length *2`, which should be a great estimation.

src/Raven.Client/Documents/BulkInsert/BulkInsertOperation.cs
```diff
@@ -305,76 +297,116 @@ public async Task StoreAsync(object entity, string id, IMetadataDictionary metad
...
-                        _asyncWrite = _requestBodyStream.WriteAsync(buffer.Array, buffer.Offset, buffer.Count, _token);
-                    }
+        private async Task WriteToStream(Action writeOperation, string documentId, CommandType currentCommand)
+        {
+            if (_continousCommandType != CommandType.None && currentCommand != _continousCommandType)
+                ThrowUnfinishedCommand(currentCommand);
+
+            try
+            {
+                writeOperation();
```
ayende: I don't particularly love this API. 
This force us to do allocations for the delegate on each call, which we would like to avoid.
Can we avoid this?

ayende (changes requested):

src/Raven.Server/Documents/Handlers/BulkInsertHandler.cs
grisha-kotler: I only call it once per the entire BulkInsert operation (the size is cached).

src/Raven.Client/Documents/BulkInsert/BulkInsertOperation.cs
grisha-kotler: Done

ayende (approved):

ppekrol (approved):

PR #10383, opened 2020-03-24
RavenDB-14006 compare exchange includes

ppekrol:


ppekrol:
TODO:

- support for queries

src/Raven.Client/Documents/Session/AbstractDocumentQuery.Includes.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            if (includes.DocumentsToInclude != null)
+            {
+                foreach (var doc in includes.DocumentsToInclude)
+                {
+                    DocumentIncludes.Add(doc);
+                }
+            }
+
+            IncludeCounters(includes.Alias, includes.CountersToIncludeBySourcePath);
```
ayende, line 55: Can't `CountersToIncludeBySourcePath` be null?

ayende (approved):
Note that we are missing JS API for this, though.
I think that is relevant for counters, timeseries, etc. 
We need a way to call that from JS, because that is the escape hatch fro complexity

src/Raven.Client/Documents/Session/AbstractDocumentQuery.Includes.cs
ppekrol: I just moved the old code, but the method 'IncludeCounters' checks that internally.

ppekrol:
I've added a TODO for this, this can be merged I think

PR #10384, opened 2020-03-25
RavenDB-14699 Expose LoadRequestTimeoutInSec configuration option for…

Danielle9897:
… Raven ETL task

### DIFF
--- a/src/Raven.Studio/wwwroot/App/views/database/tasks/editRavenEtlTask.html
+++ b/src/Raven.Studio/wwwroot/App/views/database/tasks/editRavenEtlTask.html
@@ -95,6 +95,26 @@
                                 <span class="help-block" data-bind="validationMessage: connectionStringName"></span>
                             </div>
                         </div>
+                        <div class="flex-horizontal">
+                            <div class="flex-grow">
+                                <button type="button" class="btn btn-info" data-bind="click: $root.toggleAdvancedArea">
+                                    <i class="icon-settings"></i><span>Advanced</span>
+                                </button>
+                            </div>
+                        </div>
+                        <div data-bind="collapse: $root.showAdvancedOptions">
+                            <div class="panel-body">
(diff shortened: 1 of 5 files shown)


src/Raven.Studio/wwwroot/App/views/database/tasks/editRavenEtlTask.html
```diff
@@ -95,6 +95,26 @@ <h3 data-bind="text: $parent.isAddingNewRavenEtlTask() ? `New RavenDB ETL` : `Ed
...
+                                </button>
+                            </div>
+                        </div>
+                        <div data-bind="collapse: $root.showAdvancedOptions">
+                            <div class="panel-body">
+                                <div class="form-group" data-bind="validationElement: loadRequestTimeout">
+                                    <label for="loadTimeout" class="control-label">Http Request Timeout</label>
+                                    <div class="flex-grow">
+                                        <div class="input-group">
+                                            <input type="number" min="0" class="form-control" placeholder="Use HttpClient default timeout (100s)" id="LoadTimeout" data-bind="numericInput: loadRequestTimeout" autocomplete="off">
```
ml054: for and id are not bound correctly - loadTimeout vs LoadTimeout

ml054: Do we need to say that default timeout is HttpClient timeout?

Danielle9897: 1. Done - unified for and id
2. Probably can do without explicitly saying 'HttpClient' and just say: '_Use default timeout (100s_)'
    I just needed to have some text similar to the sql-etl task where we have: '_Use SQL provider default value_'

arekpalinski: 2. Yeah, we can drop 'HttpClient' there

ml054 (approved):

PR #10386, opened 2020-03-25
RavenDB-14651 Rework time-series Client API

aviv86:

- change from `TimeSeriesFor(id)` to `TimeSeriesFor(id, name)`
- make 'tag' an optional parameter

### DIFF
--- a/src/Raven.Client/Documents/Session/AsyncSessionDocumentTimeSeries.cs
+++ b/src/Raven.Client/Documents/Session/AsyncSessionDocumentTimeSeries.cs
@@ -18,20 +18,20 @@ namespace Raven.Client.Documents.Session
 {
     public class AsyncSessionDocumentTimeSeries : SessionTimeSeriesBase, IAsyncSessionDocumentTimeSeries
     {
-        public AsyncSessionDocumentTimeSeries(InMemoryDocumentSessionOperations session, string documentId) : base(session, documentId)
+        public AsyncSessionDocumentTimeSeries(InMemoryDocumentSessionOperations session, string documentId, string name) : base(session, documentId, name)
         {
         }
 
-        public AsyncSessionDocumentTimeSeries(InMemoryDocumentSessionOperations session, object entity) : base(session, entity)
+        public AsyncSessionDocumentTimeSeries(InMemoryDocumentSessionOperations session, object entity, string name) : base(session, entity, name)
         {
         }
 
-        public async Task<IEnumerable<TimeSeriesEntry>> GetAsync(string timeseries, DateTime from, DateTime to, int start = 0, int pageSize = int.MaxValue, CancellationToken token = default)
+        public async Task<IEnumerable<TimeSeriesEntry>> GetAsync(DateTime from, DateTime to, int start = 0, int pageSize = int.MaxValue, CancellationToken token = default)
         {
             TimeSeriesDetails details;
-            
+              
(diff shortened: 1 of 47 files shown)


src/Raven.Client/Documents/Session/AsyncSessionDocumentTimeSeries.cs
```diff
@@ -18,20 +18,20 @@ namespace Raven.Client.Documents.Session
...
         {
         }
 
-        public AsyncSessionDocumentTimeSeries(InMemoryDocumentSessionOperations session, object entity) : base(session, entity)
+        public AsyncSessionDocumentTimeSeries(InMemoryDocumentSessionOperations session, object entity, string name) : base(session, entity, name)
         {
         }
 
-        public async Task<IEnumerable<TimeSeriesEntry>> GetAsync(string timeseries, DateTime from, DateTime to, int start = 0, int pageSize = int.MaxValue, CancellationToken token = default)
+        public async Task<IEnumerable<TimeSeriesEntry>> GetAsync(DateTime from, DateTime to, int start = 0, int pageSize = int.MaxValue, CancellationToken token = default)
```
karmeli87, line 29: Do we want to have also the `from` and `to` as optional?

src/Raven.Client/Documents/Session/ISessionDocumentTimeSeriesBase.cs
```diff
@@ -15,19 +15,24 @@ namespace Raven.Client.Documents.Session
...
         /// </summary>
-        void Append(string timeseries, DateTime timestamp, string tag, IEnumerable<double> values);
+        void Append(DateTime timestamp, IEnumerable<double> values, string tag = null);
 
         /// <summary>
-        /// Remove all the values in the timeseries in the range of from .. to.
+        /// Append a single value (and optional tag) to the times series at the provided time stamp
         /// </summary>
-        void Remove(string timeseries, DateTime from, DateTime to);
+        void Append(DateTime timestamp, double value, string tag = null);
```
karmeli87, line 25: What do you think about `Append` without `DateTime` something like `AppendNow`? Do we want something like that?
It can use either server or client time.

src/Raven.Client/Documents/Session/AsyncSessionDocumentTimeSeries.cs
ayende: No, they aren't optional, they are required, and we want to make the user specify it

src/Raven.Client/Documents/Session/ISessionDocumentTimeSeriesBase.cs
ayende: No, exactly because this is so hard to figure out, let's do that explicitly. 
User need to call `DateTime.UtcNow`, not really hard

ayende (approved):

src/Raven.Client/Documents/Session/ISessionDocumentTimeSeriesBase.cs
karmeli87: Just to note that if we put server's time we avoid conflicts on the the time-stamp and can handle clock/time issues.
But I guess if needed we can add this later.

ayende: Not really a good idea. Because now you added network lag to the time we measure.

PR #10387, opened 2020-03-26
RavenDB-14651 Rework time-series Client API - JS patching

aviv86:

### DIFF
--- a/src/Raven.Server/Documents/Patch/ScriptRunner.cs
+++ b/src/Raven.Server/Documents/Patch/ScriptRunner.cs
@@ -295,6 +295,8 @@ namespace Raven.Server.Documents.Patch
 
             private JsValue TimeSeries(JsValue self, JsValue[] args)
             {
+                AssertValidDatabaseContext("timeseries(doc, name)");
+
                 var originalArgs = new ArrayInstance(ScriptEngine);
                 originalArgs.FastAddProperty("length", 0, true, false, false);
                 ScriptEngine.Array.PrototypeObject.Push(originalArgs, args);
@@ -318,17 +320,18 @@ namespace Raven.Server.Documents.Patch
 
             private JsValue AppendTimeSeries(JsValue document, JsValue name, JsValue[] args)
             {
-                AssertValidDatabaseContext("append");
-                const string signature4Args = "appendTimeSeries(doc, timeseries, timestamp, values)";
-                const string signature5Args = "appendTimeSeries(doc, timeseries, timestamp, tag, values)";
+                AssertValidDatabaseContext("timeseries(doc, name).append");
+
+                const string signature2Args = "append(timestamp, values)";
+                const string signature3Args = "append(timestamp, values, tag)";
                 
                 string signature;
                 LazyStringValue lsTag = null;
                 switch (args.Length)
                 {
-                    case 2: signature = signature4Args;
+                    case 2: signature = signature2Args;
                         break;
-                    case 3: signature = signature5Args;
+                    case 3: signature = signature3Args;
                         var tagArgument = args.Last();
                         if (tagArgument != null && tagArgument.IsNull() == false && tagArgument.IsUndefined() == false)
                         {
@@ -336,7 +339,7 @@ namespace Raven.Server.Documents.Patch
                             lsTag = _jsonCtx.GetLazyString(tag);
                         }
                         break;
-                    default: throw new ArgumentException($"There is no overload with {args.Length} arguments for this method should be {signature4Args} or {signature5Args}");
+                    default: throw new ArgumentException($"There is no overload with {args.Length} arguments for this method should be {signature2Args} or {signature3Args}");
                 }
                 
                 var (id, doc) = GetIdAndDocFromArg(document, _timeSeriesSignature);
@@ -393,7 +396,7 @@ namespace Raven.Server.Documents.Patch
 
             private JsValue DeleteRangeTimeSeries(JsValue document, JsValue name, JsValue[] args)
             {
-                AssertValidDatabaseContext("deleteTimeSeries");
+                AssertValidDatabaseContext("timeseries(doc, name).remove");
 
                 const string signature = "remove(from, to)";
                 const int requiredArgs = 2;
@@ -423,8 +426,8 @@ namespace Raven.Server.Documents.Patch
             
             private JsValue GetRangeTimeSeries(JsValue document, JsValue name, JsValue[] args)
             {
-                AssertValidDatabaseContext("getRangeTimeSeries");
-                
+                AssertValidDatabaseContext("timeseries(doc, name).get");
+
                 const string signature = "get(from, to)";
                 const int requiredArgs = 2;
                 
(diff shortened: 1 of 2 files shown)


src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -293,39 +292,65 @@ private void FillDoubleArrayFromJsArray(double[] array, ArrayInstance jsArray, s
                     ++i;
                 }
             }
-            
-            private JsValue AppendTimeSeries(JsValue self, JsValue[] args)
+
+            private JsValue TimeSeries(JsValue self, JsValue[] args)
+            {
+                AssertValidDatabaseContext("timeseries(doc, name)");
```
ppekrol: Do we want to have `timeseries(name)` also and Id would be taken (IIRC) from `SingleRun.OriginalDocumentId` ?

ppekrol (approved):

src/Raven.Server/Documents/Patch/ScriptRunner.cs
ayende:
```suggestion
                AssertValidDatabaseContext(_timeSeriesSignature );
```

src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -293,39 +292,65 @@ private void FillDoubleArrayFromJsArray(double[] array, ArrayInstance jsArray, s
...
+
+                var originalArgs = new ArrayInstance(ScriptEngine);
+                originalArgs.FastAddProperty("length", 0, true, false, false);
+                ScriptEngine.Array.PrototypeObject.Push(originalArgs, args);
```
ayende, line 305: Why do you need this? Why not just use the `args` ?

src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -293,39 +292,65 @@ private void FillDoubleArrayFromJsArray(double[] array, ArrayInstance jsArray, s
...
+            }
+
+            private JsValue AppendTimeSeries(JsValue document, JsValue name, JsValue[] args)
             {
-                AssertValidDatabaseContext("appendTimeSeries");
-                const string signature4Args = "appendTimeSeries(doc, timeseries, timestamp, values)";
-                const string signature5Args = "appendTimeSeries(doc, timeseries, timestamp, tag, values)";
+                AssertValidDatabaseContext("timeseries(doc, name).append");
+
+                const string signature2Args = "append(timestamp, values)";
```
ayende:
```suggestion
                const string signature2Args = "timeseries(doc, name).append(timestamp, values)";
```

No?

src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -293,39 +292,65 @@ private void FillDoubleArrayFromJsArray(double[] array, ArrayInstance jsArray, s
...
+                const string signature3Args = "append(timestamp, values, tag)";
```
ayende: Same here

src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -293,39 +292,65 @@ private void FillDoubleArrayFromJsArray(double[] array, ArrayInstance jsArray, s
                     ++i;
                 }
             }
-            
-            private JsValue AppendTimeSeries(JsValue self, JsValue[] args)
+
+            private JsValue TimeSeries(JsValue self, JsValue[] args)
```
ayende, line 296: What would this API look like for counters?

src/Raven.Server/Documents/Patch/ScriptRunner.cs
ayende: You might be called on another document, though. Maybe loaded.

src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -369,22 +394,22 @@ private JsValue AppendTimeSeries(JsValue self, JsValue[] args)
...
-            private JsValue DeleteRangeTimeSeries(JsValue self, JsValue[] args)
+            private JsValue DeleteRangeTimeSeries(JsValue document, JsValue name, JsValue[] args)
             {
-                AssertValidDatabaseContext("deleteTimeSeries");
-                
-                const string signature = "deleteTimeSeries(doc, timeseries, from, to)";
-                const int requiredArgs = 4;
+                AssertValidDatabaseContext("timeseries(doc, name).remove");
+
+                const string signature = "remove(from, to)";
```
ayende:
```suggestion
                const string signature = "timesieres(doc, name).remove(from, to)";
```

ayende (commented):
Very minor notes, to consider / discuss.
What do you think?

src/Raven.Server/Documents/Patch/ScriptRunner.cs
ppekrol: This is why I thought to have it as an addition to the current one

src/Raven.Server/Documents/Patch/ScriptRunner.cs
aviv86: I also didn't think that this was needed, and I originaly just used the `args` here

but then I noticed that I had a single test that was always failing :
`CanAppendTimeSeriesByPatch_WithoutTag`
with InlineData : `@"timeseries(id(this), args.timeseries).append(new Date(args.timestamp), args.values);"`
(the 2nd inline data for this test, with explicit tag = null, worked fine) :
https://github.com/ravendb/ravendb/blob/v5.0/test/SlowTests/Client/TimeSeries/Patch/TimeSeriesPatchTests.cs#L174

and in this failing test, when I reached line 308 here: https://github.com/ravendb/ravendb/pull/10387/files#diff-d69adfc2e1af26dc65c54a25d0723462R308
`AppendTimeSeries(args[0], args[1], values));`
args[0] was the timestamp and args[1] was the values (i.e at this point `args` were the arguments for `AppendTimeSeries`, not the arguments for `TimeSeries`)
I tried to investigate that, but couldn't figure out why this happned, and why only in this specific case..

cloning the `args` into `originalArgs` solved this problem

src/Raven.Server/Documents/Patch/ScriptRunner.cs
aviv86: something like this ? :
`counters(doc).increment(name)`
`counters(doc).increment(name, value)`
`counters(doc).delete(name)`
`counters(doc).get(name)`

and maybe also :
`counters(doc).getAll()`
`counters(doc).getRaw(name)`

ayende: Looking at this, I don't know if that would be worth our time

ayende: @ppekrol ?

ppekrol: It is nice, but we still probably would want to keep both of the syntaxes for some time. Because of that, I think we can do it later on in 5.x cycle and deprecate old syntax in 6 or 7 even. Thoughts?

ayende: Make sense, we can delay that, then

ayende (approved):

src/Raven.Server/Documents/Patch/ScriptRunner.cs
ayende: Okay, the issue is that Jint is reusing instances. Instead of allocating a new instance of the `args`, it will reuse it.
I changed things so the value is carried in the object itself.

ayende: see https://github.com/ravendb/ravendb/pull/10404

PR #10390, opened 2020-03-27
RavenDB-13881 Show number of items for array & object

Danielle9897:


src/Raven.Studio/typescript/widgets/virtualGrid/columns/textColumn.ts
```diff
@@ -125,19 +125,19 @@ class textColumn<T> implements virtualColumn {
...
             return {
-                rawText: "[ ... ]",
+                rawText: `<span class="array-item">[${innerHtml}]</span> <sup>${cellValue.length}</sup>`,
                 typeCssClass: "token array"
             }
         }
 
         if (_.isObject(cellValue)) {
-            const value = generalUtils.escapeHtml(JSON.stringify(cellValue, null, 2));
+            const innerHtml = Object.keys(cellValue).length ? "&hellip;" : "&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;";
```
ml054: for performance reasons please compute OBject.keys(cellValue).length 1 time.

Danielle9897: done

src/Raven.Server/Web/Studio/StudioCollectionsHandler.cs
```diff
@@ -161,10 +160,10 @@ private void WriteDocument(BlittableJsonTextWriter writer, JsonOperationContext
                                 writer.WriteValue(prop.Token & BlittableJsonReaderBase.TypesMask, prop.Value);
                                 break;
                             case ValueWriteStrategy.SubstituteWithArrayStub:
-                                arraysStubs.Add(prop.Name);
+                                arraysStubs.Add(prop.Name, (prop.Value as BlittableJsonReaderArray).Length);
```
ppekrol: `prop.Value as BlittableJsonReaderArray` suggests that we are expecting that this might be `null`? If yes, then we can get NRE here, but I think this cannot be `null` so I would suggest using non-safe cast.

src/Raven.Server/Web/Studio/StudioCollectionsHandler.cs
```diff
@@ -161,10 +160,10 @@ private void WriteDocument(BlittableJsonTextWriter writer, JsonOperationContext
...
                                 break;
                             case ValueWriteStrategy.SubstituteWithObjectStub:
-                                objectsStubs.Add(prop.Name);
+                                objectsStubs.Add(prop.Name, (prop.Value as BlittableJsonReaderObject).Count);
```
ppekrol: `prop.Value as BlittableJsonReaderObject` suggests that we are expecting that this might be null? If yes, then we can get NRE here, but I think this cannot be null so I would suggest using non-safe cast.

src/Raven.Server/Web/Studio/StudioCollectionsHandler.cs
ppekrol: Previosuly we had a `HashSet` here. It suggest that we were expecting duplicates? If yes, then `.Add` will throw. You should use `[prop.Name] = value` instead.

src/Raven.Server/Web/Studio/StudioCollectionsHandler.cs
```diff
@@ -178,10 +177,18 @@ private void WriteDocument(BlittableJsonTextWriter writer, JsonOperationContext
             if (first == false)
                 writer.WriteComma();
 
+            var arrayStubsJson = new DynamicJsonValue();
+            foreach (var p in arraysStubs)
```
ppekrol: If we are not expecting duplicates, then we can build DJV right-away. Instead of using Dictionaries, you can add values to DJV.

src/Raven.Studio/typescript/widgets/virtualGrid/columns/textColumn.ts
```diff
@@ -125,19 +125,20 @@ class textColumn<T> implements virtualColumn {
...
             return {
-                rawText: "[ ... ]",
+                rawText: `<span class="array-item">[${innerHtml}]</span> <sup>${cellValue.length}</sup>`,
                 typeCssClass: "token array"
             }
         }
 
         if (_.isObject(cellValue)) {
-            const value = generalUtils.escapeHtml(JSON.stringify(cellValue, null, 2));
+            const porpertiesCount = Object.keys(cellValue).length;
```
ml054: typo

src/Raven.Server/Web/Studio/StudioCollectionsHandler.cs
Danielle9897: ok

src/Raven.Server/Web/Studio/StudioCollectionsHandler.cs
Danielle9897: ok

src/Raven.Server/Web/Studio/StudioCollectionsHandler.cs
Danielle9897: ok

src/Raven.Studio/typescript/widgets/virtualGrid/columns/textColumn.ts
Danielle9897: done

ml054 (approved):

ppekrol (approved):

PR #10399, opened 2020-03-30
RavenDB-14797: Include additional sources with each map when it is sent to CollectReferencedCollections

garayx:


src/Raven.Server/Documents/Indexes/Static/JavaScriptIndex.cs
```diff
@@ -213,11 +214,15 @@ private HashSet<CollectionName> ExecuteCodeAndCollectReferencedCollections(strin
...
 
+            var additionalSourcesString = new StringBuilder();
             if (definition.AdditionalSources != null)
             {
                 foreach (var script in definition.AdditionalSources.Values)
                 {
                     _engine.ExecuteWithReset(script);
+
+                    additionalSourcesString.Append(Environment.NewLine);
+                    additionalSourcesString.AppendLine(script);
```
ayende: Why do this? We are already creating this in the context.
Adding it for each map means that we have a LOT more scripts to go through

garayx: In `ExecuteCodeAndCollectReferencedCollections`(Line 200) we create a new program and using the visitor to execute it on each map to extract the referenced collection, if the map uses a script from additional scripts then the visitor will fail to extract the Referenced Collections, this is why I pass the map+additional scripts.

Taking a 2nd look I can remove the `_engine.ExecuteWithReset(program);` from `ExecuteCodeAndCollectReferencedCollections`
and add `_engine.ExecuteWithReset(map);`  before I concat the `additionalSourcesString` to `map` (line 238)

ayende: You can also pass the additional sources to the visitor, no?

garayx: done

ayende (approved):

ppekrol (approved):

PR #10402, opened 2020-03-30
RavenDB-14803 Register socket use same log thread if it is already run & complete setup log run on same thread

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-14803

### DIFF
--- a/src/Sparrow/Logging/LoggingSource.cs
+++ b/src/Sparrow/Logging/LoggingSource.cs
@@ -108,7 +108,14 @@ namespace Sparrow.Logging
             {
                 if (_listeners.TryAdd(source, context) == false)
                     throw new InvalidOperationException("Socket was already added?");
-                SetupLogMode(LogMode, _path, RetentionTime, RetentionSize, Compressing);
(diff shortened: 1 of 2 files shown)


src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -108,7 +108,14 @@ public async Task Register(WebSocket source, WebSocketContext context, Cancellat
             {
                 if (_listeners.TryAdd(source, context) == false)
                     throw new InvalidOperationException("Socket was already added?");
-                SetupLogMode(LogMode, _path, RetentionTime, RetentionSize, Compressing);
```
haludi, line 111: The cause of the test failure is that the `RemoveSocket` creates a new log thread async and it can happen after the test is finished and after `EndLogging` is called. The solution can be just to be sure in the test that `EndLogging` is called after the new log thread is set but using the same thread seem to be better implementation (and it fixes the test too).
Is there a benefit of creating a new thread?

ayende (approved):

PR #10403, opened 2020-03-30
RavenDB-14585 Do not allow to modify Migrator Path if defined in sett…

Danielle9897:
…ings.json

### DIFF
--- a/src/Raven.Studio/typescript/commands/database/tasks/getMigratorPathConfigurationCommand.ts
+++ b/src/Raven.Studio/typescript/commands/database/tasks/getMigratorPathConfigurationCommand.ts
@@ -3,10 +3,10 @@ import endpoints = require("endpoints");
 
 class getMigratorPathConfigurationCommand extends commandBase {
 
-    execute(): JQueryPromise<Configuration> {
+    execute(): JQueryPromise<MigratorPathConfiguration> {
         const url = endpoints.global.studioTasks.studioTasksAdminMigratorPath;
        
-        return this.query<Configuration>(url, null)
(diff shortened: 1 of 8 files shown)


src/Raven.Studio/typescript/commands/database/tasks/getMigratorPathConfigurationCommand.ts
```diff
@@ -0,0 +1,15 @@
+import commandBase = require("commands/commandBase");
+import endpoints = require("endpoints");
+
+class getMigratorPathConfigurationCommand extends commandBase {
+
+    execute(): JQueryPromise<Configuration> {
```
ml054: from where this Confgiuration type comes from?

src/Raven.Studio/typings/_studio/dto.d.ts
```diff
@@ -643,3 +643,7 @@ interface rawStackTraceResponseItem {
 }
 
 type indexStatus = "Normal" | "ErrorOrFaulty" | "Stale" | "Paused" | "Disabled" | "Idle";
+
+interface Configuration {
```
ml054: Can we rename this something less generic, like MigratorPathConfiguration

Danielle9897: done (my original thought was that maybe 'configuration' interface will represent other settings in the future)

ml054: but this is migrator specific

ml054 (approved):

PR #10406, opened 2020-03-31
RavenDB-14006 support for compare exchange value includes via JS projections

ppekrol:


src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -572,6 +581,39 @@ private JsValue IncludeDoc(JsValue self, JsValue[] args)
...
+
+                if (args[0].IsNull() || args[0].IsUndefined())
+                    return self;
+
+                if (args[0].IsArray())// recursive call ourselves
+                {
+                    var array = args[0].AsArray();
+                    foreach (var pair in array.GetOwnProperties())
+                    {
+                        args[0] = pair.Value.Value;
```
ayende, line 595: Don't we need to filter `length` here?

ayende (approved):

src/Raven.Server/Documents/Patch/ScriptRunner.cs
ppekrol: Yes! I've introduced an extension method to handle this

PR #10408, opened 2020-03-31
RavenDB-14290: fix Delete on entity doesn't work during OnBeforeDelet…

garayx:
…e event

### DIFF
--- a/src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
+++ b/src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
@@ -540,6 +540,7 @@ more responsive application.
 
         /// <summary>
         /// Marks the specified entity for deletion. The entity will be deleted when SaveChanges is called.
+        /// <para>WARNING: This method will call <see cref="IDocumentSession.Delete"/> if called during OnBeforeStore listener!</para>
         /// </summary>
         /// <typeparam name="T"></typeparam>
         /// <param name="entity">The entity.</param>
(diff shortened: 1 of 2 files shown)


src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -540,6 +540,7 @@ public static object GetDefaultValue(Type type)
 
         /// <summary>
         /// Marks the specified entity for deletion. The entity will be deleted when SaveChanges is called.
+        /// <para>WARNING: This method will call <see cref="IDocumentSession.Delete"/> if called during OnBeforeStore listener!</para>
```
ayende, line 543: What? I don't understand this comment

ayende: Calling delete will call delete?

garayx: If `Delete` (by entity) called during `OnBeforeStore` then `Delete` (by id) will be used instead
I added this warning because `Delete` (by id) don't call `OnBeforeDelete` listener

ayende: Let's remove the comment and make the `Delete(id)` raise the `OnBeforeDelete` as well

ppekrol: IIRC this is a deferred command and we do not raise events for those and OnBeforeDelete requires an entity

ayende: We can pass null for the entity, no?
And there is a different between `Delete(id)` and `Advanced.Defer()`

ppekrol: The Delete(entity) should trigger the event for sure, the Delete(id) should not, I will check tomorrow with @garayx and we will see what can be done about that

PR #10415, opened 2020-04-01
RavenDB-14830 update ubuntu base image

gregolsky:


gregolsky:
Removed few packages since they are installed in the base:
https://github.com/dotnet/dotnet-docker/blob/d68460ceab4fe635d95a28bfca410db1a1d57caa/3.1/runtime-deps/bionic/amd64/Dockerfile#L7

PR #10416, opened 2020-04-01
RavenDB-14651 : Rework time-series Client API - change structure of TimeSeriesOperation

aviv86:

### DIFF
--- a/src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesOperation.cs
+++ b/src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesOperation.cs
@@ -33,7 +33,7 @@ namespace Raven.Client.Documents.Operations.TimeSeries
 
             if (input.TryGet(nameof(Appends), out BlittableJsonReaderArray operations) && operations != null)
             {
-                result.Appends = new List<AppendOperation>();
+                var sorted = new SortedList<long, AppendOperation>();
 
                 foreach (var op in operations)
                 {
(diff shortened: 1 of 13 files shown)


src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesOperation.cs
```diff
@@ -72,40 +73,40 @@ private static void ThrowNotBlittableJsonReaderObjectOperation(object op)
 
         private static void ThrowMissingDocumentId()
         {
-            throw new InvalidDataException($"Missing '{nameof(DocumentId)}' property on 'TimeSeries'");
+            throw new InvalidDataException($"Missing '{nameof(DocumentId)}' property on 'TimeSeriesOperation'");
+        }
+        private static void ThrowMissingName()
```
ppekrol: introduce parameters here and merge those two exceptions?
`ThrowMissingProperty(nameof(DocumentId));`

ppekrol (approved):

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesOperation.cs
aviv86: done

src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
```diff
@@ -411,118 +411,80 @@ public async Task ConfigTimeSeries()
...
+                    changes++;
                 }
-
-                _appendDictionary = new Dictionary<string, SortedList<long, TimeSeriesOperation.AppendOperation>>();
-
-                foreach (var item in operation.Appends)
+                else
                 {
-                    if (_appendDictionary.TryGetValue(item.Name, out var sorted) == false)
+                    var sorted = new SortedList<long, TimeSeriesOperation.AppendOperation>();
```
karmeli87: When I think about it, in most cases it is redundant, would it be better to assume that we get a sorted data from the client.

Otherwise we will throw during the enumeration?

karmeli87: @ayende @ppekrol

aviv86: there is also the problem of having repeating timestamps in `_operation.Appends` 
(like in `CanCreateSimpleTimeSeries2` test) - which this code solves

PR #10420, opened 2020-04-02
RavenDB-14832 - Uncaught exception killed the process

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-14832


src/Raven.Server/Utils/PoolOfThreads.cs
```diff
@@ -232,6 +232,20 @@ private bool DoWork()
                     LongRunningWork.Current = _workIsDone;
                     _action(_state);
                 }
+                catch (OutOfMemoryException e)
```
ppekrol, line 235: We cannot do this like that. This will ignore OOM exception and e.g. kill the index without killing the server. User will not know that half of his indexes are not working. This needs to be handled internally in all of the actions.

PR #10433, opened 2020-04-07
RavenDB-13717 Index name overflow in Query Stats + in Delete index

Danielle9897:


ml054 (approved):
I really like this solution! It is generic. 👍

PR #10435, opened 2020-04-07
RavenDB-14851 remove time-series replication optimizations

karmeli87:
The work on the rollup/retention revealed some replication issues.
This PR hopefully will bring the replication to a correct and stable state for all cases.

### DIFF
--- a/src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
+++ b/src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
@@ -94,6 +94,37 @@ namespace Raven.Server.Documents.TimeSeries
             }
         }
 
+        public void MarkSegmentForPolicy(DocumentsOperationContext context, TimeSeriesSliceHolder slicerHolder, DateTime timestamp, 
+            string changeVector,
+            int numberOfEntries)
+        {
+            if (Configuration.Collections.TryGetValue(slicerHolder.Collection, out var config) == false)
+                return;
+
+            var currentIndex = config.GetPolicyIndexByTimeSeries(slicerHolder.Name);
+            if (currentIndex == -1) // policy not found
+                return;
+            
+            var nextPolicy = config.GetNextPolicy(currentIndex);
+            if (nextPolicy == null)
+                return;
+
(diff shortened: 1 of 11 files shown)


src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -176,6 +208,23 @@ internal async Task HandleChanges()
 
                         await AddNewPolicy(collectionName, prev, policy.Policy);
                     }
+
+                    //TODO: clean up the removed policies from the rollups table
```
ayende: Aren't they going to be cleaned up automatically when we read them, see they are not relevant and removed it?

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
```diff
@@ -374,96 +375,131 @@ bool TryRemoveRange(ref TableValueReader reader, out Table.TableValueHolder next
...
 
-                void RemoveTimeSeriesNameFromMetadata(DocumentsOperationContext ctx, string docId, string tsName)
+                void Notify()
                 {
-                    var doc = _documentDatabase.DocumentsStorage.Get(ctx, docId);
-                    if (doc == null)
-                        return;
+                    context.Transaction.AddAfterCommitNotification(new TimeSeriesChange
+                    {
+                        ChangeVector = changeVector, // change-vector can be null, if we delete an entire segments
```
ayende: Why? Shouldn't we still generate a change vector for the deletion?

src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
karmeli87: right, done.

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
karmeli87: Since the optimization was removed, so it that comment.

ppekrol (approved):

ppekrol:
@karmeli87 can you check failing test here: https://jenkins.hibernatingrhinos.com/job/PR_Tests_v5.0/730/

PR #10440, opened 2020-04-08
RavenDB-14498 added OnTopologyUpdated event to DocumentStore

ppekrol:


src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -238,13 +238,41 @@ public void Translate(object sender, FailedRequestEventArgs args)
...
 
+        private event EventHandler<TopologyUpdatedEventArgs> _onTopologyUpdated;
+
+        public event EventHandler<TopologyUpdatedEventArgs> OnTopologyUpdated
+        {
+            add
+            {
+                lock (_locker)
+                {
+                    _onTopologyUpdated += value;
```
ayende, line 254: This should be thread safe, no need for locking

ppekrol: Done: https://github.com/ravendb/ravendb/commit/b25e4814b1e65b27e456fb44c2f26acaf94bd263

PR #10443, opened 2020-04-10
RavenDB-14857 switched to newline insensitive assert

ppekrol:


ppekrol:
It was Jenkins issue after all.

PR #10446, opened 2020-04-10
RavenDB-14507 batching output reduce to collection

ppekrol:


src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceIndexWriteOperation.cs
```diff
@@ -23,12 +25,14 @@ public class OutputReduceIndexWriteOperation : IndexWriteOperation
...
-            _outputReduceToCollectionCommand = index.OutputReduceToCollection.CreateCommand(indexContext, _txHolder);
+            _outputReduceToCollectionCommandBatcher = index.OutputReduceToCollection.CreateCommandBatcher(indexContext, _txHolder);
         }
 
         public override void Commit(IndexingStatsScope stats)
         {
-            var enqueue = DocumentDatabase.TxMerger.Enqueue(_outputReduceToCollectionCommand);
+            var enqueued = new List<Task>();
+            foreach (var command in _outputReduceToCollectionCommandBatcher.CreateCommands())
+                enqueued.Add(DocumentDatabase.TxMerger.Enqueue(command));
```
arekpalinski: 1. This makes that deletions (`_reduceKeyHashesToDelete`) and puts (`_reduceDocuments`) returned by  `CreateCommands()` can be run in mixed order. In previous implementation we processed them sequentially.

I think this can cause issues in current impl - we need to call `_reduceKeyHashesToDelete` when calling `AddReduce`:

https://github.com/ravendb/ravendb/pull/10446/files#diff-b59632789139fb86226314321cc91743R56-R70

2. I'm not sure about implications of adding _everything_ immediately to TX merger. This is significant change. I thought we want to run one batch at a time to let other transactions run meanwhile. I wonder if we won't be better off by limiting the number of batches which run concurrently.

ayende: 1) Do we have to run them this way? I would rather have them run in the same tx. The issue is that is is fine is we are going from one valid state to another, but mixing deletes and inserts in such a way may mean that we move from valid state to an invalid state (just deletes, without the corresponding inserts).

2) Adding a lot of work at once to the tx merger can cause stall for other things, yes.
I would rather add a few, then wait for them to complete, then move on, etc.
We prioritize users' action over background work in general

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
```diff
@@ -10,19 +10,205 @@
 using Raven.Server.Utils;
 using Sparrow.Json;
 using Sparrow.Json.Parsing;
+using Sparrow.Platform;
 using Voron.Impl;
 
 namespace Raven.Server.Documents.Indexes.MapReduce.OutputToCollection
 {
-    public class OutputReduceToCollectionCommand : TransactionOperationsMerger.MergedTransactionCommand, IDisposable
+    public class OutputReduceToCollectionCommandBatcher : IDisposable
```
ayende, line 18: The usual way we handle this is by having a single command that we enqueue multiple times.
Each time it runs, we process a batch, and the calling code will check if it is done and if not, resend to the merger.

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceIndexWriteOperation.cs
ppekrol: Doing this sequentially now.

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
ppekrol: I'm not sure where this is handled like that? We are creating a lot of sub-commands all over the codebase.

ayende: `ExecuteRateLimitedOperations` work in this manner.

ayende (approved):

PR #10450, opened 2020-04-13
RavenDB-14831 use custom time type for retention and aggregation policy

karmeli87:


src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
```diff
@@ -22,20 +22,31 @@ public class TimeSeriesCollectionConfiguration : IDynamicJson
...
 
-        internal void Initialize()
+        internal void ValidateAndInitialize()
         {
             if (Policies.Count == 0)
                 return;
 
             Policies.Sort(TimeSeriesDownSamplePolicyComparer.Instance);
 
+            RawPolicy.RetentionTime.AssertMonthOrSeconds();
```
ml054: I wonder if we should convey implementation detail on what we assert ;-) it is impl details for me

src/Raven.Server/Documents/TimeSeries/TimeSeriesRollups.cs
```diff
@@ -766,9 +768,25 @@ void AggregateIndividualItems(IEnumerable<TimeSeriesStorage.Reader.SingleResult>
...
+                return nextRollup;
+            }
+
+            // align by months
+            var totalMonths = time.Year * 12 + time.Month - 1;
+            var integerAggPart = totalMonths / nextPolicy.AggregationTime.Months;
+            var nextInMonths = nextPolicy.AggregationTime.Months * (integerAggPart + 1);
+            var years = nextInMonths / 12;
+            var months = nextInMonths % 12;
+            return new DateTime(years, months + 1, 1).Ticks;
```
ml054: I wonder if we should also take into account the lap years?

ml054: or actually we do, as years contains values around 2k, not 1-2 right? We add to current date?

ml054 (approved):

src/Raven.Server/Documents/TimeSeries/TimeSeriesRollups.cs
karmeli87: Months and days are isolated in this solution, you can't convert from days to months and via versa.

So we don't care about lap years here, any year in the gregorian calendar has 12 months.
Unlike the hebrew calendar that sometimes has 13 months :-)

ml054: oh that make sense - thanks ;-)

ppekrol:
@karmeli87 ping me about this

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
ayende: No, this is fine

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
```diff
@@ -22,20 +22,31 @@ public class TimeSeriesCollectionConfiguration : IDynamicJson
         /// </summary>
         public RawTimeSeriesPolicy RawPolicy { get; set; } = RawTimeSeriesPolicy.Default;
 
-        internal void Initialize()
+        internal void ValidateAndInitialize()
         {
             if (Policies.Count == 0)
                 return;
 
             Policies.Sort(TimeSeriesDownSamplePolicyComparer.Instance);
```
ayende: This is now wrong, no?
`TimeSeriesDownSamplePolicyComparer` isn't taking into account the `Months`

src/Sparrow/TimeValue.cs
```diff
@@ -0,0 +1,321 @@
...
+            builder.Append(' ');
+            builder.Append(singular);
+
+            if (value == 1)
+            {
+                builder.Append(' ');
+                return;
+            }
+
+            builder.Append("s "); // lucky me, no special rules here
```
ayende, line 95: LOL

src/Sparrow/TimeValue.cs
```diff
@@ -0,0 +1,321 @@
...
+            var myBounds = GetBounds(this);
+            var otherBounds = GetBounds(other);
+
+            if (otherBounds.UpperBound < myBounds.LowerBound)
+                return 1;
+
+            if (otherBounds.LowerBound > myBounds.UpperBound)
+                return -1;
+            
+            throw new InvalidOperationException("We can't compare ");
```
ayende: Need to explain *why* we can't compare

src/Sparrow/TimeValue.cs
```diff
@@ -0,0 +1,321 @@
...
+            if (IsMin(ref other))
+            {
+                result = IsMin(ref current) ? 0 : 1;
+                return true;
+            }
+
+            return false;
+        }
+
+        private static bool IsMax(ref TimeValue time)
```
ayende: Why is this a ref?

src/Sparrow/TimeValue.cs
```diff
@@ -0,0 +1,321 @@
...
+        }
+
+        private static (int UpperBound, int LowerBound) GetBounds(TimeValue time)
+        {
+            var myUpperBound = time.Months * SecondsIn31Days + time.Seconds;
+            var myLowerBound = time.Months * SecondsIn28Days + time.Seconds;
+            return (myUpperBound, myLowerBound);
+        }
+
+        private static bool IsSpecialCompare(ref TimeValue current, ref TimeValue other, out int result)
```
ayende: Why are you passing by ref? `current` and `other` aren't changed

ayende (changes requested):

src/Sparrow/TimeValue.cs
karmeli87: It is a `struct`, but I guess he isn't big enough to benefit from passing it like this.

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesCollectionConfiguration.cs
karmeli87: We use the upper and lower bound of number of seconds in a month, so we compare it if we can, otherwise an exception is thrown.

src/Sparrow/TimeValue.cs
ayende: Yes, it is same size as the pointer, no need to do this

ppekrol (approved):

PR #10456, opened 2020-04-15
HRINT-1576

ppekrol:
- added limits to RequestExecutor context pool
- switch from Queue to Stack to constantly reuse same contexts instead of reusing all of them, this should allow to clear them more easily

### DIFF
--- a/src/Raven.Client/Http/RequestExecutor.cs
+++ b/src/Raven.Client/Http/RequestExecutor.cs
@@ -33,7 +33,6 @@ using Sparrow.Json;
 using Sparrow.Logging;
 using Sparrow.Platform;
 using Sparrow.Threading;
-using Size = Sparrow.Size;
 
 namespace Raven.Client.Http
 {
@@ -353,16 +352,14 @@ namespace Raven.Client.Http
 
             _lastReturnedResponse = DateTime.UtcNow;
 
-            var maxContextSizeToKeep = PlatformDetails.Is32Bits == false
-                ? new Size(1, SizeUnit.Megabytes)
-                : new Size(256, SizeUnit.Kilobytes);
+            Conventions = conventions.Clone();
 
             var maxNumberOfContextsToKeepInGlobalStack = PlatformDetails.Is32Bits == false
                 ? 1024
                 : 256;
 
-            ContextPool = new JsonContextPool(maxContextSizeToKeep, maxNumberOfContextsToKeepInGlobalStack);
-            Conventions = conventions.Clone();
+            ContextPool = new JsonContextPool(Conventions.MaxContextSizeToKeep, maxNumberOfContextsToKeepInGlobalStack);
+
             DefaultTimeout = Conventions.RequestTimeout;
             SecondBroadcastAttemptTimeout = conventions.SecondBroadcastAttemptTimeout;
             FirstBroadcastAttemptTimeout = conventions.FirstBroadcastAttemptTimeout;
--- a/src/Raven.Client/Http/RequestExecutor.cs
+++ b/src/Raven.Client/Http/RequestExecutor.cs
@@ -31,6 +31,7 @@ using Sparrow;
 using Sparrow.Collections;
 using Sparrow.Json;
 using Sparrow.Logging;
+using Sparrow.Platform;
 using Sparrow.Threading;
 using Size = Sparrow.Size;
 
@@ -352,7 +353,15 @@ namespace Raven.Client.Http
 
             _lastReturnedResponse = DateTime.UtcNow;
 
-            ContextPool = new JsonContextPool(new Size(8, SizeUnit.Megabytes));
+            var maxContextSizeToKeep = PlatformDetails.Is32Bits == false
+                ? new Size(1, SizeUnit.Megabytes)
+                : new Size(256, SizeUnit.Kilobytes);
+
+            var maxNumberOfContextsToKeepInGlobalStack = PlatformDetails.Is32Bits == false
+                ? 1024
+                : 256;
+
+            ContextPool = new JsonContextPool(maxContextSizeToKeep, maxNumberOfContextsToKeepInGlobalStack);
             Conventions = conventions.Clone();
             DefaultTimeout = Conventions.RequestTimeout;
             SecondBroadcastAttemptTimeout = conventions.SecondBroadcastAttemptTimeout;
--- a/src/Sparrow/Json/JsonContextPoolBase.cs
+++ b/src/Sparrow/Json/JsonContextPoolBase.cs
@@ -1,24 +1,30 @@
 ﻿using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Threading;
 using Sparrow.LowMemory;
+using Sparrow.Platform;
 using Sparrow.Threading;
 using Sparrow.Utils;
 
 namespace Sparrow.Json
 {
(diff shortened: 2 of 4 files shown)


src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -352,7 +353,15 @@ protected RequestExecutor(string databaseName, X509Certificate2 certificate, Doc
 
             _lastReturnedResponse = DateTime.UtcNow;
 
-            ContextPool = new JsonContextPool(new Size(8, SizeUnit.Megabytes));
+            var maxContextSizeToKeep = PlatformDetails.Is32Bits == false
```
grisha-kotler: Isn't it too small for 64bit?
Can we check what is the average size that we have?

ayende: Agreed, and I think we want to make this configurable for the user.

src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -152,57 +182,95 @@ public void Dispose()
...
-                if (currentTime - context.InPoolSince < idleTime)
+                if (globalStackRebuildNeeded)
                 {
-                    break;
+                    _lastGlobalStackRebuild = currentTime;
+
+                    var previousGlobalStack = _previousGlobalStack;
+
+                    _currentGlobalStack = new CountingConcurrentStack<T>();
+                    _previousGlobalStack = currentGlobalStack;
```
ayende: Isn't this a problem when we are running concurrently?
Other threads may add to their view of `_currentGlobalStack` after you called `ClearStack`

ayende: And why not simply clearing the *current* stack and be done with it?

src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -1,24 +1,30 @@
...
         private bool _disposed;
 
         protected SharedMultipleUseFlag LowMemoryFlag = new SharedMultipleUseFlag();
         private readonly CancellationTokenSource _cts = new CancellationTokenSource();
         private readonly long _maxContextSizeToKeepInBytes;
+        private readonly long _maxNumberOfContextsToKeepInGlobalStack;
 
         private readonly T[][] _perCoreContexts;
-        private readonly ConcurrentQueue<T> _globalQueue = new ConcurrentQueue<T>();
+        private CountingConcurrentStack<T> _previousGlobalStack;
```
ayende: Why the change from queue to stack?

src/Raven.Client/Http/RequestExecutor.cs
ppekrol: My thinking is:
a) this is Client API, so why it should retain big context at all?
b) this is Client API, so why it should retain high amount of contexts?

I've checked our api.ravendb.net, most of the context are below 1MB, with few that are up to 4MB of size, with total 4GB of memory kept in context pool.

src/Sparrow/Json/JsonContextPoolBase.cs
ppekrol: If we have a queue with 4000 items in it already and we are requesting let's say 20 contexts per second (synchronously even - no concurrency). It means that in 1min we will take 1200 different contexts from the pool and refresh their 'last used date', so the cleaner will not trigger, but by using stack, we will use only 1 context from the stack, so others can be safely removed by the cleaner.

src/Sparrow/Json/JsonContextPoolBase.cs
ppekrol: The idea here is that I still want to retain (maybe unnecessary?) contexts from the global stack that are not outdated. I'm doing that by creating new stack, but when allocating I'm taking values first from the 'old stack', then releasing them to the 'current stack'. I'm doing that, because I'm not removing anything from the stack during the cleanup (regular one - not the one that happens every 15min), only disposing contexts there.

The other option would be just to release everything as you wrote. It is easier, but not sure if we want that?

src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -1,25 +1,32 @@
...
         private readonly CancellationTokenSource _cts = new CancellationTokenSource();
         private readonly long _maxContextSizeToKeepInBytes;
+        private readonly long _maxNumberOfContextsToKeepInGlobalStack;
+        private long _numberOfContextsDisposedInGlobalStack;
 
         private readonly T[][] _perCoreContexts;
-        private readonly ConcurrentQueue<T> _globalQueue = new ConcurrentQueue<T>();
-        private readonly Timer _idleTimer;
+        private CountingConcurrentStack<T> _previousGlobalStack;
+        private CountingConcurrentStack<T> _currentGlobalStack = new CountingConcurrentStack<T>();
```
ayende: Why switch from queue to stack?

src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -225,26 +298,34 @@ private void Push(T context)
...
+
             // couldn't find a place for it, let's add it to the global list
-            _globalQueue.Enqueue(context);
+            if (currentGlobalStack.Count >= _maxNumberOfContextsToKeepInGlobalStack)
+            {
+                context.Dispose();
+                return;
+            }
+
+            currentGlobalStack.Push(context);
```
ayende, line 305: Here you may be pushing to the previous instance, no?

src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -152,57 +183,99 @@ public void Dispose()
...
-                if (currentTime - context.InPoolSince < idleTime)
+                if (globalStackRebuildNeeded && _numberOfContextsDisposedInGlobalStack > 0)
                 {
-                    break;
+                    _lastGlobalStackRebuild = currentTime;
+
+                    var previousGlobalStack = _previousGlobalStack;
+
+                    _numberOfContextsDisposedInGlobalStack = 0;
+                    _currentGlobalStack = new CountingConcurrentStack<T>();
```
ayende: Why are you doing this switch?

src/Raven.Client/Http/RequestExecutor.cs
ayende: This is also used internally for things like ETL, where we might have larger values.
Should we customize the conventions there?

src/Sparrow/Json/JsonContextPoolBase.cs
ayende: The easiest thing to do would be to keep the queue structure, and scan the whole queue.
If there are old stuff there, just mark it in use and dispose it.
The null pull from the queue will clear it up, no?

src/Sparrow/Json/JsonContextPoolBase.cs
ayende: IIRC, there was a reasoning behind this.
The idea is that we aren't going to end up with so many items in the global pool, because that is quite strange to have.
And when you pull from the global pool it goes to the per core one, no?
That would avoid this issue.

The auto cleanup of old instance really helps here.

src/Raven.Client/Http/RequestExecutor.cs
ppekrol: Increased to 8/2 MB (64/32 bit respectively) for server

PR #10457, opened 2020-04-16
RavenDB-14875 Tests for migration and smuggller 4.2 <-> 5

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-14875

### DIFF
--- a/test/InterversionTests/SmugglerTests.cs
+++ b/test/InterversionTests/SmugglerTests.cs
@@ -1,18 +1,20 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.IO;
+using System.Diagnostics;
+using System.Linq;
+using System.Net.Http;
+using System.Text;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


test/InterversionTests/SmugglerTests.cs
```diff
@@ -21,366 +23,25 @@ namespace InterversionTests
...
-            }
         }
 
-        [Fact]
-        public async Task CanExportFrom42AndImportToCurrent()
+        [Theory]
+        [InlineData(ExcludeOn.Non)]
+        [InlineData(ExcludeOn.Export)]
+        [InlineData(ExcludeOn.Import)]
+        public async Task CanExportFrom42AndImportTo5(ExcludeOn excludeOn)
```
ppekrol: It is not '5'. It is 'Current'. E.g. what will happen if we will move to v6? Same for other names.

ppekrol (approved):

PR #10462, opened 2020-04-17
HRINT-1576

ppekrol:
- resetting LSV in allocatedStringValues on Reset to release the materialized strings so GC can collect them
- restrict the amount of allocatedStringValues in Client API to 1024

### DIFF
--- a/src/Raven.Client/Http/RequestExecutor.cs
+++ b/src/Raven.Client/Http/RequestExecutor.cs
@@ -358,7 +358,7 @@ namespace Raven.Client.Http
                 ? 1024
                 : 256;
 
-            ContextPool = new JsonContextPool(Conventions.MaxContextSizeToKeep, maxNumberOfContextsToKeepInGlobalStack);
+            ContextPool = new JsonContextPool(Conventions.MaxContextSizeToKeep, maxNumberOfContextsToKeepInGlobalStack, 1024);
 
             DefaultTimeout = Conventions.RequestTimeout;
             SecondBroadcastAttemptTimeout = conventions.SecondBroadcastAttemptTimeout;
--- a/src/Raven.Server/ServerWide/Context/TransactionContextPool.cs
+++ b/src/Raven.Server/ServerWide/Context/TransactionContextPool.cs
@@ -4,15 +4,13 @@
 //  </copyright>
 // -----------------------------------------------------------------------
 
-using System;
-using System.Threading;
 using Sparrow;
 using Sparrow.Json;
 using Voron;
 
 namespace Raven.Server.ServerWide.Context
 {
-    public class TransactionContextPool : JsonContextPoolBase<TransactionOperationContext> ,ITransactionContextPool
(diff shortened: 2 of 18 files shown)


src/Raven.Server/ServerWide/Context/TransactionContextPool.cs
```diff
@@ -24,16 +22,22 @@ public TransactionContextPool(StorageEnvironment storageEnvironment, Size? maxCo
         protected override TransactionOperationContext CreateContext()
         {
             int initialSize;
+            int longLivedSize;
+            int maxNumberOfAllocatedStringValues;
             if (_storageEnvironment.Options.RunningOn32Bits)
             {
                 initialSize = 4096;
+                longLivedSize = 4 * 1024;
```
ppekrol, line 30: Copied this from documents operation context.

src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -106,7 +107,7 @@ public unsafe LazyStringValue AllocateStringValue(string str, byte* ptr, int siz
             }
 
             var allocateStringValue = new LazyStringValue(str, ptr, size, this);
-            if (_numberOfAllocatedStringsValues < 32 * 1_024)
+            if (_numberOfAllocatedStringsValues < _maxNumberOfAllocatedStringValues)
```
ppekrol, line 110: Each context in a pool can keep `_maxNumberOfAllocatedStringValues * sizeof(LazyStringValue)` where `sizeof(LazyStringValue)` is 80 bytes => 2.6MB

src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -1016,10 +1018,16 @@ protected internal virtual unsafe void Reset(bool forceReleaseLongLivedAllocator
                 _fieldNames.Clear();
                 CachedProperties = null; // need to release this so can be collected
             }
+
+            for (var i = 0; i < _numberOfAllocatedStringsValues; i++)
+                _allocateStringValues[i].Reset();
```
ppekrol, line 1023: Resseting not only allows to return the memory, but also to clear other fields, e.g. materialized string value

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -358,7 +358,7 @@ protected RequestExecutor(string databaseName, X509Certificate2 certificate, Doc
                 ? 1024
                 : 256;
 
-            ContextPool = new JsonContextPool(Conventions.MaxContextSizeToKeep, maxNumberOfContextsToKeepInGlobalStack);
+            ContextPool = new JsonContextPool(Conventions.MaxContextSizeToKeep, maxNumberOfContextsToKeepInGlobalStack, 1024);
```
ppekrol, line 361: Setting only 1024 in Client API. It is max 80kb per context kept in the pool.

PR #10469, opened 2020-04-19
RavenDB-14880 - Excessive Client Config after changing client read behavior

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-14880

### DIFF
--- a/src/Raven.Client/Http/ClusterRequestExecutor.cs
+++ b/src/Raven.Client/Http/ClusterRequestExecutor.cs
@@ -148,7 +148,7 @@ namespace Raven.Client.Http
             return true;
         }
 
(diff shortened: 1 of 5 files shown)


ayende:
What actually caused the issue?

grisha-kotler:
We got the topology etag directly from the configuration.
When comparing it, we used a different approach:
https://github.com/ravendb/ravendb/blob/v4.2/src/Raven.Server/Documents/DocumentDatabase.cs#L1423

This happens only when there is/was a server-wide client configuration and we disabled the specific one for the database.

ayende (approved):

ppekrol (approved):

PR #10474, opened 2020-04-21
RDBS-79 Modify HttpClientHandler connection limit to not rely on the default

haludi:
https://issues.hibernatingrhinos.com/issue/RDBS-79

### DIFF
--- a/src/Raven.Client/Http/RequestExecutor.cs
+++ b/src/Raven.Client/Http/RequestExecutor.cs
@@ -38,6 +38,8 @@ namespace Raven.Client.Http
 {
     public class RequestExecutor : IDisposable
     {
+        private const int DefaultConnectionLimit = int.MaxValue;
+
         private static Guid GlobalApplicationIdentifier = Guid.NewGuid();
 
         private const int InitialTopologyEtag = -2;
@@ -956,7 +958,7 @@ namespace Raven.Client.Http
                     })
                     {
                         TimeoutInMs = 0,
-                        DebugTag = refreshTopology ? "refresh-topology-header" : refreshClientConfiguration ? "refresh-client-configuration-header" : null
+                        DebugTag = "refresh-topology-header"
                     })
                     : Task.CompletedTask;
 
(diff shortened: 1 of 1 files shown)


src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -38,6 +38,8 @@ namespace Raven.Client.Http
 {
     public class RequestExecutor : IDisposable
     {
+        private const int MinConnectionLimit = 1024 * 10;
```
ppekrol: rename to `DefaultConnectionLimit`

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1780,17 +1782,15 @@ public virtual void Dispose()
...
-            var httpMessageHandler = new HttpClientHandler();
+            var httpMessageHandler = new HttpClientHandler {MaxConnectionsPerServer = MinConnectionLimit };
             if (httpMessageHandler.SupportsAutomaticDecompression)
             {
                 httpMessageHandler.AutomaticDecompression =
                     useCompression ?
                         DecompressionMethods.GZip | DecompressionMethods.Deflate
                         : DecompressionMethods.None;
             }
-            else if (httpMessageHandler.SupportsAutomaticDecompression == false &&
```
ppekrol, line 1791: why `httpMessageHandler.SupportsAutomaticDecompression == false` was removed?

haludi: Because the `else` handle it already, no? Or I miss here something...

ppekrol: Yes, you are correct. Didn't notice. My fault.

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -38,6 +38,8 @@ namespace Raven.Client.Http
 {
     public class RequestExecutor : IDisposable
     {
+        private const int DefaultConnectionLimit = 1024 * 10;
```
ayende: Can we make this public static value, so users have access to it?

PR #10475, opened 2020-04-21
RavenDB-14866 - Bulk insert for Time Series adjustments

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-14866

### DIFF
--- a/src/Raven.Client/Documents/BulkInsert/BulkInsertOperation.cs
+++ b/src/Raven.Client/Documents/BulkInsert/BulkInsertOperation.cs
@@ -767,7 +767,7 @@ namespace Raven.Client.Documents.BulkInsert
                                 _operation.WriteComma();
 
                             firstValue = false;
-                            _operation._currentWriter.Write(Convert.ToString(value, CultureInfo.InvariantCulture));
+                            _operation._currentWriter.Write(value.ToString("R", CultureInfo.InvariantCulture));
                         }
 
                         if (tag != null)
--- a/src/Raven.Client/Documents/BulkInsert/BulkInsertOperation.cs
+++ b/src/Raven.Client/Documents/BulkInsert/BulkInsertOperation.cs
@@ -150,6 +150,7 @@ namespace Raven.Client.Documents.BulkInsert
         public CompressionLevel CompressionLevel = CompressionLevel.NoCompression;
         private readonly JsonSerializer _defaultSerializer;
         private readonly Func<object, IMetadataDictionary, StreamWriter, bool> _customEntitySerializer;
+        private readonly int _timeSeriesBatchSize;
         private long _concurrentCheck;
 
         public BulkInsertOperation(string database, IDocumentStore store, CancellationToken token = default)
@@ -219,6 +220,7 @@ namespace Raven.Client.Documents.BulkInsert
 
             _defaultSerializer = _requestExecutor.Conventions.CreateSerializer();
             _customEntitySerializer = _requestExecutor.Conventions.BulkInsert.TrySerializeEntityToJsonStream;
+            _timeSeriesBatchSize = _conventions.BulkInsert.TimeSeriesBatchSize;
 
             _generateEntityIdOnTheClient = new GenerateEntityIdOnTheClient(_requestExecutor.Conventions,
                 entity => AsyncHelpers.RunSync(() => _requestExecutor.Conventions.GenerateDocumentIdAsync(database, entity)));
@@ -689,7 +691,6 @@ namespace Raven.Client.Documents.BulkInsert
             private readonly string _id;
             private readonly string _name;
             private bool _first = true;
-            private const int _maxTimeSeriesInBatch = 1024;
             private int _timeSeriesInBatch = 0;
 
             public TimeSeriesBulkInsert(BulkInsertOperation operation, string id, string name)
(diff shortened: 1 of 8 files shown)


src/Raven.Server/Documents/Handlers/BulkInsertHandler.cs
```diff
@@ -220,14 +226,15 @@ private long GetSize(BatchRequestParser.CommandData commandData)
...
                 case CommandType.TimeSeries:
+                case CommandType.TimeSeriesBulkInsert:
                     // we don't know the size of the change so we are just estimating
                     foreach (var append in commandData.TimeSeries.Appends)
                     {
-                        size += 2;
+                        size += sizeof(long); // DateTime
                         if (string.IsNullOrWhiteSpace(append.Tag) == false)
-                            size += 4;
+                            size += 8;
```
karmeli87: Why not `append.Tag.Length`?

src/Raven.Client/Documents/BulkInsert/BulkInsertOperation.cs
```diff
@@ -683,7 +683,7 @@ public class TimeSeriesBulkInsert : IDisposable
             private readonly string _id;
             private readonly string _name;
             private bool _first = true;
-            private const int _maxTimeSeriesInBatch = 128;
+            private const int _maxTimeSeriesInBatch = 1024;
```
karmeli87: I still think we can do better by letting the server decide on the batch size (and not the client like here).
So, should we keep it as a const so we can easily remove it if needed, or pass it as a parameter?

src/Raven.Client/Documents/BulkInsert/BulkInsertOperation.cs
```diff
@@ -769,16 +773,14 @@ private async Task AppendAsyncInternal(DateTime timestamp, IEnumerable<double> v
                             _operation._currentWriter.Write(value);
```
karmeli87: This value is a `double` which in German culture is `3,14` instead of `3.14`.
I _think_ it might be a problem and a test for it would be nice :-)

karmeli87 (approved):

src/Raven.Client/Documents/BulkInsert/BulkInsertOperation.cs
grisha-kotler: done

src/Raven.Server/Documents/Handlers/BulkInsertHandler.cs
grisha-kotler: done

src/Raven.Client/Documents/BulkInsert/BulkInsertOperation.cs
ayende: Let's do that as a parameter, yes.

src/Raven.Client/Documents/BulkInsert/BulkInsertOperation.cs
ayende: We *must* write the values using culture insensitive manner

src/Raven.Client/Documents/BulkInsert/BulkInsertOperation.cs
```diff
@@ -750,35 +748,36 @@ private async Task AppendAsyncInternal(DateTime timestamp, IEnumerable<double> v
...
                         var firstValue = true;
                         foreach (var value in values)
                         {
                             if (firstValue == false)
-                                _operation._currentWriter.Write(",");
+                                _operation.WriteComma();
 
                             firstValue = false;
-                            _operation._currentWriter.Write(value);
+                            _operation._currentWriter.Write(Convert.ToString(value, CultureInfo.InvariantCulture));
```
ppekrol: Do we need to do https://github.com/ravendb/ravendb/blob/7750eec880a0f5b185d11fce76ff74e1d3530fa5/src/Sparrow/Json/BlittableWriter.cs#L101 ?

src/Raven.Client/Documents/BulkInsert/BulkInsertOperation.cs
grisha-kotler: done

src/Raven.Client/Documents/BulkInsert/BulkInsertOperation.cs
karmeli87: I think that `EnsureDecimalPlace` is wrong?
take a look at this: https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings#GFormatString

ppekrol: We switched from G17 to R, because R is a round-trip format

karmeli87: Not according to the docs, https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings#RFormatString:
> For Double values, the "R" format specifier in some cases fails to successfully round-trip the original value

ppekrol: Please open an issue and assign it to yourself. In the meanwhile, I think that we need to use exactly the same technique there.

grisha-kotler: done

karmeli87: okay.. https://issues.hibernatingrhinos.com/issue/RavenDB-14902

PR #10476, opened 2020-04-22
RavenDB-14894

ayende:


arekpalinski (approved):
I don't see any issue with those changes. I have focused mostly on the cache introduction - I like this idea. I thought it will be totally new concept but I see we use mechanism that we already have which is good.

ayende:
Results from testing.

GetSort:
image

Lightweight thread usage:
image

I think we'll just give up on disposing on the closing of the thread. We didn't have it before, and it seems expensive.

Notice the cost of IsStale:
image

We need to account for `LastProcessedReferenceEtag`, but this is *encouraging*


ReturnRequestContext:

image

Note that we have the `RemoveAt` call there, which we should probably just skip.
We can clear the list and add the internal / external segments back directly.

Open Voron streams:
image


OTOH, we need to have another look here:
image

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1989,6 +2140,19 @@ private static void ThrowNoActiveTransactionException()
             throw new InvalidOperationException("This method requires active transaction, and no active transactions in the current context...");
         }
 
+        private IEnumerable<string> IterateCollectionNames(Transaction tx)
+        {
+            using (ContextPool.AllocateOperationContext(out JsonOperationContext context))
```
ppekrol: We can pass the context from the outside. Now we have double context allocation.

src/Raven.Server/ServerWide/Context/DocumentTransactionCache.cs
```diff
@@ -0,0 +1,25 @@
...
+        public long LastEtag;
+
+        public class CollectionCache
+        {
+            public long LastDocumentEtag;
+            public long LastTombstoneEtag;
+            public string LastChangeVector;
+        }
+
+        public Dictionary<string, CollectionCache> LastEtagsByCollection = new Dictionary<string, CollectionCache>();
```
ppekrol: OridinalIgnoreCase

src/Voron/Data/BTrees/Tree.Stream.cs
```diff
@@ -251,7 +259,10 @@ public VoronStream ReadStream(Slice key)
             using (var it = tree.Iterate())
             {
                 if (it.Seek(0) == false)
-                    return null;
+                {
+                    Debug.Assert(false);
```
ppekrol: It is hard to trace failing debug asserts when there is no message.

ppekrol (approved):

src/Raven.Server/ServerWide/ServerMetricCacher.cs
```diff
@@ -24,7 +24,7 @@ public void Initialize()
         {
             Register(MetricCacher.Keys.Server.CpuUsage, TimeSpan.FromSeconds(1), _server.CpuUsageCalculator.Calculate);
             Register(MetricCacher.Keys.Server.MemoryInfo, TimeSpan.FromSeconds(1), CalculateMemoryInfo);
-            Register(MetricCacher.Keys.Server.MemoryInfoExtended, TimeSpan.FromSeconds(1), CalculateMemoryInfoExtended);
+            Register(MetricCacher.Keys.Server.MemoryInfoExtended, TimeSpan.FromSeconds(15), CalculateMemoryInfoExtended);
```
grisha-kotler, line 27: We only need only when `CalculateMemoryInfo` returns that we are in low memory.
They need to be bound together.

src/Raven.Server/Documents/DocumentsStorage.cs
ayende: fixed

src/Raven.Server/ServerWide/Context/DocumentTransactionCache.cs
ayende: fixed

src/Voron/Data/BTrees/Tree.Stream.cs
ayende: We are actually checking a bit earlier that this can't happen.

src/Raven.Server/ServerWide/ServerMetricCacher.cs
ayende: This is for system metrics, for notifications.

ayende: This is used in `ServerTotalMemory`, not easy to make it rely on the MemoryInfo without more changes.

ppekrol: Spoke with @grisha-kotler. He said that it should be 5 second, because this is how often we check for that value in the LowMemoryNotification thread. This was probably called more often because Studio was opened and we are requesting this value when we are in low memory.

src/Raven.Server/ServerWide/Context/DocumentTransactionCache.cs
ppekrol: Have you pushed this?

src/Raven.Server/ServerWide/ServerMetricCacher.cs
ppekrol: Let's change this to 5 seconds

ppekrol (approved):

ppekrol:
Something to be aware of, this is causing some test failures.

src/Raven.Server/ServerWide/ServerMetricCacher.cs
ayende: This isn't something that we operate on, however. This is used purely for reporting purposes.
I couldn't find any use of that that we actually act on.

ayende:
Yes, I'm looking at the test failures now

src/Raven.Server/ServerWide/ServerMetricCacher.cs
grisha-kotler: This is used to determine if we are in low memory (only when the non extended version returned that we are).
It's used in the low memory notification thread.

ppekrol: @grisha-kotler you are incorrect I think, the low memory handler is accessing that directly, not via cacher

ayende:
The last optimization gives us:
image

For dynamic queries.

src/Raven.Client/Documents/Indexes/IndexFailureInformation.cs
```diff
@@ -23,7 +23,7 @@ public class IndexFailureInformation
         /// Indicates whether this is invalid index.
         /// </summary>
         /// <value><c>true</c> if this is invalid index; otherwise, <c>false</c>.</value>
-        public bool IsInvalidIndex(Func<bool> isStale)
```
ppekrol, line 26: We shouldn't change this. It was not added without a reason. Probably we are substituting one issue with another.

src/Raven.Server/Config/Categories/PatchingConfiguration.cs
```diff
@@ -13,14 +13,6 @@ public class PatchingConfiguration : ConfigurationCategory
         [ConfigurationEntry("Patching.MaxStepsForScript", ConfigurationEntryScope.ServerWideOrPerDatabase)]
         public int MaxStepsForScript { get; set; }
 
-        /// <summary>
-        /// The maximum number of javascript scripts to be cached
-        /// Default: 2048
-        /// </summary>
-        [DefaultValue(2048)]
-        [ConfigurationEntry("Patching.MaxNumberOfCachedScripts", ConfigurationEntryScope.ServerWideOrPerDatabase)]
```
ppekrol, line 21: Do we have an issue for this?

src/Raven.Server/Documents/Indexes/IndexStorage.cs
```diff
@@ -209,6 +211,32 @@ public long ReadErrorsCount()
             return DateTime.FromBinary(lastIndexingTime.Reader.ReadLittleEndianInt64());
         }
 
+        public bool IsIndexInvalid(RavenTransaction tx)
```
ppekrol, line 214: Better to implement this in Index and use IndexStorage.ReadStats witch is fairly cheap.

src/Raven.Server/Documents/Patch/ScriptRunnerCache.cs
```diff
@@ -44,6 +41,17 @@ public ScriptRunnerCache(DocumentDatabase database, RavenConfiguration configura
             LowMemoryNotification.Instance.RegisterLowMemoryHandler(this);
         }
 
+        public int NumberOfCachedScripts
+        {
+            get
+            {
+                return _cache.Values
```
ppekrol, line 48: Should we avoid LINQ here?

src/Raven.Server/ServerWide/Context/DocumentTransactionCache.cs
ppekrol: Ping

src/Raven.Client/Documents/Indexes/IndexFailureInformation.cs
ayende: I want to avoid the indirection here, and I checked, we preserve the same behavior

src/Raven.Server/Config/Categories/PatchingConfiguration.cs
ayende: yes

src/Raven.Server/Documents/Patch/ScriptRunnerCache.cs
ayende: This is only ever called from tests.

src/Raven.Client/Documents/Indexes/IndexFailureInformation.cs
ppekrol: No, you are now calculating staleness everytime instead of doing it at later stage of that function.

ayende: It doesn't actually matter, this is only called on index error. 
I don't think we need to worry about perf here.

Which means that the indirection argument is also invalid.

Need to consider which option is clearer to maintain.

This way, we don't need to worry about the state of the func that we pass, which we previous dealt with.

ppekrol (approved):
Rebased

ppekrol:
We have some tests failures here. @ayende can you take a look?

ayende:
Working on it now

PR #10477, opened 2020-04-22
RavenDB-14893 Ensure we have enough bits for new value

karmeli87:


ayende (commented):
Why do we need this?
The whole point of the temporary buffer was that we can do that check once, at the end.
As it stands now, we have to do a lot of branches in this code, which I would rather avoid.

karmeli87:
Okay, will throw a dedicated exception for that.

src/Raven.Server/Documents/TimeSeries/BitsBuffer.cs
```diff
@@ -199,7 +199,8 @@ public void SetBits(int bitsPosition, ulong value, int bitsInValue)
 
         public void AddValue(ulong value, int bitsInValue)
         {
-            Debug.Assert(HasEnoughBits(Header->UncompressedBitsPosition, bitsInValue));
+            if (HasEnoughBits(Header->UncompressedBitsPosition, bitsInValue) == false)
+                NotEnoughBitsException.Throw("Not enough space in the buffer");
```
ayende: Were you able to get this to happen?

src/Raven.Server/Documents/TimeSeries/TimeSeriesValuesSegment.cs
```diff
@@ -126,110 +126,119 @@ public static void AssertValueStatus(ulong status)
...
+                            return false;
+                        }
                     }
-                }
 
-                Memory.Copy(Header, tempHeader, copiedHeaderSize);
+                    Memory.Copy(Header, tempHeader, copiedHeaderSize);
 
-                return true;
+                    return true;
```
ayende: Look at lines 216 here, we are doing explicit checks on the temp buffer, so this shouldn't be the case.
How can you get this error?

src/Raven.Server/Documents/TimeSeries/BitsBuffer.cs
karmeli87: Yes, you can run the test.

src/Raven.Server/Documents/TimeSeries/TimeSeriesValuesSegment.cs
karmeli87: This check is too late... either the assert kills me or an `ArgumentException`/`IndexOutOfRangeException` when I'm trying to add to the `tempBitBuffer` when there is not space left.

ayende (commented):
Let's talk about it, I think that this is the wrong way to go about it.

src/Raven.Server/Documents/TimeSeries/TimeSeriesValuesSegment.cs
```diff
@@ -139,7 +139,8 @@ public bool Append(ByteStringContext allocator, int deltaFromStart, Span<double>
                 sizeof(int) + // max timestamp
                 1 + // timestamp uses ControlValue which takes 4 bits, need to account for this
                 sizeof(double) * vals.Length + vals.Length /* may use additional 2 bits per value here for the first items */ +
-                2 // previous tag position (10 bits)
+                2 + // previous tag position (10 bits)
+                1 // take into account the status
```
ayende:
```suggestion
                1 // take into account the status bit
```

ppekrol (approved):

PR #10479, opened 2020-04-22
RavenDB-14486: Add update to subscriptions Client API

garayx:


src/Raven.Client/Documents/Commands/UpdateSubscriptionCommand.cs
```diff
@@ -0,0 +1,60 @@
...
+using Raven.Client.Documents.Subscriptions;
+using Raven.Client.Http;
+using Raven.Client.Json;
+using Raven.Client.Json.Converters;
+using Raven.Client.Util;
+using Sparrow.Json;
+
+namespace Raven.Client.Documents.Commands
+{
+    public class UpdateSubscriptionCommand : RavenCommand<CreateSubscriptionResult>, IRaftCommand
```
ppekrol: internal

src/Raven.Client/Documents/Commands/UpdateSubscriptionCommand.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        private readonly SubscriptionCreationOptions _options;
+        private readonly long? _id;
+
+        public UpdateSubscriptionCommand(SubscriptionCreationOptions options, long? id = null)
+        {
+            _options = options;
+            _id = id;
```
ppekrol: I do not think that we want that? All other operations are by Name?

src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
```diff
@@ -426,6 +382,109 @@ public Task DropSubscriptionConnection()
...
+                SubscriptionState state;
+
+                if (id == null)
+                {
+                    state = Database.SubscriptionStorage.GetSubscriptionFromServerStore(options.Name);
+                    id = state.SubscriptionId;
+                }
+                else
+                {
+                    state = Database.SubscriptionStorage.GetSubscriptionFromServerStoreById(id.Value);
```
ppekrol: I do not think we want this.

src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
```diff
@@ -426,6 +382,109 @@ public Task DropSubscriptionConnection()
...
+            {
+                var json = await context.ReadForMemoryAsync(RequestBodyStream(), null);
+                var options = JsonDeserializationServer.SubscriptionCreationParams(json);
+
+                var id = GetLongQueryString("id", required: false);
+                SubscriptionState state;
+
+                if (id == null)
+                {
+                    state = Database.SubscriptionStorage.GetSubscriptionFromServerStore(options.Name);
```
ppekrol: Can this return null?

src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
```diff
@@ -426,6 +382,109 @@ public Task DropSubscriptionConnection()
 
             return NoContent();
         }
+
+        [RavenAction("/databases/*/subscriptions/update", "POST", AuthorizationStatus.ValidUser)]
+        public async Task Update()
```
ppekrol, line 387: What I think would be nice is to have the ability to create new subscription when it does not exist. Thoughts?

src/Raven.Client/Documents/Commands/UpdateSubscriptionCommand.cs
garayx: I added it because of those reasons
- support updating subscription name
- it is supported to create subscription by id:
https://github.com/ravendb/ravendb/blob/7750eec880a0f5b185d11fce76ff74e1d3530fa5/src/Raven.Client/Documents/Commands/CreateSubscriptionCommand.cs#L20

Please also note, to update the subscription, we anyway have to have the id in the handler:
https://github.com/ravendb/ravendb/pull/10479/files#diff-85b7d24872d0ec9670afa7225e9686d1R398

src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
garayx: please see the comment above 
https://github.com/ravendb/ravendb/pull/10479#discussion_r413139945

src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
garayx: yes this is if the subscription name is nulloremtpty (we are updating by id)
https://github.com/ravendb/ravendb/pull/10479/files#diff-5c857b96c03026c44f6ab3e6c11b0cdaR139

src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
garayx: I am not sure we want to mix it. 
Now the user can catch SubscriptionDoesNotExistException and send a Create command if he want a update/create

If we make it update/create by default then we will lack the only update scenario, ie when I want to just update existing subscription, it will create me a new one in case it doesn't exist, which seems odd.

src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
garayx: no, it will throw SubscriptionDoesNotExistException if ther eis no such subscription 
https://github.com/ravendb/ravendb/blob/v4.2/src/Raven.Server/Documents/Subscriptions/SubscriptionStorage.cs#L346

src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
ppekrol: I'm thinking more about the usability, like we do have with indexes. I'm just doing .Execute and do not worry if the index is there or not.

Do we have .Exists method at least?

src/Raven.Client/Documents/Subscriptions/DocumentSubscriptions.cs
```diff
@@ -457,5 +457,31 @@ public Task DisableAsync(string name, string database = null, CancellationToken
             var operation = new ToggleOngoingTaskStateOperation(name, OngoingTaskType.Subscription, disable: true);
             return _store.Maintenance.ForDatabase(database ?? _store.Database).SendAsync(operation, token);
         }
+
+        /// <summary>
+        /// Update a data subscription in a database.
+        /// </summary>
+        public string Update(SubscriptionCreationOptions options, string database = null, long? id = null)
```
ppekrol: introduce `SubscriptionUpdateOptions` that inherit from `SubscriptionCreationOptions`. It will allow us to extend them in the future if needed. Maybe they will not be 1 to 1 with creation ones.

ppekrol: We can move then the Id to the options and maybe do some validation client side.

src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
ayende: Note that updating the subscription is something that you should do with some thinking about the current change vector.
Are you resetting it or preserving it?

garayx: there is no exists method.

garayx: I think we can add Overwrite option to SubscriptionUpdateOptions ?

garayx: > Note that updating the subscription is something that you should do with some thinking about the current change vector.
> Are you resetting it or preserving it?

It obey the value of `SubscriptionCreationOptions.ChangeVector`, see test:
https://github.com/ravendb/ravendb/pull/10479/files#diff-2a118f7ce3e7b0101c6c2c0894fefaeaR433

ppekrol: Yes, we can add Overwrite property there.

src/Raven.Client/Documents/Commands/UpdateSubscriptionCommand.cs
```diff
@@ -0,0 +1,57 @@
...
+using Raven.Client.Documents.Subscriptions;
+using Raven.Client.Http;
+using Raven.Client.Json;
+using Raven.Client.Json.Converters;
+using Raven.Client.Util;
+using Sparrow.Json;
+
+namespace Raven.Client.Documents.Commands
+{
+    internal class UpdateSubscriptionCommand : RavenCommand<CreateSubscriptionResult>, IRaftCommand
```
ppekrol: CreateSubscriptionResult -> UpdateSubscriptionResult

ppekrol (approved):

src/Raven.Client/Documents/Commands/UpdateSubscriptionCommand.cs
garayx: CreateSubscriptionResult just have a subscription name
https://github.com/ravendb/ravendb/blob/v4.2/src/Raven.Client/Documents/Subscriptions/CreateSubscriptionResult.cs#L5

ppekrol: Just inherit from it, that is all.

PR #10482, opened 2020-04-23
RavenDB-14909 - Unexpected exception occurred during indexing

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-14909

### DIFF
--- a/src/Voron/Exceptions/VoronUnrecoverableErrorException.cs
+++ b/src/Voron/Exceptions/VoronUnrecoverableErrorException.cs
@@ -10,7 +10,7 @@ using Voron.Impl;
 
 namespace Voron.Exceptions
 {
-    public class VoronUnrecoverableErrorException : Exception
+    public class VoronUnrecoverableErrorException : ApplicationException
     {
         public static void Raise(LowLevelTransaction tx, string message)
         {
--- a/src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexWriter.cs
+++ b/src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexWriter.cs
@@ -65,33 +65,24 @@ namespace Raven.Server.Documents.Indexes.Persistence.Lucene
             {
                 _indexWriter.Commit(state);
             }
-            catch (EarlyOutOfMemoryException)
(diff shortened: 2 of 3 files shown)


arekpalinski:
@grisha-kotler Can you elaborate on the root cause? As I understand we didn't recreate the writer after some exception what resulted in invalid Voron state of Lucene input / ouput files

grisha-kotler:
We had a `VoronUnrecoverableErrorException`, the exception that was thrown was a `SystemException`. We tried to execute another indexing batch without recreating the writer.
(modifying when the writer is already in an invalid state).

src/Voron/Exceptions/VoronUnrecoverableErrorException.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace Voron.Exceptions
 {
-    public class VoronUnrecoverableErrorException : Exception
+    public class VoronUnrecoverableErrorException : ApplicationException
```
arekpalinski, line 13: Why do we need that?

grisha-kotler: Lucene is wrapping the exception with a SystemException except of IOException, SystemException and ApplicationException

https://github.com/ravendb/lucenenet/blob/3.0.3-netcore/src/Lucene.Net/Index/IndexWriter.cs#L4636

arekpalinski: Can you please add a comment here explaining what is the reason? Otherwise it's easy to forget about it

arekpalinski: Nevermind, we'll have it in github

ayende:
I really don't like this change, can we avoid modifying the exception hierarchy here?

ppekrol:
@grisha-kotler how Lucene is wrapping those exceptions? Can we check inner exception?

grisha-kotler:
https://github.com/ravendb/lucenenet/blob/3.0.3-netcore/src/Lucene.Net/Index/IndexWriter.cs#L4636

We can check the inner exception.

PR #10488, opened 2020-04-25
RavenDB-14913 do not calculate memory stats always when retrieving index stats

ppekrol:


ayende:
Can you post profiler details?

src/Raven.Server/Documents/Queries/Dynamic/DynamicQueryToIndexMatcher.cs
```diff
@@ -57,61 +59,103 @@ public Explanation(string index, string reason)
...
+
+                    list ??= new List<AutoIndexDefinitionBase>();
+
+                    list.Add((AutoIndexDefinitionBase)index.Definition);
+                }
+
+                return list?.Count > 0;
+            }
+
+            bool TryConvertToMatches(out Dictionary<DynamicQueryMatchType, List<DynamicQueryMatchResult>> dictionary)
```
ayende: Do we have to do it like this?
Can't we just use a local variable for storing the complete option (and if finding another one, switching if we have a better ranking)?

ppekrol: Actually this isn't even close to main cost there. I think I can optimize it even further.

BTW. Here are the results:
image

ppekrol: The main cost is here:

image

We are getting stats to check if index is invalid.

1. We probably do not need to take whole stats
2. GetMemoryStats is taking 77% of the time from which 58% are taken by GetDirectoryName!

ppekrol: Still need to check where we actually want to retrieve that Memory stats, but:

image

We are down from 49300ms to 10500ms now.

ayende: I have another way to do that, and I think it reduces a lot of the collection code

ayende:
See my impl here: https://github.com/ravendb/ravendb/pull/10476/commits/25bff973a815950525a2e067af642b16d16a75c7

PR #10490, opened 2020-04-26
RavenDB-14564 Query from multiple time-series at once

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-14564

### DIFF
--- a/src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesConfiguration.cs
+++ b/src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesConfiguration.cs
@@ -6,7 +6,7 @@ namespace Raven.Client.Documents.Operations.TimeSeries
 {
     public class TimeSeriesConfiguration : IDynamicJson
     {
-        internal const char TimeSeriesRollupSeparator = '@';
+        internal const char TimeSeriesRollupSeparator = '_';
         public Dictionary<string, TimeSeriesCollectionConfiguration> Collections { get; set; }
 
         public TimeSpan PolicyCheckFrequency { get; set; } = TimeSpan.FromMinutes(10);
--- a/src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesConfiguration.cs
+++ b/src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesConfiguration.cs
@@ -6,7 +6,7 @@ namespace Raven.Client.Documents.Operations.TimeSeries
 {
     public class TimeSeriesConfiguration : IDynamicJson
     {
-        internal const char TimeSeriesRollupSeparator = '_';
+        internal const char TimeSeriesRollupSeparator = '@';
         public Dictionary<string, TimeSeriesCollectionConfiguration> Collections { get; set; }
 
         public TimeSpan PolicyCheckFrequency { get; set; } = TimeSpan.FromMinutes(10);
--- a/src/Raven.Client/Documents/Queries/TimeSeries/TimeSeriesAggregation.cs
+++ b/src/Raven.Client/Documents/Queries/TimeSeries/TimeSeriesAggregation.cs
@@ -27,12 +27,17 @@ namespace Raven.Client.Documents.Queries.TimeSeries
 
     public enum AggregationType
     {
-        First,
-        Last,
-        Min,
-        Max,
-        Sum,
-        Count,
-        Average
+        // The order here matters.
+        // When executing an aggregation query over rolled-up series,
+        // we take just the appropriate aggregated value from each entry, 
+        // according to the aggregation's position in this enum (e.g. AggregationType.Min => take entry.Values[2])
+
+        First = 0,
+        Last = 1,
+        Min = 2,
+        Max = 3,
+        Sum = 4,
+        Count = 5,
+        Average = 6
     }
 }
--- a/src/Raven.Client/Documents/Queries/TimeSeries/TimeSeriesAggregation.cs
+++ b/src/Raven.Client/Documents/Queries/TimeSeries/TimeSeriesAggregation.cs
@@ -27,12 +27,12 @@ namespace Raven.Client.Documents.Queries.TimeSeries
 
     public enum AggregationType
     {
-        Min,
-        Max,
-        Average,
         First,
         Last,
+        Min,
+        Max,
         Sum,
-        Count
+        Count,
+        Average
     }
 }
--- a/src/Raven.Server/Documents/Queries/AST/Query.cs
+++ b/src/Raven.Server/Documents/Queries/AST/Query.cs
@@ -612,7 +612,8 @@ namespace Raven.Server.Documents.Queries.AST
                         throw new ArgumentOutOfRangeException("Unknown aggregation operation: " + Aggregation);
                 }
 
-                _count[i] = (long)_count[i] + values[index + (int)AggregationType.Count].Count;
+                val = values[index + (int)AggregationType.Count];
+                _count[i] = (long)_count[i] + val.Count;
             }
         }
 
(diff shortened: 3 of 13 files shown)


src/Raven.Server/Documents/Queries/AST/Query.cs
```diff
@@ -123,6 +124,7 @@ public unsafe class TimeSeriesFunction
         public List<(QueryExpression, StringSegment?)> Select;
         public StringSegment? LoadTagAs;
         public TimeSpan? Offset;
+        public bool Roolups;
```
ayende:
```suggestion
        public bool Rollups;
```

src/Raven.Client/Documents/Queries/TimeSeries/TimeSeriesAggregation.cs
```diff
@@ -27,12 +27,12 @@ public class TimeSeriesRangeAggregation
 
     public enum AggregationType
```
ayende, line 28: Add comment here specifying that the order matters.
And user explicit values.

src/Raven.Server/Documents/Queries/AST/Query.cs
```diff
@@ -432,6 +448,120 @@ public void Step(Span<double> values)
...
+            }
+
+            for (int i = 0; i < originalNumOfValues; i++)
+            {
+                var index = i * 6;
+                StatefulTimestampValue val;
+                switch (Aggregation)
+                {
+                    case AggregationType.Min:
+                        val = values[index + (int)AggregationType.Min];
```
ayende: You can replace all of these with a single command before the swtich
```suggestion
                        val = values[index + (int)Aggregation];
```

src/Raven.Server/Documents/Queries/AST/Query.cs
```diff
@@ -432,6 +448,120 @@ public void Step(Span<double> values)
...
+            }
+
+            for (int i = 0; i < originalNumOfValues; i++)
+            {
+                var index = i * 6;
+                double val;
+                switch (Aggregation)
+                {
+                    case AggregationType.Min:
+                        val = values[index + (int)AggregationType.Min];
```
ayende, line 640: Can move all of these before the switch

src/Raven.Server/Documents/Queries/Results/TimeSeriesRetriever.cs
```diff
@@ -598,15 +603,30 @@ object GetValue(QueryExpression expression, SingleResult singleResult)
...
+                            if (_currentIsRaw)
+                                return singleResult.Values.Span[index];
+
+                            // we are working with a rolled-up series
+                            // here an entry has 6 different values (min, max, first, last, sum, count) per each 'original' measurement 
+                            // we need to return the average value (sum / count)
+                            index *= 6;
+                            if (index + (int)AggregationType.Count >= singleResult.Values.Length)
+                                return null;
+                            return singleResult.Values.Span[index + (int)AggregationType.Sum] / singleResult.Values.Span[index + (int)AggregationType.Count];
```
ayende, line 611: Can this cause divide by zero?

test/SlowTests/Client/TimeSeries/Query/QueryFromMultipleTimeSeries.cs
```diff
@@ -0,0 +1,1588 @@
...
+                await database.TimeSeriesPolicyRunner.DoRetention();
+
+                await VerifyFullPolicyExecution(store, config.Collections["Users"]);
+
+                using (var session = store.OpenSession())
+                {
+                    var query = session.Advanced.RawQuery<TimeSeriesAggregationResult>(@"
+declare timeseries out() 
+{
+    from Heartrate
```
ayende, line 166: We give no indication that we switch to the rolled up?

ayende (changes requested):

src/Raven.Server/Documents/Queries/Results/TimeSeriesRetriever.cs
```diff
@@ -870,5 +890,107 @@ private static DynamicJsonValue AddTimeSeriesResult(TimeSeriesFunction func, Tim
...
+            public MultiReader(TimeSeriesRetriever retriever, string documentId,
+                string source, TimeSpan? offset)
+            {
+                if (string.IsNullOrEmpty(source)) 
+                    throw new ArgumentNullException(nameof(source));
+                _source = source;
+                _docId = documentId ?? throw new ArgumentNullException(nameof(documentId));
+                _offset = offset;
+                _current = 0;
+                _retriever = retriever;
```
karmeli87: Can we decouple this from the `retriever`? and pass `collection` and `context` instead?

karmeli87 (approved):

src/Raven.Server/Documents/Queries/Results/TimeSeriesRetriever.cs
aviv86: done

test/SlowTests/Client/TimeSeries/Query/QueryFromMultipleTimeSeries.cs
aviv86: no, we give no indication of that

src/Raven.Client/Documents/Queries/TimeSeries/TimeSeriesAggregation.cs
aviv86: done

src/Raven.Server/Documents/Queries/AST/Query.cs
aviv86: removed this

src/Raven.Server/Documents/Queries/Results/TimeSeriesRetriever.cs
aviv86: I don't think so.. but I added a check for this, just to be safe

src/Raven.Server/Documents/Queries/AST/Query.cs
```diff
@@ -464,35 +463,34 @@ private void SegmentOnRollup(Span<StatefulTimestampValue> values)
...
                     case AggregationType.First:
-                        val = values[index + (int)AggregationType.First];
                         if ((long)_count[i] == 0)
                             _values[i] = val.First;
                         break;
                     case AggregationType.Last:
-                        val = values[index + (int)AggregationType.Last];
                         _values[i] = val.Last;
                         break;
                     case AggregationType.Count:
```
ayende, line 609: Make the change to use `val` for the `_count` below as well

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesConfiguration.cs
```diff
@@ -6,7 +6,7 @@ namespace Raven.Client.Documents.Operations.TimeSeries
 {
     public class TimeSeriesConfiguration : IDynamicJson
     {
-        internal const char TimeSeriesRollupSeparator = '@';
+        internal const char TimeSeriesRollupSeparator = '_';
```
ppekrol: It means that we cannot use that char in TS name. The `_` is more popular than `@` I think?

ppekrol (approved):

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesConfiguration.cs
ayende: Where is this actually used?

src/Raven.Server/Documents/Queries/AST/Query.cs
aviv86: done

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesConfiguration.cs
aviv86: reverted it back to '@'

(The reason for the change was because we originally thought that the user will pass the roll-up names in query, and this '@' caused QueryScanner to not recognize these names as 'FieldExpression'.. We decided to go a different way so this doesn't matter anymore..)

ayende (approved):

PR #10491, opened 2020-04-26
RavenDB-14881 Added a separate endpoint to receive more information about collections

JnsSo:
https://issues.hibernatingrhinos.com/issue/RavenDB-14881

### DIFF
--- a/src/Raven.Client/Documents/Operations/CollectionStatistics.cs
+++ b/src/Raven.Client/Documents/Operations/CollectionStatistics.cs
@@ -24,8 +24,8 @@ namespace Raven.Client.Documents.Operations
             Collections = new Dictionary<string, CollectionDetails>();
         }
 
-        public int CountOfDocuments { get; set; }
-        public int CountOfConflicts { get; set; }
+        public long CountOfDocuments { get; set; }
+        public long CountOfConflicts { get; set; }
 
         public Dictionary<string, CollectionDetails> Collections { get; set; }
     }
--- a/src/Raven.Client/Documents/Operations/CollectionStatistics.cs
+++ b/src/Raven.Client/Documents/Operations/CollectionStatistics.cs
@@ -1,17 +1,70 @@
 ﻿using System.Collections.Generic;
+using Raven.Client.Util;
+using Sparrow.Json.Parsing;
 
 namespace Raven.Client.Documents.Operations
 {
-    public class CollectionStatistics
+    public class CollectionStatistics : IDynamicJson
     {
         public CollectionStatistics()
         {
             Collections = new Dictionary<string, long>();
         }
 
-        public int CountOfDocuments { get; set; }
-        public int CountOfConflicts { get; set; }
+        public long CountOfDocuments { get; set; }
+        public long CountOfConflicts { get; set; }
 
         public Dictionary<string, long> Collections { get; set; }
+
+        public virtual DynamicJsonValue ToJson()
+        {
+            return new DynamicJsonValue
+            {
+                [nameof(CountOfDocuments)] = CountOfDocuments,
+                [nameof(CountOfConflicts)] = CountOfConflicts,
+                [nameof(Collections)] = DynamicJsonValue.Convert(Collections)
+            };
+        }
+    }
+
+    public class DetailedCollectionStatistics : CollectionStatistics
+    {
+        public DetailedCollectionStatistics()
+        {
+            ExtendedCollectionDetails = new Dictionary<string, CollectionDetails>();
+        }
+
+        public Dictionary<string, CollectionDetails> ExtendedCollectionDetails { get; set; }
+
+        public override DynamicJsonValue ToJson()
+        {
+            return new DynamicJsonValue
+            {
+                [nameof(CountOfDocuments)] = CountOfDocuments,
+                [nameof(CountOfConflicts)] = CountOfConflicts,
+                [nameof(Collections)] = DynamicJsonValue.Convert(Collections),
+                [nameof(ExtendedCollectionDetails)] = DynamicJsonValue.Convert(ExtendedCollectionDetails)
+            };
+        }
+    }
+
+    public class CollectionDetails : IDynamicJson
+    {
+        public string Name { get; set; }
+        public long CountOfDocuments { get; set; }
+        public Size Size { get; set; }
+
+        public DynamicJsonValue ToJson()
+        {
+            return new DynamicJsonValue
+            {
+                [nameof(CountOfDocuments)] = CountOfDocuments,
+                [nameof(Size)] = new DynamicJsonValue
+                {
+                    [nameof(Size.HumaneSize)] = Size.HumaneSize,
+                    [nameof(Size.SizeInBytes)] = Size.SizeInBytes
+                }
+            };
+        }
     }
 }
--- a/src/Raven.Client/Documents/Operations/CollectionStatistics.cs
+++ b/src/Raven.Client/Documents/Operations/CollectionStatistics.cs
@@ -4,48 +4,30 @@ using Sparrow.Json.Parsing;
 
 namespace Raven.Client.Documents.Operations
 {
-    public class CollectionStatistics : IDynamicJson
+    public class CollectionStatistics
     {
         public CollectionStatistics()
         {
             Collections = new Dictionary<string, long>();
         }
 
-        public long CountOfDocuments { get; set; }
-        public long CountOfConflicts { get; set; }
+        public int CountOfDocuments { get; set; }
+        public int CountOfConflicts { get; set; }
 
         public Dictionary<string, long> Collections { get; set; }
-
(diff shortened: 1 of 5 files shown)


src/Raven.Client/Documents/Operations/CollectionStatistics.cs
```diff
@@ -1,17 +1,70 @@
...
     {
         public CollectionStatistics()
         {
             Collections = new Dictionary<string, long>();
         }
 
-        public int CountOfDocuments { get; set; }
-        public int CountOfConflicts { get; set; }
+        public long CountOfDocuments { get; set; }
+        public long CountOfConflicts { get; set; }
```
ayende: @ppekrol Backward compact issue here?

src/Raven.Client/Documents/Operations/CollectionStatistics.cs
```diff
@@ -1,17 +1,70 @@
...
+            return new DynamicJsonValue
+            {
+                [nameof(CountOfDocuments)] = CountOfDocuments,
+                [nameof(CountOfConflicts)] = CountOfConflicts,
+                [nameof(Collections)] = DynamicJsonValue.Convert(Collections)
+            };
+        }
+    }
+
+    public class DetailedCollectionStatistics : CollectionStatistics
```
ayende: I think it would be better to *not* inherit here.

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1799,6 +1800,30 @@ public IEnumerable<CollectionStats> GetCollections(DocumentsOperationContext con
...
+                if (collectionTable != null)
+                {
+                    TableReport tableReport = collectionTable.GetReport(false);
+                    if (tableReport != null)
+                    {
+                        collectionDetails.Size.SizeInBytes = tableReport.DataSizeInBytes;
+                    } else { }
+                } else { }
+
+            } else { }
```
ayende: What are all these `else {} ` ?

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1799,6 +1800,30 @@ public IEnumerable<CollectionStats> GetCollections(DocumentsOperationContext con
...
+
+            if (collectionName != null)
+            {
+                Table collectionTable = context.Transaction.InnerTransaction.OpenTable(DocsSchema,
+                    collectionName.GetTableName(CollectionTableType.Documents));
+
+                if (collectionTable != null)
+                {
+                    TableReport tableReport = collectionTable.GetReport(false);
+                    if (tableReport != null)
```
ayende: Under what scenario can this happen?

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1799,6 +1800,30 @@ public IEnumerable<CollectionStats> GetCollections(DocumentsOperationContext con
...
 
+        public CollectionDetails GetCollectionDetails(KeyValuePair<string, long> collection, DocumentsOperationContext context)
+        {
+            CollectionDetails collectionDetails = new CollectionDetails() { Name = collection.Key, CountOfDocuments = collection.Value, Size = new Client.Util.Size() };
+            CollectionName collectionName = GetCollection(collection.Key, throwIfDoesNotExist: false);
+
+            if (collectionName != null)
+            {
+                Table collectionTable = context.Transaction.InnerTransaction.OpenTable(DocsSchema,
+                    collectionName.GetTableName(CollectionTableType.Documents));
```
ayende: Also need to take into account:
* Conflicts, Revisions and Tombstones.

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1799,6 +1800,30 @@ public IEnumerable<CollectionStats> GetCollections(DocumentsOperationContext con
             }
         }
 
+        public CollectionDetails GetCollectionDetails(KeyValuePair<string, long> collection, DocumentsOperationContext context)
```
ayende:
```suggestion
        public CollectionDetails GetCollectionDetails(DocumentsOperationContext context, string collection)
```

src/Raven.Server/Documents/Handlers/CollectionsHandler.cs
```diff
@@ -16,29 +15,56 @@ public class CollectionsHandler : DatabaseRequestHandler
...
+        [RavenAction("/databases/*/collections/stats/detailed", "GET", AuthorizationStatus.ValidUser)]
+        public Task GetDetailedCollectionStats()
+        {
+            using (ContextPool.AllocateOperationContext(out DocumentsOperationContext context))
+            {
+                DetailedCollectionStatistics detailedCollectionStatistics = new DetailedCollectionStatistics();
+
+                FillCollectionStats(detailedCollectionStatistics, context);
+
+                using (context.OpenReadTransaction())
```
ayende: Switch these lines and make `FillCollectionStats` expect an open tx, would save two tx here.

src/Raven.Client/Documents/Operations/CollectionStatistics.cs
ppekrol: Yes, should be 'int'. We can change that in 5.0 only.

src/Raven.Client/Documents/Operations/CollectionStatistics.cs
```diff
@@ -1,17 +1,70 @@
...
+    {
+        public DetailedCollectionStatistics()
+        {
+            ExtendedCollectionDetails = new Dictionary<string, CollectionDetails>();
+        }
+
+        public Dictionary<string, CollectionDetails> ExtendedCollectionDetails { get; set; }
```
ppekrol: In this case, since we already moved that to other class and endpoint, maybe lets not inherit and substitute 'Collections' from 'CollectionStatistics' with `Dictionary<string, CollectionDetails>` and put there the etag as well.

JnsSo: Sorry, I didn't quite understand. I will remove the inheritance, but what do you mean by "substitute 'Collections' from 'CollectionStatistics' with Dictionary<string, CollectionDetails> and put there the etag as well"?

Did you mean I should remove the 'Collection' property from the 'DetailedCollectionStatistics' class, because the 'CollectionDetails' class also contains the number of documents?

src/Raven.Server/Documents/DocumentsStorage.cs
JnsSo: Isn't the DocSchema in this context always "Documents"? How can I take your mentioned types into account? Could you give me an advice, please?

ayende: Yes, `DocsSchema` is for documents. 
But a collection is also composed of other items, listed above. 
Each one of them has its own schema. See `RevisionStorage` for example.

`Revisions`, in particular, may be a significant portion of a collection size

src/Raven.Client/Documents/Operations/CollectionStatistics.cs
ayende: If we drop the inheritance between these classes, then we don't need `ExtendedCollectionDetails`, we can call the property `Collections`. That would make things simpler.

src/Raven.Server/Documents/DocumentsStorage.cs
JnsSo: Thanks for the advice.
Conflicts seems to be complicated, because they are not directly associated with a collection.
There is no CollectionTableType for conflicts. I tried to use the values of ConflictsStorage.ConflictsSchema and ConflictsStorage.ConflictsSlice to get the table + report, but the result is not referred to a collection. Or am I wrong in that?

arekpalinski: Yes, you're right. Conflicts don't have per collection table. Only the following ones:

https://github.com/ravendb/ravendb/blob/5ee2e9602d646c4f3e50633854b3eb1bbf6fb2cd/src/Raven.Server/Documents/CollectionName.cs#L64-L70

`CollectionTableType.Counters` is legacy value, currently we use `CounterGroups`

src/Raven.Client/Documents/Operations/CollectionStatistics.cs
```diff
@@ -14,4 +16,37 @@ public CollectionStatistics()
...
+
+    public class DetailedCollectionStatistics
+    {
+        public DetailedCollectionStatistics()
+        {
+            Collections = new Dictionary<string, CollectionDetails>();
+        }
+
+        public int CountOfDocuments { get; set; }
+        public int CountOfConflicts { get; set; }
```
ayende: Shouldn't you use `long` here?

ayende (approved):
Aside from the long issue, LGTM

ppekrol (approved):

PR #10492, opened 2020-04-27
RavenDB-14916 await method to avoid buffer leak

ppekrol:


ayende (approved):

ayende:
How did you run into this?

ppekrol:
During ManagedPinnedBuffer removal work. I've added a lot of code that only runs in Debug that checks if the new unmanaged buffer isn't released.

PR #10494, opened 2020-04-27
RavenDB-10412 removed ManagedPinnedBuffer

ppekrol:


ppekrol:
For review only.

- we should apply that to 5.0 only, I think
- will try to get rid of the Valid/Used

src/Raven.Server/Documents/AttachmentsStorageHelper.cs
```diff
@@ -12,58 +12,58 @@ public static class AttachmentsStorageHelper
...
-        private static unsafe string FinalizeGetHash(JsonOperationContext.ManagedPinnedBuffer cryptoState, JsonOperationContext.ManagedPinnedBuffer buffer)
+        private static unsafe string FinalizeGetHash(JsonOperationContext.MemoryBuffer cryptoState, JsonOperationContext.MemoryBuffer buffer)
         {
             var size = Sodium.crypto_generichash_bytes();
             var rc = Sodium.crypto_generichash_final(cryptoState.Pointer, buffer.Pointer, size);
             if (rc != 0)
                 throw new InvalidOperationException("Unable to hash attachment: " + rc);
 
-            return Convert.ToBase64String(buffer.Buffer.Array, buffer.Buffer.Offset, (int)size);
+            return Convert.ToBase64String(buffer.Memory.Span.Slice((int)size));
```
ayende: I think you need this:

```suggestion
            return Convert.ToBase64String(buffer.Memory.Span.Slice(0, (int)size));
```

src/Raven.Server/Rachis/Remote/RemoteConnection.cs
```diff
@@ -40,7 +41,7 @@ public RemoteConnection(string dest, string src, long term, Stream stream, Actio
             _src = src;
             _stream = stream;
             _disconnect = disconnect;
-            _buffer = JsonOperationContext.ManagedPinnedBuffer.LongLivedInstance();
+            _releaseBuffer = JsonOperationContext.MemoryBuffer.ShortTermSingleUse(out _buffer);
```
ayende: Why not use the usual context & buffer here?
I don't really like replacing `LongLivedInstance` with `ShortTermSingleUse`

src/Raven.Server/Smuggler/Documents/Handlers/SmugglerHandler.cs
```diff
@@ -97,9 +97,9 @@ public async Task PostExport()
 
 
                 DatabaseSmugglerOptionsServerSide options;
-                using (context.GetManagedBuffer(out var buffer))
+                using (context.GetMemoryBuffer(out var buffer))
                 {
-                    var firstRead = await stream.ReadAsync(buffer.Buffer.Array, buffer.Buffer.Offset, buffer.Buffer.Count);
+                    var firstRead = await stream.ReadAsync(buffer.Memory);
                     buffer.Used = 0;
                     buffer.Valid = firstRead;
```
ayende, line 104: This usage is the reason why I don't think that you _can_ remove the `Used` / `Valid` behavior
We can try to use `Slice`, but we'll need access to the original buffer when we need to refill it anyway

src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -379,50 +191,228 @@ void TryExecute(Action a)
...
+
+                Valid = 0;
+                Used = 0;
+#if DEBUG
+                _generation = generation;
+                _parent = context;
+#endif
+            }
+
+            public static ReturnBuffer ShortTermSingleUse(out MemoryBuffer buffer)
```
ayende: Why do we need this? We want to use this only for unmanaged memory

src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -379,50 +191,228 @@ void TryExecute(Action a)
...
+
+                buffer = new MemoryBuffer(memory, 0, null);
+
+                return new ReturnBuffer(bytes, memoryHandle, buffer);
+            }
+
+            internal (IDisposable ReleaseBuffer, MemoryBuffer Buffer) Clone<T>(JsonContextPoolBase<T> pool)
+                where T : JsonOperationContext
+            {
+                if (Length != Size)
```
ayende: I don't understand why we have this limitation?

src/Raven.Server/Documents/AttachmentsStorageHelper.cs
ppekrol: Yes, pushed already, but thanks :)

src/Raven.Server/Rachis/Remote/RemoteConnection.cs
ppekrol: Will rename and try to pass something here externally.

src/Raven.Server/Smuggler/Documents/Handlers/SmugglerHandler.cs
ppekrol: Yeah, I was researching if we can remove this, but I do not think it feasible.

src/Sparrow/Json/JsonOperationContext.cs
ppekrol: We have 1 usage, will try to get rid of this.

src/Sparrow/Json/JsonOperationContext.cs
ppekrol: Previosuly we had small buffers (32kb and large ones 128kb). Can be removed now.

ayende (approved):

PR #10495, opened 2020-04-27
RavenDB-14707 Send a DELETE command when IsEmptyScript is false

vvanderkroef:
This change will make sure a DELETE command will be send when the ETL task contains a script. This is needed because sometimes you want to filter what will be send to the destination database.

In the (added) test case a transform script is present that will make sure only documents with Name "Joe Doe" are loaded to the destination database. The fix makes sure the document is deleted when at a later time the name changes.

I'va also changed test case for RavenDB_9072 because the fix results in a extra DELETE command so Commands.Count changes from 4 to 5.


ppekrol:
I've added your '58553893+vvanderkroef@users.noreply.github.com' e-mail to our database of contributors.

PR #10499, opened 2020-04-29
Raven db 14789

ml054:


src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
```diff
@@ -445,6 +446,43 @@ public async Task Batch()
             }
         }
 
+        [RavenAction("/databases/*/timeseries/config", "GET", AuthorizationStatus.ValidUser)]
```
ppekrol, line 449: Move this to ConfigurationHandler and normalize the url

ppekrol: Please ignore, I see that this is consistent with Revisions.

PR #10501, opened 2020-04-30
RavenDB-14909 - avoid modifying the exception hierarchy

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-14909

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexWriter.cs
+++ b/src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexWriter.cs
@@ -67,13 +67,13 @@ namespace Raven.Server.Documents.Indexes.Persistence.Lucene
             catch (SystemException e)
             {
                 if (e.Message.StartsWith("this writer hit an OutOfMemoryError"))
-                    ThrowOutOfMemory(e);
+                    ThrowOutOfMemoryException(e);
 
                 if (e is Win32Exception win32Exception && win32Exception.IsOutOfMemory())
-                    ThrowOutOfMemory(e);
+                    ThrowOutOfMemoryException(e);
 
                 if (e.InnerException is VoronUnrecoverableErrorException)
-                    throw e.InnerException;
+                    ThrowVoronUnrecoverableErrorException(e);
 
                 throw;
             }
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexWriter.cs
```diff
@@ -73,6 +72,9 @@ public void Commit(IState state)
                 if (e is Win32Exception win32Exception && win32Exception.IsOutOfMemory())
                     ThrowOutOfMemory(e);
 
+                if (e.InnerException is VoronUnrecoverableErrorException)
+                    throw e.InnerException;
```
ayende: We are now losing the inner stack here.
Maybe throw that in a *good* exception?

src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneSuggestionIndexWriter.cs
```diff
@@ -204,6 +203,9 @@ public void Commit(IState state)
                 if (e is Win32Exception win32Exception && win32Exception.IsOutOfMemory())
                     LuceneIndexWriter.ThrowOutOfMemory(e);
 
+                if (e.InnerException is VoronUnrecoverableErrorException)
+                    throw e.InnerException;
```
ayende: Same

ayende (changes requested):

src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexWriter.cs
grisha-kotler: Do we need the inner stack? (this exception is wrapped)
https://github.com/ravendb/lucenenet/blob/3.0.3-netcore/src/Lucene.Net/Index/IndexWriter.cs#L4644

ayende: We never want to lose the stack

grisha-kotler: done

ppekrol (approved):

PR #10502, opened 2020-04-30
RavenDB-14922 - Support SAS tokens in Azure Backup

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-14922


ml054 (approved):

src/Raven.Server/Documents/PeriodicBackup/Azure/RavenAzureClient.cs
```diff
@@ -29,61 +29,137 @@ namespace Raven.Server.Documents.PeriodicBackup.Azure
...
+                if (hasParameters)
+                {
+                    url += "&";
+                }
+                else
+                {
+                    url += "?";
+                }
+
+                url += _sasToken;
```
ppekrol, line 157: Do we need to encode the token?

ppekrol (approved):

src/Raven.Server/Documents/PeriodicBackup/Azure/RavenAzureClient.cs
grisha-kotler: It is already escaped (using the Azure Portal and the Blob API)

PR #10503, opened 2020-04-30
Add to PAL the ability to build the zstd libs

gregolsky:


src/Raven.Pal/build-libs/zstd-build-deps.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+
+export ZSTD_LOG=${PWD}/build_zstd.log
+
+LIBZSTD_VER="ravendb"
+LIBZSTD_REPO="https://github.com/gregolsky/zstd.git"
```
ppekrol: We should move this to `ravendb`

.github/workflows/FastTests.yml
```diff
@@ -0,0 +1,36 @@
+name: FastTests
+
+on:
+  push:
+    branches:
+        - v5.0
+        - v4.2
+        - zstd
```
ppekrol: Should be removed?

ppekrol (approved):

.github/workflows/FastTests.yml
gregolsky: Yes, I think we might want to get that file to 4.2 branch as well right?

PR #10504, opened 2020-05-01
RavenDB-14905 Allow the use of unix ticks for time-series bulk-insert

karmeli87:


ayende (commented):
Where is this used?

karmeli87:
This will be used in the Java client, and in other clients as well.
Also needed for the tsbs benchmark, which is written in Go

ayende (approved):

ppekrol:
@karmeli87 can we write a simple test for this?

ppekrol (approved):

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesOperation.cs
```diff
@@ -64,11 +64,27 @@ internal static TimeSeriesOperation Parse(BlittableJsonReaderObject input)
             return result;
         }
 
+        private const string TimeSeriesFormat = "TimeFormat";
+        private const string UnixTimeFormatMs = "UnixTimeFormatMs";
+        private const string UnixTimeFormatNs = "UnixTimeFormatNs";
+        private static readonly long _epochTicks = new DateTime(1970, 1, 1).Ticks;
```
ppekrol: Just to be sure, this shouldn't be `new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc).Ticks` ?

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesOperation.cs
```diff
@@ -64,11 +64,27 @@ internal static TimeSeriesOperation Parse(BlittableJsonReaderObject input)
...
+        {
+            return unixNs / 100 + _epochTicks;
+        }
+
         internal static TimeSeriesOperation ParseForBulkInsert(BlittableJsonReaderObject input)
         {
             if (input.TryGet(nameof(Name), out string name) == false || name == null)
                 ThrowMissingProperty(nameof(Name));
 
+            input.TryGet(TimeSeriesFormat, out string format);
```
ppekrol: Hmm... I would prefer to convert this value to enum and if it exists parse it. If it cannot be parsed then we should throw.

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesOperation.cs
```diff
@@ -64,11 +64,27 @@ internal static TimeSeriesOperation Parse(BlittableJsonReaderObject input)
             return result;
         }
 
+        private const string TimeSeriesFormat = "TimeFormat";
+        private const string UnixTimeFormatMs = "UnixTimeFormatMs";
```
ppekrol: UnixTimeInMs?

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesOperation.cs
```diff
@@ -64,11 +64,27 @@ internal static TimeSeriesOperation Parse(BlittableJsonReaderObject input)
...
+        private const string UnixTimeFormatNs = "UnixTimeFormatNs";
```
ppekrol: UnixTimeInNs?

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesOperation.cs
```diff
@@ -87,9 +103,21 @@ internal static TimeSeriesOperation ParseForBulkInsert(BlittableJsonReaderObject
                     return null; //never hit
                 }
 
+                long time = (long)bjro[0];
```
ppekrol: This cannot be LazyNumberValue?

karmeli87: LazyNumberValue has an implicit cast to `long`

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesOperation.cs
karmeli87: Let's be on the safe side, done.

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesOperation.cs
karmeli87: done

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesOperation.cs
ayende: But that won't trigger here. That is something that the _compiler_ inserts, not runtime.

src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesOperation.cs
```diff
@@ -133,6 +171,16 @@ internal static TimeSeriesOperation ParseForBulkInsert(BlittableJsonReaderObject
             result.Appends = new List<AppendOperation>(sorted.Values);
 
             return result;
+
+            static long GetLong(object value)
```
karmeli87, line 175: Will this box? should we `Inline` this?

ayende: No, this shouldn't cause boxing. Why do you expect it to?

PR #10505, opened 2020-05-03
RavenDB-14365 use docId for time-series opertaions

karmeli87:


src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
```diff
@@ -116,7 +116,7 @@ public Task Stats()
         [RavenAction("/databases/*/timeseries", "GET", AuthorizationStatus.ValidUser)]
         public async Task Read()
         {
-            var documentId = GetStringQueryString("id");
+            var documentId = GetStringQueryString("docId");
             var names = GetStringValuesQueryString("name");
```
grisha-kotler, line 120: Shouldn't it be `"names"`?

ayende: No, because you usually specify just a single one, and add it multiple times

ayende (commented):
Note, we also need to look at the Python client to do fix ups there.

PR #10507, opened 2020-05-03
RavenDB-14729 & RavenDB-14800

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-14729
https://issues.hibernatingrhinos.com/issue/RavenDB-14800

### DIFF
--- a/src/Raven.Server/Documents/Queries/AST/Query.cs
+++ b/src/Raven.Server/Documents/Queries/AST/Query.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text;
 using Raven.Client;
 using Raven.Client.Documents.Queries.TimeSeries;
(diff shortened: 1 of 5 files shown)


src/Raven.Server/Documents/Queries/AST/Query.cs
```diff
@@ -447,20 +448,20 @@ public IEnumerable<object> GetFinalValues()
...
-                        _values[i] = (double)_values[i] / (long)_count[i];
+                        _values[i] = _values[i] / _count[i];
                     }
                     break;
                 default:
                     throw new ArgumentOutOfRangeException("Unknown aggregation operation: " + Aggregation);
             }
 
-            return _values;
+            return _values.Cast<object>();
```
karmeli87, line 464: I think those casting will box.. is it possible to get rid of the `object` and use `long` or `double` instead?

aviv86: we use those aggragated values in `AddTimeSeriesResult` :  https://github.com/ravendb/ravendb/blob/v5.0/src/Raven.Server/Documents/Queries/Results/TimeSeriesRetriever.cs#L828 
we turn them into `DynamicJsonArray`..  and `DynamicJsonArray`  only works with objects

karmeli87 (approved):

ppekrol (approved):
@aviv86 can you check failing tests?

karmeli87:
@ppekrol failing tests are not related to this PR.

PR #10508, opened 2020-05-03
RavenDB-6940 Adding ARM docker image

fbjerggaard:
https://issues.hibernatingrhinos.com/issue/RavenDB-6940

Idea is taken from https://github.com/1Jajen1/ravenDB-docker-arm


gregolsky:
I'll review this later today or tomorrow.

docker/ravendb-ubuntu-arm/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM arm32v7/ubuntu:bionic
```
gregolsky, line 1: We use https://hub.docker.com/_/microsoft-dotnet-core-runtime-deps/ as a base now. There are arm32 and arm 64 bionic images available. I think we should use them instead.

docker/ravendb-ubuntu-arm/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+RUN apt-get update \
+    && apt-get install -y \
+    && apt-get install --no-install-recommends curl bzip2 libunwind8 libcurl4 ca-certificates jq -y
```
gregolsky, line 5: If base image is adjusted then this line needs a tweak as well: https://github.com/ravendb/ravendb/blob/v4.2/docker/ravendb-ubuntu/Dockerfile#L5

docker/ravendb-ubuntu-arm/healthcheck.sh
```diff
@@ -0,0 +1,7 @@
+pgrep Raven.Server > /dev/null
```
gregolsky, line 1: Since it's still Ubuntu I would not duplicate `run-raven.sh` nor `healthcheck.sh`. Maybe we could have 2 or 3 dockerfiles in `ravendb-ubuntu` dir for each arch, a `.dockerignore` file and change `build-ubuntu` and `publish-ubuntu` to have an `-arch` switch and then call `docker build -f <file>...` for selected arch.

docker/ravendb-ubuntu-arm/Dockerfile
gregolsky: Also if we're adding support for arm32 we should probably also think of 64. So instead of `ravendb-ubuntu-arm` we should say `arm32v7` etc.

gregolsky: Umm cross that. We only need `arm32v7`, we do *not* support `arm64`.

fbjerggaard:
@gregolsky Updated the PR so it is now based on the Microsoft provided base-image and now has an `-arch` switch so we avoid duplicate files.

gregolsky:
Thank you @fbjerggaard. I took your work and did some adjustments. 
It's gonna land as ravendb/ravendb#10610 so I'm closing this one.

PR #10510, opened 2020-05-04
RavenDB-14929 - Reduce the number of writes in Output Reduce to Collection

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-14929

### DIFF
--- a/src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
+++ b/src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
@@ -260,10 +260,21 @@ namespace Raven.Server.Documents.Indexes.MapReduce.OutputToCollection
             {
                 var id = GetOutputDocumentKey(reduceKeyHash);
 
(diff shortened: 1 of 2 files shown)


ayende (approved):

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceToCollectionCommand.cs
```diff
@@ -318,6 +329,12 @@ protected override long ExecuteCmd(DocumentsOperationContext context)
             return _reduceDocuments.Count;
         }
 
+        private static unsafe bool DocumentsBinaryEqual(BlittableJsonReaderObject oldDocument, BlittableJsonReaderObject newDocument)
+        {
+            return oldDocument.Size == newDocument.Size &&
```
ppekrol, line 334: Shouldn't we use `oldDocument.Equals(newDocument)` ?

grisha-kotler: If the document size is different, I assume that the contents are different.
Since we build this document, I expect it to be with the same structure if it hasn't changed.

PR #10511, opened 2020-05-04
RavenDB-14931 Fix alignment issue on ARM with doubles 

karmeli87:
Working with unaligned doubles/floats must be done via long/int
(https://github.com/dotnet/runtime/issues/18041#issuecomment-237770456)

### DIFF
--- a/src/Raven.Server/Documents/TimeSeries/StatefulTimeStampValue.cs
+++ b/src/Raven.Server/Documents/TimeSeries/StatefulTimeStampValue.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Diagnostics;
 using System.Runtime.InteropServices;
 
 namespace Raven.Server.Documents.TimeSeries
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Documents/TimeSeries/StatefulTimeStampValue.cs
```diff
@@ -7,18 +8,87 @@ namespace Raven.Server.Documents.TimeSeries
...
         public long PreviousValue; // same as Last
 
         [FieldOffset(16)]
-        public double Max;
+        public long RawMax;
+        public double Max
+        {
+            get
+            {
+                if (RawMax == 0)
```
ayende: Why bother with this?

ayende (commented):
Does this fix the issue?
I would expect that the problem is with `vals`, not here.

src/Raven.Server/Documents/TimeSeries/StatefulTimeStampValue.cs
```diff
@@ -7,18 +8,87 @@ namespace Raven.Server.Documents.TimeSeries
...
-        public double Last;
+        public long RawFirst;
+        public double First
+        {
+            get
+            {
+                if (RawFirst == 0)
+                    return 0;
+
+                return BitConverter.Int64BitsToDouble(RawFirst);
```
ppekrol: Can we apply this only to ARM?

src/Raven.Server/Documents/TimeSeries/StatefulTimeStampValue.cs
karmeli87: Right.. will remove those

karmeli87:
I've tested it, and was able to create a time-series :)

src/Raven.Server/Documents/TimeSeries/StatefulTimeStampValue.cs
ayende: Better yet, can we ensure that we are accessing this aligned? That would be better in general

karmeli87:
For writes we can do tricks to ensure it will be aligned.
But the problem is for reads. You read the segment and you don't know if the pointer that holds those values is 4-bytes aligned or not.
But this code covers both ways.. and not error prone.

src/Raven.Server/Documents/TimeSeries/StatefulTimeStampValue.cs
karmeli87: I'll run a quick benchmark to see if this entails any performance penalty..

ayende:
Hm... good point on reads. Then the only question is what is the perf cost here.

src/Raven.Server/Documents/TimeSeries/StatefulTimeStampValue.cs
karmeli87: Same results.

ppekrol (approved):

ayende (approved):

PR #10517, opened 2020-05-05
RavenDB-14904 use UTC on the client

karmeli87:


src/Sparrow/TimeValue.cs
```diff
@@ -396,5 +396,13 @@ public static DateTime EnsureMilliseconds(this DateTime date)
 
             return date;
         }
+
+        public static DateTime EnsureUtc(this DateTime date)
+        {
+            if (date.Kind == DateTimeKind.Utc)
```
ayende:
```suggestion
            if (date.Kind != DateTimeKind.Local)
```

ayende (commented):
I haven't seen any dealing of queries? Have you touched that?

src/Sparrow/TimeValue.cs
ppekrol: We do not want to touch 'Unspecified' also.

Should this be
```
if (date.Kind == DateTimeKind.Local)
    return date.ToUniversalTime();

return date;
```
?

ayende (approved):

src/Raven.Server/Documents/Queries/Results/TimeSeriesRetriever.cs
```diff
@@ -859,15 +859,12 @@ private static DynamicJsonValue AddTimeSeriesResult(TimeSeriesFunction func, Tim
...
-        private static unsafe DateTime? ParseDateTime(string valueAsStr)
+        private static DateTime? ParseDateTime(string valueAsStr)
         {
-            fixed (char* c = valueAsStr)
-            {
-                var result = LazyStringParser.TryParseDateTime(c, valueAsStr.Length, out var dt, out _);
-                if (result != LazyStringParser.Result.DateTime)
-                    throw new ArgumentException("Unable to parse timeseries from/to values. Got: " + valueAsStr);
-                return dt;
-            }
```
aviv86, line 870: why the swich to `DateTime.TryParse` ? 
`LazyStringParser.TryParseDateTime` should be much more efficient

karmeli87: I don't think we care about efficiency here. 
And this should allow us to parse various formats like `DateTimeOffset` or simple date like `2020-04-04`

ayende: We need to have a closed list of formats that we support here.

karmeli87: Why not to support everything we can? 
Also let's decide on that list here: https://issues.hibernatingrhinos.com/issue/RavenDB-14930

ayende: Right now, you are support local dates.
`Mar 1, 2019` works, for example.
As does this, if you have the right locale settings.

DateTime.Parse("ה אייר תש\"ע")

I'm not happy with either option actually working.

karmeli87: Actually that cool! :)
But I see why you don't like it..

ayende: That is cool, but it depends on too many things.
If the user is running on local machine, it may work, but not on server, etc.

Having a close and well defined list is best

ayende: The worst case, BTW, is this: `1/10/2010` 
This can be either Oct 1 or Jan 10. And `1/10/12` can also be: `0001, Oct 12` in Japan.
Fun

ppekrol (approved):

src/Raven.Server/Documents/Queries/Results/TimeSeriesRetriever.cs
ppekrol: @karmeli87 will handle this in a separate PR.

PR #10522, opened 2020-05-05
RavenDB-14917 exposed time series segment summary values for indexes

ppekrol:


ayende (approved):

src/Raven.Server/Documents/TimeSeries/TimeSeriesValuesSegment.cs
```diff
@@ -82,6 +82,14 @@ public AllocatedMemoryData Clone(JsonOperationContext context, out TimeSeriesVal
             return memory;
         }
 
+        public TimeSeriesSegmentSummary GetSummary()
+        {
+            if (NumberOfEntries == 0)
+                return default;
```
karmeli87, line 88: Just to make sure, the `default` here is `null`s, right?
Since having `0` is not what we want.

test/SlowTests/Issues/RavenDB_14917.cs
```diff
@@ -0,0 +1,93 @@
...
+                    Assert.Equal(new[] { 2, 3.5 }, result.Min);
+                    Assert.Equal(new double[] { 5, 9 }, result.Sum);
+                    Assert.Equal(new [] { 3, 5.5 }, result.Max);
+                    Assert.Equal(2, result.Count);
+                    Assert.Equal(3, result.FirstMax);
+                    Assert.Equal(3.5, result.LastMin);
+                }
+            }
+        }
+
```
karmeli87, line 72: Can we have more tests for that? With empty segments and some deleted values?
And bit more complicated query will be nice :)
Maybe something like that:
```
from segment in ts
where segment.Min[0] > 10
select new {
     segment.Entries.Where(e => e.Values[0] > 10)
}
```

ppekrol: We do not iterate over empty segments and deleted values. We had other tests for checking that IIRC.

src/Raven.Server/Documents/TimeSeries/TimeSeriesValuesSegment.cs
ppekrol: This returns default(TimeSeriesSegmentSummary) which have:
Count = 0,
Max = null,
Min = null,
Sum = null

PR #10526, opened 2020-05-06
RavenDB-14948 Log retention time modified to check last write time to let hours resolution

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-14948

### DIFF
--- a/src/Sparrow/Logging/LoggingSource.cs
+++ b/src/Sparrow/Logging/LoggingSource.cs
@@ -363,25 +363,21 @@ namespace Sparrow.Logging
 
             public static bool TryGetDate(string fileName, out DateTime dateTime)
             {
-                var logPosition = fileName.LastIndexOf(".log", StringComparison.Ordinal);
-                if (logPosition <= 0)
+                try
                 {
-                    dateTime = default;
-                    return false;
+                    if (File.Exists(fileName))
+                    {
+                        dateTime = File.GetLastWriteTime(fileName);
+                        return true;
+                    }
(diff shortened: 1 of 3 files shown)


haludi:
For hours resolution we can't use the name of the file (unless we are adding the hour to the name)
So I used the last write time to apply the retention policy.
So if something will modify the log it will "reset" the retention timer for this file.
Maybe this is even a benefit.

src/Sparrow/Logging/LoggingSource.cs
```diff
@@ -363,25 +363,9 @@ public LogInfo(string fileName)
...
-                if (start < 0)
-                {
-                    dateTime = DateTime.MinValue;
-                    return false;
-                }
-
-                var date = fileName.Substring(start, DateFormatLength);
-                return DateTime.TryParse(date, out dateTime);
+                var fileInfo = new FileInfo(fileName);
+                dateTime = fileInfo.LastWriteTime;
```
ayende: Can this throw if the value does not exists?

ayende (approved):

src/Sparrow/Logging/LoggingSource.cs
grisha-kotler: This will throw an exception if the file doesn't exist

grisha-kotler: nevermind, this will throw for a directory

grisha-kotler (approved):

PR #10528, opened 2020-05-06
RavenDB-14930 Support various date formats for time series query

karmeli87:


src/Raven.Server/Documents/Queries/Results/TimeSeriesRetriever.cs
```diff
@@ -861,10 +862,23 @@ private static DynamicJsonValue AddTimeSeriesResult(TimeSeriesFunction func, Tim
...
-            return date.ToUniversalTime();
+            if (DateTime.TryParseExact(valueAsStr, SupportedDateTimeFormats, CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind, out var date) == false)
+                throw new ArgumentException($"Unable to parse timeseries from/to values. Got: {valueAsStr}{Environment.NewLine}" +
+                                            $"The supported time formats are:{Environment.NewLine}" +
+                                            $"{string.Join(Environment.NewLine, SupportedDateTimeFormats)}");
+            return DateTime.SpecifyKind(date, DateTimeKind.Utc);
         }
+
+        private static readonly string[] SupportedDateTimeFormats =
+        {
```
ayende, line 897: The one thing I wonder is if we should reverse the order. 
I think it tries them in order, so we want the most common ones first.

ayende (approved):

src/Raven.Server/Documents/Queries/Results/TimeSeriesRetriever.cs
karmeli87: So I guess the most common case will be `yyyy-MM-ddTHH:mm:ss` ?

ayende: The .NET one (full with ms)
Then the seconds, then just the date.

ayende (approved):

ppekrol (approved):

PR #10533, opened 2020-05-06
RavenDB-14953 - support importing legacy identities

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-14953


src/Raven.Server/Smuggler/Documents/StreamSource.cs
```diff
@@ -735,6 +735,12 @@ private IEnumerable<(string Prefix, long Value, long Index)> InternalGetIdentiti
                         continue;
                     }
 
+                    if (_buildVersionType == BuildVersionType.V3)
```
ppekrol: Can we move this to DatabaseSmuggler? We modify documents there based on version type.

ppekrol (approved):

PR #10536, opened 2020-05-07
RavenDB-14365 add time-series get overload that accept `null`s

karmeli87:


ppekrol:
@karmeli87 can you rebase?

test/SlowTests/Client/TimeSeries/Operations/TimeSeriesOperations.cs
```diff
@@ -971,7 +971,7 @@ public void ShouldThrowOnNullOrEmptyRanges()
                 var timeSeriesBatch = new TimeSeriesBatchOperation("users/ayende", timeSeriesOp);
 
                 store.Operations.Send(timeSeriesBatch);
-
+                IEnumerable<TimeSeriesRange> ranges = null;
```
ppekrol: Is this used?

src/Raven.Client/Documents/Session/Tokens/TimeSeriesIncludesToken.cs
```diff
@@ -43,12 +44,12 @@ public override void WriteTo(StringBuilder writer)
                 .Append(", ");
 
             writer.Append("'")
-                .Append(_range.From.GetDefaultRavenFormat())
+                .Append((_range.From ?? DateTime.MinValue).GetDefaultRavenFormat())
```
ppekrol: I think we need to support includes like `timeseries(<range_name>)`. Requiring here a date will be a problem for non-clr clients

src/Raven.Client/Documents/Operations/TimeSeries/GetTimeSeriesOperation.cs
```diff
@@ -77,9 +77,10 @@ public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, Serve
                 pathBuilder.Append("&name=")
                     .Append(Uri.EscapeDataString(_name))
                     .Append("&from=")
-                    .Append(_from.EnsureUtc().GetDefaultRavenFormat())
+                    .Append(_from?.EnsureUtc().GetDefaultRavenFormat())
```
ppekrol: it will be better if we will not provide `from` in the URL at all

src/Raven.Client/Documents/Operations/TimeSeries/GetMultipleTimeSeriesOperation.cs
```diff
@@ -85,9 +85,9 @@ public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, Serve
                     pathBuilder.Append("&name=")
                         .Append(Uri.EscapeDataString(range.Name))
                         .Append("&from=")
-                        .Append(range.From.EnsureUtc().GetDefaultRavenFormat())
+                        .Append(range.From?.EnsureUtc().GetDefaultRavenFormat())
```
ppekrol, line 88: it will be better if we will not provide `from` in the URL at all

src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs
```diff
@@ -135,8 +135,8 @@ public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, Serve
                 foreach (var range in _timeSeriesIncludes)
                 {
                     pathBuilder.Append("&timeseries=").Append(Uri.EscapeDataString(range.Name))
-                        .Append("&from=").Append(range.From.GetDefaultRavenFormat())
-                        .Append("&to=").Append(range.To.GetDefaultRavenFormat());
+                        .Append("&from=").Append(range.From?.GetDefaultRavenFormat())
```
ppekrol, line 138: it will be better if we will not provide `from` in the URL at all

test/SlowTests/Client/TimeSeries/Operations/TimeSeriesOperations.cs
karmeli87: nope

src/Raven.Client/Documents/Operations/TimeSeries/GetTimeSeriesOperation.cs
karmeli87: The thing is that we expect pairs of (from, to). 
For getting multiple time-series there might be a problem if the first one will pass only `from` and the second one only `to`..

ppekrol: For ranges, I agree. But in this method?

ayende (approved):

PR #10540, opened 2020-05-08
RavenDB-14964 allocate buffers from the pool and only for the indexing batch

ppekrol:


src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexWriteOperation.cs
```diff
@@ -206,5 +215,24 @@ protected class IndexWriteOperationStats
             public IndexingStatsScope AddStats;
             public IndexingStatsScope SuggestionStats;
         }
+
+        public byte[] GetBuffer(int necessarySize)
```
ayende, line 219: I think there is an issue here, You are returning an array that may be *larger* than the requested size.
But the calling code is expecting just an array, and may consider all of it to be full.

src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/AnonymousLuceneDocumentConverter.cs
```diff
@@ -77,7 +77,7 @@ protected override int GetFields<T>(T instance, LazyStringValue key, object docu
 
             if (_reduceOutput)
             {
-                instance.Add(GetReduceResultValueField(Scope.CreateJson(reduceResult, indexContext)));
+                instance.Add(GetReduceResultValueField(Scope.CreateJson(reduceResult, indexContext), writeBuffer));
```
ayende, line 80: Here we are doing something that may be very bad. We are passing the whole buffer here, but it may be larger than needed.

ayende (changes requested):

src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/AnonymousLuceneDocumentConverter.cs
ppekrol: Which is fine? We did exactly the same thing before. Check `GetReduceResultValueField` we are getting there a buffer, but to Lucene we are passing the 'reduceResult.Size'

src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexWriteOperation.cs
ppekrol: Check my other answer.

src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/AnonymousLuceneDocumentConverter.cs
ayende: No, not fine.
Previously, we created a new instance, which was guaranteed to be zero.
So if we had more data, it was just zero all the time.
Now, you have a shared buffer, which may be dirty, so the data may be *different* on the portions you didn't use.

ppekrol: But since we are copying from the beginning of the buffer till the reduceResult.Size in `GetReduceResult` method and then passing the reduceResult.Size to Lucene, so it should be fine? Previously the buffer was for the whole index lifetime of an index, so after few runs it was 'dirty' too?

ayende: I'm sorry, you are absolutely correct, I missed that part.

ayende (approved):

arekpalinski (commented):
LGTM

PR #10542, opened 2020-05-10
RavenDB-14946 Expose TimeSeriesStatistics as client operation

karmeli87:


src/Raven.Client/Documents/Operations/TimeSeries/GetTimeSeriesStatisticsOperation.cs
```diff
@@ -0,0 +1,54 @@
...
+namespace Raven.Client.Documents.Operations.TimeSeries
+{
+    public class GetTimeSeriesStatisticsOperation : IOperation<TimeSeriesStatistics>
+    {
+        private readonly string _documentId;
+        /// <summary>
+        /// Retrieve start, end and total number of entries for all time-series of a given document
+        /// </summary>
+        /// <param name="documentId"></param>
+        public GetTimeSeriesStatisticsOperation(string documentId)
```
ppekrol, line 17: check for null or empty?

karmeli87: done

PR #10544, opened 2020-05-10
RavenDB-14942 Return empty list when object doesn't have time series

karmeli87:


src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -314,8 +314,8 @@ public List<string> GetTimeSeriesFor<T>(T instance)
             var documentInfo = GetDocumentInfo(instance);
 
             if (documentInfo.Metadata.TryGet(Constants.Documents.Metadata.TimeSeries,
-                    out BlittableJsonReaderArray bjra) == false)
-                return null;
+                out BlittableJsonReaderArray bjra) == false)
+                return Enumerable.Empty<string>().ToList();
```
ayende:
```suggestion
                return new List<string>();
```

ayende (approved):

PR #10554, opened 2020-05-12
RavenDB-14987

ppekrol:
- cleanup allocated string values in context
- decreased the number of allocated string values that we keep

### DIFF
--- a/src/Sparrow/Json/JsonOperationContext.cs
+++ b/src/Sparrow/Json/JsonOperationContext.cs
@@ -57,7 +57,6 @@ namespace Sparrow.Json
         private readonly Stack<MemoryStream> _cachedMemoryStreams = new Stack<MemoryStream>();
 
         private int _numberOfAllocatedStringsValues;
-        private int _allocatedStringValuesCount;
         private readonly FastList<LazyStringValue> _allocateStringValues = new FastList<LazyStringValue>(256);
 
         private DateTime _lastAllocatedStringValueTime;
@@ -103,7 +102,7 @@ namespace Sparrow.Json
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public unsafe LazyStringValue AllocateStringValue(string str, byte* ptr, int size)
         {
-            if (_numberOfAllocatedStringsValues < _allocatedStringValuesCount)
+            if (_numberOfAllocatedStringsValues < _allocateStringValues.Count)
             {
                 var lazyStringValue = _allocateStringValues[_numberOfAllocatedStringsValues++];
                 lazyStringValue.Renew(str, ptr, size);
@@ -115,7 +114,6 @@ namespace Sparrow.Json
             {
                 _allocateStringValues.Add(allocateStringValue);
                 _numberOfAllocatedStringsValues++;
-                _allocatedStringValuesCount++;
                 _lastAllocatedStringValueTime = DateTime.UtcNow;
             }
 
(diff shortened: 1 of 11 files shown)


src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -1021,6 +1028,35 @@ protected internal virtual unsafe void Reset(bool forceReleaseLongLivedAllocator
...
+                    }
+
+                    for (var i = index; i < _allocatedStringValuesCount; i++)
+                    {
+                        var item = _allocateStringValues[i];
+                        if (item == null)
+                            break;
+
+                        _allocateStringValues[i] = null;
+                    }
```
ayende: What are you trying to do here?

src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -57,8 +57,12 @@ public PathCacheHolder(Dictionary<StringSegment, object> path, Dictionary<int, o
         private readonly Stack<MemoryStream> _cachedMemoryStreams = new Stack<MemoryStream>();
 
         private int _numberOfAllocatedStringsValues;
+        private int _allocatedStringValuesCount;
```
ayende: Why keep it as a flag, we already have the count on the list, no?

src/Sparrow/Json/JsonOperationContext.cs
ppekrol: I'm releasing the LSV instances (empty ones) so GC can collect them.

src/Sparrow/Json/JsonOperationContext.cs
ppekrol: I need this, because the underlying list can be 32k, but last non-null LSV can be at earlier index after cleanup.

src/Sparrow/Json/JsonOperationContext.cs
ayende: That is why we have `Array.Clear`, no? Why are you doing this one at a time?
And why are you probing the array?

src/Sparrow/Json/JsonOperationContext.cs
ayende: There is the size of the buffer, and there is the number of valid items there.
All of which should be managed by the list itself, not externally.

src/Sparrow/Json/JsonOperationContext.cs
ppekrol: Done

src/Sparrow/Json/JsonOperationContext.cs
ppekrol: Done

src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -1021,6 +1026,26 @@ protected internal virtual unsafe void Reset(bool forceReleaseLongLivedAllocator
...
+                    _lastAllocatedStringValueTime = now;
+
+                    var index = _allocateStringValues.Count;
+                    for (var i = 0; i < 4; i++)
+                    {
+                        index /= 2;
+                        var item = _allocateStringValues[index];
+                        if (item != null)
+                            break;
+                    }
```
ayende: What is this for? The count is already the size of the list.

ppekrol: You are correct. After introducing Trim I can make it simpler. Cutting by half now.

ayende (approved):

PR #10561, opened 2020-05-12
RavenDB-14990 improve JSON serialization performance

ppekrol:


ayende (approved):

src/Raven.Client/Json/Converters/JsonDictionaryDateTimeKeysConverter.cs
```diff
@@ -121,7 +123,12 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
 
         public override bool CanConvert(Type objectType)
         {
-            if (objectType.GetTypeInfo().IsGenericType == false)
+            return _cache.GetOrAdd(objectType, CanConvertInternal);
```
ayende, line 126: This is expensive?

src/Raven.Client/Json/Converters/JsonEnumerableConverter.cs
```diff
@@ -0,0 +1,73 @@
...
+                return false;
+
+            var jsonArrayContract = _conventions.JsonContractResolver.ResolveContract(objectType) as JsonArrayContract;
+            if (jsonArrayContract == null)
+                return false;
+
+            if (jsonArrayContract.IsMultidimensionalArray)
+                return false;
+
+            return jsonArrayContract.CollectionItemType != typeof(object);
```
ayende, line 47: What is the type is an interface? `List<IFoo>` ?

ayende (approved):

src/Raven.Client/Json/Converters/JsonDictionaryDateTimeKeysConverter.cs
ppekrol: Original implementation took 1600ms in my test. This one takes 400ms.

src/Raven.Client/Json/Converters/JsonEnumerableConverter.cs
ppekrol: Need to double check, but shouldn't be an issue.

ppekrol: Tested. It yields exactly the same JSON as before. Left code in the issue that I've used.

PR #10564, opened 2020-05-13
RavenDB-14631: add compare exchange expiration

garayx:


src/Raven.Server/ServerWide/CompareExchangeExpirationStorage.cs
```diff
@@ -0,0 +1,172 @@
...
+using Sparrow.Json;
+using Voron;
+using Voron.Data.Tables;
+
+namespace Raven.Server.ServerWide
+{
+    public class CompareExchangeExpirationStorage
+    {
+        public static string CompareExchangeByExpiration = "CompareExchangeByExpiration";
+        private static Dictionary<Slice, List<Slice>> _expired;
```
ayende: Make this an instance class, and then place an instance of the cluster state machine.
In this way, you cannot use it on multiple raft engines at the same time, and we need that, for testing, if nothing else.

src/Raven.Server/ServerWide/CompareExchangeExpirationStorage.cs
```diff
@@ -0,0 +1,172 @@
...
+                    items.Delete(reader.Id);
+                }
+
+                return false;
+            }
+        }
+
+        public static IDisposable ExpiredCleaner(ClusterOperationContext context)
+        {
+            _expired = new Dictionary<Slice, List<Slice>>();
```
ayende: This should not be a static field, or a field at all.
It should be something that you pass in the context.

src/Raven.Server/ServerWide/CompareExchangeExpirationStorage.cs
```diff
@@ -0,0 +1,172 @@
...
+                            {
+                                Slice clonedId = multiIt.CurrentKey.Clone(context.Transaction.InnerTransaction.Allocator);
+                                expiredDocs.Add(clonedId);
+
+                                yield return (clonedId, entryTicks);
+                            } while (multiIt.MoveNext());
+                        }
+                    }
+
+                    if (expiredDocs.Count > 0)
```
ayende: Remove the if here, we always wants it.

test/SlowTests/Client/UniqueValues.cs
```diff
@@ -361,5 +368,123 @@ public async Task EnsureCanCompareExchangeWithCorrectCharsEscapeInKey()
...
+                var dateTime = DateTime.Now.AddSeconds(2);
+                var str = "Test";
+                var num = 123.456;
+                var key = "egr/test/cmp/x/change";
+                using (var session = store.OpenAsyncSession(new SessionOptions
+                {
+                    TransactionMode = TransactionMode.ClusterWide
+                }))
+                {
+                    var result = session.Advanced.ClusterTransaction.CreateCompareExchangeValue(key, user);
```
ayende: What happened if I pass a string as the value here?

src/Raven.Client/Documents/Operations/CompareExchange/CompareExchangeValueResultParser.cs
```diff
@@ -56,37 +57,47 @@ public static CompareExchangeValue<T> GetSingleValue(BlittableJsonReaderObject i
...
+            if (raw == null)
+                return new CompareExchangeValue<T>(key, index, default);
+
+            MetadataAsDictionary metadata;
+            if (raw.TryGet(Constants.Documents.Metadata.Key, out BlittableJsonReaderObject bjro) == false || bjro == null)
+                metadata = new MetadataAsDictionary();
+            else
+                metadata = new MetadataAsDictionary(bjro);
+
             if (type.GetTypeInfo().IsPrimitive || type == typeof(string))
```
ayende: Here, you are changing how primitive exchange values are held, right?
How does this work with existing values?

ayende (changes requested):

src/Raven.Client/Documents/Operations/CompareExchange/CompareExchangeValueResultParser.cs
garayx: This is existing code, I just reorganized the method, got the `raw == null` clause out of all if clauses, and added metadata.

This shouldn't have any changes on current behavior of primitive exchange values

test/SlowTests/Client/UniqueValues.cs
garayx: this should work because its converted to blittable
I added a new test for it.

src/Raven.Client/Documents/Operations/CompareExchange/PutCompareExchangeValueOperation.cs
```diff
@@ -47,6 +49,7 @@ public PutCompareExchangeValueCommand(string key, T value, long index, DocumentC
                 _key = key;
                 _value = value;
                 _index = index;
+                _metadata = metadata ?? new MetadataAsDictionary();
```
ppekrol: If it is null maybe do not send it?

src/Raven.Client/Documents/Session/ClusterTransactionSession.cs
```diff
@@ -439,6 +445,30 @@ internal void UpdateValue(CompareExchangeValue<BlittableJsonReaderObject> value,
                         EntityToBlittable.PopulateEntity(_value.Value, value.Value, jsonSerializer);
                 }
             }
+
+            internal static BlittableJsonReaderObject PrepareMetadata(string key, IMetadataDictionary metadataDictionary, JsonOperationContext context)
```
ppekrol: PrepareMetadataForPut?

src/Raven.Server/Config/Categories/ClusterConfiguration.cs
```diff
@@ -106,5 +106,11 @@ public class ClusterConfiguration : ConfigurationCategory
         [DefaultValue(2048)]
         [ConfigurationEntry("Cluster.LogHistoryMaxEntries", ConfigurationEntryScope.ServerWideOnly)]
         public int LogHistoryMaxEntries { get; set; }
+
+        [Description("Time (in seconds) between expired compare exchange cleanup")]
+        [DefaultValue(5)]
+        [TimeUnit(TimeUnit.Seconds)]
+        [ConfigurationEntry("Cluster.ExpiredCompareExchangeCleanupIntervalInSec", ConfigurationEntryScope.ServerWideOnly)]
```
ppekrol: `Cluster.CompareExchangeExpiredDeleteFrequencyInSec` to match the ExpirationConfiguration

src/Raven.Server/Config/Categories/ClusterConfiguration.cs
```diff
@@ -106,5 +106,11 @@ public class ClusterConfiguration : ConfigurationCategory
         [DefaultValue(2048)]
         [ConfigurationEntry("Cluster.LogHistoryMaxEntries", ConfigurationEntryScope.ServerWideOnly)]
         public int LogHistoryMaxEntries { get; set; }
+
+        [Description("Time (in seconds) between expired compare exchange cleanup")]
+        [DefaultValue(5)]
```
ppekrol: IIRC the default for regular expiration is 60. I do not think we should go below that here.

src/Raven.Server/ServerWide/CompareExchangeExpirationStorage.cs
```diff
@@ -0,0 +1,165 @@
...
+
+                expired[ticksSlice] = list;
+
+                if (items.ReadByKey(keySlice, out var reader) == false)
+                    continue;
+
+                var storeValue = reader.Read((int)ClusterStateMachine.CompareExchangeTable.Value, out var size);
+                using var result = new BlittableJsonReaderObject(storeValue, size, context);
+
+                if (HasExpiredMetadata(keySlice.ToString(), result, out long currentTicks) == false)
```
ppekrol: Pass the Slice and materialize it only if needed inside the method?

src/Raven.Server/ServerWide/CompareExchangeExpirationStorage.cs
```diff
@@ -0,0 +1,165 @@
...
+            var expirationTree = context.Transaction.InnerTransaction.ReadTree(CompareExchangeByExpiration);
+            using (var it = expirationTree.Iterate(false))
+            {
+                if (it.Seek(Slices.BeforeAllKeys) == false)
+                    yield break;
+
+                do
+                {
+                    var entryTicks = it.CurrentKey.CreateReader().ReadBigEndianInt64();
+                    if (entryTicks > currentTicks)
```
ppekrol: Shouldn't this be `>=` ?

src/Raven.Server/ServerWide/CompareExchangeExpirationStorage.cs
ppekrol: Aren't you doing double work here? `GetExpiredValues` already returns expired values and you are checking here again if they are expired?

src/Raven.Server/ServerWide/CompareExchangeExpirationStorage.cs
```diff
@@ -0,0 +1,165 @@
...
+                var storeValue = reader.Read((int)ClusterStateMachine.CompareExchangeTable.Value, out var size);
+                using var result = new BlittableJsonReaderObject(storeValue, size, context);
+
+                if (HasExpiredMetadata(keySlice.ToString(), result, out long currentTicks) == false)
+                    continue;
+
+                if (currentTicks > expiredTicks)
+                    continue;
+
+                items.Delete(reader.Id);
```
ppekrol: Please double check if you do not have an iterator of this table opened

test/SlowTests/SlowTests.csproj
```diff
@@ -103,6 +103,7 @@
     <ProjectReference Include="..\..\src\Voron\Voron.csproj" />
     <ProjectReference Include="..\..\tools\Raven.Migrator\Raven.Migrator.csproj" />
     <ProjectReference Include="..\..\tools\Voron.Recovery\Voron.Recovery.csproj" />
+    <ProjectReference Include="..\EmbeddedTests\EmbeddedTests.csproj" />
```
ppekrol: Why is this needed?

ayende:
After you apply the changes requested, please ping me, I would like to have a walkthrough the code with you live.

src/Raven.Server/ServerWide/CompareExchangeExpirationStorage.cs
garayx: This is to handle update of `@expires` value
If the compare exchange `@expires` is updated there will be more than 1 entry for that compare exchange in the `CompareExchangeExpirationStorage`, so I have to check the value of compare exchange  `@expires` in the compare exchange table, then I can know the actual `@expires` value

src/Raven.Server/ServerWide/CompareExchangeExpirationStorage.cs
garayx: no this is right, we should delete if the time is equal

test/SlowTests/SlowTests.csproj
garayx: I reused a ravendb dump creating method from there
https://github.com/ravendb/ravendb/pull/10564/files#diff-d7adabe65d19221ae0f622e52c1b2751R552

src/Raven.Server/ServerWide/CompareExchangeExpirationStorage.cs
garayx: it is only opened for read/delete no iteration

src/Raven.Client/Documents/Operations/CompareExchange/PutCompareExchangeValueOperation.cs
garayx: if we use `PutCompareExchangeValueCommand` without passing metadata, we won't send metadata,
if we use clustertransaction, and don't add metadata, we also won't send the metadata,

ppekrol (commented):
LGTM (just the configuration entry value)

ppekrol (approved):

src/Raven.Client/Documents/Operations/CompareExchange/CompareExchangeValue.cs
```diff
@@ -1,22 +1,29 @@
...
 
-        public CompareExchangeValue(string key, long index, T value)
+        IMetadataDictionary ICompareExchangeValue.Metadata => Metadata;
+
+        public CompareExchangeValue(string key, long index, T value, IMetadataDictionary metadata = null)
         {
             Key = key;
             Index = index;
             Value = value;
+            Metadata = metadata ?? new MetadataAsDictionary();
```
ayende: If use didn't pass it, keep it as null

src/Raven.Client/Documents/Operations/CompareExchange/CompareExchangeValueResultParser.cs
```diff
@@ -56,37 +57,47 @@ public static CompareExchangeValue<T> GetSingleValue(BlittableJsonReaderObject i
 
             var type = typeof(T);
 
+            if (raw == null)
+                return new CompareExchangeValue<T>(key, index, default);
+
+            MetadataAsDictionary metadata;
+            if (raw.TryGet(Constants.Documents.Metadata.Key, out BlittableJsonReaderObject bjro) == false || bjro == null)
+                metadata = new MetadataAsDictionary();
```
ayende: Why are you creating a new instance here when it doesn't exists, keep it null

src/Raven.Client/Documents/Operations/CompareExchange/CompareExchangeValue.cs
```diff
@@ -25,5 +32,6 @@ internal interface ICompareExchangeValue
         public string Key { get; }
         public long Index { get; internal set; }
         public object Value { get; }
+        public IMetadataDictionary Metadata { get; }
```
ayende: Here, when accessed, create a new instance so users have easier time setting this

src/Raven.Client/Documents/Session/ClusterTransactionSession.cs
```diff
@@ -353,9 +356,16 @@ internal ICommandData GetCommand(DocumentConventions conventions, JsonOperationC
 
                         var entity = EntityToBlittable.ConvertToBlittableForCompareExchangeIfNeeded(_value.Value, conventions, context, jsonSerializer, documentInfo: null, removeIdentityProperty: false);
                         var entityJson = entity as BlittableJsonReaderObject;
+                        BlittableJsonReaderObject metadata = null;
+                        if (_value.Metadata.Count != 0)
```
ayende: Here check if metadata is null.
Maybe add another property to avoid creating an instance uneccesarily.

src/Raven.Client/Documents/Operations/CompareExchange/CompareExchangeValue.cs
```diff
@@ -1,22 +1,29 @@
...
         string ICompareExchangeValue.Key => Key;
 
         long ICompareExchangeValue.Index { get => Index; set => Index = value; }
 
         object ICompareExchangeValue.Value => Value;
 
-        public CompareExchangeValue(string key, long index, T value)
+        IMetadataDictionary ICompareExchangeValue.Metadata => Metadata;
+
+        public CompareExchangeValue(string key, long index, T value, IMetadataDictionary metadata = null)
```
karmeli87: This is the client API, right? 
Maybe we want an additional overload with expiration?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -2398,6 +2404,26 @@ internal bool HasCompareExchangeTombstonesWithEtagGreaterThanStartAndLowerThanOr
...
+            {
+                if (cmd.TryGet(nameof(DeleteExpiredCompareExchangeCommand.Ticks), out long ticks) == false)
+                    throw new RachisApplyException($"{nameof(DeleteExpiredCompareExchangeCommand)} must contain a {nameof(DeleteExpiredCompareExchangeCommand.Ticks)} property");
+
+                if (cmd.TryGet(nameof(CleanCompareExchangeTombstonesCommand.Take), out long take) == false)
+                    throw new RachisApplyException($"{nameof(DeleteExpiredCompareExchangeCommand)} must contain a {nameof(DeleteExpiredCompareExchangeCommand.Take)} property");
+
+                var items = context.Transaction.InnerTransaction.OpenTable(CompareExchangeSchema, CompareExchange);
+
+                result = CompareExchangeExpirationStorage.DeleteExpiredCompareExchange(context, items, ticks, take);
```
ayende: Why an out variable and not a return value?

src/Raven.Server/ServerWide/CompareExchangeExpirationStorage.cs
```diff
@@ -0,0 +1,169 @@
...
+                        }
+                    }
+
+                } while (it.MoveNext());
+            }
+        }
+
+        public static bool HasExpiredMetadata(BlittableJsonReaderObject value, out long ticks, Slice keySlice, string storageKey = null)
+        {
+            if (value.TryGetMember(Constants.Documents.Metadata.Key, out var metadata))
```
ayende:
```suggestion
            if (value.TryGetMember(Constants.Documents.Metadata.Key, out var metadata) && metadata != null)
```

ayende: Also, reverse the if

src/Raven.Server/ServerWide/CompareExchangeExpirationStorage.cs
```diff
@@ -0,0 +1,169 @@
...
+                if (expired.TryGetValue(ticksSlice, out List<Slice> list) == false)
+                {
+                    list = new List<Slice> { keySlice };
+                }
+                else
+                {
+                    list.Add(keySlice);
+                }
+
+                expired[ticksSlice] = list;
```
ayende:
```suggestion
                if (expired.TryGetValue(ticksSlice, out List<Slice> list) == false)
                {
                     expired[ticksSlice] = list = new List<Slice>();
                }
                list.Add(keySlice);
```

src/Raven.Server/ServerWide/CompareExchangeExpirationStorage.cs
```diff
@@ -0,0 +1,169 @@
...
+                if (it.Seek(Slices.BeforeAllKeys) == false)
+                    return false;
+
+                var entryTicks = it.CurrentKey.CreateReader().ReadBigEndianInt64();
+                if (entryTicks > currentTicks)
+                    return false;
+            }
+
+            return true;
+        }
```
ayende:
```suggestion

                var entryTicks = it.CurrentKey.CreateReader().ReadBigEndianInt64();
                retyrn entryTicks <= currentTicks;
            }

        }
```

ayende (changes requested):

src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
```diff
@@ -278,6 +280,12 @@ private void LogMessage(string message, Exception e = null, string database = nu
 
                     if (cleanupTombstones && _hasMoreTombstones == false)
                         _lastTombstonesCleanupTimeInTicks = now.Ticks;
+
+                    if (cleanupExpiredCompareExchange)
+                    {
+                        if (await RemoveExpiredCompareExchange(context, now.Ticks) == false)
```
karmeli87: No related to this issue and I know we do it all over the place here, but I wonder if the observer need to `await` for anything.. 
cc @ayende

src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
```diff
@@ -531,6 +539,17 @@ private CompareExchangeTombstonesCleanupState GetMaxCompareExchangeTombstonesEta
             return CompareExchangeTombstonesCleanupState.HasMoreTombstones;
         }
 
+        private async Task<bool> RemoveExpiredCompareExchange(TransactionOperationContext context, long nowTicks)
+        {
+            const int batchSize = 1024;
+            if (CompareExchangeExpirationStorage.HasExpired(context, nowTicks) == false)
+                return false;
+
+            var result = await _server.SendToLeaderAsync(new DeleteExpiredCompareExchangeCommand(nowTicks, batchSize, RaftIdGenerator.DontCareId));
```
karmeli87: If this command will be executed twice, would you end up with the same result as a single execution?
if the answer is no, you need to provide some `guid` to the command.

src/Raven.Client/Documents/Operations/CompareExchange/CompareExchangeValue.cs
ayende: I would rather have the metadata as a generic concept.

src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
ayende: We don't have to wait for it, but we can't create two of them at the same time

src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
ayende: We need a guid, yes

src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
karmeli87: https://issues.hibernatingrhinos.com/issue/RavenDB-15021

garayx:
opened a new pr by accident 
https://github.com/ravendb/ravendb/pull/10586

PR #10565, opened 2020-05-13
RavenDB-14993 - Improve memory utilization when using an encrypted database

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-14993

### DIFF
--- a/src/Sparrow/Json/JsonContextPoolBase.cs
+++ b/src/Sparrow/Json/JsonContextPoolBase.cs
@@ -352,12 +352,11 @@ namespace Sparrow.Json
                 Interlocked.Increment(ref _generation);
             }
 
-            if (_isExtremelyLowMemory.Raise() == false || lowMemorySeverity != LowMemorySeverity.ExtremelyLow)
-            {
-                // - already in extremely low memory
-                // - new low memory severity isn't ExtremelyLow
+            if (lowMemorySeverity != LowMemorySeverity.ExtremelyLow)
                 return;
-            }
+
+            if (_isExtremelyLowMemory.Raise() == false)
+                return;
 
             ClearStack(_globalStack);
 
--- a/src/Sparrow/Json/JsonContextPoolBase.cs
+++ b/src/Sparrow/Json/JsonContextPoolBase.cs
@@ -14,10 +14,12 @@ namespace Sparrow.Json
     {
         private readonly object _locker = new object();
 
-        private long _generation;
         private bool _disposed;
 
         protected SharedMultipleUseFlag LowMemoryFlag = new SharedMultipleUseFlag();
+        private bool _isExtremelyLow;
+        private long _generation;
+
         private readonly CancellationTokenSource _cts = new CancellationTokenSource();
         private readonly long _maxContextSizeToKeepInBytes;
         private readonly long _maxNumberOfContextsToKeepInGlobalStack;
@@ -345,25 +347,37 @@ namespace Sparrow.Json
 
         public void LowMemory(LowMemorySeverity lowMemorySeverity)
         {
-            if (lowMemorySeverity != LowMemorySeverity.ExtremelyLow)
-                return;
-
+            var currentlyExtremelyLow = lowMemorySeverity == LowMemorySeverity.ExtremelyLow;
             if (LowMemoryFlag.Raise() == false)
+            {
+                if (currentlyExtremelyLow && _isExtremelyLow == false)
+                {
+                    _isExtremelyLow = true;
+                    ClearMemory();
+                }
+
                 return;
+            }
 
             Interlocked.Increment(ref _generation);
+            _isExtremelyLow = currentlyExtremelyLow;
+            if (_isExtremelyLow)
+                ClearMemory();
 
-            ClearStack(_globalStack);
-
-            foreach (var coreContext in _perCoreContexts)
+            void ClearMemory()
             {
-                for (int i = 0; i < coreContext.Length; i++)
+                ClearStack(_globalStack);
+
+                foreach (var coreContext in _perCoreContexts)
                 {
-                    var context = coreContext[i];
-                    if (context != null && context.InUse.Raise())
+                    for (int i = 0; i < coreContext.Length; i++)
                     {
-                        context.Dispose();
-                        Interlocked.CompareExchange(ref coreContext[i], null, context);
+                        var context = coreContext[i];
+                        if (context != null && context.InUse.Raise())
+                        {
+                            context.Dispose();
+                            Interlocked.CompareExchange(ref coreContext[i], null, context);
+                        }
                     }
                 }
             }
@@ -384,6 +398,7 @@ namespace Sparrow.Json
         public void LowMemoryOver()
         {
             LowMemoryFlag.Lower();
+            _isExtremelyLow = false;
         }
 
         private sealed class CountingConcurrentStack<TItem>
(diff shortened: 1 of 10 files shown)


src/Voron/Impl/EncryptionBuffersPool.cs
```diff
@@ -50,76 +55,237 @@ public EncryptionBuffersPool()
...
+                Ptr = ptr,
+                Size = size,
+                InPoolSince = DateTime.UtcNow
             });
         }
 
-        public void ReleaseUnmanagedResources()
+        public void LowMemory(LowMemorySeverity lowMemorySeverity)
         {
+            if (lowMemorySeverity != LowMemorySeverity.ExtremelyLow)
```
ayende, line 106: Shouldn't we do that on low memory as well? 
Or at least have some reaction to low memory?

ppekrol (approved):

src/Voron/Impl/EncryptionBuffersPool.cs
```diff
@@ -1,125 +1,290 @@
...
-            if (_items[index].TryPop(out var allocation))
+            while (_items[index].TryPop(out var allocation))
             {
-                thread = allocation.AllocatingThread;
-                return (byte*)allocation.Ptr;
+                if (allocation.InUse.Raise() == false)
+                    continue;
+
+                thread = NativeMemory.ThreadAllocations.Value;
+                thread.Allocations += size;
```
ayende, line 64: This is wrong, I think.
You have _already_ allocated this, so you are double counting it, no?

src/Voron/Impl/EncryptionBuffersPool.cs
```diff
@@ -1,125 +1,290 @@
...
             {
-                // We don't want to pool large buffers / clear them up on low memory
+                // - don't want to pool large buffers
+                // - release all the buffers that were created before we got the low memory event
                 PlatformSpecific.NativeMemory.Free4KbAlignedMemory(ptr, size, allocatingThread);
                 return;
             }
 
+            // updating the thread allocations since we released the memory back to the pool
+            NativeMemory.UpdateMemoryStatsForThread(allocatingThread, size);
```
ayende, line 88: This is an issue about accounting.
From outside, we are always "charged" for this memory.
You are changing things so it will only account for the memory when it is out of the pool. But what about memory that is *in* the pool? That would be hidden from us.

ayende (changes requested):

src/Voron/Impl/EncryptionBuffersPool.cs
grisha-kotler: I'm updating it only for that thread and the total amount of allocated memory remains (only updated when we actually release the memory).

src/Voron/Impl/EncryptionBuffersPool.cs
grisha-kotler: I update the allocations when the buffer is released to the pool.

src/Voron/Impl/EncryptionBuffersPool.cs
grisha-kotler: Maybe update the generation in `LowMemory` and clean everything when `ExtremelyLow`?

ppekrol: I like that, we could do the same for regular pool?

src/Voron/Impl/EncryptionBuffersPool.cs
ayende: The problem here is that you are not accounting allocated memory (from the OS perspective), which will make it harder for us to know that there is pooled memory that isn't in use

src/Voron/Impl/EncryptionBuffersPool.cs
ayende: ditto

src/Voron/Impl/EncryptionBuffersPool.cs
grisha-kotler: The total unmanaged and the total size of the encryption buffer pool will be visible in the stats

src/Voron/Impl/EncryptionBuffersPool.cs
grisha-kotler: done

src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -345,13 +347,17 @@ static void ClearStack(CountingConcurrentStack<T> stack)
...
         {
-            if (lowMemorySeverity != LowMemorySeverity.ExtremelyLow)
-                return;
+            if (LowMemoryFlag.Raise())
+            {
+                Interlocked.Increment(ref _generation);
+            }
 
-            if (LowMemoryFlag.Raise() == false)
+            if (_isExtremelyLowMemory.Raise() == false || lowMemorySeverity != LowMemorySeverity.ExtremelyLow)
```
ayende: You are *always* raising the extremely low memory flag here.

src/Voron/Impl/EncryptionBuffersPool.cs
```diff
@@ -1,125 +1,295 @@
...
 
-        public void ReleaseUnmanagedResources()
+        public void LowMemory(LowMemorySeverity lowMemorySeverity)
         {
+            if (_isLowMemory.Raise())
+            {
+                Interlocked.Increment(ref _generation);
+            }
+
+            if (_isExtremelyLowMemory.Raise() == false || lowMemorySeverity != LowMemorySeverity.ExtremelyLow)
```
ayende: You are always raising the extreme low memory

test/FastTests/Voron/EncryptionBufferPool.cs
```diff
@@ -0,0 +1,186 @@
...
+    public unsafe class EncryptionBufferPool : NoDisposalNeeded
+    {
+        public EncryptionBufferPool(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        [Fact]
+        public void dont_pool_buffers_larger_than_8Mb()
+        {
+            lock (this)
```
ayende: Use a dedicated pool here, not the shared one

test/FastTests/Voron/EncryptionBufferPool.cs
```diff
@@ -0,0 +1,186 @@
...
+                }
+            }
+        }
+
+        [Theory]
+        [InlineData(LowMemorySeverity.Low)]
+        [InlineData(LowMemorySeverity.ExtremelyLow)]
+        public void clear_all_buffers_from_current_generation_on_low_memory(LowMemorySeverity lowMemorySeverity)
+        {
+            lock (this)
```
ayende: Use dedicated pool

ayende (changes requested):

src/Sparrow/Json/JsonContextPoolBase.cs
grisha-kotler: done

src/Voron/Impl/EncryptionBuffersPool.cs
grisha-kotler: done

test/FastTests/Voron/EncryptionBufferPool.cs
grisha-kotler: done

test/FastTests/Voron/EncryptionBufferPool.cs
grisha-kotler: done

ayende (approved):

ppekrol (approved):

PR #10567, opened 2020-05-13
RavenDB-14919

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-14919

### DIFF
--- a/src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs
+++ b/src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs
@@ -144,8 +144,7 @@ namespace Raven.Client.Documents.Commands
         public static void PrepareRequestWithMultipleIds(StringBuilder pathBuilder, HttpRequestMessage request, string[] ids, JsonOperationContext context)
         {
             var uniqueIds = new HashSet<string>(ids);
-            if (uniqueIds.Contains(null))
-                uniqueIds.Remove(null);
+            uniqueIds.Remove(null);
 
             // if it is too big, we drop to POST (note that means that we can't use the HTTP cache any longer)
             // we are fine with that, requests to load > 1024 items are going to be rare
--- a/src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs
+++ b/src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs
@@ -144,14 +144,13 @@ namespace Raven.Client.Documents.Commands
         public static void PrepareRequestWithMultipleIds(StringBuilder pathBuilder, HttpRequestMessage request, string[] ids, JsonOperationContext context)
         {
             var uniqueIds = new HashSet<string>(ids);
-            uniqueIds.Remove(null);
 
             // if it is too big, we drop to POST (note that means that we can't use the HTTP cache any longer)
             // we are fine with that, requests to load > 1024 items are going to be rare
-            var isGet = uniqueIds.Sum(x => x.Length) < 1024;
+            var isGet = uniqueIds.Sum(x => x?.Length ?? 0) < 1024;
             if (isGet)
             {
-                uniqueIds.ApplyIfNotNull(id => pathBuilder.Append("&id=").Append(Uri.EscapeDataString(id)));
+                uniqueIds.ApplyIfNotNull(id => pathBuilder.Append("&id=").Append(Uri.EscapeDataString(id ?? string.Empty)));
             }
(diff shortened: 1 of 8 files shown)


src/Raven.Client/Documents/Commands/GetDocumentsCommand.cs
```diff
@@ -144,6 +144,9 @@ public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, Serve
         public static void PrepareRequestWithMultipleIds(StringBuilder pathBuilder, HttpRequestMessage request, string[] ids, JsonOperationContext context)
         {
             var uniqueIds = new HashSet<string>(ids);
+            if (uniqueIds.Contains(null))
```
ayende:
```suggestion
```
This will work just as well

src/Raven.Client/Documents/Operations/Counters/GetCountersOperation.cs
```diff
@@ -104,6 +104,9 @@ public override void SetResponse(JsonOperationContext context, BlittableJsonRead
             private void PrepareRequestWithMultipleCounters(StringBuilder pathBuilder, HttpRequestMessage request, JsonOperationContext ctx)
             {
                 var uniqueNames = new HashSet<string>(_counters);
+                if (uniqueNames.Contains(null))
```
ayende:
```suggestion
```

src/Raven.Client/Documents/Operations/Counters/GetCountersOperation.cs
```diff
@@ -120,7 +123,7 @@ private void PrepareRequestWithMultipleCounters(StringBuilder pathBuilder, HttpR
                         Operations = new List<CounterOperation>()
                     };
 
-                    foreach (var counter in _counters)
+                    foreach (var counter in uniqueNames)
```
ayende, line 124: Do we have to worry about the order of operations here?

ayende: What happened if the user provided the same counter twice?

ayende: Is the user going to expect index matching on input and output?
What about the `null` here, should there be a match here?

ayende (changes requested):

src/Raven.Client/Documents/Operations/Counters/GetCountersOperation.cs
aviv86: - no, the order stays the same
- if the user provided the same counter more than once, we will drop the duplicates from the request. I don't think it makes much sense to ask for same counter twice... 
- if we have a `null` counter name here, we will get a NRE here: https://github.com/ravendb/ravendb/blob/v4.2/src/Sparrow.Server/ByteString.cs#L1268

Also, because we were using `uniqueNames` in cases of small batches (when `uniqueNames.Sum(x => x.Length) < 1024`), I think we should use the `uniqueNames` in large batches as well

ayende: I was actually asking about what we'll get on the *client* side.
Will we get the duplicate counter value twice as a result?
What happens when we have a null?

Also, I don't think that I saw protection on the server side for this?

aviv86: on the client side (before these changes) - if the user asked for a small batch with duplicates, he will get the counter value once as a result. 
If the user asked for a large batch with a duplicate counter name, he will get the counter value twice in the result. 
if the user had any `null` counter name in his request. it will fail (regardless to batch size)

And you're right, there's not much protection on the server side for this ..

aviv86: Similar problem in `GetDocumentsCommand` - if the user had any `null` ids in his request - it will fail (regardless to batch size). 
And duplicate ids are always removed in this command

ayende: Let's add protection on the server side, and return null for such an instance.
For `GetDocumentsCommand` as well.

And if we match the client side behavior now of `GetDocumentsCommand`, it is good

aviv86: done

ppekrol (approved):

PR #10571, opened 2020-05-14
RavenDB-14999 fixed NRE

ppekrol:


src/Raven.Server/Documents/Operations/Operations.cs
```diff
@@ -189,6 +189,9 @@ void ProgressNotification(IOperationProgress progress)
 
             _active.TryAdd(id, operation);
 
+            if (token == null)
```
grisha-kotler, line 192: what kind of operation doesn't have a token?

ppekrol: TransactionRecorder and generation of certificates

PR #10580, opened 2020-05-19
RavenDB-15022 added abstract layer to JSON serialization

ppekrol:
TODO
- need to handle CompareExchange
- need to look at SubscriptionWorker, we are using there JsonConvert directly
- there are quite a few places where I think we are using wrong method from IBlittableJsonConverter, need to review those


src/Raven.Client/Documents/Conventions/DocumentConventions.cs
```diff
@@ -172,14 +181,8 @@ public DocumentConventions()
 
             MaxNumberOfRequestsPerSession = 30;
 
-            JsonContractResolver = new DefaultRavenContractResolver(this);
-            CustomizeJsonSerializer = serializer => { };
-            CustomizeJsonDeserializer = serializer => { };
-
```
ayende, line 178: Note, breaking changes that we need to talk to the users about

src/Raven.Client/Documents/Conventions/ISerializationConventions.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        void Freeze(DocumentConventions conventions);
+
+        IJsonSerializer CreateSerializer();
+
+        IJsonSerializer CreateDeserializer();
+
+        IBlittableJsonConverter DefaultConverter { get; }
+
+        ISessionBlittableJsonConverter CreateConverter(InMemoryDocumentSessionOperations session);
```
ayende: Why is this in the context of a session?

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.Patch.cs
```diff
@@ -208,14 +208,15 @@ private object AddTypeNameToValueIfNeeded(Type propertyType, object value)
...
             {
-                // the type of the object that's being serialized 
+                // the type of the object that's being serialized
                 // is not the same as its declared type.
                 // so we need to include $type in json
 
-                var serializer = Conventions.CreateSerializer();
-                serializer.TypeNameHandling = TypeNameHandling.Objects;
+                var serializer = Conventions.Serialization.CreateSerializer();
+                var jsonSerializer = (JsonNetJsonSerializer)serializer;
```
ayende: Can we avoid this?

src/Raven.Client/Json/Serialization/JsonNet/Internal/BlittableJsonConverterBase.cs
```diff
@@ -0,0 +1,285 @@
...
+            {
+                throw new InvalidOperationException($"Could not populate entity.", ex);
+            }
+            finally
+            {
+                serializer.ObjectCreationHandling = old;
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
ayende: Remove this, the method is too big

src/Raven.Client/Json/Serialization/JsonNet/Internal/JsonNetBlittableEntitySerializer.cs
```diff
@@ -1,50 +1,46 @@
...
+        private readonly Lazy<GenerateEntityIdOnTheClient> _generateEntityIdOnTheClient;
 
-        private readonly GenerateEntityIdOnTheClient _generateEntityIdOnTheClient;
-
-        public JsonNetBlittableEntitySerializer(DocumentConventions conventions)
+        public JsonNetBlittableEntitySerializer(ISerializationConventions conventions)
         {
-            _generateEntityIdOnTheClient = new GenerateEntityIdOnTheClient(conventions, null);            
-            _deserializer = new LightWeightThreadLocal<JsonSerializer>(conventions.CreateDeserializer);
+            _generateEntityIdOnTheClient = new Lazy<GenerateEntityIdOnTheClient>(() => new GenerateEntityIdOnTheClient(conventions.Conventions, null));
```
ayende: why lazy?

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
ppekrol: Ok, all of the JSON.net conventions were moved to the JsonNetSerializationConventions, but I will summarize this.

ayende: I'm mostly concerned with this being something that users _are_ touching and being a barrier to update to 5.0

So we need a clear guide here.

src/Raven.Client/Documents/Conventions/ISerializationConventions.cs
ppekrol: After looking carefully at EntityToBlittable we had 2 types of conversions, first one was session related only, that was also taking into account events that we bound to session. And this is what is exposed in this converter. The second set of methods were static and was doing similar thing, but without any events or metadata handling.

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.Patch.cs
ppekrol: I will check

src/Raven.Client/Json/Serialization/JsonNet/Internal/BlittableJsonConverterBase.cs
ppekrol: Ok, just copied what we had in previous one.

src/Raven.Client/Json/Serialization/JsonNet/Internal/JsonNetBlittableEntitySerializer.cs
ppekrol: This is a problematic part. The 'DocumentConventions' might be in ctor.  Not sure how to make it convenient for the user.

I would like them to be able to do:

```
Conventions
{
    Serialization = new JsonNetSerializationConventions
    {
        // ...
    }
}
```

and I did that, but this means that I cannot require in ctor of JsonNetSerializationConventions the parent DocumentConventions.

ayende: You can set it when in the `Serialization` setter, no?

ppekrol: Not sure that I follow?

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
ppekrol: I will write a detailed guide in our docs.

src/Raven.Client/Json/Serialization/JsonNet/Internal/JsonNetBlittableEntitySerializer.cs
ayende: User writes:

```
Conventions
{
    Serialization = new JsonNetSerializationConventions
    {
        // ...
    }
}
```

And inside we do:

```
Conventsion.Serialization
{
   set { 
          _serialization = value;
          _serialization.Initiaize(this);
   }
}
```

So here we provide the right value

ppekrol: Ok, will check.

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.Patch.cs
ppekrol: added CreateSerializerOptions here with our own TypeNameHandling. This could be a problem when implementing a different serializer, but not sure if we can avoid this

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
ppekrol: Guide: https://github.com/ravendb/docs/pull/1154

PR #10584, opened 2020-05-19
RavenDB-15014 Avoid capitalizing time series roll up names

kwiato:


src/Raven.Studio/wwwroot/Content/css/pages/time-series-config.less
```diff
@@ -87,6 +87,7 @@
                     border-bottom: 1px solid @hr-border;
                     padding-bottom: @gutter-xs;
                     margin-bottom: @gutter-xs;
+                    text-transform: initial;
```
ml054: why initial, not none?

ml054 (approved):

PR #10587, opened 2020-05-20
[REVIEW ONLY] - TimeSeries ETL

grisha-kotler:


src/Raven.Client/Documents/Operations/ETL/Transformation.cs
```diff
@@ -31,12 +28,164 @@ public class Transformation
...
+
+                    CollectionToLoadCounterBehaviorFunction[collection] = functionName.Value;
+                }
+            }
+        }
+
+        internal readonly TimeSeriesTransformation TimeSeries;
+        internal class TimeSeriesTransformation
+        {
+            internal const string Marker = "timeSeries/";
```
arekpalinski: Maybe `"$timeSeries/"` for consistency with others?

src/Raven.Client/Documents/Operations/ETL/Transformation.cs
```diff
@@ -31,12 +28,164 @@ public class Transformation
...
+            {
+                _parent = parent;
+            }
+            
+            internal void Validate(List<string> errors, EtlType type)
+            {
+                IsAddingTimeSeries = AddTimeSeries.Regex.Matches(_parent.Script).Count > 0;
 
+                FillCollectionToLoadTimeSeriesBehaviorFunction(errors, type);
+            }
```
arekpalinski, line 152: We don't support sending TS to SQL, right? I think we're missing that check in `Validate` - for counters we have:


```
 if (type == EtlType.Sql)
                    {
                        errors.Add("Load counter behavior functions aren't supported by SQL ETL");
                    }

```

src/Raven.Server/Config/Categories/CoreConfiguration.cs
```diff
@@ -14,7 +14,7 @@ namespace Raven.Server.Config.Categories
     public class CoreConfiguration : ConfigurationCategory
     {
         [Description("The URLs which the server should listen to. By default we listen to localhost:8080")]
-        [DefaultValue("http://localhost:8080")]
+        [DefaultValue("http://192.168.1.21:8080")]
```
arekpalinski: Please revert it

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -130,7 +130,7 @@ public class SecurityConfiguration : ConfigurationCategory
         public TimeSetting MasterKeyExecTimeout { get; set; }
 
         [Description("If authentication is disabled, set address range type for which server access is unsecured (None | Local | PrivateNetwork | PublicNetwork).")]
-        [DefaultValue(UnsecuredAccessAddressRange.Local)]
+        [DefaultValue(UnsecuredAccessAddressRange.PrivateNetwork)]
```
arekpalinski: Please revert

src/Raven.Server/Documents/ETL/EtlTransformer.cs
```diff
@@ -164,6 +168,64 @@ private JsValue LoadCounter(JsValue self, JsValue[] args)
...
+            }
+            else
+            {
+                return JsValue.Null;
+            }
+
+            return loadCounterReference;
+        }
+        
+        //TODO Same code as in ScriptRunner
```
arekpalinski: I think we can just make `ScriptRunner.GetDateArg` an internal method and use it here - it's static anyway

src/Raven.Server/Documents/ETL/FilterTimeSeriesEnumerator.cs
```diff
@@ -0,0 +1,94 @@
...
+                var doc = _docsStorage.Get(_context, current.DocId);
+
+                if (doc != null && current.Etag > doc.Etag)
+                {
+                    Current = current;
+                    return true;
+                }
+
+                // counter has lower etag than its document - we skip it to avoid
+                // sending a counter of document that can not exist on the destination side
```
arekpalinski: comments are saying about conters - need to double check if the same logic is fine for time series

src/Raven.Server/Documents/ETL/Providers/SQL/SqlDocumentTransformer.cs
```diff
@@ -145,6 +146,11 @@ protected override void AddLoadedCounter(JsValue reference, string name, long va
             throw new NotSupportedException("Counters aren't supported by SQL ETL");
         }
 
+        protected override void AddLoadedTimeSeries(JsValue reference, string name, IEnumerable<TimeSeriesStorage.Reader.SingleResult> entries)
+        {
+            throw new NotImplementedException();
```
arekpalinski: Rather not supported exception - do we have a plan to support TS ETL to SQL now?

src/Raven.Server/Documents/ETL/Providers/SQL/SqlEtl.cs
```diff
@@ -44,16 +45,20 @@ protected override IEnumerator<ToSqlItem> ConvertCountersEnumerator(IEnumerator<
             throw new NotSupportedException("Counters aren't supported by SQL ETL");
         }
 
-        protected override bool ShouldTrackAttachmentTombstones()
+        protected override IEnumerator<ToSqlItem> ConvertTimeSeriesEnumerator(IEnumerator<TimeSeriesSegmentEntry> timeSeries, string collection)
         {
-            return false;
+            throw new NotImplementedException();
```
arekpalinski: Not supported

src/Raven.Server/Documents/TimeSeries/TimeSeriesSegmentEntry.cs
```diff
@@ -5,6 +5,10 @@ namespace Raven.Server.Documents.TimeSeries
 {
     public class TimeSeriesSegmentEntry
     {
+        public TimeSeriesSegmentEntry()
+        {
+            
+        }
```
arekpalinski: Can we remove this ctor?

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
```diff
@@ -2367,5 +2390,27 @@ private enum DeletedRangeTable
             From = 5,
             To = 6,
         }
+
+        public long GetNumberOfCounterGroupsToProcess(DocumentsOperationContext context, string collection, in long afterEtag, out long totalCount)
```
arekpalinski: Invalid method name - `GetNumberOfTimeSeriesSegmentsToProcess` ?

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
```diff
@@ -344,6 +405,159 @@ private void AddSingleCounterGroup(LazyStringValue docId, BlittableJsonReaderObj
                     _currentRun.AddCounter(docId, prop.Name, value);
             }
         }
+        
+        private void AddTimeSeries(LazyStringValue docId, Dictionary<string, IEnumerable<TimeSeriesStorage.Reader.SingleResult>> timeSeriesReaders)
+        {
+            foreach (var (timeSeriesName, entries) in timeSeriesReaders)
+            {
+                _currentRun.RemoveTimeSeries(docId, timeSeriesName, DateTime.MinValue, DateTime.MaxValue);
```
arekpalinski, line 419: Here we delete _all_ time series - is it intentional?

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
```diff
@@ -387,7 +601,78 @@ private List<CounterOperation> GetCounterOperationsFor(RavenEtlItem item)
...
+            
+            metadata.Modifications.Remove(Constants.Documents.Metadata.TimeSeries);
+            
+            var ret = new Dictionary<string, IEnumerable<TimeSeriesStorage.Reader.SingleResult>>();
+            foreach (LazyStringValue timeSeriesName in timeSeriesNames)
+            {
+                if(ShouldFilterByScriptAndGetParams(item.DocumentId, timeSeriesName, function, out (DateTime @from, DateTime to) toLoad))
+                    continue;
+                if (toLoad.from == default)
+                    toLoad = (DateTime.MinValue, DateTime.MaxValue);
```
arekpalinski: what if I return `{to : new Date(2020, 3, 28)}` - it will default to sending everything anyway

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
```diff
@@ -387,7 +601,78 @@ private List<CounterOperation> GetCounterOperationsFor(RavenEtlItem item)
...
+            var results = new List<TimeSeriesOperation>();
+            
+            foreach (LazyStringValue timeSeriesName in timeSeriesNames)
+            {
+                results.Add(new TimeSeriesOperation
+                {
+                    Name = timeSeriesName,
+                    Removals = new List<TimeSeriesOperation.RemoveOperation>
+                    {
+                        new TimeSeriesOperation.RemoveOperation {From = DateTime.MinValue, To = DateTime.MaxValue}
```
arekpalinski, line 659: You are deleting everything on every call? 

@karmeli87 I think there is a concept of TS tomstrones, right?

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
```diff
@@ -344,6 +405,159 @@ private void AddSingleCounterGroup(LazyStringValue docId, BlittableJsonReaderObj
...
+        
+        private void HandleSingleTimeSeriesSegment(LazyStringValue docId, TimeSeriesSegmentEntry segmentEntry, string function = null)
+        {
+            var timeSeriesEntries = segmentEntry.Segment.YieldAllValues(Context, segmentEntry.Start);
+            
+            if (function != null && FilterSingleTimeSeriesSegmentByScript(ref timeSeriesEntries, docId, segmentEntry, function))
+                return;
+
+            (DateTime from, DateTime to) toRemove = (default, default);
+            foreach (var entry in timeSeriesEntries)
```
arekpalinski, line 435: we are already iterating over `timeSeriesEntries ` in `FilterSingleTimeSeriesSegmentByScript` - can we handle removals of relevant entries as well?

src/Raven.Client/Documents/Operations/ETL/Transformation.cs
```diff
@@ -31,12 +28,164 @@ public class Transformation
...
+
+        internal readonly TimeSeriesTransformation TimeSeries;
+        internal class TimeSeriesTransformation
+        {
+            internal const string Marker = "timeSeries/";
+            
+            internal static class AddTimeSeries
+            {
+                internal const string Name = "addTimeSeries";
+                public const string Signature  = "addTimeSeries(timeSeriesName, from, to)";
```
ayende: Need to review this against the other JS API we have on timeseries, I'm pretty sure we aren't using `TimeSeries`

src/Raven.Client/Documents/Operations/ETL/Transformation.cs
```diff
@@ -31,12 +28,164 @@ public class Transformation
...
+                public const int ParamsCount = 2;
+                internal static readonly Regex Regex = new Regex(Name, RegexOptions.Compiled);
+            }
+            internal static class LoadTimeSeries
+            {
+                public const string Name  = "loadTimeSeries";
+                public const string Signature  = "loadTimeSeries(timeSeriesName, from, to)";
```
ayende, line 124: Same

src/Raven.Client/Documents/Operations/ETL/Transformation.cs
```diff
@@ -31,12 +28,164 @@ public class Transformation
...
+            }
+
+            private void FillCollectionToLoadTimeSeriesBehaviorFunction(List<string> errors, EtlType type)
+            {
+                var counterBehaviors = LoadTimeSeriesOfCollectionBehavior.Regex.Matches(_parent.Script);
+                if (counterBehaviors.Count == 0) 
+                    return;
+            
+                CollectionToLoadTimeSeriesBehaviorFunction = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+                foreach (Match counterBehaviorFunction in counterBehaviors)
```
ayende: Name

src/Raven.Server/Config/Categories/CoreConfiguration.cs
ayende: Remove this

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
ayende: Remove this

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
```diff
@@ -344,6 +405,159 @@ private void AddSingleCounterGroup(LazyStringValue docId, BlittableJsonReaderObj
...
+                }
+                else
+                {
+                    toLoad = (DateTime.MinValue, DateTime.MaxValue);
+                    foreach ((JsValue jsValue, PropertyDescriptor propertyDescriptor) in scriptRunnerResult.Instance.AsObject().GetOwnProperties())
+                    {
+                        var key = jsValue.AsString();
+                        switch (key)
+                        {
+                            case "from":
```
arekpalinski, line 540: I wonder if we should be case sensitive here?

src/Raven.Server/Documents/ETL/EtlProcess.cs
```diff
@@ -154,101 +156,154 @@ protected EtlProcess(Transformation transformation, TConfiguration configuration
...
+            }
+        }
+        
+        private void ExtractTimeSeries(DocumentsOperationContext context,
+            ExtractedItemsEnumerator<TExtracted> merged,
+            long fromEtag,
+            EtlStatsScope stats,
+            DisposableScope scope)
+        {
+            var lastDocEtag = stats.GetLastTransformedOrFilteredEtag(EtlItemType.Document);
```
ayende, line 284: What is we have 10000 time series append? What is the etag that would be recorded?
The document or the time series?

src/Raven.Server/Documents/ETL/EtlTransformer.cs
```diff
@@ -164,6 +168,64 @@ private JsValue LoadCounter(JsValue self, JsValue[] args)
...
+
+            var from = GetDateArg(args[1], signature, "from"); 
+            var to = GetDateArg(args[2], signature, "to"); 
+                
+            JsValue loadCounterReference = (JsValue)Transformation.TimeSeriesTransformation.Marker + timeSeriesName;
+
+            if ((Current.Document.Flags & DocumentFlags.HasTimeSeries) == DocumentFlags.HasTimeSeries)
+            {
+                var reader = Database.DocumentsStorage.TimeSeriesStorage.GetReader(Context, Current.DocumentId, timeSeriesName, from, to);
+                //TODO Maybe to return null if no entries
```
ayende, line 191: You aren't handling tombstones here, are you handling that elsewhere?

src/Raven.Server/Documents/ETL/FilterTimeSeriesEnumerator.cs
```diff
@@ -0,0 +1,94 @@
...
+
+                var doc = _docsStorage.Get(_context, current.DocId);
+
+                if (doc != null && current.Etag > doc.Etag)
+                {
+                    Current = current;
+                    return true;
+                }
+
+                // counter has lower etag than its document - we skip it to avoid
```
ayende: timeseries

src/Raven.Server/Documents/ETL/FilterTimeSeriesEnumerator.cs
```diff
@@ -0,0 +1,94 @@
...
+                _stats.RecordChangeVector(current.ChangeVector);
+                _stats.RecordLastFilteredOutEtag(current.Etag, EtlItemType.TimeSeriesSegment);
+            }
+
+            return false;
+        }
+
+        private static bool CanMoveNextAndNotExceedLastDocumentInBatch(long itemEtag, long lastProcessedDocEtagInBatch)
+        {
+            // we need to ensure we iterate counters only up to last processed doc etag in current batch
```
ayende, line 59: time series

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
```diff
@@ -344,6 +405,159 @@ private void AddSingleCounterGroup(LazyStringValue docId, BlittableJsonReaderObj
                     _currentRun.AddCounter(docId, prop.Name, value);
             }
         }
+        
+        private void AddTimeSeries(LazyStringValue docId, Dictionary<string, IEnumerable<TimeSeriesStorage.Reader.SingleResult>> timeSeriesReaders)
```
ayende, line 415: Right now we are sending the full segment when it changes, one entry at a time.
Need to consider how we can send a whole segment as a unit.
Because otherwise if we have 10 adds, we'll send batches of: 1,2,3,4,5,6,7,8,9
Total of 45.

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
```diff
@@ -344,6 +405,159 @@ private void AddSingleCounterGroup(LazyStringValue docId, BlittableJsonReaderObj
...
+                }
+            }
+            CheckAndAddToRemovals();
+            
+            void CheckAndAddToRemovals()
+            {
+                if (toRemove.@from == default)
+                    return;
+                _currentRun.RemoveTimeSeries(docId, segmentEntry.Name, toRemove.@from, toRemove.to);
+                toRemove.@from = default;
```
ayende: Can avoid `from` ?

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
ayende: huh?

test/SlowTests/Server/Documents/ETL/EtlTimeSeriesTests.cs
```diff
@@ -0,0 +1,1498 @@
...
+        private const int _waitInterval = 1000;
+
+        private abstract class CalculatorTestDataBase : IEnumerable<object[]>
+        {
+            protected const string _script = @"
+loadToUsers(this);
+
+function loadTimeSeriesOfUsersBehavior(doc, counter)
+{
+    return <return>;
```
ayende, line 50: ??

test/SlowTests/Server/Documents/ETL/EtlTimeSeriesTests.cs
```diff
@@ -0,0 +1,1498 @@
...
+}";
+            
+            protected const string _script2 = @"
+var user = loadToUsers(this);
+var timeSeries = loadTimeSeries('Heartrate', new Date(2020, 3, 26), new Date(2020, 3, 28));
```
ayende, line 55: Can I load time series from _another_ document?

src/Raven.Client/Documents/Operations/ETL/Transformation.cs
haludi: As I understand we do want to support ETL time series to SQL.
So no?

ayende: We'll not support it at this point.

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
haludi: It is temporary.
But I need to figure out how to track which time-series I need to remove 
while they are depended on the document and can be loaded entirely, none or part due to changes in the document, right?

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
haludi: https://github.com/ravendb/ravendb/pull/10587#discussion_r429633822

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
haludi: `FilterSingleTimeSeriesSegmentByScript` just define the filtretion by "iterator methods" the actual iteration is only here.

src/Raven.Client/Documents/Operations/ETL/Transformation.cs
haludi: In script runner it is `timeseries(doc, name).get(from, to)`
So I will modify it as follow:
```
var person = loadToPeople({ Name: this.Name + ' ' + this.LastName });
var ts = timeseries(name).get(from, to);
person.timeseries(name).append(ts);
```
Does that make sense?

src/Raven.Client/Documents/Operations/ETL/Transformation.cs
haludi: https://github.com/ravendb/ravendb/pull/10587#discussion_r429640129

src/Raven.Client/Documents/Operations/ETL/Transformation.cs
ayende: How does it relate to counters?

haludi: In the ETL?
We do it that way:
```
var person = loadToPeople({ Name: this.Name + ' ' + this.LastName });

person.addCounter(loadCounter('down'));
```

ayende: So what should the API be in order to have a consistent expereince?

haludi: Consistent with other time-series JS API and ETL counters?
Maybe as follow:
```
var person = loadToPeople({ Name: this.Name + ' ' + this.LastName });
var ts = timeseries(name).load(from, to);
person.addTimeseries(ts);
```

ayende: I don't really like it, but not sure what I would rather have here.

@karmeli87 @aviv86

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
arekpalinski: @karmeli87 how do we handle "TS tombstones" ?

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
arekpalinski: OK, I see

src/Raven.Client/Documents/Operations/ETL/Transformation.cs
arekpalinski: I think we should have consistent API across ETL scripts - it uses 'load' naming convention which we obviously don't use in JS patching. I prefer:

```
var ts = loadTimeSeries(timeSeriesName, from, to)
person.addTimeseries(ts);
```

instead of 

```
var ts = timeseries(name).load(from, to);
person.addTimeseries(ts);
```

because:
- we won't have so many methods that we need to introduce `timeseries(name).` wrapper
- it's consistent with handling counters in ETL scripts (`person.addCounter(loadCounter('down'));`

ayende: I like this

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
haludi: Deleting by tombstones is in progress.
But I referred to a scenario like below:
```
public async Task Test()
{
    const string script = @"
loadToUsers({ Name: this.Name + ' ' + this.LastName });

function loadTimeSeriesOfUsersBehavior(docId, counter)
{
    var user = load(docId);

    if (user.Age < 18)
    {
        return true;
    }
}";
            
    var user = new User {Id = "users/1", Age = 17};
    
    var src = GetDocumentStore(_options);
    var dest = GetDocumentStore();
    
    AddEtl(src, dest, "Users", script);

    using (var session = src.OpenAsyncSession())
    {
        await session.StoreAsync(user);
        session.TimeSeriesFor(user.Id, "Heartrate").Append(DateTime.Now, new []{58.0});
        await session.SaveChangesAsync();
    }

    using (var session = src.OpenAsyncSession())
    {
        session.Advanced.Increment<User, int>(user.Id, x => x.Age, 1);
        await session.SaveChangesAsync();
    }
}
```

src/Raven.Server/Documents/ETL/EtlProcess.cs
haludi: I hope I follow you.
As I understand it we store the general last ETag (per node)
And we track for a batch in memory per type - document, counters, attachment, and now timeseries.
Why did you specify "10000"? Because it includes more than one time-series segment?

src/Raven.Server/Documents/ETL/EtlTransformer.cs
haludi: In progress

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
haludi: We can't know at any point if there will be more appends, right?
So we can't wait for segment to be full.
Maybe a delay by time?
And we can let change the value by configuration or somehow defined it in the script

test/SlowTests/Server/Documents/ETL/EtlTimeSeriesTests.cs
haludi: Just a mark for later to replace by a specific test scenario

test/SlowTests/Server/Documents/ETL/EtlTimeSeriesTests.cs
haludi: For now no.
I mimicked the counters' behavior.
Do we want to support it?

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
arekpalinski: What commands do we send to a destination db currently after modifying a document like in above test?

haludi: Do you ask about commands related to time-series?
I just removed all (from DateTime.MinValue to DateTime.MaxValue) and readd the relevant.
But of course, it is not so efficient.
I did the same test for counters to check how we handle it there but as I understand we don't.
Should we?

arekpalinski: Yes, I meant the commands that ETL sends under the covert to a destination db.

Regarding updates to documents (like `session.Advanced.Increment<User, int>(user.Id, x => x.Age, 1);` in your test) and handling counters then - please see [here](https://issues.hibernatingrhinos.com/issue/RavenDB-11157#focus=streamItem-67-29744.0-0) the general overview - in particular "we send all counters on document update (if not script is defined or the load counter behavior function returns true)"

I don't know if we can do better for Time Series. @karmeli87 ?

arekpalinski: After thinking about it a bit more I think we can differentiate two cases:

- if a document is sent to the same collection (`users/1` from `Users` is send using `loadToUsers()`) then we preserve the document ID - > `users/1` - then if a document is updated we could skip sending `DELETE users/1` so we must not send _all_ time series

- if a document is sent to different collection then we don't know its ID on destination side so we always sent `DELETE docs where prefix ID is users/1`, then I think we have to send all time series as well

Thoughts?

ayende: That make sense, but that would be _horrible_ for large systems, no?
It might be better to disable that entirely, if we are sending to a different collection.

arekpalinski: Oren, we're talking here about the setup that someone _enabled_ it on purpose - by setting up a behavior function like:

```
function loadTimeSeriesOfUsersBehavior(docId, counter)
{
    return true; // not filtering - send every TS entry
 }
```
That won't be our default behavior. So I think it's fine - we just need to make people aware of it in the documentation.

ayende: Okay, then

haludi:
https://github.com/ravendb/ravendb/pull/10659

PR #10592, opened 2020-05-21
RavenDB-14974 add typed rolluped time-series

karmeli87:


src/Raven.Client/Documents/Session/TimeSeries/TimeSeriesEntry.cs
```diff
@@ -22,19 +24,70 @@ public class TimeSeriesEntry : TimeSeriesEntryValues
...
+                Values = Values,
+                IsRollup = true
+            };
         }
     }
 
-    public abstract class TimeSeriesAggregatedEntry : TimeSeriesEntryValues
+    public static class RollupExtensions
     {
+        public static TimeSeriesRollupEntry<T> AsRollUpEntry<T>(this TimeSeriesEntry entry) where T : TimeSeriesEntry, new()
```
ppekrol: Should be `AsRollupEntry`? Can you check if we have a proper casing in the solution? Especially public API?

src/Raven.Client/Documents/TimeSeries/TimeSeriesOperations.cs
```diff
@@ -46,6 +46,16 @@ private TimeSeriesOperations(IDocumentStore store, string database)
             return RegisterAsync(collection, name, mapping.Values.Select(f => f.Name).ToArray());
         }
         
+        /// <summary>
+        /// Register value names of a time-series
+        /// </summary>
+        /// <typeparam name="TCollection">Collection type</typeparam>
+        public Task RegisterAsync<TCollection>(string name, string[] valueNames)
```
ppekrol, line 53: Remember to add `string name = null` to other overloads.

ppekrol (approved):

src/Raven.Client/Documents/Session/TimeSeries/TimeSeriesEntry.cs
ayende: Why do we have both `ToRollupEntry` and this one?

karmeli87: changed `AsRollUpEntry`=> `AsRollupEntry`
forgot to remove `ToRollupEntry`.

src/Raven.Client/Documents/TimeSeries/TimeSeriesOperations.cs
karmeli87: done

src/Raven.Client/Documents/Session/TimeSeries/TimeSeriesEntry.cs
```diff
@@ -122,4 +182,230 @@ void IPostJsonDeserialization.PostDeserialization()
...
+                }
+            }
+
+            _innerArrayInitialized = true;
+        }
+
+        private double[][] _innerValues;
+
+        [OnDeserialized]
+        internal new void OnNewtonSoftJsonDeserialized(StreamingContext context)
```
ppekrol: @karmeli87 I just realized that after https://github.com/ravendb/ravendb/pull/10580 is merged I would like to avoid this kind of an approach. I think it would be better if you would write converter (supporting reads only).

ayende (approved):

PR #10596, opened 2020-05-21
RavenDB-7255 Alert 3 weeks before certificate expires

iftahbe:


src/Raven.Server/RavenServer.cs
```diff
@@ -843,6 +876,7 @@ private bool CertificateCallback(object sender, X509Certificate cert, X509Chain
         public void RefreshClusterCertificateTimerCallback(object state)
         {
             RefreshClusterCertificate(state, RaftIdGenerator.NewId());
+            UpdateCertificateExpirationAlert();
```
iftahbe: This will be a "sticky" alert. The user can dismiss it but it will pop up again in one hour. Also, if there are only 3 days left the alert becomes red.

src/Raven.Server/RavenServer.cs
```diff
@@ -344,6 +345,38 @@ void ConfigureKestrel(KestrelServerOptions options)
             }
         }
 
+        private void UpdateCertificateExpirationAlert(object sender = null, EventArgs args = null)
+        {
+            var remainingDays = (Certificate.Certificate.NotAfter - Time.GetUtcNow().ToLocalTime()).TotalDays;
+            if (remainingDays <= 20)
```
ppekrol: This can be negative? We want to set it to 0 then?

iftahbe: Yes, I will display a relevant message if it is already expired.

src/Raven.Server/RavenServer.cs
```diff
@@ -344,6 +345,52 @@ void ConfigureKestrel(KestrelServerOptions options)
             }
         }
 
+        private void UpdateCertificateExpirationAlert(object sender = null, EventArgs args = null)
```
ppekrol: @iftahbe maybe let's just invoke this inside `OnServerCertificateChanged` ? Instead of having two separate callbacks we will have one? You do not need `sender` and `args` then also and you do not have to deal with removing the event because we are handling that already for `OnServerCertificateChanged`.

ppekrol: Probably worth wrapping also everything in try catch and logging the error.

ppekrol (approved):

PR #10600, opened 2020-05-23
RavenDB-13921 changes to work with latest Jint

lahma:
I'm creating this PR to show changes required for the soon-to-be-ready Jint version with let/const support. Highlights:

* Esprima v2 (beta) is being used be Jint
   * Removed interfaces, concrete classes now (except for IFunction)
   * API is being refined
   * ParseProgram was obsoleted in v1, now should use either PraseScript or ParseModule (always script for Ravendb)
   * HoistingScope was removed, Jint now does the hoisting 

I try to keep this updated for time being in hope it makes things easier for you. I'm not expecting this to be merged. This will not build as there's no such Jint package that would have this changes yet (your fork).


ayende:
Thanks for this work. What are the new things in Jint that we can expect?

We are going to release v5.0 soon, and I want to know if we should do this merge before or after.

lahma:
I think the biggest thing since your latest catch-ups (rebases) is getting let/const support ( https://issues.hibernatingrhinos.com/issue/RavenDB-13541 ), this landed yesterday. So far it's been flaky to say the least and worked "mostly OK". Usually it was considered as a var and thus might have been working as expected depending on what kind of shadowing was (not) happening.

I think one thing you could also consider is to translate your projection code to utilize lambdas. `docs.map(function (d) { return d.name; })` => `docs.map(x => x.name)`.  You already have the version rebased that is required I believe (v5 branch). @ppekrol probably knows best.

I'm unsure if it's breaking to change to lambdas later. It runs faster with less context required.

Another note is that you have your own AST visitor, Esprima also has base class for that, don't know if you need to keep the duplicated version.

ppekrol:
@lahma I've compiled the NuGet (with small adjustments around MaxStatements) and made more changes to the code. Let's see how it goes first, then we can talk about doing adjustments.

Would be happy if you could review the changes made.

BTW, the `GetKey` extension and Property is a bit confusing to me.

ppekrol:
The Jint changes that I've done are here: https://github.com/ravendb/jint/tree/RavenDB-13921

lahma:
Looks good, it seems that the new constraint refactor allowed you to minimize custom code quite well. Maybe you could PR those changes to Jint with test case so it would be there without your custom modifications?

The GetKey is confusing to me too 😄 I think it would be ideal to have it internal and if you just take the current implementation to RavenDB's side and possibly tweak if needed. There was a problem when infinite recursion (stack overflow) was possible when key was tried to resolve in exception building. So must be careful when using engine to resolve the key - I don't know if your use cases need other than the easy ones (literal, identifier).

ppekrol:
Opened a separate issue for Lambdas: https://issues.hibernatingrhinos.com/issue/RavenDB-15064

PR #10605, opened 2020-05-24
RavenDB-14978 Add load balance behavior for writes. This over ride re…

ayende:
…ad balance behavior.

Fixed an issue with XXHash32 on string not taking into account the last char.

### DIFF
--- a/src/Raven.Client/Documents/Conventions/DocumentConventions.cs
+++ b/src/Raven.Client/Documents/Conventions/DocumentConventions.cs
@@ -238,7 +238,7 @@ namespace Raven.Client.Documents.Conventions
         private TimeSpan _firstBroadcastAttemptTimeout;
 
         private int _writeBalanceSeed;
-        private WriteBalanceBehavior _writeBalanceBehavior;
+        private LoadBalanceBehavior _loadBalanceBehavior;
         private ReadBalanceBehavior _readBalanceBehavior;
         private bool _preserveDocumentPropertiesNotFoundOnModel;
         private Size _maxHttpCacheSize;
@@ -383,15 +383,15 @@ namespace Raven.Client.Documents.Conventions
             }
         }
 
-        public WriteBalanceBehavior WriteBalanceBehavior
+        public LoadBalanceBehavior LoadBalanceBehavior
         {
             // We have to make this check so if admin activated this, but client code did not provide the selector,
             // it is still disabled. Relevant if we have multiple clients / versions at once.
-            get => _writeBalanceSessionContextSelector == null ? WriteBalanceBehavior.None : _writeBalanceBehavior;
+            get => _writeBalanceSessionContextSelector == null ? LoadBalanceBehavior.None : _loadBalanceBehavior;
             set
             {
                 AssertNotFrozen();
-                _writeBalanceBehavior = value;
+                _loadBalanceBehavior = value;
             }
         }
         /// <summary>
@@ -993,7 +993,7 @@ namespace Raven.Client.Documents.Conventions
                     _maxNumberOfRequestsPerSession = _originalConfiguration.MaxNumberOfRequestsPerSession ?? _maxNumberOfRequestsPerSession;
                     _readBalanceBehavior = _originalConfiguration.ReadBalanceBehavior ?? _readBalanceBehavior;
                     _identityPartsSeparator = _originalConfiguration.IdentityPartsSeparator ?? _identityPartsSeparator;
-                    _writeBalanceBehavior = _originalConfiguration.WriteBalanceBehavior ?? _writeBalanceBehavior;
+                    _loadBalanceBehavior = _originalConfiguration.WriteBalanceBehavior ?? _loadBalanceBehavior;
 
                     _originalConfiguration = null;
                     return;
@@ -1006,12 +1006,12 @@ namespace Raven.Client.Documents.Conventions
                         MaxNumberOfRequestsPerSession = MaxNumberOfRequestsPerSession,
                         ReadBalanceBehavior = ReadBalanceBehavior,
                         IdentityPartsSeparator = IdentityPartsSeparator,
-                        WriteBalanceBehavior = _writeBalanceBehavior,
+                        WriteBalanceBehavior = _loadBalanceBehavior,
                     };
 
                 _maxNumberOfRequestsPerSession = configuration.MaxNumberOfRequestsPerSession ?? _originalConfiguration.MaxNumberOfRequestsPerSession ?? _maxNumberOfRequestsPerSession;
                 _readBalanceBehavior = configuration.ReadBalanceBehavior ?? _originalConfiguration.ReadBalanceBehavior ?? _readBalanceBehavior;
-                _writeBalanceBehavior = configuration.WriteBalanceBehavior ?? _originalConfiguration.WriteBalanceBehavior ?? _writeBalanceBehavior;
+                _loadBalanceBehavior = configuration.WriteBalanceBehavior ?? _originalConfiguration.WriteBalanceBehavior ?? _loadBalanceBehavior;
                 _identityPartsSeparator = configuration.IdentityPartsSeparator ?? _originalConfiguration.IdentityPartsSeparator ?? _identityPartsSeparator;
             }
         }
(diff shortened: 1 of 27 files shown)


karmeli87 (approved):

karmeli87:
Do we also want to add a read configuration `ReadBehaviour.SameAsWrite` ?
So if context is defined read&writes will go to the same node.

ayende:
That is how it works now. If you defined writes, it will always use that one

karmeli87:
Right, the `switch (Conventions.WriteBalanceBehavior)` is before any other check, so it force reads and writes to the same node/topology.

Not sure if we need an ability to fine tune it.

ayende:
I can't think of a reasonable scenario for wanting to do both

src/Raven.Client/Documents/Session/IAdvancedDocumentSessionOperations.cs
```diff
@@ -33,6 +33,8 @@ public interface IAdvancedDocumentSessionOperations
         RequestExecutor RequestExecutor { get; }
         JsonOperationContext Context { get; }
 
+        SessionInfo SessionInfo { get; }
```
ppekrol, line 37: Do we need to expose this?

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
```diff
@@ -378,6 +381,27 @@ public ReadBalanceBehavior ReadBalanceBehavior
...
+        {
+            get => _writeBalanceSeed;
+            set
+            {
+                AssertNotFrozen();
+                _writeBalanceSeed = value;
+            }
+        }
+
+        public WriteBalanceBehavior WriteBalanceBehavior
```
ppekrol: From implementation it more looks like `LoadBalanceBehavior` than `WriteBalanceBehavior`. I think we should switch the name.

test/SlowTests/Issues/RavenDB-14978.cs
```diff
@@ -0,0 +1,93 @@
...
+                var sessionInfo = s3.Advanced.SessionInfo;
+                s3Ctx = sessionInfo.SessionId.Value;
+            }
+            
+            Assert.NotEqual(s2Ctx, s3Ctx);
+
+
+        }
+
+        public class User
```
ppekrol: internal

src/Raven.Client/Http/WriteBalanceBehavior.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Raven.Client.Http
+{
+    public enum WriteBalanceBehavior
+    {
+        None,
+        ClientContextSelection
```
ppekrol: I do not like this name, but I cannot think of a better one

test/SlowTests/Issues/RavenDB-14978.cs
```diff
@@ -0,0 +1,93 @@
...
+            int s2Ctx = -1;
+            using (var s2 = store.OpenSession())
+            {
+                var sessionInfo = s2.Advanced.SessionInfo;
+                s2Ctx = sessionInfo.SessionId.Value;
+            }
+            
+            Assert.Equal(s2Ctx, s1Ctx);
+
+            context = "users/2";
```
ppekrol, line 74: I think we are missing a convenient way for a user to set a context for a session, which should be either used directly then or passed to a `WriteBalanceSessionContextSelector` method.

E.g.
```
session.Advanced.SetContext("some value");
```

src/Raven.Client/Documents/Session/IAdvancedDocumentSessionOperations.cs
ayende: Yes, this is commonly needed when we want to properly use `ExecuteAsync` for custom commands.

src/Raven.Client/Http/WriteBalanceBehavior.cs
ayende: Updated it, let's see if it would work better.

test/SlowTests/Issues/RavenDB-14978.cs
ayende: `s3.Advanced.SessionInfo.SetSessionContext("foo");`

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
```diff
@@ -378,6 +381,27 @@ public ReadBalanceBehavior ReadBalanceBehavior
             }
         }
 
+        public int WriteBalanceSeed
```
ppekrol: We need to rename this also, right?

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
```diff
@@ -477,6 +501,21 @@ public bool UseOptimisticConcurrency
             }
         }
 
+        /// <summary>
+        /// Gets or set the function that allow to specialize the topology
+        /// selection for a particular session. Used in load balancing
+        /// scenarios.
+        /// </summary>
+        public Func<string, string> WriteBalanceSessionContextSelector
```
ppekrol: We need to rename this also, right?

src/Raven.Client/Documents/Operations/Configuration/ClientConfiguration.cs
```diff
@@ -15,6 +15,8 @@ public class ClientConfiguration
         public int? MaxNumberOfRequestsPerSession { get; set; }
 
         public ReadBalanceBehavior? ReadBalanceBehavior { get; set; }
+        
+        public LoadBalanceBehavior? WriteBalanceBehavior { get; set; }
```
ppekrol: We need to rename this also, right?

src/Raven.Client/Documents/Session/SessionInfo.cs
```diff
@@ -1,21 +1,73 @@
...
+        public int SessionId { get;}
 
         public long? LastClusterTransactionIndex { get; set; }
 
         public bool AsyncCommandRunning { get; set; }
 
         public bool NoCaching { get; set; }
 
-        public SessionInfo(int sessionId, bool asyncCommandRunning, long? lastClusterTransactionIndex = null, bool noCaching = false)
+        public SessionInfo(string sessionContextKey, int writeBalanceSeed, bool asyncCommandRunning, long? lastClusterTransactionIndex = null, bool noCaching = false)
```
ppekrol: We need to rename this also, right?

src/Raven.Client/Documents/Session/SessionInfo.cs
```diff
@@ -1,21 +1,73 @@
...
+                _clientSessionId = ++_clientSessionIdCounter;
+            }
+            else
+            {
+                _clientSessionId = (int)Hashing.XXHash32.Calculate(sessionKey,
+                    (uint)writeBalanceSeed);
+            }
+        }
+
+        public async Task<ServerNode> GetCurrentSessionNode(RequestExecutor requestExecutor)
```
ppekrol: internal?

src/Raven.Client/Documents/Session/SessionInfo.cs
```diff
@@ -1,21 +1,73 @@
...
         {
+            SetSessionContext(sessionContextKey, writeBalanceSeed);
+
             LastClusterTransactionIndex = lastClusterTransactionIndex;
-            SessionId = sessionId;
             AsyncCommandRunning = asyncCommandRunning;
             NoCaching = noCaching;
         }
+
+        public void SetSessionContext(string sessionKey, int writeBalanceSeed = 0)
```
ppekrol: The publicly available method should be IMO:
```
public void SetContext(string sessionKey)
```

and should check if sessionKey is not null or whitespace and use the original `loadBalanceSeed` from the conventions (the one passed via ctor).

ppekrol: Can we have a test for this?

ppekrol: I think we should also prevent user from setting this more than once in a lifecycle of a session. I do not see a valid scenario for this, it can only cause confusion and we should be safe by default.

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
ayende: yes

ppekrol:
@ayende I've pushed small changes (please review) and rebased.

ppekrol (approved):

ayende:
LGTM

PR #10608, opened 2020-05-24
RavenDB-15047 .NET Core lose type information when using ToArray with…

ayende:
… group by


ppekrol:
@ayende can you check `Raven.Tests.MailingList.Matthias.TransformerFails` test?

PR #10609, opened 2020-05-24
RavenDB-14965 Time-Series indexing progress - showing items/sec or do…

Danielle9897:
…cs/sec

### DIFF
--- a/src/Raven.Client/Documents/Indexes/IndexProgress.cs
+++ b/src/Raven.Client/Documents/Indexes/IndexProgress.cs
@@ -8,6 +8,8 @@ namespace Raven.Client.Documents.Indexes
 
         public IndexType Type { get; set; }
 
+        public IndexSourceType SourceType { get; set; }
+
         public Dictionary<string, CollectionStats> Collections { get; set; }
 
         public bool IsStale { get; set; }
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -2205,6 +2205,7 @@ namespace Raven.Server.Documents.Indexes
                     {
                         Name = Name,
                         Type = Type,
+                        SourceType = SourceType,
                         IndexRunningStatus = Status,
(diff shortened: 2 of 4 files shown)


src/Raven.Client/Documents/Indexes/IndexProgress.cs
```diff
@@ -8,6 +8,8 @@ public class IndexProgress
 
         public IndexType Type { get; set; }
 
+        public IndexSourceType SourceType { get; set; }
```
ml054, line 11: like I said I think we don't want to expose that as this is constant property - and should be part of index output not progress. Even if we have race and index type isn't send we simply don't show such item as we don't know about this index. we can have a look at details tomorrow.

ml054 (changes requested):

src/Raven.Client/Documents/Indexes/IndexProgress.cs
ayende: What about multi map indexes that have both time series / counters / docs?

ml054: wow, I wasn't even aware we can mix index sources. Then I assume we have to introduce new type of index source: Mixed?

ppekrol: At this point we do not support mixing the sources of maps. I will create an issue to throw when something like that happens, till we want to actually support that.

ayende: We do need to take this into account, because this is very likely going to _be_ a feature.

arekpalinski: I don't see the problem with adding `SourceType ` here, we already have `public IndexType Type { get; set; }`

arekpalinski: For now let's stay with current implementation here - introducing the mixed sources will require much more changes in the studio anyway

ml054 (approved):

PR #10622, opened 2020-05-26
RavenDB-15029 : session Raw Query should not track TimeSeries query result as document

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-15029

### DIFF
--- a/src/Raven.Server/Documents/Queries/Results/TimeSeriesRetriever.cs
+++ b/src/Raven.Server/Documents/Queries/Results/TimeSeriesRetriever.cs
@@ -724,8 +724,8 @@ namespace Raven.Server.Documents.Queries.Results
         {
             var result = new DynamicJsonValue
             {
-                ["Count"] = count, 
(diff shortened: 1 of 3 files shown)


ayende (approved):

ppekrol (approved):

src/Raven.Server/Documents/Queries/Results/TimeSeriesRetriever.cs
```diff
@@ -202,11 +199,7 @@ BlittableJsonReaderObject GetAggregatedValues()
 
                 _argumentValuesDictionary?.Clear();
 
-                return _context.ReadObject(new DynamicJsonValue
-                {
-                    ["Count"] = count,
```
karmeli87, line 207: Can we use `nameof` for the "Count" and "Results"?

aviv86: done

PR #10627, opened 2020-05-27
RavenDB-15013 TimeSeries Get : return results from Rolled-up series as well

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-15013


src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
```diff
@@ -252,7 +252,8 @@ internal static TimeSeriesRangeResult GetTimeSeriesRange(DocumentsOperationConte
         private static unsafe TimeSeriesRangeResult GetTimeSeriesRange(DocumentsOperationContext context, string docId, string name, DateTime from, DateTime to, ref int start, ref int pageSize)
         {
             List<TimeSeriesEntry> values = new List<TimeSeriesEntry>();
-            var reader = context.DocumentDatabase.DocumentsStorage.TimeSeriesStorage.GetReader(context, docId, name, from, to);
+            var collection = ExecuteTimeSeriesBatchCommand.GetDocumentCollection(context.DocumentDatabase, context, docId, false);
```
ppekrol: Try in the future when you have direct `null` or `bool` values to use named parameters. It is easier when reading the code to understand what that value is for.

ppekrol (approved):

PR #10628, opened 2020-05-27
RavenDB-15034 Use Security.Certificate.Load.Exec instead of Security.Certificate.Exec in RavenServerRunner.Run 

iftahbe:


src/Raven.Server/RavenServer.cs
```diff
@@ -1349,19 +1349,32 @@ private string GetWebUrl(string kestrelUrl)
...
-        private CertificateHolder LoadCertificate()
+        private CertificateHolder LoadCertificateAtStartup()
         {
             try
             {
-                if (string.IsNullOrEmpty(Configuration.Security.CertificateExec) == false)
+                if (string.IsNullOrEmpty(Configuration.Security.CertificateLoadExec) == false &&
+                    (string.IsNullOrEmpty(Configuration.Security.CertificateRenewExec) || string.IsNullOrEmpty(Configuration.Security.CertificateChangeExec)))
                 {
-                    throw new InvalidOperationException($"Invalid certificate configuration. The configuration property '{RavenConfiguration.GetKey(x => x.Security.CertificateExec)}' has been deprecated since RavenDB 4.2, please use '{RavenConfiguration.GetKey(x => x.Security.CertificateLoadExec)}' along with '{RavenConfiguration.GetKey(x => x.Security.CertificateRenewExec)}' and '{RavenConfiguration.GetKey(x => x.Security.CertificateChangeExec)}'. For more information, refer to the online documentation at https://ravendb.net/l/4554RZ/4.2.");
```
iftahbe, line 1358: I'm 99% sure we don't need this restriction since we check that renew exec is not null before actually starting the renew process.

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -172,8 +172,7 @@ public class SecurityConfiguration : ConfigurationCategory
 
         public bool IsCertificateConfigured =>
             string.IsNullOrWhiteSpace(CertificatePath) == false ||
-            string.IsNullOrWhiteSpace(CertificateLoadExec) == false ||
-            string.IsNullOrWhiteSpace(CertificateExec) == false;
```
ppekrol, line 176: What about backward compatibility?

ppekrol (commented):
@iftahbe are we taking into account backward compatibility? At the beginning we had only CertificateExec, which was moved to CertificateLoadExec IIRC. We still need to support both.

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
iftahbe: There is no backward compatibility in regards to this configuration property.
I don't remember why but we gave it up when we did this:
https://github.com/ravendb/ravendb/blob/v4.2/src/Raven.Server/RavenServer.cs#L1356

It's funny and a little ironic because that commit's title is "backward compatibility" :) 
https://github.com/ravendb/ravendb/pull/9041/commits/09dcccec1999fd077bf6181e3411152d30c4b1c7

PR #10631, opened 2020-05-27
RavenDB-15062 - Filtered Replication [ Review Only ]

ayende:


src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
```diff
@@ -200,7 +200,7 @@ private void Replication()
                         if (supportedFeatures.Replication.PullReplication)
                         {
                             SendPreliminaryData();
-                            if (Destination is PullReplicationAsSink)
+                            if (Destination is PullReplicationAsSink sink && sink.Mode == ReplicationMode.Pull)
```
ayende: Note, here we don't use flags, only the real value.

ayende (commented):
This adds the capability to add filters to replication hub as well as support pull & push on the hub.
The idea is that the hub contains a list of certificates as well as the paths that each of them is allowed to write / read.
Both incoming & outgoing replication will filter data (documents, counters, time series, attachments and revisions) so only documents that match the
allowed paths will be replicated.
Note that within a document, all the data about the document is either replicated or not. There is no way to filter data inside a document.

This is a *security* feature that assume that the other side is *not* trusted. We should think about ways that people can try to mess with us here.

When using pull filtered replication, the hub will filter out all the paths that the user is not allowed to see at the source.
When using push filtered replication, the hub will let the pushing node what allowed paths it will accept, and the pusher will filter them before
sending them up. The hub will also validate that the incoming data match the allowed filters. If the pusher send data that it shouldn't, we abort
the connection and force a retry.

Filtering is only enabled for certificates registered directly on the hub. If anything has a filter, all certificates must have a filter. Note that
accessing the pull replication using already trusted certificate that isn't registered on the hub means that we don't do *any* filtering and give
unfettered access.

We now have the concept of a replication hub that can be set to: Pull, Push or Pull & Push. It can accept either a pull replication. In which case
it will work normaly, but filter the results to send or accept a push replication, in which case it will send to the pusher the allowed paths and
then validate them locally.

On the other side, we have a sink. These can be configured as either Push or Pull (never both) and will function as the other side of the hub, 
accepting data from the remote hub or pushing filtered data here.

In this context, the hub is *trusted* and the end node is *not* trusted. 

Remaining work:
* How should we have such replication when using output collections from map/reduce. 
* Optimizing pull replication for larger number of certificates.
* We use the term Pull Replication Hub throughout our code. We have now changed it to also allow pushing. How to deal with this? Same with the sink
  terminology.

ayende:
One of the things we need to see if we can do is to cheat. 
Can we provide one value at authorization phase and then use another during the actual replication setup?

src/Raven.Client/Documents/Operations/Replication/ExternalReplication.cs
```diff
@@ -13,12 +13,14 @@ public ExternalReplication(string database, string connectionStringName) : base(
             
         }
 
+        public string HubDefinitionName;
```
karmeli87: Why is it on the `ExternalReplication`?

src/Raven.Client/Documents/Operations/Replication/PullReplicationDefinition.cs
```diff
@@ -1,23 +1,125 @@
...
+                    if (path.Length > 1 && path[path.Length - 2] != '/' && path[path.Length - 2] != '-')
+                        throw new InvalidOperationException(
+                            $"When using '*' at the end of the allowed path, the previous character must be '/' or '-', but got: {path} for {kvp.Key}");
+                }
+            }
+        }
+
+        public class FilteringOptions : IDynamicJson
+        {
+            public string[] AllowedPaths;
```
karmeli87: We can actually have a different filters for pull and push.
Not hard to add and I think it might be useful.
e.g. I want to pull only configuration document, but push all the data that I've collected.

src/Raven.Client/Documents/Operations/Replication/ReplicationMode.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Raven.Client.Documents.Operations.Replication
+{
+    [Flags]
+    public enum ReplicationMode
+    {
+        None = 0,
+        Pull = 1,
+        Push = 2,
+        Invalid = 4
```
karmeli87: Why do we need that?

src/Raven.Server/Documents/Replication/AllowedPathsValidator.cs
```diff
@@ -0,0 +1,107 @@
...
+                if (id.StartsWithOrdinalNoCase(prefix))
+                    return true;
+            }
+
+            return false;
+        }
+        public AllowedPathsValidator(string[] allowedPaths)
+        {
+            {
+                _allowedPathsContext = JsonOperationContext.ShortTermSingleUse();
```
karmeli87, line 80: Not familiar with the `ShortTermSingleUse`, just wanted to notice that this isn't a "short-term" usage.

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
```diff
@@ -512,6 +512,14 @@ private void InitialHandshake()
                     //The first time we start replication we need to register the destination current CV
                     case ReplicationMessageReply.ReplyType.Ok:
                         LastAcceptedChangeVector = response.Reply.DatabaseChangeVector;
+                        // AllowedPaths will have a locally provided value for filtered pull replication.
+                        // In that case, we prefer the local value over the one from the remote server
+                        if (AllowedPaths == null)
+                        {
+                            // this is used when the other side lets us know what 
+                            // paths it is going to accept from us
+                            AllowedPaths = response.Reply.AllowedPaths;
```
karmeli87: Need to validate that changing the definition on the Hub will be reflected on the already opened connections

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
```diff
@@ -535,6 +538,13 @@ private bool AddReplicationItemToBatch(ReplicationBatchItem item, OutgoingReplic
 
         private bool ShouldSkip(ReplicationBatchItem item, OutgoingReplicationStatsScope stats, SkippedReplicationItemsInfo skippedReplicationItemsInfo)
         {
+            if (_allowedPathsValidator?.ShouldAllow(item) == false)
+            {
+                stats.RecordArtificialDocumentSkip();
+                skippedReplicationItemsInfo.Update(item);
```
karmeli87, line 544: Need to write a correct log message, don't want this to mislead us.

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -180,50 +180,101 @@ public ReplicationLoader(DocumentDatabase database, ServerStore server)
...
-                    tcpConnectionOptions.Stream,
-                    "initial-replication-message",
-                    BlittableJsonDocumentBuilder.UsageMode.None,
-                    buffer))
-                {
-                    var initialRequest = JsonDeserializationServer.ReplicationInitialRequest(readerObject);
-                    if (initialRequest.PullReplicationDefinitionName != null)
+                case TcpConnectionHeaderMessage.AuthorizationInfo.AuthorizeMethod.PullReplication:
+                case TcpConnectionHeaderMessage.AuthorizationInfo.AuthorizeMethod.PushReplication:
+                    if (supportedVersions.Replication.PullReplication == false)
```
karmeli87, line 263: Shouldn't this `if` come before the `switch` ?

karmeli87: It is a new feature so we also need to add to the supported versions an option for push/filtered replication.

src/Raven.Server/ServerWide/RawDatabaseRecord.cs
```diff
@@ -248,29 +250,64 @@ public List<ExternalReplication> ExternalReplications
                 return _externalReplications;
             }
         }
+        
+        private List<RawPullReplicationDefinition> _hubPullReplications;
 
-        private List<PullReplicationDefinition> _hubPullReplications;
-
-        public List<PullReplicationDefinition> HubPullReplications
+        public List<RawPullReplicationDefinition> RawHubPullReplications
```
karmeli87: Does this change has any backward comp. issues that we care about?

src/Raven.Client/Documents/Operations/Replication/PullReplicationDefinition.cs
```diff
@@ -1,23 +1,125 @@
...
+
+            public DynamicJsonValue ToJson()
+            {
+                return new DynamicJsonValue {[nameof(AllowedPaths)] = AllowedPaths};
+            }
+        }
+    }
+
+    public class RawPullReplicationDefinition
```
karmeli87: Why we have this and `PullReplicationDefinition`  ?

src/Raven.Client/Documents/Operations/Replication/ExternalReplication.cs
ayende: Removed

src/Raven.Client/Documents/Operations/Replication/PullReplicationDefinition.cs
ayende: Good idea.

src/Raven.Client/Documents/Operations/Replication/ReplicationMode.cs
ayende: Removed

src/Raven.Server/Documents/Replication/AllowedPathsValidator.cs
ayende: That just means that it isn't using a context. This will live as long as the validator lives, which is fine.

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
ayende: Right now, IIRC, that wouldn't be the case for things like removing certificates.
Existing connections are maintained. 

But I agree: https://issues.hibernatingrhinos.com/issue/RavenDB-15090

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
ayende: done

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
ayende: No, it shouldn't. We need to do this check if the other side requested pull replication in this manner. 
There should be no way that you'll get here if you aren't supporting pull replication

src/Raven.Client/Documents/Operations/Replication/PullReplicationDefinition.cs
ayende: Mid way through refactoring things to avoid putting certificates inside the database record.

PR #10634, opened 2020-05-28
RavenDB-14974

ml054:


src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
```diff
@@ -500,14 +500,27 @@ public Task GetTimeSeriesConfig()
...
                         timeSeriesCollection[collection.Key] = collection.Value.ToJson();
                     }
+                    
+                    var configJson = new DynamicJsonValue
+                    {
+                        [nameof(timeSeriesConfig.Collections)] = timeSeriesCollection,
+                        [nameof(timeSeriesConfig.PolicyCheckFrequency)] = timeSeriesConfig.PolicyCheckFrequency
+                    };
+                                        
+                    if (timeSeriesConfig.NamedValues != null)
```
ppekrol: Better to implement inside TimeSeriesConfiguration a `ToJson` method via IDynamicJson interface and handle all of that there.

ml054: nice catch!

src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
```diff
@@ -527,9 +517,17 @@ public async Task ConfigTimeSeries()
...
+                    {
+                        return;
+                    }
+
+                    var hasCollectionsConfig = configuration.TryGet(nameof(TimeSeriesConfiguration.Collections), out BlittableJsonReaderObject collections) &&
+                                               collections?.Count > 0;
+
+                    var hasNamedValuesConfig = configuration.TryGet(nameof(TimeSeriesConfiguration.NamedValues), out BlittableJsonReaderObject namedValues);
+                    
+                    if (hasCollectionsConfig == false && hasNamedValuesConfig == false)
```
ppekrol: What if `hasCollectionsConfig == false` and `hasNamedValuesConfig == true` ? The code below can throw? Also are we doing something with `namedValues` variable ?

ppekrol (approved):

PR #10635, opened 2020-05-28
RavenDB-14960 added WaitForNonStaleResults

ppekrol:


ayende (commented):
Why would this cause any change in behavior?

ppekrol:
I'm assuming that cluster was unstable and we returned stale results here. Would prefer to see a staleness exception here to confirm that.

PR #10637, opened 2020-05-28
RavenDB-15019 added TryConvert method for number conversions

ppekrol:


src/Raven.Server/Documents/Indexes/Static/StaticIndexBase.cs
```diff
@@ -180,6 +180,76 @@ protected void AddMapInternal(string collection, string subCollecction, Indexing
             funcs.Add(map);
         }
 
+        public T? TryConvert<T>(object value)
+            where T : struct
+        {
+            if (value == null || value is DynamicNullObject)
+                return null;
```
ppekrol: Dynamic null propagation will not work if this method is used. Also the values from this method will be indexed not like DynamicNullValue.ImplicitNull ones

src/Raven.Server/Documents/Indexes/Static/StaticIndexBase.cs
```diff
@@ -180,6 +180,76 @@ protected void AddMapInternal(string collection, string subCollecction, Indexing
...
+                if (v is LazyStringValue lsv && double.TryParse(lsv, out var r))
+                    return r;
+
+                return null;
+            }
+
+            static long? TryConvertToLong(object v)
+            {
+                if (v is double d)
+                    return (long)d;
```
ppekrol, line 237: We will loose precision here if the value was double. Should we do that or fail to convert?

ayende: No, that is fine

src/Raven.Server/Documents/Indexes/Static/StaticIndexBase.cs
ayende: Yea,I don't like it, but it make sense

ayende (approved):

src/Raven.Server/Documents/Indexes/Static/StaticIndexBase.cs
grisha-kotler: Why can't we return `DynamicNullObject`?

ppekrol: It will be fine. I've somehow assumed that we will have issues when passing to methods like e.g. Convert.ToInt64 when we would not recognize the dynamic type properly.

src/Raven.Server/Documents/Indexes/Static/StaticIndexBase.cs
```diff
@@ -180,6 +180,76 @@ protected void AddMapInternal(string collection, string subCollecction, Indexing
...
+                if (v is int i)
+                    return i;
+                if (v is long l)
+                    return l;
+                if (v is float f)
+                    return f;
+                if (v is LazyStringValue lsv && double.TryParse(lsv, out var r))
+                    return r;
+
+                return null;
```
ayende, line 231: Still null

src/Raven.Server/Documents/Indexes/Static/StaticIndexBase.cs
```diff
@@ -180,6 +180,76 @@ protected void AddMapInternal(string collection, string subCollecction, Indexing
...
+                if (v is int i)
+                    return i;
+                if (v is long l)
+                    return l;
+                if (v is float f)
+                    return (long)f;
+                if (v is LazyStringValue lsv && long.TryParse(lsv, out var r))
+                    return r;
+
+                return null;
```
ayende, line 249: Still null

src/Raven.Server/Documents/Indexes/Static/StaticIndexBase.cs
ppekrol:
```
                var dbl = TryConvertToDouble(value);
                if (dbl.HasValue == false)
                    return DynamicNullObject.Null;
```

PR #10642, opened 2020-06-01
RDBCL-1342: don't update the _previousLicenseCores if it wasn't chang…

garayx:
…ed in the license

### DIFF
--- a/src/Raven.Server/Commercial/LicenseManager.cs
+++ b/src/Raven.Server/Commercial/LicenseManager.cs
@@ -514,15 +514,21 @@ namespace Raven.Server.Commercial
 
         private void SetLicense(Guid id, Dictionary<string, object> attributes)
         {
-            _previousLicenseCores = _licenseStatus.MaxCores;
-
-            _licenseStatus = new LicenseStatus
+            var newLicenseStatus = new LicenseStatus
             {
                 Id = id,
                 ErrorMessage = null,
                 Attributes = attributes,
                 FirstServerStartDate = _licenseStatus.FirstServerStartDate
             };
+
+            if (newLicenseStatus.MaxCores != _licenseStatus.MaxCores)
+            {
+                // it should be updated only if license cores were changed
+                _previousLicenseCores = _licenseStatus.MaxCores;
+            }
+
+            _licenseStatus = newLicenseStatus;
         }
 
         private async Task<bool> ContinueActivatingLicense(License license, bool skipLeaseLicense, string raftRequestId, bool ensureNotPassive, LicenseStatus newLicenseStatus)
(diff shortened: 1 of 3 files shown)


ppekrol (approved):

src/Raven.Server/Commercial/LicenseManager.cs
```diff
@@ -514,15 +514,21 @@ private void ResetLicense(string error)
...
                 Id = id,
                 ErrorMessage = null,
                 Attributes = attributes,
                 FirstServerStartDate = _licenseStatus.FirstServerStartDate
             };
+
+            if (newLicenseStatus.MaxCores != _licenseStatus.MaxCores)
+            {
+                // it should be updated only if license cores were changed
+                _previousLicenseCores = _licenseStatus.MaxCores;
```
grisha-kotler: This will update the licensed cores if we manually updated it

grisha-kotler (changes requested):

src/Raven.Server/Commercial/LicenseManager.cs
```diff
@@ -955,7 +965,7 @@ private void VerifyCoresPerNode(Dictionary<string, DetailsPerNode> detailsPerNod
 
             if (licenseChanged && _licenseStatus.MaxCores > _previousLicenseCores)
             {
-                changedNodes = detailsPerNode.Select(x => x.Key).ToList();
+                changedNodes = detailsPerNode.Where(x=>x.Value.Modified == false).Select(x => x.Key).ToList();
```
ppekrol, line 977: Shouldn't this be `.Where(x => x.Value.Modified)` ?

garayx: No, Modified means that the current number of utilized cores, was set by the user explicitly.
We only want to reassign cores that wasn't set by user.

src/Raven.Server/Commercial/LicenseManager.cs
```diff
@@ -358,7 +358,8 @@ public async Task ChangeLicenseLimits(string nodeTag, int newAssignedCores, stri
                         InstalledMemoryInGb = installedMemoryInGb,
                         UsableMemoryInGb = usableMemoryInGb,
                         BuildInfo = buildInfo,
-                        OsInfo = osInfo
+                        OsInfo = osInfo,
+                        Modified = newAssignedCores != _licenseStatus.MaxCores
```
ppekrol: Not sure that I follow this. If I have a license with 64 cores and a node with 32 cores then it will be always true? How this is 'set by user explicitly'?

grisha-kotler:
Created https://github.com/ravendb/ravendb/pull/10736 instead

PR #10649, opened 2020-06-03
RavenDB-14594 Allow to change the 5 years expiration on client certif…

Danielle9897:
…icates


Danielle9897:
Please note - 
Although the expiration date that is passed to the server is **correct** - verified with the debugger -
A full Let's Encrypt flow was not tested due to other connectivity issues from my machine.

ml054 (approved):

PR #10656, opened 2020-06-04
RavenDB-15115 make sure the appended entires are in 1ms precision

karmeli87:


karmeli87:
As a side note:

1. There is a different behavior for appending from the session vs operation.
In the operation we ensure it is in ms precision in the client and overwrite with the last value.
While in the session we append as is.

2. Another worry is, that the current exception will break replication/import. 
Not sure if we want to do things differently. 
As alternative in case of duplicates we can either skip or increase the number of values. 

Don't like either solution..

karmeli87:
Another idea is to pass `AppendOptions` that will define the behavior for duplicates.

```
AppendOptions.OnDuplicate = Throw
AppendOptions.OnDuplicate = Overwrite // harder to do, can be added later
AppendOptions.OnDuplicate = AddAsValue
AppendOptions.OnDuplicate = KeepOriginal
```

ayende:
I don't follow why this is an issue? Aren't we already doing this in the code?

karmeli87:
We have different code paths that are doing a bit different things.
Need to unify at least the operations and the session.

1. operations:
- client: **ensure milliseconds**, sort by timestamp before sending to the server
- server: sort by time stamp

2. bulk:
- client: send the original order with the original timestamp
- server: sort by time stamp

3. session:
- client: send the original timestamp with the original order
- server: sort by time stamp

4. JS
has a direct access to the append method on the server.

PR #10659, opened 2020-06-07
RavenDB-13977 ETL of TimeSeries

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-13977

https://github.com/ravendb/ravendb/pull/10587


haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15083

src/Raven.Server/Documents/ETL/EtlProcess.cs
```diff
@@ -1086,6 +1146,13 @@ public override EtlProcessProgress GetProgress(DocumentsOperationContext documen
                     counterGroupsToProcess += Database.DocumentsStorage.CountersStorage.GetNumberOfCounterGroupsToProcess(documentsContext, collection, lastProcessedEtag, out total);
                     totalCounterGroupsCount += total;
                 }
+                
+                if (ShouldTrackTimeSeries())
+                {
+                    //TODO To add deleted ranges
```
ayende: When / how is this going to be handled?

src/Raven.Server/Documents/ETL/EtlTransformer.cs
```diff
@@ -164,6 +171,40 @@ private JsValue LoadCounter(JsValue self, JsValue[] args)
...
+
+            var from = ScriptRunner.GetDateArg(args[1], signature, "from"); 
+            var to = ScriptRunner.GetDateArg(args[2], signature, "to"); 
+                
+            JsValue loadTimeSeriesReference = (JsValue)Transformation.TimeSeriesTransformation.Marker + timeSeriesName;
+
+            if ((Current.Document.Flags & DocumentFlags.HasTimeSeries) == DocumentFlags.HasTimeSeries)
+            {
+                var reader = Database.DocumentsStorage.TimeSeriesStorage.GetReader(Context, Current.DocumentId, timeSeriesName, from, to);
+                //TODO Maybe to return null if no entries
```
ayende: *can* this return null? I thought it would return empty list?

src/Raven.Server/Documents/ETL/FilterTimeSeriesDeleteRangeEnumerator.cs
```diff
@@ -0,0 +1,96 @@
...
+    {
+        private readonly IEnumerator<TimeSeriesDeletedRangeItem> _timeSeries;
+        private readonly EtlStatsScope _stats;
+        private readonly DocumentsStorage _docsStorage;
+        private readonly DocumentsOperationContext _context;
+        private readonly long _lastProcessedDocEtagInBatch;
+
+        public FilterTimeSeriesDeleteRangeEnumerator(IEnumerator<TimeSeriesDeletedRangeItem> timeSeries, EtlStatsScope stats, DocumentsStorage docsStorage, DocumentsOperationContext context, long lastProcessedDocEtagInBatch)
+        {
+            //TODO To check if should filter on collection
```
ayende: This is already done, no? I mean, the `timeSeries` enumerator already do that

src/Raven.Server/Documents/ETL/FilterTimeSeriesDeleteRangeEnumerator.cs
```diff
@@ -0,0 +1,96 @@
...
+            while (_timeSeries.MoveNext())
+            {
+                var current = _timeSeries.Current;
+
+                if (CanMoveNextAndNotExceedLastDocumentInBatch(current.Etag, _lastProcessedDocEtagInBatch) == false)
+                    return false;
+
+                TimeSeriesValuesSegment.ParseTimeSeriesKey(current.Key, _context, out var docId, out var name);
+                var doc = _docsStorage.Get(_context, docId);
+                if (doc != null && current.Etag > doc.Etag)
```
ayende: Why do you have this check?
A time series can appear both before and after the document etag.

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
```diff
@@ -173,6 +187,37 @@ private JsValue AddCounter(JsValue self, JsValue[] args)
...
+
+            var timeSeriesReference = args[0];
+
+            if (timeSeriesReference.IsNull())
+                return self;
+
+            if (timeSeriesReference.IsString() == false || timeSeriesReference.AsString().StartsWith(Transformation.TimeSeriesTransformation.Marker) == false)
+            {
+                var message =
+                    $"{Transformation.TimeSeriesTransformation.AddTimeSeries.Name} method expects to get the reference to a counter while it got argument of '{timeSeriesReference.Type}' type";
```
ayende:
```suggestion
                    $"{Transformation.TimeSeriesTransformation.AddTimeSeries.Name} method expects to get the reference to a timeseries while it got argument of '{timeSeriesReference.Type}' type";
```

ayende:
I stopped the review half way there, I'll do a full review with you once this is ready for merge.

src/Raven.Server/Documents/ETL/EtlTransformer.cs
arekpalinski: It doesn't seem so - under the covers `new TimeSeriesReader()` is called

src/Raven.Server/Documents/ETL/FilterTimeSeriesDeleteRangeEnumerator.cs
```diff
@@ -0,0 +1,96 @@
...
+                {
+                    Current = current;
+                    return true;
+                }
+
+                //TODO To check
+                // Time series deleted ranges has lower etag than its document - we skip it to avoid
+                // sending a time series of document that can not exist on the destination side
```
arekpalinski: I see that you have TODO to check this - but just to be sure - will we have the same behavior like for counters? I mean that those TS will be send together with document

src/Raven.Server/Documents/ETL/FilterTimeSeriesEnumerator.cs
```diff
@@ -0,0 +1,96 @@
+using System.Collections;
+using System.Collections.Generic;
+using Raven.Client.Documents.Operations.Counters;
+using Raven.Server.Documents.ETL.Stats;
+using Raven.Server.Documents.TimeSeries;
+using Raven.Server.ServerWide.Context;
+
+namespace Raven.Server.Documents.ETL
+{
+    //TODO To make common base class with time-series deleted range and counters
```
arekpalinski: Please do so only after you'll be sure that we have exactly the same behavior for time series like for counters.

src/Raven.Server/Documents/ETL/EtlTransformer.cs
haludi: If the time-series does not exist so `LoadTimeSeries` will return `null` (not in the current implementation).
I thought to return `null` also if the time-series does exist but there are no entries in the range.

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
```diff
@@ -233,38 +291,69 @@ public override void Transform(RavenEtlItem item, EtlStatsScope stats)
                         {
                             AddSingleCounterGroup(item.DocumentId, item.CounterGroupDocument);
                         }
-
+                        break;
+                    case EtlItemType.TimeSeries:
+                        if (_script.HasTransformation)
+                        {
+                            //TODO Is it make sense to run the main script (dry run) to know if the document is actual loaded to same collection. The second option is to check if the document can loaded to same collection according to load signature
```
arekpalinski: I'm not sure that I follow - where do we run the script to know if a doc is loaded to the same collection? `_script.IsLoadedToDefaultCollection` call doesn't _run_ the script. Can you please elaborate?

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
```diff
@@ -344,6 +432,139 @@ private void AddSingleCounterGroup(LazyStringValue docId, BlittableJsonReaderObj
...
+        }
+        
+        private void HandleSingleTimeSeriesSegment(LazyStringValue docId, TimeSeriesSegmentEntry segmentEntry, string loadBehaviourFunction = null)
+        {
+            var timeSeriesEntries = segmentEntry.Segment.YieldAllValues(Context, segmentEntry.Start, false);
+            
+            if (loadBehaviourFunction != null && FilterSingleTimeSeriesSegmentByLoadBehaviourScript(ref timeSeriesEntries, docId, segmentEntry, loadBehaviourFunction))
+                return;
+
+            //TODO What should be the behavior if the load behaviour function return false? remove all ?
```
arekpalinski: This is https://issues.hibernatingrhinos.com/issue/RavenDB-15083 right?

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
```diff
@@ -344,6 +432,139 @@ private void AddSingleCounterGroup(LazyStringValue docId, BlittableJsonReaderObj
...
+            {
+                _currentRun.AddTimeSeries(docId, segmentEntry.Name, entry);
+            }
+        }
+
+        private bool FilterSingleTimeSeriesSegmentByLoadBehaviourScript(
+            ref IEnumerable<SingleResult> timeSeriesEntries, 
+            LazyStringValue docId,
+            TimeSeriesSegmentEntry segmentEntry, 
+            string loadBehaviourFunction)
```
arekpalinski: Can we use "behavior" instead of "behaviour" for consistency with the naming that we already have?

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
```diff
@@ -344,6 +432,139 @@ private void AddSingleCounterGroup(LazyStringValue docId, BlittableJsonReaderObj
...
+                else
+                {
+                    var toLoadLocal = (From: DateTime.MinValue, To: DateTime.MaxValue);
+                    foreach ((JsValue jsValue, PropertyDescriptor propertyDescriptor) in scriptRunnerResult.Instance.AsObject().GetOwnProperties())
+                    {
+                        var key = jsValue.AsString();
+                        switch (key)
+                        {
+                            case "from":
+                            case "From":
```
arekpalinski: it won't work for `froM` for example. why not to just use `StringComparer.OrdinalIgnoreCase`?

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
```diff
@@ -344,6 +432,139 @@ private void AddSingleCounterGroup(LazyStringValue docId, BlittableJsonReaderObj
...
+                                break;
+                            case "to":
+                            case "To":
+                                if (toLoadLocal.To != DateTime.MaxValue)
+                                    throw new InvalidOperationException($"Duplicate of property `To`/`to`. docId({docId}), timeSeriesName({timeSeriesName}), function({function})");
+                                toLoadLocal.To = propertyDescriptor.Value.AsDate().ToDateTime();
+                                break;
+                            default:
+                                //TODO Maybe just ignore. Can be useful if we will add property in future and someone will run it with current version
+                                throw new InvalidOperationException($"Returned object should contain only `from` and `to` property but contain `{key}`. docId({docId}), timeSeriesName({timeSeriesName}), function({function})");
```
arekpalinski: if we add something in the future then we'll have to add a handling for that here anyway. I don't see the point to silently ignore the value here

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
```diff
@@ -387,7 +608,69 @@ private List<CounterOperation> GetCounterOperationsFor(RavenEtlItem item)
 
             return counterOperations;
         }
+        //TODO to make the signature prettier
```
arekpalinski: +1

I think the returned object deserves a dedicated class

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
```diff
@@ -568,6 +875,20 @@ public bool IsLoadedToDefaultCollection(RavenEtlItem item, string loadToCollecti
 
                 return collection?.CompareTo(loadToCollection) == 0;
             }
+            
+            public bool IsLoadedToDefaultCollection(RavenEtlItem item)
```
arekpalinski: Where is it used?

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlScriptRun.cs
```diff
@@ -210,6 +306,18 @@ public List<ICommandData> GetCommands()
                             FromEtl = true
                         });
                     }
+                    
+                    if (_timeSeriesByJsReference != null && _timeSeriesByJsReference.TryGetValue(put.Key, out var timeSeriesOperations))
+                    {
+                        foreach (var (_, operation) in timeSeriesOperations)
+                        {
+                            //TODO RavenDB-15083
```
arekpalinski: Please check Oren's answer - https://issues.hibernatingrhinos.com/issue/RavenDB-15070#focus=streamItem-67-48099.0-0

arekpalinski (commented):
I see there is still a lot of TODOs in the code but please check my current comments

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
haludi: `_script.IsLoadedToDefaultCollection` just check if there is a possibility to load the doc to the same collection (check for the existence of loadTo<collection> in the script and compare `collection` to item collection).
But if the document is not actually loaded because of a condition.
Like this for example:
```
if(this.Name === 'SomeName')
    return;
var person = loadToUsers(this);
```
We will keep trying to send time-series any time they change (I think we do that in `counters` as well).
This can be a big overhead if most of the documents are not loaded (or not loaded to the same collection).
I added `_currentRun.IsDocumentLoadedToSameCollection(item.DocumentId)` in `EtlItemType.Document` case but I need the script to run to use this function.
I don't want to actually run the script so the doc will not reload over and over for any time-series change but maybe a dry run? I tried to play with it a little bit but I think this requires a big change in the code (but not sure) so I left the comment.

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
haludi: yes

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
haludi: in `RavenEtlDocumentTransformer.Transform` in `EtlItemType.TimeSeries` case

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
arekpalinski: This sounds like non trivial change so its a bit risky. It's shouldn't be part of adding TS to ETL for sure. Please create a separate ticket for that.

arekpalinski: https://issues.hibernatingrhinos.com/issue/RavenDB-15147

src/Raven.Client/Documents/Commands/Batches/TimeSeriesBatchCommandData.cs
```diff
@@ -10,7 +10,7 @@ namespace Raven.Client.Documents.Commands.Batches
 {
     public class TimeSeriesBatchCommandData : ICommandData
     {
-        public TimeSeriesBatchCommandData(string documentId, string name, List<TimeSeriesOperation.AppendOperation> appends, List<TimeSeriesOperation.RemoveOperation> removals)
+        public TimeSeriesBatchCommandData(string documentId, string name, IList<TimeSeriesOperation.AppendOperation> appends, List<TimeSeriesOperation.RemoveOperation> removals)
```
arekpalinski: Why `IList` here for `appends` but `List` for removals? It's our public API let's try to be keep consistency

src/Raven.Client/Documents/Operations/ETL/EtlProcessState.cs
```diff
@@ -23,6 +23,8 @@ public EtlProcessState()
 
         public string NodeTag { get; set; }
 
+        public HashSet<string> SkippedTimeSeriesDocs { get; set; }
```
arekpalinski, line 29: Don't we need to keep that info per node? So shouldn't it be `Dicitonary<string,  HashSet<string>>` ? Also this is needed for specific time series case - a comment here explaining why do we need it would be very handy for someone looking at that in the future.

src/Raven.Client/Util/DisposableAsyncAction.cs
```diff
@@ -0,0 +1,36 @@
...
+
+using System;
+using System.Threading.Tasks;
+
+namespace Raven.Client.Util
+{
+    /// <summary>
+    /// A helper class that translate between Disposable and async action
+    /// </summary>
+    public class DisposableAsyncAction : IAsyncDisposable
```
arekpalinski: I see only one usage of this class and it's only in tests. I don't see why we would like to add it to the client assembly given that we don't have any usages of it there. Let's move it to `Tests.Infrastructure` project.

src/Raven.Server/Documents/ETL/EtlProcess.cs
```diff
@@ -154,101 +158,162 @@ protected EtlProcess(Transformation transformation, TConfiguration configuration
...
                 switch (type)
                 {
                     case EtlItemType.Document:
-                        var enumerators = new List<(IEnumerator<Document> Docs, IEnumerator<Tombstone> Tombstones, string Collection)>(Transformation.Collections.Count);
-
-                        if (Transformation.ApplyToAllDocuments)
-                        {
-                            var docs = Database.DocumentsStorage.GetDocumentsFrom(context, fromEtag, 0, long.MaxValue).GetEnumerator();
-                            scope.EnsureDispose(docs);
+                        ExtractDocuments(context, merged, fromEtag, stats, scope);
```
arekpalinski, line 178: Can we mark that `merged` is modified inside `ExtractXXXX` methods by adding `ref` to that parameter? I know it will be working the same but it's more explicit that items are added to it under the covers.

src/Raven.Server/Documents/ETL/EtlProcess.cs
```diff
@@ -406,22 +471,27 @@ private void EnterFallbackMode()
 
         protected abstract int LoadInternal(IEnumerable<TTransformed> items, DocumentsOperationContext context);
 
-        public bool CanContinueBatch(EtlStatsScope stats, TExtracted currentItem, int batchSize, DocumentsOperationContext ctx)
+        private bool CanContinueBatch(EtlStatsScope stats, TExtracted currentItem, int batchSize, DocumentsOperationContext ctx)
         {
-            if (currentItem.Type == EtlItemType.CounterGroup)
+            /*
+            TODO We will keep processing time-series (and counter) even if the batch end because of cpuCredit/timeout/lowMemory/batchSize. I
```
arekpalinski: Please create a ticket for that that and remove TODO comment here

src/Raven.Server/Documents/ETL/EtlProcess.cs
```diff
@@ -1013,7 +1075,7 @@ public override OngoingTaskConnectionStatus GetConnectionStatus()
                             var ravenEtlItem = testScript.IsDelete ? new RavenEtlItem(tombstone, docCollection, EtlItemType.Document) : new RavenEtlItem(document, docCollection);
 
                             var results = ravenEtl.Transform(new[] {ravenEtlItem}, context, new EtlStatsScope(new EtlRunStats()),
-                                new EtlProcessState());
+                                new EtlProcessState{SkippedTimeSeriesDocs = new HashSet<string> {testScript.DocumentId}});
```
arekpalinski, line 1073: Why a test doc is always assigned to `SkippedTimeSeriesDocs` during the script test?

src/Raven.Server/Documents/ETL/EtlTransformer.cs
```diff
@@ -113,14 +125,16 @@ private JsValue LoadToFunctionTranslator(string name, JsValue self, JsValue[] ar
...
+        protected abstract void AddLoadedTimeSeries(JsValue reference, string name, IEnumerable<SingleResult> entries);
 
         private JsValue LoadAttachment(JsValue self, JsValue[] args)
         {
             if (args.Length != 1 || args[0].IsString() == false)
                 ThrowInvalidScriptMethodCall($"{Transformation.LoadAttachment}(name) must have a single string argument");
 
             var attachmentName = args[0].AsString();
-            JsValue loadAttachmentReference = (JsValue)Transformation.AttachmentMarker + attachmentName;
+            var loadAttachmentReference = (JsValue)$"{Transformation.AttachmentMarker}{attachmentName}{Guid.NewGuid():N}";
```
arekpalinski: Why do we need do add Guid?

src/Raven.Server/Documents/ETL/EtlTransformer.cs
```diff
@@ -164,6 +178,35 @@ private JsValue LoadCounter(JsValue self, JsValue[] args)
...
+                ThrowInvalidScriptMethodCall($"{signature} must have {paramsCount} arguments");
+                
+            if(args[0].IsString() == false)
+                ThrowInvalidScriptMethodCall($"{signature}. The argument timeSeriesName must be a string");
+            var timeSeriesName = args[0].AsString();
+
+            var from = ScriptRunner.GetDateArg(args[1], signature, "from"); 
+            var to = ScriptRunner.GetDateArg(args[2], signature, "to"); 
+                
+            var loadTimeSeriesReference = (JsValue)Transformation.TimeSeriesTransformation.Marker + timeSeriesName + from.Ticks + ':' + to.Ticks;
```
arekpalinski: Maybe let's put move this to a dedicated method called `CreateLoadTimeSeriesReference()`. And probably same for attachment references and others

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
```diff
@@ -196,80 +247,112 @@ public override void Transform(RavenEtlItem item, EtlStatsScope stats)
...
             }
 
             _commands.AddRange(_currentRun.GetCommands());
         }
 
+        private bool ShouldLoadTimeSeriesWithDoc(RavenEtlItem item, EtlProcessState state)
+        {
+            //If an Etag of time-series is lower then its document Etag then replication can send the time-series before the document.
+            //In this situation Etl process will skip the time-series, mark it as skipped and will send all of it here with the document
+            return state.SkippedTimeSeriesDocs != null && state.SkippedTimeSeriesDocs.Remove(item.DocumentId);
```
arekpalinski, line 354: Is is ok to remove it from state? What happens if we ETL batch fails afterwards? 

We need to make sure it won't be persisted and we won't hold this state instance in next batches.

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
```diff
@@ -344,6 +426,196 @@ private void AddSingleCounterGroup(LazyStringValue docId, BlittableJsonReaderObj
...
+            }
+            var timeSeriesEntries = segmentEntry.Segment.YieldAllValues(Context, segmentEntry.Start, false);
+            if (loadBehaviorFunction != null && FilterSingleTimeSeriesSegmentByLoadBehaviourScript(ref timeSeriesEntries, docId, segmentEntry, loadBehaviorFunction))
+                return;
+
+            if (doc.TryGetMetadata(out var metadata) == false ||
+                metadata.TryGet(Constants.Documents.Metadata.TimeSeries, out BlittableJsonReaderArray timeSeriesNames) == false)
+                throw new InvalidOperationException($"Document {docId} has time-series but its metadata doesn't");
+
+            //TODO Time-series name should be taken from time-series storage after the relevant implementation will change
```
arekpalinski: Is there any ticket for that?

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
```diff
@@ -387,7 +659,70 @@ private List<CounterOperation> GetCounterOperationsFor(RavenEtlItem item)
 
             return counterOperations;
         }
+        
+        private Dictionary<string, TimeSeriesModifications> GetTimeSeriesFor(RavenEtlItem item, string loadBehaviorFunction)
+        {
+            if ((Current.Document.Flags & DocumentFlags.HasTimeSeries) != DocumentFlags.HasTimeSeries)
+                return null;
+            
+            //TODO Time-series name should be taken from time-series storage after the relevant implementation will change
```
arekpalinski: Same comment as above. I don't think any TODO should be merged to the master (v5.0) branch. If this is something we want to change / improve in the future let's create an issue for that and remove TODO comment here. This way we won't forget about it.

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlScriptRun.cs
```diff
@@ -66,41 +79,52 @@ public void PutFullDocument(string id, BlittableJsonReaderObject doc, List<Attac
...
                 });
             }
+            
+            if (timeSeriesOperations != null)
+            {
+                foreach (var operation in timeSeriesOperations)
+                {
+                    _fullDocuments.Add(new TimeSeriesBatchCommandData(id, operation.Name, operation.Appends, operation.Removals)
+                    {
+                        FromEtl = true
```
arekpalinski, line 89: Do we have any different behavior on server side if `TimeSeriesBatchCommandData.FromEtl` is `true`?

test/SlowTests/Server/Documents/ETL/EtlCountersTests.cs
```diff
@@ -0,0 +1,80 @@
...
+using Xunit.Abstractions;
+
+namespace SlowTests.Server.Documents.ETL
+{
+    public class EtlCountersTests : EtlTestBase
+    {
+        private const int _waitInterval = 1000;
+
+        private readonly Options _options = Debugger.IsAttached
+            ? new Options {ModifyDatabaseRecord = record => record.Settings[RavenConfiguration.GetKey(x => x.Etl.ExtractAndTransformTimeout)] = "300"}
```
arekpalinski, line 17: Why do we have this?

test/SlowTests/Server/Documents/ETL/EtlCountersTests.cs
```diff
@@ -0,0 +1,80 @@
...
+                }
+                await AssertWaitForTrueAsync(async () =>
+                {
+                    using var session = dest.OpenAsyncSession();
+                    return (await session.CountersFor(entity.Id).GetAllAsync()).Any() == false;
+                }, interval:_waitInterval);
+            }
+            catch (Exception e)
+            {
+                ThrowWithEtlErrors(src, e);
```
arekpalinski: Try catch on entire test method looks a bit weird. I understand you want to include some additional info but is it really needed? That's handy for debugging a problem with failing test but I think it's more important to immediately see which assertion has failed.

test/SlowTests/Server/Documents/ETL/EtlTimeSeriesTests.cs
```diff
@@ -0,0 +1,2480 @@
...
+                    Assert.Equal(time, actualSent.Timestamp);
+                    Assert.Equal(value, actualSent.Values.FirstOrDefault());
+                }
+            }
+            catch (Exception)
+            {
+                if (shouldSuccess)
+                    throw;
+
+                shouldSuccess = true;
```
arekpalinski: I don't like this try catch on entire method body. Can we do better? If we have `shouldSuccess: false` then shouldn't it throw on _specific_ assert or operation. I think we should handle it there instead of generic `try / catch`

test/SlowTests/Server/Documents/ETL/EtlTimeSeriesTests.cs
```diff
@@ -0,0 +1,2480 @@
...
+                    if (ts == null)
+                        return false;
+                    return ts.All(t => 
+                        (t.Timestamp >= ranges[0].from && t.Timestamp <= ranges[0].to)
+                        || (t.Timestamp >= ranges[1].from && t.Timestamp <= ranges[1].to));
+                }, interval: _waitInterval);
+            }
+            catch (Exception e)
+            {
+                ThrowWithEtlErrors(src, e);
```
arekpalinski: We need to handle it better - it's not a good practice to `try / catch` entire test

test/SlowTests/Server/Documents/ETL/Raven/RavenDB_11157_Raven.cs
```diff
@@ -965,10 +965,9 @@ public void Error_if_load_counter_behavior_func_doesnt_match_any_collection_that
 
             config.Initialize(new RavenConnectionString() { Database = "Foo", TopologyDiscoveryUrls = new[] { "http://localhost:8080" } });
 
-            List<string> errors;
-            config.Validate(out errors);
+            config.Validate(out List<string> errors);
 
-            Assert.Equal(1, errors.Count);
+            Assert.True(0 < errors.Count);
```
arekpalinski, line 970: Why can we have more that 1 error?

test/SlowTests/Server/Documents/ETL/Raven/RavenDB_9072.cs
```diff
@@ -86,7 +86,7 @@ public async Task CanTestScript()
 
                         Assert.Equal(0, result.TransformationErrors.Count);
 
-                        Assert.Equal(5, result.Commands.Count);
+                        Assert.True(result.Commands.Count >= 4);
```
arekpalinski: Why it changed?

src/Raven.Client/Documents/Commands/Batches/TimeSeriesBatchCommandData.cs
haludi: Only because those are the types of `TimeSeriesOperation`.
I will change `removals` to `IList` as well

src/Raven.Server/Documents/ETL/EtlProcess.cs
haludi: So it will process the time-series as well

src/Raven.Server/Documents/ETL/EtlTransformer.cs
haludi: If the user is calling `loadAttachment` twice to the same attachment.
```
var user = loadToUsers(this);
user.addAttachment(loadAttachment('photo'));
var person = loadToPeople(this);
person.addAttachment(loadAttachment('photo'));
```
  
The same `Stream` can't be used twice so anyway there is a need to open two separate `Stream`s.
As it was behaving so far (Before Jint changes) the `loadAttachment` return two different reference so I keep it the same.
But actually I thought maybe to do A "Lazy loading" and return just function/parameters and to relly load later in the `addAttachemt`.
In this approach, the user can use the same loaded twice 
```
var photo = loadAttachment('photo')
var user = loadToUsers(this);
user.addAttachment(photo);
var person = loadToPeople(this);
person.addAttachment(photo);
```

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
haludi: The `SkippedTimeSeriesDocs` is stored only in the end of the batch in `UpdateEtlProcessStateCommand`

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
haludi: Now there is - https://issues.hibernatingrhinos.com/issue/RavenDB-15333

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlScriptRun.cs
haludi: Yes, the batch command will not throw if the doc does not exist in the destination

test/SlowTests/Server/Documents/ETL/EtlCountersTests.cs
haludi: To give more time while debugging

test/SlowTests/Server/Documents/ETL/EtlCountersTests.cs
haludi: I found it very helpful. 
If you got failing test without the Etl error you have to rerun the test.
At best, it will just time-consuming but if the test becomes failing in once a couple of runnings the extra info is very important.
In the end, the final error includes the exact assert but you right it is not immediately.

src/Raven.Client/Documents/Commands/Batches/TimeSeriesBatchCommandData.cs
arekpalinski: ok

src/Raven.Server/Documents/ETL/EtlProcess.cs
arekpalinski: But it it mandatory to set it? If we have `addTimeSeries()` call it should add a time series anyway, right?

src/Raven.Server/Documents/ETL/EtlTransformer.cs
arekpalinski: It's ok. I was just curious why we need to ensure uniqueness - good catch.

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
arekpalinski: ok then

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlScriptRun.cs
arekpalinski: ok

test/SlowTests/Server/Documents/ETL/EtlCountersTests.cs
arekpalinski: ok, I missed `Debugger.IsAttached`

test/SlowTests/Server/Documents/ETL/EtlCountersTests.cs
arekpalinski: yeah, I understand your point but this is very strange. Can we handle it somehow better? Maybe do something in the base class `EtlTestBase` ?

test/SlowTests/Server/Documents/ETL/Raven/RavenDB_11157_Raven.cs
haludi: Yes, I added a check to test no `loadCountersOf<CollectionName>Behavior` defined without a call to load to this collection

test/SlowTests/Server/Documents/ETL/Raven/RavenDB_9072.cs
haludi: Because we don't send the delete doc operation automatically when there is a script.

src/Raven.Server/Documents/ETL/EtlProcess.cs
haludi: https://issues.hibernatingrhinos.com/issue/RavenDB-15331

src/Raven.Server/Documents/ETL/EtlProcess.cs
haludi: If the user define `loadBehaviorloadTimeSeriesOf<CollectionName>Behavior(docId, timeSeriesName)` we will not run it in document change because the decision was to process time-series only on time-series change unless there is skipped doc. So the testScript will not test it.

test/SlowTests/Server/Documents/ETL/Raven/RavenDB_9072.cs
arekpalinski: Then why not `Assert.Equal(4, result.Commands.Count);` ?

test/SlowTests/Server/Documents/ETL/Raven/RavenDB_11157_Raven.cs
arekpalinski: So how many errors do we expect? 2 ? If then then why not  `Assert.Equal(2, ... `?

src/Raven.Server/Documents/ETL/EtlProcess.cs
arekpalinski: Okay, I see. Makes sense

test/SlowTests/Server/Documents/ETL/Raven/RavenDB_11157_Raven.cs
haludi: I don't like tests that test the implementation and demand update on implementation changes.
If in the future someone will decide to add a check he will need to update this test again.
But if you prefer I will use `Equal` here.

test/SlowTests/Server/Documents/ETL/Raven/RavenDB_9072.cs
haludi: I don't like tests that test the implementation and demand update on implementation changes.
There are three loads in the script so actually as I see it it will be better to just check for 3 put commands.
But if you prefer I will use `Equal` here.

arekpalinski: I prefer to see a failing assertion and fix it once the implementation changes than silently pass an assertion while the original condition is no longer true.

Given the example here `result.Commands.Count >= 4` it's very risky because if someone make a mistake and will start sending each ETL command twice then `result.Commands.Count: 8` won't make the test to fail.

src/Raven.Client/Documents/Operations/ETL/EtlProcessState.cs
haludi: No, when a node skips time-series while ETL because its document didn't replicate yes and then the ETL pass to another node the new node will skip this time-series because it was "processed" (skipped) by the first node so the time-series will not be marked as skipped by the new node. So when the new node will get to ETL the document it will see the document time-series was skipped by the first node and so will load the time-series with it.

src/Raven.Server/Documents/ETL/EtlProcess.cs
haludi: The `merged` is immutable because it is in `using` and so cannot be used as `ref`.
I have couple of solutions to avoid that but none of them is pretty.

src/Raven.Client/Documents/Operations/ETL/EtlProcessState.cs
arekpalinski: OK. Still a comment explaining why is this collection used for would be very nice.

src/Raven.Server/Documents/ETL/EtlProcess.cs
arekpalinski: Let's leave it as is then.

test/SlowTests/Server/Documents/ETL/ClusterEtlTimeSeriesTests.cs
```diff
@@ -0,0 +1 @@
...
            
            await AssertWaitForNotNullAsync(async () =>
            {
                using var session = dest.OpenAsyncSession();
                var timeSeriesEntries = await session.TimeSeriesFor(tsOwnerId, timeSeriesName).GetAsync(time, time);
                return timeSeriesEntries?.FirstOrDefault();
            }, interval: 1000);
        }
    }
}
```
arekpalinski: Something wrong happened here - everything is in the single line

test/SlowTests/Server/Documents/ETL/EtlTestBase.cs
```diff
@@ -78,5 +100,94 @@ protected ManualResetEventSlim WaitForEtl(DocumentStore store, Func<string, EtlP
...
+            });
+        }
+
+        public override void Dispose()
+        {
+            try
+            {
+                if (_src != null)
+                {
+                    var b = Context.TestException;
```
arekpalinski: I don't see any usage of `b`

test/SlowTests/Server/Documents/ETL/EtlTestBase.cs
```diff
@@ -78,5 +100,94 @@ protected ManualResetEventSlim WaitForEtl(DocumentStore store, Func<string, EtlP
...
+                    var notifications = GetEtlErrorNotifications(_src).Result;
+
+                    var a = Context.TestOutput?.ToString();
```
arekpalinski: Same for `a`

test/SlowTests/Server/Documents/ETL/EtlTestBase.cs
```diff
@@ -78,5 +100,94 @@ protected ManualResetEventSlim WaitForEtl(DocumentStore store, Func<string, EtlP
...
+                        Context.TestOutput.WriteLine(message);
+                    }
+                }
+            }
+            catch
+            {
+                // ignored
+            }
+
+            base.Dispose();
```
arekpalinski: I would put it in `finally`. Effectively it's the same thing in current impl but to underline that his needs to be _always_ called

arekpalinski (approved):

PR #10662, opened 2020-06-07
RavenDB-15027 Display number of entries for time series

Danielle9897:


src/Raven.Studio/typescript/models/database/timeSeries/timeSeriesModel.ts
```diff
@@ -36,7 +36,13 @@ class timeSeriesModel {
         });
         
         this.timestamp.extend({
-            required: true
+            required: true,
+            validation: [
+                {
+                    validator: () => this.timestamp().utc().format() !== "Invalid date",
```
ml054: please never relay on magic string values. What if moment add locale support or we enable that? Code won't work. 

Looks like we can use this:
image

src/Raven.Studio/typescript/viewmodels/database/timeSeries/editTimeSeriesEntry.ts
```diff
@@ -48,7 +48,10 @@ class editTimeSeriesEntry extends dialogViewModelBase {
         this.dateFormattedAsUtc = ko.pureComputed(() => {
             if (model.timestamp()) {
                 const date = moment(model.timestamp());
-                return date.utc().format(editTimeSeriesEntry.utcTimeFormat) + "Z (UTC)";    
+                if (date.utc().format() === "Invalid date") {
```
ml054: same here

ml054 (changes requested):

src/Raven.Studio/typescript/models/database/timeSeries/timeSeriesModel.ts
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/database/timeSeries/editTimeSeriesEntry.ts
Danielle9897: done

ml054 (approved):

ml054 (approved):

PR #10663, opened 2020-06-08
RavenDB-10703 - allow to restore a database using a custom S3 URL

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-10703

### DIFF
--- a/src/Raven.Studio/typescript/models/database/tasks/periodicBackup/s3Settings.ts
+++ b/src/Raven.Studio/typescript/models/database/tasks/periodicBackup/s3Settings.ts
@@ -6,7 +6,9 @@ class s3Settings extends amazonSettings {
     bucketName = ko.observable<string>();
     useCustomS3Host = ko.observable<boolean>();
     customServerUrl = ko.observable<string>();    
-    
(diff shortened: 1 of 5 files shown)


src/Raven.Studio/typescript/models/database/tasks/periodicBackup/s3Settings.ts
```diff
@@ -35,6 +37,14 @@ class s3Settings extends amazonSettings {
                 this.testConnectionResult(null);
             }
         });
+
+        this.accessKeyPropertyName = ko.pureComputed(() => this.isBackBlaze() ? "Application Key ID" : "Access key");
+        this.secretKeyPropertyName = ko.pureComputed(() => this.isBackBlaze() ? "Application Key" : "Secret key");
+    }
+
+    private isBackBlaze() {
```
ml054: can we move this function to single location and make static? Right now we have this logic in 2 places. I'm affraid we forgot to change second place during any update.

grisha-kotler: done

ml054 (approved):

PR #10666, opened 2020-06-08
RavenDB-15123 when compacting indexes first optimize the lucene

ppekrol:


ppekrol:
This can take a lot of time, but sometimes it can drastically reduce the size. In my case it was 1.5GB before and 32MB afterwards.

ayende (commented):
Do we have any way to report progress here?

ppekrol:
Not without changing the Lucene.

ayende (approved):

PR #10676, opened 2020-06-09
RavenDB-15077 By default collapse database record values

ml054:


src/Raven.Studio/typescript/viewmodels/database/settings/databaseRecord.ts
```diff
@@ -68,6 +68,15 @@ class databaseRecord extends viewModelBase {
         const editorElement = $("#dbDocEditor");
         if (editorElement.length > 0) {
             this.docEditor = ko.utils.domData.get(editorElement[0], "aceEditor");
+
+            let alreadyFolded = false;
+            
+            this.docEditor.getSession().on("tokenizerUpdate", () => {
+                if (!alreadyFolded) {
+                    this.foldAll();
```
Danielle9897, line 75: 1. I suggest calling `this.toggleAutoCollapse()` instead of `foldAll `since it will sync the _Toggle Collapse_ button
2. Do we want to do this also after clicking 'Refresh' button ?

src/Raven.Studio/typescript/viewmodels/database/settings/databaseRecord.ts
```diff
@@ -64,16 +65,30 @@ class databaseRecord extends viewModelBase {
 
     compositionComplete() {
         super.compositionComplete();
+        
+        this.docEditor = aceEditorBindingHandler.getEditorBySelection($("#dbDocEditor"));
+        
+        if (this.docEditor) {
+            let alreadyFolded = false;
```
ml054: I wonder if can we can have single field: forceFold instead of refreshWasDone and alreadyFolded? 

then we init this value with true and after refresh we also set this to true.

src/Raven.Studio/typescript/viewmodels/database/settings/databaseRecord.ts
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/database/settings/databaseRecord.ts
Danielle9897: done

ml054:
I can't approve my PR, but I'm approving it. ;-)

PR #10678, opened 2020-06-10
RavenDB-15131 send `null` for min/max DateTime

karmeli87:


src/Raven.Client/Documents/Operations/TimeSeries/TimeSeriesRangeResult.cs
```diff
@@ -1,14 +1,35 @@
...
+using System.Runtime.Serialization;
 using Raven.Client.Documents.Session.TimeSeries;
+using Sparrow.Json;
 
 namespace Raven.Client.Documents.Operations.TimeSeries
 {
-    public class TimeSeriesRangeResult
+    public class TimeSeriesRangeResult : IPostJsonDeserialization
     {
         public DateTime From, To;
```
ppekrol, line 10: Wouldn't it be better to have nullable here instead of all those checks in the TimeSeriesHandler?

karmeli87: I actually don't want it to be `nullable` ... I just want to send it over the wire as `null` for the sack of other clients.

PR #10690, opened 2020-06-15
RavenDB-14688 add test execution snapshot, enable snapshots in Benchm…

topalach:
…arkTests


topalach:
`TestStage.Delta` can be safely removed because it's not used in the Metrics Dispatcher application. So it was displaced by the `TestExecution` stage.

PR #10691, opened 2020-06-15
RavenDB-14696: Don't redirect subscription to AppropriateNode before validating that the Subscription actually was created on it.

garayx:


grisha-kotler (approved):

src/Raven.Server/Documents/TcpHandlers/SubscriptionConnection.cs
```diff
@@ -353,6 +358,25 @@ private static async Task ReportExceptionToClient(SubscriptionConnection connect
                 }
                 else if (ex is SubscriptionDoesNotBelongToNodeException subscriptionDoesNotBelongException)
                 {
+                    try
+                    {
+                        using (server.ContextPool.AllocateOperationContext(out TransactionOperationContext ctx))
+                        using (ctx.OpenReadTransaction())
+                        {
+                            // check that the subscription exists on AppropriateNode
```
karmeli87, line 366: How did we end-up in this situation?

When we create the subscription we wait for it on the appropriate node:
```csharp
var node = Database.SubscriptionStorage.GetResponsibleNode(serverContext, name);
if (node != null && node != ServerStore.NodeTag)
{
        await WaitForExecutionOnSpecificNode(serverContext, ServerStore.GetClusterTopology(serverContext), node, subscriptionId);
}
```
But I don't see where we wait if `node == ServerStore.NodeTag`, so this is the issue?

karmeli87: hmm we do wait here:
`var subscriptionId = await Database.SubscriptionStorage.PutSubscription(options, GetRaftRequestIdFromQuery(), id, disabled, mentor);`

But what might actually happened is that `GetResponsibleNode` will return `null` if we went to `Candidate` state.
Since this is a new subscription,  we should do `return topology.WhoseTaskIsIt(RachisState.Follower, subscription, null);` inside the `GetResponsibleNode` function.

grisha-kotler: What if the responsible node changed after we checked the current responsible node?

karmeli87: That might happened as well, yes.

karmeli87: So why not to wait upon the tcp connection to the responsible node?

garayx: > But what might actually happened is that `GetResponsibleNode` will return `null` if we went to `Candidate` state.
> Since this is a new subscription, we should do `return topology.WhoseTaskIsIt(RachisState.Follower, subscription, null);` inside the `GetResponsibleNode` function.

thats what actually happens in GetResponsibleNode, we return WhoseTaskIsIt()

garayx: > So why not to wait upon the tcp connection to the responsible node?

there is a tcp connection to responsible node but the subscription task is not yet created on it so we get subscrption does not exist (https://issues.hibernatingrhinos.com/issue/RavenDB-14696#focus=streamItem-67-48264.0-0)

karmeli87: Actually we do wait for it at `AssertSubscriptionConnectionDetails`.. :/
> How did we end-up in this situation?

garayx: 5 nodes cluster, the subscription was created (on majority of the cluster, ABC), and returned responsible node, this node became down so the SubscriptionDoesNotBelongToNodeException thrown and we got a new AppropriateNode which could be D or E, and the task is not yet there so we get subscrption does not exist exception

karmeli87 (approved):

PR #10700, opened 2020-06-18
RavenDB-15208 delete index errors endpoint should not be a part of debug package

ppekrol:


src/Raven.Server/Documents/Handlers/IndexHandler.cs
```diff
@@ -666,7 +666,7 @@ public async Task SetPriority()
             }
         }
 
-        [RavenAction("/databases/*/indexes/errors", "DELETE", AuthorizationStatus.ValidUser, IsDebugInformationEndpoint = true)]
```
ayende, line 669: I don't like this change.

ayende (commented):
We should be fixing LocalEndpointClient, not this.

src/Raven.Server/Documents/Handlers/IndexHandler.cs
ppekrol: This was clearly a bug. It is a 'DELETE' endpoint not a 'GET' one. Copy/paste issue.

PR #10704, opened 2020-06-18
RavenDB-15176 - Backup allowed destination should be "GoogleCloud" without space

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-15176


ppekrol:
@ml054 @Danielle9897 @grisha-kotler Is Studio sending corrct value when Google Cloud backups is set-up?

grisha-kotler:
Yes. `GetDestinations` is only for internal use and displaying them in the studio (view only).

ppekrol:
What do you mean 'displaying'? We do not want to display 'GoogleCloud'.

grisha-kotler:
Fixed

PR #10707, opened 2020-06-18
RavenDB-15080 : Changes in counters storage

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-15080

- property names inside Counter Group values document are now lower-cased
- added a property inside Counter Group document which keeps the counter names in their original casing 
- some changes in incoming replication - deiffrent treatment to 4.2 counters (and their documents) 
- various bug fixes

### DIFF
--- a/src/Raven.Server/Documents/CountersStorage.cs
+++ b/src/Raven.Server/Documents/CountersStorage.cs
@@ -1,7 +1,10 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
+using System.Linq;
 using System.Runtime.InteropServices;
 using System.Text;
 using Raven.Client.Documents.Changes;
@@ -40,6 +43,7 @@ namespace Raven.Server.Documents
 
         public const string DbIds = "@dbIds";
         public const string Values = "@vals";
+        public const string CounterNames = "@names";
 
         private readonly List<ByteStringContext<ByteStringMemoryCache>.InternalScope> _counterModificationMemoryScopes =
             new List<ByteStringContext<ByteStringMemoryCache>.InternalScope>();
@@ -310,9 +314,14 @@ namespace Raven.Server.Documents
                             throw new InvalidDataException($"Counter-Group document '{counterKeySlice}' is missing '{Values}' property. Shouldn't happen");
                         }
 
+                        if (data.TryGet(CounterNames, out BlittableJsonReaderArray originalNames) == false)
+                        {
+                            throw new InvalidDataException($"Counter-Group document '{counterKeySlice}' is missing '{CounterNames}' property. Shouldn't happen");
+                        }
+
                         var counterEtag = _documentsStorage.GenerateNextEtag();
 
-                        counters.TryGetMember(lowerName, out object existingCounter); //todo aviv 
+                        counters.TryGetMember(lowerName, out object existingCounter);
 
                         if (existingCounter is BlittableJsonReaderObject.RawBlob blob &&
                             overrideExisting == false)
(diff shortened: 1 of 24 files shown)


ppekrol (approved):

ppekrol:
There are still few failing tests, but nothing major.

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -629,18 +713,43 @@ private static BlittableJsonReaderObject CreateHalfDocument(DocumentsOperationCo
                         builder.WriteValue(lsv);
                     else
                         throw new InvalidDataException("Unknown type: " + prop.Token + " when trying to split counter doc");
+
+                    propNames.Add(prop.Name);
                 }
 
                 builder.WriteObjectEnd();
 
+                builder.WritePropertyName(CounterNames);
```
ayende, line 717: Why is this an array and not a dictionary? That would allow O(1) searches, no?

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -677,12 +786,42 @@ private struct PutCountersData
...
+                        if (sourceCounterNames == null)
+                        {
+                            // 4.2 source
+                            // need to create @names array and to lower the property names in 'sourceCounters'
+
+                            var originalNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+                            var propDetails = new BlittableJsonReaderObject.PropertyDetails();
+                            sourceCounters.Modifications = new DynamicJsonValue(sourceCounters);
+
+                            for (int i = 0; i < sourceCounters.Count; i++)
```
ayende, line 797: Here to, if this was an array, it would be easier to run though.
Note that we'll usually not trigger this work, so we want to avoid those costs if possible.

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -995,6 +1167,57 @@ private void UpdateMetricsForNewCounterGroup(BlittableJsonReaderObject data)
...
+                    localCounterNames.Modifications.Add(propName);
+                }
+            }
+        }
+
+        private static LazyStringValue GetOriginalName(BlittableJsonReaderArray counterNames, string counterName)
+        {
+            LazyStringValue current = null;
+            int j;
+            for (j = 0; j < counterNames.Length; j++)
```
ayende: We can't do that scanning all the time

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -1414,6 +1647,18 @@ public string DeleteCounter(DocumentsOperationContext context, string documentId
             }
         }
 
+        private static int GetCounterNameIndex(BlittableJsonReaderArray originalNames, string lowered)
```
ayende: where is this called?

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -1995,6 +2266,35 @@ LazyStringValue ToLuceneKey(LazyStringValue documentId, LazyStringValue counterN
...
+        public static int BinarySearch(List<object> list, object key, StringComparison comparison)
+        {
+            int min = 0;
+            int max = list.Count - 1;
+
+            while (min <= max)
+            {
+                int mid = (min + max) >> 1;
+                var current = list[mid];
+                var result = string.Compare(key.ToString(), current.ToString(), comparison);
```
ayende: Why are we calling ToString() here?

ayende (changes requested):

src/Raven.Server/Documents/CountersStorage.cs
aviv86: changed from BlittableArray to BlittableObject

src/Raven.Server/Documents/CountersStorage.cs
aviv86: unused code

src/Raven.Server/Documents/CountersStorage.cs
aviv86: no one, removed this

karmeli87 (approved):

ppekrol (approved):

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -1166,55 +1156,60 @@ private void UpdateMetricsForNewCounterGroup(BlittableJsonReaderObject data)
...
                 }
             }
         }
 
-        private static LazyStringValue GetOriginalName(BlittableJsonReaderArray counterNames, string counterName)
+        private static LazyStringValue GetOriginalName(BlittableJsonReaderObject counterNames, string counterName)
+        {
+            var lower = counterName.ToLower();
+            if (counterNames.TryGet(lower, out object originalName) == false)
+                return null;
```
ayende: shouldn't we return `counterName` here instead of null?

ayende (approved):

PR #10712, opened 2020-06-19
RavenDB-15227

ppekrol:


src/Raven.Client/Documents/Conventions/DocumentConventions.cs
```diff
@@ -385,15 +385,29 @@ public int LoadBalancerContextSeed
 
         public LoadBalanceBehavior LoadBalanceBehavior
         {
-            // We have to make this check so if admin activated this, but client code did not provide the selector,
```
ppekrol, line 388: @ayende I do not think this is correct (the comment) ? IIRC we will use session counter as the key then?

ayende: No, this is correct. The session key is used only if the `LoadBalanceBehavior.UseSessionContext` is set.
That was done _explicitly_ to only opt in from the client.

Otherwise, each session will use a different node and you won't have the usual mechanics of one primary normally.

ayende (changes requested):
We should't change this behavior

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
ppekrol: This is exactly what I would expect after setting Load Balance Behavior to UseSessionContext. When context is not set (either via Selector or Manually) then each session should just use the counter. This is also a form of Load Balancing.

ayende: That will lead to strange errors. A write goes to one node and a read to the other.
That leads to support calls :-)

I would rather this be very explicit. An application must _choose_ to enable this.

ayende: And an admin setting is not enough, you may have multiple applications running.

ppekrol: What do you mean write goes to one and read goes to other? It will stick to the same node during the session?

Why you want to be so explicit? What is the difference between this and ReadBalanceBehavior? If we want to be so explicit, then I do not see a reason of having LoadBalanceBehavior in the Client Configuration.

ayende: Separate requests. PRG is a good example.
Two sessions, but same user seeing that, and data goes poof

ppekrol: So the only usage in the ClientConfiguration for this is to turn it off? Do we want that or just Seed is enough?

ayende: Yes

ppekrol:
@ayende can you review? I've applied all the requirements.

src/Raven.Client/Documents/Session/SessionInfo.cs
```diff
@@ -10,53 +10,62 @@ public class SessionInfo
...
+        internal SessionInfo(InMemoryDocumentSessionOperations session, SessionOptions options, DocumentStoreBase documentStore, bool asyncCommandRunning)
         {
-            SetContext(sessionKey, loadBalancerContextSeed);
+            if (documentStore is null)
+                throw new ArgumentNullException(nameof(documentStore));
 
-            _loadBalancerContextSeed = loadBalancerContextSeed;
+            _session = session ?? throw new ArgumentNullException(nameof(session));
+            _loadBalancerContextSeed = session.RequestExecutor.Conventions.LoadBalancerContextSeed;
+            _loadBalancerContextSelectorSet = session.Conventions.LoadBalanceBehavior == LoadBalanceBehavior.UseSessionContext && session.Conventions.LoadBalancerPerSessionContextSelector != null;
```
ayende: I think this is wrong. If the admin explicitly turned it off, even if you provide `SetContext`, it should not apply it.

src/Raven.Client/Documents/Session/SessionInfo.cs
```diff
@@ -10,53 +10,62 @@ public class SessionInfo
...
                 _sessionIdUsed = true;
-                return _clientSessionId;
+
+                return _sessionId.Value;
             }
         }
 
+        internal bool HasSessionId
+        {
+            get => _sessionId.HasValue || _loadBalancerContextSelectorSet;
```
ayende: The second time this is called, is will be true, because we set `_sessionId` from `_clientSessionIdCounter`

ayende (changes requested):

src/Raven.Client/Documents/Session/SessionInfo.cs
ppekrol: Done.

ayende (approved):

PR #10715, opened 2020-06-22
RavenDB-15196 fix import time-series when the ts with original name is missing

karmeli87:
We importing TS by the etag, so it might happened that we will import the rollup, before the raw one.

### DIFF
--- a/src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
+++ b/src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
@@ -21,6 +21,7 @@ using Sparrow.Binary;
 using Sparrow.Extensions;
 using Sparrow.Json;
 using Sparrow.Json.Parsing;
+using Sparrow.Logging;
 using Sparrow.Server;
 using Sparrow.Server.Utils;
 using Voron;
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
```diff
@@ -1451,7 +1455,16 @@ public void AddTimeSeriesNameToMetadata(DocumentsOperationContext ctx, string do
...
 
-            tsName = EnsureOriginalName(ctx, docId, tsName);
+            try
+            {
+                tsName = EnsureOriginalName(ctx, docId, tsName);
+            }
+            catch (Exception e)
+            {
+                var error = $"Unable to locate the original time-series '{tsName}' in document '{docId}'";
+                if (_logger.IsInfoEnabled)
```
ayende, line 1465: Is it okay to ignore this error, aside from the log?
What would be the behavior of the system?

karmeli87: If I don't ignore this it might take down the replication or import.
And if we fail we will use the initial `tsName` that we got from the argument.
Worst-case as I see it, is that the metadata might contain wrong casing.

ayende (approved):

PR #10716, opened 2020-06-22
RavenDB-15159 add swap usage to monitoring

gregolsky:
also added a debug dockerfile if somebody wanted to debug linux/docker (really easy to wire up in rider)


src/Raven.Server/Raven.Server.csproj
```diff
@@ -138,6 +138,11 @@
       <CopyToOutputDirectory>Never</CopyToOutputDirectory>
     </None>
   </ItemGroup>
+  <ItemGroup>
+    <Content Include="..\..\debug.Dockerfile">
+      <Link>debug.Dockerfile</Link>
```
ppekrol, line 143: Do we need that?

ppekrol (approved):

src/Raven.Server/Raven.Server.csproj
gregolsky: see PR desc

PR #10725, opened 2020-06-22
RavenDB-15240 processed timeseries (dead segments) and document tombstones should be processed separately

ppekrol:
TODO:
- review ETL
- review Replication
- review Backup


src/Raven.Server/Documents/PeriodicBackup/PeriodicBackupRunner.cs
```diff
@@ -985,8 +985,11 @@ public RunningBackup OnGoingBackup(long taskId)
 
         public string TombstoneCleanerIdentifier => "Periodic Backup";
 
-        public Dictionary<string, long> GetLastProcessedTombstonesPerCollection()
+        public Dictionary<string, long> GetLastProcessedTombstonesPerCollection(ITombstoneAware.TombstoneType tombstoneType)
         {
+            if (tombstoneType != ITombstoneAware.TombstoneType.Documents)
```
karmeli87: For replication and backup I don't think you care what type it is. You just need to know the last processed etag overall.
Right now we might delete dead segments without backing them up.

src/Raven.Server/Documents/TombstoneCleaner.cs
```diff
@@ -12,6 +12,12 @@ namespace Raven.Server.Documents
 {
     public class TombstoneCleaner : BackgroundWorkBase
     {
+        private readonly ITombstoneAware.TombstoneType[] _tombstoneTypes = new ITombstoneAware.TombstoneType[]
+        {
+            ITombstoneAware.TombstoneType.Documents,
+            ITombstoneAware.TombstoneType.TimeSeries
```
ayende, line 18: What about counters?

ayende (approved):

src/Raven.Server/Documents/TombstoneCleaner.cs
ppekrol: https://issues.hibernatingrhinos.com/issue/RavenDB-14816 we do not clean them at all

PR #10726, opened 2020-06-23
RavenDB-15191 transparently go to the rollup only for group-by queries

karmeli87:


ppekrol (approved):

src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
```diff
@@ -243,17 +243,10 @@ public async Task Read()
...
         }
 
-        internal static TimeSeriesRangeResult GetTimeSeriesRange(DocumentsOperationContext context, string docId, string name, DateTime? from, DateTime? to)
-        {
-            int start = 0, pageSize = int.MaxValue;
-            return GetTimeSeriesRange(context, docId, name, from ?? DateTime.MinValue, to ?? DateTime.MaxValue, ref start, ref pageSize);
-        }
-
-        private static unsafe TimeSeriesRangeResult GetTimeSeriesRange(DocumentsOperationContext context, string docId, string name, DateTime from, DateTime to, ref int start, ref int pageSize)
+        internal static unsafe TimeSeriesRangeResult GetTimeSeriesRange(DocumentsOperationContext context, string docId, string name, DateTime from, DateTime to, ref int start, ref int pageSize)
```
ayende, line 246: Why ref parameters here? Doesn't seem to be used for anything

ayende (approved):

src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
karmeli87: It's been used in `GetTimeSeriesRangeResults`, not sure if correctly though..

ayende: It isn't, and we should probably remove it.

karmeli87: @aviv86, I remember we have this because:
If ask page size 10 and 100 time-series I would get only the total first 10, and not 10x100, right?

karmeli87: anyway, let's do it in a separate issue:
https://issues.hibernatingrhinos.com/issue/RavenDB-15246

PR #10729, opened 2020-06-23
RDBCL-1478: Rearrange the backup timer if we got cluster down status …

garayx:
…on first backup

### DIFF
--- a/src/Raven.Server/Documents/PeriodicBackup/PeriodicBackupRunner.cs
+++ b/src/Raven.Server/Documents/PeriodicBackup/PeriodicBackupRunner.cs
@@ -288,7 +288,7 @@ namespace Raven.Server.Documents.PeriodicBackup
 
                 var backupDetails = (NextBackup)backupTaskDetails;
 
-                if (ShouldRunBackupAfterTimerCallback(backupDetails, out PeriodicBackup periodicBackup) == false)
+                if (ShouldRunBackupAfterTimerCallbackAndRescheduleIfNeeded(backupDetails, out PeriodicBackup periodicBackup) == false)
                     return;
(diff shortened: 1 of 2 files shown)


src/Raven.Server/Documents/PeriodicBackup/PeriodicBackupRunner.cs
```diff
@@ -644,6 +647,31 @@ private bool ShouldRunBackupAfterTimerCallback(NextBackup backupInfo, out Period
...
+            string msg;
+            switch (taskStatus)
+            {
+                case TaskStatus.ActiveByCurrentNode:
+                    msg = $"Backup {backupInfo.TaskId}, current status is {taskStatus}, the backup will be executed on current node.";
+                    break;
+                case TaskStatus.ClusterDown:
+                    msg = $"Backup {backupInfo.TaskId}, current status is {taskStatus}, the backup will be rescheduled on current node.";
+                    var status = GetBackupStatus(backupInfo.TaskId, periodicBackup.BackupStatus);
+                    periodicBackup.UpdateTimer(GetTimer(periodicBackup.Configuration, status));
```
grisha-kotler, line 665: if the cluster is down it will cause an infinite loop

grisha-kotler (changes requested):

src/Raven.Server/Documents/PeriodicBackup/PeriodicBackupRunner.cs
garayx: it will create a timer for next backup occurrence

grisha-kotler (approved):

PR #10730, opened 2020-06-23
RavenDB-15158 Improving the performance of database configuration ini…

arekpalinski:
…tialization by reducing the number of reflections calls since they can be called just once


ayende (approved):
Question, how much of an impact are we seeing from this?
Isn't that something that only happens once?

arekpalinski:
The actual numbers are here:
https://issues.hibernatingrhinos.com/issue/RavenDB-15158#focus=streamItem-67-48550.0-0

That was the issue - we were calling a lot of reflection on each database configuration initialization while we could do it just once - and this is how it works now.

ayende:
Thanks

PR #10733, opened 2020-06-23
RavenDB-14232: Better error when failed to update expired license on …

garayx:
…registration

### DIFF
--- a/src/Raven.Server/Commercial/LicenseManager.cs
+++ b/src/Raven.Server/Commercial/LicenseManager.cs
@@ -544,7 +544,10 @@ namespace Raven.Server.Commercial
                     }
                     catch (Exception e)
                     {
-                        throw new LicenseExpiredException($"License already expired on: {newLicenseStatus.Expiration}, and we were unable to get an updated license. Please make sure you that you have access to {ApiHttpClient.ApiRavenDbNet}.", e);
+                        if (e is HttpRequestException)
+                            throw new LicenseExpiredException($"License already expired on: {newLicenseStatus.Expiration}, and we were unable to get an updated license. Please make sure you that you have access to {ApiHttpClient.ApiRavenDbNet}.", e);
+                        
+                        throw new LicenseExpiredException($"License already expired on: {newLicenseStatus.Expiration}, and we were unable to get an updated license.", e);
                     }
                     if (license != null)
                     {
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Commercial/LicenseManager.cs
```diff
@@ -537,9 +537,15 @@ private async Task<bool> ContinueActivatingLicense(License license, bool skipLea
...
                     SetLicense(license.Id, newLicenseStatus.Attributes);
-
-                    // license expired, we'll try to update it
-                    license = await GetUpdatedLicenseInternal(license);
+                    try
+                    {
+                        // license expired, we'll try to update it
+                        license = await GetUpdatedLicenseInternal(license);
+                    }
+                    catch (Exception e)
```
grisha-kotler, line 545: `HttpRequestException`?

src/Raven.Server/Commercial/LicenseManager.cs
```diff
@@ -537,9 +537,18 @@ private async Task<bool> ContinueActivatingLicense(License license, bool skipLea
...
+                    {
+                        if (e is HttpRequestException)
```
ayende, line 547: Include IOException / SocketException here as well?

ppekrol: Probably not needed? https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.sendasync?view=netcore-3.1#System_Net_Http_HttpClient_SendAsync_System_Net_Http_HttpRequestMessage_ but maybe we should include `TaskCanceledException` ?

ayende: Absolutely `TaskCanceledException`, but...

```
var client = new HttpClient();
client.GetAsync("http://e.example.com").GetAwaiter().GetResult();
```

Gives:

```
Unhandled exception. System.Net.Http.HttpRequestException: No such host is known.
 ---> System.Net.Sockets.SocketException (11001): No such host is known.
   at System.Net.Http.ConnectHelper.ConnectAsync(String host, Int32 port, CancellationToken cancellationToken)
   --- End of inner exception stack trace ---
```

ppekrol: Yes, but the 'SocketException' is an inner ex of HttpRequestException?

ayende: Hm... that is an inner exception, never mind.

garayx: Do we want such deep error handling ? It still will throw a LicenseExpiredException with inner exception if it is not a HttpRequestException

ppekrol: @garayx what you did is fine, I think

PR #10744, opened 2020-06-25
RavenDB-15198: merge bulk insert commands when there are a lot of ope…

garayx:
…rations

### DIFF
--- a/src/Raven.Server/Documents/Handlers/BulkInsertHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/BulkInsertHandler.cs
@@ -58,6 +58,7 @@ namespace Raven.Server.Documents.Handlers
                             var array = new BatchRequestParser.CommandData[8];
                             var numberOfCommands = 0;
                             long totalSize = 0;
+                            int operationsCount = 0;
                             while (true)
                             {
                                 using (var modifier = new BlittableMetadataModifier(docsCtx))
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/Handlers/BulkInsertHandler.cs
```diff
@@ -71,7 +72,7 @@ private async Task<IOperationResult> DoBulkInsert(Action<IOperationProgress> onP
                                     // if we are going to wait on the network, flush immediately
                                     if ((task.Wait(5) == false && numberOfCommands > 0) ||
                                         // but don't batch too much anyway
-                                        totalSize > 16 * Voron.Global.Constants.Size.Megabyte)
+                                        totalSize > 16 * Voron.Global.Constants.Size.Megabyte || operationsCount >= 8192)
```
ppekrol, line 75: Do we have some numbers for this? What if this is 4/8/16/32/64/128k ?

test/BenchmarkTests/Storing/BulkInsert.cs
```diff
@@ -37,6 +38,51 @@ public async Task Large_Size_1M()
             await Large_Size_Internal(1_000_000);
         }
 
+        [Fact]
+        public async Task VeryBigTransactionsForBulkInsertOfCounters()
```
ppekrol: Meaningful names here please. `Counters_1M_1Value_1User()` or sth like that.

test/BenchmarkTests/Storing/BulkInsert.cs
```diff
@@ -37,6 +38,51 @@ public async Task Large_Size_1M()
...
+                const string baseDocId = "users/";
+                using (var bulk = store.BulkInsert())
+                {
+                    var docId = baseDocId + 1;
+                    await bulk.StoreAsync(new User() { Name = "Grisha" }, docId);
+
+                    var counters = bulk.CountersFor(docId);
+                    for (var j = 0; j < 1_000_000; j++)
+                    {
+                        var rnd = new Random(DateTime.Now.Millisecond);
```
ppekrol: no randoms. needs to be the same every day.

src/Raven.Server/Documents/Handlers/BulkInsertHandler.cs
garayx: yes I tested bulk insert of counter, ts, docs, for various values (1k to 20k), 8k gives the fastest result

PR #10751, opened 2020-06-28
RavenDB-15223 : Create schema upgrade from 4.2 to handle changes in counters storage

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-15223


src/Raven.Server/Storage/Schema/Updates/Documents/42017/From41016.cs
```diff
@@ -619,6 +624,17 @@ private static BlittableJsonReaderObject WriteNewCountersDocument(DocumentsOpera
 
                     builder.WriteObjectEnd();
 
+                    builder.WritePropertyName(CountersStorage.CounterNames);
```
karmeli87: Those methods are also used to upgrade from 41_016, would it have any effect on some-one upgrading from earlier version (let's say 4.1.x) to 5.0?

src/Raven.Server/Storage/Schema/Updates/Documents/50000/From42017.cs
```diff
@@ -1,16 +1,493 @@
...
+                readTable = new Table(CountersSchema, step.ReadTx);
+
+                var processedInCurrentTx = 0;
+
+                using (step.DocumentsStorage.ContextPool.AllocateOperationContext(out DocumentsOperationContext context))
+                {
+                    var commit = false;
+
+                    foreach (var counterGroup in GetCounters(readTable, context, currentPrefix))
+                    {
```
karmeli87, line 82: We recreate here all counters, regardless of their casing?

src/Raven.Server/Storage/Schema/Updates/Documents/50000/From42017.cs
```diff
@@ -1,16 +1,493 @@
...
+                    table.DeleteByIndex(CountersSchema.FixedSizeIndexes[CollectionCountersEtagsSlice], etag);
+                }
+            }
+            
+            toDeleteByCollection.Clear();
+        }
+
+        private static IEnumerable<CounterReplicationItem> GetCounters(Table table, DocumentsOperationContext ctx, string prefix)
+        {
+            if (prefix == null)
```
karmeli87, line 172: Why do we need the prefix?
Wouldn't we simply continue to the next document? only after we consumed all counters of the previous one?

src/Raven.Server/Storage/Schema/Updates/Documents/50000/From42017.cs
```diff
@@ -1,16 +1,493 @@
...
+                    //all entries that start after prefix
+                    yield return CreateReplicationBatchItem(ctx, ref result.Value.Reader);
+                }
+            }
+        }
+
+        private readonly ObjectPool<Dictionary<LazyStringValue, PutCountersData>> _dictionariesPool
+            = new ObjectPool<Dictionary<LazyStringValue, PutCountersData>>(() => new Dictionary<LazyStringValue, PutCountersData>(LazyStringValueComparer.Instance));
+
+        private unsafe void PutCounters(DocumentsOperationContext context, UpdateStep step, string documentId, string changeVector, BlittableJsonReaderObject sourceData, string collection)
```
karmeli87, line 201: This looks very similar to the `PutCounters` of the `CounterStorage`, I would prefer to reuse _if possible_.

karmeli87: same for the rest of the methods here

src/Raven.Server/Storage/Schema/Updates/Documents/42017/From41016.cs
aviv86: The only method that is shared between `From41016` and `From41017` is `ReadDbId()`
And yes, I intoduced some changes in `From41016` and upgrading from 4.1.x to 5.0 now works 
and is tested - unskipped `CanMigrateLegacyCounters`, `CanMigrateLegacyCountersWithMultipleDbIds`, `CanMigrateTablesWithCounterWord` tests

src/Raven.Server/Storage/Schema/Updates/Documents/50000/From42017.cs
aviv86: When we have `commit = true` we are breaking from `GetCounters()` foreach loop.
then when we return to GetCounters loop we want to continue from were we left off. 
without the prefix we will start from the beginning again..

src/Raven.Server/Storage/Schema/Updates/Documents/50000/From42017.cs
aviv86: Yes, we delete all `CounterGroup` entries and recreate them all from scratch (per document)
We can't simply lower-case the counter names (and keys) in each CounterGroup entry, 
because lower-casing a counter name changes it's lexical value and can affect the total ordering between CounterGroup entries.

e.g. say we have 2 CounterGroup entries :
- [A, B, X, Y] (key = "users/1|")
- [c, d, e, f] (key = "users/1|c")

if we were to lower the casing in each group we will get - 
- [a, b, x, y] (key = "users/1|")
- [c, d, e, f] (key = "users/1|c")

And now we are in a bad state - CounterGroup entry "users/1|" comes before "users/1|c" entry,
but "users/1|" entry contains counter names with lexical value that is greater than the first counter in the next entry ("x" and "y" are greater than "c").
Now for example if we want to get value of counter "x", we will search for it in entry "users/1|c" (because that's where we think it's supposed to be), won't find it there and return null.. 
Similar problem with increment

src/Raven.Server/Storage/Schema/Updates/Documents/50000/From42017.cs
aviv86: ok will try to reuse code from CountersStorage where possible

test/SlowTests/SlowTests.csproj
```diff
@@ -32,6 +32,8 @@
     <None Remove="Data\RavenDB_13759\Orders_ByOrderBy.zip" />
     <None Remove="Data\RavenDB_13940\storage-with-reused-journal-and-synced-data-2.zip" />
     <None Remove="Data\RavenDB_13940\storage-with-reused-journal-and-synced-data.zip" />
+    <None Remove="Data\RavenDB_15223\counters.4.2.103.ravendb-snapshot" />
```
ppekrol, line 35: This file has 40MB. If you need to create this anyway, then we can use Interversion tests to generate this on the fly.

src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -748,21 +751,26 @@ private class PutCountersData
             public ByteStringContext.InternalScope KeyScope;
         }
 
-        public bool PutCounters(DocumentsOperationContext context, string documentId, string collection, string changeVector,
-            BlittableJsonReaderObject sourceData)
+        public bool PutCounters(DocumentsOperationContext context, string documentId, string collection, string changeVector, BlittableJsonReaderObject sourceData,
```
ppekrol: I do not think this is a good idea. This is error prone. Consider in the future we will add additional items here that your current and some future schema upgrades should not ever encounter.

I think it would be better to copy the method and avoid using it, especially when changes are required internally to make it actually work.

src/Raven.Server/Storage/Schema/Updates/Documents/42017/From41016.cs
ppekrol: But it means that when 42017.cs is executed and migration was done from 41016, it will already produce the CounterNames property. Is this something that we would want? It means that 42017.cs will encounter them, but it shouldn't?

aviv86: reverted changed in `From41016`, redundant work

test/SlowTests/SlowTests.csproj
aviv86: I accidentally used a snapshot that was creadted with no compression. 
changed to optimal compression and now this file is 300KB

src/Raven.Server/Documents/CountersStorage.cs
aviv86: revereted changes in CountersStorage

ppekrol (approved):

PR #10752, opened 2020-06-28
RavenDB-14369 Restore from cloud backup causes the new db to be encry…

Danielle9897:
…pted even if it wasn't originally

### DIFF
--- a/src/Raven.Studio/typescript/models/resources/creation/databaseCreationModel.ts
+++ b/src/Raven.Studio/typescript/models/resources/creation/databaseCreationModel.ts
@@ -261,7 +261,7 @@ class databaseCreationModel {
                         } else {
                             // encrypted backup - we need license and https for encrypted db
                             
-                            encryptionSection.enabled(canCreateEncryptedDbs);
+                            encryptionSection.enabled(false);
                             encryptionSection.disableToggle(!canCreateEncryptedDbs);
                         }
                     } else { //backup is not encrypted


src/Raven.Studio/typescript/models/resources/creation/databaseCreationModel.ts
```diff
@@ -261,7 +261,7 @@ class databaseCreationModel {
                         } else {
                             // encrypted backup - we need license and https for encrypted db
                             
-                            encryptionSection.enabled(canCreateEncryptedDbs);
+                            encryptionSection.enabled(false);
```
ml054, line 264: according to comment + logic, 

if we restore encrypted backup we default to encrypted database. So I'm not sure putting false here is good

Danielle9897: It is possible to create/restore a **non-encrypted** **database** from an **encrypted**-**backup**.

@iftahbe was requesting that the toggle will not be 'enabled' automatically. 
Instead, allow the user to turn on the toggle if needed.

Maybe the comment should be refactored.

ayende: User _may_ chose to set it up as non encrypted, but it should be an explicit choice.

iftahbe: @ayende, @ml054 why if we restore an encrypted backup we should default to an encrypted database?
The cloud backup is always encrypted, even if the original DB was not encrypted.

When the user creates a new DB from backup, they should be able to choose if DB is encrypted or not.
So the toggle should default to false, and if the user wants an encrypted DB they should turn it on.

ml054: to sustain security level by default - with an option to lower this to non-encrypted database. Safe by default.

iftahbe: I disagree.
Turning on Encryption has a significant impact on performance and should only be done by people who explicitly want it.
Also, there is no way to disable encryption afterward, and it requires to create a whole new database.

ml054: Please also notice that you create encrypted database to hold sensitive data. Encrypted cannot be enabled on http. 
So it means someone imports sensitive data to https enabled server (which isn't probably dev). 

Maybe we should emphase that database is encrypted on UI? For me security is more important then performance. And we still can disable encryption during import.

iftahbe: Maybe this argument is pointless because it's just about a toggle of a button, but I really think that encryption at rest should be turned off by default. 
Just because the server is https, doesn't mean that the user needs encryption. It is actually meant for people who are afraid their physical disk will be compromised. I think it's a very small number of users. 
And for cloud instances, encryption performance is very relevant. It consumes a lot of CPU credits.

ayende: When we say encrypted when the backup is encrypted, the user still needs to selection encryption key
That is another step that they have to take to ensure that they aren't going to create encrypted by mistake.

Also, I don't think we saw meaningful performance difference with encryption, @iftahbe  ?

gregolsky: When creating a database encryption is optional. The default is not encrypted.
To be consistent we should follow the same convention in the restore.

iftahbe: We did see a little performance difference, more CPU usage. I don't have the numbers. I probably exaggerated.

Danielle9897: **Closing** PR means leaving current behavior:
https: button is enabled and turned 'On'
http: button is disabled

**Merging** PR means:
https: button is enabled but turned 'Off'
http: button is disabled

So should I close this PR or Not?

iftahbe: Close the PR

PR #10755, opened 2020-06-29
RavenDB-15069 Better error when using features from a grater major version

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15069

### DIFF
--- a/src/Raven.Server/Documents/Handlers/BatchRequestParser.cs
+++ b/src/Raven.Server/Documents/Handlers/BatchRequestParser.cs
@@ -15,6 +15,7 @@ using Raven.Client.Extensions;
 using Raven.Client.Http;
 using Raven.Client.Util;
 using Raven.Server.Documents.Patch;
+using Raven.Server.Exceptions;
 using Raven.Server.ServerWide;
 using Sparrow;
 using Sparrow.Json;
@@ -1012,11 +1013,6 @@ namespace Raven.Server.Documents.Handlers
         private static void ThrowUnexpectedEndOfStream()
         {
             throw new EndOfStreamException();
-        }
-
-        public class InvalidCommandTypeException : Exception
-        {
-            public InvalidCommandTypeException(string msg): base(msg) { }
         }
     }
 }
--- a/src/Raven.Server/Documents/Handlers/BatchRequestParser.cs
+++ b/src/Raven.Server/Documents/Handlers/BatchRequestParser.cs
@@ -905,40 +905,40 @@ namespace Raven.Server.Documents.Handlers
             switch (state.StringSize)
             {
                 case 3:
-                    if (*(short*)state.StringBuffer != 21840 ||
-                        state.StringBuffer[2] != (byte)'T')
-                        ThrowInvalidProperty(state, ctx);
-                    return CommandType.PUT;
+                    if (*(short*)state.StringBuffer == 21840 && 
+                        state.StringBuffer[2] == (byte)'T') 
+                        return CommandType.PUT;
+                    break;
 
                 case 5:
-                    if (*(int*)state.StringBuffer != 1129595216 ||
-                        state.StringBuffer[4] != (byte)'H')
-                        ThrowInvalidProperty(state, ctx);
-                    return CommandType.PATCH;
+                    if (*(int*)state.StringBuffer == 1129595216 &&
+                        state.StringBuffer[4] == (byte)'H') 
+                        return CommandType.PATCH;
+                    break;
 
                 case 6:
-                    if (*(int*)state.StringBuffer != 1162626372 ||
-                     *(short*)(state.StringBuffer + 4) != 17748)
-                        ThrowInvalidProperty(state, ctx);
-                    return CommandType.DELETE;
-
+                    if (*(int*)state.StringBuffer == 1162626372 &&
+                        *(short*)(state.StringBuffer + 4) == 17748) 
+                        return CommandType.DELETE;
+                    break;
+                    
                 case 8:
-                    if (*(long*)state.StringBuffer != 8318823012450529091)
-                        ThrowInvalidProperty(state, ctx);
-                    return CommandType.Counters;
+                    if (*(long*)state.StringBuffer == 8318823012450529091) 
+                        return CommandType.Counters;
+                    break;
                 
                 case 10:
-                    if (*(long*)state.StringBuffer != 6071222181947531586 ||
-                        *(short*)(state.StringBuffer + sizeof(long)) != 18499)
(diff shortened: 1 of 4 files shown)


src/Raven.Server/RavenServerStartup.cs
```diff
@@ -246,6 +239,35 @@ private async Task RequestHandler(HttpContext context)
...
+                    $"RavenDB does not support interaction between Client API major version 3 and Server version {RavenVersionAttribute.Instance.MajorVersion} when major version does not match. " +
+                    $"Client: {version}. Server: {RavenVersionAttribute.Instance.AssemblyVersion}",
+                    e);
+                return;
+            }
+
+            var nVersion = char.GetNumericValue(version, 0);
+            if (nVersion > RavenVersionAttribute.Instance.MajorVersion)
+            {
+                e = new ClientVersionMismatchException(
```
ppekrol: What will happen if we throw e.g. 'IndexDoesNotExistException' when Client version is 5.0.0.50?

haludi: So we will get a warning about the version mismatch.

ppekrol: Which is incorrect? Right?

src/Raven.Server/RavenServerStartup.cs
```diff
@@ -246,6 +241,35 @@ private async Task RequestHandler(HttpContext context)
...
+                    e);
+            }
+            else if (HasInvalidCommandTypeException(e))
+            {
+                RequestRouter.CheckClientVersionAndWrapException(version, ref e);
+            }
+
+            static bool HasInvalidCommandTypeException(Exception e)
+            {
+                if (e.GetType() == typeof(BatchRequestParser.InvalidCommandTypeException))
```
ayende:
```suggestion
                if (e is BatchRequestParser.InvalidCommandTypeException)
```

src/Raven.Server/Routing/RequestRouter.cs
```diff
@@ -223,18 +223,28 @@ public static bool TryGetClientVersion(HttpContext context, out Version version)
...
+            if(CheckClientVersionAndWrapException(clientVersion, ref innerException) == false)
+                throw innerException;
+        }
+
+        public static bool CheckClientVersionAndWrapException(Version clientVersion, ref Exception innerException)
+        {
+            var currentServerVersion = RavenVersionAttribute.Instance;
+
+            if (currentServerVersion.MajorVersion == clientVersion.Major &&
+                currentServerVersion.BuildVersion >= clientVersion.Revision &&
```
ayende, line 238: Why are you comparing build version to revision here?

src/Raven.Server/Routing/RequestRouter.cs
```diff
@@ -223,18 +223,28 @@ public static bool TryGetClientVersion(HttpContext context, out Version version)
...
+                currentServerVersion.BuildVersion != ServerVersion.DevBuildNumber &&
+                (clientVersion.Revision < 40 || clientVersion.Revision >= 50))
```
ayende: Why this line?

src/Raven.Server/Routing/RequestRouter.cs
haludi: There are two different classes.
The `currentServerVersion` is `RavenVersionAttribute`
And the `clientVersion` is System.Version.
So the `RavenVersionAttribute.BuildVersion` is equivalent to `System.Version.Revision`

src/Raven.Server/Routing/RequestRouter.cs
haludi: If I understand it right it is to check if the client unofficially built (in a development environment)

src/Raven.Server/Documents/Handlers/BatchRequestParser.cs
```diff
@@ -1024,5 +1014,9 @@ private static void ThrowUnexpectedEndOfStream()
             throw new EndOfStreamException();
         }
 
+        public class InvalidCommandTypeException : Exception
```
ppekrol: We have dedicated namespace for exceptions. Raven.Server.Exceptions. Move it there.

src/Raven.Server/RavenServerStartup.cs
ppekrol: +1

src/Raven.Server/RavenServerStartup.cs
```diff
@@ -246,6 +241,35 @@ private async Task RequestHandler(HttpContext context)
...
+                    return true;
+
+                if (e is AggregateException ae)
```
ppekrol, line 267: Is this a real case? Do we have aggregate exceptions anywhere? The .Any is not a cheap operaration.

src/Raven.Server/Routing/RequestRouter.cs
ayende: You are excluding only 40 .. 50 here?

haludi: This is the range of v4.x other versions will be check by the first condition

ayende: Problem is that the 6.0 client will be accepted here, accidently.

haludi: How? The major is not equal?

ayende: I don't like that line. It make no sense to me. 
The dev version we use is either 40 or 50 for dev, no? In that case, check for those using equality, please.

src/Raven.Server/RavenServerStartup.cs
haludi: Yes in the bulk insert it is. I can convert it to `foreach`

src/Raven.Server/Documents/Handlers/BatchRequestParser.cs
```diff
@@ -950,46 +951,36 @@ private static unsafe CommandType GetCommandType(JsonParserState state, JsonOper
...
-                    
-                    ThrowInvalidProperty(state, ctx);
-                    return CommandType.None;
-                
-                default:
-                    ThrowInvalidProperty(state, ctx);
-                    return CommandType.None;
+                    break;
             }
+            throw new InvalidCommandTypeException("Invalid command type: " + ctx.AllocateStringValue(null, state.StringBuffer, state.StringSize));
```
arekpalinski: Better to move it to a dedicated method - like we did with `ThrowInvalidProperty`. The reason is slightly better performance - parsing is our hot path.

ppekrol (approved):

PR #10759, opened 2020-06-29
RavenDB-15234 Dynamic fields. Client API throws exception when the fi…

ayende:
…eld is treated as a dictionary with the key starting with a digit

RavenDB-15235 Dynamic fields. Client API mistreats the key when the field is treated as a dictionary and the key contains the slash ('/') symbol


ppekrol:
We have few test failures here. @ayende can you take a look?

ayende:
looking

ayende:
Interesting issue, we have queries like this:

```
var q = session.Query<Test>()
    .Select(x => x.Headers["ABC.DEF"]);
```

Where "ABC.DEF" is a _single_ field name.

I changed it so you'll need to quote the field, like so:

```
var q = session.Query<Test>()
    .Select(x => x.Headers["'ABC.DEF'"]);
```

Otherwise we'll treat it as a compound field name. I think that this is more likely anyway, but needs to add it to the release notes

arekpalinski (approved):

PR #10760, opened 2020-06-29
HRINT-1267 - licensing enhancements

grisha-kotler:


src/Raven.Client/Properties/VersionInfo.cs
```diff
@@ -4,7 +4,7 @@
 using Raven.Client.Extensions;
 using Raven.Client.Properties;
 
-[assembly: RavenVersion(Build = "42", CommitHash = "a377982", Version = "4.2", FullVersion = "4.2.104-custom-42")]
+[assembly: RavenVersion(Build = "42", CommitHash = "a377982", Version = "4.2", FullVersion = "4.2.104-custom-42", EffectiveReleaseDateString = "2019-05-21")]
```
ayende: Isn't this automatically generated?

ayende (approved):

src/Raven.Client/Properties/VersionInfo.cs
grisha-kotler: It is

src/Raven.Client/Properties/VersionInfo.cs
```diff
@@ -4,7 +4,7 @@
 using Raven.Client.Extensions;
 using Raven.Client.Properties;
 
-[assembly: RavenVersion(Build = "42", CommitHash = "a377982", Version = "4.2", FullVersion = "4.2.104-custom-42", EffectiveReleaseDateString = "2019-05-21")]
+[assembly: RavenVersion(Build = "42", CommitHash = "a377982", Version = "4.2", FullVersion = "4.2.104-custom-42", ReleaseDateString = "2020-07-02")]
```
ayende, line 7: The one thing that pops to mind, now this is on each release, what updates this? Do we need to remember doing that?

ppekrol: @gregolsky will handle that on CI

ppekrol (approved):

PR #10770, opened 2020-06-30
RavenDB-15283 Expose time-series value names in query metadata

karmeli87:


src/Raven.Client/Constants.cs
```diff
@@ -199,6 +199,8 @@ private Metadata()
 
                 public const string TimeSeries = "@timeseries";
 
+                public const string TimeSeriesValuesNames = "@values-names";
```
ppekrol: `@timeseries-values-names` ? `@timeseries-value-names` ?

src/Raven.Server/Documents/Queries/Results/TimeSeriesRetriever.cs
```diff
@@ -726,10 +731,17 @@ private BlittableJsonReaderObject GetFinalResult(DynamicJsonArray array, long co
 
             if (addProjectionToResult)
             {
-                result[Constants.Documents.Metadata.Key] = new DynamicJsonValue
+                var metadata = new DynamicJsonValue
                 {
                     [Constants.Documents.Metadata.Projection] = true
                 };
+                var config = _context.DocumentDatabase.ServerStore.Cluster.ReadTimeSeriesConfiguration(_context.DocumentDatabase.Name);
```
ppekrol: This opens an additional tx for server store for each query. Do we need that? Is this for Studio? If yes, IIRC we have that information during the query, so we can only do that when using Studio.

ml054:
Not good, try to execute query:

```
from Countries
where id() == 'countries/PL'
select timeseries(from 'Cases'), timeseries(from 'Deaths')
```

image


Also it should support aliases, etc.

PR #10773, opened 2020-06-30
RavenDB-15175 Time series - Future dates tooltip

Danielle9897:


ml054:
move logic to general utils

ml054 (approved):

PR #10777, opened 2020-07-01
RavenDB-15276 avoid sophisticated change vector in sample data

karmeli87:


src/Raven.Server/Web/Studio/SampleDataHandler.cs
```diff
@@ -67,6 +67,15 @@ public async Task PostCreateSampleData()
                 (index, _) = await ServerStore.SendToLeaderAsync(editTimeSeries);
                 await Database.RachisLogIndexNotifications.WaitForIndexNotification(index, Database.ServerStore.Engine.OperationTimeout);
 
+                var unused = new UpdateUnusedDatabaseIdsCommand(Database.Name, new HashSet<string>
```
ppekrol, line 70: We do not want that. The Sample Data should be already cleared out. This can be used to clean it up when needed after modifications (comment out the code).

ppekrol (changes requested):

PR #10784, opened 2020-07-01
 RavenDB-15281: Make sure the DeleteTombstonesCommand  doesn't increase work counter

garayx:


ayende (approved):

src/Raven.Server/Documents/TransactionOperationsMerger.cs
```diff
@@ -927,11 +927,12 @@ private enum PendingOperations
                 break;
             } while (true);
 
-            var status = GetPendingOperationsStatus(context, pendingOps.Count == 0);
+            var currentOperationsCount = _operations.Count;
+            var status = GetPendingOperationsStatus(context, currentOperationsCount == 0);
```
ppekrol, line 931: It was 'pendingOps' which was renamed to 'executedOps'. Now you are using '_operations.Count'. Is this correct?

garayx: Yes, `pendingOps` were actually the executed ops (see line 870 we get it out of `_operations` and execute on line 897), then we check for more work using `pendingOps ` (line 930) which will always return `HasMore` and we will make another run of `TransactionOperationsMerger`. Also see the logging on line 934 it was actually counting the remaining ops using the `_operation` list.

PR #10787, opened 2020-07-02
RavenDB-15294 V5: Failed to execute Compaction with Document Compress…

ayende:
…ion enabled - Null Reference Exception

### DIFF
--- a/src/Raven.Server/Storage/Schema/Updates/Documents/50000/From42017.cs
+++ b/src/Raven.Server/Storage/Schema/Updates/Documents/50000/From42017.cs
@@ -4,6 +4,7 @@ using System.IO;
 using Raven.Client.Documents.Changes;
 using Raven.Server.Documents;
 using Raven.Server.Documents.Replication.ReplicationItems;
+using Raven.Server.Documents.Revisions;
 using Raven.Server.ServerWide.Context;
 using Raven.Server.Utils;
 using Sparrow;
@@ -13,8 +14,8 @@ using Sparrow.Json.Parsing;
 using Sparrow.Server;
 using Sparrow.Server.Utils;
 using Voron;
+using Voron.Data;
 using Voron.Data.Tables;
-using static Raven.Server.Documents.DocumentsStorage;
 using static Raven.Server.Documents.CountersStorage;
 
 
@@ -49,6 +50,8 @@ namespace Raven.Server.Storage.Schema.Updates.Documents
 
         public bool Update(UpdateStep step)
         {
+            UpdateSchemaForDocumentsAndRevisions(step);
+
             var readTable = new Table(CountersSchema, step.WriteTx);
             var countersTree = readTable.GetTree(CountersSchema.Key);
             if (countersTree == null)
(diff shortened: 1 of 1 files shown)


ayende:
I'm not sure about the solution here. 
The issue is that we already have users that moved to 5.0 beta and have schema version 50_000.
They will not have this applied.
The question is whatever we start out with 50_001 already or not.

ppekrol (approved):
We do not guarantee any data/binary compatibility during beta phase.

What can be done, is at the date of release we can move our 50_000 to 50_001 and write a schema upgrade from 50_000 to 50_001 that throws an exception with message that 50_000 was a beta.

src/Raven.Server/Storage/Schema/Updates/Documents/50000/From42017.cs
```diff
@@ -141,6 +144,23 @@ public bool Update(UpdateStep step)
...
+        private static void UpdateSchemaForDocumentsAndRevisions(UpdateStep step)
+        {
+            using var _ = step.DocumentsStorage.ContextPool.AllocateOperationContext(out DocumentsOperationContext context);
+            var collections = step.WriteTx.OpenTable(DocumentsStorage.CollectionsSchema, DocumentsStorage.CollectionsSlice);
+            foreach (var tvr in collections.SeekByPrimaryKey(Slices.BeforeAllKeys, 0))
+            {
+                var collection = DocumentsStorage.TableValueToId(context, (int)DocumentsStorage.CollectionsTable.Name, ref tvr.Reader);
+                var collectionName = new CollectionName(collection);
+                var tableTree = step.WriteTx.CreateTree(collectionName.GetTableName(CollectionTableType.Documents), RootObjectType.Table);
+                DocumentsStorage.DocsSchema.SerializeSchemaIntoTableTree(tableTree);
```
arekpalinski, line 156: 1. It took me a while to figure out what was the problem and why it failed only during compaction but I still don't fully understand it. I think it's very tricky and having a test for that case is a must.

2. It would also be nice to have a comment here explaining why we need to call `SerializeSchemaIntoTableTree`. Actually I wonder if `ISchemaUpdate` shouldn't have a required field `Description` where an author of the schema update would need to provide a few sentences explaining what is the purpose (or purposes) of this schema change. 

3. The missing piece that I don't get is: why `SerializeSchemaIntoTableTree` (which is called for `DocumentsStorage.DocsSchema`) doesn't need to be called for `DocumentsStorage.CompressedDocsSchema`? NRE was thrown during this call:

https://github.com/ravendb/ravendb/blob/2f55c5f46bb26234c6ef888666fe353ac1c85631/src/Voron/Data/Tables/Table.cs#L809-L812

so `_schema` was an instance of `CompressedDocsSchema`. I might be missing something so just want to make sure.

ayende: The issue was that we didn't serialize the `CompressedEtagsSourceIndex` value. It is a new field in 5.0.
`CompressedDocsSchema` isn't actually serialized, it is identical to `DocsSchema` with the `compressed` flag turned on.

Agree that we got to have a test here.

arekpalinski: Okay, that explains why I couldn't find where do we serialize `CompressedDocsSchema`. Makes sense. 

This means that here we could use `DocumentsStorage.CompressedDocsSchema ` instead of `DocumentsStorage.DocsSchema`, right? Just for readability by making it more clear that it's related to the compression added in 5.0

ayende: I don't think so, we aren't updating the compressed schema, we are updating the general one.

arekpalinski: ok

ayende:
Will push a test soon

ayende:
https://github.com/ravendb/ravendb/pull/10800

PR #10788, opened 2020-07-02
Raven db 15172 time series u iissues

Danielle9897:


src/Raven.Studio/typescript/viewmodels/database/settings/timeSeries.ts
```diff
@@ -141,42 +141,55 @@ class timeSeries extends viewModelBase {
...
+                         <small><br>
+                             <ul>
+                                 <li>
+                                     <strong>A new time series</strong> is created by each Aggregation Policy defined.<br>
+                                     Its name will be <code>&lt;Raw-data-time-series-name&gt;@&lt;Aggregation-policy-name&gt;</code>
+                                  </li><br>
+                                  <li>
+                                     Each entry in such auto-generated time series has the following values:<br>
+                                     <i>First, Last, Min, Max, Sum, Count</i><br>
+                                     which is rolled-up data <strong>based only on the Aggregation Time Frame defined by the policy!</strong>
```
ml054, line 154: I wonder is it worth to mention what happens when we have 2 values in input collection?

src/Raven.Studio/typescript/viewmodels/database/settings/timeSeries.ts
```diff
@@ -141,42 +141,55 @@ class timeSeries extends viewModelBase {
...
+                                 </li><br>
+                                 <li>
+                                     The data generated by the first aggregation policy is based on the Raw time series data.<br>
+                                     After that, input data for each aggregation policy is the data generated by the previous policy.
+                                 </li><br>
+                                 <li>
+                                     <strong>Important</strong>:<br>
+                                     Aggregated data is not created for time series that have entries with more than 5 values!<br>
```
ml054, line 162: @karmeli87 What if I have 4 values for most of the time, and then by mistake I push TS with 6 values. I assume time series policy won't be applied, right? Will it fail silently? Will it continue to work once I delete 2 values from the entry with 6 values? I assume we should notify user about such case (performance hint in studio?)

src/Raven.Studio/typescript/viewmodels/database/settings/timeSeries.ts
Danielle9897: Added
image

src/Raven.Studio/typescript/viewmodels/database/settings/timeSeries.ts
Danielle9897: @karmeli87 @ml054 

As discussed with @karmeli87  :
1. TS policy will Not be applied once an entry has more than 5 values
2. After you delete the 'extra' values, the policies will continue to work as before
3. I added this top warning : 

image

karmeli87: 1. The aggregation should stop working.
2. Message in the notification center should appear.
3. After fixing it (delete the extra values) it should resume the work.

ml054 (approved):

PR #10793, opened 2020-07-05
RavenDB-15246 Review ref parameters for `GetTimeSeriesRange`

efratshenhar:


ppekrol (approved):

karmeli87 (changes requested):
Here we want to test the affect of page size on the client cache.

src/Raven.Client/Documents/Session/AsyncSessionDocumentTimeSeries.cs
```diff
@@ -96,9 +101,11 @@ await ServeFromCacheOrGetMissingPartsFromServerAndMerge(from ?? DateTime.MinValu
                     new GetTimeSeriesOperation<TTValues>(DocId, Name, from, to, start, pageSize), Session._sessionInfo, token: token)
                 .ConfigureAwait(false);
 
-            if (rangeResult == null)
+            if ( (rangeResult == null) || ((rangeResult.Entries.Length == 0) && ( start > 0)))
```
karmeli87: If I requested a range that has no entries, we will not put that into the cache, right? why?

src/Raven.Client/Documents/Session/AsyncSessionDocumentTimeSeries.cs
```diff
@@ -115,6 +122,33 @@ await ServeFromCacheOrGetMissingPartsFromServerAndMerge(from ?? DateTime.MinValu
...
         }
 
+        private async Task FixRange<TTValues>(DateTime? from,
+            DateTime? to,
+            int start,
+            int pageSize,
+            TimeSeriesRangeResult<TTValues> rangeResult,
+            CancellationToken token = default)
+            where TTValues : TimeSeriesEntry
+        {
```
karmeli87: Would it make sense to this on the server-side? and spare the extra call?

src/Raven.Client/Documents/Session/AsyncSessionDocumentTimeSeries.cs
```diff
@@ -173,11 +207,17 @@ await ServeFromCacheOrGetMissingPartsFromServerAndMerge(from ?? DateTime.MinValu
             {
                 if (ranges[toRangeIndex].From <= from)
                 {
-                    if (ranges[toRangeIndex].To >= to)
+                    if ((ranges[toRangeIndex].To >= to) || (ranges[toRangeIndex].Entries.Length > pageSize))
```
karmeli87: Why do I care about the page-size here?

src/Raven.Client/Documents/Session/AsyncSessionDocumentTimeSeries.cs
```diff
@@ -220,16 +260,51 @@ await ServeFromCacheOrGetMissingPartsFromServerAndMerge(from ?? DateTime.MinValu
...
+                {
+                    tsrr.From = tsrr.Entries[0].Timestamp;
+                    changeStart = false;
+                }
+
+                if (partialPageSize != tsrr.Entries.Length) continue;
+
+                Session.IncrementRequestCount();
+
+                var temp = await Session.Operations.SendAsync(
```
karmeli87: 1. Can't the server handle this case and return the wanted results ?
2. If this call is indeed necessary, we need to collect all ranges and use `GetMultipleTimeSeriesOperation`.

karmeli87 (changes requested):

src/Raven.Client/Documents/Session/AsyncSessionDocumentTimeSeries.cs
efratshenhar: only when start is more then 0 because we can't know if we have entries when start will be 0

karmeli87: but we can know that on the server when returning the results, and change the `From` accordingly.

karmeli87: Also to consider is that if I'm doing the same request twice, I should go only once to the server.
But if I have `start` in my request, will that make me for the same request go twice to the server?

src/Raven.Client/Documents/Session/AsyncSessionDocumentTimeSeries.cs
efratshenhar: I think yes. In that case I can fix the range for start also

src/Raven.Client/Documents/Session/AsyncSessionDocumentTimeSeries.cs
efratshenhar: I will move everything to the server

src/Raven.Client/Documents/Session/AsyncSessionDocumentTimeSeries.cs
efratshenhar: If we have in cache range 1-5 with 3 entries and we ask for range 1-7 with page size 3 so we don't care that the range is bigger then what we have because we have all the results we need

karmeli87 (approved):

src/Raven.Client/Documents/Session/AsyncSessionDocumentTimeSeries.cs
```diff
@@ -332,16 +342,22 @@ private static TimeSeriesEntry[] MergeRangesWithResults(DateTime @from, DateTime
...
 
             foreach (var value in range.Entries)
             {
+                if (pageSize-- <= 0)
+                    yield break;
+
                 if (value.Timestamp > to)
                     yield break;
 
+                if (start-- > 0)
```
ppekrol: start and pageSize should be done just before the return after all other filtering (start first, then pageSize)

src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
```diff
@@ -179,7 +179,7 @@ public async Task Read()
                     return;
                 }
 
-                var rangeResult = GetTimeSeriesRange(context, documentId, name, from, to, ref start, ref pageSize);
+                var rangeResult = pageSize > 0 ? GetTimeSeriesRange(context, documentId, name, from, to, ref start, ref pageSize) : new TimeSeriesRangeResult();
```
ppekrol: Can GetTimeSeriesRange return empty result when pageSize is 0?

ppekrol (changes requested):

src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
efratshenhar: yes

ppekrol (approved):

PR #10794, opened 2020-07-05
RavenDB-15291

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-15291

### DIFF
--- a/src/Raven.Server/Storage/Schema/Updates/Documents/50000/From42017.cs
+++ b/src/Raven.Server/Storage/Schema/Updates/Documents/50000/From42017.cs
@@ -1,6 +1,7 @@
 ﻿using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using Raven.Client;
 using Raven.Client.Documents.Changes;
 using Raven.Server.Documents;
 using Raven.Server.Documents.Replication.ReplicationItems;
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Storage/Schema/Updates/Documents/50000/From42017.cs
```diff
@@ -141,6 +139,86 @@ public bool Update(UpdateStep step)
...
+        private static unsafe void UpdateDocumentCounters(UpdateStep step, DocumentsOperationContext context, string docId)
+        {
+            using (DocumentIdWorker.GetSliceFromId(context, docId, out Slice lowerDocId))
+            {
+                var docsTable = new Table(DocsSchema, step.ReadTx);
+                if (docsTable.ReadByKey(lowerDocId, out var tvr) == false) 
+                    return; // document doesn't exists 
+                
+                var tableId = tvr.Id;
+                var counterNames = step.DocumentsStorage.CountersStorage.GetCountersForDocumentList(context, step.WriteTx, docId);
```
ppekrol: better to use `GetCountersForDocument()` and do `.ToList()` 

This way we can avoid having `GetCountersForDocumentList` method

ppekrol (approved):

PR #10797, opened 2020-07-06
RavenDB-15199 Display the number of counters with a thousands separator

Danielle9897:


src/Raven.Studio/typescript/viewmodels/database/documents/editDocument.ts
```diff
@@ -1138,6 +1138,11 @@ class normalCrudActions implements editDocumentCrudActions {
     revisionsCount = ko.observable<number>();
     timeSeriesCount: KnockoutComputed<number>;
 
+    attachmentsCountBadge: KnockoutComputed<string>;
```
ml054: can we create single formatting function instead of creating many new funcions?

Danielle9897: done - I forced push

ml054: let's use this function directly in view. Like I said during our call. We don't have to have so many fields in this class. Let's keep it clean.

ml054: in html:

data-bind="text: $root.formatBadge(attachmentsCount)"

Danielle9897: ok - forced pushed again

ml054 (approved):

PR #10798, opened 2020-07-06
RavenDB-15299 Rollup retention time must be divided without a reminder with the previous one

karmeli87:


src/Raven.Server/Documents/Queries/AST/Query.cs
```diff
@@ -720,7 +720,7 @@ private void SegmentOnRollup(Span<StatefulTimestampValue> values)
                 }
 
                 val = values[index + (int)AggregationType.Count];
-                _count[i] += val.Count;
+                _count[i] += (long)val.Sum;
```
ayende, line 723: Why are you doing this? Shouldn't we worry about losing fractions?

karmeli87: This is `Sum` of the `Count`.
To count the total number of points in a rollup segment we need to use the `Sum` of the `Count` value.

ayende: But you are summing the `Sum`, not ` Count`, am I missing something?

karmeli87: In a rollup we have 6 values (First, Last, ... Count)
each of them has `StatefulTimestampValue`, in order to calculate the total points of the original TS for the given rollup segment, we need to go the Count's `StatefulTimestampValue` and use the `Sum` property.

ayende: ok

PR #10800, opened 2020-07-06
RavenDB-15294

ayende:
…at ensure we can upgrade properly


src/Raven.Server/Storage/Schema/SchemaUpgrader.cs
```diff
@@ -16,7 +16,7 @@ internal class CurrentVersion
 
             public const int ConfigurationVersion = 50_000;
 
-            public const int DocumentsVersion = 50_000;
+            public const int DocumentsVersion = 50_001;
```
ppekrol: Why we did that?

ayende: Because we have users that are already running the beta, and it isn't worth it to make people life harder

karmeli87: I don't think it is a good idea.
For instance, right now we are adding the Original TS name to the `Stats` table.

ppekrol: Actually we do not have a choice anyway, because we applied many fixes to 50_000 migration regarding counters knowing that this is something that we do not care at this moment about and we cannot at this moment assume that for counters no one was affected, but for this change someone was and one is not a problem, but second one is.

ayende: So we basically can't use `50_000` at all? Because it is in an undeterminate state where we don't know where it is at?

ppekrol: Yes, this is what I've wrote in the original PR. There is an issue for that: https://issues.hibernatingrhinos.com/issue/RavenDB-15308

ppekrol (approved):

PR #10803, opened 2020-07-06
RavenDB-15310 Fix case insensitivity in roll-up error

karmeli87:


ppekrol (approved):
Can we have a test for that?

PR #10804, opened 2020-07-06
RavenDB-15236: Don't AddAndStartOutgoingReplication if DocumentDataba…

garayx:
…se was disposed

### DIFF
--- a/src/Raven.Server/Documents/Replication/ReplicationLoader.cs
+++ b/src/Raven.Server/Documents/Replication/ReplicationLoader.cs
@@ -38,6 +38,7 @@ namespace Raven.Server.Documents.Replication
 {
     public class ReplicationLoader : IDisposable, ITombstoneAware
     {
+        private readonly ReaderWriterLockSlim _locker = new ReaderWriterLockSlim();
         public event Action<IncomingReplicationHandler> IncomingReplicationAdded;
         public event Action<IncomingReplicationHandler> IncomingReplicationRemoved;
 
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -967,6 +967,13 @@ internal void AddAndStartOutgoingReplication(ReplicationNode node, bool external
...
                 return;
             }
+
+            if (Database == null)
+            {
+                // database was disabled
+                return;
+            }
+
             var outgoingReplication = new OutgoingReplicationHandler(this, Database, node, external, info);
```
grisha-kotler: What if it happens inside `OutgoingReplicationHandler`?

grisha-kotler (changes requested):

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
ayende: `ReplicationLoader` is already alive only in the context of a database, so I don't see why this is required.
If the DB was disposed one ms afterward?

ayende (changes requested):

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
garayx: this code can run as not awaited [(code line),](https://github.com/ravendb/ravendb/blob/v4.2/src/Raven.Server/Documents/Replication/ReplicationLoader.cs#L725) so when we null the `Database` on dispose of `ReplicationLoader` this code could still run so we get NRE

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
```diff
@@ -85,15 +85,22 @@ public class OutgoingReplicationHandler : IDisposable, IReportOutgoingReplicatio
...
+                _external = external;
+                _log = LoggingSource.Instance.GetLogger<OutgoingReplicationHandler>(_database.Name);
+                _connectionInfo = connectionInfo;
+                _database.Changes.OnDocumentChange += OnDocumentChange;
+                _database.Changes.OnCounterChange += OnCounterChange;
+                _cts = CancellationTokenSource.CreateLinkedTokenSource(_database.DatabaseShutdown);
+            }
+            catch (NullReferenceException)
+            {
+                // expected when DB is disposed
```
ppekrol: I do not like this kind of coding. Please ping me and we will figure it out. If the race is between disposing and creating outgoingReplicationHandler, then we need to prevent both of those operations to run at the same time.

ayende: Yes, there should never be something like that in our code.

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
ayende: We need a better way to detect that.

src/Raven.Server/Documents/Replication/OutgoingReplicationHandler.cs
garayx: done

ppekrol (approved):

PR #10808, opened 2020-07-07
Wire replication with `RegisterBytesSent` and `RegisterBytesReceived`

efratshenhar:


src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -220,7 +220,8 @@ private void CreatePullReplicationAsHub(TcpConnectionOptions tcpConnectionOption
 
             var taskId = pullReplicationDefinition.TaskId; // every connection to this pull replication on the hub will have the same task id.
             var externalReplication = pullReplicationDefinition.ToExternalReplication(initialRequest, taskId);
-            var outgoingReplication = new OutgoingReplicationHandler(this, Database, externalReplication, external: true, initialRequest.Info)
+
+            var outgoingReplication = new OutgoingReplicationHandler(tcpConnectionOptions, this, Database, externalReplication, external: true, initialRequest.Info)
```
karmeli87: IIRC Those `tcpConnectionOptions` are already added in `DispatchDatabaseTcpConnection`.
This is pull replication, meaning the _destination_ opens the tcp connection to the source.

karmeli87 (changes requested):

karmeli87:
We also want to add some sort of identifier for the last outgoing/incoming batch.
I think for both cases (subscription & replication) we want to have the last etag that was sent/received.

src/Raven.Server/Documents/TcpHandlers/SubscriptionConnection.cs
```diff
@@ -485,6 +485,10 @@ private async Task<SubscriptionConnectionClientMessage> GetReplyFromClientAsync(
                     BlittableJsonDocumentBuilder.UsageMode.None,
                     _copiedBuffer.Buffer))
                 {
+                    if (blittable.TryGet("ChangeVector", out string cv))
+                    {
+                        TcpConnection.lastEtagReceived = cv.ToChangeVector()[0].Etag;
```
ayende: This is likely wrong.
Better to have `LastChangeVector`

src/Raven.Server/Documents/TcpHandlers/TcpConnectionOptions.cs
```diff
@@ -18,6 +18,8 @@ public class TcpConnectionOptions: IDisposable
         private MeterMetric _bytesReceivedMetric;
         private MeterMetric _bytesSentMetric;
         private readonly DateTime _connectedAt;
+        public long lastEtagSent;
```
ppekrol: private fields should start with `_`

src/Raven.Server/Documents/TcpHandlers/TcpConnectionOptions.cs
```diff
@@ -135,13 +137,16 @@ public DynamicJsonValue GetConnectionStats()
                 ["Operation"] = Operation.ToString(),
                 ["ClientUri"] = TcpClient?.Client?.RemoteEndPoint?.ToString(),
                 ["ConnectedAt"] = _connectedAt,
-                ["Duration"] = (DateTime.UtcNow - _connectedAt).ToString()
+                ["Duration"] = (DateTime.UtcNow - _connectedAt).ToString(),
+                ["lastEtagReceived"] = lastEtagReceived,
```
ppekrol: notice that everything is uppercased here

src/Raven.Server/Documents/TcpHandlers/SubscriptionConnection.cs
ppekrol: @efratshenhar this is used to identify the batches?

efratshenhar: yes

ppekrol: @ayende I think it is easier to match between source and destination via etag than CV. Thoughts?

ppekrol: @efratshenhar it probably requires Studio work in Replication Stats view (Ongoing Stats view)?

ayende: that may be the case, sure, but the code we have now is not going to work.
It is saying, give me the first cv's etag. What if we have multiple nodes?

ppekrol: So maybe it would be easier via some internal identifier?

E.g.
```
private static long _batchId = 1;

public static long GetNextBatchId()
{
    return Interlocked.Increment(ref _batchId);
}
```

ayende: How would that help the user seeing this? What does this corrleates to?

ppekrol: on A node I will see an Outgoing batch with Id 10 and on B node in Incoming I can identify it easily.

ayende: This is to see where you are on the subscription. No? In that case, etag isn't meaningful.
And an arbitrary integer won't help, you can't compare between nodes, there is only one node in subscription.
Or did I miss something?

ppekrol: This one is in subscription, you are correct, but the original task (in YT) was to match replication batches

ayende: For replication, the etag is already handled, so no issue. 
For subscription, I wonder if we even need it, we already report the last change vector, IIRC

efratshenhar: > @efratshenhar it probably requires Studio work in Replication Stats view (Ongoing Stats view)?

We can see the Etag but not the bytes sent/received

PR #10810, opened 2020-07-07
RavenDB-15113 Fix Storage Report View when Index name is very long

mateuszbartosik:


src/Raven.Studio/wwwroot/Content/css/pages/storage-report.less
```diff
@@ -437,4 +448,28 @@
...
+        padding-right: 20px;
+    }
+
+    .table-item-name {
+        display: block;
+        white-space: nowrap;
+    }
+}
+
+.position-relative {
```
ml054: can we avoid having global scoped classes in view specific file?

src/Raven.Studio/wwwroot/Content/css/pages/storage-report.less
```diff
@@ -437,4 +448,28 @@
             margin-left: -4px;
         }
     }
+
+    .th-min-width {
```
ml054: I think we should remove th from selectors. It will work even when you change this to div right?

ml054: + maybe we can define limited width on parent level and override single column with min-width? By doing this we don't have to repeat classes and we use cascading feature from CSS.

src/Raven.Studio/wwwroot/Content/css/pages/storage-report.less
mateuszbartosik: Fixed.

src/Raven.Studio/wwwroot/Content/css/pages/storage-report.less
mateuszbartosik: I've made it a bit different, with the same effect. I've also added the `title` attribute to the URL in Name column, so the full name will be visible on hover.

ml054 (approved):

PR #10814, opened 2020-07-07
RavenDB-15236 - start the async task only if we have new destinations and log in case of any error

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-15236


src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -722,12 +722,22 @@ private void HandleExternalReplication(DatabaseRecord newRecord, List<IDisposabl
...
-                StartOutgoingConnections(newDestinations, external: true);
-            });
+                Task.Run(() =>
+                {
+                    // here we might have blocking calls to fetch the tcp info.
+                    try
+                    {
+                        StartOutgoingConnections(newDestinations, external: true);
+                    }
+                    catch (Exception e)
```
ayende, line 734: Why will it fail here?

grisha-kotler: It shouldn't but if it does it's better to have it in the logs.

PR #10822, opened 2020-07-08
RavenDB-15321: Backup status for idle db should be loaded from cluster

garayx:


garayx:
https://github.com/ravendb/ravendb/pull/10824

PR #10826, opened 2020-07-09
RavenDB-14771 Revision configuration is collection case-sensitive

efratshenhar:


src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -239,11 +239,12 @@ public RevisionsCollectionConfiguration GetRevisionsConfiguration(string collect
             if (Configuration == null)
                 return ConflictConfiguration.Default;
 
-            if (Configuration.Collections != null &&
-                Configuration.Collections.TryGetValue(collection, out RevisionsCollectionConfiguration configuration))
+            foreach (var col in Configuration.Collections.
+                Where(col => string.Equals(col.Key, collection, StringComparison.InvariantCultureIgnoreCase)))
```
ppekrol: This will have a severe performance impact. The 'Collections' dictionary needs to have a proper comparer. I think you need to use `IFillFromBlittableJson` and implement it in RevisionsConfiguration.

karmeli87 (approved):

test/FastTests/Server/Documents/Revisions/RevisionsTests.cs
```diff
@@ -1087,6 +1087,86 @@ public async Task DeleteRevisionsWhenNoneExistShouldNotThrow()
...
+        public async Task CollectionCaseSensitiveTest1()
+        {
+            using (var store = GetDocumentStore())
+            {
+                var id = "user/1";
+                var configuration = new RevisionsConfiguration
+                {
+                    Collections = new Dictionary<string, RevisionsCollectionConfiguration>
+                    {
+                        ["uSErs"] = new RevisionsCollectionConfiguration { Disabled = false }
```
ppekrol, line 1100: What will happen if I will put `users` and `USERS` on the Client side and push it to the server (can we have a test for that)?

ppekrol (approved):

test/FastTests/Server/Documents/Revisions/RevisionsTests.cs
efratshenhar: Yes. I will add one

PR #10841, opened 2020-07-11
RavenDB-15336

ayende:
- Invalidation of compression cache depends on the _value_, not the schema
- Fixing recovery not using the right extension for compression dictionaries
- Make sure that we report the ActiveCandidateSection as well in the table report
- Changed the test to ensure that even under compression, there would be easily noticed size differences with revisions


ppekrol (approved):

arekpalinski (commented):
LGTM

PR #10842, opened 2020-07-12
RavenDB-15241 Time-series full backup should skip dead segments

karmeli87:


ppekrol (approved):

src/Raven.Server/Smuggler/Documents/DatabaseSource.cs
```diff
@@ -464,13 +464,17 @@ public IEnumerable<SubscriptionState> GetSubscriptions()
...
             Debug.Assert(_context != null);
+            var isFull = _startDocumentEtag == 0;
 
             if (collectionsToExport?.Count > 0)
             {
                 foreach (var collection in collectionsToExport)
                 {
                     foreach (var ts in _database.DocumentsStorage.TimeSeriesStorage.GetTimeSeriesFrom(_context, collection, _startDocumentEtag, long.MaxValue))
                     {
+                        if (isFull && ts.Segment.NumberOfLiveEntries == 0)
```
ppekrol, line 475: Actually... doesn't it affect last etag processed? Shouldn't this be handled in `ProcessTimeSeries` ?

ppekrol:
It needs to be handled differently in the core smuggler component.

PR #10853, opened 2020-07-14
RavenDB-15369 - UnhandledException during server startup

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-15369

### DIFF
--- a/src/Raven.Server/Utils/AffinityHelper.cs
+++ b/src/Raven.Server/Utils/AffinityHelper.cs
@@ -1,0 +1,213 @@
+﻿using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using System.Threading;
+using Raven.Client.Util;
+using Sparrow.Binary;
+using Sparrow.Collections;
+using Sparrow.Logging;
+using Sparrow.Platform;
+using Sparrow.Server.Platform.Posix;
+
+namespace Raven.Server.Utils
+{
+    public class AffinityHelper
+    {
+        private static readonly ReaderWriterLockSlim _affinityLocker = new ReaderWriterLockSlim();
+        private static readonly ConcurrentSet<PoolOfThreads.PooledThread> _customAffinityThreads = new ConcurrentSet<PoolOfThreads.PooledThread>();
+        private static readonly Logger _logger = LoggingSource.Instance.GetLogger<AffinityHelper>("Server");
+
+        public static void SetProcessAffinity(Process process, int cores, long? processAffinityMask, out long currentlyAssignedCores)
+        {
+            currentlyAssignedCores = Bits.NumberOfSetBits(process.ProcessorAffinity.ToInt64());
+            if (currentlyAssignedCores == cores)
+            {
+                // we already set the correct number of assigned cores
+                return;
+            }
+
+            var bitMask = 1L;
+            if (processAffinityMask == null)
+            {
+                for (var i = 0; i < cores; i++)
+                {
+                    bitMask |= 1L << i;
+                }
+            }
+            else if (Bits.NumberOfSetBits(processAffinityMask.Value) > cores)
+            {
+                var affinityMask = processAffinityMask.Value;
+                var bitNumber = 0;
+                while (cores > 0)
+                {
+                    if ((affinityMask & 1) != 0)
+                    {
+                        bitMask |= 1L << bitNumber;
+                        cores--;
+                    }
+
+                    affinityMask = affinityMask >> 1;
+                    bitNumber++;
+                }
+            }
+            else
+            {
+                bitMask = processAffinityMask.Value;
+            }
+
+            _affinityLocker.EnterWriteLock();
+
+            try
+            {
+                process.ProcessorAffinity = new IntPtr(bitMask);
+
+                // changing the process affinity resets the thread affinity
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Utils/AffinityHelper.cs
```diff
@@ -0,0 +1,213 @@
...
+using Sparrow.Collections;
+using Sparrow.Logging;
+using Sparrow.Platform;
+using Sparrow.Server.Platform.Posix;
+
+namespace Raven.Server.Utils
+{
+    public class AffinityHelper
+    {
+        private static readonly ReaderWriterLockSlim _affinityLocker = new ReaderWriterLockSlim();
```
ppekrol: You can use: EasyReaderWriterLock

ppekrol (approved):

src/Raven.Server/Utils/AffinityHelper.cs
```diff
@@ -0,0 +1,201 @@
...
+                }
+            }
+            else
+            {
+                bitMask = processAffinityMask.Value;
+            }
+
+            _affinityLocker.EnterWriteLock();
+
+            using (_affinityLocker.EnterWriteLock())
```
ayende: Why do you have two of these?

src/Raven.Server/Utils/AffinityHelper.cs
```diff
@@ -0,0 +1,201 @@
...
+            {
+                process.ProcessorAffinity = new IntPtr(bitMask);
+
+                // changing the process affinity resets the thread affinity
+                // we need to change the custom affinity threads as well
+                foreach (var pooledThread in _customAffinityThreads)
```
ayende: I'm pretty sure that once we set it at the process level, it applies to the threads too

src/Raven.Server/Utils/AffinityHelper.cs
```diff
@@ -0,0 +1,201 @@
...
+                if (result != 0)
+                    throw new InvalidOperationException(
+                        $"Failed to set affinity for thread: {pooledThread.CurrentUnmanagedThreadId}, " +
+                        $"affinity: {affinity}, result: {result}, error: {Marshal.GetLastWin32Error()}");
+            }
+        }
+
+        internal static void RemoveCustomAffinityThread(PoolOfThreads.PooledThread pooledThread)
+        {
+            using (_affinityLocker.EnterReadLock())
```
ayende: I really don't like the locking that we have here.

src/Raven.Server/Utils/AffinityHelper.cs
grisha-kotler: done

ayende (changes requested):
I don't like this approach.
It would be better to kill the locks entirely and simply retry from the current process state on failure a few time, then give up

src/Raven.Server/Utils/AffinityHelper.cs
grisha-kotler: It does. But it's changing it to the Process affinity and we need it to be 1 less (by default).

src/Raven.Server/Utils/AffinityHelper.cs
ppekrol: What is the problem of using RWLS?

grisha-kotler: moved it to `ResetThreadAffinity`

src/Raven.Server/Utils/AffinityHelper.cs
grisha-kotler: leftover from changed to `EasyReaderWriterLock`. removed.

src/Raven.Server/Utils/AffinityHelper.cs
ayende: Okay

src/Raven.Server/Utils/AffinityHelper.cs
ayende: There is no need to do _any_ locking. That is handled by the OS
We just need to handle retries.
Affinity changes can also come from outside.

grisha-kotler: done

src/Raven.Server/Utils/AffinityHelper.cs
```diff
@@ -0,0 +1,188 @@
...
+            // we need to change the custom affinity threads as well
+            foreach (var pooledThread in _customAffinityThreads)
+            {
+                SetCustomThreadAffinity(pooledThread);
+            }
+        }
+
+        internal static bool ResetThreadAffinity(PoolOfThreads.PooledThread pooledThread)
+        {
+            _customAffinityThreads.TryRemove(pooledThread);
```
ayende, line 68: What about removing this when the thread shuts down? Can we add an explicit handle for that when the thread exit?

ppekrol (approved):

src/Raven.Server/Utils/AffinityHelper.cs
grisha-kotler: done
https://github.com/ravendb/ravendb/pull/10853/commits/2039600eff1c54add86773d92b75d83a06d84c4b

ayende (approved):

PR #10856, opened 2020-07-15
RavenDB-15359  delete empty time-series stats by the cleaner

karmeli87:


ayende (approved):

ppekrol (approved):

src/Raven.Server/Documents/TimeSeries/TimeSeriesStats.cs
```diff
@@ -243,7 +252,12 @@ bool TryHandleDeadSegment()
                     start = first.Timestamp;
                 }
 
-                if (baseline <= end && end <= last)
+                var readerOfLastValue = tss.GetReader(context, slicer.DocId, slicer.Name, start, DateTime.MaxValue);
+                readerOfLastValue.Last();
```
arekpalinski, line 256: Is it ok that result returned by `Last()` isn't assigned? Just to confirm.

src/Voron/Data/Tables/Table.cs
```diff
@@ -1680,6 +1680,25 @@ public long DeleteBackwardFrom(TableSchema.FixedSizeSchemaIndexDef index, long v
             return deleted;
         }
 
+        public bool FindByIndex(TableSchema.FixedSizeSchemaIndexDef index, long value, out TableValueReader reader)
```
arekpalinski, line 1683: Given that this returns `bool` and actual result is `out` then I would rename the method to `TryFindByIndex` to follow the general convention

src/Voron/Data/Tables/Table.cs
```diff
@@ -1680,6 +1680,25 @@ public long DeleteBackwardFrom(TableSchema.FixedSizeSchemaIndexDef index, long v
...
+        {
+            AssertWritableTable();
+            reader = default;
```
arekpalinski, line 1686: This will create `TableValueReader ` instance even if we return `true`. It's minor but can we avoid that and create it only in cases where `false` is returned

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
```diff
@@ -200,8 +216,19 @@ private long PurgeSegments(long upto, DocumentsOperationContext context, Collect
...
                     pendingDeletion.MultiDelete(collectionName.Name, etagSlice);
                 }
-
                 outdated.Clear();
+
+                foreach (var tsKey in uniqueTimeSeries)
+                {
+                    if (table.SeekOnePrimaryKeyWithPrefix(tsKey, Slices.BeforeAllKeys, out _) == false)
+                    {
+                        using (Slice.External(context.Allocator, tsKey, tsKey.Size - 1, out var statsKey))
```
arekpalinski, line 225: Why `- 1` ?

src/Raven.Server/Documents/TimeSeries/TimeSeriesStats.cs
karmeli87: this is only to move the reader to the last segment.

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
karmeli87: the key slice is `docId/tsName/` and I need to transform it into `docId/tsName` 
note: the `/` represents the record separator char

PR #10867, opened 2020-07-15
RDBCL-1507 - Restoring encrypted backup times out

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RDBCL-1507

### DIFF
--- a/src/Raven.Server/Documents/PeriodicBackup/Restore/RestoreBackupTaskBase.cs
+++ b/src/Raven.Server/Documents/PeriodicBackup/Restore/RestoreBackupTaskBase.cs
@@ -201,8 +201,7 @@ namespace Raven.Server.Documents.PeriodicBackup.Restore
                     }
 
                     var databaseRecord = restoreSettings.DatabaseRecord;
-                    if (databaseRecord.Settings == null)
-                        databaseRecord.Settings = new Dictionary<string, string>();
+                    databaseRecord.Settings ??= new Dictionary<string, string>();
 
                     var runInMemoryConfigurationKey = RavenConfiguration.GetKey(x => x.Core.RunInMemory);
                     databaseRecord.Settings.Remove(runInMemoryConfigurationKey);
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Documents/PeriodicBackup/Restore/RestoreBackupTaskBase.cs
```diff
@@ -352,6 +352,45 @@ public async Task<IOperationResult> Execute(Action<IOperationProgress> onProgres
...
+                        continue;
+
+                    restoreResult.AddError("Failed to save the database record, the restore is aborted");
+                    onProgress.Invoke(restoreResult.Progress);
+                    throw;
+                }
+            }
+
+            // we are waiting for the command to propagate to this node after it was accepted by the leader
+            await _serverStore.Cluster.WaitForIndexNotification(index, TimeSpan.FromSeconds(60));
```
karmeli87: Reaching here, doesn't ensure that the command will be committed.
This ensures that it is in the the leader's log _only_, but it has to be in the majority of the nodes in order to be committed.

karmeli87: I would move this into the `try/catch`

grisha-kotler: Maybe it should be in its own `try/catch`?
What if `WriteDatabaseRecordAsync` succeeds and we only fail in `WaitForIndexNotification`?

src/Raven.Server/Documents/PeriodicBackup/Restore/RestoreBackupTaskBase.cs
```diff
@@ -295,15 +293,7 @@ public async Task<IOperationResult> Execute(Action<IOperationProgress> onProgres
 
                     // after the db for restore is done, we can safely set the db state to normal and write the DatabaseRecord
                     databaseRecord.DatabaseState = DatabaseStateStatus.Normal;
-                    var (updateIndex, _) = await _serverStore.WriteDatabaseRecordAsync(databaseName, databaseRecord, null, RaftIdGenerator.DontCareId, isRestore: true);
-                    await _serverStore.Cluster.WaitForIndexNotification(updateIndex);
-
-                    if (databaseRecord.Topology.RelevantFor(_serverStore.NodeTag))
```
ppekrol, line 301: Why this was removed?

grisha-kotler: It isn't needed anymore.
We don't wait for the database to be online in order to complete the restore operation.

PR #10872, opened 2020-07-16
RavenDB-13944 OS info details tooltip width

mateuszbartosik:


ml054:
@mateuszbartosik did you consider to use tooltip-lg (existing class instead of creating new customization) ?

mateuszbartosik:
> @mateuszbartosik did you consider to use tooltip-lg (existing class instead of creating new customization) ?

This is the class we use in RavenDB Cloud. Seems like `.tooltip-lg` class does not exist in this framework.

ml054 (approved):

PR #10881, opened 2020-07-17
RavenDB-15291 Making sure that we we increment tx marker during schem…

arekpalinski:
…a update. The problem was that we always default to 2 which in turn cause to create huge replication batch.

### DIFF
--- a/src/Raven.Server/Storage/Schema/UpdateStep.cs
+++ b/src/Raven.Server/Storage/Schema/UpdateStep.cs
@@ -1,13 +1,16 @@
-﻿using Raven.Server.Documents;
+﻿using System;
+using Raven.Server.Documents;
 using Voron.Impl;
 using Voron.Schema;
 using Raven.Server.ServerWide;
+using Raven.Server.ServerWide.Context;
 
 namespace Raven.Server.Storage.Schema
 {
     public class UpdateStep
     {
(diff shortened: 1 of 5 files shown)


src/Raven.Server/Storage/Schema/UpdateStep.cs
```diff
@@ -19,8 +21,15 @@ public UpdateStep(SchemaUpgradeTransactions transactions)
         public DocumentsStorage DocumentsStorage;
         public ClusterStateMachine ClusterStateMachine;
 
-        public void Commit()
+        public void Commit(DocumentsOperationContext txContext)
         {
+            if (txContext != null)
+            {
+                if (txContext.GetTransactionMarker() == DocumentsOperationContext.DefaultTransactionMarker)
```
karmeli87: `2` is a valid number to have during normal operation

arekpalinski: yes. I've forcepushed the fix

ppekrol (approved):

src/Raven.Server/Storage/Schema/UpdateStep.cs
arekpalinski: and now i'm checking that it's not the same as in previous tx

PR #10884, opened 2020-07-18
HRINT-1715 show correct error for not having rollups & retention

karmeli87:


src/Raven.Server/Documents/TimeSeries/TimeSeriesPolicyRunner.cs
```diff
@@ -61,12 +64,19 @@ public static TimeSeriesPolicyRunner LoadConfigurations(DocumentDatabase databas
...
+
+                if (e is LicenseLimitException lle)
+                {
+                    LicenseLimitWarning.AddLicenseLimitNotification(database.ServerStore.NotificationCenter, lle);
+                }
+                else
+                {
+                    database.NotificationCenter.Add(AlertRaised.Create(
+                        database.Name,
+                        $"policy runner error in {database.Name}", msg,
```
ppekrol: Why the title starts from lower letter here? Also it is hard to know from the title what policy runner it is?

karmeli87: fixed the latter.
There only one runner per database

PR #10885, opened 2020-07-19
RavenDB-15339 set 15 seconds default timeout 

karmeli87:
- `GetConnectionInfo` for external/pull replication need to timeout
- Also the call for first topology should timeout

### DIFF
--- a/src/Raven.Client/Documents/Commands/GetRemoteTaskTopologyCommand.cs
+++ b/src/Raven.Client/Documents/Commands/GetRemoteTaskTopologyCommand.cs
@@ -16,6 +16,7 @@ namespace Raven.Client.Documents.Commands
             _remoteDatabase = remoteDatabase ?? throw new ArgumentNullException(nameof(remoteDatabase));
             _databaseGroupId = databaseGroupId ?? throw new ArgumentNullException(nameof(databaseGroupId));
             _remoteTask = remoteTask ?? throw new ArgumentNullException(nameof(remoteTask));
+            Timeout = TimeSpan.FromSeconds(15);
         }
 
         public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, ServerNode node, out string url)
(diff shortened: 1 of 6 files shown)


src/Raven.Client/Documents/Conventions/DocumentConventions.cs
```diff
@@ -44,7 +44,8 @@ public class DocumentConventions : Client.Conventions
         internal static readonly DocumentConventions DefaultForServer = new DocumentConventions
         {
             SendApplicationIdentifier = false,
-            MaxContextSizeToKeep = new Size(PlatformDetails.Is32Bits == false ? 8 : 2, SizeUnit.Megabytes)
+            MaxContextSizeToKeep = new Size(PlatformDetails.Is32Bits == false ? 8 : 2, SizeUnit.Megabytes),
+            RequestTimeout = TimeSpan.FromSeconds(15)
```
ayende: This is a huge change from the previous value, no?

ayende (commented):
Can this cause compatability issues?

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
karmeli87: Yes, previous value was 12 hours if no timeout was provided.
So each request without timeout would be hang for 12 hours.
This is to prevent cases like firewall dropped the packet and no-one knows about it..

karmeli87: Note that this is a server-side only and used for the request executor in few places:
- Replication
- SecretKeyHandler
- RavenETL

ayende (approved):

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
ppekrol: The question is for what this RequestExecutor is used, because the implication of this is also that you cannot push/receive data for more that 15s. This is why it was changed to 12h in the first place and the idea was to 'cut down' the timeout on those operations that needed it only. We already did that in the past and implications were severe. ETL stopped working, etc. At this point I do not see a way to set it globally to such a low number, needs to be done individually for all needed commands.

ppekrol (approved):

PR #10890, opened 2020-07-19
HRINT-1723 - add TotalDirty to SNMP

grisha-kotler:


src/Sparrow/LowMemory/DirtyMemoryState.cs
```diff
@@ -5,5 +5,7 @@ public class DirtyMemoryState
         public bool IsHighDirty;
 
         public long TotalDirtyInBytes;
+
+        public long TotalDirtyInMb => new Size(TotalDirtyInBytes, SizeUnit.Bytes).GetValue(SizeUnit.Megabytes);
```
ppekrol: Move this to 4.2 and do the conversion inside the SNMP endpoint

grisha-kotler: done

PR #10892, opened 2020-07-20
RavenDB-15396 added tx-info endpoint for server storage

ppekrol:


src/Raven.Server/Documents/Handlers/Debugging/ServerTransactionDebugHandler.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+using Raven.Server.Routing;
+using Raven.Server.Web;
+using Sparrow.Json;
+using Sparrow.Json.Parsing;
+
+namespace Raven.Server.Documents.Handlers.Debugging
+{
+    public class ServerTransactionDebugHandler : RequestHandler
```
grisha-kotler, line 10: It can be in `AdminStorageHandler`

PR #10900, opened 2020-07-20
RavenDB-15395 : Rollups don't keep the original casing

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-15395

### DIFF
--- a/src/Raven.Server/Documents/TimeSeries/TimeSeriesRollups.cs
+++ b/src/Raven.Server/Documents/TimeSeries/TimeSeriesRollups.cs
@@ -197,8 +197,9 @@ namespace Raven.Server.Documents.TimeSeries
                 if (rollUpTime > currentTicks)
                     return;
 
-                DocumentsStorage.TableValueToSlice(context, (int)RollupColumns.Key, ref item.Result.Reader,out var key);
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/TimeSeries/TimeSeriesRollups.cs
```diff
@@ -197,8 +197,9 @@ internal void PrepareRollups(DocumentsOperationContext context, DateTime current
                 if (rollUpTime > currentTicks)
                     return;
 
-                DocumentsStorage.TableValueToSlice(context, (int)RollupColumns.Key, ref item.Result.Reader,out var key);
+                DocumentsStorage.TableValueToSlice(context, (int)RollupColumns.Key, ref item.Result.Reader, out var key);
                 SplitKey(key, out var docId, out var name);
+                name = context.DocumentDatabase.DocumentsStorage.TimeSeriesStorage.GetOriginalName(context, docId, name);
```
karmeli87, line 202: let's have it in try/catch and skip on error

karmeli87 (changes requested):

src/Raven.Server/Documents/TimeSeries/TimeSeriesRollups.cs
aviv86: https://github.com/ravendb/ravendb/pull/10903

PR #10903, opened 2020-07-20
RavenDB-15395 : skip on error

aviv86:


src/Raven.Server/Documents/TimeSeries/TimeSeriesRollups.cs
```diff
@@ -199,7 +199,17 @@ internal void PrepareRollups(DocumentsOperationContext context, DateTime current
...
                 SplitKey(key, out var docId, out var name);
-                name = context.DocumentDatabase.DocumentsStorage.TimeSeriesStorage.GetOriginalName(context, docId, name);
+                try
+                {
+                    name = context.DocumentDatabase.DocumentsStorage.TimeSeriesStorage.GetOriginalName(context, docId, name);
+                }
+                catch (Exception e)
+                {
+                    if (_logger.IsInfoEnabled)
+                        _logger.Info($"skipped '{item.Key}'. " + e);
```
karmeli87: should be `_logger.Info($"skipped '{item.Key}'. ", e);`

karmeli87 (approved):

karmeli87:
We don't want to skip the rollup after all and it should be deleted under the rollup command.

PR #10904, opened 2020-07-21
RavenDB-14386 Copy time series when cloning document

efratshenhar:


src/Raven.Client/Documents/Commands/Batches/CopyTimeSeriesCommandData.cs
```diff
@@ -0,0 +1,43 @@
...
+        public CopyTimeSeriesCommandData(string sourceDocumentId, string destinationDocumentId, string changeVector)
+        {
+            if (string.IsNullOrWhiteSpace(sourceDocumentId))
+                throw new ArgumentNullException(nameof(sourceDocumentId));
+            if (string.IsNullOrWhiteSpace(destinationDocumentId))
+                throw new ArgumentNullException(nameof(destinationDocumentId));
+           
+
+            Id = sourceDocumentId;
+            Name = destinationDocumentId;
```
ppekrol: Check how it is done in CopyAttachmentCommandData
The property is called 'DestinationId'

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -1001,6 +1001,43 @@ protected override long ExecuteCmd(DocumentsOperationContext context)
...
+
+                                foreach (var singleResult in reader.AllValues())
+                                {
+                                    var temp = new TimeSeriesOperation.AppendOperation()
+                                    {
+                                        Tag = singleResult.Tag,
+                                        Timestamp = singleResult.Timestamp,
+                                        Values = singleResult.Values.ToArray()
+                                    };
+                                    tso.Append(temp);
```
karmeli87: Instead of materialize all of the results at once, we can directly append them.
should be something like `AppendTimestamp(context, cmd.Id, collection, name, reader.AllValues())`

karmeli87 (changes requested):

karmeli87 (approved):

src/Raven.Client/Documents/Commands/Batches/CopyTimeSeriesCommandData.cs
```diff
@@ -0,0 +1,54 @@
...
+                throw new ArgumentNullException(nameof(destinationDocumentId));
+            if (string.IsNullOrWhiteSpace(destinationName))
+                throw new ArgumentNullException(nameof(destinationName));
+
+
+            Id = sourceDocumentId;
+            Name = sourceName;
+            DestinationId = destinationDocumentId;
+            DestinationName = destinationName;
+            ChangeVector = changeVector;
```
karmeli87: We don't need the change vector here, so please remove it from the ctor as well

src/Raven.Client/Documents/Commands/Batches/ICommandData.cs
```diff
@@ -36,6 +36,7 @@ public enum CommandType
         Counters,
         TimeSeries,
         TimeSeriesBulkInsert,
+        TimeSeriesCopy,
```
karmeli87, line 39: need to handle the `DeferInternal` method as well

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -1001,6 +1001,26 @@ protected override long ExecuteCmd(DocumentsOperationContext context)
...
+                                    cmd.DestinationId,
+                                    docCollection,
+                                    cmd.DestinationName,
+                                    reader.AllValues()
+                                );
+
+                            Reply.Add(new DynamicJsonValue
+                            {
+                                [nameof(BatchRequestParser.CommandData.Id)] = cmd.DestinationId,
+                                [nameof(BatchRequestParser.CommandData.ChangeVector)] = reader.GetCurrentSegmentChangeVector(),
```
karmeli87: you should use the CV from the `AppendTimestamp` result

src/Raven.Client/Documents/Commands/Batches/CopyTimeSeriesCommandData.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+using Raven.Client.Documents.Conventions;
+using Raven.Client.Documents.Session;
+using Sparrow.Json;
+using Sparrow.Json.Parsing;
+
+
+namespace Raven.Client.Documents.Commands.Batches
+{
+    public class CopyTimeSeriesCommandData : ICommandData
```
karmeli87: make this internal, currently we don't want to expose this to user.

karmeli87 (approved):

ppekrol (approved):

PR #10908, opened 2020-07-21
RavenDB-14776 Adding `/databases/*/admin/configuration/settings` to debug package

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-14776

### DIFF
--- a/src/Raven.Server/Config/ConfigurationEntryValue.cs
+++ b/src/Raven.Server/Config/ConfigurationEntryValue.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using Microsoft.Extensions.Configuration;
 using Raven.Client.ServerWide;
 using Raven.Server.Config.Attributes;
 using Sparrow.Json.Parsing;
@@ -45,13 +46,13 @@ namespace Raven.Server.Config
 
     public class ConfigurationEntryServerValue : ConfigurationEntryValue
     {
-        public ConfigurationEntryServerValue(RavenConfiguration configuration, ConfigurationEntryMetadata metadata, RavenServer.AuthenticationStatus authenticationStatus)
+        public ConfigurationEntryServerValue(IConfiguration configuration, ConfigurationEntryMetadata metadata, RavenServer.AuthenticationStatus authenticationStatus)
             : base(metadata)
         {
             ServerValues = new Dictionary<string, ConfigurationEntrySingleValue>();
             foreach (var key in Metadata.Keys)
             {
-                var value = configuration.GetServerWideSetting(key);
+                var value = configuration[key];
                 if (value == null)
                     continue;
 
@@ -59,7 +60,8 @@ namespace Raven.Server.Config
                 ServerValues[key] = new ConfigurationEntrySingleValue
                 {
                     Value = hasAccess && Metadata.IsSecured == false ? value : null,
-                    HasAccess = hasAccess
+                    HasAccess = hasAccess,
+                    HasValue = true,
                 };
             }
         }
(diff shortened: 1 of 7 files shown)


src/Raven.Server/Documents/Handlers/Admin/AdminConfigurationHandler.cs
```diff
@@ -18,7 +18,7 @@ namespace Raven.Server.Documents.Handlers.Admin
 {
     public class AdminConfigurationHandler : DatabaseRequestHandler
     {
-        [RavenAction("/databases/*/admin/configuration/settings", "GET", AuthorizationStatus.DatabaseAdmin)]
+        [RavenAction("/databases/*/admin/configuration/settings", "GET", AuthorizationStatus.DatabaseAdmin, IsDebugInformationEndpoint = true)]
```
haludi, line 21: I added only this endpoint and not "/admin/configuration/settings" endpoint because the second defined as `AuthorizationStatus.ClusterAdmin` and the higher `AuthorizationStatus` of the "debug-info" endpoint is `AuthorizationStatus.Operator`.
But actually it seems (for now at least) that the "/admin/configuration/settings" endpoint contains less information.
In the current implementation, the user will not be possible to get the settings if there are no databases but will get duplicate of the server setting on any database.
In addition, "/databases/*/debug/info-package" endpoint defined as `AuthorizationStatus.ValidUser` so should we ignore the setting in the debug package if the user has less permission then `AuthorizationStatus.DatabaseAdmin` as "/databases/*/admin/configuration/settings" defined?

src/Raven.Server/Config/ConfigurationEntryValue.cs
```diff
@@ -45,21 +46,22 @@ public virtual DynamicJsonValue ToJson()
...
     public class ConfigurationEntryServerValue : ConfigurationEntryValue
     {
-        public ConfigurationEntryServerValue(RavenConfiguration configuration, ConfigurationEntryMetadata metadata, RavenServer.AuthenticationStatus authenticationStatus)
+        public ConfigurationEntryServerValue(IConfiguration configuration, ConfigurationEntryMetadata metadata, RavenServer.AuthenticationStatus authenticationStatus)
             : base(metadata)
         {
             ServerValues = new Dictionary<string, ConfigurationEntrySingleValue>();
             foreach (var key in Metadata.Keys)
             {
-                var value = configuration.GetServerWideSetting(key);
```
haludi, line 54: In the server configuration, the `ServerWideSettings` is always null

src/Raven.Server/Web/System/AdminConfigurationHandler.cs
```diff
@@ -39,7 +39,7 @@ public Task GetSettings()
                 if (scope.HasValue && scope != configurationEntryMetadata.Scope)
                     continue;
 
-                var entry = new ConfigurationEntryServerValue(Server.Configuration, configurationEntryMetadata, status);
+                var entry = new ConfigurationEntryServerValue(Server.Configuration.Settings, configurationEntryMetadata, status);
```
haludi, line 42: https://github.com/ravendb/ravendb/pull/10908/files#r458143604

ayende (commented):
Won't this allow a non db admin to generate the debug package with db admin details?

ppekrol:
@Danielle9897 Can you review changes made here?
@haludi Did you test that this works properly with different level of security clearance?

haludi:
> @Danielle9897 Can you review changes made here?
> @haludi Did you test that this works properly with different level of security clearance?

I did test manually but now I added unit tests.
In the process I realize my changes also apply on "/databases/*/admin/debug/txinfo" and "/databases/*/admin/debug/cluster/txinfo" so user with only read/write permission will not get info from those endpoints.

Danielle9897 (approved):

PR #10909, opened 2020-07-21
RavenDB-15302 We have to invalidate a page from the decompressed page…

arekpalinski:
…s cache if the page that was compressed no longer needs to be compressed (after deletions of entries from it)

### DIFF
--- a/src/Voron/Data/Compression/DecompressedLeafPage.cs
+++ b/src/Voron/Data/Compression/DecompressedLeafPage.cs
@@ -58,6 +58,8 @@ namespace Voron.Data.Compression
                     using (TreeNodeHeader.ToSlicePtr(tx.Allocator, node, out var slice))
                         Original.CopyNodeDataToEndOfPage(node, slice);
                 }
+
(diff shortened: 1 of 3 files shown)


ayende (approved):

ayende:
Do we have a way to create a test for this?

arekpalinski:
I hope so. As I wrote in the ticket I will be working on a test today. The purpose of this draft PR was just to run all tests with this change.

src/Voron/Data/Compression/DecompressedLeafPage.cs
```diff
@@ -58,6 +58,8 @@ public void CopyToOriginal(LowLevelTransaction tx, bool defragRequired, bool was
                     using (TreeNodeHeader.ToSlicePtr(tx.Allocator, node, out var slice))
                         Original.CopyNodeDataToEndOfPage(node, slice);
                 }
+
+                tree.DecompressionsCache.Invalidate(PageNumber, DecompressionUsage.Write);
```
arekpalinski, line 62: This is the fix

PR #10916, opened 2020-07-22
RavenDB-15404 - Restoring encrypted backup times out

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RDBCL-1507

### DIFF
--- a/src/Raven.Server/Documents/DocumentDatabase.cs
+++ b/src/Raven.Server/Documents/DocumentDatabase.cs
@@ -303,7 +303,7 @@ namespace Raven.Server.Documents
                 PeriodicBackupRunner = new PeriodicBackupRunner(this, _serverStore, wakeup);
 
                 _addToInitLog("Initializing IndexStore (async)");
-                _indexStoreTask = IndexStore.InitializeAsync(record, _addToInitLog);
+                _indexStoreTask = IndexStore.InitializeAsync(record, index, _addToInitLog);
                 _addToInitLog("Initializing Replication");
                 ReplicationLoader?.Initialize(record);
                 _addToInitLog("Initializing ETL");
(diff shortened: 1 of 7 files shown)


karmeli87 (approved):

grisha-kotler:
Failing tests:
`SlowTests.Server.Documents.PeriodicBackup.EncryptedBackupTest.can_backup_and_restore_sample_data_encrypted`
`SlowTests.Server.Documents.PeriodicBackup.RavenDB_12012.CreateFullAndIncrementalBackupWithIndexInTheMiddle`

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -364,6 +351,19 @@ public void Initialize(InitializeOptions options = InitializeOptions.None, DateT
                         }
                     }
                 }, DatabaseShutdown);
+
+                try
+                {
+                    _notifyFeaturesChangedTask.Wait();
```
karmeli87: We weren't waiting before, because it might take a lot of time and stall the loading of other databases

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -660,159 +660,154 @@ private unsafe void DisposeInternal()
 
             var exceptionAggregator = new ExceptionAggregator(_logger, $"Could not dispose {nameof(DocumentDatabase)} {Name}");
 
-            var lockTaken = false;
-            Monitor.TryEnter(_clusterLocker, TimeSpan.FromSeconds(5), ref lockTaken);
+            // must acquire the lock in order to prevent concurrent access to index files
+            Monitor.Enter(_clusterLocker);
```
karmeli87: Do we need to block it here to prevent the concurrent access?
Maybe if lock not taking try to take it again down the line?

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -1190,57 +1185,61 @@ private void NotifyFeaturesAboutStateChange(DatabaseRecord record, long index)
...
+        }
+
+        private void TryTakeClusterLock()
+        {
+            var lockTaken = false;
+            while (lockTaken == false)
+            {
+                DatabaseShutdown.ThrowIfCancellationRequested();
+
+                Monitor.TryEnter(_clusterLocker, TimeSpan.FromSeconds(5), ref lockTaken);
```
karmeli87: We must return if the lock was taken or not, and release it only if it was taken.

karmeli87 (changes requested):

src/Raven.Server/Documents/DocumentDatabase.cs
grisha-kotler: This results in missing indexes after the database started.

src/Raven.Server/Documents/DocumentDatabase.cs
grisha-kotler: it is always taken unless there is a database shutdown

src/Raven.Server/Documents/DocumentDatabase.cs
karmeli87: what do u mean by missing indexes? on the database record?

src/Raven.Server/Documents/DocumentDatabase.cs
karmeli87: You can take the `_clusterLocker` in `NotifyFeaturesAboutXXX` methods as well and they can compete with each other..

grisha-kotler: This is how it used to work. I just unified both usages.

src/Raven.Server/Documents/DocumentDatabase.cs
grisha-kotler: The database was restored (smuggler). We start the database and the indexes aren't started (yet).
You are right, we cannot wait here. I'll fix the test.

src/Raven.Server/Documents/DocumentDatabase.cs
grisha-kotler: We must wait since this can cause concurrent access to the indexes.

arekpalinski (approved):

PR #10920, opened 2020-07-23
RavenDB-15407: fix NRE when calculating the size of snapshot restore

garayx:


grisha-kotler (changes requested):
`FileName` is the full path of the file

src/Raven.Server/Documents/PeriodicBackup/Restore/RestoreBackupTaskBase.cs
```diff
@@ -769,9 +769,7 @@ protected virtual void Dispose()
 
         public async Task<long> CalculateBackupSizeInBytes()
         {
-            var zipPath = GetBackupLocation();
-             zipPath = Path.Combine(zipPath, RestoreFromConfiguration.LastFileNameToRestore);
-            using (var zip = await GetZipArchiveForSnapshotCalc(zipPath))
+            using (var zip = await GetZipArchiveForSnapshotCalc(RestoreFromConfiguration.LastFileNameToRestore))
```
ayende, line 772: Where is the NRE fix?

ayende (approved):

src/Raven.Server/Documents/PeriodicBackup/Restore/RestoreBackupTaskBase.cs
garayx: Path.Combine was messing up the paths if we restore destination is cloud storage, if the data not found on cloud storage we return null

grisha-kotler (approved):

PR #10924, opened 2020-07-23
RavenDB-15408 IgnoreProcessorAffinityChanges should be respected

karmeli87:


src/Raven.Server/Commercial/LicenseManager.cs
```diff
@@ -785,26 +786,28 @@ private void SetAffinity(Process process, int cores, LicenseLimits licenseLimits
 
         private bool ShouldIgnoreProcessorAffinityChanges(int cores, LicenseLimits licenseLimits)
         {
-            if (IgnoreProcessorAffinityChanges == false)
-                return false;
-
-            if (licenseLimits == null)
+            if (IgnoreProcessorAffinityChanges)
```
ayende, line 789: Should ignore and then return false? Is this a case of double negation?

src/Raven.Server/Commercial/LicenseManager.cs
```diff
@@ -785,26 +786,28 @@ private void SetAffinity(Process process, int cores, LicenseLimits licenseLimits
...
 
-            if (ProcessorInfo.ProcessorCount == cores)
-                return false;
+                    if (licenseLimits != null)
+                    {
+                        licenseMessage = $"License limits: {string.Join(", ", licenseLimits.NodeLicenseDetails.Select(x => $"{x.Key}: {x.Value.UtilizedCores}/{x.Value.NumberOfCores}"))}. " + 
+                                         $"Total utilized cores: {licenseLimits.TotalUtilizedCores}. " +
+                                         $"Max licensed cores: {LicenseStatus.MaxCores}";
+                    }
+                    Console.WriteLine(basicMessage + " " + licenseMessage);
```
ayende, line 804: Should probably remove this

src/Raven.Server/Commercial/LicenseManager.cs
karmeli87: `IgnoreProcessorAffinityChanges` is static and used only for the tests not to change the affinity of the process.
As far as I understand we always want to skip setting the affinity in the test.

PR #10928, opened 2020-07-24
RavenDB-15409 run license tests & reduce raft commands

karmeli87:


src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -3237,11 +3239,8 @@ public override async Task OnSnapshotInstalledAsync(long lastIncludedIndex, Serv
             }
             token.ThrowIfCancellationRequested();
 
-            // reload license can send a notification which will open a write tx
-            serverStore.LicenseManager.ReloadLicense();
-            await serverStore.LicenseManager.PutMyNodeInfoAsync();
```
grisha-kotler, line 3242: We need this also on a full snapshot

grisha-kotler (changes requested):

src/Raven.Server/ServerWide/ClusterStateMachine.cs
karmeli87: which is exactly this line:
```
if (fullSnapshot)
          await OnValueChanges(lastIncludedIndex, nameof(PutLicenseCommand));
```

take a look here:
https://github.com/ravendb/ravendb/blob/aa68edc645e3951ff432f630ca54f43c91df86f3/src/Raven.Server/ServerWide/ServerStore.cs#L1048

grisha-kotler (approved):

PR #10930, opened 2020-07-26
RavenDB-15370 Better error message when HUB database is disabled

efratshenhar:


src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -313,13 +314,25 @@ public IncomingReplicationHandler CreateIncomingInstance(TcpConnectionOptions tc
...
             {
+
+                var exceptionSchema = JsonDeserializationClient.ExceptionSchema(readerObject);
+                if (exceptionSchema.Type.Equals("Error"))
+                {
+                    if (_log.IsInfoEnabled)
+                    {
+                        _log.Info("Hub database is disabled", new DatabaseDisabledException(exceptionSchema.Message));
+                    }
+                    throw new DatabaseDisabledException(exceptionSchema.Message);
```
karmeli87: It might be other type of exception. let's use the `ExceptionDispatcher`.

karmeli87 (approved):

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -313,13 +314,19 @@ public IncomingReplicationHandler CreateIncomingInstance(TcpConnectionOptions tc
...
             using (var readerObject = context.ParseToMemory(
                 tcpConnectionOptions.Stream,
                 "IncomingReplication/get-last-etag-message read",
                 BlittableJsonDocumentBuilder.UsageMode.None,
                 buffer))
             {
+
+                var exceptionSchema = JsonDeserializationClient.ExceptionSchema(readerObject);
+                if (exceptionSchema.Type.Equals("Error"))
+                    throw new Exception(exceptionSchema.Message);
```
ppekrol, line 328: Why not create an overload for ExceptionDispatcher.Throw to which we are passing the schema and handling that correctly?

karmeli87: ExceptionDispatcher expect a different format.. we will be handling on unify those here:
https://issues.hibernatingrhinos.com/issue/RavenDB-15455

PR #10935, opened 2020-07-27
RavenDB-15201 See inserted attachments/counters/time series in notifi…

Danielle9897:
…cation center (Bulk Insert)

### DIFF
--- a/src/Raven.Client/Documents/Operations/IOperationProgress.cs
+++ b/src/Raven.Client/Documents/Operations/IOperationProgress.cs
@@ -1,4 +1,5 @@
-﻿using Sparrow.Json.Parsing;
+﻿using System;
+using Sparrow.Json.Parsing;
 
 namespace Raven.Client.Documents.Operations
 {
@@ -41,12 +42,14 @@ namespace Raven.Client.Documents.Operations
         }
     }
 
-    public class BulkInsertProgress : IOperationProgress, IOperationProcessedDetails
+    public class BulkInsertProgress : IOperationProgress
     {
         public long Total { get; set; }
         public long BatchCount { get; set; }
         public string LastProcessedId { get; set; }
         
+        [Obsolete("Use field DocumentsProcessed instead")]
+        public long Processed { get; set; }
         public long DocumentsProcessed { get; set; }
         public long AttachmentsProcessed { get; set; }
         public long CountersProcessed { get; set; }
@@ -69,6 +72,7 @@ namespace Raven.Client.Documents.Operations
                 [nameof(Total)] = Total,
                 [nameof(BatchCount)] = BatchCount,
                 [nameof(LastProcessedId)] = LastProcessedId,
+                [nameof(Processed)] = DocumentsProcessed,
                 [nameof(DocumentsProcessed)] = DocumentsProcessed,
                 [nameof(AttachmentsProcessed)] = AttachmentsProcessed,
                 [nameof(CountersProcessed)] = CountersProcessed,
--- a/src/Raven.Client/Documents/Operations/IOperationProgress.cs
+++ b/src/Raven.Client/Documents/Operations/IOperationProgress.cs
@@ -41,17 +41,20 @@ namespace Raven.Client.Documents.Operations
         }
     }
 
-    public class BulkInsertProgress : IOperationProgress
+    public class BulkInsertProgress : IOperationProgress, IOperationProcessedDetails
     {
-        public long Processed { get; set; }
-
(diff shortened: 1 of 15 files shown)


src/Raven.Client/Documents/Operations/OperationState.cs
```diff
@@ -46,4 +46,13 @@ public enum OperationStatus
         Faulted,
         Canceled
     }
+    
+    public interface IOperationProcessedDetails
+    {
+        public long Total { get; set; }
```
ml054: do we need public in interfaces?

src/Raven.Studio/typings/_studio/dto.d.ts
```diff
@@ -528,7 +528,11 @@ interface virtualBulkOperationItem {
     id: string;
     date: string;
     duration: number;
-    items: number;
+    totalItemsProcessed: number,
+    documentsProcessed: number,
```
ml054: semicolons at the end of line.

test/SlowTests/Client/TimeSeries/BulkInsert.cs
```diff
@@ -1212,5 +1219,88 @@ void AssertError()
                 }
             }
         }
+        
+         [Fact]
```
ml054: wrong indent

ml054: don;t we already have test for that ?

src/Raven.Client/Documents/Operations/IOperationProgress.cs
```diff
@@ -41,17 +41,20 @@ public virtual DynamicJsonValue ToJson()
         }
     }
 
-    public class BulkInsertProgress : IOperationProgress
+    public class BulkInsertProgress : IOperationProgress, IOperationProcessedDetails
     {
-        public long Processed { get; set; }
-
+        public long Total { get; set; }
```
ml054, line 47: you changed processed to total. Is it breaking change? @ppekrol ?

src/Raven.Client/Documents/Operations/OperationState.cs
Danielle9897: removed

src/Raven.Studio/typings/_studio/dto.d.ts
Danielle9897: right - done

test/SlowTests/Client/TimeSeries/BulkInsert.cs
Danielle9897: 1.  fixed indentation
2.  I didn't see a test with bulk insert for all 4 types, so I added

ml054 (approved):

src/Raven.Client/Documents/Operations/IOperationProgress.cs
arekpalinski: `Processed` didn't come from `IOperationProgress` so that could be fine. Although can it break binary compatibility?

src/Raven.Client/Documents/Operations/OperationState.cs
```diff
@@ -46,4 +46,13 @@ public enum OperationStatus
         Faulted,
         Canceled
     }
+    
+    public interface IOperationProcessedDetails
```
ppekrol: Do we need this interface?

src/Raven.Client/Documents/Operations/IOperationProgress.cs
ppekrol: You cannot remove that field. You can mark it as obsolete and set the value from DocumentsProcessed in it instead.

src/Raven.Client/Documents/Operations/OperationState.cs
Danielle9897: Can do without it.
I used in 2 places: ` class BulkInsertProgress` & `class BulkOperationResult`
@ppekrol  should I remove it?

ppekrol: Yeah, let's remove it.

Danielle9897: done
1. removed this interface
2. made 'Processed' property obsolete as suggested

src/Raven.Client/Documents/Operations/IOperationProgress.cs
Danielle9897: done

PR #10936, opened 2020-07-27
RavenDB-14532 allow to specify scale in timeseries queries

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-14532

### DIFF
--- a/src/Raven.Server/Documents/Queries/AST/Query.cs
+++ b/src/Raven.Server/Documents/Queries/AST/Query.cs
@@ -536,15 +536,15 @@ namespace Raven.Server.Documents.Queries.AST
         public bool Any => _values.Count > 0;
 
         private readonly List<double> _values;
-        private readonly List<long> _count;
+        private readonly List<double> _count;
 
-        public IEnumerable<object> Count => _count.Cast<object>();
+        public IEnumerable<double> Count => _count;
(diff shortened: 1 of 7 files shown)


src/Raven.Server/Documents/Queries/AST/Query.cs
```diff
@@ -812,7 +812,9 @@ public IEnumerable<object> GetFinalValues()
                     throw new ArgumentOutOfRangeException("Unknown aggregation operation: " + Aggregation);
             }
 
-            return _values.Cast<object>();
+            return scale.HasValue 
+                ? _values.Select(v => (object)(v * scale.Value))
```
ppekrol: Can we avoid LINQ here and use local static method instead (for both of the cases) ?

karmeli87: Or can we scale it while we are adding the results?

ayende: Scaling while adding is how I would like to go about it.

aviv86: done

ppekrol (approved):

ayende (approved):

PR #10939, opened 2020-07-28
RavenDB-14988

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-14988

### DIFF
--- a/src/Raven.Server/Documents/Queries/AST/Query.cs
+++ b/src/Raven.Server/Documents/Queries/AST/Query.cs
@@ -377,9 +377,10 @@ namespace Raven.Server.Documents.Queries.AST
             }
         }
 
-        public static TimeValue ParseLastFromString(string source)
+        public static (TimeSpan TimePeriod, TimeValue Months) ParseTimePeriodFromString(string source)
         {
-            TimeValue result;
+            TimeSpan timeSpan = default;
+            TimeValue months = default;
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Documents/Queries/AST/Query.cs
```diff
@@ -377,9 +377,10 @@ private static void ParseRange(string source, ref int offset, ref RangeGroup ran
             }
         }
 
-        public static TimeValue ParseLastFromString(string source)
+        public static (TimeSpan TimePeriod, TimeValue Months) ParseTimePeriodFromString(string source)
```
ppekrol: Why can't we use TimeValue?

karmeli87: `TimeValue` has 2 native units `Seconds` & `Months`
We can either change from `Seconds` to `MilliSeconds`, but this will break compatibility and limit the max expressed value to 60 days (it is also an `int`)
Adding an additional unit of `Milliseconds` will complicate the code, so we thought that adding `TimeSpan` here is the best solution.

ayende (approved):

aviv86:
Reverted changes, throw NotSupportedException
https://issues.hibernatingrhinos.com/issue/RavenDB-14988#focus=Comments-67-49556.0-0

PR #10941, opened 2020-07-28
RavenDB-14680 Compare Exchange Tombstones are being deleted after ful…

efratshenhar:
…l backups only

### DIFF
--- a/src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
+++ b/src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
@@ -453,7 +453,10 @@ namespace Raven.Server.ServerWide.Maintenance
                     || lastRaftIndexBlittable.TryGet(nameof(PeriodicBackupStatus.LastEtag), out long? lastRaftIndex) == false)
                     continue;
 
-                if (lastIncrementalBackupDate == null || lastRaftIndex == null)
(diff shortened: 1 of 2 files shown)


src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
```diff
@@ -453,7 +453,10 @@ private long GetMaxTombstonesEtagToDelete(TransactionOperationContext context, s
                     || lastRaftIndexBlittable.TryGet(nameof(PeriodicBackupStatus.LastEtag), out long? lastRaftIndex) == false)
                     continue;
 
-                if (lastIncrementalBackupDate == null || lastRaftIndex == null)
+                if (localBackup.TryGet(nameof(PeriodicBackupStatus.LastFullBackup), out DateTime? lastfullBackupDate) == false)
```
karmeli87, line 456: I think we would want to have an identical code to 5.0 here.
cherry-picking will be problematic here, so simply copy-paste?

note that in 5.0 we don't check for `LastIncrementalBackup` at all.
@ppekrol thoughts?

ppekrol: Is this code problematic? What case is not covered?

efratshenhar: If we have only full backup

ppekrol:
Handled here: https://github.com/ravendb/ravendb/pull/11034

PR #10943, opened 2020-07-28
RavenDB-14784 added the ability to load attachment

ppekrol:
TODO:
- add 'GetContentAsStream' so during Lucene conversion we will be able to write byte[] directly without doing any encoding

### DIFF
--- a/src/Raven.Client/Documents/Indexes/IAttachmentObject.cs
+++ b/src/Raven.Client/Documents/Indexes/IAttachmentObject.cs
@@ -1,0 +1,19 @@
+﻿using System.Text;
+
+namespace Raven.Client.Documents.Indexes
+{
+    public interface IAttachmentObject
+    {
+        public string Name { get; }
+
+        public string Hash { get; }
+
+        public string ContentType { get; }
+
+        public long Size { get; }
+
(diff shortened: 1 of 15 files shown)


src/Raven.Client/Documents/Indexes/IAttachmentObject.cs
```diff
@@ -0,0 +1,19 @@
...
+
+        public string Hash { get; }
+
+        public string ContentType { get; }
+
+        public long Size { get; }
+
+        public string GetContentAsString();
+
+        public string GetContentAsString(Encoding encoding);
```
ayende, line 18: Should we expose a `Stream` here as well?

ayende (commented):
I love that in this model, we don't have to deal with refershing the index on load attachment in a special manner.
A change to the attachment _is_ a change to the document and will updte it.!

src/Raven.Client/Documents/Indexes/IAttachmentObject.cs
ppekrol: Yes, wrote that in the comment above.

ppekrol:
Done.

PR #10945, opened 2020-07-28
RavenDB-15425 - use a single HttpClient for cloud backup/restore operations

grisha-kotler:


src/Raven.Server/Documents/PeriodicBackup/RavenStorageClient.cs
```diff
@@ -9,17 +9,35 @@
...
+            {
+                AutomaticDecompression = System.Net.DecompressionMethods.None
+            };
+
+            var client = new HttpClient(handler)
+            {
+                Timeout = TimeSpan.FromHours(24)
+            };
+
+            HttpClient = client;
```
karmeli87, line 38: Using a read-only HttpClient might be a problem, since it will fail to handle DNS changes
https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests

(We address this issue in the client as well by replacing the HttpClient upon some failures)

ppekrol: We are not using MS DI, so using HttpClientFactory might not be a good idea. But we can use one HttpClient per StorageClient instance. Should be more than enough.

PR #10961, opened 2020-08-03
RavenDB-14724 Delete revisions when deleting a document

efratshenhar:


karmeli87 (approved):

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1537,17 +1537,23 @@ private static Tombstone TableValueToTombstone(JsonOperationContext context, ref
...
-                    var revisionsStorage = DocumentDatabase.DocumentsStorage.RevisionsStorage;
                     if (nonPersistentFlags.Contain(NonPersistentDocumentFlags.FromReplication) == false &&
                         (revisionsStorage.Configuration != null || flags.Contain(DocumentFlags.Resolved)))
                     {
                         revisionsStorage.Delete(context, id, lowerId, collectionName, changeVector, modifiedTicks, nonPersistentFlags, documentFlags);
                     }
                 }
 
+                if (((flags & DocumentFlags.HasRevisions) == DocumentFlags.HasRevisions) &&
+                    (revisionsStorage.Configuration == null) &&
```
ayende, line 1553: Why are you checking the configuration here? Shouldn't we check if we need to preserve the revision on delete? 
I think that is in the configuration

efratshenhar: We want to delete revision when we delete a document and we don't have revision configuration

ayende: Okay, make sense

PR #10964, opened 2020-08-04
RavenDB-15448 Time-series hash should take into account more parameters

efratshenhar:


ppekrol (approved):

src/Raven.Server/Documents/ComputeHttpEtags.cs
```diff
@@ -120,6 +120,13 @@ public static unsafe string ComputeEtagForRevisions(List<Document> revisions)
             return FinalizeHash(size, state);
         }
 
+        public static unsafe string ComputeHashForTimeSeriesResult(DateTime from, DateTime to, UIntPtr size, byte* state)
+        {
+            HashNumber(state, from.GetHashCode());
```
ayende:
```suggestion
            HashNumber(state, from.Ticks);
```

We already have support for hashing long, so let's use that instead

PR #10969, opened 2020-08-04
RavenDB-15469 - @expires and @refresh can be both defined in the metadata

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-15469

### DIFF
--- a/src/Raven.Server/Documents/Expiration/ExpirationStorage.cs
+++ b/src/Raven.Server/Documents/Expiration/ExpirationStorage.cs
@@ -45,19 +45,11 @@ namespace Raven.Server.Documents.Expiration
             if (hasExpirationDate == false && hasRefreshDate == false)
                 return;
 
-            if (hasExpirationDate && hasRefreshDate)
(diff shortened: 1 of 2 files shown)


ayende (approved):

ayende (changes requested):
Wait, why?
What is the issue with defining both of them?

grisha-kotler:
This was the original intension:
https://github.com/ravendb/ravendb/blob/e58e0ceddbf12ace06055383b58dc6c76d3b01f4/src/Raven.Server/Documents/Expiration/ExpirationStorage.cs#L45

However, it didn't work:
https://github.com/ravendb/ravendb/blob/e58e0ceddbf12ace06055383b58dc6c76d3b01f4/src/Raven.Server/Documents/Expiration/ExpirationStorage.cs#L41

Before the change, if you provide both @expires is the one which is going to be used.

ayende:
I would rather fix things so you can properly use both expires and refresh
The reason for this was that I reused some mechanisms early on, but there should be no reason they can't both work now

ayende (approved):

PR #10971, opened 2020-08-05
RavenDB-13838 WhatChanged on document metadata return wrong results

efratshenhar:


src/Raven.Client/Json/BlittableOperation.cs
```diff
@@ -56,6 +56,11 @@ public static bool EntityChanged(BlittableJsonReaderObject newObj, DocumentInfo
             {
                 if (changes == null)
                     return true;
+                if (field.Equals(LastModified) ||
+                    field.Equals(Collection) ||
```
ayende: Why are we ignoring the collection field or the id here?

ppekrol: We already do that 10 lines below for new fields.

efratshenhar: Ccollection doesn't need to be there but we are removing the id in JsonConverter.ToBlittable and we don't want to mark that as removed

ayende: ok

PR #10974, opened 2020-08-05
RavenDB-15081: Fix attachment stream getting deleted while not all re…

garayx:
…vision with this attachment were replicated


karmeli87:
We want this fix for 4.2 as well?

grisha-kotler (approved):

garayx:
moved to 4.2
https://github.com/ravendb/ravendb/pull/10977

PR #10976, opened 2020-08-06
RavenDB-15445 Notification Center Position Fix

arekpalinski:


arekpalinski:
It was fixed in 5.1 (https://github.com/ravendb/ravendb/pull/10956) while we need this for 5.0

PR #10979, opened 2020-08-06
RavenDB-15480 limit indexing batch to managed thread allocations

karmeli87:


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -3388,6 +3391,21 @@ private int MinBatchSize
...
+                    return _managedAllocationLimit.Value;
+
+                var limit = Configuration.MangedAllocationsBatchLimit;
+                
+                _managedAllocationLimit = new Lazy<Size?>(() => limit);
+
+                return _managedAllocationLimit.Value;
+            }
+        }
+
```
ayende: Why? There is nothing expensive here that we are trying to save

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
```diff
@@ -143,6 +143,13 @@ public virtual bool RunInMemory
         [ConfigurationEntry("Indexing.Analyzers.NGram.MaxGram", ConfigurationEntryScope.ServerWideOrPerDatabase)]
         public int MaxGram { get; set; }
 
+        [Description("Managed allocations limit, which an index will stop and complete the current batch")]
+        [DefaultValue(512)]
+        [SizeUnit(SizeUnit.Megabytes)]
+        [IndexUpdateType(IndexUpdateType.Refresh)]
+        [ConfigurationEntry("Indexing.MangedAllocationsBatchLimitInMb", ConfigurationEntryScope.ServerWideOrPerDatabase)]
+        public Size? MangedAllocationsBatchLimit { get; protected set; }
```
ayende: How do I set it to null?

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -191,26 +191,48 @@ public void Record(string message)
...
+            if (TimingTracking.TryAdd(name, holder) == false)
             {
                 throw new ArgumentException($"Recorder with the name '{name}' already exists");
             }
-            return new RachisLogRecorder(queue);
+            return new RachisLogRecorder(holder.TimingTracking);
+        }
+
+        public void RemoveRecorderOlderThan(DateTime after)
+        {
```
ayende, line 220: Check if empty (common) before calling LInq

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -191,26 +191,48 @@ public void Record(string message)
...
+            var toRemove = TimingTracking.Where(x => x.Value.Since <= after).Select(x => x.Key);
```
ayende: Better yet, move the linq statement to the `foreach`

ayende (changes requested):

src/Raven.Server/Documents/Indexes/Index.cs
karmeli87: this is how it is done for other stuff:
https://github.com/ravendb/ravendb/blob/8379818ad7c937cb4393321886749631741074ce/src/Raven.Server/Documents/Indexes/Index.cs#L3591

src/Raven.Server/Rachis/RachisConsensus.cs
karmeli87: wasn't sure it is safe to do. 
done.

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
karmeli87: I think one way is from the `settings.json` to do:
```
"Indexing.MangedAllocationsBatchLimitInMb":null
```

src/Raven.Server/Documents/Indexes/Index.cs
ayende: Let's just init all of that normally in the ctor

karmeli87: we don't have the configuration in the ctor
I will read it directly from the config `Configuration.MangedAllocationsBatchLimit`

ayende (approved):

ayende:
ok

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
```diff
@@ -143,6 +143,13 @@ public virtual bool RunInMemory
         [ConfigurationEntry("Indexing.Analyzers.NGram.MaxGram", ConfigurationEntryScope.ServerWideOrPerDatabase)]
         public int MaxGram { get; set; }
 
+        [Description("Managed allocations limit, which an index will stop and complete the current batch")]
+        [DefaultValue(512)]
+        [SizeUnit(SizeUnit.Megabytes)]
+        [IndexUpdateType(IndexUpdateType.Refresh)]
+        [ConfigurationEntry("Indexing.MangedAllocationsBatchLimitInMb", ConfigurationEntryScope.ServerWideOrPerDatabase)]
```
ppekrol: typo

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
ppekrol: typo

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -3504,6 +3507,17 @@ public bool ShouldReleaseTransactionBecauseFlushIsWaiting(IndexingStatsScope sta
                 }
             }
 
+            if (Configuration.MangedAllocationsBatchLimit != null)
+            {
+                var currentManagedAllocations = new Size(GC.GetAllocatedBytesForCurrentThread(), SizeUnit.Bytes);
+                var diff = currentManagedAllocations - _initialManagedAllocations;
+                if (diff > Configuration.MangedAllocationsBatchLimit.Value)
+                {
+                    stats.RecordMapCompletedReason($"Reached managed allocations limit ({Configuration.MangedAllocationsBatchLimit.Value}). Allocated {diff} in current batch");
```
ppekrol: ToString on 'Configuration.MangedAllocationsBatchLimit.Value' and 'diff' yields human readable string?

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
ppekrol: Does it work like taht or it will switch to 512?

src/Raven.Server/Documents/Indexes/Index.cs
karmeli87: yes, those are `Size`.

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
```diff
@@ -143,6 +143,13 @@ public virtual bool RunInMemory
         [ConfigurationEntry("Indexing.Analyzers.NGram.MaxGram", ConfigurationEntryScope.ServerWideOrPerDatabase)]
         public int MaxGram { get; set; }
 
+        [Description("Managed allocations limit, which an index will stop and complete the current batch")]
```
arekpalinski: The description needs to mention that this is about memory _increase_ of managed allocations. I realized that after I saw the implementation where you calculate `diff`

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
```diff
@@ -143,6 +143,13 @@ public virtual bool RunInMemory
...
+        [DefaultValue(512)]
```
arekpalinski: Why do you think `512` is the golden mean? I think that will drastically reduce the batch sizes for those who have large instances. 

I'm for limiting it but I have some doubts regarding the default value. Why not to limit the batch size (number of docs) or indexing batch size by default instead?

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
ayende: It isn't memory increase, we are allowing up to 512MB of allocations before closing the batch in this thread

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
ayende: The issue is that we saw a customer that had a LOT of allocations in the index.
That meant that we started seeing GC pauses as a result and the cluster become unstable.

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
arekpalinski: What do you think about the following description to make it clear that it's about allocations in a single batch:

"Managed allocations limit in an indexing batch after which the batch will complete and an index will continue by starting a new one"

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
arekpalinski: I understand the reasoning of limiting that and I'm for it. But I have some doubts that 512 is too low. Especially on stronger machines.

As I know @karmeli87 talked to @ppekrol yesterday and there was an idea of setting this to `null` by default and introducing a new performance hint where we could explain the pros and cons of large indexing batches and point to 3 options that we have so a user can limit an indexing batch size.

karmeli87: I not in favor for additional toggles here. 
I will run the indexing benchmark + SO import, so we will have some numbers around this change.

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
karmeli87: done

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
ppekrol: At least we should put it as a separate position in indexing performance batch stats (and expose it in the Studio) so it will be easier to trace down.

ayende: I agree that we want to see that in the stats.
Take into account that we also have to deal with multiple indexes at the same time here.
So the 512MB limit is per index, not global.

That said, the issue is that we want to avoid accumulating garbage during indexing. Because we want to reduce the amount of GC pauses.

karmeli87: So I ran the indexing benchmark and it is too small to show any difference. So I used the SO dump + indexes.
Imported all users and then deployed the indexes (only 2 out of 6 indexes are relevant)

with `null` - 3:01 min
with `512MB` - 5:30 min

ran it few time (on my machine) and got more or less same results as above.
I guess those results are expected.

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -3504,6 +3507,17 @@ public bool ShouldReleaseTransactionBecauseFlushIsWaiting(IndexingStatsScope sta
                 }
             }
 
+            if (Configuration.ManagedAllocationsBatchLimit != null)
+            {
+                var currentManagedAllocations = new Size(GC.GetAllocatedBytesForCurrentThread(), SizeUnit.Bytes);
```
grisha-kotler, line 3513: How expensive is: `GC.GetAllocatedBytesForCurrentThread()`?
This is called per each document that we index.
Don't we have this number from here:
https://github.com/ravendb/ravendb/blob/3ad961e29b41eda13444a9d5d34c4491966a74c6/src/Raven.Server/Documents/Indexes/Index.cs#L3821

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
```diff
@@ -143,6 +143,13 @@ public virtual bool RunInMemory
         [ConfigurationEntry("Indexing.Analyzers.NGram.MaxGram", ConfigurationEntryScope.ServerWideOrPerDatabase)]
         public int MaxGram { get; set; }
 
+        [Description("Managed allocations limit in an indexing batch after which the batch will complete and an index will continue by starting a new one")]
+        [DefaultValue(2048)]
```
grisha-kotler, line 147: Can this number be calculated on startup based on the total memory and the number of cores?

karmeli87: The problem is "How many objects do you have", we might have enough memory to keep 10M objects but this will cause long GC pauses, which we want to avoid.
2GB is some sort of indication that this batch has generated too many objects and should be closed.

karmeli87: we can adjust this number maybe by the CPU clock speed, but I'm sure we don't want to go there :-)

src/Raven.Server/Documents/Indexes/Index.cs
karmeli87: This is pretty light weighted and was requested by oren few years ago ;)

It is not about the memory, it is about the number of objects.

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
ayende: We don't want to go there, yes.
The issue is that we want to limit the amount of _garbage_ we have, not to limit memory usage.
We care about GC latency more than anything else.

ayende (approved):

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
```diff
@@ -143,6 +143,13 @@ public virtual bool RunInMemory
...
+        [SizeUnit(SizeUnit.Megabytes)]
+        [IndexUpdateType(IndexUpdateType.Refresh)]
+        [ConfigurationEntry("Indexing.ManagedAllocationsBatchLimitInMb", ConfigurationEntryScope.ServerWideOrPerDatabase)]
```
ppekrol: ManagedAllocationsBatchLimitInMb => ManagedAllocationsBatchSizeLimitInMb

test/FastTests/Server/Documents/Indexing/Auto/BasicAutoMapReduceIndexing.cs
```diff
@@ -80,7 +80,7 @@ public async Task MultipleReduceKeys(int numberOfUsers, string[] locations)
                 var batchStats = new IndexingRunStats();
                 var scope = new IndexingStatsScope(batchStats);
 
-                var cts = new CancellationTokenSource(TimeSpan.FromSeconds(60));
+                var cts = new CancellationTokenSource(TimeSpan.FromSeconds(600));
```
ppekrol: 600?

src/Raven.Server/Rachis/RachisConsensus.cs
ppekrol: @karmeli87 ?

src/Raven.Server/Documents/Indexes/IndexingStatsAggregator.cs
```diff
@@ -85,7 +85,8 @@ private IndexingPerformanceStats CreateIndexingPerformanceStats(bool completed)
                 SuccessCount = Stats.MapSuccesses + Stats.MapReferenceSuccesses,
                 FailedCount = Stats.MapErrors + Stats.MapReferenceErrors,
                 OutputCount = Stats.IndexingOutputs,
-                DocumentsSize = new Size(Stats.DocumentsSize)
+                DocumentsSize = new Size(Stats.DocumentsSize),
+                AllocatedBytes = Stats.AllocatedBytes
```
ppekrol, line 89: Is this displayed in the Studio?

src/Raven.Server/Documents/Indexes/Index.cs
ppekrol: Can we do it every 100 or 1000 items? Not on every document?

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
karmeli87: done

test/FastTests/Server/Documents/Indexing/Auto/BasicAutoMapReduceIndexing.cs
karmeli87: done

src/Raven.Server/Documents/Indexes/IndexingStatsAggregator.cs
karmeli87: it is not displayed, but now it should be exposed to the studio.

PR #10980, opened 2020-08-06
RavenDB-12859 added projection behavior

ppekrol:


src/Raven.Client/Documents/Queries/ProjectionBehavior.cs
```diff
@@ -0,0 +1,130 @@
...
+    {
+        /// <summary>
+        /// Try to extract value from index field (if field value is stored in index) on a failure (or when field value is not stored in index) extract value from a document
+        /// </summary>
+        Default,
+
+        /// <summary>
+        /// Try to extract value from index field (if field value is stored in index)
+        /// </summary>
+        FromIndex,
```
ayende, line 15: This will return null if it isn't there? If so, may want to add that to the comments.

src/Raven.Client/Documents/Queries/ProjectionBehavior.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        /// <summary>
+        /// Extract value from index field or throw
+        /// </summary>
+        FromIndexOrThrow,
+
+        /// <summary>
+        /// Try to extract value from document field
+        /// </summary>
+        FromDocument,
```
ayende, line 25: This will return null if it isn't there? If so, may want to add that to the comments.

src/Raven.Client/Documents/Session/IAsyncDocumentQuery.cs
```diff
@@ -110,6 +113,15 @@ public interface IAsyncDocumentQuery<T> :
         /// <param name="fields">Array of fields to load.</param>
         IAsyncDocumentQuery<TProjection> SelectFields<TProjection>(params string[] fields);
 
+        /// <summary>
+        ///     Selects the specified fields directly from the index if the are stored. If the field is not stored in index, value
+        ///     will come from document directly.
```
ayende: This comment isn't accurate, no? The rule is based on the provided `projectionBehavior`

src/Raven.Client/Documents/Session/IAsyncDocumentQuery.cs
```diff
@@ -127,7 +139,16 @@ public interface IAsyncDocumentQuery<T> :
         IAsyncDocumentQuery<TProjection> SelectFields<TProjection>();
 
         /// <summary>
-        ///     Selects a Time Series Aggregation based on 
+        ///     Selects the specified fields directly from the index if the are stored. If the field is not stored in index, value
+        ///     will come from document directly.
```
ayende: Same

src/Raven.Client/Documents/Queries/ProjectionBehavior.cs
ppekrol: Done.

src/Raven.Client/Documents/Queries/ProjectionBehavior.cs
ppekrol: Done.

src/Raven.Client/Documents/Session/IAsyncDocumentQuery.cs
ppekrol: Done.

src/Raven.Client/Documents/Session/IAsyncDocumentQuery.cs
ppekrol: Done.

PR #10982, opened 2020-08-07
RavenDB-15453

ayende:
- Fixing OOM when during recursive graph query on large graph
- Fixing tx already open when computing etag on query after the data was changed.

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -3447,17 +3447,23 @@ namespace Raven.Server.Documents.Indexes
 
                 using (_contextPool.AllocateOperationContext(out TransactionOperationContext indexContext))
                 {
+                    bool shouldClose =false;
+                    if (context.AreTransactionsOpened() == false)
+                    {
+                        context.OpenReadTransaction();
+                        shouldClose = true;
+                    }
+
                     using (indexContext.OpenReadTransaction())
                     {
-                        DocumentsTransaction  d = null;
-                        if (context.Documents.HasTransaction == false)
-                        {
-                            d= context.Documents.OpenReadTransaction();
-                        }
-
-                        using (d)
+                        try
                         {
                             return CalculateIndexEtag(context, indexContext, q, IsStale(context, indexContext));
+                        }
+                        finally
+                        {
+                            if(shouldClose)
+                                context.CloseTransaction();
                         }
                     }
                 }
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -3447,25 +3447,19 @@ namespace Raven.Server.Documents.Indexes
 
                 using (_contextPool.AllocateOperationContext(out TransactionOperationContext indexContext))
                 {
-                    bool shouldClose =false;
-                    if (context.AreTransactionsOpened() == false)
-                    {
-                        context.OpenReadTransaction();
-                        shouldClose = true;
-                    }
-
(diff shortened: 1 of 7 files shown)


test/SlowTests/Graph/RavenDB-15453.cs
```diff
@@ -0,0 +1,73 @@
...
+
+namespace SlowTests.Graph
+{
+    public class RavenDB_15453 : RavenTestBase
+    {
+        public RavenDB_15453(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        public class Item
```
ppekrol: private

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -3448,9 +3448,17 @@ public long GetIndexEtag(QueryOperationContext context, QueryMetadata q)
                 using (_contextPool.AllocateOperationContext(out TransactionOperationContext indexContext))
                 {
                     using (indexContext.OpenReadTransaction())
-                    using (context.OpenReadTransaction())
```
ppekrol, line 3451: I don ot like this change. You should use 'CalculateIndexEtag' directly. IIRC the order of opened transactions heree mattered.

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -3448,9 +3448,17 @@ public long GetIndexEtag(QueryOperationContext context, QueryMetadata q)
...
                     {
-                        return CalculateIndexEtag(context, indexContext, q, IsStale(context, indexContext));
+                        DocumentsTransaction  d = null;
+                        if (context.Documents.HasTransaction == false)
```
ppekrol: This is incorrect. QueryOperationContext operates on both Server-Storage and Database-Storage transactions. IIRC there is a method there 'AreTransactionsOpened' to check that.

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -3448,9 +3448,17 @@ public long GetIndexEtag(QueryOperationContext context, QueryMetadata q)
                 using (_contextPool.AllocateOperationContext(out TransactionOperationContext indexContext))
                 {
                     using (indexContext.OpenReadTransaction())
-                    using (context.OpenReadTransaction())
                     {
-                        return CalculateIndexEtag(context, indexContext, q, IsStale(context, indexContext));
+                        DocumentsTransaction  d = null;
```
ppekrol: `d` ?

src/Raven.Server/Documents/Indexes/Index.cs
ayende: Changed to use that

test/SlowTests/Graph/RavenDB-15453.cs
ayende: done

ppekrol (approved):

ppekrol:
It causes the 'BasicGraphQueries.CanHandleCyclesInGraph' to fail, altough the asssertion there looks suspicious? https://jenkins.hibernatingrhinos.com/job/PR_Tests_v5.0/1430/testReport/junit/(root)/(empty)/Graph_BasicGraphQueries_CanHandleCyclesInGraph/

```
Assert.Equal() Failure\r\nExpected: String[] [\"employees/5-A\", \"employees/2-A\", \"employees/2-A\"]\r\nActual:   String[] [\"employees/5-A\", \"employees/2-A\"]
```

ayende:
Yes, the test itself is wrong here.

PR #10991, opened 2020-08-11
RavenDB-15497 - Infinite loop when timeout is exceeded in WaitForIndexesAfterSaveChanges

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-15497

### DIFF
--- a/src/Raven.Server/Documents/Handlers/BatchHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/BatchHandler.cs
@@ -341,42 +341,39 @@ namespace Raven.Server.Documents.Handlers
                 indexesToWait.Add(indexToWait);
             }
 
-            using (contextPool.AllocateOperationContext(out DocumentsOperationContext context))
+            var lastEtag = lastChangeVector != null ? ChangeVectorUtils.GetEtagById(lastChangeVector, database.DbBase64Id) : 0;
+            var cutoffEtag = Math.Max(lastEtag, lastTombstoneEtag);
+
+            while (true)
             {
-                while (true)
+                var hadStaleIndexes = false;
+
+                using (contextPool.AllocateOperationContext(out DocumentsOperationContext context))
+                using (context.OpenReadTransaction())
                 {
-                    var hadStaleIndexes = false;
-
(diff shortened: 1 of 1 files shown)


ppekrol (approved):

ppekrol:
Can you add a test for this?

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -341,39 +341,39 @@ public static async Task WaitForReplicationAsync(DocumentDatabase database, Time
...
+
+            while (true)
             {
-                while (true)
-                {
-                    var hadStaleIndexes = false;
+                var hadStaleIndexes = false;
 
-                    using (context.OpenReadTransaction())
+                using (contextPool.AllocateOperationContext(out DocumentsOperationContext context))
```
arekpalinski, line 351: How moving the context allocation here fixes the infinite loop? We open the transaction on each iteration - that should be sufficient/

grisha-kotler: Two issues:
- Infinite loop as long as the relevant indexes are stale and throwOnTimeout is false, here:
https://github.com/ravendb/ravendb/blob/v4.2/src/Raven.Server/Documents/Handlers/BatchHandler.cs#L365
- If we have a long timeout, the memory allocations continue to grow

arekpalinski: So the fix for the infinite loop is not here but it was the first commit - https://github.com/ravendb/ravendb/pull/10991/commits/50e4b7e1ec4914317ab23dd929e5057485ba0cc0. Now it's clear. 

Why memory allocations continue to grow - don't we reuse already allocated memory in a given context?

arekpalinski: Okay. This is because of https://issues.hibernatingrhinos.com/issue/RavenDB-13589

grisha-kotler:
https://github.com/ravendb/ravendb/pull/10998

PR #10993, opened 2020-08-11
RavenDB-15443: Subscription worker should reconnect when subscription…

garayx:
… was updated


src/Raven.Client/Exceptions/Documents/Subscriptions/SubscriptionClosedException.cs
```diff
@@ -10,10 +10,17 @@ namespace Raven.Client.Exceptions.Documents.Subscriptions
 {
     public class SubscriptionClosedException : SubscriptionException
     {
+        internal bool TryToReconnect { get; set; }
```
ppekrol: CanReconnect

ppekrol (commented):
This should go into 4.2?

garayx:
> This should go into 4.2?

no there is no update method in 4.2

ayende (approved):

PR #10994, opened 2020-08-11
RavenDB-15166: Query provider now understands string.Compare

jamesfera:
Query provider now understands string.Compare

RavenDB-15166

https://issues.hibernatingrhinos.com/issue/RavenDB-15166


ppekrol (approved):
LGTM.

What I think we also need to add is the exception when non-supported string.Compare overload is used. I see that there are 10 of them, but we only want to support simplest one `string.Compare(A, B)`. Correct? If yes, can we block others and add a test for that?

grisha-kotler (approved):

ayende (approved):

PR #11002, opened 2020-08-13
RavenDB-15504 Lazy query - Invoke IDocumentStore.OnBeforeQuery

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15504

### DIFF
--- a/src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
+++ b/src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
@@ -20,6 +20,7 @@ using Raven.Client.Documents.Queries;
 using Raven.Client.Documents.Queries.MoreLikeThis;
 using Raven.Client.Documents.Session.Loaders;
 using Raven.Client.Documents.Session.Operations;
+using Raven.Client.Documents.Session.Operations.Lazy;
 using Raven.Client.Documents.Session.Tokens;
 using Raven.Client.Extensions;
 using Sparrow.Json;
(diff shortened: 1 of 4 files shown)


haludi:
According to the discussion [here](https://issues.hibernatingrhinos.com/issue/RavenDB-14959), I first thought the `IDocumentStore.OnBeforeQuery` event should be called one moment before the execution. 
But as it implemented now the LazyQueryOperation is created when the `Lazily` function called and so any future changes to `DocumentQuery` (one of the args in the `IDocumentStore.OnBeforeQuery` event) will not make any affect.
In addition, `AbstractDocumentQuery.BeforeQueryExecutedCallback` is also invoke when the `Lazily` function is called.
It makes for me more sense if those two would be called one moment before the execution but I think it requires too many changes that will change the current behavior and public client API

src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
```diff
@@ -224,6 +225,16 @@ public void WaitForNonStaleResults(TimeSpan? waitTimeout = null)
             Timeout = waitTimeout ?? DefaultTimeout;
         }
 
+        internal LazyQueryOperation<T> GetLazyQueryOperation()
+        {
+            var beforeQueryExecutedEventArgs = new BeforeQueryEventArgs(TheSession, this);
```
ppekrol: Shouldn't this only be called when QueryOperation is `null`? Check `InitSync` and `InitAsync` in DocumentQuery.

ppekrol (commented):
Do we need this for `CountLazily` and `CountLazilyAsync` also?

grisha-kotler (approved):

ppekrol (approved):

PR #11005, opened 2020-08-14
RavenDB-13732 added recurse support

ppekrol:


src/Raven.Server/Documents/Indexes/Static/JavaScript/RecursiveJsFunction.cs
```diff
@@ -0,0 +1,85 @@
+using System;
+using System.Collections.Generic;
+using Jint;
+using Jint.Native;
+using Jint.Native.Array;
+using Jint.Native.Function;
+
+namespace Raven.Server.Documents.Indexes.Static.JavaScript
+{
+    public class RecursiveJsFunction
```
ayende, line 10: I don't think that there is a need for this. After all, you can _already_ do recursion natively in js, no need to jump through hoops.

ppekrol: It makes us consistent with C# API, also we do not require user to write that code + code there is not that straight-forward.

ayende (approved):

PR #11011, opened 2020-08-17
RavenDB-15500 Fixing serialization of projected decimal fields

ppekrol:


ayende:
@ppekrol Thanks,didn't manage to get to it fast enough

PR #11012, opened 2020-08-17
RavenDB-15506

ayende:
- Fixing subscription usage of in operations
- Fixing @metadata references from JS when we can provide the value
- Allow to run subscription on @all_docs

### DIFF
--- a/src/Raven.Server/Documents/ETL/EtlTransformer.cs
+++ b/src/Raven.Server/Documents/ETL/EtlTransformer.cs
@@ -86,12 +86,12 @@ namespace Raven.Server.Documents.ETL
             if (args[1].IsObject() == false)
                 ThrowInvalidScriptMethodCall("loadTo(name, obj) second argument must be an object");
 
-            using (var result = new ScriptRunnerResult(DocumentScript, args[1].AsObject()))
-            {
-                LoadToFunction(args[0].AsString(), result);
-
-                return result.Instance;
-            }
+            // explicitly not disposing here, this will clear the context from the JavaScriptUtils, but this is 
+            // called _midway_ through the script, so that is not something that we want to do. The caller will
+            // already be calling that.
+            var result = new ScriptRunnerResult(DocumentScript, args[1].AsObject());
+            LoadToFunction(args[0].AsString(), result);
+            return result.Instance;
         }
 
         private JsValue LoadToFunctionTranslator(string name, JsValue self, JsValue[] args)
@@ -102,12 +102,12 @@ namespace Raven.Server.Documents.ETL
             if (args[0].IsObject() == false)
                 ThrowInvalidScriptMethodCall($"loadTo{name}(obj) argument must be an object");
 
-            using (var result = new ScriptRunnerResult(DocumentScript, args[0].AsObject()))
-            {
-                LoadToFunction(name, result);
-
-                return result.Instance;
-            }
+            // explicitly not disposing here, this will clear the context from the JavaScriptUtils, but this is 
+            // called _midway_ through the script, so that is not something that we want to do. The caller will
+            // already be calling that.
+            var result = new ScriptRunnerResult(DocumentScript, args[0].AsObject());
+            LoadToFunction(name, result);
+            return result.Instance;
         }
 
         protected abstract void AddLoadedAttachment(JsValue reference, string name, Attachment attachment);
--- a/src/Raven.Server/Documents/Patch/BlittableObjectInstance.cs
+++ b/src/Raven.Server/Documents/Patch/BlittableObjectInstance.cs
@@ -38,7 +38,6 @@ namespace Raven.Server.Documents.Patch
 
         public SpatialResult? Distance => _doc?.Distance;
         public float? IndexScore => _doc?.IndexScore;
-
         private void MarkChanged()
         {
             Changed = true;
--- a/src/Raven.Server/Documents/Patch/JintNullPropagationReferenceResolver.cs
+++ b/src/Raven.Server/Documents/Patch/JintNullPropagationReferenceResolver.cs
@@ -3,6 +3,7 @@ using Jint;
 using Jint.Native;
 using Jint.Runtime.Interop;
 using Jint.Runtime.References;
+using Raven.Client;
 
 namespace Raven.Server.Documents.Patch
 {
@@ -26,6 +27,12 @@ namespace Raven.Server.Documents.Patch
 
         public virtual bool TryPropertyReference(Engine engine, Reference reference, ref JsValue value)
         {
+            if (reference.GetReferencedName() == Constants.Documents.Metadata.Key && 
+                reference.GetBase() is BlittableObjectInstance boi)
+            {
+                value = engine.Invoke(ScriptRunner.SingleRun.GetMetadataMethod, boi);
+                return true;
+            }
             if (reference.GetReferencedName() == "reduce" &&
                 value.IsArray() && value.AsArray().Length == 0)
             {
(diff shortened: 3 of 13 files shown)


src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -845,6 +845,17 @@ public IEnumerable<Document> GetDocumentsFrom(DocumentsOperationContext context,
             }
         }
 
+        public IEnumerable<Document> GetAllDocumentsFrom(DocumentsOperationContext context, long etag)
```
ppekrol: We already have this method. Use 'GetDocumentsFrom` instead.

src/Raven.Server/Documents/Patch/BlittableObjectInstance.cs
```diff
@@ -38,6 +38,7 @@ public class BlittableObjectInstance : ObjectInstance
 
         public SpatialResult? Distance => _doc?.Distance;
         public float? IndexScore => _doc?.IndexScore;
+        public Document Document { get; set; }
```
ppekrol: is this used?

src/Raven.Server/Documents/Patch/JintNullPropagationReferenceResolver.cs
```diff
@@ -26,6 +26,12 @@ public virtual bool TryUnresolvableReference(Engine engine, Reference reference,
 
         public virtual bool TryPropertyReference(Engine engine, Reference reference, ref JsValue value)
         {
+            if (reference.GetReferencedName() == "@metadata" &&
```
ppekrol: use Constants

src/Raven.Server/Documents/Patch/JintNullPropagationReferenceResolver.cs
```diff
@@ -26,6 +26,12 @@ public virtual bool TryUnresolvableReference(Engine engine, Reference reference,
 
         public virtual bool TryPropertyReference(Engine engine, Reference reference, ref JsValue value)
         {
+            if (reference.GetReferencedName() == "@metadata" && 
+                reference.GetBase() is BlittableObjectInstance boi)
+            {
+                value = engine.Invoke("getMetadata", boi);
```
ppekrol: can we define that somewhere and use the constant?

test/SlowTests/Issues/RavenDB-15506.cs
```diff
@@ -0,0 +1,91 @@
...
+    public class RavenDB_15506 : RavenTestBase
+    {
+        public RavenDB_15506(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        private class Item
+        {
+        }
+        public class User{}
```
ppekrol: private?

test/SlowTests/Issues/RavenDB-15506.cs
```diff
@@ -0,0 +1,91 @@
...
+            }
+
+            await store.Subscriptions.CreateAsync(new SubscriptionCreationOptions()
+            {
+                Name = "TestSub",
+                Query = @"from @all_docs as doc
+where doc.'@metadata'.'@collection' in ('Items','Users')"
+            });
+            
+            var worker = store.Subscriptions.GetSubscriptionWorker(new SubscriptionWorkerOptions("TestSub"){
```
ppekrol: formatting?

src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
```diff
@@ -82,12 +83,19 @@ public class SubscriptionDocumentsFetcher
...
             {
-                foreach (var doc in _db.DocumentsStorage.GetDocumentsFrom(
-                    docsContext,
-                    _collection,
-                    startEtag + 1,
-                    0,
-                    long.MaxValue))
+                IEnumerable<Document> documents = _collection switch
+                {
+                    Constants.Documents.Collections.AllDocumentsCollection =>
```
ppekrol, line 88: just a side note: if subscription for all docs will be created with filter, then for each document we will execute JS to check if it matches or not. This can cause some performance issues.

test/SlowTests/Issues/RavenDB-15506.cs
```diff
@@ -0,0 +1,91 @@
...
+        public class User{}
+
+        [Fact]
+        public async Task CanUseInOnMetadataInSubscription()
+        {
+            using var store = GetDocumentStore();
+            using (var s = store.OpenAsyncSession())
+            {
+                await s.StoreAsync(new Item());
+                await s.StoreAsync(new User());
```
ppekrol, line 38: can we create an entity that would not match the subscription?

src/Raven.Server/Documents/DocumentsStorage.cs
ayende: That is used for replication and return the wrong result

src/Raven.Server/Documents/Patch/BlittableObjectInstance.cs
ayende: No, the `_doc` is used.

src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
ayende: Pretty much the same rule for all subscriptions.
In practice, you typically have a single large collection, so that doesn't matter

src/Raven.Server/Documents/DocumentsStorage.cs
ppekrol:
```
public IEnumerable<Document> GetDocumentsFrom(DocumentsOperationContext context, long etag, long start, long take, DocumentFields fields = DocumentFields.All)
```

Use this one.

src/Raven.Server/Documents/TcpHandlers/SubscriptionDocumentsFetcher.cs
ppekrol: I'm just saying that there is a significant difference between doing

```
from Employees
```

and
```
from @all_docs where doc.'@metadata'.'@collection' == 'Employees'
```

when it comes to performance. Not only when we are changing something in 'Employees', but also when changing (maybe more often) documents in e.g. 'Logs' collection.

This needs to be documented well and should not be advised to do so if working on one collection.

ayende: Sure, that is a good point. We can do `from @all_docs(Users, Items)` or something like that in the next itme

ppekrol:
@ayende can you check failing tests?

ayende:
Should be fine now

ppekrol (approved):

ppekrol:
@ayende the 'RavenDB_11379.Should_remove_attachment' is failing. I'm wondering if this solution is not too generic. It should only affect subscriptions? Maybe regular patching.

ayende:
Looking at the tests, but in general, I _want_ to make it global change.

src/Raven.Server/Documents/ETL/EtlTransformer.cs
```diff
@@ -86,12 +86,12 @@ private JsValue LoadToFunctionTranslator(JsValue self, JsValue[] args)
             if (args[1].IsObject() == false)
                 ThrowInvalidScriptMethodCall("loadTo(name, obj) second argument must be an object");
 
-            using (var result = new ScriptRunnerResult(DocumentScript, args[1].AsObject()))
-            {
-                LoadToFunction(args[0].AsString(), result);
-
-                return result.Instance;
-            }
+            // explicitly not disposing here, this will clear the context from the JavaScriptUtils, but this is
```
ppekrol, line 89: This is also releasing all disposables (cloned documents). Who is releasing them now?

ayende: I don't think so, the only thing it does is call to:
https://github.com/ravendb/ravendb/blob/v5.0/src/Raven.Server/Documents/Patch/ScriptRunnerResult.cs#L52

`JavaScriptUtils.Clear()` - But that is _shared_ on the whole script runner. This will be disposed by the calling script, so not a owrry

arekpalinski (approved):

PR #11013, opened 2020-08-17
RavenDB-14534 RequestExecutor.OnBeforeRequest RequestExecutor.OnAfterRequests

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-14534

### DIFF
--- a/src/Raven.Client/Documents/DocumentStoreBase.cs
+++ b/src/Raven.Client/Documents/DocumentStoreBase.cs
@@ -220,12 +220,12 @@ namespace Raven.Client.Documents
         {
             add
             {
-                AssertNotInitialized(nameof(OnFailedRequest));
+                AssertNotInitialized(nameof(OnBeforeRequest));
                 _onBeforeRequest += value;
             }
             remove
             {
-                AssertNotInitialized(nameof(OnFailedRequest));
+                AssertNotInitialized(nameof(OnBeforeRequest));
                 _onBeforeRequest -= value;
             }
         }
@@ -235,12 +235,12 @@ namespace Raven.Client.Documents
         {
             add
             {
-                AssertNotInitialized(nameof(OnFailedRequest));
+                AssertNotInitialized(nameof(OnSucceedRequest));
                 _onSucceedRequest += value;
             }
             remove
             {
-                AssertNotInitialized(nameof(OnFailedRequest));
+                AssertNotInitialized(nameof(OnSucceedRequest));
                 _onSucceedRequest -= value;
             }
         }
--- a/src/Raven.Client/Documents/DocumentStoreBase.cs
+++ b/src/Raven.Client/Documents/DocumentStoreBase.cs
@@ -215,6 +215,36 @@ namespace Raven.Client.Documents
                 _onFailedRequest -= value;
             }
         }
+        private event EventHandler<BeforeRequestEventArgs> _onBeforeRequest;
+        public event EventHandler<BeforeRequestEventArgs> OnBeforeRequest
+        {
+            add
+            {
+                AssertNotInitialized(nameof(OnFailedRequest));
+                _onBeforeRequest += value;
+            }
+            remove
+            {
+                AssertNotInitialized(nameof(OnFailedRequest));
+                _onBeforeRequest -= value;
+            }
+        }
+        
+        private event EventHandler<SucceedRequestEventArgs> _onSucceedRequest;
+        public event EventHandler<SucceedRequestEventArgs> OnSucceedRequest
+        {
+            add
+            {
+                AssertNotInitialized(nameof(OnFailedRequest));
+                _onSucceedRequest += value;
+            }
+            remove
+            {
+                AssertNotInitialized(nameof(OnFailedRequest));
+                _onSucceedRequest -= value;
+            }
+        }
+        
 
         private event EventHandler<TopologyUpdatedEventArgs> _onTopologyUpdated;
         public event EventHandler<TopologyUpdatedEventArgs> OnTopologyUpdated
@@ -289,6 +319,8 @@ namespace Raven.Client.Documents
         protected internal void RegisterEvents(RequestExecutor requestExecutor)
         {
             requestExecutor.OnFailedRequest += _onFailedRequest;
+            requestExecutor.OnBeforeRequest += _onBeforeRequest;
+            requestExecutor.OnSucceedRequest += _onSucceedRequest;
             requestExecutor.OnTopologyUpdated += _onTopologyUpdated;
         }
 
--- a/src/Raven.Client/Http/RequestExecutor.cs
+++ b/src/Raven.Client/Http/RequestExecutor.cs
@@ -171,9 +171,6 @@ namespace Raven.Client.Http
             }
         }
 
-        public event EventHandler<HttpRequestMessage> OnBeforeRequest;
-        public event EventHandler<HttpResponseMessage> OnAfterRequests;
-        
         public event EventHandler<(long RaftCommandIndex, ClientConfiguration Configuration)> ClientConfigurationChanged;
 
         private class FailedRequestTranslator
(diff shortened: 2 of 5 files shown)


src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -889,10 +892,12 @@ protected virtual bool TryLoadFromCache(JsonOperationContext context)
 
                 SetRequestHeaders(sessionInfo, cachedChangeVector, request);
 
+                OnBeforeRequest?.Invoke(this, request);
                 var response = await SendRequestToServer(chosenNode, nodeIndex, context, command, shouldRetry, sessionInfo, request, url, token).ConfigureAwait(false);
                 if (response == null) // the fail-over mechanism took care of this
```
ayende, line 899: What about failures here? How would you deal with the correlation?

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -171,6 +171,9 @@ public TimeSpan FirstBroadcastAttemptTimeout
             }
         }
 
+        public event EventHandler<HttpRequestMessage> OnBeforeRequest;
+        public event EventHandler<HttpResponseMessage> OnAfterRequests;
```
ayende: Not a good idea, we need to send the request as well to the after callback.
Otherwise, not way to correlate

ayende: `BeforeRequest` and `AfterRequest`

The `On` prefix is for the invocation method, usually

test/FastTests/Client/RequestExecutorTests.cs
```diff
@@ -0,0 +1,99 @@
...
+            }
+            catch
+            {
+                // ignored
+            }
+
+            Assert.Equal(expected, actual);
+        }
+        
+        private class FirsFailCommand : PutDocumentCommand
```
ayende:
```suggestion
        private class FirstFailCommand : PutDocumentCommand
```

ayende (changes requested):

haludi:
I wondered what should be the behavior when a request fails in general and particular when we retry.
If we pass the `HttpRequestMessage` just to read so it makes sense to invoke `OnBeforeRequest` just once.
But if we allow the user to modify `HttpRequestMessage` so on any `CreateRequest` the `OnBeforeRequest` should be called.
For `OnAfterRequests` when a request fails we can invoke it with `(HttpResponseMessage)null` or as it implemented now just not invoke it as there is `OnFailedRequest` that invoked (by the way `OnFailedRequest` invoked only on the first retries and not on the last one).
In this case, maybe the name should be `OnSucceedRquest`

src/Raven.Client/Http/RequestExecutor.cs
haludi: We have `OnFailedRequest` so may it worth for consistency?
In addition, maybe while we have `OnFailedRequest` the `OnAfterRequests` should be `SucceedRequest`?

src/Raven.Client/Http/RequestExecutor.cs
haludi: If the response cannot be null so the user can get it from the response, no?

src/Raven.Client/Http/RequestExecutor.cs
haludi: https://github.com/ravendb/ravendb/pull/11013#discussion_r471467897

ayende: Good idea, yes

haludi: So, for "On" prefix I will add a property `FailedRequest` to wrap the `OnFailedRequest` and I will remove the "On" prefix from the two new. Is it make sense?

haludi: We already have `FailedRequest` that is obsolete
```
[Obsolete("Use OnFailedRequest instead")]
public event Action<string, Exception> FailedRequest
```

ppekrol: All our events start from 'On' so this is fine, but we need to have 'BeforeRequestEventArgs' and 'AfterRequestEventArgs'

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -171,6 +171,9 @@ public TimeSpan FirstBroadcastAttemptTimeout
             }
         }
 
+        public event EventHandler<HttpRequestMessage> OnBeforeRequest;
```
ppekrol: We need those also in DocumentStoreBase. Check 'OnFailedRequest' there.

src/Raven.Client/Http/RequestExecutor.cs
haludi: What about adding the `HttpRequestMessage` to `FailedRequestEventArgs` for correlation too?

ayende:
I think we want to call before on each time that we make a remote call, so not just once per logical request.
Maybe add some additional arguments that indicate the failure rate?

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -242,6 +242,45 @@ public void Translate(object sender, FailedRequestEventArgs args)
             }
         }
 
+        public event EventHandler<BeforeRequestEventArgs> _onBeforeRequest;
+        public event EventHandler<BeforeRequestEventArgs> OnBeforeRequest
+        {
+            add
+            {
+                lock (_locker)
```
ayende: those aren't actually required. adding/ removing entries from delegate is thread asfe

haludi: So I will remove it from `OnFailedRequest` as well

src/Raven.Client/Documents/DocumentStoreBase.cs
```diff
@@ -215,6 +215,36 @@ private void AssertNotInitialized(string property)
                 _onFailedRequest -= value;
             }
         }
+        private event EventHandler<BeforeRequestEventArgs> _onBeforeRequest;
+        public event EventHandler<BeforeRequestEventArgs> OnBeforeRequest
+        {
+            add
+            {
+                AssertNotInitialized(nameof(OnFailedRequest));
```
ppekrol: OnBeforeRequest

src/Raven.Client/Documents/DocumentStoreBase.cs
```diff
@@ -215,6 +215,36 @@ private void AssertNotInitialized(string property)
...
+                _onBeforeRequest += value;
+            }
+            remove
+            {
+                AssertNotInitialized(nameof(OnFailedRequest));
```
ppekrol: OnBeforeRequest

src/Raven.Client/Documents/DocumentStoreBase.cs
```diff
@@ -215,6 +215,36 @@ private void AssertNotInitialized(string property)
...
+                _onBeforeRequest -= value;
+            }
+        }
+        
+        private event EventHandler<SucceedRequestEventArgs> _onSucceedRequest;
+        public event EventHandler<SucceedRequestEventArgs> OnSucceedRequest
+        {
+            add
+            {
+                AssertNotInitialized(nameof(OnFailedRequest));
```
ppekrol: OnSucceedRequest

src/Raven.Client/Documents/DocumentStoreBase.cs
```diff
@@ -215,6 +215,36 @@ private void AssertNotInitialized(string property)
...
+                _onSucceedRequest += value;
+            }
+            remove
+            {
+                AssertNotInitialized(nameof(OnFailedRequest));
```
ppekrol: OnSucceedRequest

ppekrol (approved):

PR #11019, opened 2020-08-18
RavenDB-15509 Cannot delete by query from studio

Danielle9897:


src/Raven.Studio/typescript/viewmodels/database/query/query.ts
```diff
@@ -757,6 +766,9 @@ class query extends viewModelBase {
             return;
         }
         
+        // if index was defined outside of current view then we don't know that at this point, so must fetch
+        this.fetchAllIndexes(this.activeDatabase());
```
ml054: wait it means, we *have to* fetch *all* indexes on *every* query?

arekpalinski: https://issues.hibernatingrhinos.com/issue/RavenDB-15509#focus=Comments-67-49974.0-0

src/Raven.Studio/typescript/viewmodels/database/query/query.ts
```diff
@@ -644,8 +635,15 @@ class query extends viewModelBase {
         if (!indexName)
             return null;
 
-        const indexes = this.indexes() || [];
-        return indexes.find(i => i.Name === indexName);
+        let currentIndex = this.indexes() ? this.indexes().find(i => i.Name === indexName) : null;
+        
+        if (!currentIndex) {
+            // fetch indexes since this view may not be up-to-date if index was defined outside of studio
+            this.fetchAllIndexes(this.activeDatabase());
```
ml054: it won't work - fetch is async operation, so currentIndex will still be falsy in next line.

src/Raven.Studio/typescript/viewmodels/database/query/query.ts
```diff
@@ -635,8 +634,28 @@ class query extends viewModelBase {
...
-        const indexes = this.indexes() || [];
-        return indexes.find(i => i.Name === indexName);
+        let currentIndex = this.indexes() ? this.indexes().find(i => i.Name === indexName) : null;
+        
+        if (currentIndex) {
+            return currentIndex;
+        }
+        
+        // fetch indexes since this view may not be up-to-date if index was defined outside of studio
+        const deferred = $.Deferred();
```
ml054: create $.Deferred<string> and resolve with proper value - instead of creating variable for that.

src/Raven.Studio/typescript/viewmodels/database/query/query.ts
```diff
@@ -635,8 +634,28 @@ class query extends viewModelBase {
...
+            .done(() => {
+                currentIndex = this.indexes() ? this.indexes().find(i => i.Name === indexName) : null;
+                deferred.resolve();
+            })
+            .fail(() => {
+                currentIndex = null;
+                deferred.resolve();
+            });
+
+        $.when(deferred).done(() => {
```
ml054: here you simply return deferred.

src/Raven.Studio/typescript/viewmodels/database/query/query.ts
```diff
@@ -313,13 +314,15 @@ class query extends viewModelBase {
             this.queriedIndex() ? appUrl.forVisualizer(this.activeDatabase(), this.queriedIndex()) : null);
 
         this.isMapReduceIndex = ko.pureComputed(() => {
-            const currentIndex = this.getCurrentIndex();
+            this.getQueriedIndexInfo();
```
ml054, line 317: ?

src/Raven.Studio/typescript/viewmodels/database/query/query.ts
```diff
@@ -313,13 +314,15 @@ class query extends viewModelBase {
...
+            const currentIndex = this.queriedIndexInfo();
             return !!currentIndex && (currentIndex.Type === "AutoMapReduce" || currentIndex.Type === "MapReduce" || currentIndex.Type === "JavaScriptMapReduce");
         });
 
         this.queryResultsContainMatchingDocuments = ko.pureComputed(() => {
-            const currentIndex = this.getCurrentIndex();
-            return !!currentIndex && currentIndex.SourceType !== "TimeSeries" && currentIndex.SourceType !== "Counters" && !this.isMapReduceIndex();
+            this.getQueriedIndexInfo();
```
ml054, line 323: ?

src/Raven.Studio/typescript/viewmodels/database/query/query.ts
ml054: we should call methods with side effects inside pure computed.

src/Raven.Studio/typescript/viewmodels/database/query/query.ts
ml054: we should call methods with side effects inside pure computed. as it will be invoked on every change.

src/Raven.Studio/typescript/viewmodels/database/query/query.ts
ml054: probably we should call this method in run query - but one time.

Danielle9897: done, moved to runQuery

ml054 (approved):

arekpalinski (approved):

PR #11024, opened 2020-08-19
RavenDB-15007 Distributed counters can overflow

karmeli87:


src/Raven.Server/Documents/CountersStorage.cs
```diff
@@ -1020,22 +1020,30 @@ private void MergeBlobAndDeleteVector(DocumentsOperationContext context, DbIdsHo
...
             var count = localCounterValues.Length / SizeOfCounterValues;
             long value = 0;
-
+            long uncheckedValue = 0;
             try
             {
                 for (int index = 0; index < count; index++)
                 {
+                    if (capOnOverflow)
+                        uncheckedValue = unchecked(value + ((CounterValues*)localCounterValues.Ptr)[index].Value);
```
ayende: This is confusing, better to run the sum operation again on the `catch` clause and use that there

ayende (approved):

ppekrol (approved):

PR #11027, opened 2020-08-19
RavenDB-15508 & RavenDB-15071 Searching for access certs in hub replication and cleaning up certs when deleting hub

ayende:


src/Raven.Server/Commercial/LicenseManager.cs
```diff
@@ -800,7 +800,9 @@ private bool ShouldIgnoreProcessorAffinityChanges(int cores, LicenseLimits licen
                                          $"Total utilized cores: {licenseLimits.TotalUtilizedCores}. " +
                                          $"Max licensed cores: {LicenseStatus.MaxCores}";
                     }
-                    Console.WriteLine(basicMessage + " " + licenseMessage);
+
+                    if (Logger.IsOperationsEnabled)
```
ppekrol: Please revert this. This is used only in tests and should be written to console.

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -3482,12 +3472,24 @@ public IEnumerable<BlittableJsonReaderObject> GetReplicationHubCertificateByHub(
...
+                var blittable = GetReplicationCertificateAccessObject(context, ref val.Reader);
+                string thumbprint = key.ToString().Substring(prefixString.Length);
+                if (filter != null)
+                {
+                    blittable.TryGet("Name", out string name);
+                    var match =
+                        thumbprint.Contains(filter, StringComparison.OrdinalIgnoreCase) ||
+                        name?.Contains(filter, StringComparison.OrdinalIgnoreCase) == true;
+
+                    if(!match)
```
ppekrol: change to `match == false`. This is the form that we use elsewhere.

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -3482,12 +3472,24 @@ public IEnumerable<BlittableJsonReaderObject> GetReplicationHubCertificateByHub(
 
             foreach (var (key, val) in certs.SeekByPrimaryKeyPrefix(prefix, Slices.Empty, start))
             {
+                var blittable = GetReplicationCertificateAccessObject(context, ref val.Reader);
+                string thumbprint = key.ToString().Substring(prefixString.Length);
+                if (filter != null)
+                {
+                    blittable.TryGet("Name", out string name);
```
ppekrol: nameof

src/Raven.Server/ServerWide/Commands/UpdateDatabaseCommand.cs
```diff
@@ -19,7 +21,9 @@ protected UpdateDatabaseCommand(string databaseName, string uniqueRequestId) : b
             DatabaseName = databaseName;
         }
 
-        public abstract string UpdateDatabaseRecord(DatabaseRecord record, long etag);
+        public abstract void UpdateDatabaseRecord(DatabaseRecord record, long etag);
+        
+        public virtual void CleanupLeftovers(ClusterOperationContext ctx, Table items,  Logger clusterAuditLog){}
```
ppekrol: I would prefer a more generic name like `AfterDatabaseRecordUpdate` or sth like that. Maybe it would be better just to change the name of the `UpdateDatabaseRecord` and handle everything there? I do not like the idea of passing info from one method to another via private fields.

Also can we fix formatting?

ayende: We need to separate because the `UpdateDatabaseRecord` may fail (consistently) across nodes.

ppekrol (approved):

PR #11029, opened 2020-08-19
RavenDB-15513 BlittableJsonReaderObject.GetHashCode cause a LOT of co…

ayende:
…llisions


arekpalinski (approved):
We need to check performance in grafana after merging this.

PR #11031, opened 2020-08-20
RavenDB-15177 remove unused dbIds upon counter splitting

karmeli87:


aviv86 (commented):
Looks good. Will review again when SchemaUpgrade part is fixed

Also, I think that maybe we should introduce changes in `InternalPutCounter` to avoid
adding a new dbId and new partial value if that value is 0.

aviv86 (approved):

PR #11033, opened 2020-08-20
RavenDB-15518 Deserialization of 'HubName' when creating Sink Task

ayende:


src/Raven.Client/Documents/Operations/Replication/PullReplicationAsSink.cs
```diff
@@ -10,9 +10,12 @@ public class PullReplicationAsSink : ExternalReplicationBase
         public string CertificatePassword;
 
         [Obsolete("PullReplicationAsSink.HubDefinitionName is not supported anymore. Will be removed in next major version of the product. Use HubName instead.")]
-        public string HubDefinitionName { get => HubName; set => HubName = value; }
+        public string HubDefinitionName { get; set; }
 
-        public string HubName;
+        private string _hubName;
+#pragma warning disable 618 // disable use obselete property
+        public string HubName { get => _hubName ?? HubDefinitionName; set => _hubName = value; }
```
ppekrol, line 17: Maybe it is better to use 'IFillFromBlittableJson' or 'IPostJsonDeserialization' ?

ayende: The problem happens during _serialization_, not deserialization, though.

ppekrol: Then maybe we want to do:
```
public string HubName { get => _hubName ?? (_hubName = HubDefinitionName); set => _hubName = value; }
```
?

ayende: Why does that matter?

ppekrol: Hmm, actually now when I'm thinking about this, it doesn't.

PR #11034, opened 2020-08-20
RavenDB-14680: Fix compare exchange tombstones cleaner not cleaning when there is only full backup, don't clean when backup status is error

garayx:


src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
```diff
@@ -443,18 +443,34 @@ private long GetMaxTombstonesEtagToDelete(TransactionOperationContext context, s
...
+                {
+                    // no backup status for this task
                     continue;
+                }
 
-                if (singleBackupStatus.TryGet(nameof(PeriodicBackupStatus.LocalBackup), out BlittableJsonReaderObject localBackup) == false
-                    || singleBackupStatus.TryGet(nameof(PeriodicBackupStatus.LastRaftIndex), out BlittableJsonReaderObject lastRaftIndexBlittable) == false)
+                if (singleBackupStatus.TryGet(nameof(PeriodicBackupStatus.Error), out BlittableJsonReaderObject error) == false || error != null)
+                {
+                    // this backup task last status is error
```
ppekrol: When we will have more than 1 backup, and one of them will be in error state, according to the PR description it should not clean anything. Now it will just ignore that one and continue with whatever others had. Is this what we want?

src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
```diff
@@ -443,18 +443,34 @@ private long GetMaxTombstonesEtagToDelete(TransactionOperationContext context, s
...
                     continue;
+                }
+
+                if (singleBackupStatus.TryGet(nameof(PeriodicBackupStatus.LastFullBackupInternal), out DateTime? lastFullBackupInternal) == false || lastFullBackupInternal == null)
```
ppekrol, line 451: I know that this is unchangeable and the moment, but why this is called 'Internal' ?

garayx: LastFullBackupInternal is always updated and used internally (setup next backup timer, idle database)

PR #11035, opened 2020-08-23
RavenDB-15353 attach logs to test

karmeli87:


karmeli87:
will try to apply other fixes before doing the above.

PR #11036, opened 2020-08-24
RavenDB-15521 be sure that CV is applied to the metadata

ppekrol:


test/SlowTests/Issues/RavenDB_15521.cs
```diff
@@ -0,0 +1,44 @@
...
+                    SimpleDoc doc = new SimpleDoc() { Id = "TestDoc", Name = "State1" };
+                    session.Store(doc);
+
+                    string attachment = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
+                    var stream = new MemoryStream(Encoding.UTF8.GetBytes(attachment));
+                    session.Advanced.Attachments.Store(doc, "TestAttachment", stream);
+
+                    session.SaveChanges();
+                    var changeVector1 = session.Advanced.GetChangeVectorFor(doc);
+                    session.Advanced.Refresh(doc);
```
ayende, line 31: Not sure that this is a good idea.
What happened if there were _other_ changes to the document in the meantime?

ppekrol: What do you mean? Refresh is fetching data as it is from the Server.

PR #11047, opened 2020-08-25
RavenDB-15517: fix deadlock in PeriodicBackupRunner

garayx:


src/Raven.Server/Documents/PeriodicBackup/PeriodicBackup.cs
```diff
@@ -77,25 +77,37 @@ private void CancelFutureTasks()
             }
         }
 
-        public void UpdateTimer(Timer newBackupTimer, bool discardIfDisabled = false)
+        public void UpdateTimer(Timer newBackupTimer, bool discardIfDisabled = false, bool underLock = true)
```
ppekrol: There are 6 usages for this method. You changed 1 which could mean that for other 5 you just disabled locking. Is this what you wanted? I suggest removing the default value from `underLock` and reviewing other usages.

ppekrol: Having it as a required parameter will also help in the future to avoid potential problem. Someone will have to consider what value to put there instead of going with 'defaults'.

grisha-kotler: Can we use a regular `lock` to avoid this?

garayx: you can't use a lock in this code path because you already under it

grisha-kotler: A semaphore is not reentrant but the lock is.
https://github.com/ravendb/ravendb/blob/a4d8887f76796e9fb83b370c8f1f18a1b3d12bb6/src/Raven.Server/Documents/PeriodicBackup/PeriodicBackup.cs#L14

ppekrol: Regular lock might be an issue here, because you need to be sure that we are releasing from the same thread. I would suggest leaving the code as is.

PR #11048, opened 2020-08-26
RavenDB-14681 Allow special characters in FindCollectionName

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-14681

### DIFF
--- a/src/Raven.Client/Extensions/StringExtensions.cs
+++ b/src/Raven.Client/Extensions/StringExtensions.cs
@@ -1,4 +1,6 @@
+using System;
 using System.Globalization;
+using System.Text;
 
 namespace Raven.Client.Extensions
 {
@@ -57,6 +59,35 @@ namespace Raven.Client.Extensions
             }
 
             return true;
+        }
+        
+        private static readonly char[] LiteralSymbolsToEscape = { '\'', '\"', '\\', '\a', '\b', '\f', '\n', '\r', '\t', '\v' };
+        private static readonly string[] LiteralEscapedSymbols = { @"\'", @"\""", @"\\", @"\a", @"\b", @"\f", @"\n", @"\r", @"\t", @"\v" };
+
+        public static string EscapeString(string value) 
+        {
+            var builder = new StringBuilder(6 * value.Length);
+            foreach (var c in value)
+            {
+                builder.Append(EscapeChar(c));
+            }
+            return builder.ToString();
+        }
+        
+        public static string EscapeChar(char c)
+        {
+            var index = Array.IndexOf(LiteralSymbolsToEscape, c);
+
+            if (index != -1)
+                return LiteralEscapedSymbols[index];
+
+            if (char.IsLetterOrDigit(c) == false 
+                && char.IsWhiteSpace(c) == false 
+                && char.IsSymbol(c) == false 
+                && char.IsPunctuation(c) == false)
+                return @"\u" + ((int)c).ToString("x4");
+
+            return c.ToString();
         }
     }
 }
--- a/src/Raven.Client/Extensions/StringExtensions.cs
+++ b/src/Raven.Client/Extensions/StringExtensions.cs
@@ -64,16 +64,6 @@ namespace Raven.Client.Extensions
         private static readonly char[] LiteralSymbolsToEscape = { '\'', '\"', '\\', '\a', '\b', '\f', '\n', '\r', '\t', '\v' };
         private static readonly string[] LiteralEscapedSymbols = { @"\'", @"\""", @"\\", @"\a", @"\b", @"\f", @"\n", @"\r", @"\t", @"\v" };
 
-        public static string EscapeString(string value)
-        {
-            if (string.IsNullOrEmpty(value))
-                return value;
-            
-            var builder = new StringBuilder(6 * value.Length);
-            EscapeStringInternal(builder, value);
-            return builder.ToString();
-        }
-        
         public static void EscapeString(StringBuilder builder, string value)
         {
             if (string.IsNullOrEmpty(value))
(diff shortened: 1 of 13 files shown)


src/Raven.Server/Documents/Indexes/Static/CurrentIndexingScope.cs
```diff
@@ -89,29 +89,14 @@ public unsafe dynamic LoadDocument(LazyStringValue keyLazy, string keyString, st
...
-                        return DynamicNullObject.Null;
-                    // we intentionally don't dispose of the scope here, this is being tracked by the references
-                    // and will be disposed there.
-                    Slice.From(_documentsContext.Allocator, keyString, out keySlice);
-                }
-
-                // making sure that we normalize the case of the key so we'll be able to find
-                // it in case insensitive manner
-                _documentsContext.Allocator.ToLowerCase(ref keySlice.Content);
+                var key = keyLazy ?? keyString;
```
ppekrol: Please revert this.

src/Raven.Server/Documents/Queries/Results/MapQueryResultRetriever.cs
```diff
@@ -47,7 +48,10 @@ public override bool TryGetKey(Lucene.Net.Documents.Document input, IState state
 
         protected override Document DirectGet(Lucene.Net.Documents.Document input, string id, DocumentFields fields, IState state)
         {
-            return DocumentsStorage.Get(_context, id, fields);
+            using (Slice.From(_context.Allocator, id, out var idSlice))
+            {
+                return DocumentsStorage.Get(_context, idSlice, fields);
```
ppekrol: Why? Previosuly we did 'DocumentIdWorker.GetSliceFromId', now just 'Slice.From'

haludi: As I saw it the id here already escaped and converted to lower.
If the id has special char '\v' it becomes '\u000B'.
The `DocumentIdWorker.GetSliceFromId` function do toLower so '\u000B' becomes '\u000b' so the document not found.

src/Raven.Server/Documents/Indexes/Static/CurrentIndexingScope.cs
haludi: The main target was to escape the key so its document can be found.
In the beginning I just replace `Slice.From` whit `DocumentIdWorker.GetSliceFromId`.
```
Slice keySlice;
if (keyLazy != null)
{
    if (keyLazy.Length == 0)
        return DynamicNullObject.Null;

    // we intentionally don't dispose of the scope here, this is being tracked by the references
    // and will be disposed there.
   DocumentIdWorker.GetSliceFromId(_documentsContext, keyLazy, out var keySlice);
}
else
{
    if (keyString.Length == 0)
        return DynamicNullObject.Null;
    // we intentionally don't dispose of the scope here, this is being tracked by the references
    // and will be disposed there.
   DocumentIdWorker.GetSliceFromId(_documentsContext, keyString, out var keySlice);
}
```

So what do you me to revert? 
The common code for `LazyString` and `string`?
The use of `DocumentIdWorker.GetSliceFromId`?

src/Raven.Client/Extensions/StringExtensions.cs
```diff
@@ -58,5 +60,34 @@ public static bool IsIdentifier(string token, int start, int length)
 
             return true;
         }
+        
+        private static readonly char[] LiteralSymbolsToEscape = { '\'', '\"', '\\', '\a', '\b', '\f', '\n', '\r', '\t', '\v' };
+        private static readonly string[] LiteralEscapedSymbols = { @"\'", @"\""", @"\\", @"\a", @"\b", @"\f", @"\n", @"\r", @"\t", @"\v" };
+
+        public static string EscapeString(string value)
```
ayende: Make this pass a StringBuilder, you are doing needless allocations

src/Raven.Client/Extensions/StringExtensions.cs
```diff
@@ -58,5 +60,34 @@ public static bool IsIdentifier(string token, int start, int length)
...
+            var index = Array.IndexOf(LiteralSymbolsToEscape, c);
+
+            if (index != -1)
+                return LiteralEscapedSymbols[index];
+
+            if (char.IsLetterOrDigit(c) == false 
+                && char.IsWhiteSpace(c) == false 
+                && char.IsSymbol(c) == false 
+                && char.IsPunctuation(c) == false)
+                return @"\u" + ((int)c).ToString("x4");
```
ayende: More needless allocations, should use the string builder

src/Raven.Server/Documents/Indexes/Static/CurrentIndexingScope.cs
ayende: You are not calling to lower, which can cause issues.
Why allocate a string if we don't have to?

src/Sparrow/Json/Parsing/JsonParserState.cs
```diff
@@ -45,15 +45,19 @@ public static int VariableSizeIntSize(int value)
...
+            return FindEscapePositionsMaxSize(str.AsSpan(), out escapedCount);
+        }
+        
+        public static int FindEscapePositionsMaxSize(ReadOnlySpan<char> str, out int escapedCount)
         {
             var count = 0;
             var controlCount = 0;
 
-            for (int i = 0; i < str.Length; i++)
+            foreach (var value in str)
```
arekpalinski: Not sure if usage of `for` instead of `foreach` wasn't intentional because of slightly better performance. Let's get back to `for` just in case.

PR #11050, opened 2020-08-26
RavenDB-15532 Support for ARM64 build

ayende:


ayende:
@gregolsky I think we also need to update the downloads?

ppekrol:
What are the performance numbers for this? Last time when we checked it was slower than arm32 and we decided to wait till .NET Core catches up.

ayende:
We have users that want to run there because of memory limits, not related to performance.

ppekrol:
libsodium, zstd, etc are working?

ayende:
That said, import `users.dump` gives us:

image

image

This is on `a1.large` instance on AWS, 2 cores, 4 GB RAM.

8GB gp2 drive with 100 IOPS

ayende:
image


Inserting `posts-001.dump` in 00:00:29.1865386

ayende:
zstd works

ayende:
Encryption & Decryption works :-)

karmeli87:
One step closer to run on android? :-)

gregolsky:
need to wire up on CI and other release related things

ppekrol:
> One step closer to run on android? :-)

.NET 6 might help with that

build.ps1
```diff
@@ -6,6 +6,7 @@ param(
     [switch]$MacOs,
     [switch]$Osx,
     [switch]$Rpi,
+    [switch]$Arm64,
```
ppekrol: @gregolsky when dealing with this rename this to LinuxArm64, in .NET 5 we will have WindowsArm64 as well.

gregolsky: ok gotcha

PR #11056, opened 2020-08-27
RavenDB-15539 Allow user to tell the session to ignore changes from e…

ayende:
…ntities

### DIFF
--- a/src/Raven.Client/Documents/Conventions/DocumentConventions.cs
+++ b/src/Raven.Client/Documents/Conventions/DocumentConventions.cs
@@ -213,6 +213,8 @@ namespace Raven.Client.Documents.Conventions
         private bool _saveEnumsAsIntegers;
         private char _identityPartsSeparator;
         private bool _disableTopologyUpdates;
+
+        private Func<InMemoryDocumentSessionOperations, object, string, bool> _shouldIgnoreEntityChanges;
         private Func<MemberInfo, bool> _findIdentityProperty;
         private Func<string, string> _transformTypeCollectionNameToDocumentIdPrefix;
         private Func<string, object, Task<string>> _asyncDocumentIdGenerator;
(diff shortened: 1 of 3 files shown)


src/Raven.Client/Documents/Conventions/DocumentConventions.cs
```diff
@@ -252,6 +254,16 @@ public DocumentConventions()
         private Size _maxContextSizeToKeep;
         private ISerializationConventions _serialization;
 
+        public EventHandler<ShouldIgnoreEntityChangesEventArgs> ShouldIgnoreEntityChanges
```
arekpalinski: That is the first time we use `EventHandler / EventArgs` in our conventions. Why not to change to `Func` returning `bool`?

test/SlowTests/Issues/RavenDB-15539.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        [Fact]
+        public void CanIgnoreChanges()
+        {
+            using var store = GetDocumentStore(new Options
+            {
+                ModifyDocumentStore = s =>
+                {
+                    s.Conventions.ShouldIgnoreEntityChanges
+                        = (sender, args) => args.ShouldIgnoreChanges =args.Entity is User u && u.IgnoreChanges;
```
arekpalinski: For me it would be much more natural to just return `bool` here instead of assigning it to `ShouldIgnoreChanges`. The convention name is `ShouldIgnoreEntityChanges` anyway so the following is simpler:
```
 s.Conventions.ShouldIgnoreEntityChanges
                        => (id, entity) => entity is User u && u.IgnoreChanges;

```

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
ayende: There is a recommendation against returning bool from events. Because you may have multiple subscribers, and only the last one will be respected.

ayende: https://docs.microsoft.com/en-us/dotnet/csharp/event-pattern#returning-values-from-event-subscribers

ayende: That said, we are returning values from other events here

arekpalinski: Hmm.. okay, but the issue is that this is something totally different from the rest of our API.

Does it make sense to have multiple subscribers here? I think we were always assuming that there will be only one. Maybe we should explicitly forbid multiple subscribers and throw then.

ayende:
Updated to have the same sig as the other events

PR #11067, opened 2020-08-31
RavenDB-15427 Waiting for compressed file to be delete by time retention

haludi:
RavenDB-15427

### DIFF
--- a/test/SlowTests/SparrowTests/LoggingSourceTests.cs
+++ b/test/SlowTests/SparrowTests/LoggingSourceTests.cs
@@ -81,7 +81,7 @@ namespace SlowTests.SparrowTests
                     await logger.OperationsAsync("Some message");
                 }
                 afterEndFiles = Directory.GetFiles(path);
-                return afterEndFiles.Any(f => toCheckLogFiles.Any(tc => tc.fileName.Contains(f) && tc.shouldExist == false));
+                return IsContainsShouldNotExistFiles(afterEndFiles, toCheckLogFiles, compressing);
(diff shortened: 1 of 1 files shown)


grisha-kotler (approved):

test/SlowTests/SparrowTests/LoggingSourceTests.cs
```diff
@@ -81,7 +81,7 @@ await using (File.Create(fileName))
                     await logger.OperationsAsync("Some message");
                 }
                 afterEndFiles = Directory.GetFiles(path);
-                return afterEndFiles.Any(f => toCheckLogFiles.Any(tc => tc.fileName.Equals(f) && tc.shouldExist == false));
+                return afterEndFiles.Any(f => toCheckLogFiles.Any(tc => tc.fileName.Contains(f) && tc.shouldExist == false));
```
arekpalinski: @haludi Can you please explain why we need change change?

haludi: The `Assert`s below should be invoked after the retention policy is applied.
In the previous implementation, the test waits until there is no logfile with just the `.log` prefix but while in this test case the log files are compressed this check is not relevant so I changed it to wait for `.log.gz` log files are removed too

arekpalinski: Would `StartsWith()` work as well here? If yes then wouldn't it be a slightly safer check? Maybe let's add a comment explaining that logs are compressed here so they have `.log.gz` extension

PR #11069, opened 2020-09-01
RavenDB-15530 upon TS conflict prefer deleted values

karmeli87:


src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -458,11 +458,14 @@ private async Task ExecuteClusterTransactionTask()
                 var topology = ServerStore.LoadDatabaseTopology(Name);
                 if (topology.Promotables.Contains(ServerStore.NodeTag))
                 {
-                    await Task.Delay(1000);
+                    await Task.Delay(1000, DatabaseShutdown);
```
arekpalinski, line 461: That can throw `TaskCancelledException`. Is it ok? I see the code below just catches everything.

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -458,11 +458,14 @@ private async Task ExecuteClusterTransactionTask()
...
+                    if (DatabaseShutdown.IsCancellationRequested)
```
arekpalinski: Whats the point of checking it here since we're passing `DatabaseShutdown` one line above?

karmeli87: remvoed

src/Raven.Server/Documents/DocumentDatabase.cs
karmeli87: `TaskCancelledException` is caught higher in the stack by `catch (OperationCanceledException)`

arekpalinski (approved):

aviv86 (approved):

PR #11071, opened 2020-09-01
RavenDB-14033 Only test for unreproducible issue

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-14033

### DIFF
--- a/test/Tests.Infrastructure/RavenTestBase.cs
+++ b/test/Tests.Infrastructure/RavenTestBase.cs
@@ -628,7 +628,7 @@ namespace FastTests
                 }
                 catch
                 {
-                    if (sw.ElapsedMilliseconds <= timeout)
+                    if (sw.ElapsedMilliseconds > timeout)
                     {
                         throw;
                     }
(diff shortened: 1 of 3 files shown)


test/Tests.Infrastructure/RavenTestBase.cs
```diff
@@ -628,7 +628,7 @@ protected async Task<T> WaitForValueAsync<T>(Func<Task<T>> act, T expectedVal, i
                 }
                 catch
                 {
-                    if (sw.ElapsedMilliseconds <= timeout)
+                    if (sw.ElapsedMilliseconds > timeout)
```
arekpalinski, line 631: Why? We should throw then, shouldn't we?

haludi: I think the purpose here it not throw if the time of `timeout` not passed. 
What is the benefit of not throwing when the time passed?

arekpalinski: Sorry, there are two similar conditions there and I didn't notice that this one is inside `catch`.

PR #11072, opened 2020-09-02
RavenDB-10228 IO Stats graph doesn't look good when having a lot of gaps

kwiato:


src/Raven.Studio/wwwroot/Content/css/themes/light-theme-colors-override.less
```diff
@@ -1,3 +1,23 @@
...
+@gray-700: #8085A1;
+@gray-600: #999DB7;
+@gray-500: #AFB2C6;
+@gray-400: #C5C8D4;
+@gray-300: #DBDDE3;
+@gray-200: #E5E6EA;
+@gray-100: #f6f7f7;
+@white: #ffffff;
+
+/*@gray-base: @black;
```
ml054, line 13: remove?

kwiato: No, there is some problem with less compilation, the colors are not assigned correctly.  Hovewer this is the way I want it defined once we move to SCSS.

ml054: ok

ml054 (approved):

PR #11073, opened 2020-09-02
RavenDB-13548 Better auth message in chrome

Danielle9897:


src/Raven.Server/Web/System/StudioHandler.cs
```diff
@@ -152,9 +152,24 @@ public Task FavIcon()
         [RavenAction("/auth-error.html", "GET", AuthorizationStatus.UnauthenticatedClients)]
         public Task StudioAuthError()
         {
+            string reason;
+            string reasonAddon;
             var error = GetStringQueryString("err");
+
+            var stringBreak = error.IndexOf("<br>");
```
ml054: can we find better solution for that? This is really fragile approach.

Danielle9897: do you mean use some pre-defined constant instead of ```"<br>"``` ?

src/Raven.Server/Routing/RequestRouter.cs
```diff
@@ -375,12 +382,19 @@ private static void ThrowUnknownAuthStatus(RavenServer.AuthenticationStatus stat
             RavenServer.AuthenticateConnection feature,
             AuthorizationStatus authorizationStatus)
         {
+            TryGetUserAgent(context, out var userAgent);
+            var userAgentChrome = userAgent.Contains("chrome");
```
ayende: Why do we care? We can simply always include this

ayende: Note that the same message is releavnt for FireFox, Safari, etc

src/Raven.Server/Routing/RequestRouter.cs
```diff
@@ -477,5 +497,8 @@ private static void DrainRequest(JsonOperationContext ctx, HttpContext context)
                 }
             }
         }
+        
+        private const string ChromeCertificateMessage = "<br>Your certificate store may be cached by Chrome. " +
```
ayende: By the browser, instead of chome.
And the "private browsing tab, Ctrl+Shift+N in Chrome"

src/Raven.Server/Routing/RequestRouter.cs
ml054: good point! + we don't need this detection which is also fragile

src/Raven.Server/Web/System/StudioHandler.cs
ml054: in general we should escape those characters to avoid XSS.

src/Raven.Server/Routing/RequestRouter.cs
Danielle9897: np
(this chrome detection was inserted only because that was the issue definition)

src/Raven.Server/Web/System/StudioHandler.cs
Danielle9897: using `Environment.NewLine` instead

src/Raven.Server/Routing/RequestRouter.cs
Danielle9897: Changed to:

`Your certificate store may be cached by the browser. Create a new private browsing tab, which will not cache any certificates. (Ctrl+Shift+N in Chrome, Ctrl+Shift+P in Firefox).`

src/Raven.Server/Web/HtmlUtil.cs
```diff
@@ -65,8 +65,23 @@ private static string GetUnsecuredAccessFlagsHtml()
             return unsecuredAccessFlagsHtml;
         }
         
-        public static string RenderStudioAuthErrorPage(string reason, string reasonAddon)
+        public static string RenderStudioAuthErrorPage(string error)
         {
+            string reason;
+            string reasonAddon;
+            
+            var stringBreak = error.IndexOf(Environment.NewLine);
```
ml054: Do we need to break this at all? Can't we simply display this and single text? Such solutions are hard to maintain in future. Btw. you can split text using .Split method on string.

Danielle9897: 1. The break was added because it seemed more nice & clear like this.
   See  image:  https://issues.hibernatingrhinos.com/issue/RavenDB-13548#focus=Comments-67-50351.0-0
2. I didn't want to use split because I only wanted to get 2 parts for the html,
   and not an array of strings in case someone added more than one 'break' in the message text.

ml054: split has paramater to limit items count

Danielle9897: awesome - done

ml054 (approved):

PR #11074, opened 2020-09-02
RavenDB-15516 sinks should not affect hubs global change vector

karmeli87:
Note that we do not change the change vector of the document itself, but only the global database change vector.

### DIFF
--- a/src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
+++ b/src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
@@ -9,6 +9,7 @@ using System.Net.Sockets;
 using System.Threading;
 using System.Threading.Tasks;
 using Raven.Client.Documents.Operations.Attachments;
+using Raven.Client.Documents.Operations.Replication;
 using Raven.Client.Documents.Replication;
 using Raven.Client.Documents.Replication.Messages;
 using Raven.Client.Extensions;
@@ -66,6 +67,7 @@ namespace Raven.Server.Documents.Replication
         private IncomingReplicationStatsAggregator _lastStats;
 
         public readonly string PullReplicationName;
+        public readonly PullReplicationMode Mode;
 
         public bool PullReplication => PullReplicationName != null;
         private readonly DisposeOnce<SingleAttempt> _disposeOnce;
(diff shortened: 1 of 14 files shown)


src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -1302,6 +1331,70 @@ protected override long ExecuteCmd(DocumentsOperationContext context)
                 }
             }
 
+            private static string ReplaceUnknownEntriesWithSinkTag(DocumentsOperationContext context, ref string changeVector)
+            {
+                var globalDbIds = context.LastDatabaseChangeVector?.ToChangeVectorList()?.Select(x => x.DbId).ToList();
```
ayende, line 1336: Don't you also need `?.ToList(); `?

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -1302,6 +1331,70 @@ protected override long ExecuteCmd(DocumentsOperationContext context)
...
+                }
+
+                changeVector = newIncoming.SerializeVector();
+
+                return knownEntries.Count > 0 ? 
+                    knownEntries.SerializeVector() : 
+                    null;
+            }
+
+            private static void ReplaceKnownSinkEntries(DocumentsOperationContext context, ref string changeVector)
```
ayende, line 1366: Why do you need this?

ayende (commented):
Also need to handle large change vectors, no?

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
karmeli87: consider the following: (A - sink, B - hub)

1. doc create on sink (A:1- ... )
2. replicated to hub (SINK:1- ..)
3. modified on the the hub (SINK:1- .., B:1- ... )
4. replicated back to the sink (A:1-.., B:1-..)

I would like to avoid the tag SINK on the origin.

ayende: We compare the db id, so no issue there.
And not a big worry from our perspective, because sink B will have the SINK from node A anyway, no?

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
karmeli87: I don't think that `Select(x => x.DbId)` can generate `null`

karmeli87:
> Also need to handle large change vectors, no?

I think we should handle large change vectors separately. 
I'm still not sure what would be the best solution.

ayende:
Sort by etag descending, chop the bottom half once it is past 512 bytes ?

arekpalinski:
@ayende Could you check this PR after Karmel's fixes?

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
arekpalinski: Correct. It can return empty enumerable but not null.

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
arekpalinski: @karmeli87 ?

karmeli87: I would like to avoid a situation where the sink node have the `SINK` tag on its own dbid.
I agree it is more a cosmetic issue, since we compare dbids, but it is "cleaner".

arekpalinski:
@ayende I'm merging it since we need that changes for testing. If you have more comment they will be addressed in separate PR.

ayende:
Looks good to me. Still need to do some more tests here.

PR #11078, opened 2020-09-03
RavenDB-15538: Missing attachment break replication

garayx:


grisha-kotler:
Can you add a test?

garayx:
> Can you add a test?

https://issues.hibernatingrhinos.com/issue/RavenDB-15538#focus=Comments-67-50344.0-0

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
```diff
@@ -213,24 +213,82 @@ public bool ExecuteReplicationOnce(TcpConnectionOptions tcpConnectionOptions, Ou
...
+                                docItem.Flags.Contain(DocumentFlags.HasAttachments))
                             {
+                                var type = (docItem.Flags & DocumentFlags.Revision) == DocumentFlags.Revision ? AttachmentType.Revision : AttachmentType.Document;
+                                foreach (var attachment in _parent._database.DocumentsStorage.AttachmentsStorage.GetAttachmentsForDocument(documentsContext, type, docItem.Id, docItem.ChangeVector))
+                                {
+                                    // we need to filter attachments that are been sent in the same batch as the document
+                                    if (attachment.Etag >= prevLastEtag)
+                                    {
+                                        Debug.Assert(docItem.TransactionMarker == attachment.TransactionMarker, $"{nameof(docItem)}.{nameof(docItem.TransactionMarker)} == {nameof(attachment)}.{nameof(attachment.TransactionMarker)}");
+                                        // treat the rare case when revision's attachment was created under different TransactionMarker
```
arekpalinski, line 243: Why is that a _valid_ case? When can that happen?

garayx: We got it on few of our DBs I currently work on reproducing it in a test, but it seems to be OK to have different tx marker for document and its attachment, but should not happen for revision and its attachment?

ayende: A revision's attachment can be added by a previous revision, so that would be different.

ppekrol:
Status?

grisha-kotler:
@garayx is working on a test

PR #11082, opened 2020-09-06
RavenDB-15564 Getting a distinct numeric (decimal) value will try to …

ayende:
…parse the returned string using the local culture


grisha-kotler:
Can we have a test for this?

ayende (commented):
LGTM

Issue #11064, opened 2020-08-28
OnAfterConversationToEntity not called for the same session where it was registered


mcrio:
Hi,

Not sure the described behavior is expected. Please can you check.

Subscribing to `OnAfterConversationToEntity` using the Session object:`session.Advanced.DocumentStore.OnAfterConversionToEntity`  does not fire the event listener for that session.

The event listener does get executed for the next session created.

**Test**

Listener1 is registered directly on the `store` object.
Listener2 is registered by using `session.Advanced`.
```
    public class RavenDbTest : RavenTestDriver
    {
        private readonly ITestOutputHelper _output;

        public RavenDbTest(ITestOutputHelper output)
        {
            _output = output;
        }

        [Fact]
        public async Task ShouldCallEventListeners()
        {
            var listenerOneCalled = 0;
            var listenerTwoCalled = 0;

            IDocumentStore store = GetDocumentStore();

            // Register listener 1
            store.OnAfterConversionToEntity += (sender, args) =>
            {
                _output.WriteLine("incrementing listenerOne");
                listenerOneCalled += 1;
            };

            // Insert data
            {
                IAsyncDocumentSession session = store.OpenAsyncSession();
                await session.StoreAsync(new Entity("bob", "bob"));
                await session.SaveChangesAsync();
            }

            // FIRST LOAD
            {
                IAsyncDocumentSession session = store.OpenAsyncSession();

                // Register listener 2
                session.Advanced.DocumentStore.OnAfterConversionToEntity += (sender, args) =>
                {
                    _output.WriteLine("incrementing listenerTwo");
                    listenerTwoCalled += 1;
                };

                _output.WriteLine("Executing 1st LoadAsync()");
                Entity entity = await session.LoadAsync<Entity>("bob");
                entity.Id.Should().Be("bob");
            }

            // SECOND LOAD
            {
                IAsyncDocumentSession session = store.OpenAsyncSession();

                _output.WriteLine("Executing 2nd LoadAsync()");
                Entity entity = await session.LoadAsync<Entity>("bob");
                entity.Id.Should().Be("bob");
            }

            listenerOneCalled.Should().Be(2, "we made 2 calls to LoadAsync()");
            listenerTwoCalled.Should().Be(2, "we made 2 calls to LoadAsync()");
        }

        public class Entity
        {
            public Entity(string id, string name)
            {
                Id = id;
                Name = name;
            }

            private Entity()
            {
            }

            public string Id { get; }

            public string Name { get; }
[...]

PR #11083, opened 2020-09-06
RavenDB-15554 OnAfterConversionToEntity is missing on the session (on…

ayende:
…ly on document store)


arekpalinski:
Can we add a test here that the user sent us - https://github.com/ravendb/ravendb/issues/11064 ?

arekpalinski:
I will add it

arekpalinski:
@ayende I've tried to add the test that user has provided but I see his assumption wasn't valid. Not sure what test should be added to validate what you added here.

ayende:
Maybe add separate handling just for that sesion?

PR #11085, opened 2020-09-06
RavenDB-15041: Add ConditionalLoad and ConditionalLoadAsync methods t…

garayx:
…o Client API

### DIFF
--- a/src/Raven.Client/Documents/Session/AsyncDocumentSession.ConditionalLoad.cs
+++ b/src/Raven.Client/Documents/Session/AsyncDocumentSession.ConditionalLoad.cs
@@ -1,0 +1,52 @@
+using System;
+using System.Net;
+using System.Threading;
+using System.Threading.Tasks;
+using Raven.Client.Documents.Commands;
+using Sparrow.Json;
+
+namespace Raven.Client.Documents.Session
+{
+    public partial class AsyncDocumentSession
+    {
+        /// <inheritdoc />
+        public async Task<(T Entity, string ChangeVector)> ConditionalLoadAsync<T>(string id, string changeVector, CancellationToken token = default)
+        {
+            using (AsyncTaskHolder())
+            {
+                if (id == null)
+                    throw new ArgumentNullException(nameof(id));
+
+                if (Advanced.IsLoaded(id))
+                {
+                    var e = await LoadAsync<T>(id, token).ConfigureAwait(false);
+                    var cv = Advanced.GetChangeVectorFor(e);
+                    return (e, cv);
+                }
+
+                if (changeVector == null)
+                    return default;
+
+                IncrementRequestCount();
+                var cmd = new ConditionalGetDocumentsCommand(id, changeVector);
+                await Advanced.RequestExecutor.ExecuteAsync(cmd, Advanced.Context, sessionInfo: null, token).ConfigureAwait(false);
+
+                switch (cmd.StatusCode)
+                {
+                    case HttpStatusCode.NotModified:
+                        return (default, changeVector); // value not changed
+                    case HttpStatusCode.NotFound:
+                        return default; // value is missing
+                }
+
+                if (cmd.Result.Results.Length == 0)
+                    return (default, cmd.Result.ChangeVector);
+
+                var documentInfo = DocumentInfo.GetNewDocumentInfo((BlittableJsonReaderObject)cmd.Result.Results[0]);
+                var r = TrackEntity<T>(documentInfo);
+
+                return (r, cmd.Result.ChangeVector);
+            }
+        }
+    }
+}
--- a/src/Raven.Client/Documents/Session/AsyncDocumentSession.ConditionalLoad.cs
+++ b/src/Raven.Client/Documents/Session/AsyncDocumentSession.ConditionalLoad.cs
@@ -25,7 +25,7 @@ namespace Raven.Client.Documents.Session
                 }
 
                 if (string.IsNullOrEmpty(changeVector))
-                    throw new InvalidOperationException($"The requested document with id '{id} is not loaded in to the session and could not conditional load when {nameof(changeVector)} is null or empty.");
+                    throw new InvalidOperationException($"The requested document with id '{id} is not loaded into the session and could not conditional load when {nameof(changeVector)} is null or empty.");
 
                 IncrementRequestCount();
                 var cmd = new ConditionalGetDocumentsCommand(id, changeVector);
--- a/src/Raven.Client/Documents/Session/AsyncDocumentSession.ConditionalLoad.cs
+++ b/src/Raven.Client/Documents/Session/AsyncDocumentSession.ConditionalLoad.cs
@@ -14,7 +14,7 @@ namespace Raven.Client.Documents.Session
         {
             using (AsyncTaskHolder())
             {
-                if (id == null)
+                if (string.IsNullOrEmpty(id))
                     throw new ArgumentNullException(nameof(id));
 
                 if (Advanced.IsLoaded(id))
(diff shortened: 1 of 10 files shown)


src/Raven.Client/Documents/Session/AsyncDocumentSession.ConditionalLoad.cs
```diff
@@ -0,0 +1,52 @@
...
+                    throw new ArgumentNullException(nameof(id));
+
+                if (Advanced.IsLoaded(id))
+                {
+                    var e = await LoadAsync<T>(id, token).ConfigureAwait(false);
+                    var cv = Advanced.GetChangeVectorFor(e);
+                    return (e, cv);
+                }
+
+                if (changeVector == null)
```
ayende: I think we want to throw that here. Or just continue.
Thoughts?

ayende: Right now we are returning that the item does not exists

src/Raven.Client/Documents/Session/AsyncDocumentSession.ConditionalLoad.cs
```diff
@@ -0,0 +1,52 @@
...
+
+                switch (cmd.StatusCode)
+                {
+                    case HttpStatusCode.NotModified:
+                        return (default, changeVector); // value not changed
+                    case HttpStatusCode.NotFound:
+                        return default; // value is missing
+                }
+
+                if (cmd.Result.Results.Length == 0)
```
ayende: Under what scenario that can happen?

test/FastTests/Client/ConditionalLoadTest.cs
```diff
@@ -0,0 +1,200 @@
...
+                    cv = newSession.Advanced.GetChangeVectorFor(user);
+                    Assert.NotNull(user);
+                    Assert.Equal(user.Name, "RavenDB");
+                    user.Name = "RavenDB 5.1";
+                    newSession.SaveChanges();
+                }
+
+                using (var newestSession = store.OpenSession())
+                {
+                    var user = newestSession.ConditionalLoad<User>("users/1", cv);
```
ayende: This should be under `Advanced`, not at the top

ayende (changes requested):

src/Raven.Client/Documents/Session/AsyncDocumentSession.ConditionalLoad.cs
garayx: removed

src/Raven.Client/Documents/Session/AsyncDocumentSession.ConditionalLoad.cs
garayx: Currently we allow to check if the entity loaded into session with `changeVector=null`
I think if it's not loaded then we should throw an exception

src/Raven.Client/Documents/Session/IAdvancedSessionOperations.ConditionalLoad.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Raven.Client.Documents.Session
+{
+    /// <summary>
+    ///     Advanced synchronous session operations
+    /// </summary>
+    public partial interface IAdvancedSessionOperations
+    {
+        /// <summary>
+        ///     Conditional load the specified entity with the specified id and changeVector.
```
arekpalinski: I think we need to explain what is the condition actually. So user will know that if etag matches it won't send document over the wire.

src/Raven.Client/Documents/Session/AsyncDocumentSession.ConditionalLoad.cs
```diff
@@ -0,0 +1,49 @@
...
+
+                if (Advanced.IsLoaded(id))
+                {
+                    var e = await LoadAsync<T>(id, token).ConfigureAwait(false);
+                    var cv = Advanced.GetChangeVectorFor(e);
+                    return (e, cv);
+                }
+
+                if (string.IsNullOrEmpty(changeVector))
+                    throw new InvalidOperationException($"The requested document with id '{id} is not loaded in to the session and could not conditional load when {nameof(changeVector)} is null or empty.");
```
arekpalinski: in to -> into

arekpalinski:
And please resolve the conflict

arekpalinski (approved):

src/Raven.Client/Documents/Session/AsyncDocumentSession.ConditionalLoad.cs
```diff
@@ -0,0 +1,49 @@
...
+                IncrementRequestCount();
+                var cmd = new ConditionalGetDocumentsCommand(id, changeVector);
+                await Advanced.RequestExecutor.ExecuteAsync(cmd, Advanced.Context, sessionInfo: null, token).ConfigureAwait(false);
+
+                switch (cmd.StatusCode)
+                {
+                    case HttpStatusCode.NotModified:
+                        return (default, changeVector); // value not changed
+                    case HttpStatusCode.NotFound:
+                        return default; // value is missing
```
grisha-kotler, line 39: Should this be added to `TrackEntity` as we do for `Load`?
@ayende

grisha-kotler (changes requested):

src/Raven.Client/Documents/Session/AsyncDocumentSession.ConditionalLoad.cs
ayende: Why, that is why we have `IsLoaded` for, no?

src/Raven.Client/Documents/Session/AsyncDocumentSession.ConditionalLoad.cs
ayende: I think so, we now know it doesn't exist. 
It should be added to `missing`, no?

garayx: sending 'Load' operation 2 times with the same non-existent ID will send 2 requests to the server, therefore it is not added to `TrackEntity` if the the entity doesn't exists on server

arekpalinski: Why do you think the 2nd call of `ConditionalLoadAsync` with the same non-existent ID should sent the request to the server. We already know that it doesn't exist, don't we? So we should answer immediately without asking the server.

ayende: Yes, it should update `_knownMissingIds` and for that matter, the `ConditionalLoad` should also check it.

grisha-kotler: `_knownMissingIds` is used in `Delete` and `Include` pnly.
If you `Load` an entity with the same id twice (non-existing one) it will generate 2 requests.
We should probably call `RegisterMissing` for `Load` a well.

ayende: That isn't how it works. `LoadOperation` calls to `CheckIfIdAlreadyIncluded` which handles this.
No two calls here.

ayende: Oh, we aren't _registering_ in the known missing in `Load`, yes, we need that too.

garayx: https://issues.hibernatingrhinos.com/issue/RavenDB-15615

I will add it to the ConditionalLoad as well

PR #11086, opened 2020-09-06
RavenDB-15300 Creating new server-wide backup while only updating the name fixed

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15300

### DIFF
--- a/src/Raven.Client/ServerWide/Operations/Configuration/ServerWideBackupConfiguration.cs
+++ b/src/Raven.Client/ServerWide/Operations/Configuration/ServerWideBackupConfiguration.cs
@@ -6,7 +6,6 @@ namespace Raven.Client.ServerWide.Operations.Configuration
     public class ServerWideBackupConfiguration : PeriodicBackupConfiguration
     {
         internal static string NamePrefix = "Server Wide Backup";
-        internal string ToRemoveFromDatabaseRecord { get; set; }
 
         public override DynamicJsonValue ToJson()
         {
--- a/src/Raven.Client/ServerWide/Operations/Configuration/ServerWideBackupConfiguration.cs
+++ b/src/Raven.Client/ServerWide/Operations/Configuration/ServerWideBackupConfiguration.cs
@@ -6,7 +6,8 @@ namespace Raven.Client.ServerWide.Operations.Configuration
     public class ServerWideBackupConfiguration : PeriodicBackupConfiguration
     {
         internal static string NamePrefix = "Server Wide Backup";
-        
+        internal string ToRemoveFromDatabaseRecord { get; set; }
+
         public override DynamicJsonValue ToJson()
         {
             var json = base.ToJson();
--- a/src/Raven.Server/ServerWide/ClusterStateMachine.cs
+++ b/src/Raven.Server/ServerWide/ClusterStateMachine.cs
@@ -3063,7 +3063,9 @@ namespace Raven.Server.ServerWide
 
             const string dbKey = "db/";
             var toUpdate = new List<(string Key, BlittableJsonReaderObject DatabaseRecord, string DatabaseName)>();
-            long? oldTaskId = null;
+            (long? TaskId, string name) old = default;
+            
+            var allSeverWideBackupNames = GetSeverWideBackupNames(context);
 
             using (Slice.From(context.Allocator, dbKey, out var loweredPrefix))
             {
@@ -3080,34 +3082,21 @@ namespace Raven.Server.ServerWide
 
                     if (oldDatabaseRecord.TryGet(nameof(DatabaseRecord.PeriodicBackups), out BlittableJsonReaderArray backups))
                     {
+                        var editedConfigFound = false;
                         foreach (BlittableJsonReaderObject backup in backups)
                         {
-                            if (backup.TryGet(nameof(PeriodicBackupConfiguration.Name), out string backupName)  == false || backupName == null)
-                                throw new RachisInvalidOperationException(
-                                    $"This {nameof(PeriodicBackupConfiguration)} of {result.Key} has no value for {nameof(PeriodicBackupConfiguration.Name)}. This should not happens\n{backup}");
-                            
-                            if (serverWideBackupConfiguration.ToRemoveFromDatabaseRecord != null)
-                            {
-                                var toRemove = PutServerWideBackupConfigurationCommand.GetTaskNameForDatabase(serverWideBackupConfiguration.ToRemoveFromDatabaseRecord);
-                                if(backupName.Equals(toRemove, StringComparison.OrdinalIgnoreCase))
-                                    continue;
-                            }
-                                
-                            if(backupName.Equals(periodicBackupConfiguration.Name, StringComparison.OrdinalIgnoreCase))
-                            {
-                                if (backup.TryGet(nameof(PeriodicBackupConfiguration.TaskId), out long taskId))
-                                    oldTaskId = taskId;
-
+                            //Even though we rebuild the whole list of the configurations just one should be modified
+                            //In addition the same configuration should be modified in all databases  
+                            if (editedConfigFound == false && CheckIfEditedConfig(backup, periodicBackupConfiguration, allSeverWideBackupNames, ref old, ref editedConfigFound))
                                 continue;
-                            }
-
+                            
                             newBackups.Add(backup);
                         }
                     }
 
                     using (oldDatabaseRecord)
                     {
-                        periodicBackupConfiguration.TaskId = oldTaskId ?? serverWideBackupConfiguration.TaskId;
+                        periodicBackupConfiguration.TaskId = old.TaskId ?? serverWideBackupConfiguration.TaskId;
                         newBackups.Add(periodicBackupConfiguration.ToJson());
 
                         oldDatabaseRecord.Modifications = new DynamicJsonValue(oldDatabaseRecord)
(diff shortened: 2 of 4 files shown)


src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -3082,7 +3082,18 @@ private void UpdateDatabasesWithNewServerWideBackupConfiguration(TransactionOper
...
                         foreach (BlittableJsonReaderObject backup in backups)
                         {
-                            if (IsServerWideBackupWithTaskName(backup, periodicBackupConfiguration.Name))
+                            if (backup.TryGet(nameof(PeriodicBackupConfiguration.Name), out string backupName)  == false || backupName == null)
+                                throw new RachisInvalidOperationException(
+                                    $"This {nameof(PeriodicBackupConfiguration)} of {result.Key} has no value for {nameof(PeriodicBackupConfiguration.Name)}. This should not happens\n{backup}");
+                            
+                            if (serverWideBackupConfiguration.ToRemoveFromDatabaseRecord != null)
+                            {
+                                var toRemove = PutServerWideBackupConfigurationCommand.GetTaskNameForDatabase(serverWideBackupConfiguration.ToRemoveFromDatabaseRecord);
```
grisha-kotler: You can move it up to avoid the `serverWideBackupConfiguration.ToRemoveFromDatabaseRecord != null` check

grisha-kotler (changes requested):

src/Raven.Client/ServerWide/Operations/Configuration/ServerWideBackupConfiguration.cs
```diff
@@ -6,7 +6,8 @@ namespace Raven.Client.ServerWide.Operations.Configuration
     public class ServerWideBackupConfiguration : PeriodicBackupConfiguration
     {
         internal static string NamePrefix = "Server Wide Backup";
-        
+        internal string ToRemoveFromDatabaseRecord { get; set; }
```
arekpalinski: Why do we need this? How do we handle update of a name of regular (not server wide) backup tasks?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -3082,7 +3082,18 @@ private void UpdateDatabasesWithNewServerWideBackupConfiguration(TransactionOper
                     {
                         foreach (BlittableJsonReaderObject backup in backups)
                         {
-                            if (IsServerWideBackupWithTaskName(backup, periodicBackupConfiguration.Name))
```
arekpalinski, line 3085: 1) Shouldn't this method ensure that the configuration name starts with `ServerWideBackupConfiguration.NamePrefix` ?
2) You removed the usage of this method here but there is another one in `DeleteServerWideBackupConfigurationFromAllDatabases` is its usage correct there?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -3082,7 +3082,18 @@ private void UpdateDatabasesWithNewServerWideBackupConfiguration(TransactionOper
...
+                                    $"This {nameof(PeriodicBackupConfiguration)} of {result.Key} has no value for {nameof(PeriodicBackupConfiguration.Name)}. This should not happens\n{backup}");
+                            
+                            if (serverWideBackupConfiguration.ToRemoveFromDatabaseRecord != null)
+                            {
+                                var toRemove = PutServerWideBackupConfigurationCommand.GetTaskNameForDatabase(serverWideBackupConfiguration.ToRemoveFromDatabaseRecord);
+                                if(backupName.Equals(toRemove, StringComparison.OrdinalIgnoreCase))
+                                    continue;
+                            }
+                                
+                            if(backupName.Equals(periodicBackupConfiguration.Name, StringComparison.OrdinalIgnoreCase))
```
arekpalinski: Shouldn't we rely on `TaskId` instead of `Name` to detect that this is an update?

src/Raven.Client/ServerWide/Operations/Configuration/ServerWideBackupConfiguration.cs
arekpalinski: Additionally it's hard to know what is it actually about - is about TaskId to remove or TaskName to remove.

haludi: In the database record, the `TaskId` is constant while in the server-wide configuration the `TaskId` change on any update to the configuration so we can't use the `TaskId` from the server-wide to know which configuration to edit (remove) in the database record (usually, we use the name but it is not useful when the user edits the name).

One option is to change this behavior and make the `TaskId` of server-wide backup constant as well.
This kind of change will not "help" to already exist server-wide backup configuration which already has different `TaskId` from database record `TaskId`. In this solution, we maybe can use a scheme upgrade.

In the current solution, I am using the name from the server-wide configuration update. This doesn't need any scheme upgrade

src/Raven.Server/ServerWide/ClusterStateMachine.cs
haludi: The `TaskId` from the server-wide backup can be different from the `TaskId` from the database records
https://github.com/ravendb/ravendb/pull/11086#discussion_r484883827

src/Raven.Server/ServerWide/ClusterStateMachine.cs
haludi: I didn't remove the behavior of the method just the usage.
I just wanted to use the backup name twice for the remove check and for the edit check
and while the code in `IsServerWideBackupWithTaskName` is very small so there is a small benefit of reuse it.

By the way, I thought to get the name as `LazyStringValue` but I didn't find an overload of `LazyStringValue.Equal` with `StringComparison` and I don't know if it so relevant here

src/Raven.Client/ServerWide/Operations/Configuration/ServerWideBackupConfiguration.cs
arekpalinski: Do I understand correctly that the server wide backup configuration has _constant_ `TaskId` in the database record? Can you please point me to the relevant code? Why is that?

@grisha-kotler What do you think?

haludi: Here https://github.com/ravendb/ravendb/pull/11086/files#diff-ad98abd3baf8509e485969ace15cd336R3098
```
if (backup.TryGet(nameof(PeriodicBackupConfiguration.TaskId), out long taskId))
                                    oldTaskId = taskId;
```

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -3080,25 +3082,27 @@ private void UpdateDatabasesWithNewServerWideBackupConfiguration(TransactionOper
...
+                            isBackupToEditFound = true;
+
+                            if (backup.TryGet(nameof(PeriodicBackupConfiguration.TaskId), out long taskId))
+                                periodicBackupConfiguration.TaskId = taskId;
                         }
                     }
 
                     using (oldDatabaseRecord)
                     {
-                        periodicBackupConfiguration.TaskId = oldTaskId ?? index;
```
grisha-kotler, line 3099: If it's a new one, you are missing the index `?? index`

grisha-kotler (changes requested):

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -3110,7 +3114,49 @@ private void UpdateDatabasesWithNewServerWideBackupConfiguration(TransactionOper
                 }
             }
 
-            ApplyDatabaseRecordUpdates(toUpdate, type, oldTaskId ?? index, index, items, context);
+            ApplyDatabaseRecordUpdates(toUpdate, type, old.TaskId ?? serverWideBackupConfiguration.TaskId, serverWideBackupConfiguration.TaskId, items, context);
+        }
+
+        private static bool IsBackupToEdit(BlittableJsonReaderObject backup,
```
arekpalinski:
```suggestion
        private static bool IsServerWideBackupToEdit(BlittableJsonReaderObject backup, 
```

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -3110,7 +3114,49 @@ private void UpdateDatabasesWithNewServerWideBackupConfiguration(TransactionOper
...
+
+            if (backupName.StartsWith(ServerWideBackupConfiguration.NamePrefix) == false) 
+                return false;
+            
+            if (backupName.Equals(periodicBackupConfigName, StringComparison.OrdinalIgnoreCase))
+            {
+                //server-wide backup to update when name is not modified
+                return true;
+            } 
+            if(severWideBackupNames.Contains(backupName) == false)
```
arekpalinski: minor - space after if

grisha-kotler (approved):

src/Raven.Server/ServerWide/ClusterStateMachine.cs
haludi: Instead of using `index` I used `serverWideBackupConfiguration.TaskId`

arekpalinski (approved):

PR #11093, opened 2020-09-08
RavenDB-15524 Min swap size configuration

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15524

### DIFF
--- a/src/Raven.Server/Config/Categories/MemoryConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/MemoryConfiguration.cs
@@ -20,8 +20,6 @@ namespace Raven.Server.Config.Categories
             UseTotalDirtyMemInsteadOfMemUsage = PlatformDetails.RunningOnDocker;
 
             EnableHighTemporaryDirtyMemoryUse = false; //MemoryInformation.TotalPhysicalMemory.GetValue(SizeUnit.Gigabytes) >= 2;
-
-            MinSwapSize = MemoryInformation.TotalPhysicalMemory;
         }
 
         [Description("The minimum amount of available memory RavenDB will attempt to achieve (free memory lower than this value will trigger low memory behavior)")]
@@ -73,11 +71,5 @@ namespace Raven.Server.Config.Categories
         [TimeUnit(TimeUnit.Seconds)]
         [ConfigurationEntry("Memory.TemporaryDirtyMemoryChecksPeriodInSec", ConfigurationEntryScope.ServerWideOnly)]
         public TimeSetting TemporaryDirtyMemoryChecksPeriod { get; set; }
-        
-        [Description("The minimum swap size. If the swap size is lower a notification will arise")]
-        [DefaultValue(DefaultValueSetInConstructor)]
-        [SizeUnit(SizeUnit.Megabytes)]
-        [ConfigurationEntry("Memory.MinSwapSize", ConfigurationEntryScope.ServerWideOnly)]
-        public Size MinSwapSize { get; set; }
     }
 }
--- a/src/Raven.Server/Config/Categories/MemoryConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/MemoryConfiguration.cs
@@ -20,6 +20,8 @@ namespace Raven.Server.Config.Categories
             UseTotalDirtyMemInsteadOfMemUsage = PlatformDetails.RunningOnDocker;
 
             EnableHighTemporaryDirtyMemoryUse = false; //MemoryInformation.TotalPhysicalMemory.GetValue(SizeUnit.Gigabytes) >= 2;
+
+            MinSwapSize = MemoryInformation.TotalPhysicalMemory;
         }
 
         [Description("The minimum amount of available memory RavenDB will attempt to achieve (free memory lower than this value will trigger low memory behavior)")]
@@ -71,5 +73,11 @@ namespace Raven.Server.Config.Categories
         [TimeUnit(TimeUnit.Seconds)]
         [ConfigurationEntry("Memory.TemporaryDirtyMemoryChecksPeriodInSec", ConfigurationEntryScope.ServerWideOnly)]
         public TimeSetting TemporaryDirtyMemoryChecksPeriod { get; set; }
+        
+        [Description("The minimum swap size. If the swap size is lower a notification will arise")]
+        [DefaultValue(DefaultValueSetInConstructor)]
+        [SizeUnit(SizeUnit.Megabytes)]
+        [ConfigurationEntry("Memory.MinSwapSize", ConfigurationEntryScope.ServerWideOnly)]
+        public Size MinSwapSize { get; set; }
     }
 }
--- a/src/Raven.Server/Config/Categories/PerformanceHintsConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/PerformanceHintsConfiguration.cs
@@ -2,11 +2,16 @@ using System.ComponentModel;
 using Raven.Server.Config.Attributes;
 using Raven.Server.Config.Settings;
 using Sparrow;
+using Sparrow.LowMemory;
 
 namespace Raven.Server.Config.Categories
 {
     public class PerformanceHintsConfiguration : ConfigurationCategory
     {
+        public PerformanceHintsConfiguration()
+        {
+            MinSwapSize = MemoryInformation.TotalPhysicalMemory;
+        }
         [Description("The size of a document after which it will get into the huge documents collection")]
         [DefaultValue(5)]
         [SizeUnit(SizeUnit.Megabytes)]
@@ -33,5 +38,11 @@ namespace Raven.Server.Config.Categories
         [DefaultValue(30)]
         [TimeUnit(TimeUnit.Seconds)]
         public TimeSetting TooLongRequestThreshold { get; set; }
+        
+        [Description("The minimum swap size. If the swap size is lower a notification will arise")]
+        [DefaultValue(DefaultValueSetInConstructor)]
+        [SizeUnit(SizeUnit.Megabytes)]
+        [ConfigurationEntry("PerformanceHints.Memory.MinSwapSizeInMb", ConfigurationEntryScope.ServerWideOnly)]
+        public Size MinSwapSize { get; set; }
     }
 }
--- a/src/Raven.Server/Config/Categories/PerformanceHintsConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/PerformanceHintsConfiguration.cs
@@ -47,7 +47,7 @@ namespace Raven.Server.Config.Categories
         [TimeUnit(TimeUnit.Seconds)]
         public TimeSetting TooLongRequestThreshold { get; set; }
         
-        [Description("The minimum swap size. If the swap size is lower a notification will arise")]
+        [Description("The minimum swap size (for linux only). If the swap size is lower a notification will arise")]
         [DefaultValue(DefaultValueSetInConstructor)]
         [SizeUnit(SizeUnit.Megabytes)]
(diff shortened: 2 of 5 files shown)


src/Raven.Server/Config/Categories/MemoryConfiguration.cs
```diff
@@ -71,5 +73,11 @@ public MemoryConfiguration()
         [TimeUnit(TimeUnit.Seconds)]
         [ConfigurationEntry("Memory.TemporaryDirtyMemoryChecksPeriodInSec", ConfigurationEntryScope.ServerWideOnly)]
         public TimeSetting TemporaryDirtyMemoryChecksPeriod { get; set; }
+        
+        [Description("The minimum swap size. If the swap size is lower a notification will arise")]
+        [DefaultValue(DefaultValueSetInConstructor)]
+        [SizeUnit(SizeUnit.Megabytes)]
+        [ConfigurationEntry("Memory.MinSwapSize", ConfigurationEntryScope.ServerWideOnly)]
+        public Size MinSwapSize { get; set; }
```
arekpalinski: - let's put it in `PerformanceHintsConfiguration.cs` file
- configuration key - `"PerformanceHints.Memory.MinSwapSizeInMb"`

src/Raven.Server/Config/Categories/MemoryConfiguration.cs
```diff
@@ -20,6 +20,8 @@ public MemoryConfiguration()
             UseTotalDirtyMemInsteadOfMemUsage = PlatformDetails.RunningOnDocker;
 
             EnableHighTemporaryDirtyMemoryUse = false; //MemoryInformation.TotalPhysicalMemory.GetValue(SizeUnit.Gigabytes) >= 2;
+
+            MinSwapSize = MemoryInformation.TotalPhysicalMemory;
```
arekpalinski: Isn't it too much? The recommendation in the issue (https://issues.hibernatingrhinos.com/issue/RavenDB-15524) was "half the size of the memory?".  

We don't have it configured in our cloud this way. @gregolsky told me we have 1 - 2 GB of swap. So especially on stronger machines we'll be getting that alert.

CC @grisha-kotler

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -707,6 +707,9 @@ public void Initialize()
             {
                 NotificationCenter.Add(alertRaised);
             }
+            
+            if(PlatformDetails.RunningOnDocker == false)
```
arekpalinski: minor: space after `if`

arekpalinski: Shouldn't we check this on Linux?

arekpalinski (changes requested):

src/Raven.Server/Config/Categories/MemoryConfiguration.cs
haludi: In windows, I was a recommendation for 1.5X of the physical memory
And in Linux I found more detailed recommendation:
Physical | Swap 
  <2        | X2
 2-8        | X1
 8-64      | X1/2

So as constant ratio default the most reasonable value to me was X1 
But actually we can use a dynamic ratio

src/Raven.Server/ServerWide/ServerStore.cs
haludi: I am not sure I follow your question.
The check here is because first I check it on docker on Linux and I got the host value of Swap so it is not relevant unless we will implement the `MemoryInformation.GetMemoryInfo` differently .

ayende: Did you docker instance had limited memory?
What about:
https://docs.microsoft.com/en-us/dotnet/api/system.gcmemoryinfo.totalavailablememorybytes?view=netcore-3.1

haludi: The physical memory is correct but the swap is incorrect (as I saw )

src/Raven.Server/Config/Categories/MemoryConfiguration.cs
arekpalinski: Note that this is about minimal swap size, not the recommended size. Unless we want to change that alert to recommend the swap size instead of alerting on the minimal.

Since we're setting the default value dynamically it doesn't have to be constant. You can calculate it dynamically based on the platform and available physical memory size.

src/Raven.Server/ServerWide/ServerStore.cs
arekpalinski: @haludi I got the feeling that this is relevant only for Linux. But it's not. It verifies the page file size on windows as well - is that correct?

src/Raven.Server/Config/Categories/PerformanceHintsConfiguration.cs
```diff
@@ -2,11 +2,16 @@
...
 using Sparrow;
+using Sparrow.LowMemory;
 
 namespace Raven.Server.Config.Categories
 {
     public class PerformanceHintsConfiguration : ConfigurationCategory
     {
+        public PerformanceHintsConfiguration()
+        {
+            MinSwapSize = MemoryInformation.TotalPhysicalMemory;
```
arekpalinski: What about my comment regarding this - https://github.com/ravendb/ravendb/pull/11093#discussion_r485551261?

arekpalinski:
Compilation check has failed. Can you check that?

```
Server/NotificationCenter/NotificationCenterTests.cs(559,58): error CS1061: 'MemoryConfiguration' does not contain a definition for 'MinSwapSize' and no accessible extension method 'MinSwapSize' accepting a first argument of type 'MemoryConfiguration' could be found (are you missing a using directive or an assembly reference?) [/mnt/c/jenkins/workspace/PR_Compile_v4.2/s/test/SlowTests/SlowTests.csproj]
```

src/Raven.Server/ServerWide/ServerStore.cs
arekpalinski: Did you see my question?

ppekrol:
Status?

src/Raven.Server/ServerWide/ServerStore.cs
haludi: > @haludi I got the feeling that this is relevant only for Linux. But it's not. It verifies the page file size on windows as well - is that correct?

My check? yes.
Do you think it should be applied only on Linux?

arekpalinski: If your check verifies the size of a page file on windows then it's okay. Although then the hint message shouldn't mention about "swap" but "the page file"

haludi:
> Status?

I spoked with @grisha-kotler and the decision is 
in Linux notify due to the configuration
and in Windows, while the PageFile can be flexible to just check if there is a FilePage (by CommitedAvailabe - TotalPhysicalMemory > 0) and notify if not.

src/Raven.Server/Config/Categories/PerformanceHintsConfiguration.cs
```diff
@@ -33,5 +46,11 @@ public class PerformanceHintsConfiguration : ConfigurationCategory
         [DefaultValue(30)]
         [TimeUnit(TimeUnit.Seconds)]
         public TimeSetting TooLongRequestThreshold { get; set; }
+        
+        [Description("The minimum swap size. If the swap size is lower a notification will arise")]
```
ppekrol: Write that this only applied to Linux

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -725,6 +727,35 @@ public void Initialize()
             InitializationCompleted.Set();
         }
 
+        private void CheckSwapOrPageFileAndRaiseNotification()
+        {
+            if (PlatformDetails.RunningOnLinux)
+            {
+                if (PlatformDetails.RunningOnDocker)
```
ppekrol, line 734: On Windows-docker we are good?

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -725,6 +727,35 @@ public void Initialize()
...
+            if (PlatformDetails.RunningOnLinux)
+            {
+                if (PlatformDetails.RunningOnDocker) 
+                    return;
+
+                var swapSize = MemoryInformation.GetMemoryInfo().TotalSwapSize;
+                if (swapSize < Configuration.PerformanceHints.MinSwapSize)
+                    NotificationCenter.Add(AlertRaised.Create(null,
+                        "Low swap size",
+                        $"The current swap size is {swapSize} and it is lower then the threshold defined {Configuration.PerformanceHints.MinSwapSize}",
```
ppekrol:
```
The current swap size is '{swapSize}' and it is lower than the threshold defined '{Configuration.PerformanceHints.MinSwapSize}'.
```

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -725,6 +727,35 @@ public void Initialize()
...
+                return;
+            }
+
+            if(PlatformDetails.RunningOnPosix == false)
+            {
+                var memoryInfo = MemoryInformation.GetMemoryInfo();
+                if(memoryInfo.TotalCommittableMemory - memoryInfo.TotalPhysicalMemory <= Sparrow.Size.Zero)
+                    NotificationCenter.Add(AlertRaised.Create(null,
+                        "No PageFile available",
+                        "Your system has no PageFile. There should be a PageFile for RavenDb to work properly",
```
ppekrol:
```
Your system has no PageFile. It is recommended to have a PageFile in order for Server to work properly.
```

src/Raven.Server/ServerWide/ServerStore.cs
haludi: In Windows, we only check if there is a PageFile without looking at the size but I will run it on Windows Docker to make sure it works properly

ppekrol: Does it work?

src/Raven.Server/Config/Categories/PerformanceHintsConfiguration.cs
ppekrol: Ping

src/Raven.Server/Config/Categories/PerformanceHintsConfiguration.cs
```diff
@@ -33,5 +46,11 @@ public class PerformanceHintsConfiguration : ConfigurationCategory
         [DefaultValue(30)]
         [TimeUnit(TimeUnit.Seconds)]
         public TimeSetting TooLongRequestThreshold { get; set; }
+        
+        [Description("The minimum swap size (for linux only). If the swap size is lower a notification will arise")]
```
ppekrol: linux -> Linux

src/Raven.Server/Config/Categories/PerformanceHintsConfiguration.cs
```diff
@@ -2,11 +2,24 @@
...
 namespace Raven.Server.Config.Categories
 {
     public class PerformanceHintsConfiguration : ConfigurationCategory
     {
+        public PerformanceHintsConfiguration()
+        {
+            if (PlatformDetails.RunningOnLinux)
+            {
+                var threshold = new Size(8, SizeUnit.Gigabytes);
+                MinSwapSize = MemoryInformation.TotalPhysicalMemory < threshold
```
ppekrol, line 17: Are we sure that we want to raise an alert when on 128GB machine swap is lower than 64GB?

haludi: No, if the half of the physical memory is greater than 8GB so we set it to 8GB

ppekrol: You are correct 🥇

PR #11095, opened 2020-09-09
RavenDB-15366 More details on failing tests

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15366

I couldn't reproduce
But it seems that the client thinks there where modifications to the document so I added information to facilitate further investigations


test/FastTests/Client/FirstClassPatch.cs
```diff
@@ -712,6 +712,29 @@ public async Task CanAddToArrayAsync()
                     Assert.Equal(loaded.Numbers[7], 202);
                 }
             }
+
+            async Task SaveChangesWithTryCatch(IAsyncDocumentSession session, User loaded)
```
arekpalinski: I know this is needed only for investigation but can we write it as helper method so we'll avoid repeating it multiple times. Once we figure out what is the problem it will be easier remove just a single method

arekpalinski:
BTW `commit/whitespace` check has failed because of tabs detection in `RavenDB-11089.cs` file. See: https://jenkins.hibernatingrhinos.com/job/PR_Conventions/10057/console

Issue #11108, opened 2020-09-14
[3.5] Could not get token for bulk insert on BulkInsert even with anonymous access


danielmarbach:
One of our clients is getting

```
Exception:
Could not get token for bulk insert
StackTrace:
   at Raven.Client.Document.RemoteBulkInsertOperation.<StartBulkInsertAsync>d__23.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Raven.Client.Document.RemoteBulkInsertOperation.<DisposeAsync>d__45.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Raven.Client.Document.ChunkedRemoteBulkInsertOperation.<DisposeAsync>d__19.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at ServiceControl.Audit.Auditing.AuditPersister.<Persist>d__2.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at ServiceControl.Audit.Auditing.AuditIngestionComponent.<<-ctor>b__8_0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NServiceBus.Transport.RabbitMQ.MessagePump.<Process>d__29.MoveNext()
Source:
Raven.Client.Lightweight
TargetSite:
Void MoveNext()
A task was canceled.
Data :System.Collections.DictionaryEntry
StackTrace:
[...]

PR #11110, opened 2020-09-14
Construct TCS with TaskCreationOptions.RunContinuationsAsynchronously in the PushContent

danielmarbach:
Potentially related to #11108 

I have not been able to investigate this in very thorough details but to me it seems due to the sync over async and blocking collection bridging not marking the TCS with `TaskCreationOptions.RunContinuationsAsynchronously` seems to be dangerous and potentially can cause deadlock scenarios under load

The whole code path is very similar to https://devblogs.microsoft.com/premier-developer/the-danger-of-taskcompletionsourcet-class/

I'll leave it up to the experts to decide whether this change makes sense or not

### DIFF
--- a/Raven.Client.Lightweight/Connection/Implementation/HttpJsonRequest.cs
+++ b/Raven.Client.Lightweight/Connection/Implementation/HttpJsonRequest.cs
@@ -906,7 +906,7 @@ namespace Raven.Client.Connection.Implementation
         private class PushContent : HttpContent
         {
             private readonly Action<Stream, TaskCompletionSource<object>> action;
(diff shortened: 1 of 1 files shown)


arekpalinski (approved):

ayende (approved):

grisha-kotler (approved):

arekpalinski:
@danielmarbach Are you running on .net core or full .net framework? I had to make the following PR https://github.com/ravendb/ravendb/pull/11120 to fix the build since our client project for full .net is targeting .net 4.5

danielmarbach:
Running on .NET 4.6 and higher. That's why I assumed the enum is available

PR #11114, opened 2020-09-15
Bulk insert session async all the way where possible

danielmarbach:
It seems to me the bulk insertion due to the sync over async problem is suffering quite a bit of starvation issues under load. This is a quick attempt to make things better by providing async write overloads that allows proper async flow in case of

```
            // if we haven't flushed the previous one yet, we will force
            // a disposal of both the previous one and the one before, to avoid
```

technically this is a breaking change but I assumed the low level interfaces can change given that they are not intended for public consumption.

### DIFF
--- a/Raven.Client.Lightweight/Document/RemoteBulkInsertOperation.cs
+++ b/Raven.Client.Lightweight/Document/RemoteBulkInsertOperation.cs
@@ -40,6 +40,8 @@ namespace Raven.Client.Document
 
         void Write(string id, RavenJObject metadata, RavenJObject data, int? dataSize = null);
 
+        Task WriteAsync(string id, RavenJObject metadata, RavenJObject data, int? dataSize = null);
+
         Task WaitForLastTaskToFinish();
 
         Task<int> DisposeAsync();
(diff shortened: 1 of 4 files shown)


danielmarbach:
I also found this gem here

```
            if (previousTask != null && waitedForPreviousTask == false)
            {
                Total += previousTask.Result;
                waitedForPreviousTask = true;
            }
```

which means `FlushBatch` is synchronously blocking on `DisposeAsync` of previous operations in certain scenarios. Is the assumption of this code that because we are always on a background thread this is fine?

Raven.Client.Lightweight/Document/RemoteBulkInsertOperation.cs
```diff
@@ -335,6 +337,54 @@ public virtual void Write(string id, RavenJObject metadata, RavenJObject data, i
             throw new TimeoutException("Could not flush in the specified timeout, server probably not responding or responding too slowly.\r\nAre you writing very big documents?");
         }
 
+        public virtual async Task WriteAsync(string id, RavenJObject metadata, RavenJObject data, int? dataSize = null)
```
danielmarbach, line 340: I wish I had value task

Raven.Tests.Web/Controllers/Commands/AsyncCommandsController.cs
```diff
@@ -125,7 +125,7 @@ public Task<HttpResponseMessage> GetBulkInsertOperation()
         {
             using (var operation = DocumentStore.AsyncDatabaseCommands.GetBulkInsertOperation(new BulkInsertOptions(), DocumentStore.Changes()))
             {
-                operation.Write(Guid.NewGuid().ToString(), new RavenJObject(), new RavenJObject());
+                operation.WriteAsync(Guid.NewGuid().ToString(), new RavenJObject(), new RavenJObject());
```
arekpalinski: Either `await` or user `Write()`. You still have it.

arekpalinski (commented):
You've added a new method to `ILowLevelBulkInsertOperation` interface so it's fine. As you pointed it's low level API anyway.

Raven.Tests.Web/Controllers/Commands/AsyncCommandsController.cs
```diff
@@ -125,7 +125,7 @@ public Task<HttpResponseMessage> GetBulkInsertOperation()
         {
             using (var operation = DocumentStore.AsyncDatabaseCommands.GetBulkInsertOperation(new BulkInsertOptions(), DocumentStore.Changes()))
             {
-                operation.Write(Guid.NewGuid().ToString(), new RavenJObject(), new RavenJObject());
```
grisha-kotler, line 128: You are disposing of the operation before the write is done

Raven.Tests.Web/Controllers/Commands/AsyncCommandsController.cs
danielmarbach: Ah sorry. Oversight. Will fix

danielmarbach: Fixed

PR #11123, opened 2020-09-17
RavenDB-15621 - Reduce the response size of the /admin/debug/memory/stats endpoint

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-15621


src/Raven.Server/Documents/Handlers/Debugging/MemoryDebugHandler.cs
```diff
@@ -190,14 +190,16 @@ public Task MemorySmaps()
         [RavenAction("/admin/debug/memory/stats", "GET", AuthorizationStatus.Operator, IsDebugInformationEndpoint = true)]
```
ppekrol, line 190: This is a debug information endpoint, it means that we will not include threads ans mappings for debug package. Is taht what we want? cc @karmeli87

grisha-kotler: I changed it to be `true` by default.

PR #11133, opened 2020-09-19
RavenDB-15074 Allow to pass a flag that will avoid the feedback notic…

ml054:
…e in the Studio


ayende (commented):
Need to also update `EmbeddedServer.OpenStudioInBrowser()`

ml054:
updated EmbeddedServer as well - thanks

PR #11145, opened 2020-09-20
RavenDB-15632 Optimizing memory utilization for tens of thousands on Voron instances

ayende:


src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -323,13 +294,9 @@ public virtual void Dispose()
...
 
-                foreach (var coreContext in _perCoreContexts)
+                foreach (var context in _perCoreCache.EnumerateAndClear())
                 {
-                    for (int i = 0; i < coreContext.Length; i++)
-                    {
-                        coreContext[i]?.Dispose();
-                        coreContext[i] = null;
-                    }
+                    context.Dispose();
```
karmeli87, line 299: I know it was in the original code like this, but I wonder if we need to be on the safe side and do:
`if (context.InUse.Raise())`

karmeli87: Not sure if it possible, but for instance we use the context pool in the request executor, what if we dispose it when some requests still pending?

ayende: I'm actually doing that here on _purpose_. The idea is that if we are calling `Dispose()`, the _caller_ is responsible for making sure that no one is using these. So we don't want to do a check, we want it to fail.

karmeli87: But this will fail with AVE, right? maybe add a debug message to the destructor?

ayende: If you have concurrent usage, yes, it would. And that is what I expect it to do.
This is a use after free scenario, which is expected to fail _hard_.
It didn't so far, mind, so we are good.

grisha-kotler:
LGTM.
It would be nice if we could store this data as unmanaged memory.

ayende:
What data are you talking about?

grisha-kotler:
IO metrics and indexing metrics

ayende:
I'm not sure how helpful that would have been. The issue was that we have a LOT of them. That cause memory utilization.
The fact that they are managed meant that we had an easy time tracking them, which is good.

grisha-kotler:
The metrics are useful. Enabling them will require a database restart.

ayende:
The issue is the total memory cost, not where this is stored. This isn't the case with the GC causing problem.
And this is expected to only be turned off in strange conditions.

PR #11149, opened 2020-09-20
RavenDB-15589 Implement aggressive caching on lazy queries

ayende:


src/Raven.Client/Documents/Commands/MultiGet/MultiGetCommand.cs
```diff
@@ -13,21 +13,55 @@ namespace Raven.Client.Documents.Commands.MultiGet
...
             _baseUrl = $"{node.Url}/databases/{node.Database}";
+            url = $"{_baseUrl}/multi_get";
+
+            var aggressiveCacheOptions = _requestExecutor.AggressiveCaching.Value;
+            if (aggressiveCacheOptions?.Mode == AggressiveCacheMode.TrackChanges)
+            {
+                Result = new List<GetResponse>();
+                foreach (var command in _commands)
+                {
+                    if(command.SkipAggressiveCache) break;
```
ppekrol: Should be `continue` ?

src/Raven.Client/Documents/Commands/MultiGet/GetRequest.cs
```diff
@@ -22,6 +22,8 @@ public class GetRequest
         /// </summary>
         public string Query { get; set; }
         
+        public bool SkipAggressiveCache { get; set; }
```
ppekrol: It is called 'CanCacheAggressively' in Commands. Let's stick to that.

src/Raven.Client/Documents/Commands/MultiGet/MultiGetCommand.cs
```diff
@@ -13,21 +13,55 @@ namespace Raven.Client.Documents.Commands.MultiGet
...
+                    var cacheKey = GetCacheKey(command, out string _);
+                    using (var cachedItem = _cache.Get(ctx, cacheKey, out _, out var cached))
+                    {
+                        if (cached == null || 
+                            cachedItem.Age > aggressiveCacheOptions.Duration ||
+                            cachedItem.MightHaveBeenModified)
+                            break;
+
```
ppekrol, line 49: Double-check. Do we need to apply logic here from 'RequestExecutor.TryGetFromCache' ?

src/Raven.Client/Documents/Commands/MultiGet/GetRequest.cs
ayende: I wanted to have this set to false by default, so we don't have to touch it in all call sites.

src/Raven.Client/Documents/Commands/MultiGet/MultiGetCommand.cs
ayende: No, the first one that isn't find for us, we'll break and the `Result`'s count will be different than `_commands` so we'll go to the server for all items.

src/Raven.Client/Documents/Commands/MultiGet/MultiGetCommand.cs
ayende: That is where I took the logic from, yes.

src/Raven.Client/Documents/Commands/MultiGet/GetRequest.cs
ppekrol: Set it to `true` in ctor like we have it for RavenCommand?

src/Raven.Client/Documents/Commands/MultiGet/MultiGetCommand.cs
ppekrol: OK

src/Raven.Client/Documents/Commands/MultiGet/MultiGetCommand.cs
ppekrol: There was additional logic for delete, but it is not possible to have that case here, wanted to double-check.

ppekrol (approved):

PR #11150, opened 2020-09-21
RavenDB-15640 Allow to disable remote api calls for secured deployments

ayende:


src/Raven.Server/Config/Categories/LicenseConfiguration.cs
```diff
@@ -46,5 +46,16 @@ public class LicenseConfiguration : ConfigurationCategory
         [DefaultValue(false)]
         [ConfigurationEntry("License.SkipLeasingErrorsLogging", ConfigurationEntryScope.ServerWideOnly)]
         public bool SkipLeasingErrorsLogging { get; set; }
+        
+        
+        [Description("EXPERT ONLY. Disable automatic updating of the license")]
+        [DefaultValue(false)]
+        [ConfigurationEntry("License.DisableAutoLicenceUpdate", ConfigurationEntryScope.ServerWideOnly)]
```
ppekrol: There is a typo here, but I think it would be better to change this to `License.DisableAutoUpdate`. Thoughts?

src/Raven.Server/Config/Categories/LicenseConfiguration.cs
```diff
@@ -46,5 +46,16 @@ public class LicenseConfiguration : ConfigurationCategory
...
+        public bool DisableAutoLicenceUpdate { get; set; }
+        
+        [Description("EXPERT ONLY. Disable checking the support options for the license")]
+        [DefaultValue(false)]
+        [ConfigurationEntry("License.DisableLicenseSupportMode", ConfigurationEntryScope.ServerWideOnly)]
```
ppekrol: Can we change that to `License.DisableLicenseSupportCheck` ?

ppekrol:
Do we need this in 4.2?

src/Raven.Server/Config/Categories/LicenseConfiguration.cs
ayende: done

src/Raven.Server/Config/Categories/LicenseConfiguration.cs
ayende: done

ayende:
I don't think we need this in 4.2 at this point

ppekrol (approved):

PR #11151, opened 2020-09-21
RavenDB-15596 ETL fixes

karmeli87:
- process ETL items by the etag order
- Break large time series ETL into several batches
- Filter only after extraction, otherwise we might move the etag too far and skip items


src/Raven.Server/Documents/ETL/EtlProcess.cs
```diff
@@ -100,6 +99,12 @@ public static EtlProcessState GetProcessState(DocumentDatabase database, string
         }
     }
 
+    public interface IExtractEnumerator<out T> : IEnumerator<T> where T : ExtractedItem
```
arekpalinski: Let's extract it to separate file.

src/Raven.Server/Documents/ETL/EtlProcess.cs
```diff
@@ -286,34 +262,27 @@ private IEnumerable<TExtracted> Extract(DocumentsOperationContext context, long
             EtlStatsScope stats,
             DisposableScope scope)
         {
-            var lastProcessedDocEtag = stats.GetLastTransformedOrFilteredEtag(EtlItemType.Document);
-            var toEtag = lastProcessedDocEtag > 0 
-                ? lastProcessedDocEtag 
-                // If there was no document transformed in current batch
-                // it means we're done with documents, we can send all time series
-                : long.MaxValue;
```
arekpalinski, line 294: Just to make sure. We no longer need that because after your changes docs and timeseries are processed in etag order (not as it was before that we were processing docs and their time series were processed later on). Is that correct?

src/Raven.Server/Documents/ETL/Providers/Raven/Enumerators/AttachmentTombstonesToRavenEtlItems.cs
```diff
@@ -0,0 +1,62 @@
...
+            _context = context;
+            _tombstones = tombstones;
+            _collections = collections ?? throw new ArgumentNullException(nameof(collections));
+        }
+
+        public bool Filter()
+        {
+            var tombstone = _tombstones.Current;
+            if (tombstone.Type != Tombstone.TombstoneType.Attachment)
+                throw new InvalidOperationException();
```
arekpalinski: Please provide exception message. Also you might extract the exception to a dedicated method to get slightly better perf. `Filter` is expected to be called quite frequently.

src/Raven.Server/Documents/ETL/EtlProcess.cs
```diff
@@ -502,8 +455,7 @@ private bool CanContinueBatch(EtlStatsScope stats, TExtracted currentItem, int b
                 return false;
             }
 
-            if (currentItem.Type == EtlItemType.Document &&
```
arekpalinski, line 505: Why did you remove `currentItem.Type == EtlItemType.Document` check? We should check `Database.Configuration.Etl.MaxNumberOfExtractedDocuments` only for the docs

src/Raven.Server/Documents/ETL/Providers/Raven/Enumerators/AttachmentTombstonesToRavenEtlItems.cs
arekpalinski: And fix the missing exception message in other places as well.

src/Raven.Server/Documents/ETL/EtlProcess.cs
karmeli87: yes

src/Raven.Server/Documents/ETL/Providers/Raven/Enumerators/AttachmentTombstonesToRavenEtlItems.cs
karmeli87: Frequent calls was my concern as well, but it is not suppose to be super performant anyway, right?
`Filter` will be called as much as `MoveNext`.

src/Raven.Server/Documents/ETL/EtlProcess.cs
karmeli87: we do, one line below.

src/Raven.Server/Documents/ETL/Providers/Raven/Enumerators/AttachmentTombstonesToRavenEtlItems.cs
arekpalinski: Yeah. Just moving the `throw` clause to a separate method and calling it here might let it optimize stuff by inlining.

src/Raven.Server/Documents/ETL/EtlProcess.cs
arekpalinski: Okay. Makes sense, you don't have to check that `currentItem` is a document

ppekrol (approved):

src/Raven.Server/Documents/ETL/EtlProcess.cs
```diff
@@ -201,81 +194,61 @@ private IEnumerable<TExtracted> Extract(DocumentsOperationContext context, long
...
-            {
-                var attachmentTombstones = Database.DocumentsStorage
-                    .GetTombstonesFrom(context, AttachmentsStorage.AttachmentsTombstones, fromEtag, 0, long.MaxValue).GetEnumerator();
-                scope.EnsureDispose(attachmentTombstones);
-
-                attachmentTombstones = new FilterTombstonesEnumerator(attachmentTombstones, stats, Tombstone.TombstoneType.Attachment, context,
-                    fromCollections: Transformation.ApplyToAllDocuments ? null : Transformation.Collections);
 
-                merged.AddEnumerator(ConvertTombstonesEnumerator(attachmentTombstones, null, EtlItemType.Document));
+                if (ShouldTrackAttachmentTombstones())
```
haludi, line 223: You should do check and track attachments in `Transformation.ApplyToAllDocuments` as well, no?

src/Raven.Server/Documents/ETL/EtlProcess.cs
```diff
@@ -706,29 +654,26 @@ public void Run()
...
-                                    List<TTransformed> ExtractTransform(EtlItemType type)
-                                    {
-                                        var extracted = Extract(context, loadLastProcessedEtag + 1, type, stats);
-                                        return Transform(extracted, context, stats, state);
-                                    }
+                                    var transformations = Transform(merged, context, stats, state);
 
-                                    var successfulLoad = false;
-
-                                    if (transformations.Count > 0)
```
haludi, line 726: In a case, there is no transformed item you still act as "didWork"

karmeli87: yes, because there might be stuff that we skipped, but we need to update the etag.

src/Raven.Server/Documents/ETL/EtlProcess.cs
```diff
@@ -431,7 +403,7 @@ public bool Load(IEnumerable<TTransformed> items, DocumentsOperationContext cont
 
                     stats.RecordLastLoadedEtag(stats.LastTransformedEtags.Values.Max());
 
-                    Statistics.LoadSuccess(stats.NumberOfTransformedItems.Sum(x => x.Value));
+                    Statistics.LoadSuccess(count);
```
arekpalinski, line 406: It's for stats only but I don't think that the usage of `count` is correct. `LoadInternal` returns the number of commands (for example we might have DELETE and PUT for a single transformed item) while we should record `transformed - filtered` items here

arekpalinski: and we should use the same value for below `stats.RecordLoadSuccess()`

karmeli87: To me it is not well defined.

does sending a tombstone counts as `Load` ? (it doesn't count as `Transformed`, it counts as `NumberOfTransformedTombstones`)
Load of a time-series is segment based, entries based or is it count as 1 regardless to the number of segments (`TransformedItems` is segment based btw.)
Same for counters.

Not sure that `count` is the best choice though..

arekpalinski: As we discussed there are other problems with using the transformed number of items - for example we might transform time series segment but filter out all it's entries under the covers.

So the usage of `count` had some drawbacks but appears to be the best number for stats purpose.

arekpalinski (approved):

src/Raven.Server/Documents/ETL/Providers/Raven/Enumerators/CountersToRavenEtlItems.cs
```diff
@@ -1,26 +1,39 @@
...
             _counters = counters;
             _collection = collection;
         }
 
+        private bool Filter(RavenEtlItem item)
+        {
+            var doc = _context.DocumentDatabase.DocumentsStorage.Get(_context, item.DocumentId, DocumentFields.Default);
+
+            // counter has lower etag than its document - we skip it to avoid
+            // sending a counter of document that can not exist on the destination side
```
ayende, line 26: what will send the counter to the other side afterward? won't this cause us to miss a write?

ayende (commented):
LGTM

ayende (approved):

src/Raven.Server/Documents/ETL/Providers/Raven/Enumerators/CountersToRavenEtlItems.cs
karmeli87: we always load the counters together with the document itself.
https://github.com/ravendb/ravendb/blob/4569e8a7bf34cc94c722b85bd16e6f342d3df663/src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs#L217

PR #11158, opened 2020-09-22
RavenDB-15642 updated to .NET 5.0 RC1

ppekrol:


build.ps1
```diff
@@ -212,16 +212,16 @@ Foreach ($target in $targets) {
...
 
     if ($target.Name -eq "windows-x64") {
-        Validate-AssemblyVersion $(Join-Path -Path $specOutDirs.Server -ChildPath "Raven.Server.dll" ) $versionObj
-        Validate-AssemblyVersion $(Join-Path -Path $specOutDirs.Server -ChildPath "Sparrow.dll" ) $versionObj
-        Validate-AssemblyVersion $(Join-Path -Path $specOutDirs.Server -ChildPath "Voron.dll" ) $versionObj
-        Validate-AssemblyVersion $(Join-Path -Path $specOutDirs.Rvn -ChildPath "rvn.dll" ) $versionObj
-        Validate-AssemblyVersion $(Join-Path -Path $specOutDirs.Migrator -ChildPath "Raven.Migrator.dll" ) $versionObj
-        Validate-AssemblyVersion $(Join-Path -Path $specOutDirs.Drtools -ChildPath "Voron.Recovery.dll" ) $versionObj
-        Validate-AssemblyVersion $(Join-Path -Path $specOutDirs.Debug -ChildPath "Raven.Debug.dll" ) $versionObj
+        #Validate-AssemblyVersion $(Join-Path -Path $specOutDirs.Server -ChildPath "Raven.Server.exe" ) $versionObj
```
ayende: This is still required, no?

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -177,6 +177,7 @@ public class SecurityConfiguration : ConfigurationCategory
 
         public bool AuthenticationEnabled => IsCertificateConfigured;
 
+#if !RVN
```
ayende, line 180: why do we need this?

src/Raven.Server/Raven.Server.csproj
```diff
@@ -35,46 +35,47 @@
...
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
-    <None Include="..\..\libs\librvnpal\librvnpal.arm.64.so" Link="librvnpal.arm.64.so">
+  </ItemGroup>
+  <ItemGroup Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Linux)))'">
+    <None Include="..\..\libs\libsodium\libsodium.linux.x64.so" Link="libsodium.linux.x64.so">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
-    <None Include="..\..\libs\librvnpal\librvnpal.win.x64.dll" Link="librvnpal.win.x64.dll">
+    <None Include="..\..\libs\libsodium\libsodium.arm.32.so" Link="libsodium.arm.32.so">
```
ayende, line 66: ??

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
ppekrol: I removed project reference in RVN to Raven.Server and using file linking there. By doing this the Single File App for RVN is 30mb instead of 130mb

build.ps1
ppekrol: I will have to check how to do it, but this is now throwing exception.

src/Raven.Server/Raven.Server.csproj
ppekrol: I'm attaching all .so for linux, all dlls for windows and all dylib for osx. Previosuly all distributions had all the DLLs. Not sure how in MSBuild detect that I'm building for ARM (x32 and x64) to include even less libs for Linux. Note taht we are talking here about ~1mb only so I consider this an optimization (the ARM detection)

PR #11159, opened 2020-09-22
RavenDB-15020 : Additional time series functions

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-15020


src/Raven.Client/Documents/Queries/TimeSeries/TimeSeriesAggregation.cs
```diff
@@ -145,7 +145,7 @@ public class TimeSeriesAggregationResult<T> : TimeSeriesQueryStreamResultBase<Ti
     public class TimeSeriesRangeAggregation : IPostJsonDeserialization, ITimeSeriesQueryStreamEntry
     {
         public long[] Count;
-        public double[] Max, Min, Last, First, Average, Sum;
+        public double[] Max, Min, Last, First, Average, Sum, Percentile, Slope;
```
karmeli87: Wonder if we should call it `Derivative` (or maybe `Change`?) instead of `Slope`.

src/Raven.Server/Documents/Queries/Results/TimeSeries/TimeSeriesAggregation.cs
```diff
@@ -0,0 +1,414 @@
...
+                        break;
+                    case AggregationType.Last:
+                        _values[i] = val;
+                        break;
+                    case AggregationType.Count:
+                        break;
+                    case AggregationType.Percentile:
+                        Debug.Assert(_rankedValues != null);
+                        var currentRankedList = _rankedValues[i];
+                        var loc = currentRankedList.BinarySearch(val);
```
karmeli87: I would prefer to use `SortedDictionary` with a comparer to handle duplicates.

src/Raven.Client/Documents/Queries/TimeSeries/TimeSeriesAggregation.cs
ayende: Isn't slope the right math term?

src/Raven.Server/Documents/Queries/Results/TimeSeries/TimeSeriesAggregation.cs
ayende: Agreed, also, and `insert` here will cause us to move a lot of values, potentially.

src/Raven.Server/Documents/Queries/Results/TimeSeries/TimeSeriesAggregation.cs
```diff
@@ -0,0 +1,414 @@
...
+
+            if (scale.HasValue == false) 
+                return _finalValues = _values;
+
+            return _finalValues = _values.Select(x => x * scale.Value);
+        }
+
+        private IEnumerable<double> GetPercentile(double scale)
+        {
+            foreach (var list in _rankedValues)
```
ayende: This rounds down, right? Do we need some manner to adjust the value based on the distance?
So if we have [1,2,4] and we ask for 75% percentile, we'll get [2] here. But that actually represent the 50%.
Should we adjust that so it would be 3 as the result?

aviv86: No actually we round up : `index = (int)Math.Ceiling(_percentileFactor * list.Count)`
so in your example we'll get `4` as the result. (`2` is wrong by any definition)

There are different definitions for percentile and methods for calculating percentile.
By the definition that I chose to implement the result is always a part of the data set.
So if we want to adjust the result to be `3` we need to use a different method..

I wrote more about it in the issue page : https://issues.hibernatingrhinos.com/issue/RavenDB-15020#focus=Comments-67-51024.0-0

src/Raven.Client/Documents/Queries/TimeSeries/TimeSeriesAggregation.cs
aviv86: I think `Slope` is the right math term. 
`Change` is inaccurate - we are measuring **the rate** of change
and `Derivative` is the rate of change when the distance (delta X) tends towards 0, no?

src/Raven.Server/Documents/Queries/Results/TimeSeries/TimeSeriesAggregation.cs
ayende: Okay, that make sense.
If it isn't hard, let's go with extrapolation. Note that we need to document this.

src/Raven.Server/Documents/Queries/Results/TimeSeries/TimeSeriesAggregation.cs
aviv86: done

src/Raven.Server/Documents/Queries/Results/TimeSeries/TimeSeriesAggregation.cs
aviv86: done

aviv86:
- Added `StandardDeviation` support
- Made some refactoring in `TimeSeriesAggregation`  (introduced `ITimeSeriesAggregation`)
- Addressed PR comments

src/Raven.Server/Documents/Queries/Results/TimeSeries/AverageAggregation.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            for (int i = 0; i < values.Length; i++)
+            {
+                _values[i] += values[i];
+                _count[i]++;
+            }
+        }
+
+        IEnumerable<double> ITimeSeriesAggregation.GetFinalValues(DateTime? @from, DateTime? to, double? scale = null)
+        {
```
ppekrol, line 84: Can you check the warning?
```
src/Raven.Server/Documents/Queries/Results/TimeSeries/AverageAggregation.cs#L83
The default value specified for parameter 'scale' will have no effect because it applies to a member that is used in contexts that do not allow optional arguments
```

Probably just remove `= null` from `scale` here. It is an explicit interface impl.

src/Raven.Server/Documents/Queries/Results/TimeSeries/SlopeAggregation.cs
```diff
@@ -0,0 +1,144 @@
...
+                if (i < oldCount)
+                {
+                    _first[i] = val.First;
+                }
+
+                _values[i] = val.Last;
+            }
+        }
+
+        IEnumerable<double> ITimeSeriesAggregation.GetFinalValues(DateTime? @from, DateTime? to, double? scale = null)
```
ppekrol: Warning here. Explicit interface impl.

src/Raven.Server/Documents/Queries/Results/TimeSeries/TimeSeriesAggregation.cs
```diff
@@ -0,0 +1,274 @@
...
+
+        public string Name { get; }
+
+        public int NumberOfValues { get; }
+
+        public AggregationType Aggregation { get; }
+
+    }
+
+    public class TimeSeriesAggregationBase
```
ppekrol: abstract?

ppekrol (approved):

src/Raven.Server/Documents/Queries/Results/TimeSeries/AverageAggregation.cs
aviv86: done

src/Raven.Server/Documents/Queries/Results/TimeSeries/SlopeAggregation.cs
aviv86: done

src/Raven.Server/Documents/Queries/Results/TimeSeries/TimeSeriesAggregation.cs
aviv86: done

ayende (approved):

src/Raven.Client/Documents/Queries/TimeSeries/TimeSeriesAggregation.cs
```diff
@@ -145,7 +145,7 @@ public class TimeSeriesAggregationResult<T> : TimeSeriesQueryStreamResultBase<Ti
     public class TimeSeriesRangeAggregation : IPostJsonDeserialization, ITimeSeriesQueryStreamEntry
     {
         public long[] Count;
-        public double[] Max, Min, Last, First, Average, Sum;
+        public double[] Max, Min, Last, First, Average, Sum, Percentile, Slope, StandardDeviation;
```
karmeli87, line 148: Should we maybe shorten this to StdDev?

src/Raven.Server/Documents/Queries/Results/TimeSeries/SlopeAggregation.cs
```diff
@@ -0,0 +1,144 @@
...
+            }
+
+            var oldCount = _values.Count;
+            InitValuesIfNeeded(values.Length);
+
+            for (int i = 0; i < values.Length; i++)
+            {
+                var val = values[i];
+
+                if (i < oldCount)
```
karmeli87: for consistency with above `oldCount <= i`

karmeli87 (approved):
Very Nice! I like the `TimeSeriesAggregation` abstraction :-)

src/Raven.Client/Documents/Queries/TimeSeries/TimeSeriesAggregation.cs
ayende: yes

ppekrol: No, then Average should be Avg, etc. Either use full names of abbreviations. Not both.

ppekrol: Actually it is already mixed, nvm.

ml054: May I ask what's definition used for slope?

aviv86: The rate of change within a range (only works with group by)
i.e. `slope = dY/dX`, dY = `range.LastVal - range.FirstVal`, dX = `(range.End - range.Start).TotalMilliseconds`

ml054: ok I thought  it uses linear regression

PR #11162, opened 2020-09-22
RavenDB-15643 - Configuration value isn't visible when using an operator certificate

grisha-kotler:


ppekrol:
That is by design. You cannot view Server-wide value if you are not Cluster Admin, but you are aware that it is set.

ayende:
So the question is now, why is this is a secured value?

ppekrol:
It isn't. It is a Server-wide value and those are not visible to Operator by design.

Secured values are not visible to anyone at all times.

grisha-kotler:
Why a user cannot view the configuration value? (not edit)

PR #11166, opened 2020-09-23
RavenDB-15615 Track not found documents on `Load`

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15615


ayende (approved):
Need to also add it to conditional load, no?

ppekrol:
> Need to also add it to conditional load, no?

Will be handled here https://issues.hibernatingrhinos.com/issue/RavenDB-15652

PR #11171, opened 2020-09-24
RavenDB-15538: Missing attachment break replication #2

garayx:


src/Raven.Server/Documents/Replication/ReplicationItems/AttachmentReplicationItem.cs
```diff
@@ -109,18 +109,28 @@ public override unsafe void Read(DocumentsOperationContext context, IncomingRepl
 
         protected override ReplicationBatchItem CloneInternal(JsonOperationContext context)
         {
+            MemoryStream stream;
```
grisha-kotler:
```suggestion
            MemoryStream stream = null;
```

src/Raven.Server/Documents/Replication/ReplicationItems/AttachmentReplicationItem.cs
```diff
@@ -109,18 +109,28 @@ public override unsafe void Read(DocumentsOperationContext context, IncomingRepl
...
+            if (Stream == null)
+                stream = null;
+            else
```
grisha-kotler:
```suggestion
            if (Stream != null)
```

ppekrol (approved):

src/Raven.Server/Documents/AttachmentsStorage.cs
```diff
@@ -789,6 +797,32 @@ public ByteStringContext.InternalScope GetAttachmentPrefix(DocumentsOperationCon
             return scope;
         }
 
+        public static AttachmentType GetAttachmentTypeByKey(Slice keySlice)
+        {
+            var index = 0;
+            var found = false;
+            for (int i = 0; i < keySlice.Size; i++)
+            {
+                if (Convert.ToChar(keySlice[i]) == SpecialChars.RecordSeparator)
```
karmeli87, line 806: Hmm.. we have places that we do
 `var idEnd = key.Content.IndexOf(SpecialChars.RecordSeparator);` 

or 
```
var val = bytes[i];
if (val != SpecialChars.RecordSeparator)
```
and I wonder if it is correct.. in case we have some unicode char might we identify it wrong?

karmeli87 (approved):

PR #11182, opened 2020-09-29
RavenDB-15538: Missing attachment break replication 4.2

garayx:


garayx:
5.0: https://github.com/ravendb/ravendb/pull/11171

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -1550,6 +1551,66 @@ public int GetNextReplicationStatsId()
...
+        internal TestingStuff _forTestingPurposes;
+        internal TestingStuff ForTestingPurposesOnly()
+        {
+            if (_forTestingPurposes != null)
+                return _forTestingPurposes;
+
+            return _forTestingPurposes = new TestingStuff();
+        }
+
+        internal class TestingStuff
```
karmeli87: Why do we need all this? can't we use `GetReplicationItems` in both dbs?

garayx: done

test/SlowTests/Issues/RavenDB_15538.cs
```diff
@@ -0,0 +1,302 @@
...
+        {
+            if (item.Type == ReplicationBatchItem.ReplicationItemType.Attachment)
+            {
+                item.Stream.Position = 0;
+                var stream = new MemoryStream();
+                item.Stream.CopyTo(stream);
+                stream.Position = 0;
+                item.Stream.Position = 0;
+
+                return new ReplicationBatchItem
```
karmeli87, line 268: For the 5.x please use the `Clone` method

karmeli87 (approved):

PR #11187, opened 2020-09-30
RavenDB-15617 prevent putting same timestamp twice

karmeli87:


src/Raven.Server/Documents/TimeSeries/TimeSeriesRollups.cs
```diff
@@ -529,52 +485,130 @@ protected override long ExecuteCmd(DocumentsOperationContext context)
                             AlertType.RollupExceedNumberOfValues, NotificationSeverity.Warning, $"{item.DocId}/{item.Name}", new ExceptionDetails(e));
 
                         context.DocumentDatabase.NotificationCenter.Add(alert);
```
ayende, line 489: Can this cause us to spam the user with alerts?

karmeli87: it will generate a message per (problematic) ts

ayende: That can be a lot of them, no?

karmeli87: So how to notify to the user about the issue? 
Also I think there is an open issue in the studio to combine alerts by the type.

ayende: We can use the key on the "Collection/TimeSeries", no? That would reduce the number of unique matches.

karmeli87: Without the docId, since we assume that all ts in the same collection will have the same structure? 
I'm fine with that :)

ayende: yes

karmeli87: done

PR #11188, opened 2020-09-30
RavenDB-12471 More exception to return HttpStatusCode.ServiceUnavailable

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-12471

https://github.com/ravendb/ravendb/pull/8198#pullrequestreview-183179828

### DIFF
--- a/src/Raven.Server/RavenServerStartup.cs
+++ b/src/Raven.Server/RavenServerStartup.cs
@@ -336,10 +336,12 @@ namespace Raven.Server
             }
 
             if (exception is LowMemoryException ||
+                exception is HighDirtyMemoryException || 
                 exception is OutOfMemoryException ||
                 exception is VoronUnrecoverableErrorException ||
-                exception is DiskFullException ||
-                exception is HighDirtyMemoryException)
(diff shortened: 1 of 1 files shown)


src/Raven.Server/RavenServerStartup.cs
```diff
@@ -336,18 +336,19 @@ private static void MaybeSetExceptionStatusCode(HttpContext httpContext, ServerS
...
+                exception is QuotaException ||
+                exception is DiskFullException)
             {
                 response.StatusCode = (int)HttpStatusCode.ServiceUnavailable;
                 return;
             }
 
             if (exception is DocumentConflictException ||
                 exception is ConflictException ||
-                exception is ConcurrencyException ||
```
grisha-kotler, line 350: why was it removed?

haludi: Actually, it was a ReSharper warning.  `ConcurrencyException` is `ConflictException` so it always returns `false`

grisha-kotler (approved):

PR #11190, opened 2020-10-04
RavenDB-14787 Check for null argument in client commands

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-14787


ml054:
do we want to update other commands with new nice utitlity method? To be consistent?

ppekrol:
I'm against doing this. Why?
- 90% of the people will do not remember that it exist
- Dev cannot use ReSharper, VS IntelliSense, Roslynator automatic fixes now

Please use those automatic fixes instead, so the end result would be for example:

```
 _cache = cache ?? throw new ArgumentNullException(nameof(cache));
```

ml054:
ok I agree, so we should remove this utility method.

PR #11192, opened 2020-10-04
RavenDB-15667 Hub Replication Access Info is Not exported/imported wh…

efratshenhar:
…en exporting/importing a database


ml054:
do we want to have toggle for that on UI? Right now we export / import that everytime.

karmeli87:
This was only a studio issue? do we a have server-side test for this?

efratshenhar:
yes

ppekrol:
> do we want to have toggle for that on UI? Right now we export / import that everytime.

We should expose this in the UI

Danielle9897:
Opened https://issues.hibernatingrhinos.com/issue/RavenDB-15680

PR #11197, opened 2020-10-05
RavenDB-15507: Support for multiple collections in Subscriptions

garayx:


src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
```diff
@@ -70,7 +71,7 @@ public async Task Try()
                         case Constants.Documents.SubscriptionChangeVectorSpecialStates.LastDocument:
                             using (context.OpenReadTransaction())
                             {
-                                state.ChangeVectorForNextBatchStartingPoint = Database.DocumentsStorage.GetLastDocumentChangeVector(context.Transaction.InnerTransaction, context, sub.Collection);
+                                state.ChangeVectorForNextBatchStartingPoint = sub.Collections.Max(x => Database.DocumentsStorage.GetLastDocumentChangeVector(context.Transaction.InnerTransaction, context, x));
```
ayende, line 74: What is the meaning of Max on change vectors?

src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
```diff
@@ -491,7 +492,7 @@ private async Task CreateInternal(BlittableJsonReaderObject bjro, SubscriptionCr
                         options.ChangeVector = null;
                         break;
                     case Constants.Documents.SubscriptionChangeVectorSpecialStates.LastDocument:
-                        options.ChangeVector = Database.DocumentsStorage.GetLastDocumentChangeVector(context.Transaction.InnerTransaction, context, sub.Collection);
+                        options.ChangeVector = sub.Collections.Max(x => Database.DocumentsStorage.GetLastDocumentChangeVector(context.Transaction.InnerTransaction, context, x));
```
ayende, line 495: Same, what is the meaning here?

test/SlowTests/Client/Subscriptions/SubscriptionsBasic.cs
```diff
@@ -770,6 +770,295 @@ public async Task ShouldThrow()
...
+        }
+
+        [Fact]
+        public async Task SubscriptionOnMultipleCollectionsWillReturnResultsWhenAnyOfTheCollectionsChange()
+        {
+            using (var store = GetDocumentStore())
+            {
+                var allId = await store.Subscriptions.CreateAsync(new SubscriptionCreationOptions
+                {
+                    Query = "from (Users, People, PersonWithAddresses, GeekPeople) as doc"
```
ayende, line 930: I don't like this approach. There is no good way for us to do more interesting stuff here.
Better to do something like multi map with providing multiple queries.
There is also the issue of how to handle this on the client side? I don't like the kind we have right now

ayende (changes requested):

ayende:
This need a proper design to consider the proper UX for this feature.

PR #11200, opened 2020-10-05
RavenDB-15482 Editing nested fields under Settings in the Database Re…

Danielle9897:
…cord


ml054:
I wonder if we should handle that on client or server side?

Danielle9897:
Will move implementation to studio side

PR #11203, opened 2020-10-06
RavenDB-15045 Lighter health check endpoint

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15045

### DIFF
--- a/src/Raven.Client/Documents/Operations/DatabaseHealthCheckOperation.cs
+++ b/src/Raven.Client/Documents/Operations/DatabaseHealthCheckOperation.cs
@@ -5,7 +5,7 @@ using Sparrow.Json;
 
 namespace Raven.Client.Documents.Operations
 {
-    public class DatabaseHealthCheckOperation : IMaintenanceOperation
+    internal class DatabaseHealthCheckOperation : IMaintenanceOperation
     {
         public RavenCommand GetCommand(DocumentConventions conventions, JsonOperationContext context)
         {
@@ -16,7 +16,7 @@ namespace Raven.Client.Documents.Operations
         {
             public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, ServerNode node, out string url)
             {
-                url = $"{node.Url}/databases/{node.Database}/is-loaded";
+                url = $"{node.Url}/databases/{node.Database}/healthcheck";
                 return new HttpRequestMessage { Method = HttpMethod.Get };
             }
         }
--- a/src/Raven.Client/Documents/Operations/DatabaseHealthCheckOperation.cs
+++ b/src/Raven.Client/Documents/Operations/DatabaseHealthCheckOperation.cs
@@ -1,0 +1,24 @@
+using System.Net.Http;
+using Raven.Client.Documents.Conventions;
+using Raven.Client.Http;
+using Sparrow.Json;
+
+namespace Raven.Client.Documents.Operations
+{
+    public class DatabaseHealthCheckOperation : IMaintenanceOperation
+    {
+        public RavenCommand GetCommand(DocumentConventions conventions, JsonOperationContext context)
+        {
+            return new DatabaseHealthCheckCommand();
+        }
+
+        private class DatabaseHealthCheckCommand : RavenCommand
+        {
+            public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, ServerNode node, out string url)
+            {
+                url = $"{node.Url}/databases/{node.Database}/is-loaded";
+                return new HttpRequestMessage { Method = HttpMethod.Get };
+            }
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/Handlers/StatsHandler.cs
```diff
@@ -51,7 +51,13 @@ public Task Stats()
 
             return Task.CompletedTask;
         }
-
+        
+        [RavenAction("/databases/*/is-loaded", "GET", AuthorizationStatus.ValidUser, IsDebugInformationEndpoint = true)]
+        public Task IsLoaded()
+        {
+            return Task.CompletedTask;
```
ayende, line 59: Please return something here. This is a debug endpoint, and we expect some JSON

ayende (changes requested):

src/Raven.Server/Documents/Handlers/StatsHandler.cs
ppekrol: I think it would be better to rename it to `/databases/*/alive` or `/databases/*/healthcheck` and return no-content. There is no need to return anything.

src/Raven.Client/Documents/Operations/DatabaseHealthCheckOperation.cs
```diff
@@ -0,0 +1,24 @@
+using System.Net.Http;
+using Raven.Client.Documents.Conventions;
+using Raven.Client.Http;
+using Sparrow.Json;
+
+namespace Raven.Client.Documents.Operations
+{
+    public class DatabaseHealthCheckOperation : IMaintenanceOperation
```
ppekrol: internal

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -51,7 +51,7 @@ public class RequestExecutor : IDisposable
         private static readonly ConcurrentDictionary<string, Lazy<HttpClient>> GlobalHttpClientWithCompression = new ConcurrentDictionary<string, Lazy<HttpClient>>();
         private static readonly ConcurrentDictionary<string, Lazy<HttpClient>> GlobalHttpClientWithoutCompression = new ConcurrentDictionary<string, Lazy<HttpClient>>();
 
-        private static readonly GetStatisticsOperation FailureCheckOperation = new GetStatisticsOperation(debugTag: "failure=check");
+        private static readonly DatabaseHealthCheckOperation FailureCheckOperation = new DatabaseHealthCheckOperation();
```
ppekrol, line 56: What about backward compatibility? If you are connecting to a server version that does not have this endpoint then what will happen?

haludi:
> http://live-test.ravendb.net/databases/demo/healthcheck
> 
> This fails, how is this backward compatible?

The stats endpoint is still available.
The backward compatibility is only in the client - https://github.com/ravendb/ravendb/pull/11203/files#diff-e6c80c683c49344dd29a830753547288R1743

ayende:
In this case, that needs to go on 5.1, not 4.2, since we expect a 4.2 client to work with older 4.2 server

haludi:
> In this case, that needs to go on 5.1, not 4.2, since we expect a 4.2 client to work with older 4.2 server

As I check it, it works with an older server.
We try to use the new endpoint and if it fails with `ClientVersionMismatchException` we try with the stats endpoint.
This is not efficient while if the client communicates with an older server it always sends two requests.
But we use this health check only if a command failed.
I thought maybe to cache the decision which endpoint to use per `RequestExecutor`

ayende:
Okay, I missed that. Can you change it so we'll only do the first check once, and then remember that this server is older?

ayende:
Also, did you test that it works against older servers and newer ones?

haludi:
> Also, did you test that it works against older servers and newer ones?

I checked it against a server with and without the dedicate health check endpoint

haludi:
> Okay, I missed that. Can you change it so we'll only do the first check once, and then remember that this server is older?

At what level? In the `RequesExceuter`? session? document store?
Also, there can be a cluster with mix or the server can change while the client still working

ayende:
Per RequestExecuter per url

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1732,9 +1735,32 @@ private async Task CheckNodeStatusCallback(NodeStatus nodeStatus)
...
+                {
+                    await ExecuteAsync(serverNode, nodeIndex, context, FailureCheckOperation.GetCommand(Conventions, context), shouldRetry: false, sessionInfo: null,
+                        token: CancellationToken.None).ConfigureAwait(false);
+                }
+                else
+                {
+                    await ExecuteOldHealthCheck().ConfigureAwait(false);
+                }
+            }
+            catch (ClientVersionMismatchException e) when (e.Message.Contains("RouteNotFoundException"))
```
ppekrol: nameof(RouteNotFoundException)

PR #11205, opened 2020-10-06
RavenDB-15482 Flatten nested fields under Settings when editing the D…

Danielle9897:
…atabase Record

### DIFF
--- a/src/Raven.Studio/typescript/viewmodels/database/settings/databaseRecord.ts
+++ b/src/Raven.Studio/typescript/viewmodels/database/settings/databaseRecord.ts
@@ -151,7 +151,7 @@ class databaseRecord extends viewModelBase {
             .then(result => {
                 if (result.can) {
                     let dto = JSON.parse(this.documentText());
-                    dto = this.handleNestedSettings(dto);
+                    dto.Settings = this.flattenSettings(dto.Settings, "");
                     
                     new saveDatabaseSettingsCommand(this.activeDatabase(), dto, dto.Etag)
                         .execute()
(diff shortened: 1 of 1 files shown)


ml054:
Please make flatten code recursive.

src/Raven.Studio/typescript/viewmodels/database/settings/databaseRecord.ts
```diff
@@ -160,6 +162,51 @@ class databaseRecord extends viewModelBase {
...
+        if (!settings) {
+            return dto;
+        }
+        
+        const settingsKeys = Object.keys(settings);
+        if (settingsKeys.length === 0) {
+            return dto;
+        }
+
+        let flattenedSettings = { fs: {} };
```
ml054: I don't see the point to detect if we should flatten or not. Just check if settings are present if yes then call flatten and this is end of story. 

```
  function flattenObject(obj, parent, res = {}){
    for(let key in obj){
        let propName = parent ? parent + '.' + key : key;
        if(typeof obj[key] == 'object'){
            flattenObj(obj[key], propName, res);
        } else {
            res[propName] = obj[key];
        }
    }
    return res;
}
```

ml054: see: https://stackoverflow.com/questions/44134212/best-way-to-flatten-js-object-keys-and-values-to-a-single-depth-array

Danielle9897: done

ml054 (approved):

PR #11207, opened 2020-10-06
RavenDB-14286 Require inspection on new `RavenCommand`s for parallel execution if can be checked for fastest node

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-14286

### DIFF
--- a/test/FastTests/Client/RavenCommandTest.cs
+++ b/test/FastTests/Client/RavenCommandTest.cs
@@ -1,0 +1,97 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Raven.Client.Http;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace FastTests.Client
+{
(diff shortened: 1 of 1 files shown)


test/FastTests/Client/RavenCommandTest.cs
```diff
@@ -0,0 +1,97 @@
...
+        public RavenCommandTest(ITestOutputHelper output) : base(output)
+        {
+        }
+        
+        [Fact]
+        public void WhenCommandCanBeCheckedForFastestNode_ItCanRunInParallel()
+        {
+            var expected = new[]
+            {
+                "GetClusterTopologyCommand", "GetDatabaseTopologyCommand", "GetNodeInfoCommand", "GetRawStreamResultCommand", "GetTcpInfoCommand",
```
ml054, line 21: can we scan this using reflection - list is good for today - but it can easily be out of sync tomorrow.

PR #11210, opened 2020-10-07
RavenDB-15645 avoid segment split for different number of values

karmeli87:
If we append 10 entries when the first has 3 values and the rest only one,
we would call segment split 9 times.

### DIFF
--- a/src/Raven.Server/Documents/TimeSeries/SegmentHeader.cs
+++ b/src/Raven.Server/Documents/TimeSeries/SegmentHeader.cs
@@ -18,6 +18,14 @@ namespace Raven.Server.Documents.TimeSeries
         [FieldOffset(13)]
         public byte NumberOfValues;
         [FieldOffset(14)]
-        public fixed byte Reserved[2];
+        public SegmentVersion Version;
+        [FieldOffset(15)]
+        public fixed byte Reserved[1];
+    }
+
+    public enum SegmentVersion : byte
+    {
+        V50000 = 0,
+        V50001 = 1
     }
 }
--- a/src/Raven.Server/Documents/TimeSeries/SegmentHeader.cs
+++ b/src/Raven.Server/Documents/TimeSeries/SegmentHeader.cs
@@ -26,6 +26,7 @@ namespace Raven.Server.Documents.TimeSeries
     public enum SegmentVersion : byte
     {
         V50000 = 0,
-        V50001 = 1
+        V50001 = 1,
+        Current = V50001
     }
 }
--- a/src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
+++ b/src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
@@ -1215,7 +1215,7 @@ namespace Raven.Server.Documents.TimeSeries
 
                             if (EnsureNumberOfValues(segmentHolder.ReadOnlySegment.NumberOfValues, ref current))
                             {
-                                if (TryAppendToCurrentSegment(context, segmentHolder, appendEnumerator, out var newValueFetched))
+                                if (TryAppendToCurrentSegment(context, segmentHolder, appendEnumerator, current, out var newValueFetched))
                                     break;
 
                                 if (newValueFetched)
@@ -1302,16 +1302,15 @@ namespace Raven.Server.Documents.TimeSeries
             return segmentNumberOfValues == current.Values.Length;
         }
 
-        private bool TryAppendToCurrentSegment(
-            DocumentsOperationContext context,
+        private bool TryAppendToCurrentSegment(DocumentsOperationContext context,
             TimeSeriesSegmentHolder segmentHolder,
             IEnumerator<SingleResult> appendEnumerator,
+            SingleResult current,
             out bool newValueFetched)
         {
             var segment = segmentHolder.ReadOnlySegment;
             var slicer = segmentHolder.SliceHolder;
 
-            var current = appendEnumerator.Current;
             var lastTimestamp = segment.GetLastTimestamp(segmentHolder.BaselineDate);
             var nextSegmentBaseline = BaselineOfNextSegment(segmentHolder, current.Timestamp) ?? DateTime.MaxValue;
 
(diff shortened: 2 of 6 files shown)


src/Raven.Server/Documents/TimeSeries/SegmentHeader.cs
```diff
@@ -18,6 +18,14 @@ public unsafe struct SegmentHeader
...
-        public fixed byte Reserved[2];
+        public SegmentVersion Version;
+        [FieldOffset(15)]
+        public fixed byte Reserved[1];
+    }
+
+    public enum SegmentVersion : byte
+    {
+        V50000 = 0,
+        V50001 = 1
```
ppekrol: can we also have `Current = V50001` and use that in the Initialize?

karmeli87: done

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
```diff
@@ -2038,6 +2037,26 @@ private static void EnsureStatsAndDataIntegrity(DocumentsOperationContext contex
...
+                foreach (var state in segment.SegmentValues.Span)
+                {
+                    Debug.Assert(double.IsNaN(state.First) == false, "First is NaN");
+                    Debug.Assert(double.IsNaN(state.Last) == false, "Last is NaN");
+                    Debug.Assert(double.IsNaN(state.Min) == false, "Min is NaN");
+                    Debug.Assert(double.IsNaN(state.Max) == false, "Max is NaN");
+                    Debug.Assert(double.IsNaN(state.Sum) == false, "Sum is NaN");
+                    Debug.Assert(double.IsNaN(state.Count) == false, "Count is NaN");
+                    Debug.Assert(state.Count > 0, "Count is Zero");
+                }
```
ayende, line 2052: Why do we have all of those asserts here? If we are validating this, let's move all of this to a `[Conditional("DEBUG")]` method.

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
```diff
@@ -2038,6 +2037,26 @@ private static void EnsureStatsAndDataIntegrity(DocumentsOperationContext contex
...
+
+                if (name.Contains(TimeSeriesConfiguration.TimeSeriesRollupSeparator))
+                {
+                    var noNaN = segment.YieldAllValues(context, baseline: default, includeDead: false).All(x => x.Values.ToArray().All(y => double.IsNaN(y) == false));
+                    Debug.Assert(noNaN, "Rollup has NaN");
+                }
```
ayende, line 2058: Same here, this is debug only code, no?

src/Raven.Server/Documents/TimeSeries/TimeSeriesValuesSegment.cs
```diff
@@ -468,7 +473,13 @@ public IEnumerable<SingleResult> YieldAllValues(JsonOperationContext context, By
                     if (status == Dead && includeDead == false)
                         continue;
 
-                    var cur = baseline.AddMilliseconds(ts);
+                    var next = baseline.AddMilliseconds(ts);
+                    if (next == current && 
+                        Version == SegmentVersion.V50000) // fix legacy issue RavenDB-15617
```
ayende, line 478: What is the difference between the versions?

src/Raven.Server/Documents/TimeSeries/TimeSeriesValuesSegment.cs
```diff
@@ -569,6 +580,13 @@ public Span<byte> AsSpan()
             }
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static bool IsValid(ulong status, long value)
+        {
+            // valid = live and not NaN
+            return status == Live && (value & 0x7FFFFFFFFFFFFFFF) > 0x7FF0000000000000 == false;
```
ayende, line 587: wouldn't it be better to use `double.IsNaN(Convert.LongToDoubleBits(value))` ?

src/Raven.Server/Documents/TimeSeries/TimeSeriesValuesSegment.cs
```diff
@@ -747,6 +772,23 @@ private static void ThrowValuesOutOfRange(int numberOfValues)
...
+                return false;
+
+            if (NumberOfLiveEntries == 0)
+                return false;
+
+            foreach (var value in SegmentValues.Span)
+            {
+                if (double.IsNaN(value.Last))
+                    return true;
+            }
```
ayende, line 791: Why do we check all the items in the array?

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
karmeli87: It is all under `[Conditional("DEBUG")]`

src/Raven.Server/Documents/TimeSeries/TimeSeriesValuesSegment.cs
karmeli87: it is the impl of `double.IsNaN`, so I avoid the double conversion (from `long` to `double` and then back from `double` to `long`)

src/Raven.Server/Documents/TimeSeries/TimeSeriesValuesSegment.cs
karmeli87: different issue...  we might have duplicate timestamps :/

src/Raven.Server/Documents/TimeSeries/TimeSeriesValuesSegment.cs
karmeli87: Eventually this check is used to decide whether we can use the `StatefulTimestampValueSpan` of the segment or not.

ayende: can you add a comment explaining the reasoning

ayende (approved):

ayende:
Need schema upgrade

ayende:
We are reusing a reserved value. We don't _have_ to do anything, but it is a concern, no?

karmeli87:
Yes, we basically need only to bump the version

PR #11229, opened 2020-10-11
RavenDB-15693 Allow to boost on complex queries

ayende:


test/SlowTests/Issues/RavenDB-15693.cs
```diff
@@ -0,0 +1,43 @@
...
+                .AndAlso()
+                .OpenSubclause()
+                .Search(x => x.StrVal2, "b")
+                .OrElse()
+                .Search(x => x.StrVal3, "search")
+                .CloseSubclause()
+                .Boost(0.2m);
+            var queryBoost = q.ToString();
+            
+            Assert.Equal("from 'Docs' where search(StrVal1, $p0) and boost(search(StrVal2, $p1) or search(StrVal3, $p2), $p3)", queryBoost);
```
ppekrol, line 33: Query without boost would look like:
```
from 'Docs' where search(StrVal1, $p0) and (search(StrVal2, $p1) or search(StrVal3, $p2))
```

The missing `( )` inside the boost expression will not affect how Lucene generates (and then interprets) query?

ayende: No, the `()` is there to ensure order of operations, the `boost()` serves the same purpose.

ppekrol:
@ayende can you check failing tests? Also does it affect backward compatibility?

ayende:
Rebased on the latest and re-run those queries, they all work. Very strange. Will test again on CI.

I don't believe that this should have any impact on backward compatability of queries. Such queries were invalid, so cannot happen.

src/Raven.Client/Documents/Session/Tokens/CloseSubclauseToken.cs
```diff
@@ -4,6 +4,7 @@ namespace Raven.Client.Documents.Session.Tokens
 {
     public sealed class CloseSubclauseToken : QueryToken
     {
+        public string BoostParameterName { get; set; }
```
ppekrol: Issue with failing tests is here. We are using singleton.

src/Raven.Client/Documents/Session/Tokens/OpenSubclauseToken.cs
```diff
@@ -9,9 +9,12 @@ private OpenSubclauseToken()
         }
 
         public static readonly OpenSubclauseToken Instance = new OpenSubclauseToken();
+        public string BoostParameterName { get; set; }
```
ppekrol: Issue with failing tests is here. We are using singleton.

ppekrol:
Fixed.

PR #11235, opened 2020-10-12
RavenDB-15548 Custom combine to client config indexes

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15548

As I see it if we want to simplify the backward compatibility the following solution can be possible.
The client doesn't need to know about the changes at all.
And as an addition, we reduce config update while if there are **database** client config and the server-wide config changed there is no need to update.

This solution will work smoothly (I think) on a fresh cluster.
But when a user upgrades there are two problems:
 * When a `RequestExecutor` created before the upgrade it can hold `Etag` (hash combine from database & server-wide index) that would be considered as more updated in the new implementation while it is not.
 * When the cluster is combined of upgraded and unupgraded nodes so it can cause to endless client config update.

Another solution can be instead of replating the old `Etag` that sent to the client just add new property (to the header).
But this will require changes in the client as well

### DIFF
--- a/src/Raven.Server/Documents/DocumentDatabase.cs
+++ b/src/Raven.Server/Documents/DocumentDatabase.cs
@@ -1428,9 +1428,21 @@ namespace Raven.Server.Documents
 
         public bool HasClientConfigurationChanged(long index)
         {
-            var actual = Hashing.Combine(_lastClientConfigurationIndex, ServerStore.LastClientConfigurationIndex);
+            return _lastClientConfigurationIndex != -1 
+                ? IsServerIndexBiggerThenClient(_lastClientConfigurationIndex & 0xFFFFFFFF, index & 0xFFFFFFFF) 
+                : IsServerIndexBiggerThenClient(ServerStore.LastClientConfigurationIndex & 0xFFFFFFFF, (index >> 4*8));
+        }
 
-            return index != actual;
+        private static bool IsServerIndexBiggerThenClient(long server, long client)
+        {
+            var diff = server - client;
+            var abs = Math.Abs(diff);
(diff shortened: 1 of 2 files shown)


ayende:
The value is already per server, no? And the old clients will simply see that the value changed and update it,no?

haludi:
#### The value is already per server, no? 
The value is per `RequestExecutor` and the origin issue is because a value the client got from one node sent to another node while one of the nodes is not up to date.

#### And the old clients will simply see that the value changed and update it, no?
I am not sure I am following the question but if I do
let say the current client configuration index for database is x and for server-wide is y
and the Hash.Combine(x,y) is greater (`IsServerIndexBiggerThenClient`) than the result of the new "combine" method (`CombineClientConfigurationIndexes)`.
The client (`RequesExecutor`) send the value it has and the server check and decide the client has more up to date client config so it doesn't send to the client `context.HttpContext.Response.Headers[Constants.Headers.RefreshClientConfiguration] = "true";` and the client just remains with the old configuration.

ayende:
Can we keep track of the node id that gave us the client config?
That will allow us to know if different servers give different results.

I don't like trying to overload the values of the long in this manner, you are going to get into trouble after a while.
Consider a long running server that will reach 4 billion changes at some point?

haludi:
#### Can we keep track of the node id that gave us the client config? That will allow us to know if different servers give different results.
But this will not let us know which one is more up to date. 
Maybe I can save a short history and check it before updating the config?

#### I don't like trying to overload the values of the long in this manner, you are going to get into trouble after a while. Consider a long running server that will reach 4 billion changes at some point?

This was handled in `IsServerIndexBiggerThenClient`.
It will get into trouble only if the difference between the new value and the old value is more then 2Gb.

PR #11242, opened 2020-10-12
RavenDB-15700 Fixing map-reduce index debug endpoint used by map-redu…

arekpalinski:
…ce visualizer. Taking into account that there might be empty aggregation results.


src/Raven.Server/Documents/Indexes/Debugging/IndexDebugExtensions.cs
```diff
@@ -426,7 +431,10 @@ private static BlittableJsonReaderObject GetReduceResult(ulong reduceKeyHash, In
                      .Query(query, null, fieldsToFetch, new Reference<int>(), new Reference<int>(), retriever, ctx, null, CancellationToken.None)
                     .ToList();
 
-                if (result.Count != 1)
+                if (result.Count == 0)
+                    return context.ReadObject(new DynamicJsonValue(), "debug-reduce-result");
+
+                if (result.Count > 1)
```
ayende, line 437: But that is what happened here, no? We had two values for the same reduce key

arekpalinski: No, what happened here was the case above - `result.Count == 0` because we filtered out the results in a given page during the reduce phase.

ayende:
image

ayende:
See here: https://issues.hibernatingrhinos.com/issue/RavenDB-15702

arekpalinski:
Interesting. I didn't run into this using the same (I believe) set of data. I'll check .ravendump file you attached.

At glance it looks we have the reduce key hash collision (which is valid case and properly handled in the indexing code, but debug endpoint needs to take it into account as well).

PR #11257, opened 2020-10-15
RavenDB-15616 use the initial value if not found in the json

karmeli87:


karmeli87:
Open questions if this change is acceptable:
1. Should we replace all `Expression.Default` within `GetValue`
2. Do we want to backport this

ayende:
What is the code equivalent to this change?
Why do we _need_ to do anything here?

ppekrol:
@karmeli87 can we have a test for this?

From what I see I'm assuming that if I'm setting sth in the ctor and I do not have it in JSON then instead of setting default value, it will use the one from ctor?

karmeli87:
This issue is, that stuff like: https://github.com/ravendb/ravendb/blob/378ae39eb1531a243f6d968a47d02721294fd62d/src/Raven.Client/Documents/Operations/Replication/PullReplicationAsSink.cs#L9
are not respected if the property is missing from the json (blittable).
In that case we initialize to default, so this change should retain the initial value.

the eqv. is something like that:
```
if(tryGetFromJson(...) == false)
{
  try
  {
     return readTheCurrentValue();
  }
  catch
  {
  return default;
  }
}
```

ayende:
Instead of generating this code, can we do that in line, and then just reuse the captured value? 
We can ensure that we only do that to `ValueType`, not to class instances (which may then be shared)

ayende:
In other words, during the serialization construction, run this code, get the value, and then use that?

karmeli87:
after changes we generate the following:
```
.Lambda #Lambda1<System.Func`2[Sparrow.Json.BlittableJsonReaderObject,FastTests.Issues.RavenDB_7493+Foo]>(Sparrow.Json.BlittableJsonReaderObject $json)
{
    .Block(
        System.String $String,
        FastTests.Issues.RavenDB_7493+Foo $result) {
        $result = .New FastTests.Issues.RavenDB_7493+Foo();
        .If (
            .Call $json.TryGet(
                "Bar",
                $String)
        ) {
            $result.Bar = $String
        } .Else {
            .Default(System.Void)
        };
        $result.ListOfStrings = .Call Sparrow.Json.JsonDeserializationBase.ToCollectionOfString(
            $json,
            "ListOfStrings");
        $result
    }
}
```

karmeli87:
Just noticed that `List` or `Dictionary` are _always_ instantiated, even if they are not in the `json` or initially `null`.
https://github.com/ravendb/ravendb/blob/ce71904c0cc6d4ac33be8a22d732da3c94e4c374/src/Sparrow/Json/JsonDeserializationBase.cs#L650
But I suggest to have different ticket for that in case we even what to fix it.

ppekrol (approved):

ayende (approved):

ayende:
The one thing that worries me about this is that we might be using a custom comparer in the case of a dictionary. But it would be easier to just forbid that then support it.

ppekrol:
We already have 'JsonDeserializationStringDictionaryAttribute', but I'm wondering if it is not better to take whatever was created in ctor (e.g. dict with custom comparer) and if it is null then create default instance?

ayende:
I would rather avoid this completely, given that we have no use for that right now

PR #11262, opened 2020-10-16
RavenDB-15632

ppekrol:
- dispose the command in ETL to release the blittables
- set max cache size for global pool items

### DIFF
--- a/src/Sparrow/Json/BlittableJsonDocumentBuilder.cs
+++ b/src/Sparrow/Json/BlittableJsonDocumentBuilder.cs
@@ -4,6 +4,7 @@ using System.Runtime.CompilerServices;
 using Sparrow.Collections;
 using Sparrow.Exceptions;
 using Sparrow.Json.Parsing;
+using Sparrow.Platform;
 using Sparrow.Threading;
 
 namespace Sparrow.Json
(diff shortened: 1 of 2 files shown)


src/Sparrow/Json/BlittableJsonDocumentBuilder.cs
```diff
@@ -12,16 +13,20 @@ public sealed class BlittableJsonDocumentBuilder : IDisposable
     {
         private class GlobalPoolItem
         {
+            private static int MaxCacheSize = PlatformDetails.Is32Bits == false
+                ? 8 * 1024
+                : 2 * 1024;
```
ayende, line 18: Should this be this high on 32 bits?
```suggestion
                : 512;
```

ayende (approved):

src/Sparrow/Json/BlittableJsonDocumentBuilder.cs
ppekrol: Not sure if we want to be that conservative here? In other places the ratio is 1:4 so I thought that we could use that ratio here too.

ppekrol:
We found a better approach. Will open a PR soon.

PR #11265, opened 2020-10-16
RavenDB-15632 Cache the PropertyTag for the duration of a single object construction and reuse immediately

ppekrol:


src/Sparrow/Json/BlittableJsonDocumentBuilder.cs
```diff
@@ -115,24 +58,22 @@ public void Reset()
             _debugTag = null;
             _mode = UsageMode.None;
             _continuationState.Clear();
```
ppekrol: Shouldn't we return items to the cache here?

src/Sparrow/Json/BlittableJsonDocumentBuilder.cs
```diff
@@ -115,24 +58,22 @@ public void Reset()
...
         public void Renew(string debugTag, UsageMode mode)
         {
-            _writeToken = default(WriteToken);
+            _writeToken = default;
             _debugTag = debugTag;
             _mode = mode;
 
-            _continuationState.Clear();          
-            _cacheItem.Reset();
+            _continuationState.Clear();
```
ppekrol: Shouldn't we return items to the cache here?

ayende (approved):

PR #11270, opened 2020-10-19
RavenDB-15622 database LastAccessTime updated upon each transaction

karmeli87:


ppekrol (approved):

src/Raven.Server/Documents/TombstoneCleaner.cs
```diff
@@ -158,6 +158,8 @@ internal class DeleteTombstonesCommand : TransactionOperationsMerger.MergedTrans
 
             public long NumberOfTombstonesDeleted { get; private set; }
 
+            public override bool UpdateAccessTime => false;
```
ayende, line 161: Can we do this a regular field and we set it? That avoids the virtual method call at all time

karmeli87: done: https://github.com/ravendb/ravendb/pull/11272

PR #11276, opened 2020-10-19
RavenDB-15730 - Failed to start the server after machine restart during initialization

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-15730

### DIFF
--- a/src/Raven.Server/ServerWide/ClusterStateMachine.cs
+++ b/src/Raven.Server/ServerWide/ClusterStateMachine.cs
@@ -2808,21 +2808,11 @@ namespace Raven.Server.ServerWide
             }
         }
 
-        public static unsafe BlittableJsonReaderObject ReadInternal<T>(TransactionOperationContext<T> context, out long etag, Slice key, bool ignoreMissingTable = false)
+        public static unsafe BlittableJsonReaderObject ReadInternal<T>(TransactionOperationContext<T> context, out long etag, Slice key)
             where T : RavenTransaction
         {
             var items = context.Transaction.InnerTransaction.OpenTable(ItemsSchema, Items);
-            if (items == null && ignoreMissingTable)
-            {
-                // can happen if we fail on first run of the server
-                // we created the system store but failed to create all tables
-                etag = 0;
-                return null;
-            }
-
-            Debug.Assert(items != null);
-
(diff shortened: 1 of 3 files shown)


src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -2808,10 +2808,20 @@ public BlittableJsonReaderObject Read<T>(TransactionOperationContext<T> context,
             }
         }
 
-        public static unsafe BlittableJsonReaderObject ReadInternal<T>(TransactionOperationContext<T> context, out long etag, Slice key)
+        public static unsafe BlittableJsonReaderObject ReadInternal<T>(TransactionOperationContext<T> context, out long etag, Slice key, bool ignoreMissingTable = false)
             where T : RavenTransaction
         {
             var items = context.Transaction.InnerTransaction.OpenTable(ItemsSchema, Items);
+            if (items == null && ignoreMissingTable)
```
ayende: Why aren't we checking and then creating the tables, then?

grisha-kotler: We do that after the _Schema Upgrade_.

ppekrol: Why not simply always return `etag = 0` and `null` when items are null instead of introducing a parameter?

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -2466,18 +2466,18 @@ public NodeInfo GetNodeInfo()
 
         public License LoadLicense()
         {
-            return LoadLicense(ContextPool);
+            return LoadLicense(ContextPool, ignoreMissingTable: false);
```
ppekrol: Should be `true` ?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
grisha-kotler: done

src/Raven.Server/ServerWide/ServerStore.cs
grisha-kotler: removed

PR #11280, opened 2020-10-20
RavenDB-15145 When adding PullReplicationAsSink with a certificate wi…

efratshenhar:
…thout a private key, give good error

### DIFF
--- a/src/Raven.Client/Documents/Operations/Replication/UpdatePullReplicationAsSinkOperation.cs
+++ b/src/Raven.Client/Documents/Operations/Replication/UpdatePullReplicationAsSinkOperation.cs
@@ -24,11 +24,13 @@ namespace Raven.Client.Documents.Operations.Replication
             if (pullReplication.CertificateWithPrivateKey != null)
             {
                 var certBytes = Convert.FromBase64String(pullReplication.CertificateWithPrivateKey);
-                var certificate = new X509Certificate2(certBytes,
+                using (var certificate = new X509Certificate2(certBytes,
                     pullReplication.CertificatePassword,
-                    X509KeyStorageFlags.Exportable | X509KeyStorageFlags.MachineKeySet);
-                if (certificate.HasPrivateKey == false)
-                    throw new AuthorizationException("Certificate with private key is required");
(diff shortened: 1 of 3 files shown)


src/Raven.Client/Documents/Operations/Replication/UpdatePullReplicationAsSinkOperation.cs
```diff
@@ -18,6 +20,16 @@ public class UpdatePullReplicationAsSinkOperation : IMaintenanceOperation<Modify
         public UpdatePullReplicationAsSinkOperation(PullReplicationAsSink pullReplication)
         {
             _pullReplication = pullReplication;
+
+            if (pullReplication.CertificateWithPrivateKey != null)
+            {
+                var certBytes = Convert.FromBase64String(pullReplication.CertificateWithPrivateKey);
+                var certificate = new X509Certificate2(certBytes,
```
ayende: We need to dispose this instance, I think

test/SlowTests/Smuggler/BackupDatabaseRecordTests.cs
```diff
@@ -39,6 +40,13 @@ public BackupDatabaseRecordTests(ITestOutputHelper output) : base(output)
         public async Task CanExportAndImportDatabaseRecord()
         {
             var file = Path.GetTempFileName();
+            var dummy = GenerateAndSaveSelfSignedCertificate(true);
+            string privateKey;
+            using (var pullReplicationCertificate =
+                new X509Certificate2(dummy.ServerCertificatePath, (string)null, X509KeyStorageFlags.MachineKeySet | X509KeyStorageFlags.Exportable))
```
ppekrol, line 46: Is this correct? To dispose the certificate created by `GenerateAndSaveSelfSignedCertificate` ? It is used by the whole test suite. Maybe it is fine, because we are creating a new instance here, but I would like to double check.

karmeli87: I think it is fine, since we use `dummy.ServerCertificatePath` (which is a path and not the certificate itself).

PR #11285, opened 2020-10-20
RavenDB-13479: Add include time series last range by time or count

garayx:
for query, load and subscription


src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
```diff
@@ -109,16 +95,36 @@ public Task Stats()
...
                     writer.WriteEndObject();
-                }    
+                }
             }
-            
+
             return Task.CompletedTask;
         }
 
+        internal static void GetTimesSeriesNames(Document document, List<string> timeSeriesNames)
```
ppekrol: change return type to List<string>

src/Raven.Server/Documents/Includes/IncludeTimeSeriesCommand.cs
```diff
@@ -37,33 +46,138 @@ public void Fill(Document document)
                 if (Results.ContainsKey(docId))
                     continue;
 
-                var rangeResults = GetTimeSeriesForDocument(docId, kvp.Value);
-
-                Results.Add(docId, rangeResults);
-
+                var count = kvp.Value.Count(x => x.Name == Constants.TimeSeries.All);
```
ppekrol: avoid Linq, by the way... kv.Value.Any is better because it looks for a first occurence

ppekrol: it looks like you are doing this calculation for each document and it can be done once in ctor?

src/Raven.Server/Documents/Includes/IncludeTimeSeriesCommand.cs
```diff
@@ -37,33 +46,138 @@ public void Fill(Document document)
...
+                        Results.Add(docId, rangeResults);
+                        break;
+                    }
+                    case 1:
+                    {
+                        // get all ts names
+                        var timeSeriesNames = new List<string>();
+                        TimeSeriesHandler.GetTimesSeriesNames(document, timeSeriesNames);
+                        var arr = new HashSet<AbstractTimeSeriesRange>();
+                        switch (kvp.Value.First())
```
ppekrol: why first?

src/Raven.Server/Documents/Includes/IncludeTimeSeriesCommand.cs
garayx: refactored this code

src/Raven.Server/Documents/Includes/IncludeTimeSeriesCommand.cs
garayx: if there is a request for range on `@all_timeseries` we forbidding to use additional ts includes, so it always going to be single.

ppekrol (approved):

ppekrol:
LGTM, @garayx can you rebase?

src/Raven.Server/Documents/Handlers/TimeSeriesHandler.cs
```diff
@@ -411,55 +417,105 @@ internal static async Task WriteTimeSeriesRangeResults(DocumentsOperationContext
...
+                if (first == false)
+                    writer.WriteComma();
+
+                first = false;
+
+                writer.WritePropertyName(name);
 
-                            await writer.MaybeOuterFlushAsync();
+                writer.WriteStartArray();
+                {
```
karmeli87: redundant ?

src/Raven.Server/Documents/Includes/IncludeTimeSeriesCommand.cs
```diff
@@ -37,33 +46,130 @@ public void Fill(Document document)
...
-                {
-                    list.Add(timeSeriesRangeResult);
+                    case TimeSeriesRange r:
+                        result = TimeSeriesHandler.GetTimeSeriesRange(_context, docId, r.Name, r.From ?? DateTime.MinValue, r.To ?? DateTime.MaxValue, ref start, ref pageSize);
+                        break;
+                    case TimeSeriesTimeRange tr:
+                        {
+                            var stats = GetTimeSeriesStats(docId, tr.Name);
+                            if (stats.Count == 0)
+                                continue;
```
karmeli87, line 106: Do we want to skip or add an empty result?  Do we have a test for this case?

src/Raven.Server/Documents/Includes/IncludeTimeSeriesCommand.cs
```diff
@@ -37,33 +46,130 @@ public void Fill(Document document)
...
+                            }
+
+                            result = TimeSeriesHandler.GetTimeSeriesRange(_context, docId, tr.Name, from, to, ref start, ref pageSize);
+                        }
+                        break;
+                    case TimeSeriesCountRange cr:
+                        {
+                            var stats = GetTimeSeriesStats(docId, cr.Name);
+                            if (stats.Count == 0)
+                                continue;
```
karmeli87, line 126: same

src/Raven.Server/Documents/Includes/IncludeTimeSeriesCommand.cs
```diff
@@ -37,33 +46,130 @@ public void Fill(Document document)
...
+
+                            if (stats.Count <= cr.Count)
```
karmeli87: Why not to pass cr.Count to page size? or `TimeSeriesCountRange` is something else?

I think it should have `From` & `Count` ?

garayx:
@ppekrol 
done

aviv86 (approved):
LGTM

src/Raven.Server/Documents/Includes/IncludeTimeSeriesCommand.cs
```diff
@@ -1,29 +1,37 @@
...
             _context = context;
             _timeSeriesRangesBySourcePath = timeSeriesRangesBySourcePath;
 
+            _timeSeriesStatsPerDocumentId = new Dictionary<string, Dictionary<string, (long Count, DateTime Start, DateTime End)>>(StringComparer.OrdinalIgnoreCase);
             Results = new Dictionary<string, Dictionary<string, List<TimeSeriesRangeResult>>>(StringComparer.OrdinalIgnoreCase);
         }
 
         public void Fill(Document document)
         {
+            if (document == null)
```
karmeli87, line 32: When can it be `null` ?

src/Raven.Client/Documents/Session/Loaders/IncludeBuilder.cs
```diff
@@ -492,24 +619,138 @@ private void IncludeTimeSeries(string alias, string name, DateTime? from, DateTi
...
-        public bool Equals(TimeSeriesRange x, TimeSeriesRange y)
+        public bool Equals(AbstractTimeSeriesRange x, AbstractTimeSeriesRange y)
         {
             return string.Equals(x?.Name, y?.Name, StringComparison.OrdinalIgnoreCase);
         }
 
-        public int GetHashCode(TimeSeriesRange obj)
+        public int GetHashCode(AbstractTimeSeriesRange obj)
         {
             return obj.Name.GetHashCode();
```
karmeli87: Wonder if this should be `ToLowerInvariant'

src/Raven.Server/Documents/TcpHandlers/SubscriptionConnection.cs
```diff
@@ -159,6 +162,12 @@ public async Task InitAsync()
                     throw new SubscriptionInvalidStateException($"Subscription with ID '{SubscriptionId}' cannot be opened because it requires the protocol to support Counter Includes.");
             }
 
+            if (_supportedFeatures.Subscription.TimeSeriesIncludes == false)
+            {
+                if (Subscription.TimeSeriesIncludes != null && Subscription.TimeSeriesIncludes.TimeSeries.Count > 0)
+                    throw new SubscriptionInvalidStateException($"Subscription with ID '{SubscriptionId}' cannot be opened because it requires the protocol to support TimeSeries Includes.");
```
karmeli87, line 168: This is if an old client try to connect, right? What will happened here?

karmeli87:
LGTM too!, Just few nitpicking comments :-)

src/Raven.Server/Documents/Includes/IncludeTimeSeriesCommand.cs
garayx: skip,
for reference on `TimeSeriesRange` if `result == null` dictionary returned without empty result

src/Raven.Server/Documents/Includes/IncludeTimeSeriesCommand.cs
garayx: its case when last count requested is bigger then total count of ts, so I return all
no, the TimeSeriesRangeType is Last

src/Raven.Server/Documents/Includes/IncludeTimeSeriesCommand.cs
garayx: if requested doc id wasnt found in the storage

src/Raven.Server/Documents/TcpHandlers/SubscriptionConnection.cs
garayx: the worker will be disconnected

PR #11306, opened 2020-10-23
RavenDB-15765 Avoid recursion in in queries (with adjustments)

ppekrol:


src/Raven.Server/Documents/Queries/LuceneIntegration/TermsMatchQuery.cs
```diff
@@ -69,7 +68,7 @@ public override bool Next(IState state)
                         {
                             break; // search the next term
                         }
-                    }
+                    } while (termEnum.Next(state));
```
ayende, line 71: This isn't correct, you remain on the same term that you are already know is too small

ayende:
This is superseded by: https://github.com/ravendb/ravendb/pull/11308

PR #11307, opened 2020-10-23
RavenDB-14695 Reduce the cost of in queries

ayende:


ayende:
@ppekrol Note that https://github.com/ravendb/ravendb/pull/11308 is the same, but for v4.2, your call on how to best merge these.

ayende:
Will merge to 4.2 first

PR #11308, opened 2020-10-24
RavenDB-15765 Avoid recursion in in queries

ayende:
RavenDB-14695 Modify the manner in which we handle in queries.
If we have a low amount of items, we'll use a lazy approach with a Disjunction operation
If we have a large amount of items (> 128), we'll read them to a bitmap and scan them from there. Results show significant improvement in several key scenarios.


arekpalinski:
Related Lucene PR: https://github.com/ravendb/lucenenet/pull/18

src/Raven.Server/Documents/Queries/LuceneIntegration/TermsMatchQuery.cs
```diff
@@ -19,115 +22,191 @@ public TermsMatchQuery(string field, List<string> matches)
...
+                Similarity similarity = _parent.GetSimilarity(_searcher);
+                if(_parent.Matches.Count > 128)
+                    return new LazyInitTermMatchScorer(_parent, reader, state, similarity);
+
+                var scorers = new Scorer[_parent.Matches.Count];
+                int index = 0;
+                foreach (var match in _parent.Matches)
+                {
+                    var termDocs = reader.TermDocs(new Term(_parent.Field, match),state);
+                    var scorer = new TermScorer(this, termDocs,similarity, reader.Norms(_parent.Field, state));
```
arekpalinski: Can we call ` reader.Norms(_parent.Field, state)` just once before the loop and pass the returned array here?

src/Raven.Server/Documents/Queries/LuceneIntegration/TermsMatchQuery.cs
```diff
@@ -19,115 +22,191 @@ public TermsMatchQuery(string field, List<string> matches)
...
+                _queryWeight *= norm;
+            }
+
+            public override Scorer Scorer(IndexReader reader, bool scoreDocsInOrder, bool topScorer, IState state)
+            {
+                Similarity similarity = _parent.GetSimilarity(_searcher);
+                if(_parent.Matches.Count > 128)
+                    return new LazyInitTermMatchScorer(_parent, reader, state, similarity);
+
+                var scorers = new Scorer[_parent.Matches.Count];
```
arekpalinski: It would be nice to use `ArrayPool` here but I don't see where we could return it back to the pool.

src/Raven.Server/Documents/Queries/LuceneIntegration/TermsMatchQuery.cs
```diff
@@ -19,115 +22,191 @@ public TermsMatchQuery(string field, List<string> matches)
...
-                {
-                    _endEnum = true;
-                    return;
-                }
-                MoveToCurrentTerm(state);
+                _similarity = similarity;
             }
 
-            private void MoveToCurrentTerm(IState state)
+            private EagerTermMatchScorer InitIfNeeded()
```
arekpalinski, line 134: Maybe add `[MethodImpl(MethodImplOptions.AggressiveInlining)]` ?

src/Raven.Server/Documents/Queries/LuceneIntegration/TermsMatchQuery.cs
ayende: Yes, there should be no reason to do so.

src/Raven.Server/Documents/Queries/LuceneIntegration/TermsMatchQuery.cs
ayende: We can probably override the `DisjunctionMaxScorer` methods and detect when it is at end. Not sure how useful that would be.

src/Raven.Server/Documents/Queries/LuceneIntegration/TermsMatchQuery.cs
ayende: Probably not, this is small enough that we should let the JIT decide.

PR #11309, opened 2020-10-25
RavenDB-15548 Modify database client configuration Etag to use same incremental Etag as server wide client configuration

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15548

Two remain issues will remain as discussed:
* Mix cluster (some node with this fix and some not) can cause unnecessary client configuration update and/or not update a necessary configuration.
* If the client cached a higher value then the current incremental ETag (returned from Hash.Combine of database and server-wide client configuration) then the server will think the client has a more up to date client configuration.

### DIFF
--- a/src/Raven.Server/ServerWide/ClusterStateMachine.cs
+++ b/src/Raven.Server/ServerWide/ClusterStateMachine.cs
@@ -1345,12 +1345,12 @@ namespace Raven.Server.ServerWide
                         }
                     }
 
-                    bool isClientConfigChanged;
+                    bool shouldSetClientConfigEtag;
                     var dbId = Constants.Documents.Prefix + addDatabaseCommand.Name;
                     using (var oldDatabaseRecord = Read(context, dbId, out _))
                     {
                         VerifyUnchangedTasks(oldDatabaseRecord);
-                        isClientConfigChanged = IsClientConfigChanged(newDatabaseRecord, oldDatabaseRecord);
+                        shouldSetClientConfigEtag = ShouldSetClientConfigEtag(newDatabaseRecord, oldDatabaseRecord);
                     }
 
                     using (var databaseRecordAsJson = UpdateDatabaseRecordIfNeeded())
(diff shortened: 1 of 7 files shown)


ayende (commented):
Something that we can also do.
We can check the max value of the _cluster_ etag, and if the value is greater than that, and we are in a stable cluster (follower / leader with more than 10 sec), overwrite this, no?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1449,6 +1459,17 @@ BlittableJsonReaderObject UpdateDatabaseRecordIfNeeded()
             }
         }
 
+        private static bool IsClientConfigChanged(BlittableJsonReaderObject newDatabaseRecord, BlittableJsonReaderObject oldDatabaseRecord)
```
ppekrol: HasClientConfigurationChanged

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1449,6 +1459,17 @@ BlittableJsonReaderObject UpdateDatabaseRecordIfNeeded()
...
+        {
+            const string clientPropName = nameof(DatabaseRecord.Client);
+            if (newDatabaseRecord.TryGet(clientPropName, out BlittableJsonReaderObject newDbClientConfig) == false || newDbClientConfig == null)
```
ppekrol, line 1465: Not sure that I follow. If new record does not have a configuration then it didn't change (assuming that the old one has) ?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1342,82 +1360,74 @@ private unsafe List<string> AddDatabase(TransactionOperationContext context, Bli
...
+                                    throw new RachisInvalidOperationException(
+                                        $"Cannot update {task} configuration with DatabaseRecord. Please use a dedicated operation to update the {task} configuration.");
                             }
                         }
+
                     }
 
                     BlittableJsonReaderObject UpdateDatabaseRecordIfNeeded()
                     {
+                        if (isClientConfigChanged)
```
ppekrol: hasClientConfigurationChanged

src/Raven.Server/ServerWide/ClusterStateMachine.cs
haludi: The goal is to give the new client configuration the ETag of the database record.
And to do that only if the client configuration changed.
If there is no new client configuration so there is no configuration to set the ETag.
Maybe if I will change the method name (and variable) to `ShouldSetClientConfigEtag` or something similar it will be more clear?

ppekrol: Let's go with that. Yes.

PR #11310, opened 2020-10-25
RavenDB-15706 better error when using bulk insert without db

efratshenhar:


karmeli87 (approved):

src/Raven.Client/Documents/DocumentStore.cs
```diff
@@ -157,8 +157,7 @@ public override RequestExecutor GetRequestExecutor(string database = null)
         {
             AssertInitialized();
 
-            if (database == null)
-                database = Database;
+            database ??= Database ?? throw new ArgumentNullException(nameof(Database), "Document store database name cannot be empty");
```
ppekrol: This needs to be done explicitly in BulkInsert. Like we do in Session.
Check here:

https://github.com/ravendb/ravendb/blob/52bb0beee6fe674b6f245433bbc82e442709e7d2/src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs#L217-L219

ppekrol (commented):
Also having a test would be nice.

src/Raven.Client/Documents/BulkInsert/BulkInsertOperation.cs
```diff
@@ -217,6 +217,8 @@ public BulkInsertOperation(string database, IDocumentStore store, CancellationTo
 
             _token = token;
             _conventions = store.Conventions;
+            if (string.IsNullOrWhiteSpace(database))
+                throw new ArgumentNullException(nameof(database), "Document store database name cannot be empty");
```
karmeli87: Please make it more user friendly, similar to what we have here: 
https://github.com/ravendb/ravendb/blob/f2fad77b13ce1f94a80b84c549fd615ad70db1c5/src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs#L1119

PR #11311, opened 2020-10-25
RavenDB-15741 Move EncryptionBuffersPool to per core model.

ayende:
* We'll keep track of encryption buffer by power of 2 _page sizes_. In other words, 1,2,4,8,16 .. 128 pages (8KB ... 1 MB)
* Beyond that, we'll just allocate / deallocate directly.
* All data is shared on a per core model, instead of concurrent stack model

### DIFF
--- a/src/Voron/Impl/EncryptionBuffersPool.cs
+++ b/src/Voron/Impl/EncryptionBuffersPool.cs
@@ -5,23 +5,24 @@ using System.Linq;
 using System.Threading;
 using Sparrow;
 using Sparrow.Binary;
+using Sparrow.Json;
 using Sparrow.Json.Parsing;
 using Sparrow.LowMemory;
 using Sparrow.Server;
 using Sparrow.Server.Platform;
 using Sparrow.Threading;
 using Sparrow.Utils;
+using Voron.Global;
 
 namespace Voron.Impl
 {
     public unsafe class EncryptionBuffersPool : ILowMemoryHandler
     {
         public static EncryptionBuffersPool Instance = new EncryptionBuffersPool();
-
-        private readonly long _maxBufferSizeToKeepInBytes = new Size(8, SizeUnit.Megabytes).GetValue(SizeUnit.Bytes);
+        private const int MaxNumberOfPagesToCache = 128; // 128 * 8K = 1 MB, beyond that, we'll not both 
         private readonly MultipleUseFlag _isLowMemory = new MultipleUseFlag();
         private readonly MultipleUseFlag _isExtremelyLowMemory = new MultipleUseFlag();
-        private readonly ConcurrentStack<NativeAllocation>[] _items;
+        private readonly PerCoreContainer<NativeAllocation>[] _items;
         private readonly Timer _cleanupTimer;
         private DateTime _lastAllocationsRebuild = DateTime.UtcNow;
         private long _generation;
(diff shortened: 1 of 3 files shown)


src/Voron/Impl/EncryptionBuffersPool.cs
```diff
@@ -77,7 +79,8 @@ public void Return(byte* ptr, long size, NativeMemory.ThreadStats allocatingThre
             size = Bits.PowerOf2(size);
             Sodium.sodium_memzero(ptr, (UIntPtr)size);
 
-            if (Disabled || size > _maxBufferSizeToKeepInBytes || (_isLowMemory.IsRaised() && generation < Generation))
+            if (Disabled || size / Constants.Storage.PageSize > MaxNumberOfPagesToCache ||
```
karmeli87, line 82: You are modified the `size` and If the size is larger than 1MB, Will you return here more than you allocated??

karmeli87: And zero it beyond the range.

ayende: It is fine to return more than the requested size, yes.
We aren't zeroing the data here, the caller will do that when they return the data back.

ayende:
@karmeli87 Thanks, that was indeed an issue

src/Voron/Impl/EncryptionBuffersPool.cs
```diff
@@ -32,31 +33,32 @@ public unsafe class EncryptionBuffersPool : ILowMemoryHandler
...
 
-        public byte* Get(int size, out NativeMemory.ThreadStats thread)
+        public byte* Get(int numberOfPages, out int size, out NativeMemory.ThreadStats thread)
         {
-            size = Bits.PowerOf2(size);
+            numberOfPages = Bits.PowerOf2(numberOfPages); ;
+            size = numberOfPages * Constants.Storage.PageSize;
 
-            if (Disabled || size > _maxBufferSizeToKeepInBytes)
+            if (Disabled || numberOfPages > MaxNumberOfPagesToCache)
```
karmeli87: Even here you will be over-allocating alot, just like before.
let's say you requested 275,000 pages (~2.1GB), next power of 2 of that is 2^19 = 524,288 pages which is 4GB of size.
I guess you need to keep the original numberOfPages..

ayende: Sigh, sorry, that is me being silly. I wanted to avoid power of 2 if the size is greater than > 1MB

src/Voron/Impl/Paging/CryptoPager.cs
```diff
@@ -198,9 +198,10 @@ internal override void UnprotectPageRange(byte* start, ulong size, bool force =
 
             var pageHeader = (PageHeader*)pagePointer;
 
-            var size = VirtualPagerLegacyExtensions.GetNumberOfPages(pageHeader) * Constants.Storage.PageSize;
+            int numberOfPages = VirtualPagerLegacyExtensions.GetNumberOfPages(pageHeader);
+            var size = numberOfPages * Constants.Storage.PageSize;
```
ppekrol: Is this used?

ppekrol (approved):
LGTM

PR #11313, opened 2020-10-25
RavenDB-15689 Add a copy button to examples

Danielle9897:


src/Raven.Studio/typescript/common/copyToClipboard.ts
```diff
@@ -1,6 +1,16 @@
 ﻿import messagePublisher = require("common/messagePublisher");
 
 class copyToClipboard {
+
+    constructor(private htmlElement: HTMLElement, private textArray: Array<string>) {
```
ml054: why do we need html element here and some list?

src/Raven.Studio/typescript/common/copyToClipboard.ts
```diff
@@ -1,6 +1,16 @@
...
+        this.htmlElement = htmlElement;
+        this.textArray = textArray;
+    }
+    
+    copyText(arrayEntry: number, msg?: string) {
+        copyToClipboard.copy(this.textArray[arrayEntry], msg, this.htmlElement)
```
ml054: missing ;

src/Raven.Studio/wwwroot/Content/css/bootstrap/variables-body.less
```diff
@@ -815,3 +815,6 @@
 @database-settings-dirty-item-color: @color-2-2;
 @database-settings-invalid-item-color: @color-1;
 @database-origin-key-color: @color-2-2;
+
+@copy-button-color: @gray-light;
```
ml054, line 819: why new definions? Can't we use existing (for example we have copy on document edit page) - it those copy styles are scoped, then we should move them to global scope and reuse.

Danielle9897: done - refactored so that all places use the same global definition

src/Raven.Studio/typescript/common/copyToClipboard.ts
Danielle9897: done

src/Raven.Studio/typescript/common/copyToClipboard.ts
Danielle9897: done - refactored.

ml054 (approved):

PR #11316, opened 2020-10-27
RavenDB-15669 properly handle hash if no TS entries found

karmeli87:


ppekrol:
Does this mirror the behavior of counters and documents?

karmeli87:
well..
image

~~also we don't throw at the endpoint if the page size is zero..~~

PR #11326, opened 2020-10-29
RavenDB-15732 : Time series group by date start

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-15732

### DIFF
--- a/src/Raven.Server/Documents/Queries/AST/Query.cs
+++ b/src/Raven.Server/Documents/Queries/AST/Query.cs
@@ -228,8 +228,8 @@ namespace Raven.Server.Documents.Queries.AST
 
             if (ticks > _start.Ticks)
             {
-                double distance = ticks - _start.Ticks;
-                var ticksToAdd = (long)Math.Floor(distance / Ticks) * Ticks;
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Documents/Queries/AST/Query.cs
```diff
@@ -221,12 +221,21 @@ public void InitializeFullRange(DateTime start, DateTime end)
...
         {
-            var ticks = start.Ticks;
+            var ticks = first.Ticks;
             ticks -= (ticks % AlignedTicks);
-            _start = new DateTime(ticks, start.Kind);
+
+            if (ticks > _start.Ticks)
+            {
+                double distance = ticks - _start.Ticks;
+                var ticksToAdd = (long)Math.Floor(distance / Ticks) * Ticks;
```
karmeli87: This should yield same result without type conversions, right?
`var distance = ticks - _start.Ticks; ` // keep it long
`var ticksToAdd = ((distance / Ticks) * Ticks)`

karmeli87 (approved):

PR #11327, opened 2020-10-29
RavenDB-15297 Edit certificate without the need to send the actual certificate through the wire

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15297

### DIFF
--- a/src/Raven.Client/ServerWide/Operations/Certificates/CertificateDefinition.cs
+++ b/src/Raven.Client/ServerWide/Operations/Certificates/CertificateDefinition.cs
@@ -4,55 +4,55 @@ using Sparrow.Json.Parsing;
 
 namespace Raven.Client.ServerWide.Operations.Certificates
 {
-public class CertificateDefinition : CertificateMetadata
-{
-    public string Certificate;
-    public string Password;
-
-    public DynamicJsonValue ToJson(bool metadataOnly = false)
+    public class CertificateDefinition : CertificateMetadata
     {
-        var jsonValue = base.ToJson();
-        if (metadataOnly == false)
+        public string Certificate;
+        public string Password;
+
+        public DynamicJsonValue ToJson(bool metadataOnly = false)
         {
-            jsonValue[nameof(Certificate)] = Certificate;
+            var jsonValue = base.ToJson();
+            if (metadataOnly == false)
+            {
+                jsonValue[nameof(Certificate)] = Certificate;
+            }
+            return jsonValue;
         }
-        return jsonValue;
     }
-}
 
-public class CertificateMetadata
-{
-    public string Name;
-    public SecurityClearance SecurityClearance;
-    public string Thumbprint;
-    public DateTime? NotAfter;
-    public Dictionary<string, DatabaseAccess> Permissions = new Dictionary<string, DatabaseAccess>(StringComparer.OrdinalIgnoreCase);
-    public List<string> CollectionSecondaryKeys = new List<string>();
-    public string CollectionPrimaryKey = string.Empty;
-    public string PublicKeyPinningHash;
-
-    public DynamicJsonValue ToJson()
+    public class CertificateMetadata
     {
-        var permissions = new DynamicJsonValue();
-        
-        if (Permissions != null)
-            foreach (var kvp in Permissions)
-                permissions[kvp.Key] = kvp.Value.ToString();
+        public string Name;
+        public SecurityClearance SecurityClearance;
+        public string Thumbprint;
+        public DateTime? NotAfter;
+        public Dictionary<string, DatabaseAccess> Permissions = new Dictionary<string, DatabaseAccess>(StringComparer.OrdinalIgnoreCase);
+        public List<string> CollectionSecondaryKeys = new List<string>();
+        public string CollectionPrimaryKey = string.Empty;
+        public string PublicKeyPinningHash;
 
-        var jsonValue = new DynamicJsonValue
+        public DynamicJsonValue ToJson()
         {
-            [nameof(Name)] = Name,
-            [nameof(Thumbprint)] = Thumbprint,
-            [nameof(SecurityClearance)] = SecurityClearance,
-            [nameof(Permissions)] = permissions,
-            [nameof(NotAfter)] = NotAfter,
-            [nameof(CollectionSecondaryKeys)] = CollectionSecondaryKeys,
-            [nameof(CollectionPrimaryKey)] = CollectionPrimaryKey,
-            [nameof(PublicKeyPinningHash)] = PublicKeyPinningHash
-        };
-        return jsonValue;
+            var permissions = new DynamicJsonValue();
+        
+            if (Permissions != null)
+                foreach (var kvp in Permissions)
+                    permissions[kvp.Key] = kvp.Value.ToString();
+
+            var jsonValue = new DynamicJsonValue
+            {
+                [nameof(Name)] = Name,
+                [nameof(Thumbprint)] = Thumbprint,
+                [nameof(SecurityClearance)] = SecurityClearance,
+                [nameof(Permissions)] = permissions,
+                [nameof(NotAfter)] = NotAfter,
+                [nameof(CollectionSecondaryKeys)] = CollectionSecondaryKeys,
+                [nameof(CollectionPrimaryKey)] = CollectionPrimaryKey,
+                [nameof(PublicKeyPinningHash)] = PublicKeyPinningHash
+            };
+            return jsonValue;
+        }
     }
-}
 
     public enum DatabaseAccess
     {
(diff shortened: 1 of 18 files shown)


ppekrol:
It is causing a lot of test failures. Can you check those?

src/Raven.Client/ServerWide/Operations/Certificates/CertificateDefinition.cs
```diff
@@ -4,18 +4,35 @@
...
 {
-    public class CertificateDefinition
+    public class CertificateDefinition : CertificateMetadata
     {
-        public string Name;
         public string Certificate;
         public string Password;
+        public string PublicKeyPinningHash;
+        public string CollectionPrimaryKey = string.Empty;
+        public List<string> CollectionSecondaryKeys = new List<string>();
```
ayende: What is that?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -2236,8 +2236,31 @@ public IEnumerable<string> GetCertificateThumbprintsFromLocalState(TransactionOp
                 } while (it.MoveNext());
             }
         }
+        
+        public static IEnumerable<(string Key, BlittableJsonReaderObject Cert)> GetCertificateFromLocalState(TransactionOperationContext context)
```
ayende:
```suggestion
        public static IEnumerable<(string Key, BlittableJsonReaderObject Cert)> GetAllCertificatesFromLocalState(TransactionOperationContext context)
```

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
```diff
@@ -240,9 +240,8 @@ public async Task Put()
 
                 try
                 {
-                    var _ = string.IsNullOrEmpty(certificate.Password)
-                        ? new X509Certificate2(certBytes, (string)null, X509KeyStorageFlags.MachineKeySet)
-                        : new X509Certificate2(certBytes, certificate.Password, X509KeyStorageFlags.MachineKeySet);
+                    var password = string.IsNullOrEmpty(certificate.Password) ? null : certificate.Password;
+                    _ = new X509Certificate2(certBytes, password, X509KeyStorageFlags.MachineKeySet);
```
ayende: We have to dispose this value.

src/Raven.Client/ServerWide/Operations/Certificates/CertificateDefinition.cs
haludi: Do you ask about `CollectionSecondaryKeys`?
As I understand it if the user supplies PFX that contains more than one certificate we save each one separately but we present all of them as secondary except to the first.

But your question makes me think about how the decision here should be.
The idea here is to give the user sufficient information to call for the edit but the original problem was to prevent sending sensitive data through the wire so maybe it would make more sense to divide it by sensitivity?
So if we want to keep `CertificateDefinition` unchanged to not break anything:
```
public class CertificateDefinition : InsensitiveCertificateDefinition
{
    public string Certificate;
    public string Password;

    public DynamicJsonValue ToJson(bool metadataOnly = false)
    {
        var jsonValue = base.ToJson();
        if (metadataOnly == false)
        {
            jsonValue[nameof(Certificate)] = Certificate;
        }
        return jsonValue;
    }
}

public class InsensitiveCertificateDefinition
{
    public string Name;
    public SecurityClearance SecurityClearance;
    public string Thumbprint;
    public DateTime? NotAfter;
    public Dictionary<string, DatabaseAccess> Permissions = new Dictionary<string, DatabaseAccess>(StringComparer.OrdinalIgnoreCase);
    public List<string> CollectionSecondaryKeys = new List<string>();
    public string CollectionPrimaryKey = string.Empty;
    public string PublicKeyPinningHash;

    public DynamicJsonValue ToJson()
    {
        var permissions = new DynamicJsonValue();
        
        if (Permissions != null)
            foreach (var kvp in Permissions)
                permissions[kvp.Key] = kvp.Value.ToString();

        var jsonValue = new DynamicJsonValue
        {
            [nameof(Name)] = Name,
            [nameof(Thumbprint)] = Thumbprint,
            [nameof(SecurityClearance)] = SecurityClearance,
            [nameof(Permissions)] = permissions,
            [nameof(NotAfter)] = NotAfter,
            [nameof(CollectionSecondaryKeys)] = CollectionSecondaryKeys,
            [nameof(CollectionPrimaryKey)] = CollectionPrimaryKey,
            [nameof(PublicKeyPinningHash)] = PublicKeyPinningHash;
        };
        return jsonValue;
    }
}
```

ayende: I think that there is some issues with the reason for this feature.
This is _not_ about security. We assume the following:

* The connection is encrypted between us and the server.
* The certificate we keep is the public info only.

This is mostly so we'll not need to download and send back the certificate on each time. Users want to be able to provide a way to edit permissions on a user using just the thumbprint of the certificate.

ayende (changes requested):
Basically, what we mostly need is just to be able to edit the permissions without sending the full cert over.

src/Raven.Client/ServerWide/Operations/Certificates/CertificateDefinition.cs
haludi: But the user still needs to get the old configuration so he could edit it and resend to the `edit` endpoint.
So it is all about the size of the request/response here?

ayende: I think that this is mostly about the ease of use.
What if we change things up.
Add an `EditPermissions` operation that would allow simply replacing the permissions by thumbprint?

ayende: Without updating anything else?

ppekrol: I think that this operation is already lightweight enough. Why making it more complex, adding an endpoint or any other additional logic?

```
string thumbprint, string name, Dictionary<string, DatabaseAccess> permissions, SecurityClearance clearance
```

src/Raven.Client/ServerWide/Operations/Certificates/CertificateDefinition.cs
```diff
@@ -4,41 +4,55 @@
 
 namespace Raven.Client.ServerWide.Operations.Certificates
 {
-    public class CertificateDefinition
+public class CertificateDefinition : CertificateMetadata
```
ppekrol: formatting

src/Raven.Client/ServerWide/Operations/Certificates/CertificateDefinition.cs
```diff
@@ -4,41 +4,55 @@
...
-                [nameof(CollectionPrimaryKey)] = CollectionPrimaryKey,
-                [nameof(CollectionSecondaryKeys)] = CollectionSecondaryKeys
-            };
+            jsonValue[nameof(Certificate)] = Certificate;
         }
+        return jsonValue;
     }
+}
+
+public class CertificateMetadata
```
ppekrol: formatting

src/Raven.Client/ServerWide/Operations/Certificates/EditClientCertificateOperation.cs
```diff
@@ -0,0 +1,73 @@
...
+namespace Raven.Client.ServerWide.Operations.Certificates
+{
+    public class EditClientCertificateOperation : IServerOperation
+    {
+        private readonly string _thumbprint;
+        private readonly Dictionary<string, DatabaseAccess> _permissions;
+        private readonly string _name;
+        private readonly SecurityClearance _clearance;
+
+        public EditClientCertificateOperation(string thumbprint, string name, Dictionary<string, DatabaseAccess> permissions, SecurityClearance clearance)
```
ppekrol: introduce nested `Parameters` class (we do not that in other places) and use those. Easier to extend in the future.

src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
```diff
@@ -240,9 +240,9 @@ public async Task Put()
 
                 try
                 {
-                    var _ = string.IsNullOrEmpty(certificate.Password)
-                        ? new X509Certificate2(certBytes, (string)null, X509KeyStorageFlags.MachineKeySet)
-                        : new X509Certificate2(certBytes, certificate.Password, X509KeyStorageFlags.MachineKeySet);
+                    var password = string.IsNullOrEmpty(certificate.Password) ? null : certificate.Password;
+                    var certificate2 = new X509Certificate2(certBytes, password, X509KeyStorageFlags.MachineKeySet);
```
ppekrol: what is wrong with using `using` ?

src/Raven.Client/ServerWide/Operations/Certificates/CertificateDefinition.cs
haludi: So if it is not about sensitive data maybe I will remove the `metadataOnly` if it is not necessary (include the division to CertificateDefinition & CertificateMetadata)?

ppekrol: No

PR #11332, opened 2020-11-01
RavenDB-15784: Rename attachment of conflicted document if it has dup…

garayx:
…licate name

### DIFF
--- a/src/Raven.Server/Documents/AttachmentsStorage.cs
+++ b/src/Raven.Server/Documents/AttachmentsStorage.cs
@@ -544,6 +544,27 @@ namespace Raven.Server.Documents
             return attachments;
         }
 
+        public List<AttachmentDetails> GetAttachmentDetailsForDocument(DocumentsOperationContext context, Slice lowerDocumentId)
+        {
+            var attachments = new List<AttachmentDetails>();
+            using (GetAttachmentPrefix(context, lowerDocumentId, AttachmentType.Document, Slices.Empty, out Slice prefixSlice))
+            {
+                foreach (var attachment in GetAttachmentsForDocument(context, prefixSlice))
+                {
+                    attachments.Add(new AttachmentDetails
+                    {
+                        Name = attachment.Name,
+                        Hash = attachment.Base64Hash.ToString(),
+                        ContentType = attachment.ContentType,
+                        Size = attachment.Size,
+                        ChangeVector = attachment.ChangeVector,
+                        DocumentId = lowerDocumentId.ToString()
+                    });
+                }
+            }
+            return attachments;
+        }
+
         public (long AttachmentCount, long StreamsCount) GetNumberOfAttachments(DocumentsOperationContext context)
         {
             // We count in also revision attachments
@@ -957,6 +978,32 @@ namespace Raven.Server.Documents
                 Hash = hash,
                 Size = attachment.Size
             };
+        }
+
+        public string ResolveAttachmentName(DocumentsOperationContext context, Slice lowerId, string name, string hash, string contentType)
+        {
+            const string prefix = "RESOLVED";
+            var count = 0;
+            string newName = $"{prefix}_#{count}_{name}";
+
+            var table = context.Transaction.InnerTransaction.OpenTable(AttachmentsSchema, AttachmentsMetadataSlice);
+            using (DocumentIdWorker.GetLowerIdSliceAndStorageKey(context, contentType, out Slice lowerContentType, out Slice contentTypePtr))
+            using (Slice.From(context.Allocator, hash, out Slice base64Hash))
+            {
+                while (true)
+                {
+                    using (DocumentIdWorker.GetLowerIdSliceAndStorageKey(context, newName, out Slice lowerName, out Slice namePtr))
+                    using (GetAttachmentPartialKey(context, lowerId.Content.Ptr, lowerId.Size, lowerName.Content.Ptr, lowerName.Size, AttachmentType.Document, changeVector: null, out Slice partialKeySlice))
+                    {
+                        if (table.SeekOnePrimaryKeyPrefix(partialKeySlice, out _) == false)
+                            break;
+
+                        newName = $"{prefix}_#{++count}_{name}";
+                    }
+                }
+            }
+
+            return newName;
         }
 
         public void DeleteAttachment(DocumentsOperationContext context, string documentId, string name, LazyStringValue expectedChangeVector, bool updateDocument = true)
--- a/src/Raven.Server/Documents/AttachmentsStorage.cs
+++ b/src/Raven.Server/Documents/AttachmentsStorage.cs
@@ -184,7 +184,7 @@ namespace Raven.Server.Documents
                 using (GetAttachmentKey(context, lowerDocumentId.Content.Ptr, lowerDocumentId.Size, lowerName.Content.Ptr, lowerName.Size, base64Hash,
                     lowerContentType.Content.Ptr, lowerContentType.Size, AttachmentType.Document, Slices.Empty, out Slice keySlice))
                 {
-                    Debug.Assert(base64Hash.Size == 44, $"Hash size should be 44 but was: {keySlice.Size}");
+                    Debug.Assert(base64Hash.Size == 44, $"Hash size should be 44 but was: {base64Hash.Size}");
 
                     DeleteTombstoneIfNeeded(context, keySlice);
 
@@ -299,7 +299,7 @@ namespace Raven.Server.Documents
         /// </summary>
         public void PutDirect(DocumentsOperationContext context, Slice key, Slice name, Slice contentType, Slice base64Hash, string changeVector = null)
         {
-            Debug.Assert(base64Hash.Size == 44, $"Hash size should be 44 but was: {key.Size}");
+            Debug.Assert(base64Hash.Size == 44, $"Hash size should be 44 but was: {base64Hash.Size}");
 
             var newEtag = _documentsStorage.GenerateNextEtag();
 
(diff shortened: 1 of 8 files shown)


src/Raven.Server/Documents/DocumentPutAction.cs
```diff
@@ -422,6 +424,33 @@ private static void ThrowInvalidDocumentId(string id)
...
+                    if (metadata.TryGet(Constants.Documents.Metadata.Attachments, out BlittableJsonReaderArray currentDocumentAttachments))
+                    {
+                        // actualAttachments can contain attachments of current document which were directly put, we want to check if there are attachment with duplicate name
+                        var attachmentsChanged = false;
+                        foreach (BlittableJsonReaderObject bjro in currentDocumentAttachments)
+                        {
+                            var currentAttachment = JsonDeserializationClient.AttachmentName(bjro);
+
+                            if (actualAttachments.Items.Count(o =>
+                                o is DynamicJsonValue attachment && attachment[nameof(AttachmentName.Name)] is LazyStringValue lzv && lzv == currentAttachment.Name) > 1)
```
ayende: The types here are _very_ exact. Can it be other things? 
This looks very fragile.

src/Raven.Server/Documents/AttachmentsStorage.cs
```diff
@@ -901,6 +901,64 @@ public AttachmentDetails MoveAttachment(DocumentsOperationContext context, strin
             return result;
         }
 
+        public AttachmentDetails RenameAttachment(DocumentsOperationContext context, Slice lowerId, string name, string hash, string contentType, AttachmentType attachmentType, string newName)
```
ayende, line 925: Why is this required? The tag on the attachment is for debug only.
We don't actually need to rename it.

src/Raven.Server/Documents/DocumentPutAction.cs
```diff
@@ -422,6 +424,33 @@ private static void ThrowInvalidDocumentId(string id)
...
+                            {
+                                // rename currentAttachment which is already saved in storage
+                                _documentsStorage.AttachmentsStorage.RenameAttachment(
```
ayende: What if we have three duplicates?

ayende (changes requested):

src/Raven.Server/Documents/DocumentPutAction.cs
garayx: this is returned from the attachments storage:
https://github.com/ravendb/ravendb/blob/ff622ad5b817a4245e8011e0a5779599236f1511/src/Raven.Server/Documents/AttachmentsStorage.cs#L146
and the values are from table

src/Raven.Server/Documents/AttachmentsStorage.cs
garayx: yes the unique value is key, but we don't allow attachments to have same name when it is added from studio or Client API. Conflicts in replication can cause such scenario, which in place can break other things like patch.

ayende: What happens if we have _another_ conflict? Then we'll have a duplicate on `resolved_name`, no?

src/Raven.Server/Documents/DocumentPutAction.cs
ayende: That is fragile. If we'll change this in the future, there is no indication that this is related.

src/Raven.Server/Documents/DocumentPutAction.cs
garayx: The only chance for this is when the doc already has attachments with duplicate names (got conflicted before this PR merged) and gets into conflict again after this PR merged. For normal doc, after this PR there is always going to be max 2 duplicates.

We can treat this issue by handling here the > 2 duplicates case or do schema update to rename attachments with duplicate  name

src/Raven.Server/Documents/DocumentPutAction.cs
garayx: done

src/Raven.Server/Documents/AttachmentsStorage.cs
garayx: done

ayende (approved):

src/Raven.Server/Documents/Replication/ConflictManager.cs
```diff
@@ -85,9 +86,16 @@ public ConflictManager(DocumentDatabase database, ResolveConflictOnReplicationCo
...
-
-                    return;
+                    try
+                    {
+                        _conflictResolver.PutResolvedDocument(documentsContext, resolved, conflictedDoc, localDocumentTuple.Document);
+                        return;
+                    }
+                    catch (AttachmentConflictException e)
+                    {
+                        if (_log.IsInfoEnabled)
```
ayende, line 96: Under what scenario do we want this behavior? We should always resolve things automatically

src/Raven.Server/Documents/Replication/ConflictManager.cs
```diff
@@ -137,14 +146,26 @@ public static void AssertChangeVectorNotNull(string conflictedChangeVector)
                 InvalidConflictWhenThereIsNone(conflict.Id);
 
             conflictedDocs.Add(conflict.Clone());
+            conflictedDocs = conflictedDocs.OrderBy(x => x.Doc?.GetHashCode()).ToList();
```
ayende, line 149: This is effectively random, why do you do this?
Also, it isn't guaranteed to be the same on different nodes

src/Raven.Server/Documents/Replication/ConflictManager.cs
```diff
@@ -137,14 +146,26 @@ public static void AssertChangeVectorNotNull(string conflictedChangeVector)
...
                 scriptResolver,
                 conflictedDocs,
                 documentsContext.GetLazyString(collection), out var resolved))
             {
-                _conflictResolver.PutResolvedDocument(documentsContext, resolved, conflict);
+                try
+                {
+                    _conflictResolver.PutResolvedDocument(documentsContext, resolved, conflict, relevantLocalDoc.Document);
+                }
+                catch (AttachmentConflictException e)
```
ayende, line 161: I don't think we want this

src/Raven.Server/Documents/Replication/ResolveConflictOnReplicationConfigurationChange.cs
```diff
@@ -400,6 +412,125 @@ public DocumentConflict ResolveToLatest(List<DocumentConflict> conflicts)
...
+                        }
+
+                        foreach (var match in matches)
+                        {
+                            // find which name of the matched attachments doesn't exist in the current metadata
+                            if (CountInAttachmentsList(match.Name, currentDocumentAttachments) == 0)
+                            {
+                                if (_database.DocumentsStorage.AttachmentsStorage.RenameAttachment(context, lowerId, match.Name, match.Hash, match.ContentType, AttachmentType.Document, currentAttachment.Name) == null)
+                                    throw new AttachmentConflictException($"Could not rename attachment '{match.Name}' to '{currentAttachment.Name}' in attachments storage of document '{lowerId}', attachment has the following: Hash: '{currentAttachment.Hash}', ContentType: '{currentAttachment.ContentType}'.", lowerId.ToString());
```
ayende, line 453: Can use `_1`, `_2`, etc instead.
Can use random value, etc.

We can't throw here.

src/Raven.Server/Documents/Replication/ResolveConflictOnReplicationConfigurationChange.cs
```diff
@@ -400,6 +412,125 @@ public DocumentConflict ResolveToLatest(List<DocumentConflict> conflicts)
...
+                        RemoveFromAttachmentsToRemoveListIfNeeded(currentAttachment.Name);
+                        continue;
+                    case 0:
+                        // attachment exists in metadata, does not exist in storage (renamed/added in conflict resolution)
+                        // try to find attachments in attachment storage with both same hash and content type
+                        var matches = storageAttachmentsDetails.Where(x => x.Hash == currentAttachment.Hash && x.ContentType == currentAttachment.ContentType).ToList();
+                        if (matches.Count == 0)
+                        {
+                            // attachment is in the metadata but doesn't exist in the attachments storage
+                            throw new AttachmentConflictException($"Attachment '{currentAttachment.Name}' was found in '{nameof(DocumentFlags.Resolved)}' document '{lowerId}' metadata, but was not found in the attachments storage using the following: Hash: '{currentAttachment.Hash}', ContentType: '{currentAttachment.ContentType}'.", lowerId.ToString());
```
ayende, line 444: Shouldn't this be a missing attachment error?

src/Raven.Server/Documents/Replication/ResolveConflictOnReplicationConfigurationChange.cs
```diff
@@ -400,6 +412,125 @@ public DocumentConflict ResolveToLatest(List<DocumentConflict> conflicts)
...
+                                
+                                RemoveFromAttachmentsToRemoveListIfNeeded(match.Name);
+                                break;
+                            }
+                        }
+
+                        break;
+                    case 2:
+                        // there are attachments with duplicate names in storage
+                        var duplicates = storageAttachmentsDetails.Where(x => x.Name == currentAttachment.Name && (x.Hash != currentAttachment.Hash || x.ContentType != currentAttachment.ContentType)).ToList();
```
ayende, line 463: Why do you care about the attachment name? Only the hash matters to us

ayende (changes requested):
Talk to me about this, please

ppekrol:
This will be handled differently. @karmeli87 and @garayx spoke.

src/Raven.Server/Documents/Replication/ConflictManager.cs
karmeli87: Right... we need some deterministic way to order them.
Maybe rely on the CV for that? (But I guess we need to order it internally as well)

ayende: Ordering by CV will be a good idea, yes.

src/Raven.Server/Documents/Replication/ResolveConflictOnReplicationConfigurationChange.cs
garayx: This is to prevent from adding an attachment in conflict resolver script
https://github.com/ravendb/ravendb/pull/11332/files#diff-0274c68798c060bc7f04679d2b76c93d9d1f92404872601fcf7d4e7f547b224fR1951

src/Raven.Server/Documents/Replication/ResolveConflictOnReplicationConfigurationChange.cs
garayx: The duplicate names was the original issue we had to solve, because duplicate names can break patch and should be disallowed

src/Raven.Server/Documents/Replication/ResolveConflictOnReplicationConfigurationChange.cs
garayx: `RenameAttachment` should not fail, its a precaution so it can be resolved manually if something goes wrong and we cannot add 1,2, random value, because this is result of script of the `ScriptResolver`

src/Raven.Server/Documents/Replication/ConflictManager.cs
garayx: this is to mark the doc as conflicted

garayx:
https://github.com/ravendb/ravendb/pull/11434

PR #11333, opened 2020-11-01
RavenDB-15797 The client test should not wait for client configuration to be changed while it is not changed in the server

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15797

### DIFF
--- a/test/SlowTests/Issues/RavenDB-14978.cs
+++ b/test/SlowTests/Issues/RavenDB-14978.cs
@@ -4,6 +4,7 @@ using System.Threading.Tasks;
 using Nito.AsyncEx;
 using Raven.Client.Documents;
 using Raven.Client.Documents.Conventions;
+using Raven.Client.Documents.Operations.Configuration;
 using Raven.Client.Http;
 using Tests.Infrastructure;
 using Xunit;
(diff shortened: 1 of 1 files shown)


test/SlowTests/Issues/RavenDB-14978.cs
```diff
@@ -38,21 +38,47 @@ public async Task Can_setup_write_load_balancing()
...
+            };
 
             var (index, _) = await CreateDatabaseInCluster(databaseName, 3, leader.WebUrl);
             await WaitForRaftIndexToBeAppliedInCluster(index, TimeSpan.FromSeconds(30));
 
-            using (var s0 = store.OpenSession())
-            {
-                s0.Load<User>("test/1");
-            }
+            await store.Maintenance.SendAsync(new PutClientConfigurationOperation(configuration), cts.Token);
```
ayende, line 75: where are you wait to be notified about this?

haludi: I am testing the behavior of `LoadBalanceBehavior.UseSessionContext`.
And there is no need to wait while in the new implementation the server sends in the header of `PutClientConfigurationOperation` request the need to update so the client does it immediately after this request
https://github.com/ravendb/ravendb/blob/1936579b7a63b726e036eac4de2558e99ce0141e/src/Raven.Server/Documents/Handlers/Admin/AdminConfigurationHandler.cs#L143

ayende: The update is done in the background, no? What happens if this takes too long?
That would make this a racy test, no?

haludi: We do await for both the client configuration update task and the topology update task
https://github.com/ravendb/ravendb/blob/3d7a25151b978e717edd9f3fab2e965b085b4a87/src/Raven.Client/Http/RequestExecutor.cs#L897

ayende: Okay, then LGTM

PR #11334, opened 2020-11-02
RavenDB-15144 Allow to define pull replication sink in the studio wit…

Danielle9897:
…hout certificate


src/Raven.Studio/typescript/viewmodels/database/tasks/editReplicationSinkTask.ts
```diff
@@ -42,11 +44,37 @@ class editReplicationSinkTask extends viewModelBase {
...
+    canActivate(args: any) {
+        return $.when<any>(super.canActivate(args))
+            .then(() => {
+                const deferred = $.Deferred<canActivateResultDto>();
+
+                if (this.canDefineCertificates) {
+                    new getCertificatesCommand()
+                        .execute()
+                        .done(certificatesInfo => {
+                            const serverCertificate = certificatesInfo.Certificates.find(cert => cert.Name === "Server Certificate");
```
ml054: maybe we should use some const here ?

ml054 (approved):

src/Raven.Studio/typescript/viewmodels/database/tasks/editReplicationSinkTask.ts
Danielle9897: done

PR #11340, opened 2020-11-03
RavenDB-15786 Increasing the number of  MinThreads on Thread Pool (de…

arekpalinski:
…fault value is set to the number of processors on a system) on server startup to alleviate potential thread starvation problem when serving many requests.

### DIFF
--- a/src/Raven.Server/Config/Categories/ServerConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/ServerConfiguration.cs
@@ -1,11 +1,18 @@
 using System.ComponentModel;
 using Raven.Server.Config.Attributes;
 using Raven.Server.Config.Settings;
+using Sparrow.Platform;
+using Sparrow.Utils;
 
 namespace Raven.Server.Config.Categories
 {
     public class ServerConfiguration : ConfigurationCategory
     {
+        public ServerConfiguration()
+        {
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Config/Categories/ServerConfiguration.cs
```diff
@@ -1,11 +1,18 @@
...
+using Sparrow.Platform;
+using Sparrow.Utils;
 
 namespace Raven.Server.Config.Categories
 {
     public class ServerConfiguration : ConfigurationCategory
     {
+        public ServerConfiguration()
+        {
+            ThreadPoolMinThreads = (PlatformDetails.Is32Bits ? 2 : 4) * ProcessorInfo.ProcessorCount;
```
ayende, line 13: IIRC, that would actually reduce the number of small machines, no?
The default is ~25 ?

arekpalinski: No. The default is the number of processors on a system. 

https://docs.microsoft.com/en-us/dotnet/api/system.threading.threadpool.setminthreads?view=netcore-3.1

arekpalinski:
Note: `Server.ThreadPool.MinThreads` should be set to `null` on `v5.0` once https://github.com/ravendb/ravendb/pull/11321 gets merged.

PR #11347, opened 2020-11-03
RavenDB-15741 use EnumerateAndClear when cleaning EncryptionBufferPool on low memory, if not we will only clear allocations from current core

ppekrol:


ppekrol:
This fixes: `FastTests.Voron.EncryptionBufferPool.dont_pool_buffers_larger_than_8Mb`

We clear the json context pool same way:
https://github.com/ravendb/ravendb/blob/399342c91116fff469a72a71dac1f5e17c4f90e9/src/Sparrow/Json/JsonContextPoolBase.cs#L330-L334

ayende (approved):

PR #11349, opened 2020-11-04
RavenDB-15794 Circle spatial field not created for javascript index

ayende:


ayende:
Need this on 4.2, not 5.0

PR #11362, opened 2020-11-05
RavenDB-7070 updated to .NET 5.0 RTM

ppekrol:


ppekrol:
1. Had to disable Trimming (investigating)
2. Test Single File App on Linux

PR #11364, opened 2020-11-06
RavenDB-15822 added endpoints to monitor active TCP connections

ppekrol:


src/Raven.Server/Web/System/AdminTcpConnectionDebugInfoHandler.cs
```diff
@@ -0,0 +1,123 @@
...
+                context.Write(writer, djv);
+            }
+
+            return Task.CompletedTask;
+
+            static DynamicJsonValue ToDynamic(TcpStatistics stats)
+            {
+                var result = new DynamicJsonValue();
+                if (stats == null)
+                    return result;
```
ayende: shouldn't we return null here? Or a message that the platform is not supported?
On what platform isn't this supported?

src/Raven.Server/Web/System/AdminTcpConnectionDebugInfoHandler.cs
```diff
@@ -0,0 +1,123 @@
...
+                    return null;
+                }
+            }
+        }
+
+        [RavenAction("/admin/debug/info/tcp/active-connections", "GET", AuthorizationStatus.Operator, IsDebugInformationEndpoint = true)]
+        public Task ActiveConnections()
+        {
+            var properties = IPGlobalProperties.GetIPGlobalProperties();
+            var connections = properties.GetActiveTcpConnections();
```
ayende: This can throw, no?

ayende (changes requested):

src/Raven.Server/Web/System/AdminTcpConnectionDebugInfoHandler.cs
ppekrol: Done. Not sure which OS can throw this. Docs state it can throw that ex: https://docs.microsoft.com/en-us/dotnet/api/system.net.networkinformation.ipglobalproperties.gettcpipv6statistics?view=netcore-3.1

Win7 maybe?

src/Raven.Server/Web/System/AdminTcpConnectionDebugInfoHandler.cs
ppekrol: No info in docs, but I got ex on a property where there was also no info, so I've added helper method to retrieve this safely.

src/Sparrow.Server/Extensions/TcpExtensions.cs
```diff
@@ -0,0 +1,275 @@
...
+            try
+            {
+                if (properties == null)
+                    return null;
+
+                return properties.GetActiveTcpConnections();
+            }
+            catch (PlatformNotSupportedException)
+            {
+                return null;
```
ayende, line 31: Should we return empty array here?

src/Raven.Server/Web/System/AdminTcpConnectionDebugInfoHandler.cs
```diff
@@ -0,0 +1,112 @@
...
+                result[nameof(stats.SegmentsSent)] = stats.GetSegmentsSentSafely();
+
+                return result;
+            }
+        }
+
+        [RavenAction("/admin/debug/info/tcp/active-connections", "GET", AuthorizationStatus.Operator, IsDebugInformationEndpoint = true)]
+        public Task ActiveConnections()
+        {
+            var properties = TcpExtensions.GetIPGlobalPropertiesSafely();
```
ayende, line 63: This can return null

src/Raven.Server/Monitoring/Snmp/Objects/Server/1.15/TcpActiveConnections.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class TcpActiveConnections : ScalarObjectBase<Gauge32>
+    {
+        public TcpActiveConnections()
+            : base(SnmpOids.Server.TcpActiveConnections)
+        {
+        }
+
+        protected override Gauge32 GetData()
+        {
+            var properties = TcpExtensions.GetIPGlobalPropertiesSafely();
```
ayende, line 15: Can return null

src/Sparrow.Server/Extensions/TcpExtensions.cs
```diff
@@ -0,0 +1,275 @@
...
+    {
+        public static IPGlobalProperties GetIPGlobalPropertiesSafely()
+        {
+            try
+            {
+                return IPGlobalProperties.GetIPGlobalProperties();
+            }
+            catch
+            {
+                return null;
```
ayende, line 16: Should we return an empty object, instead?

src/Raven.Server/Monitoring/Snmp/Objects/Server/1.15/TcpActiveConnections.cs
ppekrol: Every extension method that I've added checks that.

src/Raven.Server/Web/System/AdminTcpConnectionDebugInfoHandler.cs
ppekrol: Every extension method that I've added checks that.

src/Sparrow.Server/Extensions/TcpExtensions.cs
ppekrol: No, I'm taking that into account in other extension methods. Also all of the implementations are internal.

src/Sparrow.Server/Extensions/TcpExtensions.cs
ppekrol: I do not think so.

PR #11366, opened 2020-11-07
RavenDB-15824 Optimize client side JSON serialization

ayende:
- Avoid allocating arrays when getting blittable properties, use native memory instead
- Reduce the cost of CanConvert calls in JsonConvert
- When detecting changes in blittable, use the order of properties in the document, not lexical properties. User can modify that if needed more easily


ppekrol (approved):
LGTM, but from failing tests it looks like we need to have 4 converters. 2 for serialization and 2 for deserialization?

arekpalinski (approved):

ppekrol:
@ayende can you check failing test?

PR #11370, opened 2020-11-08
RavenDB-15346 Retry CPU credits sync on startup

iftahbe:


src/Raven.Server/RavenServer.cs
```diff
@@ -513,14 +513,37 @@ private void StartMonitoringCpuCredits()
...
+                    {
+                        currentRetry++;
+
+                        // If we weren't able to sync at startup, we'll assume we have enough credits and let RavenDB perform in the meantime.
+                        // We don't want RavenDB to (false) throttle unless it knows for sure that it doesn't have credits.
+                        CpuCreditsBalance.RemainingCpuCredits = CpuCreditsBalance.MaxCredits;
+                        
+                        if (currentRetry > 5)
+                        {
+                            throw;
```
ayende: What will happen at this point?

src/Raven.Server/Web/System/AdminCpuCreditsHandler.cs
```diff
@@ -26,6 +28,29 @@ public Task UpdateCpuCredits()
             }
         }
 
+        [RavenAction("/admin/cpu-credits/sync", "POST", AuthorizationStatus.ClusterAdmin)]
+        public Task SyncCpuCredits()
+        {
+            try
+            {
+                Server.UpdateCpuCreditsFromExec();
```
ayende: This is something that we really want to avoid. We are turning everything to async, so let's make this async and wait in the long running thread, instead.

src/Raven.Server/RavenServer.cs
```diff
@@ -513,14 +513,37 @@ private void StartMonitoringCpuCredits()
...
+                        }
+                    }
+                    Task.Delay(TimeSpan.FromMinutes(1));
```
ayende: Why do we retry at this length?

ayende (changes requested):

src/Raven.Server/RavenServer.cs
iftahbe: Nothing will happen. It will be caught in the catch block, logged, and ignored. Then we'll try in the next time we have to sync (30 min)

src/Raven.Server/RavenServer.cs
iftahbe: I figured we don't get a result because of the cloud provider. We need to wait but for how long? Well, I just thought that one minute is a long enough period to get new results from the provider
.

src/Raven.Server/Web/System/AdminCpuCreditsHandler.cs
iftahbe: Yes of course.

src/Raven.Server/RavenServer.cs
ayende: But at this point, we have 5 minutes that went by with no visible action, right?

src/Raven.Server/RavenServer.cs
ayende: The question is, why do we need to wait at all? We already have a periodic check, no?
Why hold up the thread for so long?

src/Raven.Server/RavenServer.cs
iftahbe: Ok. I merged it with the existing loop.

src/Raven.Server/RavenServer.cs
iftahbe: I merged the logic with the main loop, it will try to sync every minute until we get a valid result.

src/Raven.Server/RavenServer.cs
```diff
@@ -452,6 +460,12 @@ public double RemainingCpuCredits
                 set => Interlocked.Exchange(ref _remainingCpuCredits, value);
             }
 
+            public int ForceSync
+            {
+                get => Interlocked.CompareExchange(ref _forceSync, 0, 0);
```
ayende: Huh? Why not `Interlocked.Read` ?

iftahbe: Because Interlocked.Read doesn't work for int, just long.
This is infact like Interlocked.Read. It performs an atomic read operation with no writing.

iftahbe: Does it matter? Do you want me to change the type to long and use Interlocked.Read?

ayende: It mutates the data (more expensive. Can use `Thread.VolatileRead` instead, I think

ppekrol: Few questions:
- why this is not bool?
- getting/setting bool is atomic
- why do we care here about interlocked/volatile/etc?

iftahbe: Actually, we don't really care.
The reason I did this is to be consistent with the above RemainingCpuCredits variable where it's actually important. So I guess I'll just use a normal bool there.

ppekrol (approved):

PR #11372, opened 2020-11-08
RavenDB-15803 fix race in PoolOfThreads and reduce work for the TaskExecutor

karmeli87:


src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -332,7 +332,7 @@ public void Initialize(InitializeOptions options = InitializeOptions.None, DateT
 
                 _serverStore.StorageSpaceMonitor.Subscribe(this);
 
-                TaskExecutor.Execute( _ =>
+                ThreadPool.QueueUserWorkItem( _ =>
```
ayende, line 335: This is here on purpose. 
The idea is that even if the Thread Pool is backed up, we have a dedicated thread that would still churn through those tasks.
We need to make sure that we aren't doing that twice, mind you.

ayende (changes requested):
TaskExecuter is important when we are running close to the system resources.
But the tasks it executes need to understand that they may run twice, is that the issue here?

arekpalinski:
The 2 commits fixing the race conditions LGTM. Thanks @karmeli87 for explaining that to me.

src/Raven.Server/Documents/DocumentDatabase.cs
karmeli87: The task executor make a competition between the dedicated thread and a TP thread.

The problem that we had, is that the Task Executor run a never ending task, it got stuck, all operations completed by the TP, but the queue of the Task Executor exploded.

karmeli87: So I left only non-blocking operations to be run by the Task Executor

ayende: Okay, yes. `TaskExecuter`  is for very short notifications, yes.

ayende: Like here,, no `NotifyFeaturesAboutStateChange` ?

karmeli87: it is not short :-) 
1. It try to get a lock .. in a loop.
2. It handles every feature change..

ayende (approved):

ppekrol (approved):

PR #11377, opened 2020-11-09
RavenDB-15415 Index progress modified to show items instead of documents to support Counter/TimeSeries index

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15415

### DIFF
--- a/src/Raven.Client/Documents/Indexes/IndexProgress.cs
+++ b/src/Raven.Client/Documents/Indexes/IndexProgress.cs
@@ -20,18 +20,17 @@ namespace Raven.Client.Documents.Indexes
 
         public class CollectionStats
         {
-            public long LastProcessedDocumentEtag { get; set; }
+            public long LastProcessedItemEtag { get; set; }
 
-            public long NumberOfDocumentsToProcess { get; set; }
+            public long NumberOfItemsToProcess { get; set; }
 
-            public long TotalNumberOfDocuments { get; set; }
+            public long TotalNumberOfItems { get; set; }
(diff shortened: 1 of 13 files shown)


src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
```diff
@@ -2014,6 +2014,29 @@ private static void MarkSegmentAsPendingDeletion(DocumentsOperationContext conte
             }
         }
 
+        public long GetNumberOfTimeSeriesSegmentsToProcess(DocumentsOperationContext context, string collection, in long afterEtag, out long totalCount)
```
grisha-kotler, line 2017: Is it segments or the number of time series?

src/Raven.Client/Documents/Indexes/IndexProgress.cs
```diff
@@ -20,18 +20,17 @@ public class IndexProgress
 
         public class CollectionStats
         {
-            public long LastProcessedDocumentEtag { get; set; }
+            public long LastProcessedItemEtag { get; set; }
```
ppekrol, line 23: This breaks backward compatibility. You can mark fields as obsolete, but the value should still be set in them.

src/Raven.Server/Documents/TimeSeries/TimeSeriesStorage.cs
haludi: Segments

PR #11384, opened 2020-11-10
RavenDB-15713 Create SQL writer lazily

karmeli87:


src/Raven.Server/Documents/ETL/EtlProcess.cs
```diff
@@ -659,21 +659,24 @@ public void Run()
...
+                                            Extract(context, merged, nextEtag, EtlItemType.CounterGroup, stats, scope);
                                     
-                                    if (ShouldTrackTimeSeries())
-                                        Extract(context, merged, nextEtag, EtlItemType.TimeSeries, stats, scope);
+                                        if (ShouldTrackTimeSeries())
+                                            Extract(context, merged, nextEtag, EtlItemType.TimeSeries, stats, scope);
 
-                                    var transformations = Transform(merged, context, stats, state);
+                                        transformations = Transform(merged, context, stats, state);
+                                    }
```
arekpalinski: 1. I don't think we can do that. If there is no transformation script we just put a document as is. So we need to have the read transaction open while sending data to a destination:

https://github.com/ravendb/ravendb/blob/18a605ad027854aae791f060265cd21e30d064e7/src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs#L277

2. How this code could cause any regression - https://issues.hibernatingrhinos.com/issue/RavenDB-15713 - since we have it from the beginning?

arekpalinski: Now I understand how you found https://github.com/ravendb/ravendb/pull/11385. Because tests here are causing this. I believe because you're tying to access data after disposing the read transaction

karmeli87: You are correct on both comments :-)
Still investigating..

karmeli87:
The issue was, that prior to 5.1 we skipped the `Load` completely, so the SQL writer wasn't  created. (which can take few seconds!)
Now we don't skip because we are getting `IEnumerable`.

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -507,9 +507,10 @@ protected void DisposeAllFailedNodesTimers()
         public void Execute<TResult>(
             RavenCommand<TResult> command,
             JsonOperationContext context,
-            SessionInfo sessionInfo = null)
+            SessionInfo sessionInfo = null,
+            CancellationToken token = default)
```
ppekrol: Not a good idea for public method in request executor. Use async one or write a helper method in replication loader.

src/Raven.Server/Documents/ETL/Providers/SQL/SqlEtl.cs
```diff
@@ -80,18 +80,23 @@ protected override int LoadInternal(IEnumerable<SqlTableWithRecords> records, Do
...
+                LogStats(stats, table);
 
-                writer.Commit();
+                count += stats.DeletedRecordsCount + stats.InsertedRecordsCount;
+            }
+
+            if (lazyWriter.IsValueCreated)
+            {
+                lazyWriter.Value.Commit();
+                lazyWriter.Value.Dispose();
```
ppekrol: use `using` to dispose. What will happen if commit will throw?

arekpalinski (approved):

PR #11385, opened 2020-11-10
RavenDB-15842 Fix Stack overflow in Voron.Impl.LowLevelTransaction.ThrowObjectDisposed

karmeli87:


ayende (approved):

src/Voron/Impl/LowLevelTransaction.cs
```diff
@@ -582,7 +582,8 @@ private Page GetPageInternal(long pageNumber)
         private void ThrowObjectDisposed()
         {
             if (_disposed.HasFlag(TxState.Disposed))
-                ThrowObjectDisposed();
+                throw new ObjectDisposedException("Transaction is already disposed");
```
arekpalinski, line 585: Wow! Nice catch.

PR #11387, opened 2020-11-10
RavenDB-15714  Disable the Renew License button when user does not ha…

Danielle9897:
…ve sufficient privileges

### DIFF
--- a/src/Raven.Server/Web/Studio/LicenseHandler.cs
+++ b/src/Raven.Server/Web/Studio/LicenseHandler.cs
@@ -1,7 +1,6 @@
 ﻿using System.Net;
 using System.Reflection;
 using System.Threading.Tasks;
-using Microsoft.AspNetCore.Http.Features.Authentication;
 using Raven.Server.Commercial;
 using Raven.Server.Config.Categories;
 using Raven.Server.Json;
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Web/Studio/LicenseHandler.cs
```diff
@@ -58,11 +59,18 @@ public Task GetLicenseConfigurationSettings()
...
+                {
+                    canConfigLicense = false;
+                }
+                
                 var djv = new DynamicJsonValue
                 {
-                    [nameof(LicenseConfiguration.CanRenew)] = ServerStore.Configuration.Licensing.CanRenew,
-                    [nameof(LicenseConfiguration.CanActivate)] = ServerStore.Configuration.Licensing.CanActivate,
-                    [nameof(LicenseConfiguration.CanForceUpdate)] = ServerStore.Configuration.Licensing.CanForceUpdate
+                    [nameof(LicenseConfiguration.CanRenew)] = ServerStore.Configuration.Licensing.CanRenew && canConfigLicense,
```
ppekrol: Wouldn't it be better to get auth status from elsewhere? It would give you more flexibility in the UI e.g.
License can be renewed, but you are not auth as Cluster Admin is a different case than License cannot be renewed at all.

ml054 (approved):
studio part looks good

src/Raven.Server/Web/Studio/LicenseHandler.cs
Danielle9897: done

ml054 (approved):

PR #11400, opened 2020-11-11
RavenDB-15791 Certificates View: Enhancements

Danielle9897:


src/Raven.Studio/typescript/viewmodels/manage/certificates.ts
```diff
@@ -217,8 +235,11 @@ class certificates extends viewModelBase {
     }
     
     enterEditCertificateMode(itemToEdit: unifiedCertificateDefinition) {
-        this.model(certificateModel.fromDto(itemToEdit));
-        this.model().validationGroup.errors.showAllMessages(false);
+        if (!_.includes(itemToEdit.Thumbprints, this.serverCertificateThumbprint())) {
```
ml054: do we need this check here? I think we should just disable button?

src/Raven.Studio/typescript/viewmodels/manage/certificates.ts
```diff
@@ -391,8 +413,17 @@ class certificates extends viewModelBase {
                 });
                 
                 mergedCertificates = _.sortBy(mergedCertificates, x => x.Name.toLocaleLowerCase());
-                this.updateCache(mergedCertificates);
-                this.certificates(mergedCertificates);
+                
+                // place the server certificate and client certificate first in list
+                let serverCert = mergedCertificates.find(x => _.includes(x.Thumbprints, this.serverCertificateThumbprint()));
```
ml054: use const?

ml054 (approved):

src/Raven.Studio/typescript/viewmodels/shell.ts
```diff
@@ -258,25 +259,24 @@ class shell extends viewModelBase {
 
         if (this.clientCertificate() && this.clientCertificate().Name) {
             
-            const dbAccess = certificateModel.resolveDatabasesAccess(this.clientCertificate())
-                .map(x => `<div>${x}</div>`)
-                .join("");
-            
+            const dbAccess = certificateModel.resolveDatabasesAccess(this.clientCertificate());
```
ml054, line 262: update this method to never return null.

src/Raven.Studio/typescript/viewmodels/manage/certificates.ts
Danielle9897: removed

src/Raven.Studio/typescript/viewmodels/manage/certificates.ts
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/shell.ts
Danielle9897: done

ml054 (approved):

PR #11409, opened 2020-11-12
RavenDB-15851 Studio is not sending Additional Assemblies to get inde…

Danielle9897:
…x fields so the compiler fails

### DIFF
--- a/src/Raven.Server/Web/Studio/StudioIndexHandler.cs
+++ b/src/Raven.Server/Web/Studio/StudioIndexHandler.cs
@@ -141,11 +141,15 @@ namespace Raven.Server.Web.Studio
             json.TryGet(nameof(AdditionalAssembly.PackageVersion), out string packageVersion);
             json.TryGet(nameof(AdditionalAssembly.PackageSourceUrl), out string packageSourceUrl);
                 
-            json.TryGet(nameof(AdditionalAssembly.Usings), out BlittableJsonReaderArray usingsArray);
             var usings = new HashSet<string>();
-            foreach (var item in usingsArray)
+            json.TryGet(nameof(AdditionalAssembly.Usings), out BlittableJsonReaderArray usingsArray);
+            
+            if (usingsArray != null)
             {
-                usings.Add(item.ToString());
+                foreach (var item in usingsArray)
+                {
+                    usings.Add(item.ToString());
+                }
             }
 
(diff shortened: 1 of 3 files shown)


src/Raven.Studio/typescript/viewmodels/database/indexes/editIndex.ts
```diff
@@ -241,9 +241,14 @@ class editIndex extends viewModelBase {
 
     private updateIndexFields() {
         const map = this.editedIndex().maps()[0].map();
+
         const additionalSourcesDto = {} as dictionary<string>;
         this.editedIndex().additionalSources().forEach(x => additionalSourcesDto[x.name()] = x.code());
-        new getIndexFieldsFromMapCommand(this.activeDatabase(), map, additionalSourcesDto)
+
+        const additionalAssembliesDto: Array<Raven.Client.Documents.Indexes.AdditionalAssembly> = [];
```
ml054: please use map instead of foreach.

const additionalAssembliesDto = this.editedIndex().additionalAssemblies().map(x => x.toDto());

Danielle9897: done

ml054 (approved):
studio part looks good

src/Raven.Server/Web/Studio/StudioIndexHandler.cs
```diff
@@ -112,6 +113,58 @@ public async Task PostIndexFields()
...
+
+        private static AdditionalAssembly GetAssembly(BlittableJsonReaderObject json)
+        {
+            json.TryGet(nameof(AdditionalAssembly.AssemblyName), out string assemblyName);
+            json.TryGet(nameof(AdditionalAssembly.AssemblyPath), out string assemblyPath);
+            json.TryGet(nameof(AdditionalAssembly.PackageName), out string packageName);
+            json.TryGet(nameof(AdditionalAssembly.PackageVersion), out string packageVersion);
+            json.TryGet(nameof(AdditionalAssembly.PackageSourceUrl), out string packageSourceUrl);
+                
+            json.TryGet(nameof(AdditionalAssembly.Usings), out BlittableJsonReaderArray usingsArray);
```
ppekrol: usingsArray can be null. You need to wrap the code below in `if`

Danielle9897: done

PR #11412, opened 2020-11-15
RavenDB-15606 Add "/admin/stats" end point to debug package

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15606


ppekrol:
This should go into 4.2

PR #11419, opened 2020-11-15
RavenDB-15814 When node is passive and license is not activated, stat…

Danielle9897:
…e so on the License View

### DIFF
--- a/src/Raven.Studio/typescript/viewmodels/shell/about.ts
+++ b/src/Raven.Studio/typescript/viewmodels/shell/about.ts
@@ -8,10 +8,15 @@ import accessManager = require("common/shell/accessManager");
 import forceLicenseUpdateCommand = require("commands/licensing/forceLicenseUpdateCommand");
 import getLatestVersionInfoCommand = require("commands/version/getLatestVersionInfoCommand");
 import getLicenseConfigurationSettingsCommand = require("commands/licensing/getLicenseConfigurationSettingsCommand");
+import clusterTopologyManager = require("common/shell/clusterTopologyManager");
+import appUrl = require("common/appUrl");
 
 class about extends viewModelBase {
 
     accessManager = accessManager.default.aboutView;
+    
+    clusterViewUrl = appUrl.forCluster();
+    passiveNode = ko.pureComputed(() => clusterTopologyManager.default.nodesCount() === 0);
     
     licenseCssClass = license.licenseCssClass;
     supportCssClass = license.supportCssClass;
--- a/src/Raven.Studio/typescript/viewmodels/shell/about.ts
+++ b/src/Raven.Studio/typescript/viewmodels/shell/about.ts
@@ -16,7 +16,7 @@ class about extends viewModelBase {
     accessManager = accessManager.default.aboutView;
     
     clusterViewUrl = appUrl.forCluster();
-    passiveNode = ko.pureComputed(() => clusterTopologyManager.default.nodesCount() === 0);
+    passiveNode = ko.pureComputed(() => clusterTopologyManager.default.topology().isPassive());
     
     licenseCssClass = license.licenseCssClass;
     supportCssClass = license.supportCssClass;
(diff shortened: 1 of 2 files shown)


src/Raven.Studio/typescript/viewmodels/shell/about.ts
```diff
@@ -8,11 +8,16 @@ import accessManager = require("common/shell/accessManager");
...
 import getLicenseConfigurationSettingsCommand = require("commands/licensing/getLicenseConfigurationSettingsCommand");
+import clusterTopologyManager = require("common/shell/clusterTopologyManager");
+import appUrl = require("common/appUrl");
 
 class about extends viewModelBase {
 
     accessManager = accessManager.default.aboutView;
     
+    clusterViewUrl = appUrl.forCluster();
+    passiveNode = ko.pureComputed(() => clusterTopologyManager.default.nodesCount() === 0);
```
ml054: isn't better to use: https://github.com/ravendb/ravendb/blob/135743f79f7a52bfb0f747c432689039f8ed4f15/src/Raven.Studio/typescript/models/database/cluster/clusterTopology.ts#L10 
from 
https://github.com/ravendb/ravendb/blob/v5.0/src/Raven.Studio/typescript/common/shell/clusterTopologyManager.ts#L13

src/Raven.Studio/wwwroot/App/views/shell/about.html
```diff
@@ -55,6 +55,22 @@
...
         </div>
+        <div data-bind="visible: passiveNode">
+            <div class="bg-info text-info flex-horizontal padding">
+                <div class="flex-start"><i class="icon-info"></i></div>
+                <div>The running server is in a <strong>Passive State</strong>, it is not part of a cluster yet.<br>
+                     In order to work, it is required that the server is part of a cluster.<br><br>
+                    Either one of the following can be done to <strong>Bootstrap a Cluster</strong>:<br>
+                     <ul>
+                         <li>Create a new database</li>
+                         <li>Register a license (if not registered yet)</li>
```
ml054: do we know if license is registered or no?

src/Raven.Studio/wwwroot/App/views/shell/about.html
```diff
@@ -55,6 +55,22 @@
                 </div>
             </div>
         </div>
+        <div data-bind="visible: passiveNode">
+            <div class="bg-info text-info flex-horizontal padding">
+                <div class="flex-start"><i class="icon-info"></i></div>
+                <div>The running server is in a <strong>Passive State</strong>, it is not part of a cluster yet.<br>
```
ml054: I'd join last sentence with this one.

src/Raven.Studio/wwwroot/App/views/shell/about.html
ml054: "In order to work" - what's not working?

src/Raven.Studio/typescript/viewmodels/shell/about.ts
Danielle9897: done

src/Raven.Studio/wwwroot/App/views/shell/about.html
Danielle9897: Refactored text to be:

image

src/Raven.Studio/wwwroot/App/views/shell/about.html
Danielle9897: > do we know if license is registered or no?

As far as I see, at this point, we don't have this info
The '`registered`' observable https://github.com/ravendb/ravendb/blob/efba92603d4ba571907188e038c03b0b91250e13/src/Raven.Studio/typescript/viewmodels/shell/about.ts#L189
returns relevant info **only** when we have a cluster

ml054 (approved):

ml054 (approved):

PR #11422, opened 2020-11-16
RavenDB-15625 stay in rehab with disabled index

efratshenhar:


src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
```diff
@@ -1363,6 +1363,12 @@ private string FindMostUpToDateNode(List<string> nodes, string database, Diction
                     return false;
                 }
 
+                if (mentorIndexStats.State != IndexState.Disabled && currentIndexStats.State == IndexState.Disabled)
+                {
```
ayende: Under what scenario can this happen?

ayende: Also, shouldn't we also check for indexing errors?

efratshenhar: We can change index to disabled only in one node. when that node tries to come back from rehab we don't want to promote him until the index is enabled.
index errors are being handled in https://issues.hibernatingrhinos.com/issue/RavenDB-15588

ayende: That would be a change in the same area of code, right?

efratshenhar: yes

ppekrol: It means that we will stay forever in Rehab in this situation, is this what we want?

ayende: What is the behavior that we _want_ here?

ppekrol (approved):
Can we rebase this?

efratshenhar:
Done

PR #11425, opened 2020-11-16
RDBCL-1656 copy native binaries outside of publish pipeline

gregolsky:


gregolsky:
@ppekrol Dumps do work if those 2 files are there in the package, but I'm wondering whether we need some more files. See:
```
Certain additional binaries may be optionally included with the app to enable debugging scenarios. The work to not require these binaries as separate files on disk alongside the app is ongoing.

For F5 debugging on VS and VS-Core: mscordbi.dll, libmscordbi.so
For Linux mini-dumps: createdump, libmscordaccore.so
For ETW / LTTng: clretwrc.dll, libcoreclrtraceptprovider.so
For exception stack trace source information, error string resources on Windows: Microsoft.DiaSymReader.Native.amd64.dll, mscorrc.debug.dll, mscorrc.dll
```

scripts/package.ps1
```diff
@@ -49,15 +49,45 @@ function GetRavenArchiveFileName ( $version, $target, $packageSuffix ) {
...
+    }
+
+}
+
 function LayoutServerPackage ( $packageDir, $projectDir, $packOpts ) {
     $target = $packOpts.Target
     CopyStudioPackageToServerOutputDirectory $packOpts
+
+    if ($target.IsUnix) {
+        # TODO get 'net5.0' from somewhere
```
gregolsky, line 79: @ppekrol do you know any good spot we can parse this out from? csproj?

PR #11426, opened 2020-11-16
RavenDB-15875 - Improve the performance of initial index run

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-15875


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -146,8 +146,11 @@ public abstract class Index : ITombstoneAware, IDisposable, ILowMemoryHandler
         private readonly ManualResetEventSlim _logsAppliedEvent = new ManualResetEventSlim();
 
         private DateTime? _lastQueryingTime;
+
         public DateTime? LastIndexingTime { get; private set; }
 
+        public long LastDatabaseEtagOnIndexCreation { get; private set; }
```
ppekrol: protected?

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -679,6 +682,10 @@ internal ExitWriteLock DrainRunningQueries()
 
                 OnInitialization();
 
+                LastDatabaseEtagOnIndexCreation = _environment.IsNew
```
ppekrol: Simplify it. Just load from the storage, if it is not there, calculate it and save it.

src/Raven.Server/Documents/Indexes/IndexItem.cs
```diff
@@ -29,6 +29,8 @@ public abstract class IndexItem : IDisposable
 
         public readonly IndexItemType ItemType;
 
+        public bool Referenced;
```
ppekrol: Reference? IsReference?

src/Raven.Server/Documents/Indexes/LastDatabaseEtag.cs
```diff
@@ -0,0 +1,39 @@
+﻿using Raven.Server.ServerWide.Context;
+using Voron;
+
+namespace Raven.Server.Documents.Indexes
+{
+    public class LastDatabaseEtag
```
ppekrol: We have IndexStorage for that. Move it there.

src/Voron/StorageEnvironment.cs
```diff
@@ -120,6 +120,8 @@ public static IDisposable GetStaticContext(out ByteStringContext ctx)
 
         private readonly long[] _validPages;
 
+        public bool IsNew { get; }
```
ppekrol, line 123: not needed?

src/Raven.Server/Documents/Indexes/MapIndexBase.cs
```diff
@@ -58,7 +58,7 @@ public override int HandleMap(IndexItem indexItem, IEnumerable mapResults, Index
                 mustDelete = _filters.Add(indexItem.LowerId) == false;
             }
 
-            if (mustDelete)
+            if ((LastDatabaseEtagOnIndexCreation < indexItem.Etag || indexItem.Referenced) && mustDelete)
```
ppekrol: You could store last etag per each of the collection needed, then you could also apply that for references?

arekpalinski (approved):
Aside from Pawel's comments - LGTM

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -679,6 +682,10 @@ internal ExitWriteLock DrainRunningQueries()
 
                 OnInitialization();
 
+                LastDatabaseEtagOnIndexCreation = _environment.IsNew 
+                    ? LastDatabaseEtag.Save(DocumentDatabase, this)
```
ayende: I don't particularly like the `Save` here. You are are mutating data in a way that is very subtle and hard to see. 
I would rather create this explicitly.

src/Raven.Server/Documents/Indexes/Index.cs
grisha-kotler: Cannot do that since this will break existing indexes.
It must be a new index or an index after a reset.

src/Raven.Server/Documents/Indexes/IndexItem.cs
grisha-kotler: done

src/Raven.Server/Documents/Indexes/MapIndexBase.cs
grisha-kotler: I need the etag of the _referenced_ document when indexing a reference.
Actually, we can skip indexing references entirely until we reach the `LastDatabaseEtagOnIndexCreation`

src/Raven.Server/Documents/Indexes/LastDatabaseEtag.cs
grisha-kotler: done

src/Raven.Server/Documents/Indexes/Index.cs
grisha-kotler: done

src/Raven.Server/Documents/Indexes/Index.cs
grisha-kotler: Done

src/Raven.Server/Documents/Indexes/Workers/HandleReferences.cs
```diff
@@ -212,6 +212,12 @@ private unsafe bool HandleItems(ActionType actionType, QueryOperationContext que
                                     hasChanges = true;
                                     inMemoryStats.UpdateLastEtag(lastEtag, isTombstone);
 
+                                    if (_index._indexStorage.LowerThanLastDatabaseEtagOnIndexCreation(referencedDocument.Etag))
```
ppekrol: There is a mistake here, the variable should be called 'referencedItem'. It can be a compare exchange.

Do the following:
1. `referencedDocument` => `referencedItem`
2.  change that condition to: `if (ItemsAndReferencesAreUsingSameEtagPool && _index._indexStorage.LowerThanLastDatabaseEtagOnIndexCreation(referencedItem.Etag))`

grisha-kotler: done

grisha-kotler: Actually, this isn't true.
The `referencedItem` is always a `referencedDocument` since you can only have `LoadDocument`.

PR #11434, opened 2020-11-17
RavenDB-15784: Resolve duplicate attachments in conflicted documents

garayx:


src/Raven.Server/Documents/AttachmentsStorage.cs
```diff
@@ -901,6 +922,124 @@ public AttachmentDetails MoveAttachment(DocumentsOperationContext context, strin
...
+            Attachment attachment;
+            using (DocumentIdWorker.GetLowerIdSliceAndStorageKey(context, name, out Slice lowerName, out Slice namePtr))
+            using (DocumentIdWorker.GetLowerIdSliceAndStorageKey(context, contentType, out Slice lowerContentType, out Slice contentTypePtr))
+            using (Slice.From(context.Allocator, hash, out Slice base64Hash)) // Hash is a base64 string, so this is a special case that we do not need to escape
+            using (GetAttachmentKey(context, lowerId.Content.Ptr, lowerId.Size, lowerName.Content.Ptr, lowerName.Size, base64Hash, lowerContentType.Content.Ptr,
+                lowerContentType.Size, AttachmentType.Document, Slices.Empty, out Slice keySlice))
+            {
+                var table = context.Transaction.InnerTransaction.OpenTable(AttachmentsSchema, AttachmentsMetadataSlice);
+                if (table.SeekOnePrimaryKeyPrefix(keySlice, out TableValueReader tvr) == false)
+                    return null;
```
ayende: Shouldn't we throw here as well?

src/Raven.Server/Documents/AttachmentsStorage.cs
```diff
@@ -901,6 +922,124 @@ public AttachmentDetails MoveAttachment(DocumentsOperationContext context, strin
...
+            using (GetAttachmentKey(context, lowerId.Content.Ptr, lowerId.Size, lowerName.Content.Ptr, lowerName.Size, base64Hash, lowerContentType.Content.Ptr,
+                lowerContentType.Size, AttachmentType.Document, Slices.Empty, out Slice keySlice))
+            {
+                var table = context.Transaction.InnerTransaction.OpenTable(AttachmentsSchema, AttachmentsMetadataSlice);
+                if (table.SeekOnePrimaryKeyPrefix(keySlice, out TableValueReader tvr) == false)
+                    return false;
+
+                var attachment = TableValueToAttachment(context, ref tvr);
+                if (attachment == null)
+                    AttachmentDoesNotExistException.ThrowFor(docId, name);
```
ayende: Why are you throwing here?

src/Raven.Server/Documents/Replication/ConflictManager.cs
```diff
@@ -137,6 +138,7 @@ public static void AssertChangeVectorNotNull(string conflictedChangeVector)
                 InvalidConflictWhenThereIsNone(conflict.Id);
 
             conflictedDocs.Add(conflict.Clone());
+            conflictedDocs = conflictedDocs.OrderByDescending(x => x.ChangeVector.ToChangeVectorList().Select(cv => cv.GetHashCode()).OrderByDescending(k => k).GetEnumerableHashCode()).ToList();
```
ayende: No, you need to sort it by the string of the change vector, not this. I'm not sure how this will sort at all.

src/Raven.Server/Documents/Replication/ResolveConflictOnReplicationConfigurationChange.cs
```diff
@@ -400,6 +406,57 @@ public DocumentConflict ResolveToLatest(List<DocumentConflict> conflicts)
...
+            {
+                if (group.Count() == 1)
+                    continue;
+
+                resolvedAttachmentsMetadata ??= GetAttachmentsFromMetadata(resolved.Doc);
+                var found = false;
+
+                foreach (var attachment in group)
+                {
+                    if (found == false && resolvedAttachmentsMetadata.Any(x => x.Name == attachment.Name && x.Hash == attachment.Hash && x.ContentType == attachment.ContentType))
```
ayende, line 425: I don't understand what you are doing here, can you explain?

src/Raven.Server/Documents/Replication/ResolveConflictOnReplicationConfigurationChange.cs
```diff
@@ -400,6 +406,57 @@ public DocumentConflict ResolveToLatest(List<DocumentConflict> conflicts)
...
+                }
+            }
+
+            List<AttachmentName> GetAttachmentsFromMetadata(BlittableJsonReaderObject bjro)
+            {
+                var list = new List<AttachmentName>();
+                if (bjro != null && resolved.Doc.TryGet(Constants.Documents.Metadata.Key, out BlittableJsonReaderObject resolvedMetadata) &&
+                    resolvedMetadata.TryGet(Constants.Documents.Metadata.Attachments, out BlittableJsonReaderArray resolvedArray))
+                {
+                    list = (from BlittableJsonReaderObject att in resolvedArray select JsonDeserializationClient.AttachmentName(att)).ToList();
```
ayende:
```suggestion
                    return (from BlittableJsonReaderObject att in resolvedArray select JsonDeserializationClient.AttachmentName(att)).ToList();
```

src/Raven.Server/Documents/Replication/ConflictManager.cs
garayx: This is when we resolve by script, we need to get same order of conflicted docs on all nodes/dbs, I used the hash of CV because it is the same for `A:1, B:1` and `B:1, A:1`, then ordered by `EnumerableHash`.

src/Raven.Server/Documents/Replication/ResolveConflictOnReplicationConfigurationChange.cs
garayx: In the storage we already have all attachments for particular document id (both conflict + local).
In resolved document we have the attachments metadata which can be modified in script resolver.

I get the attachments from storage to `storageAttachmentsDetails` then group by name, if I have more then 1 attachment in group, this mean I have duplicate attachments. Then I retrieve the `@metadata.@attachments` for resolved document and for each attachment in the group (duplicates) I check if I have the attachment in resolved metadata, if I do I mark it as `found` (so I don't have to check the resolved attachment so all next duplicates), for each next duplicate I "resolve" it according to the next logic: 
1. if `conflictResovler` doesn't have script I delete all duplicates, this mean only the attachment found in resolved metadata will be kept.
2. if  `conflictResovler` do have a script I keep all the duplicates and rename them, by this we don't let the use modify the attachment using the resolver script.

src/Raven.Server/Documents/AttachmentsStorage.cs
garayx: I think we should not throw here at all to not break the replication

src/Raven.Server/Documents/AttachmentsStorage.cs
garayx: I think we should not throw here at all to not break the replication

src/Raven.Server/Documents/Replication/ConflictManager.cs
ayende: You are calling GetHashCode, which isn't guaranteed to be identical
Note that the change vector _has_ to be the same across nodes.
There is no way you can get `A:1, B1` and `B:1, A1` in our system. The node that generated this decided on the format, and all else will follow that

src/Raven.Server/Documents/Replication/ResolveConflictOnReplicationConfigurationChange.cs
ayende: My issue is that you aren't syncing between the storage and document attachments.
And why are you comparing to the document values on hash & content type? What does the script give you in this case?

src/Raven.Server/Documents/Replication/ConflictManager.cs
karmeli87: @ayende you can have the situation of `A:1, B1` and `B:1, A1` and we consider this as an identical change vector.
Consider two nodes send simultaneously and each of them resolve it independently.  

But I agree that `GetHashCode` on a string is not deterministic, I would say order them lexicographically by the dbId.

src/Raven.Server/Documents/AttachmentsStorage.cs
```diff
@@ -901,6 +922,124 @@ public AttachmentDetails MoveAttachment(DocumentsOperationContext context, strin
...
+            using (Slice.From(context.Allocator, hash, out Slice base64Hash)) // Hash is a base64 string, so this is a special case that we do not need to escape
+            using (GetAttachmentKey(context, lowerId.Content.Ptr, lowerId.Size, lowerName.Content.Ptr, lowerName.Size, base64Hash, lowerContentType.Content.Ptr,
+                lowerContentType.Size, AttachmentType.Document, Slices.Empty, out Slice keySlice))
+            {
+                var table = context.Transaction.InnerTransaction.OpenTable(AttachmentsSchema, AttachmentsMetadataSlice);
+                if (table.SeekOnePrimaryKeyPrefix(keySlice, out TableValueReader tvr) == false)
+                    return null;
+
+                attachment = TableValueToAttachment(context, ref tvr);
+                if (attachment == null)
```
karmeli87: I this possible?

src/Raven.Server/Documents/AttachmentsStorage.cs
karmeli87: Is this even possible?

src/Raven.Server/Documents/Replication/ResolveConflictOnReplicationConfigurationChange.cs
```diff
@@ -400,6 +406,57 @@ public DocumentConflict ResolveToLatest(List<DocumentConflict> conflicts)
 
             return latestDoc;
         }
+
+        private void ResolveAttachmentsConflicts(DocumentsOperationContext context, DocumentConflict resolved)
+        {
+            using var scope = Slice.External(context.Allocator, resolved.LowerId, out var lowerId);
+            var storageAttachmentsDetails = _database.DocumentsStorage.AttachmentsStorage.GetAttachmentDetailsForDocument(context, lowerId);
+            List<AttachmentName> resolvedAttachmentsMetadata = null;
+            bool resolveToLatest = ScriptConflictResolversCache.TryGetValue(resolved.Collection, out ScriptResolver scriptResolver) == false || scriptResolver == null;
```
karmeli87: Note that you can define both `ResolveToLatest` and a script.
In that case, we try the script first and if we not succeed we will resolve to latest

src/Raven.Server/Documents/Replication/ConflictManager.cs
ayende: You cannot get that. Only one server can generate a change vector, and the version it generates is what it will use.

src/Raven.Server/Documents/Replication/ResolveConflictOnReplicationConfigurationChange.cs
garayx: The storage already have all attachments from all conflicts, when we put the resolved doc we recreate the attachments metadata from storage
https://github.com/ravendb/ravendb/blob/v4.2/src/Raven.Server/Documents/DocumentPutAction.cs#L395

In this method I retrieve the attachments from storage so I have the merged attachments list, and manipulate the according to conflict resolution strategy.

I compare the attachment from storage to attachment from resolved doc's metadata by name, hash, key (those 3 are part of the attachment key in the table). In this way I can know that I don't have to delete/rename this attachment in storage because it is in resolved metadata.

src/Raven.Server/Documents/Replication/ConflictManager.cs
garayx: changed to sort on string, and fixed the situation of `A:1, B1` and `B:1, A1` in cv

src/Raven.Server/Documents/AttachmentsStorage.cs
```diff
@@ -901,6 +922,116 @@ public AttachmentDetails MoveAttachment(DocumentsOperationContext context, strin
...
+                throw new ArgumentException($"Context must be set with a valid transaction before calling {nameof(DeleteAttachmentDirect)}", nameof(context));
+
+            using (DocumentIdWorker.GetLowerIdSliceAndStorageKey(context, name, out Slice lowerName, out Slice namePtr))
+            using (DocumentIdWorker.GetLowerIdSliceAndStorageKey(context, contentType, out Slice lowerContentType, out Slice contentTypePtr))
+            using (Slice.From(context.Allocator, hash, out Slice base64Hash)) // Hash is a base64 string, so this is a special case that we do not need to escape
+            using (GetAttachmentKey(context, lowerId.Content.Ptr, lowerId.Size, lowerName.Content.Ptr, lowerName.Size, base64Hash, lowerContentType.Content.Ptr,
+                lowerContentType.Size, AttachmentType.Document, Slices.Empty, out Slice keySlice))
+            {
+                var table = context.Transaction.InnerTransaction.OpenTable(AttachmentsSchema, AttachmentsMetadataSlice);
+                table.SeekOnePrimaryKeyPrefix(keySlice, out TableValueReader tvr);
```
karmeli87: if not found return false?

src/Raven.Server/Documents/Replication/ResolveConflictOnReplicationConfigurationChange.cs
```diff
@@ -400,11 +407,65 @@ public DocumentConflict ResolveToLatest(List<DocumentConflict> conflicts)
...
+                            continue;
+                        }
+                        // rename duplicates
+                        var newName = _database.DocumentsStorage.AttachmentsStorage.ResolveAttachmentName(context, lowerId, attachment.Name);
+                        _database.DocumentsStorage.AttachmentsStorage.RenameAttachment(context, lowerId, attachment.Name, attachment.Hash, attachment.ContentType, AttachmentType.Document, newName);
+                    }
+                }
+            }
+
+            static List<AttachmentName> GetAttachmentsFromMetadata(BlittableJsonReaderObject bjro)
```
karmeli87: instead, reuse `GetAttachmentsNameAndHash` or `GetAttachmentsFromDocumentMetadata` ? (move those methods `AttachmentsStorage`)

src/Raven.Server/Documents/AttachmentsStorage.cs
```diff
@@ -901,6 +922,116 @@ public AttachmentDetails MoveAttachment(DocumentsOperationContext context, strin
             return result;
         }
 
+        public AttachmentDetails RenameAttachment(DocumentsOperationContext context, Slice lowerId, string name, string hash, string contentType, AttachmentType attachmentType, string newName)
```
karmeli87: This looks very much as `MoveAttachment`, why not to use it?

src/Raven.Server/Documents/AttachmentsStorage.cs
```diff
@@ -901,6 +922,116 @@ public AttachmentDetails MoveAttachment(DocumentsOperationContext context, strin
...
+                        break;
+
+                    newName = $"{prefix}_#{++count}_{name}";
+                }
+            }
+
+            return newName;
+        }
+
+        public bool DeleteAttachment(DocumentsOperationContext context, Slice lowerId, string name, string hash, string contentType, AttachmentType attachmentType)
```
karmeli87: Why not to use the already existing `DeleteAttachment` ?

garayx: it uses partial key, added option to use full key

src/Raven.Server/Documents/AttachmentsStorage.cs
garayx: it uses partial key, added option to use full key

src/Raven.Server/Documents/Replication/ChangeVectorExtensions.cs
```diff
@@ -28,6 +29,8 @@ public static string SerializeVector(this ChangeVectorEntry[] self)
             if (self == null)
                 return null;
 
+            self = self.OrderBy(x => x.DbId, StringComparer.OrdinalIgnoreCase).ToArray();
```
ayende:
```suggestion
            Array.Sort(self, (x,y) => string.CompareOrdinal(x.DbId, y.DbId));
```

Let's avoid allocating a new array here each time

src/Raven.Server/Documents/Replication/ChangeVectorExtensions.cs
```diff
@@ -40,17 +43,7 @@ public static string SerializeVector(this ChangeVectorEntry[] self)
...
-
-            var sb = new StringBuilder();
-            for (int i = 0; i < self.Count; i++)
-            {
-                if (i != 0)
-                    sb.Append(", ");
-                self[i].Append(sb);
-            }
-            return sb.ToString();
+            return SerializeVector(self.ToArray());
```
ayende: Why do we allocate an array here?

src/Raven.Server/Documents/Replication/ChangeVectorExtensions.cs
garayx: done

src/Raven.Server/Documents/Replication/ChangeVectorExtensions.cs
garayx: done

karmeli87 (approved):

ppekrol (approved):
@garayx can you rebase?

ayende (approved):

PR #11440, opened 2020-11-18
RavenDB-15475 Adding unused database Ids to import/export for prevent ing "conflict" in import destination cluster replication

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15475

To reproduce the issue there need to be two steps before the replication:
1. The document and its revision have to come from export/backup & import.
2. last revision has to be a deletion revision.
* One obligate state - the first revision has a "conflict" with the deletion.

Reason:
1. If the document came from the same database the `UnusedDatabaseIds` solve the conflict.
2. The `import` deal first with documents so if the document is not deleted the first revision will be created by the destination of the import and be the same as the final document so all the rest of the revision will be treated as conflict but the final result be the desired document.

One way to solve it is to import the `UnusedDatabaseIds` as well (as proposed in this PR).
Another way can be to remove the unused database Ids from the document in the import process but this will require specific treatment in documents that there change vector composed of just unused database ids.

### DIFF
--- a/src/Raven.Client/Documents/Smuggler/SmugglerResult.cs
+++ b/src/Raven.Client/Documents/Smuggler/SmugglerResult.cs
@@ -254,9 +254,6 @@ namespace Raven.Client.Documents.Smuggler
                 if (ClientConfigurationUpdated)
                     json[nameof(ClientConfigurationUpdated)] = ClientConfigurationUpdated;
 
-                if (ConflictSolverConfigUpdated)
-                    json[nameof(ConflictSolverConfigUpdated)] = ClientConfigurationUpdated;
-
                 if (PeriodicBackupsUpdated)
                     json[nameof(PeriodicBackupsUpdated)] = PeriodicBackupsUpdated;
 
@@ -329,8 +326,8 @@ namespace Raven.Client.Documents.Smuggler
                 if (ClientConfigurationUpdated)
                     sb.AppendLine("- Client");
                 
-                if (ClientConfigurationUpdated)
-                    sb.AppendLine("- Unused Database Ids");
+                if (UnusedDatabaseIdsUpdated)
+                    sb.AppendLine("- Unused Database IDs");
 
                 if (sb.Length == 0)
                     return string.Empty;
--- a/src/Raven.Client/Documents/Smuggler/SmugglerResult.cs
+++ b/src/Raven.Client/Documents/Smuggler/SmugglerResult.cs
@@ -229,6 +229,8 @@ namespace Raven.Client.Documents.Smuggler
             public bool SqlConnectionStringsUpdated { get; set; }
 
             public bool ClientConfigurationUpdated { get; set; }
+            
+            public bool UnusedDatabaseIdsUpdated { get; set; }
 
             public override DynamicJsonValue ToJson()
             {
@@ -275,6 +277,9 @@ namespace Raven.Client.Documents.Smuggler
 
                 if (HubPullReplicationsUpdated)
                     json[nameof(HubPullReplicationsUpdated)] = HubPullReplicationsUpdated;
+                
+                if (UnusedDatabaseIdsUpdated)
+                    json[nameof(UnusedDatabaseIdsUpdated)] = UnusedDatabaseIdsUpdated;
 
                 return json;
             }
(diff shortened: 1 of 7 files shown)


src/Raven.Server/Smuggler/Documents/StreamSource.cs
```diff
@@ -206,6 +206,15 @@ public DatabaseRecord GetDatabaseRecord()
                     }
                 }
 
+                if (reader.TryGet(nameof(databaseRecord.UnusedDatabaseIds), out BlittableJsonReaderArray unusedDatabaseIds) &&
+                    unusedDatabaseIds != null)
+                {
+                    foreach (LazyStringValue id in unusedDatabaseIds)
```
karmeli87: I wonder if this can be `LazyCompressedStringValue`. Would it be safer to do: 
`foreach (string id in unusedDatabaseIds)`?
We have implicit conversion to `string` for both cases..

haludi: I have tried that but it throws an exception so I will change it to 
```
foreach (var id in unusedDatabaseIds)
{
    Debug.Assert(id is LazyStringValue || id is LazyCompressedStringValue);
    databaseRecord.UnusedDatabaseIds.Add(id.ToString());
}
```
Ok?

karmeli87: I think it would be better either to skip it and log an error or fail the import entirely.
We are doing this before we even imported a single doc right? so let's fail it completely..

src/Raven.Client/Documents/Smuggler/SmugglerResult.cs
```diff
@@ -323,6 +328,9 @@ public override string ToString()
 
                 if (ClientConfigurationUpdated)
                     sb.AppendLine("- Client");
+                
+                if (ClientConfigurationUpdated)
```
karmeli87: should be `UnusedDatabaseIdsUpdated` ?

karmeli87 (changes requested):

src/Raven.Client/Documents/Smuggler/SmugglerResult.cs
```diff
@@ -323,6 +328,9 @@ public override string ToString()
...
+                    sb.AppendLine("- Unused Database Ids");
```
ppekrol: Ids -> IDs

ppekrol (approved):

haludi:
There is a race condition if a "problematic" document replicated before the `UnusedDatabaseId`.

* A "problematic" document is a document that was deleted and its first revision has "conflict" (only if we don't remove the unused ids) with the others revision in particular with the deletion revision.

grisha-kotler:
@karmeli87 Can we remove the unused database ids on import?

karmeli87:
@grisha-kotler then you will have the problem this PR is fixing, no?
@haludi The race isn't specific for `UnusedDatabaseId`, it might happened for any other configuration as well.
For restoring from backup you can prepare the database record and send already with all configuration.
For importing you need to wait for the last raft index to be applied on all nodes of the cluster.
Let's handle it in a separate issue.

haludi:
> @grisha-kotler then you will have the problem this PR is fixing, no?
> @haludi The race isn't specific for `UnusedDatabaseId`, it might happened for any other configuration as well.
> For restoring from backup you can prepare the database record and send already with all configuration.
> For importing you need to wait for the last raft index to be applied on all nodes of the cluster.
> Let's handle it in a separate issue.

Restore - https://issues.hibernatingrhinos.com/issue/RavenDB-15962
Import - https://issues.hibernatingrhinos.com/issue/RavenDB-15963

src/Raven.Client/Documents/Smuggler/SmugglerResult.cs
```diff
@@ -252,9 +254,6 @@ public override DynamicJsonValue ToJson()
                 if (ClientConfigurationUpdated)
                     json[nameof(ClientConfigurationUpdated)] = ClientConfigurationUpdated;
 
-                if (ConflictSolverConfigUpdated)
```
ppekrol, line 255: Why this is removed? I see that the value is incorrect

haludi: I read the value and it seemed like a duplicate of the condition above.
Just an embarrassing mistake.

PR #11445, opened 2020-11-20
RavenDB-15877 Optimizing between queries

ayende:
Changing the settings for between queries that are used as secondary filter to avoid preloading a lot of data in most common situations.

### DIFF
--- a/src/Raven.Server/Documents/Queries/QueryBuilder.cs
+++ b/src/Raven.Server/Documents/Queries/QueryBuilder.cs
@@ -36,6 +36,12 @@ namespace Raven.Server.Documents.Queries
     {
         private static readonly KeywordAnalyzer KeywordAnalyzer = new KeywordAnalyzer();
 
+        private static readonly MultiTermQuery.ConstantScoreAutoRewrite _secondaryBetweenRewriteMethod = new MultiTermQuery.ConstantScoreAutoRewrite
+        {
+            TermCountCutoff = BooleanQuery.MaxClauseCount, // can be set using Query.MaxClauseCount
+            DocCountPercent = 200 // disable
+        };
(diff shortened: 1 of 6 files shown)


src/Raven.Server/Documents/Queries/QueryBuilder.cs
```diff
@@ -36,6 +36,12 @@ public static class QueryBuilder
     {
         private static readonly KeywordAnalyzer KeywordAnalyzer = new KeywordAnalyzer();
 
+        private static readonly MultiTermQuery.ConstantScoreAutoRewrite _secondaryBetweenRewriteMethod = new MultiTermQuery.ConstantScoreAutoRewrite
+        {
+            TermCountCutoff = BooleanQuery.MaxClauseCount, // can be set using Query.MaxClauseCount
```
ppekrol, line 41: we can change that in the config, shouldn't we take that from there?

ppekrol (approved):

src/Raven.Server/Documents/Queries/QueryBuilder.cs
ayende: We are doing that. The reason this is done this way is that this is the limit _inside lucene_

ppekrol: My bad, now I got this. For future reference: This assumes that ServerStore (the place where we are setting this) is initialized faster than QueryBuilder.

ayende: That must be the case, we cannot create a query builder without creating a db, which require a server

arekpalinski (approved):

PR #11448, opened 2020-11-22
RavenDB-15588 Stay in rehab with index error

efratshenhar:


src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
```diff
@@ -1363,6 +1363,14 @@ private string FindMostUpToDateNode(List<string> nodes, string database, Diction
                     return false;
                 }
 
+                if (currentIndexStats.State == IndexState.Error)
+                {
+                    if (mentorIndexStats.State == IndexState.Error)
+                        continue;
+                    reason = $"Index '{mentorIndex.Key}' is in state '{currentIndexStats.State}'";
```
ayende, line 1370: Is there any more information that we can provide here. What _was_ the error, etc.

efratshenhar: I don't think we save the reason after we log the state change

ayende (approved):

karmeli87 (approved):

PR #11450, opened 2020-11-22
RavenDB-15844  Database list view - Alerts and perf hints should be l…

Danielle9897:
…inks

### DIFF
--- a/src/Raven.Studio/wwwroot/App/views/resources/databases.html
+++ b/src/Raven.Studio/wwwroot/App/views/resources/databases.html
@@ -106,8 +106,9 @@
     </div>
     <div class="flex-grow scroll" data-bind="if: databases().sortedDatabases().length, visible: databases().sortedDatabases().length">
         <div data-bind="foreach: databases().sortedDatabases">
-            <div id="databaseItem" class="panel panel-hover database-item" data-bind="click: $root.databasePanelClicked, scrollTo: isCurrentlyActiveDatabase(), visible: !filteredOut(),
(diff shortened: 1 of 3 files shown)


src/Raven.Studio/wwwroot/App/views/resources/databases.html
```diff
@@ -280,16 +283,22 @@ <h3 class="database-name">
     </div>
     <div class="properties-item item-last-backup" data-bind="visible: $root.createIsLocalDatabaseObservable(name)()">
         <div data-bind="attr: { class: 'properties-value value-only ' + backupStatus() }">
-            <div class="set-size" title="Last backup time"
-                 data-bind="css: { 'link-disabled': !canNavigateToDatabase() || isBeingDeleted() }">
+            <a class="set-size text-danger" title="Navigate to Ongoing Tasks view to create a backup"
```
ml054, line 286: Shouldn't we redirect to new backups view ?

ml054: oh it is for 5.0 so maybe we should update this on 5.1?

Danielle9897: yes - it should be updated for 5.1

ml054: please create separate issue for that, ok ? we update that after this will be merged into 5.1.

ml054 (approved):

src/Raven.Studio/wwwroot/App/views/resources/databases.html
Danielle9897: https://issues.hibernatingrhinos.com/issue/RavenDB-15896

PR #11452, opened 2020-11-23
RavenDB-15735 Suggest X509KeyStorageFlags.MachineKeySet when the private key in null

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15735

### DIFF
--- a/src/Raven.Client/Documents/DocumentStore.cs
+++ b/src/Raven.Client/Documents/DocumentStore.cs
@@ -220,7 +220,7 @@ namespace Raven.Client.Documents
 
             RequestExecutor.ValidateUrls(Urls, Certificate);
 
-            if(Certificate.PrivateKey == null)
+            if(Certificate != null && Certificate.PrivateKey == null)
                 throw new InvalidOperationException(
                     $"The supplied {Certificate.FriendlyName} certificate contains no private key. " +
                     "Constructing the certificate with the 'X509KeyStorageFlags.MachineKeySet' flag may solve this problem.");
--- a/src/Raven.Client/Documents/DocumentStore.cs
+++ b/src/Raven.Client/Documents/DocumentStore.cs
@@ -220,6 +220,11 @@ namespace Raven.Client.Documents
 
             RequestExecutor.ValidateUrls(Urls, Certificate);
 
+            if(Certificate.PrivateKey == null)
+                throw new InvalidOperationException(
+                    $"The supplied {Certificate.FriendlyName} certificate contains no private key. " +
+                    "Constructing the certificate with the 'X509KeyStorageFlags.MachineKeySet' flag may solve this problem.");
+
             try
             {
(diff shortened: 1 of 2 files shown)


src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1356,10 +1356,24 @@ internal HttpRequestMessage CreateRequest<TResult>(JsonOperationContext ctx, Ser
...
+                        .Append("Forbidden access to ").Append(chosenNode.Database).Append("@").Append(chosenNode.Url).Append(", ");
+                    if (Certificate == null)
+                    {
+                        builder.Append("a certificate is required. ");
+                    }
+                    else if(Certificate.PrivateKey != null)
+                    {
+                        builder.Append(Certificate.FriendlyName).Append(" does not have permission to access it or is unknown. ");
+                    }
+                    else
```
ayende, line 1369: Those are already in the message from the server, no?

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1356,10 +1356,24 @@ internal HttpRequestMessage CreateRequest<TResult>(JsonOperationContext ctx, Ser
...
+                    {
+                        builder.Append("The certificate ").Append(Certificate.FriendlyName)
```
ayende, line 1371: This is the part where we tell you that the provided certificate is missing the private key, which is what is important, no?

src/Raven.Client/Documents/DocumentStore.cs
```diff
@@ -220,6 +220,11 @@ public override IDocumentStore Initialize()
 
             RequestExecutor.ValidateUrls(Urls, Certificate);
 
+            if(Certificate.PrivateKey == null)
```
ayende: And if the Certificate is null?

ayende (changes requested):

src/Raven.Client/Http/RequestExecutor.cs
haludi: {""Type"":""InvalidAuth"",""Message"":""This server requires client certificate for authentication, but none was provided by the client.""}

This is the message we got from the server.
I accidentally appended the `msg` twice in the beginning and in the end so I will remove it from the beginning so it will be as before.

src/Raven.Client/Http/RequestExecutor.cs
haludi: Append(" contains no private key. Constructing the certificate with the 'X509KeyStorageFlags.MachineKeySet' flag may solve this problem. "); 
https://github.com/ravendb/ravendb/pull/11452/files#diff-8eb5dec408a858236268c82f3fe1523a8d957370e6a24bef4e21079f70d36ae1R1373

ayende (approved):

PR #11453, opened 2020-11-23
RavenDB-14634 various fixes

karmeli87:


src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -2585,7 +2585,7 @@ private async Task<(long Index, object Result)> SendToLeaderAsyncInternal(Comman
         private async Task<(long Index, object Result)> SendToLeaderAsyncInternal(TransactionOperationContext context, CommandBase cmd)
         {
             //I think it is reasonable to expect timeout twice of error retry
-            var timeoutTask = TimeoutManager.WaitFor(Engine.OperationTimeout, _shutdownNotification.Token);
+            var timeoutTask = TimeoutManager.WaitFor(TimeSpan.FromMinutes(60), _shutdownNotification.Token);
```
ayende: Why is that?

karmeli87: we try to hunt why we got random timeout failures that cause to 4-5 tests simultaneously to fail

ppekrol: Something that @karmeli87 forgot to write here. This PR will not be merged. Just a pure bug hunting code.

@karmeli87 maybe track how much time we wait with stopwatch and if it is > Engine.OperationTimeout do a console writeline with some debug info? Thoughts?

ppekrol (approved):

src/Raven.Client/Documents/Operations/DatabaseHealthCheckOperation.cs
```diff
@@ -14,11 +15,16 @@ public RavenCommand GetCommand(DocumentConventions conventions, JsonOperationCon
 
         private class DatabaseHealthCheckCommand : RavenCommand
         {
+            public DatabaseHealthCheckCommand()
+            {
+                Timeout = TimeSpan.FromSeconds(15); // maybe even less?
```
ayende, line 20: Shouldn't we have a single location to define the default timeout?

src/Raven.Server/Commercial/LicenseManager.cs
```diff
@@ -468,7 +468,7 @@ public static LicenseStatus GetLicenseStatus(License license)
             return licenseStatus;
         }
 
-        public void TryActivateLicense(bool throwOnActivationFailure)
+        public async ValueTask TryActivateLicenseAsync(bool throwOnActivationFailure)
```
ayende: Why is this an `async ValueTask`?

src/Raven.Server/Rachis/RachisLogHistory.cs
```diff
@@ -379,6 +379,18 @@ private static unsafe string ReadGuid(Table.TableValueHolder entryHolder)
             return guid;
         }
 
+        public bool ContainsGuid(ClusterOperationContext context, string guid)
```
ayende: Guid is a specific term, but you are passing just ids here, which can be any strings

src/Raven.Server/ServerWide/Maintenance/ClusterObserver.cs
```diff
@@ -322,21 +322,45 @@ private void LogMessage(string message, Exception e = null, string database = nu
...
+                        }
+
+                        await _engine.PutAsync(cmd);
+                    }
                 }
             }
         }
 
+
+        private static string GetGuid(Dictionary<string, long> dic)
```
ayende: This is not a guid.

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -2359,15 +2359,15 @@ public void AssignNodesToDatabase(ClusterTopology clusterTopology, DatabaseRecor
             return SendToLeaderAsync(addDatabaseCommand);
         }
 
-        public async Task EnsureNotPassiveAsync(string publicServerUrl = null, string nodeTag = "A", bool skipLicenseActivation = false)
+        public async ValueTask EnsureNotPassiveAsync(string publicServerUrl = null, string nodeTag = "A", bool skipLicenseActivation = false)
```
ayende: Why?

ayende (changes requested):

src/Raven.Client/Documents/Operations/DatabaseHealthCheckOperation.cs
ppekrol: No, that would affect too many things. All internal'ish commands have this timeout set manually, In theory we could introduce sth like RavenInternalCommand which would set it by default.

src/Raven.Server/Rachis/RachisLogHistory.cs
karmeli87: Okay, so would something like: `CommandUniqueId` would be a more proper term?

src/Raven.Server/Commercial/LicenseManager.cs
karmeli87: In the common case `TryActivateLicenseAsync` just like `EnsureNotPassiveAsync` are expected to return without actually doing any `async` work.

src/Raven.Server/Rachis/RachisLogHistory.cs
ayende: Yes

src/Raven.Server/Commercial/LicenseManager.cs
ayende: See here: https://blog.marcgravell.com/2019/08/prefer-valuetask-to-task-always-and.html

In general, if we have to call `await` internally, that isn't saving a lot.
And if we are going to await internally and outside, there is little point in here.
Also, this isn't likely to be in anything that is in the hot path

karmeli87: I read this and others few times. TBH I'm still confused :-)

As I understand, the bottom line of that post is to use `ValueTask` _always_ (it is even in the title). But to be aware of the API changes.

Yet, as long as `AsyncTaskMethodBuilder` and friends aren't showing up in the profiler `Task` should be fine to use.

ayende: In general, in our internal API, prefer `Task` unless we know that:

* This is called frequently.
* We expect (and will write code) to handle the completion in a synchronous manner.

For external API:

* If they are likely to complete with no async work, use `ValueTask`

PR #11456, opened 2020-11-23
RavenDB-15903 - support TimeSpan in projections

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-15903


ayende (commented):
What about when the user wants to do aggregation? That was the first reason for this ask, no?

grisha-kotler:
The issue was that we didn't support `TimeSpan` in projections.
It was transalated to: 
```
select {
    Lasts:new System.TimeSpan(book.Calls.map((function(record){return record.Lasts.Ticks;})).reduce(function(a, b) { return a + b; }, 
}
```
I added a test to check this as well

aviv86 (approved):

PR #11461, opened 2020-11-24
RavenDB-15909 - fix NRE in stream query

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-15909


ayende:
What was the cause of this?

grisha-kotler:
Missing `Renew` after this https://github.com/ravendb/ravendb/blob/c20462a4b95e6c634f5a30b977271fcd20ac06e5/src/Raven.Client/Documents/Session/Operations/StreamOperation.cs#L576

PR #11466, opened 2020-11-25
RavenDB-15586 fixing leak in EncryptionBuffersPool that could occur when TryPush fails in per core container

ppekrol:


src/Voron/Impl/EncryptionBuffersPool.cs
```diff
@@ -66,6 +67,9 @@ public EncryptionBuffersPool()
 
                 thread = NativeMemory.ThreadAllocations.Value;
                 thread.Allocations += size;
+
+                Debug.Assert(size == allocation.Size, "size == allocation.Size");
```
grisha-kotler: Can you add the `size` and the `allocation.Size`?

ayende (approved):

grisha-kotler:
This is needed in 5.0 as well

src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -973,7 +978,12 @@ protected internal virtual unsafe void Reset(bool forceReleaseLongLivedAllocator
             for (var i = 0; i < _numberOfAllocatedStringsValues; i++)
                 _allocateStringValues[i].Reset();
 
-            _perCoreLazyStringValuesList.Push(_allocateStringValues);
+            if (_perCoreLazyStringValuesList.TryPush(_allocateStringValues) == false)
```
ayende: Need to check that this isn't already null. 
Pretty sure that this is why we are getting test failures.

ayende (changes requested):

src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -112,7 +112,12 @@ public JsonOperationContext(int initialSize, int longLivedSize, int maxNumberOfA
                 TryDispose(_arenaAllocatorForLongLivedValues);
                 if (_allocateStringValues != null)
                 {
-                    _perCoreLazyStringValuesList.Push(_allocateStringValues);
+                    if (_perCoreLazyStringValuesList.TryPush(_allocateStringValues) == false)
```
ayende, line 115:
```suggestion
                    if (_allocateStringValues != null && _perCoreLazyStringValuesList.TryPush(_allocateStringValues) == false)
```

src/Sparrow/Json/JsonOperationContext.cs
ayende:
```suggestion
            if (_allocateStringValues != null && _perCoreLazyStringValuesList.TryPush(_allocateStringValues) == false)
```

ayende (changes requested):

src/Sparrow/Json/JsonOperationContext.cs
ppekrol: This is checked one line up?

src/Sparrow/Json/JsonOperationContext.cs
ppekrol: Done

arekpalinski (approved):

PR #11470, opened 2020-11-26
RavenDB-15927 Collection queries are reading their entire input befor…

ayende:
…e they start running

### DIFF
--- a/src/Raven.Server/Documents/CollectionRunner.cs
+++ b/src/Raven.Server/Documents/CollectionRunner.cs
@@ -117,9 +117,6 @@ namespace Raven.Server.Documents
                                 }
 
                                 ids.Enqueue(document.Id);
-
-                                if (ids.Count >= OperationBatchSize)
-                                    break;
                             }
                         }
                     }
@@ -161,6 +158,8 @@ namespace Raven.Server.Documents
             {
                 if (_operationQuery == null)
                     _operationQuery = ConvertToOperationQuery(_collectionQuery);
+                
+                _operationQuery.PageSize = (int)batchSize;
 
                 return new CollectionQueryEnumerable(Database, Database.DocumentsStorage, new FieldsToFetch(_operationQuery, null),
                     collectionName, _operationQuery, null, context, null, null, new Reference<int>())
--- a/src/Raven.Server/Documents/CollectionRunner.cs
+++ b/src/Raven.Server/Documents/CollectionRunner.cs
@@ -164,6 +164,7 @@ namespace Raven.Server.Documents
                 return new CollectionQueryEnumerable(Database, Database.DocumentsStorage, new FieldsToFetch(_operationQuery, null),
                     collectionName, _operationQuery, null, context, null, null, new Reference<int>())
                 {
+                    Fields = fields,
                     InternalQueryOperationStart = start
                 };
             }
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/CollectionRunner.cs
```diff
@@ -85,36 +85,42 @@ public virtual Task<IOperationResult> ExecuteDelete(string collectionName, long
...
+                                {
+                                    end = true;
+                                    break;
+                                }
 
-                            ids.Enqueue(document.Id);
+                                ids.Enqueue(document.Id);
+
+                                if (ids.Count >= OperationBatchSize)
+                                    break;
```
ppekrol: The whole problem is that the collection query is:
1. Not supporting `DocumentField` - should be easy to do?
2. Shouldn't it support paging there, like the others inside 'GetDocuments' ?

src/Raven.Server/Documents/CollectionRunner.cs
```diff
@@ -85,36 +85,42 @@ public virtual Task<IOperationResult> ExecuteDelete(string collectionName, long
                     {
                         foreach (var document in GetDocuments(context, collectionName, startEtag, internalQueryOperationStart, OperationBatchSize, isAllDocs, DocumentFields.Id))
                         {
-                            internalQueryOperationStart++;
+                            using (document)
```
grisha-kotler, line 88: We dispose here only the document id and we collect here `OperationBatchSize` which is 1024 IIRC.

src/Raven.Server/Documents/CollectionRunner.cs
grisha-kotler: We already request `OperationBatchSize` in the `foreach`

src/Raven.Server/Documents/CollectionRunner.cs
ayende: I want to ensure that if we have encrypted database, we return the buffers ASAP

src/Raven.Server/Documents/CollectionRunner.cs
ayende: Yeah, that wasn't handled for `CollectionQueryRunner`, fixing it.

ppekrol:
@ayende I've made some adjustments, can you review?

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -697,58 +696,84 @@ public static long ReadLastEtag(Transaction tx)
         }
 
         public IEnumerable<Document> GetDocumentsStartingWith(DocumentsOperationContext context, string idPrefix, string matches, string exclude, string startAfterId,
-            long start, long take, string collection)
+            long start, long take, string collection, DocumentFields fields = DocumentFields.All)
         {
-            foreach (var doc in GetDocumentsStartingWith(context, idPrefix, matches, exclude, startAfterId, start, take))
+            var isAllDocs = collection == Constants.Documents.Collections.AllDocumentsCollection;
+            var requestedDataField = fields.HasFlag(DocumentFields.Data);
+            if (isAllDocs == false && requestedDataField == false)
```
ayende, line 703: why do we need this? to read the metadat?

ayende (commented):
LGTM

src/Raven.Server/Documents/DocumentsStorage.cs
ppekrol: Yes

arekpalinski:
LGTM

PR #11472, opened 2020-11-26
RavenDB-15915 Timeseries visualisation is broken

Danielle9897:


src/Raven.Studio/typescript/models/database/timeSeries/timeSeriesQueryResult.ts
```diff
@@ -1,7 +1,17 @@
 /// <reference path="../../../../typings/tsd.d.ts" />
 
 class timeSeriesQueryResult {
+    
+    queryHasGroupByTag = ko.observable<boolean>();
```
ml054: why we need observable here?

Danielle9897: removed

ml054 (approved):

PR #11475, opened 2020-11-26
RavenDB-15909 - Prevent NRE on session dispose after streaming

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-15909

### DIFF
--- a/src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
+++ b/src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
@@ -42,7 +42,7 @@ namespace Raven.Client.Documents.Session
     public abstract partial class InMemoryDocumentSessionOperations : IDisposable
     {
         internal long _asyncTasksCounter;
-        internal int? _maxDocsCountOnCachedRenewSession = null;
+        internal int _maxDocsCountOnCachedRenewSession = 16 * 1024;
         protected readonly RequestExecutor _requestExecutor;
         private OperationExecutor _operationExecutor;
         private readonly IDisposable _releaseOperationContext;
--- a/src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
+++ b/src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
@@ -42,6 +42,7 @@ namespace Raven.Client.Documents.Session
     public abstract partial class InMemoryDocumentSessionOperations : IDisposable
     {
         internal long _asyncTasksCounter;
+        internal int? _maxDocsCountOnCachedRenewSession = null;
         protected readonly RequestExecutor _requestExecutor;
         private OperationExecutor _operationExecutor;
         private readonly IDisposable _releaseOperationContext;
--- a/src/Raven.Client/Documents/Session/Operations/StreamOperation.cs
+++ b/src/Raven.Client/Documents/Session/Operations/StreamOperation.cs
@@ -90,7 +90,7 @@ namespace Raven.Client.Documents.Session.Operations
             return enumerator;
         }
 
-        private class YieldStreamResults : IAsyncEnumerator<BlittableJsonReaderObject>, IEnumerator<BlittableJsonReaderObject>
+        internal class YieldStreamResults : IAsyncEnumerator<BlittableJsonReaderObject>, IEnumerator<BlittableJsonReaderObject>
         {
             public YieldStreamResults(InMemoryDocumentSessionOperations session, StreamResult response, bool isQueryStream, bool isAsync, StreamQueryStatistics streamQueryStatistics)
             {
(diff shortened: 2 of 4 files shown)


src/Raven.Client/Documents/Session/Operations/StreamOperation.cs
```diff
@@ -356,12 +358,13 @@ private static void HandleStreamQueryStats(JsonOperationContext context, StreamR
...
             private void CheckIfContextNeedsToBeRenewed()
             {
-                if (_docsCountOnCachedRenewSession <= 16 * 1024)
+                if (_docsCountOnCachedRenewSession <= _maxDocsCountOnCachedRenewSession)
                 {
                     if (_cachedItemsRenew)
                     {
                         _session.Context.CachedProperties = new CachedProperties(_session.Context);
                         ++_docsCountOnCachedRenewSession;
+                        _cachedItemsRenew = false;
```
grisha-kotler, line 369: This isn't related to the fix to the NRE but needed to prevent us doing unnecessary work

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -42,6 +42,7 @@ namespace Raven.Client.Documents.Session
     public abstract partial class InMemoryDocumentSessionOperations : IDisposable
     {
         internal long _asyncTasksCounter;
+        internal int? _maxDocsCountOnCachedRenewSession = null;
```
arekpalinski: Why setting to null by default? In the test you can override it anyway

test/SlowTests/Issues/RavenDB-15909.cs
```diff
@@ -47,5 +50,44 @@ public async Task Cached_Properties_Should_Renew_Correctly()
...
+                    for (var i = 0; i < documentsCount; i++)
+                    {
+                        var expandoObject = new ExpandoObject();
+                        for (var j = 0; j < CachedProperties.CachedPropertiesSize; j++)
+                        {
+                            expandoObject.TryAdd((i + j).ToString(), null);
+                        }
+
+                        await bulkInsert.StoreAsync(expandoObject, i.ToString());
+                        expandoObject.Remove("Id", out _);
```
arekpalinski, line 71: Why do we call `Remove` here?

src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -923,10 +923,19 @@ protected internal virtual void Renew()
...
 
-            if (_perCorePathCache.TryPull(out _activeAllocatePathCaches) == false)
-                _activeAllocatePathCaches = new PathCache();
-            if (_perCoreLazyStringValuesList.TryPull(out _allocateStringValues) == false)
-                _allocateStringValues = new FastList<LazyStringValue>(256);
+            if (_activeAllocatePathCaches == null)
+            {
+                if (_perCorePathCache.TryPull(out _activeAllocatePathCaches) == false)
+                    _activeAllocatePathCaches = new PathCache();
+            }
```
arekpalinski, line 930: If `_activeAllocatePathCaches  != null` then don't we need to clean. We're gonna reuse the existing instance.

src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -923,10 +923,19 @@ protected internal virtual void Renew()
...
+
+            if (_allocateStringValues == null)
+            {
+                if (_perCoreLazyStringValuesList.TryPull(out _allocateStringValues) == false)
+                    _allocateStringValues = new FastList<LazyStringValue>(256);
+
+                _numberOfAllocatedStringsValues = 0;
+            }
```
arekpalinski, line 938: Same here - don't we need to clean / dispose `_allocateStringValue`. As I understand we just want to reuse `_allocateStringValues ` instance to avoid allocation of new `FastList`. But shouldn't we make sure it doesn't contain any old LSV items?

src/Sparrow/Json/JsonOperationContext.cs
arekpalinski: After talking to Pawel now I realized my comment isn't relevant  because we're in Renew method - not Reset

src/Sparrow/Json/JsonOperationContext.cs
arekpalinski: After talking to Pawel now I realized my comment isn't relevant because we're in Renew method - not Reset

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
arekpalinski: Fixed in commit https://github.com/ravendb/ravendb/pull/11475/commits/61ec6797f4052f5d3b9ca3b9189b1d715d128b44

test/SlowTests/Issues/RavenDB-15909.cs
grisha-kotler: Currently there is no need to do that. The previous code was reusing the same object.

PR #11482, opened 2020-11-28
RavenDB-15936 fixed DirectoryNotFoundException

ppekrol:


src/Sparrow.Server/LowMemory/CheckPageFileOnHdd.cs
```diff
@@ -391,17 +394,31 @@ public static string PosixIsSwappingOnHddInsteadOfSsd()
...
                 string foundRotationalDiskDrive = null;
                 for (int i = 0; i < swaps.Length; i++)
                 {
                     if (swaps[i].IsDeviceSwapFile)
                         continue; // we do not check swap file, only partitions
 
-                    // remove numbers at end of string (i.e.: /dev/sda5 ==> sda)
-                    var disk = _regExRemoveNumbers.Replace(swaps[i].DeviceName, "").Replace("/dev/", "");
+                    if (TryFindDisk(swaps[i].DeviceName, out var disk) == false)
+                        return null;
```
ppekrol: Not sure about returning `null` here. Should it be `continue` ?

src/Sparrow.Server/LowMemory/CheckPageFileOnHdd.cs
```diff
@@ -391,17 +394,31 @@ public static string PosixIsSwappingOnHddInsteadOfSsd()
...
+
                     var filename = $"/sys/block/{disk}/queue/rotational";
-                    var isHdd = KernelVirtualFileSystemUtils.ReadNumberFromFile(filename);
+                    if (File.Exists(filename) == false)
+                        return null;
```
ppekrol: Not sure about returning `null` here. Should it be `continue` ?

src/Sparrow.Server/LowMemory/CheckPageFileOnHdd.cs
```diff
@@ -391,17 +394,31 @@ public static string PosixIsSwappingOnHddInsteadOfSsd()
...
 
+                    var isHdd = KernelVirtualFileSystemUtils.ReadNumberFromFile(filename);
                     if (isHdd == -1)
                         return null;
```
ppekrol, line 423: How this can return -1? It should only be 0 or 1? On error this method returns long.MaxValue

src/Sparrow.Server/LowMemory/CheckPageFileOnHdd.cs
```diff
@@ -418,12 +435,19 @@ public static string PosixIsSwappingOnHddInsteadOfSsd()
                 if (foundRotationalDiskDrive != null)
                 {
                     // search if ssd drive is available
-                    HashSet<string> disks = KernelVirtualFileSystemUtils.GetAllDisksFromPartitionsFile();
-                    foreach (var disk in disks)
+                    foreach (var partitionDisk in KernelVirtualFileSystemUtils.GetAllDisksFromPartitionsFile())
                     {
                         //ignore ramdisks (ram0..ram15 etc)
-                        if (disk.Equals("ram", StringComparison.OrdinalIgnoreCase)) continue;
+                        if (partitionDisk.StartsWith("ram", StringComparison.OrdinalIgnoreCase))
```
ppekrol, line 441: Changed to StartsWith just in case.

src/Sparrow.Server/LowMemory/CheckPageFileOnHdd.cs
```diff
@@ -434,6 +458,27 @@ public static string PosixIsSwappingOnHddInsteadOfSsd()
...
+                    disk = null;
+
+                    if (string.IsNullOrWhiteSpace(deviceName))
+                        return false;
+
+                    deviceName = deviceName.Replace("/dev/", string.Empty);
+
+                    foreach (var block in blocks)
+                    {
+                        if (deviceName.Contains(block))
```
gregolsky, line 473: Not sure here. EndsWith or equals even (since you replce `/dev/`)?

src/Sparrow.Server/LowMemory/CheckPageFileOnHdd.cs
gregolsky: If the swap entry isn't related to any swapfile nor device, then what is it? I'd `LogInformation()` here that we could not determine what's with that swap.

And +1 for continue.

gregolsky:
Would be great to have a test with `KernelVirtualFileSystemUtils` and other device listing facilities underneath faked.

src/Sparrow.Server/LowMemory/CheckPageFileOnHdd.cs
gregolsky: +1 for continue.

PR #11486, opened 2020-11-29
RavenDB-15912 on data move should forget about cached compressed ids

karmeli87:


src/Voron/Data/RawData/ActiveRawDataSmallSection.cs
```diff
@@ -176,11 +177,11 @@ public string DebugDump(RawDataSmallPageHeader* pageHeader)
                         if (oldSize->IsCompressed)
                         {
                             using var __ = Table.DecompressValue(_transaction, entryPos, size, out var buffer);
-                            OnDataMoved(prevId, newId, buffer.Ptr, buffer.Length);
+                            OnDataMoved(prevId, newId, buffer.Ptr, buffer.Length, compressed: true);
```
ayende, line 180: Why not call this here?
` _tx.CachedDecompressedBuffersByStorageId.Remove(previousId);`

ayende (changes requested):

src/Voron/Data/RawData/ActiveRawDataSmallSection.cs
karmeli87: this is only one place.
You have this one when the page is defragged and you have when the density is lower than 0.15 we merge sections.

karmeli87: both which call `OnDataMoved`

ayende (approved):

src/Voron/Data/RawData/ActiveRawDataSmallSection.cs
arekpalinski: I like Karmel's approach. It's less error prone.

arekpalinski (approved):

PR #11487, opened 2020-11-29
RavenDB-15913 handle exceptions in timers

karmeli87:


src/Sparrow/Utils/NativeMemoryCleaner.cs
```diff
@@ -103,6 +103,10 @@ public void CleanNativeMemory(object state)
                     }
                 }
             }
+            catch (OutOfMemoryException)
```
ayende: We should handle all errors here, no?

src/Voron/Impl/EncryptionBuffersPool.cs
```diff
@@ -223,6 +223,10 @@ private void CleanupTimer(object _)
                     }
                 }
             }
+            catch (OutOfMemoryException)
```
ayende: Need to handle all errors here

ayende (changes requested):

src/Sparrow/Json/JsonContextPoolBase.cs
```diff
@@ -270,6 +270,10 @@ private void Cleanup(object _)
                         _globalStack.Push(context);
                 }
             }
+            catch (OutOfMemoryException)
```
ayende: What if we get `EarlyOutOfMemoryException`?

ayende (changes requested):

src/Sparrow/Utils/NativeMemoryCleaner.cs
karmeli87: Wasn't sure if we want to handle exceptions like double dispose or let it crash here intentionally.

src/Sparrow/Json/JsonContextPoolBase.cs
karmeli87: As far as I can tell we can get EOOM only when we allocating. Maybe I missing something but I don't see a code path here that might throw it.
Do will still want to safe-guard against it?

src/Sparrow/Utils/NativeMemoryCleaner.cs
ayende: It isn't so much as that error, but needing to handle _any_ error here, because otherwise we kill the process.

PR #11491, opened 2020-11-30
RavenDB-14600 Allow to include on the output of facets

ayende:


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -2987,6 +2988,51 @@ private void QueryInternalPreparation(IndexQueryServerSide query)
                             {
                                 result.Results = reader.FacetedQuery(facetQuery, queryContext.Documents, GetOrAddSpatialField, token.Token);
                                 result.TotalResults = result.Results.Count;
+
+                                if (query.Metadata.Includes?.Length > 0)
+                                {
+                                    var cmd = new IncludeDocumentsCommand(this.DocumentDatabase.DocumentsStorage, queryContext.Documents, query.Metadata.Includes, true);
+                                    foreach (string include in query.Metadata.Includes)
```
ppekrol: All of this should be handled by `cmd.Gather`. You exposed a lot of private fields, which is an indication that something is not handled the way it should.

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -2987,6 +2988,51 @@ private void QueryInternalPreparation(IndexQueryServerSide query)
...
+                                                {
+                                                    // expected, we can ignore this
+                                                    continue;
+                                                }
+                                                IncludeUtil.GetDocIdFromInclude(obj, new StringSegment(path), cmd.IncludedIds, DocumentDatabase.IdentityPartsSeparator);
+                                            }
+                                        }
+                                    }
+
+                                    queryContext.Documents.OpenReadTransaction();
```
ppekrol: I think that for consistency, when there are includes, we should not close the transaction before executing faceted query. It will give you results from snapshot where staleness was also calculated.

test/SlowTests/SlowTests/RavenDB-14600.cs
```diff
@@ -0,0 +1,88 @@
...
+    {
+        public RavenDB_14600(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        [Fact]
+        public void CanIncludeFacetResult()
+        {
+            using var store = GetDocumentStore();
+            store.Maintenance.Send(new CreateSampleDataOperation(Raven.Client.Documents.Smuggler.DatabaseItemType.Documents | Raven.Client.Documents.Smuggler.DatabaseItemType.Indexes));
```
ppekrol: Can we avoid creating that? It takes considerable amount of time.

test/SlowTests/SlowTests/RavenDB-14600.cs
```diff
@@ -0,0 +1,88 @@
...
+                    };
+                Index("EmployeeByDay", FieldIndexing.Exact);
+            }
+        }
+        
+        [Fact]
+        public void CanIncludeComplexFacetResult()
+        {
+            using var store = GetDocumentStore();
+            store.Maintenance.Send(new CreateSampleDataOperation(Raven.Client.Documents.Smuggler.DatabaseItemType.Documents ));
```
ppekrol: Can we avoid creating that? It takes considerable amount of time.

src/Raven.Server/Documents/Indexes/Index.cs
ayende: I don't want to do that, because we are not using `BlittableJsonObject` here, we are using the facet value, and I'm not sure if we want to introduce that behavior that deeply here.

src/Raven.Server/Documents/Indexes/Index.cs
ayende: I don't think we _open_ a document transaction here. We are only using index transaction, different altogether.

test/SlowTests/SlowTests/RavenDB-14600.cs
ayende: We'll just need to create this manually, instead, to be able to test the result.

ppekrol:
@ayende I've made some cleanup. Please review.

ayende (commented):
Looks much better

PR #11495, opened 2020-12-01
RavenDB-15944: add CultureInfo to HashCalculator

garayx:


src/Raven.Client/Documents/Queries/HashCalculator.cs
```diff
@@ -293,6 +294,9 @@ private void WriteParameterValue(object value, int level)
                     }
 
                     break;
+                case CultureInfo ci:
```
ayende, line 297: Why do we need to have a special  behavior here?
Why did the query hang?

garayx: `CultureInfo` has `Parent` field which has `Parent` as well and pointing to itself and so on, so it got stuck in `WriteParameterValue` method

https://github.com/ravendb/ravendb/blob/91ad9a0257e2d643a3def910b1acbdc60037569c/src/Raven.Client/Documents/Queries/HashCalculator.cs#L170

ayende: That means that any recursive structure will kill us, no?
There is supposed to be a limit of 100 here, no?
Also, why are we doing this manually instead of calling the seiralizer?

ppekrol: We can call serializer, but this code is called very frequently, so we need to accept performance implications of that.

The other option would be to have a HashSet<object> of visited instances, thoughts?

ayende: The problem is the correctness issue. On the server side, we do JSON serialization of complex values.
We need to do the same here.

ayende: Note that complex parameters are rare.

ppekrol: I will check it out

PR #11497, opened 2020-12-02
RavenDB-15911 Cleanup Temp directory on storage environment startup

arekpalinski:


src/Voron/StorageEnvironmentOptions.cs
```diff
@@ -695,13 +695,15 @@ public override unsafe void WriteHeader(string filename, FileHeader* header)
...
             }
 
-            public void DeleteAllTempBuffers()
+            public void DeleteAllTempFiles()
             {
                 if (Directory.Exists(TempPath.FullPath) == false)
                     return;
 
-                foreach (var file in Directory.GetFiles(TempPath.FullPath, "*.buffers"))
+                foreach (var file in Directory.GetFiles(TempPath.FullPath))
```
ayende: Why do we need this? Do we have other files here?

ayende: Never mind, the temp stream files.

ayende (approved):

src/Voron/StorageEnvironmentOptions.cs
arekpalinski: Yes. They are supposed to be deleted on close but if the process is aborted (e.g. because of AVE) then they might stay here.

arekpalinski:
After talking to Pawel I changed the impl. Instead of deleting all files in Temp dir I'm doing this selectively only for .buffers and .tmp files - this way we have explicit control over what we consider temporary files.

ppekrol (approved):

ayende (approved):

PR #11498, opened 2020-12-02
RavenDB-15941 Cluster log truncation should be limited by time

karmeli87:
- fix possible truncation of an uncommitted entry
- don't keep the write tx too long during truncation
- don't open the write tx on the leader if there is nothing to truncate


ayende (commented):
What is the context for this?

karmeli87:
I had a single node with 2M committed entries, which was "stuck" on the log truncation after booting up.

PR #11506, opened 2020-12-03
RavenDB-15976 Rate flow control for indexes journal writes.

redknightlois:
Concurrent writes to the indexes now can be synchronized at the database or server level. The functionality is implemented at the Voron level so while now the only synchronizing writes happen when indexing and can be controlled through configuration:
Indexing.JournalSynchronizationMode = {Immediate|Database|Server} = Database
Indexing.JournalMaxConcurrentWrites = 4
Indexing.JournalMaxConcurrentWritesSizeInMb = 128
This same mechanism can be used for any other purpose at the Journal level.


src/Raven.Server/Config/Categories/IndexingConfiguration.cs
```diff
@@ -222,6 +223,25 @@ public virtual bool RunInMemory
         [ConfigurationEntry("Indexing.NuGetPackageSourceUrl", ConfigurationEntryScope.ServerWideOnly)]
         public string NuGetPackageSourceUrl { get; set; }
 
+        [Description("Synchronization mode for the indexes")]
+        [DefaultValue(IndexWriteSynchronizationMode.Database)]
+        [IndexUpdateType(IndexUpdateType.None)]
+        [ConfigurationEntry("Indexing.JournalSynchronizationMode", ConfigurationEntryScope.ServerWideOnly)]
```
ayende, line 229: Any reason we shouldn't make it a per db configuration?

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
```diff
@@ -222,6 +223,25 @@ public virtual bool RunInMemory
         [ConfigurationEntry("Indexing.NuGetPackageSourceUrl", ConfigurationEntryScope.ServerWideOnly)]
         public string NuGetPackageSourceUrl { get; set; }
 
+        [Description("Synchronization mode for the indexes")]
```
ayende:
```suggestion
        [Description("Determine how RavenDB should coordinate expensive journal writes for indexes. This is used to reduce I/O contention with many indexes.")]
```

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -168,6 +171,20 @@ public DocumentDatabase(string name, RavenConfiguration configuration, ServerSto
                 });
                 _hasClusterTransaction = new AsyncManualResetEvent(DatabaseShutdown);
                 IdentityPartsSeparator = '/';
+
+                this.IndexWritesSynchronization = new StorageEnvironmentSynchronization();
```
ayende: Move this to the `case`'s `default`?

src/Raven.Server/Indexing/IndexWriteSynchronizationMode.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Raven.Server.Indexing
+{
+    public enum IndexWriteSynchronizationMode
+    {
+        Immediate,
```
ayende:
```suggestion
        None,
```

src/Voron/Impl/Journal/JournalWriter.cs
```diff
@@ -28,49 +28,87 @@ public unsafe class JournalWriter : IJournalWriter
...
+            _options = options;            
             FileName = filename;
             _log = LoggingSource.Instance.GetLogger<JournalWriter>(options.BasePath.FullPath);
 
+            if (writeSynchronization == null)
+                writeSynchronization = new StorageEnvironmentSynchronization();
+
+            // The write synchronization mechanism will require a max size and to be aligned to 4kb.  
+            _writeSynchronization = writeSynchronization;
+            _maxNumberOf4KbInSingleWrite = writeSynchronization.MaxWriteSize / 4096;
```
ayende:
```suggestion
            _maxNumberOf4KbInSingleWrite = writeSynchronization.MaxWriteSize /  (4L * Constants.Size.Kilobyte);
```

Let's keep this consistent.

src/Voron/Impl/Journal/JournalWriter.cs
```diff
@@ -28,49 +28,87 @@ public unsafe class JournalWriter : IJournalWriter
...
+                            _workingSetQuotaLogged = true;
+                        }
+                    }
+
+                    posBy4Kb += _maxNumberOf4KbInSingleWrite;
+                    currentBufferPtr += _maxNumberOf4KbInSingleWrite * 4L * Constants.Size.Kilobyte;
+                }
 
-                if (error == ERROR_WORKING_SET_QUOTA && _log.IsOperationsEnabled && _workingSetQuotaLogged == false)
+                // Then we will write the last chunk.
```
ayende: We can remove this.

src/Voron/Impl/Journal/JournalWriter.cs
```diff
@@ -28,49 +28,87 @@ public unsafe class JournalWriter : IJournalWriter
...
             using (var metrics = _options.IoMetrics.MeterIoRate(FileName.FullPath, IoMetrics.MeterType.JournalWrite, numberOf4Kb * 4L * Constants.Size.Kilobyte))
             {
                 sp = Stopwatch.StartNew();
 
-                var result = Pal.rvn_write_journal(_writeHandle, p, numberOf4Kb * 4L * Constants.Size.Kilobyte, posBy4Kb * 4L * Constants.Size.Kilobyte, out var error);
-                if (result != PalFlags.FailCodes.Success)
-                    PalHelper.ThrowLastError(result, error, $"Attempted to write to journal file - Path: {FileName.FullPath} Size: {numberOf4Kb * 4L * Constants.Size.Kilobyte}, numberOf4Kb={numberOf4Kb}");
+                // We will separate them in chunks of max size. 
+                long iterations = numberOf4Kb / _maxNumberOf4KbInSingleWrite;
+                long leftovers = numberOf4Kb % _maxNumberOf4KbInSingleWrite;
```
ayende: You can assert that this `leftovers` is zero here, we should always be passing 4KB aligned buffers here.

src/Voron/StorageEnvironment.cs
```diff
@@ -38,8 +38,49 @@ namespace Voron
...
+            this.MaxWriteSize = int.MaxValue;
+        }
+
+        public StorageEnvironmentSynchronization(int concurrentWrites, long maxWriteSize = long.MaxValue)
+        {
+            this._writeSemaphore = new Semaphore(concurrentWrites, concurrentWrites);
+            this.MaxWriteSize = maxWriteSize;
+        }        
+
+        public SynchronizationSection Enter()
```
ayende, line 85: We are going to need to plug this into the I/O metrics, so we can see how long we waited for this.

src/Voron/StorageEnvironment.cs
```diff
@@ -38,8 +38,49 @@ namespace Voron
...
+    public class StorageEnvironmentSynchronization
     {
+        public struct SynchronizationSection : IDisposable
+        {
+            private readonly Semaphore _context;
+
+            public SynchronizationSection(Semaphore context)
+            {
+                this._context = context;
+                if (this._context != null)
```
ayende: Under what scenario this can be null and we want to accept that?

src/Voron/StorageEnvironment.cs
```diff
@@ -38,8 +38,49 @@ namespace Voron
...
+                    this._context.WaitOne();
```
ayende: We need to make sure that this is an _interruptible_ wait.
Right now, if the index is waiting for a write, we can't shut down the server, for example.

src/Voron/Impl/Journal/JournalWriter.cs
```diff
@@ -28,49 +28,87 @@ public unsafe class JournalWriter : IJournalWriter
...
 
-                sp.Stop();
+                byte* currentBufferPtr = buffer;
+                for (long i = 0; i < iterations; i++ )
+                {
+                    using (_writeSynchronization.Enter())
```
ayende: I worry about fairness here. The semaphore isn't fair, which means that a tight _release_ and _acquire_ is going to see the same index keeping the lock for a long time.
That isn't necessarily a bad idea, mind, because we'll complete this one faster and there is a limit to how long we can stall here, but need to consider this explicitly.

ayende (commented):
We need to see what kind of performance difference this makes.
The scenarios in question should be:

* One db - one index
* One db - 10 indexes
* 5 dbs, - 1 index each
* 5 dbs - 10 indexes each

And then we want to measure:
* Amount of IOPS generated
* Time for the index to complete.

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
redknightlois: We could. But it makes a lot more sense to do that at the server level as that would give better behavior at the server level. I would say that unless we find the case where it really matters to control a single database write synchronization I would leave it Server wide.

src/Raven.Server/Documents/DocumentDatabase.cs
redknightlois: Yes, I am a bit hesitant if `switch` `default` because they tend to mask errors, but in this case which is essentially an ignore, there is no reason why not.

src/Voron/Impl/Journal/JournalWriter.cs
redknightlois: I suffered a bit after coming back the lack of comments, agree this particular one is not so important, but I will be writing a lot more from now on.

src/Voron/StorageEnvironment.cs
redknightlois: `None` is the scenario. We don't need to pay the synchronization cost if it is not used.

src/Voron/StorageEnvironment.cs
redknightlois: Makes sense, where do we get that interruption callback from?

src/Voron/Impl/Journal/JournalWriter.cs
redknightlois: I thought exactly about that, performance-wise hitting the same index/file is always faster than jumping around on sector addressable media. Therefore we kinda want it to be a bit unfair if possible.

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
ayende: `ServerWideOnly` vs `ServerWideOrDatabase` still allows you to define it globally.
This matters, because db only configuration is something that you can do in the cloud without needing manual assitance.

src/Voron/Impl/Journal/JournalWriter.cs
ayende: That isn't about the comments, that is about handling the last portion, there isn't going to be one.
It is always 4KB aligned.

src/Voron/StorageEnvironment.cs
ayende: The easiest is to the use the `DatabaseShutdown` cancellation token. IIRC, there is one for each indexc.

src/Voron/Impl/Journal/JournalWriter.cs
ayende: Need a comment explaining that, I think.

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
ppekrol: I also think it should be `ServerWideOrDatabase`

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
```diff
@@ -222,6 +223,25 @@ public virtual bool RunInMemory
...
+        [DefaultValue(4)]
+        [IndexUpdateType(IndexUpdateType.None)]
+        [ConfigurationEntry("Indexing.JournalMaxConcurrentWrites", ConfigurationEntryScope.ServerWideOnly)]
+        public int JournalMaxConcurrentWrites { get; set; }
+
+        [Description("Maximum size in Megabytes of the concurrent writes for the indexes")]
+        [DefaultValue(128)]
+        [IndexUpdateType(IndexUpdateType.None)]
+        [ConfigurationEntry("Indexing.JournalMaxConcurrentWritesSizeInMb", ConfigurationEntryScope.ServerWideOnly)]
+        public int JournalMaxConcurrentWritesSizeInMegabytes { get; set; }
```
ppekrol: Use `Size` type e.g.
https://github.com/ravendb/ravendb/blob/785eb6a41c1faf45331cfe17ebe8a216d0845c6a/src/Raven.Server/Config/Categories/StorageConfiguration.cs#L54-L58

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -96,6 +97,8 @@ public void ResetIdleTime()
             _lastIdleTicks = DateTime.MinValue.Ticks;
         }
 
+        public StorageEnvironmentSynchronization IndexWritesSynchronization { get; private set; }
```
ppekrol: This should go into IndexStore

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -177,7 +179,13 @@ public ServerStore(RavenConfiguration configuration, RavenServer server)
 
             HasFixedPort = Configuration.Core.ServerUrls == null ||
                            Uri.TryCreate(Configuration.Core.ServerUrls[0], UriKind.Absolute, out var uri) == false ||
-                           uri.Port != 0;
+                           uri.Port != 0;                       
+
+            this.IndexWritesSynchronization = new StorageEnvironmentSynchronization(
```
ppekrol: remove `this`

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -168,6 +171,22 @@ public DocumentDatabase(string name, RavenConfiguration configuration, ServerSto
...
+
+                switch (serverStore.Configuration.Indexing.JournalSynchronizationMode)
+                {
+                    case IndexWriteSynchronizationMode.Server:
+                        this.IndexWritesSynchronization = serverStore.IndexWritesSynchronization;
+                        break;
+                    case IndexWriteSynchronizationMode.Database:
+                        this.IndexWritesSynchronization = new StorageEnvironmentSynchronization(
+                                  serverStore.Configuration.Indexing.JournalMaxConcurrentWrites,
+                                  serverStore.Configuration.Indexing.JournalMaxConcurrentWritesSizeInMegabytes * Sparrow.Global.Constants.Size.Megabyte,
```
ppekrol: use `Size` type as a parameter, we do not want to deal with conversions manually. You will get it from the configuration after applying my previous remark.

src/Sparrow.Server/Meters/IoMetrics.cs
```diff
@@ -10,12 +10,12 @@ public enum MeterType
         {
             Compression,
             JournalWrite,
+            JournalWait,
```
ppekrol, line 13: We need to check if this requires any adjustments in Studio IO Metrics graph

src/Voron/Impl/Journal/JournalWriter.cs
```diff
@@ -28,49 +28,96 @@ public unsafe class JournalWriter : IJournalWriter
...
-                sp.Stop();
+                IoMeterBuffer.DurationMeasurement writeMetrics;
+
+                byte* currentBufferPtr = buffer;
+                for (long i = 0; i < iterations; i++ )
+                {
+                    writeMetrics = _options.IoMetrics.MeterIoRate(FileName.FullPath, IoMetrics.MeterType.JournalWait, _maxNumberOf4KbBlocks * 4L * Constants.Size.Kilobyte);
+                    using (_writeSynchronization.Enter())
+                    {
+                        writeMetrics.SetFileSize(_maxNumberOf4KbBlocks * (4L * Constants.Size.Kilobyte));
```
ppekrol: is this correct? each iteration is writing less?

src/Voron/StorageEnvironment.cs
```diff
@@ -38,8 +38,54 @@ namespace Voron
...
+
+    public class StorageEnvironmentSynchronization
     {
+        public struct SynchronizationSection : IDisposable
+        {
+            private readonly SemaphoreSlim _context;
+
+            public SynchronizationSection(SemaphoreSlim context, CancellationToken cancellationToken)
+            {
+                this._context = context;
```
ppekrol: in our code styling we do not use `this`

src/Voron/StorageEnvironment.cs
```diff
@@ -38,8 +38,54 @@ namespace Voron
...
+
+            void IDisposable.Dispose()
+            {
+                if (this._context != null)
+                    _context.Release();
+            }
+        }
+
+        private readonly SemaphoreSlim _writeSemaphore;
+        public readonly long MaxWriteSize;
```
ppekrol: rename to `MaxWriteSizeInBytes` ? Maybe leave it as `Size` ?

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
ppekrol: The property then should be `JournalMaxConcurrentWritesSize` and the configuration entry `Indexing.JournalMaxConcurrentWritesSizeInMb`

src/Voron/StorageEnvironment.cs
redknightlois: Like `MaxWriteSizeInBytes` more, because it is more explicit.

ppekrol: https://github.com/ravendb/ravendb/blob/v5.1/src/Sparrow/Size.cs

src/Voron/Impl/Journal/JournalWriter.cs
redknightlois: No `_maxNumberOf4KbBlocks` comes from the constructor, it won't change. The first iteration will write always the max number of blocks possible through configuration. The only case that is not the case is when `leftovers` is non-zero. The only 2 conditions when it might happen is smaller than biggest block write or when we processed everything and we are left with a partial block at the end.

src/Sparrow.Server/Meters/IoMetrics.cs
redknightlois: Yes, who would be the one to hand over this ticket when this is done, or do I create a new ticket?

ppekrol: Assign to @arekpalinski he will distrubte

src/Voron/Impl/Journal/JournalWriter.cs
redknightlois: Not really, those are 4kb aligned blocks :) a zero on both means no write in effect.

src/Voron/StorageEnvironment.cs
redknightlois: Ahhhh you mean to use the `Size` type... I wasn't understanding... roger that

src/Voron/Impl/Journal/JournalWriter.cs
ayende: The key here is that we want to remove the code to handle the left overs

redknightlois: Leftovers are 4kb aligned blocks that need to be written. Unless what you mean is that we could simplify that code instead of using 2 sequences, use a single that handle both cases.

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
```diff
@@ -222,6 +223,26 @@ public virtual bool RunInMemory
         [ConfigurationEntry("Indexing.NuGetPackageSourceUrl", ConfigurationEntryScope.ServerWideOnly)]
         public string NuGetPackageSourceUrl { get; set; }
 
+        [Description("Determine how RavenDB should coordinate expensive journal writes for indexes. This is used to reduce I/O contention with many indexes.")]
+        [DefaultValue(IndexWriteSynchronizationMode.Database)]
+        [IndexUpdateType(IndexUpdateType.None)]
+        [ConfigurationEntry("Indexing.JournalSynchronizationMode", ConfigurationEntryScope.ServerWideOnly)]
```
ayende, line 229:
```suggestion
        [ConfigurationEntry("Indexing.JournalSynchronizationMode", ConfigurationEntryScope.ServerWideOrDatabase)]
```

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
```diff
@@ -222,6 +223,26 @@ public virtual bool RunInMemory
...
+        public IndexWriteSynchronizationMode JournalSynchronizationMode { get; set; }
+
+        [Description("Maximum allowed concurrent writes for the indexes")]
+        [DefaultValue(4)]
```
ayende, line 233: Isn't that a high value, shouldn't we favor 1 / 2 ?

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -84,6 +88,22 @@ public IndexStore(DocumentDatabase documentDatabase, ServerStore serverStore)
 
             var stoppedConcurrentIndexBatches = _documentDatabase.Configuration.Indexing.NumberOfConcurrentStoppedBatchesIfRunningLowOnMemory;
             StoppedConcurrentIndexBatches = new SemaphoreSlim(stoppedConcurrentIndexBatches);
+
+            switch (_serverStore.Configuration.Indexing.JournalSynchronizationMode)
+            {
+                case IndexWriteSynchronizationMode.Server:
+                    IndexWritesSynchronization = serverStore.IndexWritesSynchronization;
```
ayende, line 95: Shouldn't we create a new instance here that uses `DatabaseShutdown`, instead? But the same semaphore as the server?

src/Voron/Impl/Journal/JournalWriter.cs
```diff
@@ -28,50 +29,75 @@ public unsafe class JournalWriter : IJournalWriter
...
-            _options = options;
+            _options = options;            
             FileName = filename;
             _log = LoggingSource.Instance.GetLogger<JournalWriter>(options.BasePath.FullPath);
 
+            if (writeSynchronization == null)
+                writeSynchronization = new StorageEnvironmentSynchronization();
+
+            // The write synchronization mechanism will require a max size and to be aligned to 4kb.  
+            _writeSynchronization = writeSynchronization;
```
ayende, line 50:
```suggestion
            // The write synchronization mechanism will require a max size and to be aligned to 4kb.  
            _writeSynchronization = writeSynchronization ?? new StorageEnvironmentSynchronization();
```

src/Voron/Impl/Journal/JournalWriter.cs
```diff
@@ -28,50 +29,75 @@ public unsafe class JournalWriter : IJournalWriter
...
+                long blocksToWrite = Math.Min(numberOf4KbBlocks, _maxNumberOf4KbBlocks);
 
-                var result = Pal.rvn_write_journal(_writeHandle, p, numberOf4Kb * 4L * Constants.Size.Kilobyte, posBy4Kb * 4L * Constants.Size.Kilobyte, out var error);
-                if (result != PalFlags.FailCodes.Success)
-                    PalHelper.ThrowLastError(result, error, $"Attempted to write to journal file - Path: {FileName.FullPath} Size: {numberOf4Kb * 4L * Constants.Size.Kilobyte}, numberOf4Kb={numberOf4Kb}");
+                IoMeterBuffer.DurationMeasurement writeMetrics = _options.IoMetrics.MeterIoRate(FileName.FullPath, IoMetrics.MeterType.JournalWait, blocksToWrite * 4L * Constants.Size.Kilobyte);
+                using (_writeSynchronization.Enter())
+                {
+                    writeMetrics.SetFileSize(blocksToWrite * (4L * Constants.Size.Kilobyte));
+                    writeMetrics.Dispose();
```
ayende, line 75: What happens on error? Can we wrap this in a `using` ?

src/Voron/StorageEnvironment.cs
```diff
@@ -38,8 +38,58 @@ namespace Voron
...
+            // the semaphore it can keep it around under contention to finish what it has started. 
+            private readonly SemaphoreSlim _context;
+
+            public SynchronizationSection(SemaphoreSlim context, CancellationToken cancellationToken)
+            {
+                _context = context;
+                if (_context != null)
+                {
+                    _context.Wait(cancellationToken);
+                    cancellationToken.ThrowIfCancellationRequested();
```
ayende, line 58: Shouldn't we always call this?

ayende (changes requested):

PR #11508, opened 2020-12-04
Async read and write support for counting stream

danielmarbach:
We are still getting occasionally the #11114 and I'm trying to get to the bottom of it. This is another path where a sync method is used that could benefit from using an already available async path.


ayende (approved):

danielmarbach:
Is there anything left to do here from my side to get this in? Unfortunately I don't even know what the message convention status check actually says because I don't see the result.

ppekrol:
@danielmarbach LGTM, but @grisha-kotler will do a final review

PR #11509, opened 2020-12-04
Avoid GetWaiter ChunkedRemoteBulkInsertOperation

danielmarbach:
Something that I noticed during my investigations. I couldn't understand why doing sync over async is required in the ctor of ChunkedRemoteBulkInsertOperation because judging by the code flow when we are creating a chunked operation it is always the root of the remote bulk operations therefore I think it safer to just create the underlying remote bulk operation.

### DIFF
--- a/Raven.Client.Lightweight/Document/ChunkedRemoteBulkInsertOperation.cs
+++ b/Raven.Client.Lightweight/Document/ChunkedRemoteBulkInsertOperation.cs
@@ -41,11 +41,7 @@ namespace Raven.Client.Document
             this.changes = changes;
             currentChunkSize = 0;
             RemoteBulkInsertOperationSwitches = 0;
-            using (NoSynchronizationContext.Scope())
-            {
-                var currentAsync = GetBulkInsertOperation().ConfigureAwait(false);
-                current= currentAsync.GetAwaiter().GetResult();
-            }
+            current = CreateBulkInsertOperation(cachedPreviousEmptyTask);
         }
 
         public Guid OperationId


ayende (commented):
IIRC, the logic here was that we want to get an immediate error if the db is down / bad db name, etc.

danielmarbach:
@ayende how would this happen? Judging by the code we would previously always land in

```
            if (current == null)
                return current = CreateBulkInsertOperation(cachedPreviousEmptyTask);
```

isn't the only place where you would detect such a scenario anyway when doing

```
        private void SubscribeToBulkInsertNotifications(IDatabaseChanges changes)
        {
            subscription = changes
                .ForBulkInsert(OperationId)
                .Subscribe(this);
        }
```

which is still called on this code path?

ayende:
This was a while ago, but the logic was meant to handle the case that you are writing to a missing database.
In that case, we would error immediately, instead of down the line.
Note that the subscription is actually async in nature too, which will not be created if the bulk insert isn't created.
Need to test how this behaves in the context of error handling with the db does not exists.

danielmarbach:
Ok so the current code before this change was calling.

```private async Task<RemoteBulkInsertOperation> GetBulkInsertOperation()
        {
            if (current == null)
                return current = CreateBulkInsertOperation(cachedPreviousEmptyTask);

            if (processedItemsInCurrentOperation < options.ChunkedBulkInsertOptions.MaxDocumentsPerChunk)
                if (options.ChunkedBulkInsertOptions.MaxChunkVolumeInBytes <= 0 || currentChunkSize < options.ChunkedBulkInsertOptions.MaxChunkVolumeInBytes)
                {
                    return current;
                }
            // if we haven't flushed the previous one yet, we will force
            // a disposal of both the previous one and the one before, to avoid
            // consuming a lot of memory, and to have _too_ much concurrency.
            if (previousTask != null)
            {
                await previousTask.ConfigureAwait(false);
            }
            previousTask = current.DisposeAsync();
            currentChunkSize = 0;
            processedItemsInCurrentOperation = 0;
            current = CreateBulkInsertOperation(previousTask);
            return current;
        }
```

and as far as I understand in the construction scenario we would always land on the fist condition and then return after creation. The rest of the creation logic is according to my potentially weak understanding of the code path only called when reentering `GetBulkInsertOperation` at later stages of the bulk insertion process

grisha-kotler (approved):

PR #11510, opened 2020-12-04
Make bulk insert writes against the server non-blocking

danielmarbach:


ayende (approved):

danielmarbach:
Is there anything left to do here from my side to get this in? Unfortunately I don't even know what the message convention status check actually says because I don't see the result.

grisha-kotler:
There are failing tests:
`bulk_insert_with_duplicates`
`can_finish_reduce`

Raven.Client.Lightweight/Document/RemoteBulkInsertOperation.cs
```diff
@@ -281,7 +280,7 @@ private void WriteQueueToServer(Stream stream, BulkInsertOptions options, Cancel
                         break;
                 }
 
-                FlushBatch(stream, batch);
+                return FlushBatch(stream, batch);
```
grisha-kotler: You cannot exit here since you are still processing the queue (this would fail after more than 512 docs in `BulkInsert`)

grisha-kotler:
Pushed my changes to your branch

grisha-kotler (approved):

Raven.Client.Lightweight/Document/RemoteBulkInsertOperation.cs
danielmarbach: Good catch. Missed that

Raven.Client.Lightweight/Document/RemoteBulkInsertOperation.cs
```diff
@@ -259,15 +258,16 @@ private Task WriteQueueToServer(Stream stream, BulkInsertOptions options, Cancel
 
                     if (document == null) // marker
                     {
-                        return FlushBatch(stream, batch);
+                        await FlushBatch(stream, batch).ConfigureAwait(false);
```
danielmarbach, line 261: I really should have cought this. Drowning in shame 😂

danielmarbach:
LGTM

ppekrol:
Did few runs on CI. All of them are hanging during `Raven.Tests` execution.

danielmarbach:
All tests are hanging? Is this something that is only on this PR or already something on v3. 5 branch?

danielmarbach:
I see that you are using XUnit. XUnit has a custom sync context to achieve test parallelization. The custom sync context can lead to deadlocks when sync over async is used.

danielmarbach:
What also has changed here is that previously it created a dedicated background thread that was not from the thread pool. With these changes the intent is that this is no longer required because the code path has no abilities to free up the thread. What's possible though is that while previously the `TrySetResult` was guaranteed to be called on that dedicated background thread. To migitate the risk we are using the TaskCompletionFactory internally though that should address and insert TaskCreationOptions RunContinuationsAsync into the PushContent

danielmarbach:
Any input on the above?

ayende:
All the tests passed, I'm okay with merging this.

ayende:
@grisha-kotler @ppekrol

danielmarbach:
Rebased

danielmarbach:
I guess this is the last one. Any insights why the recent tests failed? @ayende

ayende:
Looks like this is hanging here:

image


The test in question is: `CanExportImportSmugglerMaxSplitExportFileSize`, I'll check if this is reproducible in the single test.

ayende:
Yes, looks reproducible.

danielmarbach:
So this is the ChunkedBulkInsertOperations right? The change that deadlocks there is from

https://github.com/ravendb/ravendb/pull/11114

which has already been merged. Why would that deadlock given that you opt out from the sync context?

danielmarbach:
I've been looking through the code and the test failing and I don't know why suddenly the behavior should fundamentally change. The loop is still executed as a fire&forget offloaded task. I wonder if there is some kind of a race that the code has that in certain cases is exposed and making it async exaggerates the problem. I think I will stop here and close the PR

PR #11516, opened 2020-12-06
RavenDB-15983 Duplicate fields using CreateField return a single result

ayende:


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -3396,6 +3396,8 @@ private int MinBatchSize
         }
 
         private DateTime _lastCheckedFlushLock;
+        
+        public int CreatedFieldsCount;
```
ppekrol: I think this should be a part of CurrentIndexingScope it is created per batch?

src/Raven.Server/Documents/Indexes/Workers/MapDocuments.cs
```diff
@@ -79,6 +79,7 @@ public MapDocuments(Index index, DocumentsStorage documentsStorage, IndexStorage
                             {
                                 while (true)
                                 {
+                                    _index.CreatedFieldsCount = 0;
```
ppekrol: We have other works also. This needs to be reset per document, correct? We have Cleanup worker (do not think it is needed there), but for sure in HandleItems (HandleReferences) one? What about Map-Reduce?

src/Raven.Server/Documents/Indexes/Index.cs
ayende: No, this is per document, not per batch.
This is needed here because I need to reset it in the right place and this is the simplest location to track this.

ayende (commented):
Should be better now

ppekrol (approved):

PR #11517, opened 2020-12-06
RavenDB-15983 Duplicate fields using CreateField return a single result

ayende:


ppekrol:
There is already 4.2 PR opened: https://github.com/ravendb/ravendb/pull/11516

ayende:
Yes, there is a rename, so it would conflict when trying to merge

PR #11518, opened 2020-12-06
RavenDB-15538: Replicate all document attachments if there are MissingAttachmentsInLastBatch

garayx:


src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
```diff
@@ -379,6 +346,66 @@ public bool ExecuteReplicationOnce(TcpConnectionOptions tcpConnectionOptions, Ou
             }
         }
 
+        private bool CanContinueBatch(HashSet<short> missingTx, TimeSpan delay, ReplicationBatchItem item, long currentNext, long size, DocumentsOperationContext documentsContext,
```
karmeli87: Can we make a class out most of the args here? like `ReplicationState`

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
```diff
@@ -379,6 +346,66 @@ public bool ExecuteReplicationOnce(TcpConnectionOptions tcpConnectionOptions, Ou
...
+                missingTx.Remove(lastTransactionMarker);
+
+                if (missingTx.Count != 0)
+                {
+                    return true;
+                }
+            }
+
+            var batchSize = _parent._database.Configuration.Replication.MaxItemsCount;
+            var maxSizeToSend = _parent._database.Configuration.Replication.MaxSizeToSend;
```
karmeli87: And maybe put those in that class.

karmeli87 (approved):

ppekrol (approved):

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
```diff
@@ -379,6 +356,62 @@ public bool ExecuteReplicationOnce(TcpConnectionOptions tcpConnectionOptions, Ou
             }
         }
 
+        private bool CanContinueBatch(ReplicationState state, ref long next)
+        {
+            if (MissingAttachmentsInLastBatch)
```
ayende, line 361: Isn't this called for any item? In that case, what is the meaning of removing it one at a time?

garayx: this code called only when `item.TransactionMarker != replicationState.LastTransactionMarker`
then we try to remove `replicationState.LastTransactionMarker`, if there are `state.MissingTxMarkers` we have to continue the replication batch, else we allow to interrupt it if needed

PR #11534, opened 2020-12-09
RavenDB-15749 Map view for Spatial queries

Danielle9897:


src/Raven.Studio/typescript/viewmodels/database/query/query.ts
```diff
@@ -184,6 +186,12 @@ class query extends viewModelBase {
     showTimeSeriesGraph: KnockoutComputed<boolean>;
     showPlotButton: KnockoutComputed<boolean>;
     
+    isSpatialQuery = ko.observable<boolean>();
+    spatialDataFound = ko.observable<boolean>();
+    spatialMap = ko.observable<spatialQueryMap>();
+    showMapView: KnockoutComputed<boolean>;
+    viewOnMapClicked = ko.observable<boolean>();
```
ml054: I don't think we need that field? What if someone clicks on the map twice?

src/Raven.Studio/typescript/viewmodels/database/query/query.ts
```diff
@@ -1168,6 +1179,45 @@ class query extends viewModelBase {
...
     }
     
+    private onSpatialLoaded(queryResults: pagedResultExtended<document>) {
+        const spatialProperties = queryResults.additionalResultInfo.SpatialProperties;
+        if (spatialProperties && queryResults.items.length) {
+            this.isSpatialQuery(true);
+            this.spatialDataFound(true);
+
+            // Each spatial map model will contain the layer of markers per spatial properties pair
+            let spatialMapModels: spatialMapLayerModel[] = [];
```
ml054: const?

src/Raven.Studio/typescript/viewmodels/database/query/query.ts
```diff
@@ -1168,6 +1179,45 @@ class query extends viewModelBase {
...
+                const latitudeProperty = spatialProperties[i].LatitudeProperty;
+                const longitudeProperty = spatialProperties[i].LongitudeProperty;
+
+                let pointsArray: geoPoint[] = [];
+                for (let i = 0; i < queryResults.items.length; i++) {
+                    const item = queryResults.items[i];
+
+                    const latitudeValue = _.get(item, latitudeProperty) as number;
+                    const longitudeValue = _.get(item, longitudeProperty) as number;
+                    this.spatialDataFound(!!latitudeValue && !!longitudeValue);
```
ml054: what if all values are (0,0)? Do we need this spatial data found?

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
```diff
@@ -0,0 +1,104 @@
+import viewModelBase = require("viewmodels/viewModelBase");
+import spatialMapLayerModel = require("models/database/query/spatialMapLayerModel");
+import document = require("models/database/documents/document");
+import documentMetadata = require("models/database/documents/documentMetadata");
+import {Layer, MarkerClusterGroup} from "leaflet";
```
ml054: does this kind of import work?

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    compositionComplete() {
+        super.compositionComplete();
+        this.createMap();
+    }
+    
+    createMap() {
+
+        ////////////////////////////
```
ml054: please remove those comment banners - either use single line comments:
```
// Define base tile layers
```

or create separate methods and call them one by one.

ml054: let's keep coding style consistent.

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
```diff
@@ -0,0 +1,104 @@
...
+        const baseLayers = {
+            "Streets Map": osmMap,
+            "Topography Map": otmMap
+        };
+
+        //////////////////////////////////////////////////////////////
+        // Define group layers (layer per spatial point from query)
+        //////////////////////////////////////////////////////////////
+
+        let dataLayers = {};
```
ml054: const?

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
```diff
@@ -0,0 +1,104 @@
...
+        let markersGroups: MarkerClusterGroup[] = [];
+
+        this.geoData().forEach((layerModel) => {
```
ml054, line 49: maybe use map instead of foreach and pushing to array?

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
```diff
@@ -0,0 +1,104 @@
...
+            
+            let markers: Layer[] = [];
+            let markersGroup = L.markerClusterGroup();
+
+            const generatePopupContent = (doc: document) => {
```
ml054: move this function outside of foreach block

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
```diff
@@ -0,0 +1,104 @@
...
+                const docDto = doc.toDto(true);
+                const metaDto = docDto["@metadata"];
+                documentMetadata.filterMetadata(metaDto);
+
+                let text = JSON.stringify(docDto, null, 4);
+                text = Prism.highlight(text, (Prism.languages as any)["javascript"]);
+
+                const textHtml = `<div>
+                                    <h4>Document: ${doc.getId()}</h4>
```
ml054: escaping! Please remeber about that every time you concat user defined input with html tags.

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
```diff
@@ -0,0 +1,104 @@
...
+            }
+
+            layerModel.geoPoints().forEach((point) => {
+                const pointMarker = L.marker([point.latitude, point.longitude], { title: point.popupContent.getId() })
+                                     .bindPopup(generatePopupContent(point.popupContent), { 'className' : 'custom-popup', 'maxWidth': 600, 'maxHeight': 400 } );
+                markers.push(pointMarker);
+            });
+
+            markersGroup.addLayers(markers);
+            const controlLayerText = `<span>Point fields:</span>
```
ml054: escape

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
ml054: + it doesnt have any type. Please use type: Record<...>

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
```diff
@@ -0,0 +1,104 @@
...
+                                      </div>`;
+            (dataLayers as any)[controlLayerText] = markersGroup;
+            markersGroups.push(markersGroup);
+        })
+
+        //////////////////////
+        // Define the map 
+        //////////////////////
+
+        let map = L.map('mapid', {preferCanvas: true});      // init map
```
ml054: please use double quotes whenever possible

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
```diff
@@ -0,0 +1,104 @@
...
+        osmMap.addTo(map);                                   // init base layer to use
+        L.control.layers(baseLayers, dataLayers).addTo(map); // init legend control
+        markersGroups.forEach(group => map.addLayer(group)); // init markers groups
+        
+        let bounds = markersGroups[0].getBounds();
```
ml054: const?

ml054: can you start with empty bound L.bounds([])? 
Then we don;t need special treatment for first and other elements?

src/Raven.Studio/wwwroot/App/views/database/query/query.html
```diff
@@ -216,36 +222,42 @@ <h2 class="on-base-background">
                 <div class="panel-body absolute-fill js-time-series-graph-container" data-bind="if: showTimeSeriesGraph(), visible: showTimeSeriesGraph()">
                     <div class="absolute-fill flex-vertical" data-bind="compose: $root.currentTab"></div>
                 </div>
+                <div class="panel-body absolute-fill js-time-series-graph-container" data-bind="if: showMapView(), visible: showMapView()">
```
ml054: js-time-series-graph-container" data-bind="if: showMapView()

time series and map view?

src/Raven.Studio/wwwroot/App/views/database/query/query.html
```diff
@@ -216,36 +222,42 @@ <h2 class="on-base-background">
...
             </div>
         </div>
         <div class="flex-window-head">
-            <button type="button" class="btn btn-sm" 
+            <button type="button" class="btn btn-sm"
                     data-bind="css: { active: currentTab() === 'results'}, click: goToResultsTab">
                 Results &nbsp; <span class="label" data-bind="text: totalResultsForUi().toLocaleString() + (hasMoreUnboundedResults() ? '+' : ''), css: { 'label-primary': totalResultsForUi(), 'label-default': !totalResultsForUi() }"></span>
             </button>
+            <button type="button" class="btn btn-sm" data-bind="click: $root.goToSpatialMapTab, css: { active: showMapView }, visible: viewOnMapClicked">
+                <span>Spatial Map</span> &nbsp; <span class="label" data-bind="text: totalResultsForUi().toLocaleString() + (hasMoreUnboundedResults() ? '+' : ''), css: { 'label-primary': totalResultsForUi(), 'label-default': !totalResultsForUi() }"></span>
```
ml054: this is probably not true since we don't load all records to memory, right?

src/Raven.Studio/typescript/viewmodels/database/query/query.ts
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
Danielle9897: `map.fitBounds()` expects a `LatLngBoundsExpression `expression which is **either** what you get from a `markersGroup.getBounds()` method and then you can extend it to include bounds of another group,
**or** a plain list of all the points in the data.

So it is **either** doing it as I did here where each group bounds extends the other, 
**or** making something like:

```
let allPoints: any[] = [];
        this.geoData().forEach(x => {
            x.geoPoints().forEach(y => {
                allPoints.push([y.latitude, y.longitude]);
            })
        })

map.fitBounds(allPoints, {padding: [50, 50]});
```
`L.bounds()` returns `Bounds` which represents a rectangular area in pixel coordinates
and is Not passed to` map.fitBounds`

src/Raven.Studio/wwwroot/App/views/database/query/query.html
Danielle9897: 1. Now showing the number of markers in the bottom tab (and Not the number of resulting documents)
2. Also showing the number per layer in the legend control

image
image

src/Raven.Studio/wwwroot/App/views/database/query/query.html
Danielle9897: removed that ts class

src/Raven.Studio/typescript/viewmodels/database/query/query.ts
Danielle9897: Removed the _View on Map_ button.
Removed this field as well.

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
ml054: ?

Danielle9897: ok - done - with `L.latLngBounds([];`

src/Raven.Studio/typescript/viewmodels/database/query/query.ts
Danielle9897: `spatialDataFound `removed.
Only markers that are found are pushed to the array.

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
Danielle9897: done
`const dataLayers: Control.LayersObject = {};`

src/Raven.Studio/wwwroot/App/views/database/query/query.html
ml054: great idea

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -3317,6 +3317,11 @@ private static void ThrowInvalidField(string f)
             result.LastQueryTime = _lastQueryingTime ?? DateTime.MinValue;
             result.ResultEtag = CalculateIndexEtag(queryContext, indexContext, q, result.IsStale);
             result.NodeTag = DocumentDatabase.ServerStore.NodeTag;
+            
+            if (q.SpatialResults != null)
+            {
+                result.SpatialProperties = q.SpatialResults.ToArray();
```
arekpalinski: Naming inconsistency - so those are sparial properties or results?

src/Raven.Server/Json/BlittableJsonTextWriterExtensions.cs
```diff
@@ -230,6 +231,20 @@ public static void WriteFacetedQueryResult(this BlittableJsonTextWriter writer,
             writer.WriteEndObject();
         }
 
+        public static void WriteSpatialPropertyResult(this AbstractBlittableJsonTextWriter writer, JsonOperationContext context, SpatialProperty result)
```
arekpalinski: It can be `private`

src/Raven.Server/Documents/Queries/QueryMetadata.cs
```diff
@@ -2381,6 +2393,21 @@ private void HandleSum(List<QueryExpression> arguments, BlittableJsonReaderObjec
 
                 _metadata.AddWhereField(new QueryFieldName(f.FieldValue, f.IsQuoted), parameters);
             }
+
+            private string GetPathWithAliasName(string propertyPath)
```
arekpalinski: What does it actually do? Can you show concrete example? Will that work for fields with nesting like `Foo.Bar.Baz.Bee` ?

src/Raven.Client/Documents/Queries/QueryResultBase.cs
```diff
@@ -84,5 +84,10 @@ public abstract class QueryResultBase<TResult, TInclude>
         /// Detailed timings for various parts of a query (Lucene search, loading documents, transforming results) - if requested.
         /// </summary>
         public QueryTimings Timings { get; set; }
+        
+        /// <summary>
+        /// List of longitude & latitude document properties names in a spatial query
+        /// </summary>
+        public SpatialProperty[] SpatialProperties { get; set; }
```
arekpalinski: Do we need to add it to `QueryResultBase`? I don't see a similar property for time series here. How do we handle that for time series?

src/Raven.Server/Documents/Queries/QueryMetadata.cs
```diff
@@ -2317,14 +2319,24 @@ public void HandleSpatial(string methodName, List<QueryExpression> arguments, bo
...
                             });
+
+                            if (_metadata.SelectFields != null)
+                            {
+                                latitudePropertyPath = GetPathWithAliasName(latitudePropertyPath);
+                                longitudePropertyPath = GetPathWithAliasName(longitudePropertyPath);
+                            }
+                            
+                            _metadata.SpatialResults ??= new List<SpatialProperty>();
+                            _metadata.SpatialResults.Add(new SpatialProperty(latitudePropertyPath, longitudePropertyPath));
```
arekpalinski: This is only needed for studio. Technically we could probably do it only if it's explicitly requested by the studio. Although query metadata are cached so I think it's okay to have this information always in metadata - it will be calculated just once per query.

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -3317,6 +3317,11 @@ private static void ThrowInvalidField(string f)
             result.LastQueryTime = _lastQueryingTime ?? DateTime.MinValue;
             result.ResultEtag = CalculateIndexEtag(queryContext, indexContext, q, result.IsStale);
             result.NodeTag = DocumentDatabase.ServerStore.NodeTag;
+            
+            if (q.SpatialResults != null)
```
arekpalinski: This means that _every_ time we answer a spatial query we'll be including `SparialProperties` as well. We'd prefer to include it _only_ if it's explicitly requested by the studio. I think we do it this way for time series

arekpalinski (changes requested):

src/Raven.Server/Json/BlittableJsonTextWriterExtensions.cs
Danielle9897: done

src/Raven.Server/Documents/Indexes/Index.cs
Danielle9897: done - renamed to `SpatialProperties`

src/Raven.Server/Documents/Queries/QueryMetadata.cs
Danielle9897: If for example, you query and select some fields, e.g.: 
```
spatial.point(Foo.Bar.Baz.Been.Latitude, Foo.Bar.Baz.Bee.Longitude)
...
select Foo as AliasName
```

then this method will return
```
AliasName.Bar.Baz.Bee.Latitude
```
and
```
AliasName.Bar.Baz.Bee.Longitude
```
so that the studio will know where to look for the values in the results

src/Raven.Client/Documents/Queries/QueryResultBase.cs
Danielle9897: moved `SpatialProperties` to `QueryResultServerSide.cs`

src/Raven.Server/Documents/Indexes/Index.cs
Danielle9897: Added a url param `addSpatialProperties` 
(similar to `addTimeSeriesNames` param)

src/Raven.Client/Documents/Queries/QueryResultBase.cs
arekpalinski: I like current impl much better.

src/Raven.Server/Documents/Indexes/Index.cs
arekpalinski: Great!

src/Raven.Server/Documents/Queries/QueryMetadata.cs
arekpalinski: What if do `select Foo.Bar.Baz.Been.Latitude as AliasName`? Won't this method fail then?

arekpalinski: Note  `propertyPath.IndexOf(".");`  might return `-1` if `"."` isn't found

src/Raven.Server/Documents/Queries/QueryMetadata.cs
```diff
@@ -2317,14 +2319,24 @@ public void HandleSpatial(string methodName, List<QueryExpression> arguments, bo
...
-                                latitude,
-                                longitude
+                                latitudePropertyPath,
+                                longitudePropertyPath
                             });
+
+                            if (_metadata.SelectFields != null)
+                            {
+                                latitudePropertyPath = GetPathWithAliasName(latitudePropertyPath);
+                                longitudePropertyPath = GetPathWithAliasName(longitudePropertyPath);
```
ayende: Aren't you changing them here _after_ the were already included in the `fieldOptions` ?

src/Raven.Server/Documents/Queries/QueryMetadata.cs
Danielle9897: refactored code - works now for any variation:
`select Foo.Bar.Baz.Bee.Latitude as AliasName`
`select Foo.Bar.Baz as AliasName`
`select Foo.Bar`
etc.

src/Raven.Server/Documents/Queries/QueryMetadata.cs
Danielle9897: pls see latest commit - code was refactored
either way - I'm not changing the `latitudePropertyPath` & `longitudePropertyPath `
I'm just getting the relevant path for the _metadata which in turn is used by the studio to locate the geodata in the resulting document.

arekpalinski:
@Danielle9897 Please rebase it against latest v5.1 branch so tests will manage to run.

Danielle9897:
rebased on latest 5.1 & forced pushed

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
```diff
@@ -0,0 +1,142 @@
+import viewModelBase = require("viewmodels/viewModelBase");
+import spatialMarkersLayerModel = require("models/database/query/spatialMarkersLayerModel");
+import document = require("models/database/documents/document");
+import documentMetadata = require("models/database/documents/documentMetadata");
+import {Control, LatLngExpression, Layer, MarkerClusterGroup} from "leaflet";
```
ml054, line 5: please make sure this works when you build project with build.ps1 script.

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
```diff
@@ -0,0 +1,142 @@
...
+        this.polygonsLayer(polygons);
+    }
+
+    compositionComplete() {
+        super.compositionComplete();
+        this.createMap();
+    }
+    
+    private createMap() {
+
```
ml054: remove empty line

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
```diff
@@ -0,0 +1,142 @@
...
+        const osmMap = this.getStreetMapTileLayer();
+        const otmMap = this.getTopographyMapTileLayer();
+        const baseLayers =  {
```
ml054: remove double spaces

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
```diff
@@ -0,0 +1,142 @@
...
+                                    <hr>
+                                    <pre>${text}</pre>
+                              </div>`;
+            return textHtml;
+        }
+        
+        const dataLayers: Control.LayersObject = {};
+        const markersGroups: MarkerClusterGroup[] = [];
+
+        this.markersLayers().forEach((markersLayer) => {
```
ml054: remove ( )

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
```diff
@@ -0,0 +1,142 @@
...
+            const markers: Layer[] = [];
+            const markersGroup = L.markerClusterGroup();
+
+            markersLayer.geoPoints().forEach((point) => {
```
ml054: remove ( )

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
```diff
@@ -0,0 +1,142 @@
...
+                const pointMarker = L.marker([point.Latitude, point.Longitude], { title: point.PopupContent.getId() })
```
ml054: double quotes

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
```diff
@@ -0,0 +1,142 @@
...
+                                     .bindPopup(generatePopupContent(point.PopupContent), { 'className' : 'custom-popup', 'maxWidth': 600, 'maxHeight': 400 } );
```
ml054: you generate popup content from popup content? I think we should have better name for PopupContent

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
ml054: can we avoid those foreach and use map instead?

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
```diff
@@ -0,0 +1,142 @@
...
+                                          <span>${genUtils.escapeHtml(markersLayer.latitudeProperty)}</span><br>
+                                          <span>${genUtils.escapeHtml(markersLayer.longitudeProperty)}</span>
+                                      </div>`;
+
+            // A markersGroup layer is per spatial point from query
+            (dataLayers as any)[controlLayerText] = markersGroup;
+            markersGroups.push(markersGroup);
+        })
+        
+        const polyArray: Layer[] = [];
```
ml054: can we use map?

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
```diff
@@ -0,0 +1,142 @@
...
+        for (let i = 0; i < this.polygonsLayer().length; i++) {
+            const poly = this.polygonsLayer()[i];
+            const polyItem = L.polygon(poly.vertices,
+                { color: spatialPolygonModel.colors[i % spatialPolygonModel.colors.length] });
+            polyArray.push(polyItem);
+        }
+
+        const circleArray: Layer[] = [];
```
ml054: same here

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
```diff
@@ -0,0 +1,142 @@
...
+        L.control.layers(baseLayers, dataLayers).addTo(map);
+        
+        markersGroups.forEach(group => map.addLayer(group));
+        map.addLayer(polyLayer);
+        map.addLayer(circleLayer);
+
+        const mapBounds = L.latLngBounds([]);
+        markersGroups.forEach(group => mapBounds.extend(group.getBounds()));
+       
+        polyArray.forEach(poly => mapBounds.extend((poly as L.Polygon).getBounds()));
```
ml054: do we need cast here? Maybe both methods has getBounds method?

test/SlowTests/Tests/Spatial/DynamicSpatialQueries.cs
```diff
@@ -0,0 +1,447 @@
...
+namespace SlowTests.Tests.Spatial
+{
+    public class DynamicSpatialQueries : RavenTestBase
+    {
+        public DynamicSpatialQueries(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        [Fact]
+        public void VerifySingleSpatialPropertyInResultsTest()
```
ml054: do we need word test at the end? I think most test cases don't have that.

Danielle9897: oh - I thought that this is a convention - I can remove - ok done

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
Danielle9897: `generatePopupContent` => `generatePopupHtml`

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/database/query/spatialQueryMap.ts
Danielle9897: done

ml054 (approved):

PR #11537, opened 2020-12-10
RavenDB-16001 Fix certificates view - support the case when a client …

Danielle9897:
…certificate is not provided


src/Raven.Studio/typescript/viewmodels/manage/certificates.ts
```diff
@@ -415,7 +415,10 @@ class certificates extends viewModelBase {
                 const serverCert = mergedCertificates.find(x => _.includes(x.Thumbprints, this.serverCertificateThumbprint()));
                 const clientCert = mergedCertificates.find(x => _.includes(x.Thumbprints, this.clientCertificateThumbprint()));
                 const orderedCertificates = mergedCertificates.filter(x => x !== serverCert && x !== clientCert);
-                orderedCertificates.unshift(clientCert);
+                
+                if(!!clientCert) {
```
ml054: please format code and no need to use !! in condition

Danielle9897: right - done

ml054 (approved):

PR #11553, opened 2020-12-14
RavenDB-15998 handle cluster tx conflicts and test fixes

karmeli87:


src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1773,6 +1776,37 @@ public void EnsureLastEtagIsPersisted(DocumentsOperationContext context, long do
             return (newEtag, changeVector);
         }
 
+        public static bool FlagsProperlySet(DocumentFlags flags, string changeVector, out string reason)
```
ppekrol: I'm worried that in DEBUG we will be doing that. What is the slowdown?

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1773,6 +1776,37 @@ public void EnsureLastEtagIsPersisted(DocumentsOperationContext context, long do
...
+            var cvArray = changeVector.ToChangeVector();
+
+            if (flags.Contain(DocumentFlags.FromClusterTransaction))
+            {
+                if (cvArray.Length != 1)
+                {
+                    reason = $"FromClusterTransaction, expect change vector of length 1, {changeVector}";
+                    return false;
+                }
+                if (cvArray[0].NodeTag != "RAFT".ParseNodeTag())
```
ppekrol: `RAFT` looks like a constant. Don't we have a place for those?

ppekrol: Is `ParseNodeTag` for each call necessary?

src/Raven.Server/Documents/DocumentsStorage.cs
ayende: Yes, we want that with `Conditional`, I think

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
```diff
@@ -1418,6 +1418,18 @@ protected override long ExecuteCmd(DocumentsOperationContext context)
                                             _replicationInfo.Logger.Info(
                                                 $"Conflict check resolved to Conflict operation, resolving conflict for doc = {item.Id}, with change vector = {item.ChangeVector}");
 
+                                        if (hasLocalClusterTx == hasRemoteClusterTx)
```
ayende, line 1421: Can this be that both are false here?

src/Raven.Server/Documents/DocumentsStorage.cs
ppekrol: Sure, but we are using this in Debug only, I'm worried that this can get us severe slowdown during Debug tests, but maybe I'm not right

src/Raven.Server/Documents/Replication/IncomingReplicationHandler.cs
karmeli87: yes, both false is the common case, normal conflict

src/Raven.Server/Documents/DocumentsStorage.cs
karmeli87: I don't think there is a severe slowdown here... 
I guess the most consuming part is parsing the change-vector and most of our tests has a change-vector of 1 or 2 entries.

PR #11556, opened 2020-12-14
RavenDB-15930: Long startup time with large transactions

garayx:


grisha-kotler (approved):

ayende (approved):

src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -186,6 +186,7 @@ public bool RecoverDatabase(TransactionHeader* txHeader, Action<string> addToIni
                 journalToStartReadingFrom == -1)
                 journalToStartReadingFrom++;
 
+            var recycleLastJournal = false;
```
arekpalinski: This isn't actually about recycling it about it indicates if current (last) journal should be _deleted_ or not. Unless I don't follow the code correctly.

src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -394,14 +402,15 @@ private void CleanupNewerInvalidJournalFiles(long lastSyncedJournal)
             }
         }
 
-        private void RecoverCurrentJournalSize(AbstractPager pager)
+        private bool RecoverCurrentJournalSizeAndCheckIfLargerThanMaxLogFileSize(AbstractPager pager)
```
arekpalinski: Instead of returning `bool` let's return it as `out wasMoreThanMaxFileSize` parameter. This way it'll obvious what is it about. Currently since `Recover...` method returns `true/false` it might indicate that it failed when returning `false`.

src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -336,6 +339,11 @@ public bool RecoverDatabase(TransactionHeader* txHeader, Action<string> addToIni
                         _journalApplicator.AddJournalToDelete(journalFile);
                         toDelete.Add(journalFile);
                     }
+                    else if (recycleLastJournal)
```
arekpalinski: Shouldn't we also verify that this is really last (or current) journal by checking it's number? Also some debug assertion live above?

src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -196,7 +197,9 @@ public bool RecoverDatabase(TransactionHeader* txHeader, Action<string> addToIni
                     using (var recoveryPager = _env.Options.CreateTemporaryBufferPager(journalRecoveryName, initialSize))
                     using (var pager = _env.Options.OpenJournalPager(journalNumber, logInfo))
                     {
-                        RecoverCurrentJournalSize(pager);
+                        var isMoreThanMaxFileSize = RecoverCurrentJournalSizeAndCheckIfLargerThanMaxLogFileSize(pager);
+                        if (journalNumber == logInfo.CurrentJournal)
```
arekpalinski, line 201: At this point we're sure that `CurrentJournal` is the last one, right? Maybe add some debug assertion which will verify that.

src/Voron/StorageEnvironmentOptions.cs
```diff
@@ -572,6 +572,14 @@ private void AttemptToReuseJournal(VoronPathSetting desiredPath, long desiredSiz
 
                         try
                         {
+                            var journalSize = new FileInfo(filename).Length;
+                            if (journalSize > MaxLogFileSize && desiredSize <= MaxLogFileSize)
```
arekpalinski: 1. There is `File.Exists(filename)` below - should we check it first?
2. You can create `new FileInfo(filename)` instance and use its methods here - `Exists` or `MoveTo`.
3. Can we simplify the condition to `journalSize > MaxLogFileSize && desiredSize > journalSize ` - I think it's easier to read.

arekpalinski: 4. Just to make sure - technically we could just skip adding a journal to `_journalsForReuse` list if it's too large. But I think you didn't do it on purpose because next attempt to create a journal could also request such big size. Is that correct?

src/Voron/Impl/Journal/WriteAheadJournal.cs
garayx: yes we are sure because of the for clause
https://github.com/ravendb/ravendb/blob/7096a085e174f62f001d6cfc5df95f68186e1392/src/Voron/Impl/Journal/WriteAheadJournal.cs#L190

src/Voron/Impl/Journal/WriteAheadJournal.cs
garayx: `recycleLastJournal` is assigned iff currecnt journal number is `logInfo.CurrentJournal` which is the last journal numbeer

src/Voron/StorageEnvironmentOptions.cs
garayx: 1. done
2. done
3. its not equal to current condition
4. we want to keep the big journals in case we have big tx in future, as well I wanted to do minimal changes to current journal recycle behavior

src/Voron/Impl/Journal/WriteAheadJournal.cs
arekpalinski: okay

src/Voron/Impl/Journal/WriteAheadJournal.cs
arekpalinski: Ok. I see you added Debug.Assert so we're good

ppekrol (approved):

src/Voron/StorageEnvironmentOptions.cs
arekpalinski: 3. OK. I understand it after your clarification in 4.

arekpalinski (approved):

PR #11560, opened 2020-12-15
 RavenDB-16022 We can't assume that when we call DirectAdd on a value…

arekpalinski:
…, the data we previously read will not be modified.

https://issues.hibernatingrhinos.com/issue/RavenDB-16022


ayende (commented):
How did we run into this?

ayende (approved):

ayende (commented):
Okay, at least only in tests...

ppekrol (approved):

PR #11561, opened 2020-12-15
RavenDB-16020 & RavenDB-16021

redknightlois:
The changes revolve around diminishing the IO impact of unbuffered access (in the case of Voron-Lucene's writing process) and improving the sync behavior and heuristic to make it more predictable. The old policy would suggest a Sync after every flush, the new policy will under low load wait by default 120 secs and under load after a certain amount of bytes are ready to be sync. In writing heavy environments waiting to sync larger amounts is better. The sync behavior doesn't show a measurable impact on IO starved systems or in the indexing process (which is more compute-bound) but it does on high-performance bulk inserting systems.

### DIFF
--- a/src/Raven.Server/Config/Categories/StorageConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/StorageConfiguration.cs
@@ -91,6 +91,12 @@ namespace Raven.Server.Config.Categories
         [ConfigurationEntry("Storage.SyncJournalsCountThreshold", ConfigurationEntryScope.ServerWideOrPerDatabase)]
         public int SyncJournalsCountThreshold { get; set; }
 
+        [Description("Max amount of unsynchronized bytes before requesting a sync and removing unused journal files.")]
+        [DefaultValue(256)]
+        [SizeUnit(SizeUnit.Megabytes)]
+        [ConfigurationEntry("Storage.SyncJournalsMaxSizeInMegabytesThreshold", ConfigurationEntryScope.ServerWideOrPerDatabase)]
+        public Size SyncJournalsMaxSizeInMegabytesThreshold { get; set; }
+
         /// <summary>
         /// Specifies the time interval between each IoMetrics Cleaner run
         /// </summary>
--- a/src/Raven.Server/Indexing/VoronIndexOutput.cs
+++ b/src/Raven.Server/Indexing/VoronIndexOutput.cs
@@ -55,7 +55,7 @@ namespace Raven.Server.Indexing
                 if (options.Encryption.IsEnabled)
                     return new TempCryptoStream(_fileTempPath, ignoreSetLength: true);
 
-                return SafeFileStream.Create(_fileTempPath, FileMode.CreateNew, FileAccess.ReadWrite, FileShare.ReadWrite, 4096, FileOptions.DeleteOnClose);
+                return SafeFileStream.Create(_fileTempPath, FileMode.CreateNew, FileAccess.ReadWrite, FileShare.ReadWrite, Voron.Global.Constants.Size.Megabyte, FileOptions.DeleteOnClose);
             }
             catch (IOException ioe) when (ioe.IsOutOfDiskSpaceException())
             {
--- a/src/Voron/Impl/Journal/WriteAheadJournal.cs
+++ b/src/Voron/Impl/Journal/WriteAheadJournal.cs
@@ -579,7 +579,7 @@ namespace Voron.Impl.Journal
             }
 
 
-            public void ApplyLogsToDataFile(CancellationToken token, TimeSpan timeToWait)
+            public void ApplyLogsToDataFile(CancellationToken token, TimeSpan timeToWait, bool suggestSync = true)
             {
                 if (token.IsCancellationRequested)
                     return;
@@ -706,7 +706,8 @@ namespace Voron.Impl.Journal
 
                     ApplyJournalStateAfterFlush(token, lastProcessedJournal, lastFlushedTransactionId, unusedJournals, byteStringContext);
 
-                    _waj._env.SuggestSyncDataFile();
+                    if (suggestSync)
+                        _waj._env.SuggestSyncDataFile();
                 }
                 finally
                 {
--- a/src/Voron/StorageEnvironment.cs
+++ b/src/Voron/StorageEnvironment.cs
@@ -122,6 +122,8 @@ namespace Voron
 
         public bool IsNew { get; }
 
+        public Size SyncJournalsMaxSizeInMegabytesThreshold { get; private set; }
+
         public StorageEnvironment(StorageEnvironmentOptions options)
         {
             try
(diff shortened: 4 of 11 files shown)


src/Raven.Server/Config/Categories/StorageConfiguration.cs
```diff
@@ -91,6 +90,12 @@ public class StorageConfiguration : ConfigurationCategory
         [ConfigurationEntry("Storage.SyncJournalsCountThreshold", ConfigurationEntryScope.ServerWideOrPerDatabase)]
         public int SyncJournalsCountThreshold { get; set; }
 
+        [Description("Number of Journals files for each storage, before forcing a sync and removing unused journal files.")]
```
arekpalinski: Wrong description I think

src/Voron/StorageEnvironment.cs
```diff
@@ -196,14 +199,12 @@ private async Task IdleFlushTimer()
...
-                        else if (Journal.Applicator.TotalWrittenButUnsyncedBytes != 0)
-                            SuggestSyncDataFile();
-                    }
-                    else
-                    {
-                        await TimeoutManager.WaitFor(TimeSpan.FromMilliseconds(1000), cancellationToken).ConfigureAwait(false);
                     }
+
+                    if (Journal.Applicator.TotalWrittenButUnsyncedBytes > SyncJournalsMaxSizeInMegabytesThreshold.GetValue(SizeUnit.Bytes))
+                        SuggestSyncDataFile();
```
arekpalinski, line 205: But now we might be suggesting data sync when we have a running write transaction at the same time, don't we? Is it what we want? Note that this is `IdleFlushTimer`

src/Raven.Server/Config/Categories/StorageConfiguration.cs
```diff
@@ -40,10 +40,9 @@ public class StorageConfiguration : ConfigurationCategory
         public int MaxConcurrentFlushes { get; set; }
 
         [Description("Time to sync after flash in seconds")]
-        [DefaultValue(30)]
+        [DefaultValue(120)]
         [TimeUnit(TimeUnit.Seconds)]
-        [ConfigurationEntry("Storage.TimeToSyncAfterFlashInSec", ConfigurationEntryScope.ServerWideOrPerDatabase)]
-        [ConfigurationEntry("Storage.TimeToSyncAfterFlushInSec", ConfigurationEntryScope.ServerWideOrPerDatabase)]
```
arekpalinski: We had 2 separate option on purpose. Probably someone make a typo 'flash' instead of 'flush' initially so we fixed it by adding _another_ entry here. But we can't remove the previous one because of backward compatibility.

arekpalinski (commented):
Can you share the perf results after introducing those changes?

src/Raven.Server/Indexing/VoronIndexOutput.cs
```diff
@@ -55,7 +55,7 @@ private Stream InitFileStream(StorageEnvironmentOptions options)
                 if (options.Encryption.IsEnabled)
                     return new TempCryptoStream(_fileTempPath, ignoreSetLength: true);
 
-                return SafeFileStream.Create(_fileTempPath, FileMode.CreateNew, FileAccess.ReadWrite, FileShare.ReadWrite, 4096, FileOptions.DeleteOnClose);
+                return SafeFileStream.Create(_fileTempPath, FileMode.CreateNew, FileAccess.ReadWrite, FileShare.ReadWrite, Voron.Global.Constants.Size.Megabyte, FileOptions.DeleteOnClose);
```
ayende, line 58: I think that this may be a bad idea, see:
https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/IO/FileStream.cs#L749

The issue is that we are creating a 1MB managed buffer on each such call.

src/Voron/StorageEnvironment.cs
ayende: Yes, I think this should be inside the if.

src/Raven.Server/Indexing/VoronIndexOutput.cs
ayende: Relevant: https://ayende.com/blog/171875/production-postmortem-the-case-of-the-memory-eater-and-high-load

src/Raven.Server/Config/Categories/StorageConfiguration.cs
redknightlois: Ahh OK

src/Raven.Server/Indexing/VoronIndexOutput.cs
redknightlois: Are these guys serious??? Yes, it is a very bad idea. Now the question is how many times that is being called.

src/Voron/StorageEnvironment.cs
redknightlois: That would never happen under stress. In fact, it was the reason why the `Sync` would happen without any real heuristic. After every `Flush` we were issuing a `Sync` request that would hammer the IO.

src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -691,7 +691,8 @@ public void ApplyLogsToDataFile(CancellationToken token, TimeSpan timeToWait)
 
                     ApplyJournalStateAfterFlush(token, lastProcessedJournal, lastFlushedTransactionId, unusedJournals, byteStringContext);
 
-                    _waj._env.SuggestSyncDataFile();
```
redknightlois, line 709: @arekpalinski @ayende  This was the reason why actually the heuristic was running even if there was a write transaction in place immediately after a flush. As long as there was at least 1 full journal it would apply the Sync. What I did is to make the `Sync` predictable so we can actually apply a heuristic, we can decide on a different one current is very very naive, it will initiate no matter what as soon as there are more written bytes than 256Mb on the data file.

src/Raven.Server/Config/Categories/StorageConfiguration.cs
```diff
@@ -91,6 +91,12 @@ public class StorageConfiguration : ConfigurationCategory
         [ConfigurationEntry("Storage.SyncJournalsCountThreshold", ConfigurationEntryScope.ServerWideOrPerDatabase)]
         public int SyncJournalsCountThreshold { get; set; }
 
+        [Description("Max amount of unsynchronized bytes before requesting a sync and removing unused journal files.")]
+        [DefaultValue(256)]
+        [SizeUnit(SizeUnit.Megabytes)]
+        [ConfigurationEntry("Storage.SyncJournalsMaxSizeInMegabytesThreshold", ConfigurationEntryScope.ServerWideOrPerDatabase)]
```
ppekrol, line 97: Consistency & conventions: 
- `Storage.SyncJournalsMaxSizeThresholdInMb` for ConfigurationEntry
- `SyncJournalsMaxSizeThreshold` for property name

src/Voron/StorageEnvironment.cs
```diff
@@ -122,6 +122,8 @@ public static IDisposable GetStaticContext(out ByteStringContext ctx)
 
         public bool IsNew { get; }
 
+        public Size SyncJournalsMaxSizeInMegabytesThreshold { get; private set; }
```
ppekrol, line 125: SyncJournalsMaxSizeThreshold

src/Voron/StorageEnvironmentOptions.cs
```diff
@@ -182,6 +182,8 @@ public long MaxScratchBufferSize
             }
         }
 
+        public Size SyncJournalsMaxSizeInMegabytesThreshold { get; set; }
```
ppekrol, line 185: SyncJournalsMaxSizeThreshold

PR #11565, opened 2020-12-16
Make replication informer async

danielmarbach:
Part of #11543

### DIFF
--- a/Raven.Client.Lightweight/Connection/ReplicationInformer.cs
+++ b/Raven.Client.Lightweight/Connection/ReplicationInformer.cs
@@ -82,7 +82,10 @@ namespace Raven.Client.Connection
                     return;
 
                 if (refreshReplicationInformationTask != null)
+                {
                     await refreshReplicationInformationTask.ConfigureAwait(false);
+                    return;
+                }
 
                 refreshReplicationInformationTask = RefreshReplicationInformationInternalAsync(url, getReplicationDestinations).
                     ContinueWith(task =>
--- a/Raven.Client.Lightweight/Connection/ReplicationInformer.cs
+++ b/Raven.Client.Lightweight/Connection/ReplicationInformer.cs
@@ -5,7 +5,6 @@
 // -----------------------------------------------------------------------
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Raven.Abstractions;
@@ -108,28 +107,34 @@ namespace Raven.Client.Connection
         public override void UpdateReplicationInformationFromDocument(JsonDocument document)
         {
             var replicationDocument = document.DataAsJson.JsonDeserialization<ReplicationDocumentWithClusterInformation>();
-            ReplicationDestinations = replicationDocument.Destinations.Select(x =>
-            {
-                var url = string.IsNullOrEmpty(x.ClientVisibleUrl) ? x.Url : x.ClientVisibleUrl;
-                if (string.IsNullOrEmpty(url))
-                    return null;
-                if (x.CanBeFailover() == false)
-                    return null;
-                if (string.IsNullOrEmpty(x.Database))
-                    return new OperationMetadata(url, x.Username, x.Password, x.Domain, x.ApiKey, x.ClusterInformation);
+            var destinations = new List<OperationMetadata>(replicationDocument.Destinations.Count);
 
-                return new OperationMetadata(
-                    MultiDatabase.GetRootDatabaseUrl(url) + "/databases/" + x.Database + "/",
-                    x.Username,
-                    x.Password,
-                    x.Domain,
-                    x.ApiKey,
-                    x.ClusterInformation);
-            })
+            foreach (var destination in replicationDocument.Destinations)
+            {
+                OperationMetadata operationMetadata;
+                var url = string.IsNullOrEmpty(destination.ClientVisibleUrl) ? destination.Url : destination.ClientVisibleUrl;
                 // filter out replication destination that don't have the url setup, we don't know how to reach them
                 // so we might as well ignore them. Probably private replication destination (using connection string names only)
-                .Where(x => x != null)
-                .ToList();
+                if (string.IsNullOrEmpty(url))
+                    continue;
+                if (destination.CanBeFailover() == false)
+                    continue;
+                if (string.IsNullOrEmpty(destination.Database))
+                    operationMetadata = new OperationMetadata(url, destination.Username, destination.Password, destination.Domain, destination.ApiKey, destination.ClusterInformation);
+
+                operationMetadata = new OperationMetadata(
+                    MultiDatabase.GetRootDatabaseUrl(url) + "/databases/" + destination.Database + "/",
+                    destination.Username,
+                    destination.Password,
+                    destination.Domain,
+                    destination.ApiKey,
+                    destination.ClusterInformation);
+
+                destinations.Add(operationMetadata);
+            }
+
+            ReplicationDestinations = destinations;
+
             foreach (var replicationDestination in ReplicationDestinations)
             {
                 FailureCounter value;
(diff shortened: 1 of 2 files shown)


Raven.Client.Lightweight/Connection/ReplicationInformer.cs
```diff
@@ -44,22 +44,29 @@ public ReplicationInformer(DocumentConvention conventions, HttpJsonRequestFactor
...
                 if (lastReplicationUpdate.Add(Conventions.TimeToWaitBetweenReplicationTopologyUpdates) > SystemTime.UtcNow)
-                    return new CompletedTask();
+                    return;
             }
-            lock (replicationLock)
+
+            try
             {
+                // Do a quick synchronous check before we resort to async/await with the state-machine overhead.
+                if (!replicationLock.Wait(0))
```
danielmarbach, line 68: Previously this acquired a replication lock object while `RefreshReplicationInformationInternal` acquired `lock(this)` so I assumed it is OK to assume only one operation is ever allowed into this critical section. Please advise

Raven.Client.Lightweight/Connection/ReplicationInformer.cs
```diff
@@ -72,20 +79,22 @@ private Task UpdateReplicationInformationIfNeededInternalAsync(string url, Func<
...
 
-                var taskCopy = refreshReplicationInformationTask;
-                if (taskCopy != null)
-                    return taskCopy;
+                if (refreshReplicationInformationTask != null)
+                    await refreshReplicationInformationTask.ConfigureAwait(false);
 
-                return refreshReplicationInformationTask = Task.Factory.StartNew(() => 
-                    RefreshReplicationInformationInternal(url, getReplicationDestinations)).
+                refreshReplicationInformationTask = RefreshReplicationInformationInternalAsync(url, getReplicationDestinations).
```
danielmarbach, line 102: If I understood the code correctly the rationale is to do the heavy lifting in the background but keep track of it so I kept that consistent

Raven.Client.Lightweight/Connection/ReplicationInformer.cs
```diff
@@ -72,20 +79,22 @@ private Task UpdateReplicationInformationIfNeededInternalAsync(string url, Func<
...
 
                 if (lastReplicationUpdate.Add(Conventions.TimeToWaitBetweenReplicationTopologyUpdates) > SystemTime.UtcNow && force == false)
-                    return new CompletedTask();
+                    return;
 
-                var taskCopy = refreshReplicationInformationTask;
-                if (taskCopy != null)
-                    return taskCopy;
+                if (refreshReplicationInformationTask != null)
+                    await refreshReplicationInformationTask.ConfigureAwait(false);
```
danielmarbach: If I understood correctly next refresh should await the previous refresh task if available and then return to not trigger subsequent refreshes

Raven.Client.Lightweight/Connection/ReplicationInformer.cs
```diff
@@ -140,26 +155,34 @@ protected override string GetServerCheckUrl(string baseUrl)
...
-            RefreshReplicationInformationInternal(serverClient.Url, () => AsyncHelpers.RunSync(() => serverClient.DirectGetReplicationDestinationsAsync(new OperationMetadata(serverClient.Url, serverClient.PrimaryCredentials, null), null)));
+            AsyncHelpers.RunSync(() =>
+                RefreshReplicationInformationInternalAsync(serverClient.Url, () => serverClient.DirectGetReplicationDestinationsAsync(new OperationMetadata(serverClient.Url, serverClient.PrimaryCredentials, null), null)));
         }
 
         public override void RefreshReplicationInformation(ServerClient serverClient)
         {
-            RefreshReplicationInformationInternal(serverClient.Url, () => serverClient.DirectGetReplicationDestinations(new OperationMetadata(serverClient.Url, serverClient.PrimaryCredentials, null), null));
+            AsyncHelpers.RunSync(() =>
+                RefreshReplicationInformationInternalAsync(serverClient.Url, () => Task.FromResult(serverClient.DirectGetReplicationDestinations(new OperationMetadata(serverClient.Url, serverClient.PrimaryCredentials, null), null))));
```
danielmarbach, line 180: Wasn't entirely sure here whether you would prefer to offload `serverClient.DirectGetReplicationDestinations` or not. Given the sync nature of this code I went straight to not offload. Please advise

Raven.Client.Lightweight/Connection/ReplicationInformer.cs
ayende: Yes

Raven.Client.Lightweight/Connection/ReplicationInformer.cs
ayende: Yes, looks okay

ayende (approved):

danielmarbach:
By the way, this has slightly more allocation costs due to more async machinery involved. Yet I think this is negligible in comparison to the previously sync over async problem this code had. In addition to that, I removed the LINQ overhead in this call to compensate allocs. That being said I haven't measured the before and after nor would I know how to properly isolate this to measure other than doing long-running profiling sessions which would be a significant effort on my side.

grisha-kotler (approved):

danielmarbach:
Rebased

@grisha-kotler @ayende @ppekrol anything that hinders this from being merged?

ppekrol:
Hi @danielmarbach 

We have scheduled 5 CI builds and all of them got stuck. We are looking into that at the moment.

ayende:
Looks like we are hanging here: 
image

ayende:
This is pretty consistently getting stuck there. I took a brief look and it looks like the `refreshReplicationInformationTask` is never completed. It is possible that this is related to `SemaphoreSlim` usage, but I'm not going to dive too deeply into this.

This command reproduce this quite easily:  

` C:\Work\ravendb-3.5\Raven.Tests> ..\SharedLibs\xunit\xunit.console.clr4.exe .\bin\Debug\Raven.Tests.dll`

I don't see much benefit in this change at this time. This is happening on an infrequent basis and shouldn't have much of an impact.

ayende:
You also need to handle disposed cancellation token, but after fixing that, we have hangs here. 
I don't think that this is a change that we want to push forward, given the critical nature of the location and the impact.


```
 	Raven.Abstractions.dll!Raven.Abstractions.Util.AsyncHelpers.ExclusiveSynchronizationContext.BeginMessageLoop() Line 153	C#
 	Raven.Abstractions.dll!Raven.Abstractions.Util.AsyncHelpers.RunSync(System.Func<System.Threading.Tasks.Task> task) Line 45	C#
 	Raven.Client.Lightweight.dll!Raven.Client.Connection.AdminServerClient.EnsureDatabaseExists(string name, bool ignoreFailures) Line 38	C#
 	Raven.Client.Lightweight.dll!Raven.Client.Document.DocumentStore.Initialize(bool ensureDatabaseExists) Line 472	C#
 	Raven.Tests.Helpers.dll!Raven.Tests.Helpers.RavenTestBase.NewRemoteDocumentStore(bool fiddler, Raven.Server.RavenDbServer ravenDbServer, string databaseName, bool runInMemory, string dataDirectory, string requestedStorage, bool enableAuthentication, bool ensureDatabaseExists, System.Action<Raven.Client.Document.DocumentStore> configureStore, string activeBundles, System.Collections.Generic.IEnumerable<Raven.Client.Indexes.AbstractIndexCreationTask> indexes, System.Collections.Generic.IEnumerable<Raven.Client.Indexes.AbstractTransformerCreationTask> transformers, System.Collections.Generic.IEnumerable<System.Collections.IEnumerable> seedData, bool noStaleQueries) Line 358	C#
>	Raven.Tests.dll!Raven.Tests.Faceted.LazyFacets.Default_operator_not_honoured_remote_store_ToFacetsLazy() Line 22	C#

```

danielmarbach:
Fair enough

danielmarbach:
I do think though some of those hangs could have been mitigated opting out from context capturing with

```
            using (NoSynchronizationContext.Scope())
            {
```

Pretty sure the reason of these hangs is XUnit custom context.

PR #11577, opened 2020-12-17
RavenDB-15930: Add condition to debug assert clause

garayx:


src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -363,8 +363,9 @@ public bool RecoverDatabase(TransactionHeader* txHeader, Action<string> addToIni
 #if DEBUG
                 if (instanceOfLastFlushedJournal == null)
                 {
-                    Debug.Assert(toDelete.Count == 0, $"Last flushed journal (number: {lastFlushedJournal}) doesn't exist so we didn't call {nameof(_journalApplicator.SetLastFlushed)} but" +
-                                                      $" there are still some journals to delete ({string.Join(", ", toDelete.Select(x => x.Number))}. )");
+                    Debug.Assert(toDelete.Count == 0 || deleteLastJournal, $"Last flushed journal (number: {lastFlushedJournal}) doesn't exist so we didn't call {nameof(_journalApplicator.SetLastFlushed)}" +
```
garayx: the journal doesn't get flushed when its empty or it's `lastReadHeaderPtr == null`, but we still want to delete such journals if their size if bigger than `MaxLogFileSize`

PR #11578, opened 2020-12-17
Remote changes base

danielmarbach:
A demonstration how #11543 could be solved

### DIFF
--- a/Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
+++ b/Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
@@ -105,7 +105,7 @@ namespace Raven.Client.Changes
         {
             foreach (var value in Counters.ValuesSnapshot)
             {
-                value.Task.Wait();
+                value.Task.GetAwaiter().GetResult();
             }
         }
 
--- a/Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
+++ b/Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
@@ -73,7 +73,7 @@ namespace Raven.Client.Changes
                         {
                             task.AssertNotFailed();
                             return this as TChangesApi;
-                        });
+                        }, CancellationToken.None, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
         }
 
         protected TConventions Conventions { get; private set; }
@@ -275,7 +275,7 @@ namespace Raven.Client.Changes
                     {
                         logger.ErrorException("Got error from server connection for " + url + " on id " + id, e);
                     }
-                });
+                }, CancellationToken.None, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
         }
 
 
@@ -335,7 +335,7 @@ namespace Raven.Client.Changes
                         value.Error(task.Exception);
                     }
                     Counters.Clear();
-                });
+                }, CancellationToken.None, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
         }
 
         public void OnNext(string dataFromConnection)
--- a/Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
+++ b/Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
@@ -257,7 +257,7 @@ namespace Raven.Client.Changes
             }
         }
 
-        struct Work : IDisposable
+        sealed class Work : IDisposable
         {
             readonly HttpJsonRequest request;
             public readonly TaskCompletionSource<bool> DoneTask;
--- a/Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
+++ b/Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
@@ -30,7 +30,7 @@ namespace Raven.Client.Changes
         private readonly string url;
         private readonly OperationCredentials credentials;
         private readonly HttpJsonRequestFactory jsonRequestFactory;
-        private readonly Action onDispose;                
+        private readonly Action onDispose;
 
         private IDisposable connection;
         private DateTime lastHeartbeat;
@@ -43,7 +43,7 @@ namespace Raven.Client.Changes
         private const int Idle = 0;
 
         // This is the StateCounters, it is not related to the counters database
-        protected readonly AtomicDictionary<TConnectionState> Counters = new AtomicDictionary<TConnectionState>(StringComparer.OrdinalIgnoreCase);        
+        protected readonly AtomicDictionary<TConnectionState> Counters = new AtomicDictionary<TConnectionState>(StringComparer.OrdinalIgnoreCase);
 
         protected RemoteChangesClientBase(
             string url,
@@ -65,7 +65,7 @@ namespace Raven.Client.Changes
             this.url = url;
             this.credentials = new OperationCredentials(apiKey, credentials);
             this.jsonRequestFactory = jsonRequestFactory;
-            this.onDispose = onDispose;            
+            this.onDispose = onDispose;
             Conventions = conventions;
             Task = EstablishConnection()
                         .ObserveException()
@@ -346,7 +346,7 @@ namespace Raven.Client.Changes
             var value = ravenJObject.Value<RavenJObject>("Value");
             var type = ravenJObject.Value<string>("Type");
             if (logger.IsDebugEnabled)
-            logger.Debug("Got notification from {0} id {1} of type {2}", url, id, dataFromConnection);
+                logger.Debug("Got notification from {0} id {1} of type {2}", url, id, dataFromConnection);
 
             switch (type)
             {
--- a/Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
+++ b/Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
@@ -243,7 +243,7 @@ namespace Raven.Client.Changes
                 {
                     try
                     {
-                        await work.SendTask.ConfigureAwait(false);
+                        await work.ExecuteAsync().ConfigureAwait(false);
                     }
                     catch (Exception)
                     {
@@ -261,14 +261,14 @@ namespace Raven.Client.Changes
         {
             readonly HttpJsonRequest request;
             public readonly TaskCompletionSource<bool> DoneTask;
-            public readonly Task SendTask;
 
-            public Work(TaskCompletionSource<bool> doneTask, HttpJsonRequest request, Task sendTask)
+            public Work(TaskCompletionSource<bool> doneTask, HttpJsonRequest request)
             {
                 DoneTask = doneTask;
                 this.request = request;
-                SendTask = sendTask;
             }
+
+            public Task ExecuteAsync() => request.ExecuteRequestAsync();
 
             public void Dispose()
             {
@@ -301,7 +301,7 @@ namespace Raven.Client.Changes
                     AvoidCachingRequest = true
                 };
                 var request = jsonRequestFactory.CreateHttpJsonRequest(requestParams);
-                workQueue.Enqueue(new Work(done, request, request.ExecuteRequestAsync()));
+                workQueue.Enqueue(new Work(done, request));
 
                 syncSource.TrySetResult(true);
             }
--- a/Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
+++ b/Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
@@ -278,30 +278,39 @@ namespace Raven.Client.Changes
 
         protected Task Send(string command, string value)
         {
-            logger.Info("Sending command {0} - {1} to {2} with id {3}", command, value, url, id);
-            var sendUrlBuilder = new StringBuilder();
-            sendUrlBuilder.Append(url);
-            sendUrlBuilder.Append("/changes/config?id=");
-            sendUrlBuilder.Append(id);
-            sendUrlBuilder.Append("&command=");
-            sendUrlBuilder.Append(command);
-
-            if (string.IsNullOrEmpty(value) == false)
+            try
             {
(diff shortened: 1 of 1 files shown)


Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
```diff
@@ -201,83 +213,133 @@ private bool DisposeHeartbeatTimer()
...
                 {
-                    return new CompletedTask(e).Task.ObserveException();
+                    try
+                    {
+                        await work.SendTask.ConfigureAwait(false);
+                        work.DoneTask.TrySetResult(true);
+                    }
+                    catch (Exception e)
+                    {
+                        work.DoneTask.TrySetException(e);
```
danielmarbach: This line could also be removed if the previous behavior of swalling exceptions is of interest/desired. Then the TrySet would need to move into a finally

Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
```diff
@@ -201,83 +213,133 @@ private bool DisposeHeartbeatTimer()
...
-                var sendTask = lastSendTask;
-                if (sendTask != null)
+                try
                 {
-                    return sendTask.ContinueWith(_ =>
-                    {
-                        Send(command, value);
-                    });
+                    await syncSource.Task.ConfigureAwait(false);
+                    syncSource = TaskCompletionSourceFactory.Create<bool>();
```
ayende, line 223: You are trying to do a manual reset event, no?
Better to use:
https://github.com/ravendb/ravendb/blob/v3.5/Raven.Abstractions/Util/AsyncManualResetEvent.cs

Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
ayende: The issue is the breaking change, I think

Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
```diff
@@ -201,83 +213,133 @@ private bool DisposeHeartbeatTimer()
...
+                    }
+                    catch (Exception e)
+                    {
+                        work.DoneTask.TrySetException(e);
+                    }
                 }
             }
         }
 
+        struct Work : IDisposable
```
ayende: Why is this a struct? It is copied around quite a lot, no?

Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
```diff
@@ -201,83 +213,133 @@ private bool DisposeHeartbeatTimer()
...
-                    }
-                    catch (Exception e)
-                    {
-                        logger.ErrorException("Got error from server connection for " + url + " on id " + id, e);
-                    }
-                });
-        }
+            try
+            {
+                await Send("disconnect", null).ConfigureAwait(false);
```
ayende, line 339: This can fail, need to try/catch this, no?

Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
```diff
@@ -201,83 +213,133 @@ private bool DisposeHeartbeatTimer()
...
 
+                tokenSource.Cancel();
+                // will drain and handle disconnect command
+                await worker.ConfigureAwait(false);
+
+                if (connection != null)
+                    connection.Dispose();
+
+                tokenRegistration.Dispose();
+                tokenSource.Dispose();
```
ayende: Needs to be in a finally

ayende (changes requested):

Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
ayende: Then again, I wonder how many people actually do wait on this

Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
```diff
@@ -201,83 +213,133 @@ private bool DisposeHeartbeatTimer()
...
+                    syncSource = TaskCompletionSourceFactory.Create<bool>();
                 }
-
-                try
+                catch (TaskCanceledException)
                 {
-                    var sendUrl = url + "/changes/config?id=" + id + "&command=" + command;
-                    if (string.IsNullOrEmpty(value) == false)
-                        sendUrl += "&value=" + Uri.EscapeUriString(value);
+                    // Swallowing the task cancellation in case we are stopping work.
```
grisha-kotler, line 229: Shouldn't you exit here and `SetException` for all tasks?

Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
```diff
@@ -201,83 +213,133 @@ private bool DisposeHeartbeatTimer()
...
-        }
+            try
+            {
+                await Send("disconnect", null).ConfigureAwait(false);
 
+                tokenSource.Cancel();
+                // will drain and handle disconnect command
+                await worker.ConfigureAwait(false);
+
+                if (connection != null)
```
grisha-kotler, line 352: Should be in the finally (only the `Send` can fail here)

grisha-kotler (changes requested):

Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
danielmarbach: Doesn't seem to use the TCS factory and therefore doesn't use RunContinuationsAsync

Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
danielmarbach: That's why I mentioned it. I can keep swallowing the exception

Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
danielmarbach: I first wanted to make this readonly struct but then realized this project lacks language support for that. Need to have another look

Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
danielmarbach: No why? I could add a break and then we don't try to empty twice. Judging by the disconnect in dispose we always want to complete all enqueued things right?

danielmarbach:
Addressed comments where I saw fit and rebased

ppekrol:
Hi @danielmarbach 

Following two tests are failing in this PR:
image

Could you take a look?

ppekrol:
@grisha-kotler Could you review?

danielmarbach:
I'm on vacation and I think I have done quite a bit of heavy lifting already. I gave edit rights on the PR. That being said I did test runs with this version of the code and not only do we get very nice throughput improvements due to less contention it also removes the starving problem entirely

danielmarbach:
Pretty sure the tests failing here are due to

```
            Task = EstablishConnection()
                        .ObserveException()
                        .ContinueWith(task =>
                        {
                            task.AssertNotFailed();
                            return this as TChangesApi;
                        }, CancellationToken.None, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
```

This change should be a non-problematic change but the problem is that the tests in questions are doing a very horrible sync over async

```
                    changes.ForDocument("users/1").Task.Result
                           .Subscribe(documentChangeNotification => canProceedEvent.Signal());

                    changes.ForAllIndexes().Task.Result
                           .Subscribe(indexChangeNotification => canProceedEvent.Signal());
```

which then in combination with the XUnit async context causes things to starve.

danielmarbach:
So given https://github.com/ravendb/ravendb/pull/11578#issuecomment-749235186 and that solving would require to asyncify all tests that use the Task exposed probably the best avenue right now to proceed with this is to remove the `TaskContinuationOptions.ExecuteSynchronously` on the connection task

danielmarbach:
Ok apparently there was always an issue with the code that the last unsubscribe will fail

image

That's why there was a very broad try/catch in there. Will add the same to preserve the behavior

danielmarbach:
I pushed a commit the restores the previous behavior although I think this is very fishy. It seems the order of dispose is somehow of and therefore the unwatching commands (`onZero`) will never be sent to the server. Given that this was a problem also before I decided to not further look into this problem and keep the behavior consistent by adding a try catch. I also used the `CompletedTask` to keep it more or less consistent with the rest of the code base

Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
```diff
@@ -201,83 +213,149 @@ private bool DisposeHeartbeatTimer()
...
+            }
 
-                    var requestParams = new CreateHttpJsonRequestParams(null, sendUrl, HttpMethods.Get, credentials, Conventions)
-                    {
-                        AvoidCachingRequest = true
-                    };
-                    var request = jsonRequestFactory.CreateHttpJsonRequest(requestParams);
-                    lastSendTask = request.ExecuteRequestAsync().ObserveException();
+            // drain
+            await WorkUntilEmpty().ConfigureAwait(false);
```
ayende, line 234: Need try catch for `TaskCancelledException` here, no?

Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
```diff
@@ -201,83 +213,149 @@ private bool DisposeHeartbeatTimer()
...
-                var sendTask = lastSendTask;
-                if (sendTask != null)
+                try
                 {
-                    return sendTask.ContinueWith(_ =>
-                    {
-                        Send(command, value);
-                    });
-                }
+                    await syncSource.Task.ConfigureAwait(false);
```
ayende, line 222: Do we need to handle `TaskCancelledException` here?

ayende (changes requested):
A key issue here is why do we even have this?
The whole point of the `Send` calling itself was to have a single concurrent request to the server.
What you did was change that it would allow multiple concurrent requests, then you do background work to process them.
But I don't understand why we even need to process them.
There is nothing that you actually *do* in the `Worker`. 

Am I missing someting?

ayende:
By the way, I don't think we *need* to have a single concurrent send, so that can simplify the code.

danielmarbach:
> 
> 
> A key issue here is why do we even have this?
> The whole point of the `Send` calling itself was to have a single concurrent request to the server.
> What you did was change that it would allow multiple concurrent requests, then you do background work to process them.
> But I don't understand why we even need to process them.
> There is nothing that you actually _do_ in the `Worker`.
> 
> Am I missing someting?

Yeah that was an oversight from my end. I moved now the actual execution into the worker itself. Then it would be aligned with the previous code.

danielmarbach:
> 
> 
> By the way, I don't think we _need_ to have a single concurrent send, so that can simplify the code.

If that is true I could have spent way less time on this by ripping out the lock and the self-calling to just directly fire&forget the send? Is that what you are hinting at?

Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
danielmarbach: `WorkUntilEmpty` doesn't observe the the token or the done task. So I don't think it is needed

Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
danielmarbach: Ctor does

```
tokenRegistration = tokenSource.Token.Register(() => syncSource.TrySetCanceled());
```

which cancels the the task represented by the sync source. Would you much rather prefer to do

```
tokenRegistration = tokenSource.Token.Register(() => syncSource.TrySetResult());
```

instead and remove the catch?

danielmarbach:
@ayende do you mean something like https://github.com/ravendb/ravendb/pull/11634 ? Or throwing in a semaphoreslim into the changes outlined in the alternative PR should the only be a single send allowed?

danielmarbach:
I've also added https://github.com/ravendb/ravendb/pull/11635 that uses a semaphore slim. Drawback is that it would not preserve the order of executes which the concurrent queue approach does

ayende:
Unless I miss something, the order doesn't matter.
In the current model not that there isn't any guaranteed order without moving the code to the worker. But given that I don't believe that this is required (the tests should tell us), we can go with the simpler option with the semaphore.

danielmarbach:
> 
> 
> Unless I miss something, the order doesn't matter.
> In the current model not that there isn't any guaranteed order without moving the code to the worker. But given that I don't believe that this is required (the tests should tell us), we can go with the simpler option with the semaphore.

@ayende My knowledge is not that deep in RavenDB to counter that argument. At first I thought ordering was deemed to be desired due to the fact that inheritors from the `RemoteChangesBase` sending commands from the client to the server that are related to watching and unwatching streams of changes happening on the server. So I thought if `watch` and `unwatch` could be reordered that would not be desired from a client/server interaction perspective. Technically the continuation chain built by

```
                    return sendTask.ContinueWith(_ =>
                    {
                        Send(command, value);
                    });
```

would guarantee the ordering. But then I realized that `lock(this)` has weak ordering semantics so it is only roughly in FIFO order and not guaranteed to be in FIFO order so technically the code before never could guarantee FIFO ordering. Though the version with the `SemaphoreSlim` doesn't guaranteed any order at all. 

> If multiple threads are blocked, there is no guaranteed order, such as FIFO or LIFO, that controls when threads enter the semaphore.

Do you think that could turn out to be an issue or not due to the previous code not guaranteeing ordering neither?

ayende:
I don't expect that to be the case. People who are relying on the order are likely to be very rare, and the major reason we had this was for simplicity of the code, not functionality.

ayende:
When re-running the tests, they are all passing.

PR #11582, opened 2020-12-18
RavenDB-16035 Lazy query +  caching returns index does not exists

ayende:


src/Raven.Client/Documents/Commands/MultiGet/MultiGetCommand.cs
```diff
@@ -24,12 +24,24 @@ public MultiGetCommand(HttpCache cache, List<GetRequest> commands)
...
             _commands = commands ?? throw new ArgumentNullException(nameof(commands));
             ResponseType = RavenCommandResponseType.Raw;
+            _cachedValues = new List<(HttpCache.ReleaseCacheItem, BlittableJsonReaderObject)>(commands.Count);
+        }
+        
+        private readonly List<(HttpCache.ReleaseCacheItem, BlittableJsonReaderObject)> _cachedValues;
+
+        ~MultiGetCommand()
+        {
+            ReleaseCachedValues();
```
ppekrol: Wouldn't this cause dispose of BJRO after context was released?

ayende: Hm... good point. That might cause issues, I guess. In that case, it is better to let the context do the cleanup, I guess.

arekpalinski (approved):

PR #11585, opened 2020-12-18
RavenDB-15924 Wrong query results or exception when querying for quot…

ayende:
…es or/and slashes

### DIFF
--- a/src/Raven.Server/Documents/Queries/QueryBuilder.cs
+++ b/src/Raven.Server/Documents/Queries/QueryBuilder.cs
@@ -35,6 +35,7 @@ namespace Raven.Server.Documents.Queries
     public static class QueryBuilder
     {
         private static readonly KeywordAnalyzer KeywordAnalyzer = new KeywordAnalyzer();
+        private static readonly char[] WhiteSpaceCharacters = new[]{' ', '\t'};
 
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/Queries/QueryBuilder.cs
```diff
@@ -34,6 +34,7 @@ namespace Raven.Server.Documents.Queries
     public static class QueryBuilder
     {
         private static readonly KeywordAnalyzer KeywordAnalyzer = new KeywordAnalyzer();
+        private static readonly char[] WhiteSpaceCharacters = new[]{' ', '\t'};
```
ppekrol: is this used?

ppekrol (commented):
Do we really want to introduce those changes to 4.2 at this stage?

src/Raven.Server/Documents/Queries/QueryBuilder.cs
ayende: No, removed.

ayende:
I'm fine with making this with 5.0 / 5.1 instead.

arekpalinski (approved):
I don't see anything wrong here. Although it's very difficult to spot a bug during a review in that piece of code. I trust our test suite but that's potentially risky change IMO.

PR #11587, opened 2020-12-20
RavenDB-15796 Resolving document conflict in studio removes the attac…

efratshenhar:
…hment

### DIFF
--- a/src/Raven.Server/Documents/DocumentPutAction.cs
+++ b/src/Raven.Server/Documents/DocumentPutAction.cs
@@ -169,12 +169,7 @@ namespace Raven.Server.Documents
                         _documentDatabase.DocumentsStorage.RevisionsStorage.Put(context, id, document, flags, nonPersistentFlags, changeVector, modifiedTicks, configuration, collectionName);
                     }
                 }
-                if (document.TryGet(Constants.Documents.Metadata.Key, out BlittableJsonReaderObject metadata) == true &&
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/Handlers/BatchRequestParser.cs
```diff
@@ -113,6 +113,7 @@ private static void ReturnBuffer(CommandData[] cmds)
             using (var parser = new UnmanagedJsonParser(ctx, state, "bulk_docs"))
             using (var modifier = new BlittableMetadataModifier(ctx))
             {
+                modifier.OperateOnTypes |= DatabaseItemType.Attachments;
```
ppekrol: Why we have this here? I would prefer to add this via ctor and with appropriate comment

src/Raven.Server/Documents/DocumentPutAction.cs
```diff
@@ -168,7 +169,12 @@ public DocumentPutAction(DocumentsStorage documentsStorage, DocumentDatabase doc
                         _documentDatabase.DocumentsStorage.RevisionsStorage.Put(context, id, document, flags, nonPersistentFlags, changeVector, modifiedTicks, configuration, collectionName);
                     }
                 }
-
+                if (document.TryGet(Constants.Documents.Metadata.Key, out BlittableJsonReaderObject metadata) == true &&
```
karmeli87: This should be handle in the `RecreateAttachments`, no?

karmeli87 (approved):

PR #11588, opened 2020-12-20
RavenDB-15901 support boost in JS indexes

ppekrol:


src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/JintLuceneDocumentConverter.cs
```diff
@@ -144,36 +186,87 @@ protected override int GetFields<T>(T instance, LazyStringValue key, LazyStringV
...
+                    return value;
+
+                return new BoostedValue
+                {
+                    Boost = boost.Value,
+                    Value = value
+                };
+            }
+
+            static void BoostDocument(T instance, int numberOfCreatedFields, float? boost)
```
ayende, line 207: Why do we care about the number of fields?

ppekrol: This is copied from the anonymous one. We are not boosting 'auxiliary' fields like `id()` or `reduceKey()`

PR #11589, opened 2020-12-20
RavenDB-15979: tighten subscription test

garayx:


src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
```diff
@@ -198,8 +198,10 @@ await requestExecutor.ExecuteAsync(_redirectNode, null, context, command, should
                     {
                         await requestExecutor.ExecuteAsync(command, context, sessionInfo: null, token: token).ConfigureAwait(false);
                         tcpInfo = command.Result;
-
-                        _redirectNode = requestExecutor.Topology.Nodes.Where(x => tcpInfo.Urls.Contains(x.Url)).FirstOrDefault();
+                        if (tcpInfo.NodeTag != null)
+                        {
+                            _redirectNode = requestExecutor.Topology.Nodes.FirstOrDefault(x => x.ClusterTag == tcpInfo.NodeTag);
+                        }
```
ayende, line 204: Can you explain this change?

ppekrol: We were comparing oranges to apples.
`Node.Url` is e.g. `localhost:8080` and `TcpInfo.Urls` are `localhost:12345;localhost:12346`

ayende: Won't that mean that we are now breaking backward compact with servers that won't send this?

ppekrol: It was always returning `null` anyway, so it should not be a problem?

ayende (approved):

PR #11591, opened 2020-12-21
RavenDB-16038 Allow to project the last modified date from queries

ayende:


src/Raven.Client/Util/JavascriptConversionExtensions.cs
```diff
@@ -2412,24 +2412,42 @@ private MetadataSupport()
...
 
-                context.PreventDefault();
-                var writer = context.GetWriter();
-                using (writer.Operation(methodCallExpression))
+                switch (methodCallExpression.Method.Name)
                 {
-                    writer.Write("getMetadata(");
-                    context.Visitor.Visit(methodCallExpression.Arguments[0]);
-                    writer.Write(")");
+                    case "LastModified":
```
ppekrol: can we use nameof? easier to trace

src/Raven.Client/Util/JavascriptConversionExtensions.cs
```diff
@@ -2412,24 +2412,42 @@ private MetadataSupport()
...
+                    {
+                        context.PreventDefault();
+                        var writer = context.GetWriter();
+                        using (writer.Operation(methodCallExpression))
+                        {
+                            writer.Write("getMetadata(");
+                            context.Visitor.Visit(methodCallExpression.Arguments[0]);
+                            writer.Write(")['@last-modified']");
```
ppekrol: can we use constant?

src/Raven.Client/Util/JavascriptConversionExtensions.cs
```diff
@@ -2412,24 +2412,42 @@ private MetadataSupport()
...
+                        }
+                        break;
+                    }
+                    case "Metadata":
+                    case "GetMetadataFor":
```
ppekrol: can we use nameof?

ppekrol (approved):

src/Raven.Client/Util/JavascriptConversionExtensions.cs
ayende:
```suggestion
                            writer.Write(")['" + Constants.Documents.Metadata.LastModified + "']");
```

This is what it will look like with a constant, the string version is better, I think

ppekrol: I'm more concerned about finding all places with the usages. It got handy few times. And compiler should remove the concatenations and use const here IIRC?

ayende: I can't think of any scenario where this would be something we'll need. But ... 🤷

PR #11592, opened 2020-12-21
RavenDB-15993 Separate context for builder in StreamOperation & reset+renew context on too high cachedProperties amount & memory size

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15993

The main issue here is not the memory even so there is a little bit of issue with memory.
But the cause is multiple unique property names.
This leads to a bigger problem where we can have thousands of cached properties and we spend very significant time searching in the inner collections in CachedProperties.

### DIFF
--- a/src/Raven.Client/Documents/Session/Operations/StreamOperation.cs
+++ b/src/Raven.Client/Documents/Session/Operations/StreamOperation.cs
@@ -166,7 +166,7 @@ namespace Raven.Client.Documents.Session.Operations
 
         internal class TimeSeriesStreamEnumerator : IAsyncEnumerator<BlittableJsonReaderObject>, IEnumerator<BlittableJsonReaderObject> 
         {
-            private readonly InMemoryDocumentSessionOperations _session;
+            private readonly JsonOperationContext _context;
             private readonly PeepingTomStream _peepingTomStream;
             private readonly UnmanagedJsonParser _parser;
             private readonly JsonParserState _state;
@@ -174,9 +174,9 @@ namespace Raven.Client.Documents.Session.Operations
             private readonly CancellationToken _token;
             private BlittableJsonReaderObject _current;
 
-            public TimeSeriesStreamEnumerator(InMemoryDocumentSessionOperations session, PeepingTomStream peepingTomStream, UnmanagedJsonParser parser, JsonParserState state, JsonOperationContext.MemoryBuffer buffer, CancellationToken token = default)
+            public TimeSeriesStreamEnumerator(JsonOperationContext context, PeepingTomStream peepingTomStream, UnmanagedJsonParser parser, JsonParserState state, JsonOperationContext.MemoryBuffer buffer, CancellationToken token = default)
             {
-                _session = session;
+                _context = context;
                 _peepingTomStream = peepingTomStream;
                 _parser = parser;
                 _state = state;
@@ -191,7 +191,7 @@ namespace Raven.Client.Documents.Session.Operations
 
             public async Task InitializeAsync()
             {
-                var property = UnmanagedJsonParserHelper.ReadString(_session.Context, _peepingTomStream, _parser, _state, _buffer);
+                var property = UnmanagedJsonParserHelper.ReadString(_context, _peepingTomStream, _parser, _state, _buffer);
 
                 if (property.StartsWith(Constants.TimeSeries.QueryFunction) == false)
                     UnmanagedJsonParserHelper.ThrowInvalidJson(_peepingTomStream);
@@ -241,7 +241,7 @@ namespace Raven.Client.Documents.Session.Operations
                     return false;
                 }
 
-                using (var builder = new BlittableJsonDocumentBuilder(_session.Context, BlittableJsonDocumentBuilder.UsageMode.None, "readArray/singleResult", _parser, _state))
+                using (var builder = new BlittableJsonDocumentBuilder(_context, BlittableJsonDocumentBuilder.UsageMode.None, "readArray/singleResult", _parser, _state))
                 {
                     await UnmanagedJsonParserHelper.ReadObjectAsync(builder, _peepingTomStream, _parser, _buffer, _token).ConfigureAwait(false);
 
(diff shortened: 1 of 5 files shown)


test/SlowTests/Core/Streaming/QueryStreaming.cs
```diff
@@ -236,6 +237,71 @@ public void TestFailingProjection()
...
+            var objs = Enumerable.Range(0, 100000).Select(_ => new TestObj
+            {
+                Prop = new Dictionary<string, string>
+                {
+                    {Guid.NewGuid().ToString("n"), "someValue"},
+                    {Guid.NewGuid().ToString("n"), "someValue"},
+                    {Guid.NewGuid().ToString("n"), "someValue"}
+                }
+            }).ToArray();
+
```
ayende: Can avoid materializing this by operating on the enumerable directly?

haludi: I am not sure I understand your comment.
Do you want it without `ToArray`?

ayende: Yes, then you can simple `foreach` on the values without creating a big array

src/Raven.Client/Documents/Session/Operations/StreamOperation.cs
```diff
@@ -277,6 +277,7 @@ internal class YieldStreamResults : IAsyncEnumerator<BlittableJsonReaderObject>,
             public YieldStreamResults(InMemoryDocumentSessionOperations session, StreamResult response, bool isQueryStream, bool isTimeSeriesStream, bool isAsync, StreamQueryStatistics streamQueryStatistics, CancellationToken token = default)
             {
                 _response = response ?? throw new InvalidOperationException("The index does not exists, failed to stream results");
+                _contextDispose = session.RequestExecutor.ContextPool.AllocateOperationContext(out _builderContext);
```
ppekrol: `_returnContext`

src/Raven.Client/Documents/Session/Operations/StreamOperation.cs
```diff
@@ -403,20 +412,20 @@ public async ValueTask<bool> MoveNextAsync()
                     return false;
                 }
 
-                using (var builder = new BlittableJsonDocumentBuilder(_session.Context, BlittableJsonDocumentBuilder.UsageMode.ToDisk, "readArray/singleResult", _parser, _state))
+                _builder = new BlittableJsonDocumentBuilder(_builderContext, BlittableJsonDocumentBuilder.UsageMode.ToDisk, "readArray/singleResult", _parser, _state);
```
ppekrol: it is creating builder every time in contrast to `MoveNext`

src/Raven.Client/Documents/Session/Operations/StreamOperation.cs
```diff
@@ -403,20 +412,20 @@ public async ValueTask<bool> MoveNextAsync()
...
                 {
                     if (_cachedItemsRenew == false)
```
ppekrol: I do not see builder.Reset and builder.Renew here?

test/SlowTests/Core/Streaming/QueryStreaming.cs
```diff
@@ -236,6 +237,85 @@ public void TestFailingProjection()
             }
         }
 
+        class TestObj
```
ppekrol: private

src/Raven.Client/Documents/Session/Operations/StreamOperation.cs
```diff
@@ -571,21 +580,23 @@ private static void HandleStreamQueryStats(JsonOperationContext context, StreamR
 
             private void CheckIfContextNeedsToBeRenewed()
             {
-                if (_docsCountOnCachedRenewSession <= _maxDocsCountOnCachedRenewSession)
+                if (_builderContext.CachedProperties.PropertiesDiscovered > 1024 ||
```
ayende: The `1024` should be a static field so user can set it

src/Raven.Client/Documents/Session/Operations/StreamOperation.cs
ayende: That is the same behavior as before, no? But where is it closed?
Did you mean: 

```suggestion
                _builder ??= new BlittableJsonDocumentBuilder(_builderContext, BlittableJsonDocumentBuilder.UsageMode.ToDisk, "readArray/singleResult", _parser, _state);
```

src/Raven.Client/Documents/Session/Operations/StreamOperation.cs
```diff
@@ -580,24 +575,19 @@ private static void HandleStreamQueryStats(JsonOperationContext context, StreamR
...
-                if (_cachedItemsRenew)
-                {
-                    _builderContext.CachedProperties.Reset();
-                    _builderContext.CachedProperties.Renew();
 
-                    ++_docsCountOnCachedRenewSession;
-                    _cachedItemsRenew = false;
-                }
+                if (_builder.NeedResetPropertiesCache())
+                    _builderContext.CachedProperties = new CachedProperties(_builderContext);
```
grisha-kotler: What are we doing with the old `CachedProperties`?

haludi: I don't see `Dispose` for `CachedProperties` and as I understand it we do the same when we call context reset&renew with extra memory free (that we don't need here until we have high memory usage and we will reset&renew the whole context).
I mimicked the behavior from `UnmanagedJsonParserHelper.ReadArrayToMemory` and it seems to me to be more efficient.

haludi: I don't see `Dispose` for `CachedProperties` and as I understand it we do the same when we call context reset&renew with extra memory free (that we don't need here until we have high memory usage and we will reset&renew the whole context).
I mimicked the behavior from `UnmanagedJsonParserHelper.ReadArrayToMemory` and it seems to me to be more efficient.

ayende: I would rather reset the `CachedProeprties`, so we'll clear the values, but not allocate new memory.
That would reduce the amount of memory we allocate

ayende (changes requested):

src/Raven.Client/Documents/Session/Operations/StreamOperation.cs
haludi: Reset the `CachedProeprties` doesn't clean all collections in `CachedProeprties` and particularly doesn't clean `_propertiesSortOrder` and `_propertyNameToId` that cause the problem.
I think it is on purpose so may I separate it to `SoftReset` and `HardReset` so the `HardReset` will clean it all?

ayende: There are two modes here. Reset, where we remember the old properties and clear, where we don't.
The whole idea is that we can reduce memory allocations.

ayende (approved):

PR #11597, opened 2020-12-22
RavenDB-15879 Adding standard deviation of 100MB to swap size check

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15879

### DIFF
--- a/src/Raven.Server/ServerWide/ServerStore.cs
+++ b/src/Raven.Server/ServerWide/ServerStore.cs
@@ -734,8 +734,9 @@ namespace Raven.Server.ServerWide
                 if (PlatformDetails.RunningOnDocker) 
                     return;
 
+                var errorThreshold = new Sparrow.Size(128, SizeUnit.Megabytes);
                 var swapSize = MemoryInformation.GetMemoryInfo().TotalSwapSize;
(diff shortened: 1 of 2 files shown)


src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -735,7 +735,7 @@ private void CheckSwapOrPageFileAndRaiseNotification()
                     return;
 
                 var swapSize = MemoryInformation.GetMemoryInfo().TotalSwapSize;
-                if (swapSize < Configuration.PerformanceHints.MinSwapSize)
+                if (swapSize < Configuration.PerformanceHints.MinSwapSize - new Sparrow.Size(100, SizeUnit.Megabytes))
```
ppekrol: Let's change to 128 and introduce a variable called 'errorThreshold' or sth like that.

Also there is a fialing test.

haludi: Do you want the variable outside the function (static or field)?

ppekrol: Just variable.
```
var swapSize = MemoryInformation.GetMemoryInfo().TotalSwapSize;
var errorThreshold = new Sparrow.Size(128, SizeUnit.Megabytes);
if (swapSize < Configuration.PerformanceHints.MinSwapSize - errorThreshold )
```

ppekrol (approved):

PR #11599, opened 2020-12-22
RavenDB-16034 fix MissingRevision3 test

efratshenhar:


test/SlowTests/Issues/RavenDB-15535.cs
```diff
@@ -202,6 +202,14 @@ public async Task MissingRevisions3()
 
                 var val = await WaitForValueAsync(async () => await GetMembersCount(store, database), 2, 20000);
                 Assert.Equal(2, val);
+
+                var delCount = WaitForValue(() =>
```
karmeli87: Please use `WaitForValueAsync` so we use `async` all the way.

PR #11603, opened 2020-12-23
RavenDB-15890 Cluster-wide disable index

efratshenhar:


src/Raven.Server/ServerWide/Commands/Indexes/SetIndexStateCommand.cs
```diff
@@ -32,9 +32,14 @@ public override string UpdateDatabaseRecord(DatabaseRecord record, long etag)
...
+                return null;
             }
-
-            return null;
+            if (record.Indexes.TryGetValue(IndexName, out IndexDefinition indexDefinition))
+            {
+                indexDefinition.State = State;
+                return null;
+            }
+            throw new InvalidOperationException($"Can not change index: {IndexName} state. Index doesn't exist");
```
karmeli87: Do we have a test for that? I'm afraid it might kill the state machine and we will unable to perform further cluster operations.

karmeli87 (changes requested):

src/Raven.Server/ServerWide/Commands/Indexes/SetIndexStateCommand.cs
efratshenhar: It kills the state machine. removing it

karmeli87 (approved):

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -848,13 +848,18 @@ private void UpdateIndex(IndexDefinition definition, Index existingIndex, IndexD
...
         {
             if (definition.LockMode.HasValue && (indexDifferences & IndexDefinitionCompareDifferences.LockMode) != 0)
                 existingIndex.SetLock(definition.LockMode.Value);
 
             if (definition.Priority.HasValue && (indexDifferences & IndexDefinitionCompareDifferences.Priority) != 0)
                 existingIndex.SetPriority(definition.Priority.Value);
+
+            if (definition.State.HasValue && (indexDifferences & IndexDefinitionCompareDifferences.State) != 0)
+            {
+                existingIndex.SetState((IndexState)definition.State);
```
ppekrol: definition.State.Value

ppekrol (commented):
LGTM, but we need to apply that also to auto indexes (AutoIndexDefinition)?

efratshenhar:
We already have state in AutoIndexDefinition. setIndexState command was for auto indexes only

karmeli87:
@efratshenhar after talking to pawel there are few open things :
1. Find all usages of `LockMode`/`Priority` and check that we handle the `Disable` there as well.
2. Need to create an endpoint for this, or extend the existing one for index disabling.
3. Add this to the studio (will be handled by the studio team :-) )

src/Raven.Server/Documents/Indexes/Static/MapIndexDefinition.cs
```diff
@@ -122,9 +123,8 @@ public static IndexDefinition Load(StorageEnvironment environment, out long vers
                     definition.Name = ReadName(reader);
                     definition.LockMode = ReadLockMode(reader);
                     definition.Priority = ReadPriority(reader);
-
                     version = ReadVersion(reader);
-
+                    definition.State = version < 50000 ? IndexState.Normal : ReadState(reader);
```
karmeli87: Do we need to bump the index version?
@ppekrol does index rebuild is needed to update the version of an existing index?

src/Raven.Server/Documents/Indexes/MapReduce/Auto/AutoMapReduceIndexDefinition.cs
```diff
@@ -171,6 +172,7 @@ public static AutoMapReduceIndexDefinition LoadFromJson(BlittableJsonReaderObjec
         {
             var lockMode = ReadLockMode(reader);
             var priority = ReadPriority(reader);
+            var state = ReadState(reader);
```
karmeli87, line 175: need to check for version here?

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -2093,6 +2093,35 @@ public virtual void SetState(IndexState state, bool inMemoryOnly = false, bool i
                 if (State == state)
                     return;
 
+                using (DocumentDatabase.ServerStore.ContextPool.AllocateOperationContext(out TransactionOperationContext ctx))
+                using (ctx.OpenReadTransaction())
+                using (var rawRecord = DocumentDatabase.ServerStore.Cluster.ReadRawDatabaseRecord(ctx, DocumentDatabase.Name))
+                {
+                    //If state is different then staticDef.State/autoDef.State this is not a cluster command.
+                    //In this case we can change the state only if the index state in the cluster is normal or idle
+                    if (TryFindIndexDefinition(Definition.Name, rawRecord, out IndexDefinition staticDef, out AutoIndexDefinition autoDef))
```
karmeli87: Is it possible not to find the index def?

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -2093,6 +2093,35 @@ public virtual void SetState(IndexState state, bool inMemoryOnly = false, bool i
...
+                    {
+                        if (staticDef != null)
```
karmeli87: Would it be the simpler : 
```
if(staticDef?.State == IndexState.Disabled)
   return; // log as well
```

I don't think that `Error` is a valid state for the definition..
@ppekrol Do we want to allow to create an index with an errored state?

src/Raven.Server/Documents/Indexes/MapReduce/Auto/AutoMapReduceIndexDefinition.cs
efratshenhar: yes

src/Raven.Server/Documents/Indexes/IndexDefinitionBase.cs
```diff
@@ -449,6 +456,14 @@ protected static IndexPriority ReadPriority(BlittableJsonReaderObject reader)
             return (IndexPriority)priorityAsInt;
         }
 
+        protected static IndexState ReadState(BlittableJsonReaderObject reader)
+        {
+            if (reader.TryGet(nameof(State), out int StateAsInt) == false)
+                throw new InvalidOperationException("No persisted state");
```
karmeli87: Let's return `Normal` here.

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -2093,6 +2093,35 @@ public virtual void SetState(IndexState state, bool inMemoryOnly = false, bool i
                 if (State == state)
                     return;
 
+                using (DocumentDatabase.ServerStore.ContextPool.AllocateOperationContext(out TransactionOperationContext ctx))
```
karmeli87: Let's remove this completely, last state wins.

src/Raven.Server/Documents/Handlers/Admin/AdminIndexHandler.cs
```diff
@@ -269,6 +269,25 @@ public Task Disable()
             return NoContent();
         }
 
+        [RavenAction("/databases/*/admin/indexes/set-state", "POST", AuthorizationStatus.DatabaseAdmin)]
+        public async Task SetState()
```
karmeli87: Use the regular EP with an additional parameter

karmeli87: Also need to update the client command

src/Raven.Client/Documents/Operations/Indexes/DisableIndexOperation.cs
```diff
@@ -2,42 +2,67 @@
...
 
 namespace Raven.Client.Documents.Operations.Indexes
 {
     public class DisableIndexOperation : IMaintenanceOperation
     {
-        private readonly string _indexName;
+        private readonly Parameters _parameters;
 
-        public DisableIndexOperation(string indexName)
+        public DisableIndexOperation(string indexName, bool clusterWide = false)
```
karmeli87: Let's maintain binary compatibility, and create a separate overload.

src/Raven.Client/Documents/Operations/Indexes/DisableIndexOperation.cs
```diff
@@ -2,42 +2,67 @@
...
+                    Method = HttpMethod.Post,
+                    Content = new BlittableJsonContent(stream =>
+                    {
+                        ctx.Write(stream, _parameters);
+                    })
                 };
             }
         }
+
+        public class Parameters
```
karmeli87: internal

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -882,7 +882,9 @@ internal IndexCreationOptions GetIndexCreationOptions(object indexDefinition, In
             if (indexDef != null)
             {
                 differences = existingIndex.Definition.Compare(indexDef);
-                //Index definition.state can be different from index state so we need to check both
+                //We need to check differences between old and new definition
```
karmeli87, line 885: The key is that we need to figure out the difference between the _in-memory_ state and the definition state.
And not only compare the old def with the new def.

src/Raven.Client/Documents/Operations/Indexes/DisableIndexOperation.cs
karmeli87: Also, where is the Enable op?

grisha-kotler: Maybe create a new operation?
ToggleIndexState?

ppekrol: No

src/Raven.Client/Documents/Operations/Indexes/DisableIndexOperation.cs
```diff
@@ -2,42 +2,82 @@
...
+                if (parameters == null)
+                    throw new ArgumentNullException(nameof(parameters));
+
+                _parameters = DocumentConventions.Default.Serialization.DefaultConverter.ToBlittable(parameters, context);
             }
 
             public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, ServerNode node, out string url)
             {
-                url = $"{node.Url}/databases/{node.Database}/admin/indexes/disable?name={Uri.EscapeDataString(_indexName)}";
+                url = $"{node.Url}/databases/{node.Database}/admin/indexes/disable";
```
karmeli87: This will break backward comp. for studio and users .(I don't think we care about the studio since, we need to re-work it)
Although I like the parameters approach here we need to maintain the back comp.

cc @ppekrol

ppekrol: You need to send the name anyway. I think it would be better just to include both of them in the query string.

src/Raven.Server/Documents/Handlers/Admin/AdminIndexHandler.cs
```diff
@@ -244,29 +244,47 @@ public Task Start()
         }
 
         [RavenAction("/databases/*/admin/indexes/enable", "POST", AuthorizationStatus.DatabaseAdmin)]
-        public Task Enable()
+        public async Task Enable()
         {
+            var raftRequestId = GetRaftRequestIdFromQuery();
             var name = GetStringQueryString("name");
+            var clusterWide = GetBoolValueQueryString("clusterWide") ?? false;
```
ppekrol: remember to set `required: false`

src/Raven.Server/Documents/Handlers/Admin/AdminIndexHandler.cs
```diff
@@ -244,29 +244,47 @@ public Task Start()
...
+            NoContentStatus();
         }
 
         [RavenAction("/databases/*/admin/indexes/disable", "POST", AuthorizationStatus.DatabaseAdmin)]
-        public Task Disable()
+        public async Task Disable()
         {
+            var raftRequestId = GetRaftRequestIdFromQuery();
             var name = GetStringQueryString("name");
+            var clusterWide = GetBoolValueQueryString("clusterWide") ?? false;
```
ppekrol: remember to set `required: false`

PR #11606, opened 2020-12-23
RavenDB-16003 tests stabilizations

karmeli87:


src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -113,7 +113,16 @@ public int HandleDatabaseRecordChange(DatabaseRecord record, long raftIndex)
                 }
                 catch (OperationCanceledException e)
                 {
+                    indexLock.Release();
```
ppekrol: Why? You didn't get a lock

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -113,7 +113,16 @@ public int HandleDatabaseRecordChange(DatabaseRecord record, long raftIndex)
...
                     _documentDatabase.RachisLogIndexNotifications.NotifyListenersAbout(raftIndex, e);
+                    try
+                    {
+                        index.Dispose();
```
ppekrol: You didn't get a lock and database is being shutdown, which would dispose the index anyway

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -126,6 +135,15 @@ public int HandleDatabaseRecordChange(DatabaseRecord record, long raftIndex)
                     _documentDatabase.RachisLogIndexNotifications.NotifyListenersAbout(raftIndex, e);
                     if (Logger.IsInfoEnabled)
                         Logger.Info($"Could not start index `{index.Name}`", e);
+
+                    try
+                    {
+                        index.Dispose();
```
ppekrol: Why? You didn't managed to start the index (any reason) and you are disposing it

src/Raven.Server/Documents/Indexes/IndexStore.cs
karmeli87: you are correct

src/Raven.Server/Documents/Indexes/IndexStore.cs
karmeli87: The env is initialized even if we fail to start the index... so it would make sense to dispose it, no?

test/Tests.Infrastructure/TestBase.cs
```diff
@@ -692,7 +692,6 @@ protected string NewDataPath([CallerMemberName] string prefix = null, string suf
             var path = RavenTestHelper.NewDataPath(prefix, 0, forceCreateDir);
 
             GlobalPathsToDelete.Add(path);
-            _localPathsToDelete.Add(path);
```
ppekrol, line 695: Does it mean that single test is not cleaning after execution? It is done only after whole suite? If yes, we do not want that.

karmeli87: The main issue is that we might have a concurrent file deletion which result in `AccessDenied`, between deleting this at the end of the test and the database deletion.

1. We do clean after each test, each db creation should delete itself.
2. This is why too global place to put paths for deletion, if we have leftovers we need to modify the tests to clean after themselves.

ppekrol: 1. This is not only used for databases, also used internally for other things like smuggler files, etc. This is the way tests clean after themselves.
2. Global is 'just in case'
3. What do you mean concurrent file deletion? The tests should first delete the database, after that we are cleaning the paths (which in most cases should be already deleted)

karmeli87: 1. For smuggler and friends `GetTempFileName` should be used so we will add it to `_localPathsToDelete`.
2. You got a point, the concurrent file deletion is only a symptom to the problem.
The deletion is a multi-step operation, we assume that the files are deleted when the command (or the next one) is committed. which is racy and need to be fixed.

So you want me to keep this line after all?

test/Tests.Infrastructure/RavenLowLevelTestBase.cs
```diff
@@ -26,7 +27,7 @@ namespace FastTests
 {
     public abstract class RavenLowLevelTestBase : TestBase
     {
-        private readonly List<string> _databases = new List<string>();
+        private readonly ConcurrentSet<string> _databases = new ConcurrentSet<string>();
```
ayende, line 30: what is used concurrently?

karmeli87:
Will separate the commits, since some of them need further investigation.

PR #11607, opened 2020-12-24
RavenDB-15917 Cached useOldFailureCheckOperation to static

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15917

### DIFF
--- a/src/Raven.Client/Http/RequestExecutor.cs
+++ b/src/Raven.Client/Http/RequestExecutor.cs
@@ -54,7 +54,7 @@ namespace Raven.Client.Http
 
         private static readonly GetStatisticsOperation BackwardCompatibilityFailureCheckOperation = new GetStatisticsOperation(debugTag: "failure=check");
         private static readonly DatabaseHealthCheckOperation FailureCheckOperation = new DatabaseHealthCheckOperation();
-        private ConcurrentSet<string> _useOldFailureCheckOperation;
+        private static ConcurrentSet<string> _useOldFailureCheckOperation;
 
         private readonly SemaphoreSlim _updateDatabaseTopologySemaphore = new SemaphoreSlim(1, 1);
         private readonly SemaphoreSlim _updateClientConfigurationSemaphore = new SemaphoreSlim(1, 1);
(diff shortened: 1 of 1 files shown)


haludi:
The only way I can see that prevents us to cache the decision to use the old health check is if the health check fails for another reason then version mismatch.
As I understand it, when a request from a client fails for `RouteNotFoundException` and the client has a bigger version than the sever wrap the exception with `ClientVersionMismatchException`.
I can switch the health check failure catch condition so it will check **any** exception if contains `RouteNotFoundException` in its massage (now we check only if the exception type is `ClientVersionMismatchException`) just to be sure but I don't like it.

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -54,7 +54,7 @@ public class RequestExecutor : IDisposable
 
         private static readonly GetStatisticsOperation BackwardCompatibilityFailureCheckOperation = new GetStatisticsOperation(debugTag: "failure=check");
         private static readonly DatabaseHealthCheckOperation FailureCheckOperation = new DatabaseHealthCheckOperation();
-        private ConcurrentSet<string> _useOldFailureCheckOperation;
+        private static ConcurrentSet<string> _useOldFailureCheckOperation;
```
ayende, line 57: That isn't safe. What happens when we have two concurrent document stores to different servers?

ayende (changes requested):

src/Raven.Client/Http/RequestExecutor.cs
haludi: Different server with different Urls so the only downside is a little bit bigger dictionary and so a little bit more time to add/search, no?

ayende: The value of the `string` is the URL, node the node id?

haludi: What do you mean by "node id"? The tag?
The [current implementation](https://github.com/ravendb/ravendb/pull/11607/files/8df03d1ae961740b66fe814e0eb725502bb9dd80#diff-8eb5dec408a858236268c82f3fe1523a8d957370e6a24bef4e21079f70d36ae1R1768) uses the URL includes the protocol (http/s).

ayende: If the URL is there, fine by me

PR #11612, opened 2020-12-28
RDBS-92 de-reference index resources under low mem

karmeli87:


src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexSearcherHolder.cs
```diff
@@ -155,6 +162,22 @@ public IndexSearcher GetIndexSearcher(IState state)
...
 
+            public void MoveBackToLazy()
+            {
+                if (_lazyIndexSearcher.IsValueCreated == false)
+                    return;
+
+                using (_lazyIndexSearcher.Value)
+                using (_lazyIndexSearcher.Value.IndexReader)
+                { }
+
```
ayende, line 182: Here we have a disposed object.

src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexSearcherHolder.cs
```diff
@@ -155,6 +162,22 @@ public IndexSearcher GetIndexSearcher(IState state)
                 Dispose();
             }
 
+            public void MoveBackToLazy()
+            {
+                if (_lazyIndexSearcher.IsValueCreated == false)
+                    return;
+
+                using (_lazyIndexSearcher.Value)
+                using (_lazyIndexSearcher.Value.IndexReader)
```
ayende: What if someone is using this?

src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexSearcherHolder.cs
```diff
@@ -155,6 +162,22 @@ public IndexSearcher GetIndexSearcher(IState state)
...
+                { }
+
+                _lazyIndexSearcher = new Lazy<IndexSearcher>(() =>
+                {
+                    Debug.Assert(_indexSearcherInitializationState != null);
+                    return _recreateSearcher(_indexSearcherInitializationState);
```
ayende: Not thread safe

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -1558,7 +1558,7 @@ public void RunIdleOperations()
         {
             foreach (var index in _indexes)
             {
-                index.Cleanup();
+                index.Cleanup(everything: true);
```
ayende: Not sure that this should be here, that runs on a timer, no?

src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexSearcherHolder.cs
```diff
@@ -143,8 +148,13 @@ public IndexSearcherHoldingState(Transaction tx, Func<IState, IndexSearcher> rec
 
             public IndexSearcher GetIndexSearcher(IState state)
             {
-                _indexSearcherInitializationState = state;
-                return _lazyIndexSearcher.Value;
+                Interlocked.Increment(ref Usage);
+
+                lock (this)
```
ayende: This will serialize all queries, we cannot allow it.

src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexSearcherHolder.cs
```diff
@@ -155,6 +165,30 @@ public IndexSearcher GetIndexSearcher(IState state)
...
 
+            public void MoveBackToLazy()
+            {
+                if (_lazyIndexSearcher.IsValueCreated == false)
+                    return;
+
+                var old = _lazyIndexSearcher.Value;
+
+                lock (this)
+                {
```
ayende, line 184: Here check that old is the same as the current value, otherwise it was already switched

ayende (changes requested):

src/Raven.Server/Documents/Indexes/IndexStore.cs
karmeli87: This runs on a timer, but only if the database is idle

ppekrol: In will happen every 10min (forced), which means that this will clear the searcher every 10min? Not sure if we want this.

src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexSearcherHolder.cs
```diff
@@ -155,6 +172,41 @@ public IndexSearcher GetIndexSearcher(IState state)
...
+                    if (old != _lazyIndexSearcher)
+                        return;
+                    
+                    _isMovingToLazy.Raise();
+
+                    if (Volatile.Read(ref Usage) > 0)
+                    {
+                        _isMovingToLazy.Lower();
```
ppekrol: shouldn't this be in `try finally` ?

src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexSearcherHolder.cs
```diff
@@ -155,6 +172,41 @@ public IndexSearcher GetIndexSearcher(IState state)
...
+                        return;
+                    }
+
+                    _lazyIndexSearcher = new Lazy<IndexSearcher>(() =>
+                    {
+                        Debug.Assert(_indexSearcherInitializationState != null);
+                        return _recreateSearcher(_indexSearcherInitializationState);
+                    }); 
+
+                    _isMovingToLazy.Lower();
```
ppekrol: shouldn't this be in try finally ?

src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexSearcherHolder.cs
```diff
@@ -65,43 +64,50 @@ internal IndexSearcherHoldingState GetStateHolder(Transaction tx)
...
-
                 return state;
             }
 
             throw new InvalidOperationException($"Could not get an index searcher state holder for transaction {txId}");
         }
 
-        public void Cleanup(long oldestTx)
+
+        public void Cleanup(long oldestTx, bool everything = false)
```
arekpalinski: What's the meaning of `"everything"`? What does it mean - "cleanup but not everything". Can we have more meaningful name or maybe an enum here.

src/Raven.Server/Documents/Indexes/IndexStore.cs
arekpalinski: I also have some doubts about it

karmeli87: `RunIdleOperations` is called only if the database is idle, and in that case we have no issue to clean this, I think.
Or do I miss something?

src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexSearcherHolder.cs
karmeli87: Normal clean will keep one `IndexSearcherHoldingState` which might hold an `IndexSearcher`.
`everything` will clean even that last one.

I don't like the name as well, do you have an idea for a better name?

src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexSearcherHolder.cs
karmeli87: done

src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexSearcherHolder.cs
karmeli87: done

src/Raven.Server/Documents/Indexes/IndexStore.cs
ppekrol: Check the second condition here:
https://github.com/ravendb/ravendb/blob/b860e8ee4bb8a4a1070fdd9b5f094eccd7a3095b/src/Raven.Server/ServerWide/ServerStore.cs#L2380

arekpalinski (approved):

ppekrol (approved):

PR #11613, opened 2020-12-28
Prerequisites script requires Mono but it is no longer required by .Net 5.0

redknightlois:
Also changed the null checks to use the safe pattern.

### DIFF
--- a/install_build_prerequisites.sh
+++ b/install_build_prerequisites.sh
@@ -10,7 +10,7 @@ UBUNTU_CODENAME=$(lsb_release -c | cut -d ":" -f2 | sed 's/\t//g')
 UBUNTU_VERSION=$(lsb_release -r | cut -d ":" -f2 | sed 's/\t//g')
 
 if [[ ! "$UBUNTU_VERSION" =~ ^1[468]\.04$ ]] ; then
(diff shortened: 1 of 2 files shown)


ayende (commented):
There are also things in the `install_build_pre...`

redknightlois:
Should we update also the build prerequisites for the new UBUNTU_VERSION? 

```
    elif [ "$UBUNTU_VERSION" = "18.04" ] ; then
        curl https://packages.microsoft.com/config/ubuntu/18.04/prod.list | sudo tee /etc/apt/sources.list.d/microsoft.list
```
Like:
```
    elif [ "$UBUNTU_VERSION" = "20.04" ] ; then
        curl https://packages.microsoft.com/config/ubuntu/20.04/prod.list | sudo tee /etc/apt/sources.list.d/microsoft.list        
    elif [ "$UBUNTU_VERSION" = "20.10" ] ; then
        curl https://packages.microsoft.com/config/ubuntu/20.10/prod.list | sudo tee /etc/apt/sources.list.d/microsoft.list        
```
Also, the prerequisites build installs SDK 2.1

redknightlois:
Mhhh the entire prerequisites need a revamp... probably not the best person to do those changes. cc @ppekrol

PR #11616, opened 2020-12-29
RavenDB-16064  Fix admin logs filtering

Danielle9897:


src/Raven.Studio/typescript/viewmodels/manage/adminLogs.ts
```diff
@@ -183,31 +183,33 @@ class adminLogs extends viewModelBase {
...
-            
-            const filteredItems = fromFilterChange
-                ? this.allData.filter(filterFunction)
-                : this.pendingMessages.filter(filterFunction);
-            
-            this.listController().pushElements(filteredItems);
-        } else {
-            this.listController().pushElements(fromFilterChange ? this.allData : this.pendingMessages);
         }
+        return x => true;
```
ml054: return null

src/Raven.Studio/typescript/viewmodels/manage/adminLogs.ts
```diff
@@ -183,31 +183,33 @@ class adminLogs extends viewModelBase {
...
-        const errorsOnly = this.onlyErrors();
-
         if (fromFilterChange) {
             this.listController().reset();
         }
 
+        const filterFunction = this.getFilterFunction();
+
+        const filteredItems = fromFilterChange ?
+            this.allData.filter(filterFunction) :
```
ml054: filter only if filter function is not null

src/Raven.Studio/typescript/viewmodels/manage/adminLogs.ts
```diff
@@ -348,16 +350,19 @@ class adminLogs extends viewModelBase {
...
+            this.listController().pushElements([...filteredItems, msg]);
+            this.pendingMessages.length = 0;
+            
         } else {
             this.pendingMessages.push(msg);
         }
     }
     
-    private onAppendPendingMessages() {
+    private onMessageAdded() {
```
ml054: name is wrong ihmo.

src/Raven.Studio/typescript/viewmodels/manage/adminLogs.ts

src/Raven.Studio/typescript/viewmodels/manage/adminLogs.ts

src/Raven.Studio/typescript/viewmodels/manage/adminLogs.ts
Danielle9897: we used to just append to `this.pendingMessages`
but now there is also the `showMessageNow `flow
and then we don't append to `this.pendingMesages`, but **add** to `listController`

ml054 (approved):

PR #11618, opened 2020-12-30
RavenDB-16052 Frequent dashboard style recalculation causing  20% CPU

ml054:


arekpalinski:
@ml054 Shouldn't it be for 5.0 as marked in the issue - https://issues.hibernatingrhinos.com/issue/RavenDB-16052 ?

ml054:
right, will fix that

karmeli87:
BTW I think we have this issue in 4.2 as well?

PR #11619, opened 2020-12-30
RavenDB-16065: Tweak behavior on SubscriptionDoesNotBelongToNodeException, tighten DistributedRevisionsSubscription test

garayx:


src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
```diff
@@ -377,7 +377,7 @@ private void AssertConnectionState(SubscriptionConnectionServerMessage connectio
                         }
                     }
                     throw new SubscriptionDoesNotBelongToNodeException(
-                        $"Subscription With Id '{_options.SubscriptionName}' cannot be processed by current node, it will be redirected to {appropriateNode}]{Environment.NewLine}Reasons:{string.Join(Environment.NewLine, reasonsDictionary.Select(x => $"{x.Key}:{x.Value}"))}{Environment.NewLine}Message: {connectionStatus.Message}",
+                        $"Subscription With Id '{_options.SubscriptionName}' cannot be processed by current node, it will be redirected to {appropriateNode}]{Environment.NewLine}Reasons:{string.Join(Environment.NewLine, reasonsDictionary.Select(x => $"{x.Key}:{x.Value}"))}{Environment.NewLine}Exception: {connectionStatus.Exception}",
```
ayende: We should also add the exception as an inner exception, which is better mode all around

src/Raven.Server/ServerWide/Commands/Subscriptions/AcknowledgeSubscriptionBatchCommand.cs
```diff
@@ -41,9 +41,14 @@ protected override BlittableJsonReaderObject GetUpdatedValue(long index, RawData
             var subscription = JsonDeserializationCluster.SubscriptionState(existingValue);
 
             var topology = record.Topology;
-            var lastResponsibleNode = GetLastResponsibleNode(HasHighlyAvailableTasks, topology, NodeTag);
-            if (topology.WhoseTaskIsIt(RachisState.Follower, subscription, lastResponsibleNode) != NodeTag)
-                throw new SubscriptionDoesNotBelongToNodeException($"Can't update subscription with name {subscriptionName} by node {NodeTag}, because it's not its task to update this subscription");
+            var appropriateNode = topology.WhoseTaskIsIt(RachisState.Follower, subscription, GetLastResponsibleNode(HasHighlyAvailableTasks, topology, NodeTag));
+            if (appropriateNode != NodeTag)
```
ayende, line 45: Wrap this in {}

ayende (changes requested):

src/Raven.Client/Documents/Subscriptions/SubscriptionWorker.cs
garayx: It is actually a ToString() of the exception
https://github.com/ravendb/ravendb/blob/ba98e423c8737b417fd6e93015f41be68a0de904/src/Raven.Client/Documents/Subscriptions/SubscriptionWorkerOptions.cs#L72

I can create a general `new Exception(message: connectionStatus.Exception)` and set it as a inner exception, but this will request a new ctor of `SubscriptionDoesNotBelongToNodeException`

ayende: A better option is to have an inner exception, yes. That allow to better track what is going on and simplify the actual error message.

PR #11621, opened 2020-12-31
RavenDB-15494 More informaion on failing test for further investigation

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15494

### DIFF
--- a/test/SlowTests/Issues/RavenDB_15409.cs
+++ b/test/SlowTests/Issues/RavenDB_15409.cs
@@ -1,9 +1,13 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
+using FastTests;
 using Raven.Server.Rachis;
 using Raven.Server.ServerWide.Commands;
+using Sparrow.Json;
+using Sparrow.Logging;
 using Tests.Infrastructure;
 using Xunit;
 using Xunit.Abstractions;
@@ -15,12 +19,15 @@ namespace SlowTests.Issues
         [Fact]
         public async Task DoNotCallUpdateLicenseLimitsCommandOnEveryLeaderChange()
         {
+            await using var logStream = new MemoryStream();
+            LoggingSource.Instance.AttachPipeSink(logStream);
+
             var (servers, leader) = await CreateRaftCluster(3);
             await WaitForRaftIndexToBeAppliedInCluster(9, TimeSpan.FromSeconds(15));
             var expected = new HashSet<long>();
             foreach (var server in servers)
             {
-                expected.Add(CountOfRaftCommandByType(server, nameof(UpdateLicenseLimitsCommand)));
+                expected.Add(GetRaftCommandByType(server, nameof(UpdateLicenseLimitsCommand)).Count());
             }
 
             Assert.Single(expected);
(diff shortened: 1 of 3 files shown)


test/SlowTests/Issues/RavenDB_15409.cs
```diff
@@ -15,12 +19,15 @@ public class RavenDB_15409 : ClusterTestBase
         [Fact]
         public async Task DoNotCallUpdateLicenseLimitsCommandOnEveryLeaderChange()
         {
+            await using var logStream = new MemoryStream();
+            LoggingSource.Instance.AttachPipeSink(logStream);
```
ayende: Won't this cause us to also gather data from other tests? What if there are more locations for this?

ayende: This will also dispose the stream, but still keep it attached?

ayende (changes requested):

test/SlowTests/Issues/RavenDB_15409.cs
haludi: Yes, but it may give us  the information we need and after fixing the test/code we can remove it.

test/SlowTests/Issues/RavenDB_15409.cs
```diff
@@ -12,15 +18,67 @@ namespace SlowTests.Issues
...
+        }
+
+
         [Fact]
         public async Task DoNotCallUpdateLicenseLimitsCommandOnEveryLeaderChange()
         {
+            using var socket = new DummyWebSocket();
+
             var (servers, leader) = await CreateRaftCluster(3);
+            var _ = LoggingSource.Instance.Register(socket, new LoggingSource.WebSocketContext(), CancellationToken.None);
```
ppekrol, line 75: How this is unregistered?

haludi: We unregister when we try to send to the socket and get an error so the close inside `Dispose` and `CloseAndGetLogsAsync` handle that.
https://github.com/haludi/ravendb/blob/RavenDB-15494/src/Sparrow/Logging/LoggingSource.cs#L918
https://github.com/haludi/ravendb/blob/RavenDB-15494/src/Sparrow/Logging/LoggingSource.cs#L949

test/SlowTests/Issues/RavenDB_15409.cs
```diff
@@ -12,15 +18,67 @@ namespace SlowTests.Issues
 {
     public class RavenDB_15409 : ClusterTestBase
     {
+        private class DummyWebSocket : WebSocket
+        {
+            private static WebSocketReceiveResult Result { get; } = new(1, WebSocketMessageType.Text, true);
+            private readonly TaskCompletionSource<WebSocketReceiveResult> _completionSource = new();
```
arekpalinski: Please pass `TaskCreationOptions.RunContinuationsAsynchronously`

test/SlowTests/Issues/RavenDB_15409.cs
```diff
@@ -12,15 +18,67 @@ namespace SlowTests.Issues
...
+            private readonly MemoryStream _stream = new();
+            private int _isClose;
```
arekpalinski: _isClose -> _isClosed

test/Tests.Infrastructure/RavenTestBase.cs
```diff
@@ -142,23 +142,18 @@ protected long LastRaftIndexForCommand(RavenServer server, string commandType)
             return updateIndex;
         }
 
-        protected long CountOfRaftCommandByType(RavenServer server, string commandType)
+        protected IEnumerable<DynamicJsonValue> GetRaftCommandByType(RavenServer server, string commandType = null)
```
arekpalinski: Given that the parameter `commandType` is optional I would rather changed the method name to `GetRaftCommand`.

arekpalinski (approved):

PR #11622, opened 2020-12-31
RavenDB-15742 : OLAP ETL

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-15742


src/Raven.Client/Documents/Operations/ETL/EtlType.cs
```diff
@@ -3,6 +3,7 @@
     public enum EtlType
     {
         Raven,
-        Sql
+        Sql,
+        S3
```
ayende: I don't like the S3 designation here. Maybe Parquet, instead?

src/Raven.Client/Documents/Operations/ETL/S3ConnectionString.cs
```diff
@@ -0,0 +1,29 @@
...
+    public class S3ConnectionString : ConnectionString
+    {
+        public override ConnectionStringType Type => ConnectionStringType.S3;
+
+        public S3Settings S3Settings { get; set; } 
+
+        protected override void ValidateImpl(ref List<string> errors)
+        {
+            if (S3Settings == null)
+                errors.Add($"{nameof(S3Settings)} cannot be null");
```
ayende: we have that in the backup already, no?
Can we merge the impl?

src/Raven.Client/Documents/Operations/OngoingTasks/GetTaskInfoResult.cs
```diff
@@ -14,6 +14,7 @@ public enum OngoingTaskType
         Replication,
         RavenEtl,
         SqlEtl,
+        S3Etl,
```
ayende: ParquetEtl ?

src/Raven.Server/Documents/ETL/EtlLoader.cs
```diff
@@ -212,6 +220,13 @@ private void HandleChangesSubscriptions()
                             connectionStringNotFound = true;
 
                         break;
+                    case EtlType.S3:
+                        s3Config = config as S3EtlConfiguration;
```
ayende: Need to think about how we can integrate this with the user. Can we use a script to get the details, as well?
That way they can generate that on the fly. Like we do for backups.

src/Raven.Server/Documents/ETL/Providers/S3/RowGroups.cs
```diff
@@ -0,0 +1,149 @@
...
+        public Field[] Fields { get; private set; }
+
+        public List<RowGroup> Groups { get; }
+
+        private const int MaxItemsPerGroup = 50_000;
+
+        public void Add(ToS3Item item)
+        {
+            var group = GetCurrentGroup(item);
+            var propCount = item.Properties.Count;
```
ayende: What happens when different items have different number of fields?
What happens if the order of the field isn't fixed?

src/Raven.Server/Documents/ETL/Providers/S3/RowGroups.cs
```diff
@@ -0,0 +1,149 @@
...
+                        throw new ArgumentOutOfRangeException();
+                }
+            }
+
+            group.Count++;
+        }
+
+        private RowGroup GetCurrentGroup(ToS3Item item)
+        {
+            if (Groups.Count == 0 || Groups[^1].Count == MaxItemsPerGroup)
```
ayende: That looks great

src/Raven.Server/Documents/ETL/Providers/S3/S3Etl.cs
```diff
@@ -0,0 +1,171 @@
...
+                // todo upload parquet file to s3/bucketName/TableName/PartitionKey/ and delete file from disc 
+            }
+
+            return count;
+        }
+
+        private string GetTempPath()
+        {
+            var dir = Configuration.TempDirectoryPath ?? Path.GetTempPath();
+            var fileName = $"{Guid.NewGuid()}.parquet";
```
ayende: Should have - db name, etl name, date as well.

src/Raven.Server/Documents/ETL/Providers/S3/S3Etl.cs
```diff
@@ -0,0 +1,171 @@
...
+        public override bool ShouldTrackCounters() => false;
+
+        public override bool ShouldTrackTimeSeries() => false;
+
+        protected override EtlTransformer<ToS3Item, RowGroups> GetTransformer(DocumentsOperationContext context)
+        {
+            return new S3DocumentTransformer(Transformation, Database, context, Configuration);
+        }
+
+        protected override int LoadInternal(IEnumerable<RowGroups> records, DocumentsOperationContext context)
```
ayende: What creates the batches here?

test/SlowTests/Server/Documents/ETL/S3EtlTests.cs
```diff
@@ -0,0 +1,170 @@
...
+                    var etlDone = WaitForEtl(store, (n, statistics) => statistics.LoadSuccesses != 0);
+
+
+                    var script = @"
+var orderDate = new Date(this.OrderedAt);
+var year = orderDate.getFullYear();
+var month = orderDate.getMonth();
+var key = new Date(year, month);
+
+loadToOrders(key,
```
ayende: what is the meaning of `Orders` here?

test/SlowTests/Server/Documents/ETL/S3EtlTests.cs
```diff
@@ -0,0 +1,170 @@
...
+            {
+                Name = connectionStringName,
+                ConnectionStringName = connectionStringName,
+                TempDirectoryPath = path,
+                ETLFrequency = frequency,
+                Transforms =
+                {
+                    new Transformation
+                    {
+                        Name = "MonthlyOrders",
```
ayende: What is the meaning of `MonthlyOrders` here?

src/Raven.Server/Documents/ETL/Providers/S3/S3Etl.cs
```diff
@@ -0,0 +1,171 @@
...
+        public readonly S3EtlMetricsCountersManager S3Metrics = new S3EtlMetricsCountersManager();
+
+        private Timer _timer;
+
+        public S3Etl(Transformation transformation, S3EtlConfiguration configuration, DocumentDatabase database, ServerStore serverStore)
+            : base(transformation, configuration, database, serverStore, S3EtlTag)
+        {
+            Metrics = S3Metrics;
+
+            _timer = new Timer(_ => _waitForChanges.Set(), null, TimeSpan.Zero, configuration.ETLFrequency);
```
ayende: If I understand this properly, we are handling the delay of this kind of ETL using the timer?
So we'll start running on startup, then every `ETLFrequency` ?

karmeli87: This is a good point. 
We need to persist last time it was run and adjust the timer, so we will not start on each server restart.

ayende: Take into account that this may _move_ between nodes as well, so we have to account for that as well.

src/Raven.Client/Documents/Operations/ETL/EtlProcessState.cs
```diff
@@ -27,7 +27,8 @@ public EtlProcessState()
         // When there is time-series without document (that can happen if the time-series was replicated but its document didn't yet)
         // we mark the time-series of the document as skipped so when we load the document we will load all its time-series with it
         public HashSet<string> SkippedTimeSeriesDocs { get; set; }
-        
+        public long LastBatchTime { get; set; }
```
arekpalinski: Any reason why is it `long` instead of `DateTime`?

src/Raven.Client/Documents/Operations/ETL/ParquetEtlConfiguration.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Raven.Client.Documents.Operations.ETL
+{
+    // todo 
+
+    public class ParquetEtlConfiguration : EtlConfiguration<ParquetEtlConnectionString>
+    {
+        public string TempDirectoryPath { get; set; }
+
+        public TimeSpan ETLFrequency { get; set; }
```
arekpalinski: Maybe `RunFrequency`? We are already in "ETL" context

src/Raven.Client/Documents/Operations/ETL/ParquetEtlConnectionString.cs
```diff
@@ -0,0 +1,58 @@
...
+                    if (S3Settings == null)
+                    {
+                        errors.Add($"{nameof(S3Settings)} cannot be null");
+                        return;
+                    }
+                    if (S3Settings.HasSettings() == false)
+                        errors.Add($"{nameof(S3Settings)} has no setting");
+                    return;
+                default:
+                    throw new ArgumentOutOfRangeException();
```
arekpalinski:
```suggestion
                    throw new ArgumentOutOfRangeException(nameof(Destination), Destination, "Unsupported destination type");
```

src/Raven.Client/Documents/Operations/ETL/RavenConnectionString.cs
```diff
@@ -12,7 +12,7 @@ public class RavenConnectionString : ConnectionString
 
         public override ConnectionStringType Type => ConnectionStringType.Raven;
 
-        protected override void ValidateImpl(ref List<string> errors)
+        public override void ValidateImpl(ref List<string> errors)
```
arekpalinski: Why public? For public usage we have ` public bool Validate(ref List<string> errors)`

src/Raven.Server/Commercial/LicenseManager.cs
```diff
@@ -1310,6 +1310,11 @@ public void AssertCanAddSqlEtl()
             const string message = "Your current license doesn't include the SQL ETL feature";
             throw GenerateLicenseLimit(LimitType.SqlEtl, message);
         }
+        
+        public void AssertCanAddParquetEtl()
+        {
+            // todo
```
arekpalinski: Please create a dedicated issue for handling this by @grisha-kotler

src/Raven.Server/Documents/ETL/EtlLoader.cs
```diff
@@ -83,14 +86,15 @@ private void OnProcessRemoved(EtlProcess process)
...
+            List<ParquetEtlConfiguration> newS3Destinations,
             List<EtlProcess> toRemove)
         {
             lock (_loadProcessedLock)
             {
                 _databaseRecord = record;
                 RavenDestinations = _databaseRecord.RavenEtls;
                 SqlDestinations = _databaseRecord.SqlEtls;
-
+                S3Destinations = _databaseRecord.ParquetEtls;
```
arekpalinski: Why `S3Destinations` as the name? S3 is potentially just one of the destinations for Parquet ELT, isn't it?

src/Raven.Server/Documents/ETL/EtlLoader.cs
```diff
@@ -213,7 +221,13 @@ private void HandleChangesSubscriptions()
                             connectionStringNotFound = true;
 
                         break;
-
+                    case EtlType.Parquet:
+                        s3Config = config as ParquetEtlConfiguration;
```
arekpalinski: Same here - why `s3Config`, not `parquetConfig`?

src/Raven.Server/Documents/ETL/EtlProcess.cs
```diff
@@ -709,11 +711,13 @@ public void Run()
                         statsAggregator.Complete();
                     }
 
-                    if (didWork)
+                    if (didWork || ShouldUpdateOnLastBatch)
                     {
+                        var batchTimeMilliseconds = didWork ? 0 : startTime.Ticks / 10_000;
```
arekpalinski: Why dealing with ticks here? What's the pros instead of using `DateTime`?

src/Raven.Server/Documents/ETL/EtlProcess.cs
```diff
@@ -709,11 +711,13 @@ public void Run()
                         statsAggregator.Complete();
                     }
 
-                    if (didWork)
+                    if (didWork || ShouldUpdateOnLastBatch)
```
arekpalinski: I don't follow why we're _always_ updating the ETL state here for ParquetETL? Why `didWork ` was not sufficient?

src/Raven.Server/Documents/ETL/Providers/Parquet/ParquetEtl.cs
```diff
@@ -0,0 +1,260 @@
...
+        private void UploadToS3(S3Settings s3Settings, Stream stream, string path)
+        {
+            var progress = new Progress();
+            using (var client = new RavenAwsS3Client(s3Settings, progress, logger: null, CancellationToken))
+            {
+                var prefix = string.IsNullOrWhiteSpace(s3Settings.RemoteFolderName)
+                    ? string.Empty
+                    : $"{s3Settings.RemoteFolderName}/";
+
+                var key = Path.Combine(prefix, path);
```
arekpalinski: Didn't we want to add some ETL process specific prefix? In order to be able to distinguish between two Parquete ETL file uploaded to the same folder on S3?

src/Raven.Server/Documents/ETL/Providers/Parquet/ParquetDocumentTransformer.cs
```diff
@@ -0,0 +1,146 @@
...
+            {
+                var name = Transformation.LoadTo + collection;
+                DocumentScript.ScriptEngine.SetValue(name, new ClrFunctionInstance(DocumentScript.ScriptEngine, name,
+                    (value, args) => LoadToS3FunctionTranslator(collection, value, args)));
+            }
+        }
+
+        protected override string[] LoadToDestinations { get; }
+
+        protected override void LoadToFunction(string tableName, ScriptRunnerResult colsAsObject)
```
arekpalinski: I wonder what is the meaning of "table" used as parameter name in context of Parquet ETL?

src/Raven.Server/Documents/ETL/Providers/Parquet/ParquetDocumentTransformer.cs
```diff
@@ -0,0 +1,146 @@
...
+            };
+
+            var rowGroups = GetOrAdd(tableName, key);
+            rowGroups.Add(s3Item);
+
+            _stats.IncrementBatchSize(result.Size);
+        }
+
+        protected override void AddLoadedAttachment(JsValue reference, string name, Attachment attachment)
+        {
```
arekpalinski: I assume attachments aren't gonna be supported?

src/Raven.Server/Documents/ETL/Providers/Parquet/ParquetDocumentTransformer.cs
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        protected override void AddLoadedCounter(JsValue reference, string name, long value)
+        {
+            throw new NotSupportedException("Counters aren't supported by SQL ETL");
```
arekpalinski: by SQL ETL -> by Parquet ETL

src/Raven.Server/Documents/ETL/Providers/Parquet/ParquetDocumentTransformer.cs
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        protected override void AddLoadedTimeSeries(JsValue reference, string name, IEnumerable<SingleResult> entries)
+        {
+            throw new NotSupportedException("Time series aren't supported by SQL ETL");
```
arekpalinski: by SQL ETL -> by Parquet ETL

src/Raven.Server/Documents/ETL/Providers/Parquet/ParquetEtl.cs
```diff
@@ -0,0 +1,260 @@
...
+        public ParquetEtl(Transformation transformation, ParquetEtlConfiguration configuration, DocumentDatabase database, ServerStore serverStore)
+            : base(transformation, configuration, database, serverStore, ParquetEtlTag)
+        {
+            Metrics = S3Metrics;
+            _configuration = configuration;
+
+            var etlFrequency = _configuration.ETLFrequency;
+            var dueTime = GetDueTime(etlFrequency);
+
+            _timer = new Timer(_ => _waitForChanges.Set(), null, dueTime, etlFrequency);
```
arekpalinski: Does it needs to be controlled directly by the timer? This mean's that you're forcing to run ETL even if there isn't anything do to.

I was thinking that that actual implementation will be more like this:

- wait for document changes - same as in current impl
- once you get a signal about changes then:
```
if (now - time_since_last_run > ETLFrequency) 
{
  RunParquetETL() 
}
else  
{ 
  WaitAppropriateTimeToReachETLFrequency();
  continue; // to actually run it now
}
```

src/Raven.Server/Documents/ETL/Providers/Parquet/ParquetEtl.cs
```diff
@@ -0,0 +1,260 @@
...
+
+        protected override IEnumerator<ToParquetItem> ConvertDocsEnumerator(DocumentsOperationContext context, IEnumerator<Document> docs, string collection)
+        {
+            return new DocumentsToParquetItems(docs, collection);
+        }
+
+        protected override IEnumerator<ToParquetItem> ConvertTombstonesEnumerator(DocumentsOperationContext context, IEnumerator<Tombstone> tombstones, string collection, bool trackAttachments)
+        {
+            return EmptyEnumerator; // todo
+            throw new NotSupportedException("Tombstones aren't supported by S3 ETL");
```
arekpalinski: S3 ETL ?

arekpalinski: In other places as well

src/Raven.Server/Documents/ETL/Providers/Parquet/ParquetEtl.cs
```diff
@@ -0,0 +1,260 @@
...
+            {
+                Logger.Info($"[{Name}] Inserted {group.Count} records to '{name}/{key}' table " +
+                            $"from the following documents: {string.Join(", ", group.Ids)}");
+            }
+        }
+
+        private string GetPath(RowGroups group, out string remotePath)
+        {
+            var dir = Configuration.TempDirectoryPath ?? Path.GetTempPath();
+            var fileName = $"{Database.Name}_{Guid.NewGuid()}.parquet";
```
arekpalinski: I think it would make sense to include ETL process name too

src/Raven.Server/Documents/ETL/Providers/Parquet/ParquetEtl.cs
```diff
@@ -0,0 +1,260 @@
...
+                        case DataType.Int64:
+                            array = ((List<long>)data).ToArray();
+                            break;
+                        case DataType.String:
+                            array = ((List<string>)data).ToArray();
+                            break;
+                        case DataType.Float:
+                        case DataType.Double:
+                        case DataType.Decimal:
+                            array = ((List<double>)data).ToArray();
```
arekpalinski: Can we use `ArrayPool` to avoid allocating so many arrays?

src/Raven.Server/Documents/ETL/Providers/Parquet/ParquetEtl.cs
```diff
@@ -0,0 +1,260 @@
...
+                }
+
+
+                // todo delete file from disc 
+            }
+
+            return count;
+        }
+
+        private void WriteGroup(ParquetWriter parquetWriter, RowGroup group, RowGroups rowGroups)
```
arekpalinski: Wouldn't it make sense to move this method and its implementation to `ParquetWriter` as well?

src/Raven.Server/Documents/ETL/Providers/Parquet/ParquetEtl.cs
```diff
@@ -0,0 +1,260 @@
...
+            {
+                case ParquetEtlDestination.S3:
+                    var s3Settings = BackupTask.GetBackupConfigurationFromScript(connection.S3Settings, x => JsonDeserializationServer.S3Settings(x), 
+                        Database, updateServerWideSettingsFunc: null, serverWide: false);
+
+                    // todo
+                    //UploadToS3(s3Settings, stream, path);
+                    break;
+                default:
+                    throw new ArgumentOutOfRangeException();
```
arekpalinski: Please provide param name and value to ArgumentOutOfRangeException

src/Raven.Server/Documents/ETL/Providers/Parquet/ParquetEtl.cs
```diff
@@ -0,0 +1,260 @@
...
+            var progress = new Progress();
+            using (var client = new RavenAwsS3Client(s3Settings, progress, logger: null, CancellationToken))
+            {
+                var prefix = string.IsNullOrWhiteSpace(s3Settings.RemoteFolderName)
+                    ? string.Empty
+                    : $"{s3Settings.RemoteFolderName}/";
+
+                var key = Path.Combine(prefix, path);
+
+                client.PutObject(key, stream, new Dictionary<string, string>
```
arekpalinski: I think we need to expose the progress of upload to user sso they will know that ETL process isn't stuck.

src/Raven.Server/Documents/ETL/Providers/Parquet/ParquetDocumentTransformer.cs
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        private RowGroups GetOrAdd(string tableName, string key)
+        {
+            var name = tableName + "_" + key;
+
+            if (_tables.TryGetValue(name, out var table) == false)
+            {
+                _tables[name] = table = new RowGroups(tableName, key);
```
arekpalinski: This dictionary can be potentially very large because it keeps _all_ records in memory, right? As I understand the batches in Parquet ETL will be significantly larger because they aggregate data modified in the last few hours (depending on `ETLFrequency` configuration).

Do we have a way to write the result of LoadTo() immediately to the temporary parquet file? So we would not need keep it in memory

src/Raven.Server/Documents/ETL/Providers/Parquet/ParquetEtl.cs
arekpalinski: Also I think we should let user know that if the upload takes more time that `ETLFrequency` - so he might need to adjust that setting.

src/Raven.Server/Documents/ETL/Providers/Parquet/RowGroups.cs
```diff
@@ -0,0 +1,256 @@
...
+
+        public Dictionary<string, DataField> Fields => _fields ??= GenerateDataFields();
+
+        public List<RowGroup> Groups { get; }
+
+        private readonly Dictionary<string, DataType> _dataFields;
+
+        private Dictionary<string, DataField> _fields;
+
+        private const int MaxItemsPerGroup = 50_000;
```
arekpalinski: Shouldn't it be configurable for a user? (advanced setting)

src/Raven.Server/Documents/ETL/Providers/Parquet/ParquetDocumentTransformer.cs
```diff
@@ -0,0 +1,146 @@
...
+
+        public override List<RowGroups> GetTransformedResults()  
+        {
+            return _tables.Values.ToList();
+        }
+
+        public override void Transform(ToParquetItem item, EtlStatsScope stats, EtlProcessState state)
+        {
+            _stats = stats;
+            if (item.IsDelete)
```
arekpalinski: Given that we do nothing on delete I then we can not iterate over tombstones at all, right?

src/Raven.Server/Documents/ETL/Providers/Parquet/RowGroups.cs
```diff
@@ -0,0 +1,256 @@
...
+                    case BlittableJsonToken.String:
+                        propType = DataType.String;
+                        data ??= new List<string>();
+                        break;
+                    case BlittableJsonToken.Boolean:
+                        propType = DataType.Boolean;
+                        data ??= new List<bool>();
+                        break;
+                    default:
+                        throw new ArgumentOutOfRangeException();
```
arekpalinski: please provide parameters to ArgumentOutOfRangeException

src/Raven.Server/Documents/ETL/Providers/Parquet/RowGroups.cs
```diff
@@ -0,0 +1,256 @@
...
+                        break;
+                    case BlittableJsonToken.Null:
+                        if (newField)
+                            UpdateField(propType, propName, data, group);
+                        else
+                            AddDefaultData(data, dataType, 1);
+                        continue;
+                    case BlittableJsonToken.Integer:
+                        propType = DataType.Int64;
+                        data ??= new List<long>();
```
arekpalinski: To make sue -  is everything a list here because parquet format is column based?

src/Raven.Server/ServerWide/Commands/ConnectionStrings/RemoveConnectionStringCommand.cs
```diff
@@ -61,4 +61,22 @@ public override void UpdateDatabaseRecord(DatabaseRecord record, long etag)
             record.SqlConnectionStrings.Remove(ConnectionStringName);
         }
     }
+
+    public class RemoveOlapConnectionStringCommand : RemoveConnectionStringCommand<ParquetEtlConnectionString>
```
arekpalinski: Is it used anywhere? It's the first time I see "OLAP"

arekpalinski (commented):
After reviewing entire PR I have some doubts about the implementation. My concern is that the code is super "parquet" oriented - starting from small things like enum `EtlType.Parquet` and ending at the main class `ParquetEtl`.

Actually I thought that it's gonna be more "OLAP" oriented and parquet format will be just one of the possible output providers (the only one - in the first implementation). That could give us some flexibility in adding more formats in the future.

Same approach when it comes to the upload - right now just S3 is supported but in the future we might have more. So S3 isn't Parquet ETL specific.

src/Raven.Server/Documents/ETL/Providers/Parquet/ParquetDocumentTransformer.cs
arekpalinski: Another thing that is concerning is the duration of the transaction. We're processing entire batch in single transaction right?

src/Raven.Server/Documents/ETL/Providers/Parquet/ParquetEtl.cs
ayende: Yes, some prefix (used defined?) is required

src/Raven.Server/Documents/ETL/Providers/Parquet/ParquetDocumentTransformer.cs
ayende: Not planning to

src/Raven.Server/Documents/ETL/Providers/Parquet/ParquetDocumentTransformer.cs
ayende: I would like to build them on the fly, yes.
We can write multiple RowGroups, I think

ayende:
The idea is that we are going to be generating Parquet files, the choices are Parquet, Avro or ORC, pretty much.
Avro is row based, and not likely fit this need. Parquet is better supported than ORC.

The name OLAP ETL can be used, yes, and we should have a single option file type. But in essence, I don't think we'll be adding
any additional formats easily. 

And the upload process *must* support anything that the backup does, and it should use the exact same code.

src/Raven.Client/Documents/Operations/ETL/EtlProcessState.cs
arekpalinski: Still same question ^

src/Raven.Client/Documents/Operations/ETL/OLAP/OlapEtlConfiguration.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public bool KeepFilesOnDisk { get; set; }
+        
+        public string PartitionFieldName { get; set;  }
+
+        public string CustomPrefix { get; set; }
+
+        public List<OlapEtlTable> OlapTables { get; set; }
+
+        public override string GetDestination() => ConnectionStringName;
```
arekpalinski: Just a name of connection string isn't very meaningful. Maybe let's extract S3 details here - e.g. bucket name. That's the real destination of OLAP ETL files

src/Raven.Client/Documents/Operations/ETL/OLAP/OlapEtlConfiguration.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public override EtlType EtlType => EtlType.Olap;
+
+        public override bool UsingEncryptedCommunicationChannel()
+        {
+            throw new NotImplementedException();
```
arekpalinski, line 30: We need to have it handled when outputting data from encrypted db

src/Raven.Server/Documents/ETL/Providers/OLAP/OlapDocumentTransformer.cs
```diff
@@ -0,0 +1,169 @@
...
+            if (args.Length != 2)
+                ThrowInvalidScriptMethodCall($"loadTo{name}(key, obj) must be called with exactly 2 parameters");
+
+            if (args[0].IsDate() == false)
+                ThrowInvalidScriptMethodCall($"loadTo{name}(key, obj) argument 'key' must be a date object");
+            
+            if (args[1].IsObject() == false)
+                ThrowInvalidScriptMethodCall($"loadTo{name}(key, obj) argument 'obj' must be an object");
+
+            var key = args[0].AsDate().ToDateTime().ToString(DateFormat); //todo
```
arekpalinski: What about this `todo`?

src/Raven.Server/Documents/ETL/Providers/OLAP/OlaptEtl.cs
```diff
@@ -0,0 +1,202 @@
...
+        private static readonly IEnumerator<ToOlapItem> EmptyEnumerator = Enumerable.Empty<ToOlapItem>().GetEnumerator();
+
+        protected override IEnumerator<ToOlapItem> ConvertDocsEnumerator(DocumentsOperationContext context, IEnumerator<Document> docs, string collection)
+        {
+            return new DocumentsToOlapItems(docs, collection);
+        }
+
+        protected override IEnumerator<ToOlapItem> ConvertTombstonesEnumerator(DocumentsOperationContext context, IEnumerator<Tombstone> tombstones, string collection, bool trackAttachments)
+        {
+            return EmptyEnumerator;
```
arekpalinski: Tombstones are just ignored, right? This is important to write about that explicitly in the documentation

src/Raven.Server/Documents/ETL/Providers/OLAP/OlaptEtl.cs
```diff
@@ -0,0 +1,202 @@
...
+            if (_s3Settings != null)
+            {
+                var key = GetKey(_s3Settings, path);
+                UploadToS3(_s3Settings, stream, key);
+            }
+        }
+
+        private void UploadToS3(S3Settings s3Settings, Stream stream, string key)
+        {
+            using (var client = new RavenAwsS3Client(s3Settings, progress: null, Logger, CancellationToken))
```
arekpalinski: `progress: null` - can we expose it to the user? We already know how many files has been uploaded, and from here we might know the upload progress of current file, right?

If this is feasible please create a dedicates issue to handle that

src/Raven.Server/Documents/ETL/Providers/OLAP/OlaptEtl.cs
```diff
@@ -0,0 +1,202 @@
...
+            if (string.IsNullOrWhiteSpace(Configuration.CustomPrefix) == false)
+                prefix = $"{prefix}/{Configuration.CustomPrefix}";
+
+            return $"{prefix}/{path}";
+        }
+
+
+        private string GetPath(OlapTransformedItems transformed, out string remotePath)
+        {
+            var fileName = $"{Database.Name}__{Guid.NewGuid()}.{transformed.Format}";
```
arekpalinski: Double `_` characters in a file name?

src/Raven.Server/Documents/ETL/Providers/OLAP/ParquetTransformedItems.cs
```diff
@@ -0,0 +1,445 @@
...
+            _dataFields = new Dictionary<string, DataType>();
+
+            if (string.IsNullOrEmpty(partitionColumn))
+                partitionColumn = DefaultPartitionColumn;
+
+            DocumentIdColumn = string.IsNullOrEmpty(idColumn)
+                ? DefaultIdColumn
+                : idColumn;
+
+            Prefix = $"{CollectionName}/{partitionColumn}%3D{PartitionKey}";
```
arekpalinski: What is `%3D`? Why it needs to be provided this way? Please create a dedicated const with proper name for that

src/Raven.Server/Documents/ETL/Providers/OLAP/ParquetTransformedItems.cs
```diff
@@ -0,0 +1,445 @@
...
+
+        public Dictionary<string, DataField> Fields => _fields ??= GenerateDataFields();
+
+        public List<RowGroup> Groups { get; }
+
+        private readonly Dictionary<string, DataType> _dataFields;
+
+        private Dictionary<string, DataField> _fields;
+
+        private const int MaxItemsPerGroup = 50_000;
```
arekpalinski: Didn't we want to make it configurable (as an advanced option)?

src/Raven.Server/Documents/ETL/Providers/OLAP/ParquetTransformedItems.cs
```diff
@@ -0,0 +1,445 @@
...
+                var newField = _dataFields.TryGetValue(propName, out var dataType) == false;
+                group.Data.TryGetValue(propName, out var data);
+
+                switch (prop.Type)
+                {
+                    case BlittableJsonToken.StartObject:
+                    case BlittableJsonToken.StartArray:
+                    case BlittableJsonToken.CompressedString:
+                    case BlittableJsonToken.EmbeddedBlittable:
+                        // todo
```
arekpalinski: What `todo` here?

src/Raven.Server/Documents/ETL/Providers/OLAP/ParquetTransformedItems.cs
```diff
@@ -0,0 +1,445 @@
...
+                            UpdateField(propType, propName, data, group);
+                        else
+                            AddDefaultData(data, dataType, 1);
+                        continue;
+                    case BlittableJsonToken.Integer:
+                        propType = DataType.Int64;
+                        data ??= new List<long>();
+                        break;
+                    case BlittableJsonToken.LazyNumber:
+                        // todo
```
arekpalinski: What `todo` here?

src/Raven.Server/Raven.Server.csproj
```diff
@@ -144,6 +144,9 @@
     <PackageReference Include="NuGet.Protocol" Version="5.8.1" />
     <PackageReference Include="Oracle.ManagedDataAccess.Core" Version="2.19.101" />
     <PackageReference Include="Portable.BouncyCastle" Version="1.8.9" />
+    <PackageReference Include="Oracle.ManagedDataAccess.Core" Version="2.19.100" />
```
arekpalinski: We already have it referenced - see above:

`<PackageReference Include="Oracle.ManagedDataAccess.Core" Version="2.19.101" />`

src/Raven.Server/Raven.Server.csproj
```diff
@@ -144,6 +144,9 @@
...
+    <PackageReference Include="Parquet.Net" Version="3.8.4" />
+    <PackageReference Include="Portable.BouncyCastle" Version="1.8.8" />
```
arekpalinski: We already have it referenced - see above:

`<PackageReference Include="Portable.BouncyCastle" Version="1.8.9" />`

src/Raven.Client/Documents/Operations/ETL/OLAP/OlapEtlConfiguration.cs
ayende: Note that this ETL isn't about S3, it can go to SFTP, for example.

src/Raven.Client/Documents/Operations/ETL/OLAP/OlapEtlConfiguration.cs
ayende: For S3, Azure, GCP storage - can assume this is safe.
Local path, also safe.
For FTP - abort
SFTP - Allow

src/Raven.Server/Documents/ETL/Providers/OLAP/OlaptEtl.cs
ayende: Yes, that is a good point to have, they might be _large_.
Also need to make sure we handle resuming uploads.

src/Raven.Client/Documents/Operations/ETL/OLAP/OlapEtlConfiguration.cs
arekpalinski: Exactly - we should pull the details about connection from the connection string - not returning the connection string name. 

We already do it for SQL and Raven ETLs via `Connection` property

src/Raven.Studio/typescript/commands/database/settings/getConnectionStringInfoCommand.ts
```diff
@@ -28,6 +28,10 @@ class getConnectionStringInfoCommand extends commandBase {
     static forSqlEtl(db: database, connectionStringName: string) {
         return new getConnectionStringInfoCommand(db, "Sql", connectionStringName);
     }
+
+    static forOlapEtl(db: database, connectionStringName: string) {
+        return new getConnectionStringInfoCommand(db, "Olap", connectionStringName);
```
ml054, line 33: OLAP

src/Raven.Studio/typescript/models/database/tasks/ongoingTaskOlapEtlTransformationModel.ts
```diff
@@ -0,0 +1,91 @@
...
+                Disabled: false,
+                Name: "",
+                Script: ""
+            }, true, false);
+    }
+
+    toDto(): Raven.Client.Documents.Operations.ETL.Transformation {
+        return {
+            ApplyToAllDocuments: false,
+            Collections: [this.collection()],
```
ml054: why we support single collection? see: https://github.com/ravendb/ravendb/blob/c329d57f62a480d9b66e0bcc5c110b286d676c3b/src/Raven.Studio/typescript/models/database/tasks/ongoingTaskRavenEtlTransformationModel.ts#L59

src/Raven.Studio/typescript/models/database/tasks/periodicBackup/backupConfiguration.ts
```diff
@@ -1,6 +1,7 @@
 ﻿/// <reference path="../../../../../typings/tsd.d.ts"/>
 import localSettings = require("models/database/tasks/periodicBackup/localSettings");
-import s3Settings = require("models/database/tasks/periodicBackup/s3Settings");
+// import s3Settings = require("models/database/tasks/periodicBackup/s3Settings");
```
ml054: remove

src/Raven.Studio/typescript/models/database/tasks/smugglerDatabaseRecord.ts
```diff
@@ -20,8 +20,10 @@ class smugglerDatabaseRecord {
     includeExternalReplications = ko.observable<boolean>(true);
     includeRavenConnectionStrings = ko.observable<boolean>(true);
     includeSqlConnectionStrings = ko.observable<boolean>(true);
+    includeOlapConnectionStrings = ko.observable<boolean>(true);
```
ml054, line 23: Is server side ready?

src/Raven.Studio/typescript/viewmodels/database/settings/connectionStrings.ts
```diff
@@ -195,47 +233,86 @@ class connectionStrings extends viewModelBase {
...
-                this.editedSqlEtlConnectionString(new connectionStringSqlEtlModel(result.SqlConnectionStrings[connectionStringName], false, this.getTasksThatUseThisString(connectionStringName, "Sql")));
-                this.editedSqlEtlConnectionString().connectionString.subscribe(() => this.clearTestResult());
-                this.editedRavenEtlConnectionString(null);
+                this.editedOlapEtlConnectionString(new connectionStringOlapEtlModel(result.OlapConnectionStrings[connectionStringName], false, this.getTasksThatUseThisString(connectionStringName, "Olap")));
+                this.onOlapEtl();
             });
     }
     
+    private onOlapEtl() {
+        this.editedOlapEtlConnectionString().s3Settings().bucketName.subscribe(() => this.clearTestResult());
```
ml054, line 303: this is ugly. Can't we move that logic to s3settings object?

src/Raven.Studio/wwwroot/App/views/database/tasks/createOngoingTask.html
```diff
@@ -42,6 +42,12 @@ <h5>ETL</h5>
                             <i class="icon-sql-etl"></i><br />SQL ETL
                         </a>
                     </div>
+                    <!-- todo... olap-icon...-->
```
ml054: icon

src/Raven.Studio/wwwroot/App/views/partial/backupDestinationLocal.html
```diff
@@ -0,0 +1,58 @@
...
+                            <br/>
+                            Path: <strong data-bind="text: $parents[1].localSettings().container() === 'backup' && $parents[1].isServerWide() ? FullPath + '\\{DATABASE-NAME}\\' : FullPath"></strong><br />
+                                                        
+                            <!-- ko if: Error -->
+                            Error: <strong data-bind="text: Error"></strong>
+                            <!-- /ko -->
+                            <!-- ko ifnot: Error -->
+                            Free Disk Space:
+                            <span data-bind="if: FreeSpaceHumane">
+                                                        <strong data-bind="text: FreeSpaceHumane"></strong>
```
ml054, line 43: indent here is wrong?

src/Raven.Studio/wwwroot/App/views/shell/about.html
```diff
@@ -149,6 +149,7 @@ <h4>Backups</h4>
                     <div class="name">Encrypted backups</div><div class="value"><i data-bind="attr: { class: licenseAttribute('HasEncryptedBackups') }"></i></div>
                     <h4>ETL</h4>
                     <div class="name">SQL ETL</div><div class="value"><i data-bind="attr: { class: licenseAttribute('HasSqlEtl') }"></i></div>
+                    <div class="name">OLAP ETL</div><div class="value"><i data-bind="attr: { class: licenseAttribute('HasOlapEtl') }"></i></div>
```
ml054, line 155: we need to make sure that will be final license attribute name

src/Raven.Studio/typescript/models/database/tasks/ongoingTaskOlapEtlListModel.ts
```diff
@@ -0,0 +1,36 @@
...
+    connectionStringDefined = ko.observable<boolean>(true); // needed for template in the ongoing tasks list view
+        
+    constructor(dto: Raven.Client.Documents.Operations.OngoingTasks.OngoingTaskOlapEtlListView) {
+        super();
+
+        this.update(dto);
+        this.initializeObservables();
+
+        this.connectionStringsUrl = appUrl.forConnectionStrings(activeDatabaseTracker.default.database(), "olap", this.connectionStringName());
+        console.log("url === " + this.connectionStringsUrl);
```
ml054: remove

src/Raven.Studio/typescript/models/database/tasks/ongoingTaskOlapEtlTransformationModel.ts
```diff
@@ -0,0 +1,91 @@
...
+        this.validationGroup = ko.validatedObservable({
+            script: this.script,
+            collection: this.collection
+        });
+    }
+
+    update(dto: Raven.Client.Documents.Operations.ETL.Transformation, isNew: boolean, resetScript: boolean) {
+        this.name(dto.Name);
+        this.script(dto.Script); 
+        this.collection(dto.Collections[0]);
```
ml054: why ?

src/Raven.Studio/typescript/models/database/tasks/periodicBackup/localSettings.ts
```diff
@@ -2,18 +2,23 @@
...
 
+type container =  "backup" | "connectionString";
+
 class localSettings extends backupSettings {
     folderPath = ko.observable<string>();
     folderPathHasFocus = ko.observable<boolean>(false);
-
-    constructor(dto: Raven.Client.Documents.Operations.Backups.LocalSettings) {
+    
+    container = ko.observable<container>();
```
ml054, line 11: what's the purpose of that?

ml054: can we avoid having that?

src/Raven.Studio/typescript/viewmodels/database/settings/connectionStrings.ts
```diff
@@ -166,21 +199,26 @@ class connectionStrings extends viewModelBase {
...
     }
 
     confirmDelete(connectionStringName: string, connectionStringtype: Raven.Client.Documents.Operations.ConnectionStrings.ConnectionStringType) {
-        const stringType = connectionStringtype === "Raven" ? "RavenDB" : "SQL";
-        this.confirmationMessage("Are you sure?",
-            `Do you want to delete ${stringType} ETL connection string: <br><strong>${generalUtils.escapeHtml(connectionStringName)}</strong>`, {
-            buttons: ["Cancel", "Delete"],
-            html: true
-        })
+        const stringType = connectionStringtype === "Raven" ? "RavenDB" :  connectionStringtype === "Sql" ? "SQL" : "Olap";
```
ml054, line 210: I've seen this login in some other place - I think it is time to move that to some static method with maps that .

src/Raven.Studio/typescript/viewmodels/database/settings/connectionStrings.ts
```diff
@@ -264,7 +345,7 @@ class connectionStrings extends viewModelBase {
 
         if (sqlConnectionString) {
             if (this.isValid(sqlConnectionString.testConnectionValidationGroup)) {
-                eventsCollector.default.reportEvent("ravenDB-SQL-connection-string", "test-connection");
+                eventsCollector.default.reportEvent("SQL-connection-string", "test-connection");
```
ml054, line 352: please avoid changing that - we already have events in GA - we want to correlate them!

src/Raven.Studio/typescript/viewmodels/database/settings/connectionStrings.ts
```diff
@@ -279,7 +360,7 @@ class connectionStrings extends viewModelBase {
     onTestConnectionRaven(urlToTest: discoveryUrl) {
         this.clearTestResult();
         const ravenConnectionString = this.editedRavenEtlConnectionString();
-        eventsCollector.default.reportEvent("ravenDB-ETL-connection-string", "test-connection");
+        eventsCollector.default.reportEvent("ravenDB-connection-string", "test-connection");
```
ml054, line 367: same here.

src/Raven.Studio/typescript/viewmodels/database/tasks/editOlapEtlTask.ts
```diff
@@ -0,0 +1,583 @@
...
+import connectionStringOlapEtlModel = require("models/database/settings/connectionStringOlapEtlModel");
+import ongoingTaskOlapEtlEditModel = require("models/database/tasks/ongoingTaskOlapEtlEditModel");
+import ongoingTaskOlapEtlTransformationModel = require("models/database/tasks/ongoingTaskOlapEtlTransformationModel");
+import ongoingTaskOlapEtlTableModel = require("models/database/tasks/ongoingTaskOlapEtlTableModel");
+import aceEditorBindingHandler = require("common/bindingHelpers/aceEditorBindingHandler");
+import getPossibleMentorsCommand = require("commands/database/tasks/getPossibleMentorsCommand");
+import jsonUtil = require("common/jsonUtil");
+import popoverUtils = require("common/popoverUtils");
+import tasksCommonContent = require("models/database/tasks/tasksCommonContent");
+import backupSettings = require("../../../models/database/tasks/periodicBackup/backupSettings");
```
ml054: update import

src/Raven.Studio/typescript/viewmodels/database/tasks/editOlapEtlTask.ts
```diff
@@ -0,0 +1,583 @@
...
+
+        if (args.taskId) {
+            // 1. Editing an Existing task
+            this.isAddingNewOlapEtlTask(false);
+
+            getOngoingTaskInfoCommand.forOlapEtl(this.activeDatabase(), args.taskId)
+                .execute()
+                .done((result: Raven.Client.Documents.Operations.OngoingTasks.OngoingTaskOlapEtlDetails) => {
+                    this.editedOlapEtl(new ongoingTaskOlapEtlEditModel(result));
+                    this.showAdvancedOptions(!!result.Configuration.MaxNumberOfItemsInRowGroup);
```
ml054: let's move that logic to separate method - it will be helpful if we add extra conditions in future.

src/Raven.Studio/typescript/viewmodels/database/tasks/editOlapEtlTask.ts
```diff
@@ -0,0 +1,583 @@
...
+            {
+                content: `<small>Set custom prefix</small>`
+            });
+
+        popoverUtils.longWithHover($(".run-frequency"),
+            {
+                content: `<small>Frequency at which this task will be triggered</small>`
+            });
+
+        popoverUtils.longWithHover($(".max-items-in-row"),
```
ml054: this is under advanced so may not be visible by default (please make sure tooltip is working when you defining new OLAP)

src/Raven.Studio/typescript/viewmodels/database/tasks/editOlapEtlTask.ts
```diff
@@ -0,0 +1,583 @@
...
+            else {
+                const olapTableToUpdate = this.editedOlapEtl().olapTables().find(x => x.tableName() === this.olapTableSelectedForEdit().tableName());
+                this.overwriteExistingOlapTable(olapTableToUpdate, newOlapTable);
+                overwriteAction.resolve(true);
+            }
+        } 
+
+        overwriteAction.done(() => {
+            this.editedOlapEtl().olapTables.sort((a, b) => a.tableName().toLowerCase().localeCompare(b.tableName().toLowerCase()));
+            if (overwriteAction) { this.editedOlapTableSandbox(null); }
```
ml054: format code

src/Raven.Studio/wwwroot/App/views/database/settings/connectionStrings.html
```diff
@@ -67,23 +67,56 @@ <h4 class="text-elipsis">
...
+                <form class="flex-form padding" autocomplete="off">
                     <div data-bind="with: editedRavenEtlConnectionString">
                         <div data-bind="compose: 'database/settings/connectionStringRaven.html'"></div>
                         <div class="form-group" data-bind="if: tasksThatAreUsingThisConnectionString().length > 0">
                             <label class="control-label">Used in Tasks:</label> 
                             <div data-bind="foreach: tasksThatAreUsingThisConnectionString">
                                 <div class="task">
-                                    <a data-bind="attr: { 'href': $root.taskEditLink($data.taskId, $parent.connectionStringName()) }" title="Click to edit task.."><span class="text-info small" data-bind="text: $data.taskName"></span></a>
+                                    <a data-bind="attr: { 'href': $root.taskEditLink($data.taskId, $parent.connectionStringName()) }"
+                                       title="Click to edit task.."><span class="text-info small" data-bind="text: $data.taskName"></span></a>
```
ml054, line 118: triple dots?

src/Raven.Studio/wwwroot/App/views/database/settings/connectionStrings.html
```diff
@@ -94,7 +127,20 @@ <h3 class="flex-grow" data-bind="visible: editedSqlEtlConnectionString">SQL Conn
                             <label class="control-label">Used in Tasks:</label>
                             <div data-bind="foreach: tasksThatAreUsingThisConnectionString">
                                 <div class="task"> 
-                                    <a data-bind="attr: { 'href': $root.taskEditLink($data.taskId, $parent.connectionStringName()) }" title="Click to edit task.."><span class="text-info small" data-bind="text: $data.taskName"></span></a>
+                                    <a data-bind="attr: { 'href': $root.taskEditLink($data.taskId, $parent.connectionStringName()) }"
+                                       title="Click to edit task.."><span class="text-info small" data-bind="text: $data.taskName"></span></a>
```
ml054: triple dots?

src/Raven.Studio/wwwroot/App/views/database/settings/connectionStrings.html
```diff
@@ -94,7 +127,20 @@ <h3 class="flex-grow" data-bind="visible: editedSqlEtlConnectionString">SQL Conn
...
+                        </div>
+                    </div>
+                    <div data-bind="with: editedOlapEtlConnectionString">
+                        <div data-bind="compose: 'database/settings/connectionStringOlap.html'"></div>
+                        <div class="form-group" data-bind="if: tasksThatAreUsingThisConnectionString().length > 0">
+                            <label class="control-label">Used in Tasks:</label>
+                            <div data-bind="foreach: tasksThatAreUsingThisConnectionString">
+                                <div class="task">
+                                    <a data-bind="attr: { 'href': $root.taskEditLink($data.taskId, $parent.connectionStringName()) }"
+                                       title="Click to edit task.."><span class="text-info small" data-bind="text: $data.taskName"></span></a>
```
ml054, line 130: triple dots? + $data.taskName -> taskName

src/Raven.Studio/wwwroot/App/views/database/tasks/editOlapEtlTask.html
```diff
@@ -0,0 +1,294 @@
...
+                                </div>
+                            </div>
+                            <div class="form-group">
+                                <label class="control-label">&nbsp;</label>
+                                <div class="toggle keep-files-on-disk">
+                                    <input id="keepFiles" type="checkbox" data-bind="checked: keepFilesOnDisk" />
+                                    <label for="keepFiles">Keep files on disk</label>
+                                </div>
+                            </div>
+                            <div class="form-group">
```
ml054: I wonder if this should go to advanced?

src/Raven.Studio/wwwroot/App/views/database/tasks/editOlapEtlTask.html
```diff
@@ -0,0 +1,294 @@
...
+                                </div>
+                            </div>
+                        </div>
+                        <div data-bind="collapse: $root.showAdvancedOptions">
+                            <div class="panel-body">
+                                <div class="form-group max-items-in-row">
+                                    <label for="maxNumber" class="control-label">Max Items in Row</label>
+                                    <div class="flex-grow" data-bind="validationElement: maxNumberOfItemsInRowGroup">
+                                        <div class="input-group">
+                                            <input type="number" min="1" class="form-control" placeholder="Enter a number (or leave empty for null)" id="maxNumber"
```
ml054: leave empty for null? What null means?

ml054: can we avoid id="maxNumber"?

src/Raven.Studio/wwwroot/App/views/database/tasks/editOlapEtlTask.html
```diff
@@ -0,0 +1,294 @@
...
+            <div class="flex-horizontal flex-header">
+                <div class="flex-grow"><h3>OLAP Tables</h3></div>
+                <div><button type="button" class="btn btn-info" data-bind="click: addNewOlapTable, disable: showEditOlapTableArea()"><i class="icon-plus"></i><span>Add Table</span></button></div>
+            </div>
+            <div class="olap-tables-list margin-bottom">
+                <!-- OLAP table edit area -->
+                <div class="panel" data-bind="visible: showEditOlapTableArea()">
+                    <div class="padding padding-sm flex-horizontal item" data-bind="with: editedOlapTableSandbox()">
+                        <div class="flex-grow" data-bind="validationElement: tableName"> 
+                            <!-- todo 1: why not red on the FIRST Time ? only after saving at least once.. -->
```
ml054: TODO

src/Raven.Studio/wwwroot/App/views/database/tasks/editSqlEtlTask.html
```diff
@@ -312,7 +314,7 @@ <h3 class="text-center text-muted" data-bind="css: { 'text-danger': !transformat
                             <div class="form-group flex-horizontal">
                                 <label class="control-label">Document ID</label>
                                 <div class="flex-grow" data-bind="validationElement: documentId">
-                                    <input type="text" id="documentId" data-bind="textInput: documentId" class="form-control" placeholder="Document ID"
+                                    <input type="text" id="documentId" data-bind="textInput: documentId" class="form-control" placeholder="Document ID" autocomplete="off"
```
ml054, line 317: don't we already have autocomplete off on entire form?

src/Raven.Client/Documents/Operations/ETL/EtlProcessState.cs
aviv86: changed to `DateTime?`

arekpalinski: Yes, let's go with `DateTime?`

src/Raven.Studio/typescript/models/database/tasks/ongoingTaskOlapEtlListModel.ts
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/database/tasks/editOlapEtlTask.ts
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/database/tasks/editOlapEtlTask.ts
Danielle9897: removed

src/Raven.Studio/typescript/viewmodels/database/tasks/editOlapEtlTask.ts
Danielle9897: removed

src/Raven.Studio/typescript/viewmodels/database/tasks/editOlapEtlTask.ts
Danielle9897: done

src/Raven.Studio/wwwroot/App/views/database/tasks/createOngoingTask.html
Danielle9897: done

src/Raven.Studio/wwwroot/App/views/database/tasks/editOlapEtlTask.html
Danielle9897: advanced removed

src/Raven.Studio/wwwroot/App/views/database/tasks/editSqlEtlTask.html
Danielle9897: there is autocomplete on entire 'form' which is on the left side of screen
this was missing on the right side

src/Raven.Studio/wwwroot/App/views/shell/about.html
Danielle9897: @grisha-kotler ?

src/Raven.Client/Documents/Operations/ETL/OLAP/OlapConnectionString.cs
```diff
@@ -0,0 +1,63 @@
...
+        public LocalSettings LocalSettings { get; set; }
+
+        public S3Settings S3Settings { get; set; }
+
+        protected override void ValidateImpl(ref List<string> errors)
+        {
+            if (S3Settings != null)
+            {
+                if (S3Settings.HasSettings() == false)
+                    errors.Add($"{nameof(S3Settings)} has no valid setting");
```
arekpalinski: Shouldn't we tell user exactly what is not valid?

src/Raven.Client/Documents/Operations/ETL/OLAP/OlapConnectionString.cs
```diff
@@ -0,0 +1,63 @@
...
+            }
+
+            if (LocalSettings == null)
+            {
+                errors.Add($"Missing both {nameof(LocalSettings)} and {nameof(S3Settings)}");
+                return;
+            }
+
+            if (LocalSettings.HasSettings() == false)
+                errors.Add($"{nameof(LocalSettings)} has no valid setting");
```
arekpalinski, line 41: Same here - can we provide a user with concrete field error?

src/Raven.Client/Documents/Operations/ETL/OLAP/OlapEtlConfiguration.cs
arekpalinski: Still `NotImplementedException`?

src/Raven.Client/Documents/Smuggler/DatabaseItemType.cs
```diff
@@ -40,20 +40,22 @@ public enum DatabaseRecordItemType
...
-        LockMode = 1 << 18
+        OlapEtls = 1 << 11,
+        Client = 1 << 12,
+        Sorters = 1 << 13,
+        SinkPullReplications = 1 << 14,
+        HubPullReplications = 1 << 15,
+        TimeSeries = 1 << 16,
+        DocumentsCompression = 1 << 17,
+        Analyzers = 1 << 18,
+        LockMode = 1 << 19
```
arekpalinski: Just to make sure - we don't store those enums anywhere (e.g. in an export file)? Asking about because you modified existing values so I wanna be sure it doesn't break the import compatibility of older .ravendump files?

src/Raven.Server/Commercial/LicenseManager.cs
```diff
@@ -1311,6 +1323,12 @@ public void AssertCanAddSqlEtl()
             throw GenerateLicenseLimit(LimitType.SqlEtl, message);
         }
         
+        public void AssertCanAddOlapEtl()
+        {
+            // RavenDB-16242
```
arekpalinski: https://issues.hibernatingrhinos.com/issue/RavenDB-16242 is fixed

src/Raven.Server/Config/Categories/EtlConfiguration.cs
```diff
@@ -41,5 +41,10 @@ public class EtlConfiguration : ConfigurationCategory
         [SizeUnit(SizeUnit.Megabytes)]
         [ConfigurationEntry("ETL.MaxBatchSizeInMb", ConfigurationEntryScope.ServerWideOrPerDatabase)]
         public Size? MaxBatchSize { get; protected set; }
+
+        [Description("Max number of extracted documents in OLAP ETL batch")]
+        [DefaultValue(50 * 1024)]
```
arekpalinski: Maybe let's use power of 2 as the default - 65536?

src/Raven.Server/Documents/ETL/Providers/OLAP/OlapDocumentTransformer.cs
```diff
@@ -0,0 +1,245 @@
...
+                ThrowLoadParameterIsMandatory(nameof(key));
+
+            var result = runnerResult.TranslateToObject(Context);
+            var props = new List<SqlColumn>(result.Count);
+            var prop = new BlittableJsonReaderObject.PropertyDetails();
+
+            for (var i = 0; i < result.Count; i++)
+            {
+                result.GetPropertyByIndex(i, ref prop);
+                props.Add(new SqlColumn
```
arekpalinski: SqlColumn? Maybe create a dedicated class?

src/Raven.Server/Documents/ETL/Providers/OLAP/OlapDocumentTransformer.cs
```diff
@@ -0,0 +1,245 @@
...
+            };
+
+            var transformed = GetOrAdd(tableName, key);
+            transformed.AddItem(olapItem);
+
+            _stats.IncrementBatchSize(result.Size);
+        }
+
+        protected override void AddLoadedAttachment(JsValue reference, string name, Attachment attachment)
+        {
```
arekpalinski, line 114: Shouldn't we throw `NotSupportedException` here?

src/Raven.Server/Documents/ETL/Providers/OLAP/OlapDocumentTransformer.cs
```diff
@@ -0,0 +1,245 @@
...
+                    OlapEtlFileFormat.Parquet => (table = new ParquetTransformedItems(tableName, key, _tmpFilePath, _fileNamePrefix, _config, _logger)),
+                    _ => throw new ArgumentOutOfRangeException(nameof(OlapEtlConfiguration.Format))
+                };
+            }
+
+            return table;
+        }
+
+
+        private JsValue LoadToS3FunctionTranslator(string name, JsValue[] args)
```
arekpalinski: Why does it have "S3" in the name? I think it's not s3 specific, right?

src/Raven.Server/Documents/ETL/Providers/OLAP/OlapDocumentTransformer.cs
```diff
@@ -0,0 +1,245 @@
...
+
+        public override IEnumerable<OlapTransformedItems> GetTransformedResults()  
+        {
+            return _tables.Values;
+        }
+
+        public override void Transform(ToOlapItem item, EtlStatsScope stats, EtlProcessState state)
+        {
+            _stats = stats;
+            if (item.IsDelete)
```
arekpalinski: Given that we don't do anything with deleted docs, do we actually extract tombstones? If yes, then can we skip it completely?

src/Raven.Server/Documents/ETL/Providers/OLAP/OlapEtlMetricsCountersManager.cs
```diff
@@ -0,0 +1,10 @@
+using Raven.Server.Documents.ETL.Metrics;
+
+namespace Raven.Server.Documents.ETL.Providers.OLAP
+{
+    // RavenDB-16309
```
arekpalinski, line 5: Just a reminder about RavenDB-16309

src/Raven.Server/Documents/ETL/Providers/OLAP/OlaptEtl.cs
```diff
@@ -0,0 +1,250 @@
...
+
+        private void UpdateTimerInternal(NextBackup nextRun)
+        {
+            if (Configuration.Disabled)
+                return;
+
+            _timer?.Dispose();
+
+            if (Logger.IsOperationsEnabled)
+                Logger.Operations($"Next OLAP ETL run is in {nextRun.TimeSpan.TotalMinutes} minutes.");
```
arekpalinski: 1. Shouldn't we add `Name` to the message?
2. How often is it going to be called? Asking to make sure it won't flood the logs (since it's on operations level)

src/Raven.Server/Documents/ETL/Providers/OLAP/OlapDocumentTransformer.cs
arekpalinski: I found it later that we do skip extraction of tombstones. Then please add here a comment that will explain that actually this must never happen. You can even add Debug.Assert here

src/Raven.Server/Documents/ETL/Providers/OLAP/OlaptEtl.cs
```diff
@@ -0,0 +1,250 @@
...
+                }
+            };
+
+            var backupUploader = new BackupUploader(uploaderSettings, new RetentionPolicyBaseParameters(), Logger, uploadResult, onProgress: ProgressNotification, _operationCancelToken);
+            backupUploader.Execute();
+        }
+
+        private static void ProgressNotification(IOperationProgress progress)
+        {
+            // todo RavenDB-16341
```
arekpalinski: Just reminder about TODO

src/Raven.Server/Raven.Server.csproj
```diff
@@ -143,6 +143,7 @@
     <PackageReference Include="NuGet.Protocol" Version="5.8.1" />
     <PackageReference Include="Oracle.ManagedDataAccess.Core" Version="3.21.1" />
     <PackageReference Include="Portable.BouncyCastle" Version="1.8.10" />
+    <PackageReference Include="Parquet.Net" Version="3.8.4" />
```
arekpalinski: Make sure we're using the latest `Parquet.Net`

src/Raven.Server/Web/System/OngoingTasksHandler.cs
```diff
@@ -847,6 +863,34 @@ private IEnumerable<OngoingTask> CollectEtlTasks(DatabaseRecord databaseRecord,
...
+                        TaskConnectionStatus = connectionStatus,
+                        TaskState = taskState,
+                        MentorNode = olapEtl.MentorNode,
+                        ResponsibleNode = new NodeId
+                        {
+                            NodeTag = tag,
+                            NodeUrl = clusterTopology.GetUrlFromTag(tag)
+                        },
+                        ConnectionStringName = olapEtl.ConnectionStringName,
+                        //Destination = olapEtl.GetDestination(), // todo ... waiting for RavenDB-16435
```
arekpalinski, line 891: TODO reminder

src/Raven.Studio/typescript/models/database/tasks/ongoingTaskOlapEtlTransformationModel.ts
Danielle9897: done, now support multiple

src/Raven.Server/Documents/ETL/Providers/OLAP/OlaptEtl.cs
aviv86: Added `Name` to the message. 
This will be called once a minute at most

src/Raven.Server/Raven.Server.csproj
aviv86: updated to 3.8.6

src/Raven.Client/Documents/Operations/ETL/OLAP/OlapEtlConfiguration.cs
aviv86: created a separate issue to handle that https://issues.hibernatingrhinos.com/issue/RavenDB-16627

src/Raven.Server/Documents/ETL/Providers/OLAP/OlaptEtl.cs
aviv86: https://issues.hibernatingrhinos.com/issue/RavenDB-16341

src/Raven.Server/Documents/ETL/Providers/OLAP/OlaptEtl.cs
arekpalinski: Once a minute is quite often actually. For me it looks rather like debug logs so I'd change it to `Information` level. What do you think?

src/Raven.Client/Documents/Smuggler/DatabaseItemType.cs
ppekrol: To stay on the safe side, let's revert this to original values and add OlapEtl and OlapConnectionStrings at the end

src/Raven.Server/Config/Categories/EtlConfiguration.cs
```diff
@@ -41,5 +41,10 @@ public class EtlConfiguration : ConfigurationCategory
         [SizeUnit(SizeUnit.Megabytes)]
         [ConfigurationEntry("ETL.MaxBatchSizeInMb", ConfigurationEntryScope.ServerWideOrPerDatabase)]
         public Size? MaxBatchSize { get; protected set; }
+
+        [Description("Max number of extracted documents in OLAP ETL batch")]
+        [DefaultValue(65536)]
```
ppekrol: Sometimes it is just easier to write `64 * 1024`

src/Raven.Server/Documents/ETL/Providers/OLAP/ParquetTransformedItems.cs
```diff
@@ -0,0 +1,712 @@
...
+                    var data = kvp.Value;
+                    Array array;
+
+                    switch (field.DataType)
+                    {
+                        case DataType.Boolean:
+                            array = _boolArr ??= new bool[data.Count];
+                            break;
+                        case DataType.Byte:
+                            array = _byteArr ??= new byte[data.Count];
```
ppekrol, line 137: Should we use ArrayPool for those?

src/Raven.Server/Json/JsonDeserializationServer.cs
```diff
@@ -44,7 +44,11 @@
 using Raven.Server.Web.Studio;
 using Raven.Server.Web.System;
 using Sparrow.Json;
+using Raven.Server.ServerWide;
```
ppekrol: Let's fix those warnings

src/Raven.Server/ServerWide/Commands/ConnectionStrings/PutConnectionStringCommand.cs
```diff
@@ -70,4 +71,33 @@ public override void UpdateDatabaseRecord(DatabaseRecord record, long etag)
...
+        public override void UpdateDatabaseRecord(DatabaseRecord record, long etag)
+        {
+            try
+            {
+                record.OlapConnectionStrings[ConnectionString.Name] = ConnectionString;
+
+            }
+            catch (Exception e)
+            {
+                Console.WriteLine(e);
```
ppekrol: Remove `try catch`

PR #11623, opened 2020-12-31
RavenDB-11014 Rolling Index Deployment

iftahbe:
This is just a draft for review, to see that the general approach is ok and I'm on the right track. 
@karmeli87 I would like your comments please.


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1465,6 +1478,37 @@ protected void ExecuteIndexing()
             }
         }
 
+        private void FinishStaggeredDeploymentForThisNode(IndexDefinitionBase definition)
+        {
+            var nodeTag = DocumentDatabase.ServerStore.NodeTag;
+            var record = DocumentDatabase.ReadDatabaseRecord();
```
ayende: Do we need any error handling here?

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1465,6 +1478,37 @@ protected void ExecuteIndexing()
...
+            if (record.StaggeredIndexes.TryGetValue(definition.Name, out var staggeredIndex) == false)
+                return;
+
+            if (staggeredIndex.TryGetValue(nodeTag, out var staggeredIndexStatus) == false)
+                return;
+
+            if (staggeredIndexStatus != StaggeredIndexStatus.Running)
+                return;
+
+            // todo We need to make sure we only run this command once for each finished node
```
ayende: Why does that matter? what would happen if this runs multiple times?

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1465,6 +1478,37 @@ protected void ExecuteIndexing()
...
+            Task.Run(async () =>
+                {
+                    try
+                    {
+                        await DocumentDatabase.ServerStore.SendToLeaderAsync(new PutStaggeredIndexCommand(record.DatabaseName, Definition.Name, RaftIdGenerator.NewId()))
+                            .ConfigureAwait(false);
+                    }
+                    catch (Exception e)
+                    {
+                        Console.WriteLine(e);
```
ayende: Should remove this.

Note that the current `throw` will _kill the server_ if there is an error here, which is expected.

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -356,6 +362,17 @@ private void HandleChangesForStaticIndexes(DatabaseRecord record, long index, Li
             }
         }
 
+        private bool ShouldSkipThisNode(DatabaseRecord record, string indexName)
+        {
+            if (record.StaggeredIndexes.TryGetValue(indexName, out var staggeredIndexesByNodeTag) == false)
+                return false; // Can this happen? Should we throw or ignore?
```
ayende: Yes, this can happen if there is a race in the record. Shouldn't throw, just start processing it.

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -356,6 +362,17 @@ private void HandleChangesForStaticIndexes(DatabaseRecord record, long index, Li
...
+
+            if (staggeredIndexesByNodeTag.TryGetValue(_serverStore.NodeTag, out var staggeredIndexStatus) == false)
+                return false; // Can this happen? Should we throw or ignore?
```
ayende: Yes, can happen, when we add a node to the database while the stagerred index is running.
Note that we need to decide what to do here, I'm leaning toward starting this immediately, instead.

src/Raven.Server/ServerWide/Commands/Indexes/PutIndexCommand.cs
```diff
@@ -50,6 +58,31 @@ public override string UpdateDatabaseRecord(DatabaseRecord record, long etag)
             return null;
         }
 
+        private bool InitializeStaggeredDeployment(DatabaseRecord record)
+        {
+            // What happens here though? No members.. do we resend the command? How do we make sure this will run again?
+            if (record.Topology.Members.Count == 0)
```
ayende: This can happen if all the members are on rehab, but we can still create new indexes in this mode.
I think that in this case, we need to use AllMembers, not Members.

Need to also take into account the fact that we also have rehab and promotable instances as well.

src/Raven.Server/ServerWide/Commands/Indexes/PutIndexCommand.cs
```diff
@@ -50,6 +58,31 @@ public override string UpdateDatabaseRecord(DatabaseRecord record, long etag)
...
+                return false;
+
+            record.StaggeredIndexes ??= new Dictionary<string, Dictionary<string, StaggeredIndexStatus>>();
+
+            if (record.StaggeredIndexes.TryAdd(Definition.Name, new Dictionary<string, StaggeredIndexStatus>()) == false)
+            {
+                //staggered index already being deployed, no need to add it to indexes again
```
ayende: What happens if I update the definition of an index using staggered deployed?

src/Raven.Server/ServerWide/Commands/Indexes/PutIndexCommand.cs
```diff
@@ -50,6 +58,31 @@ public override string UpdateDatabaseRecord(DatabaseRecord record, long etag)
...
+                return false;
+            }
+
+            // It's guaranteed that topology members are ordered the same on all nodes
```
ayende: Not _quite_, mind. It is the same if they are in sync, they don't have to be.
And generally speaking, we simply need to use one, we don't actually care which.

src/Raven.Server/ServerWide/Commands/Indexes/PutIndexCommand.cs
ayende: What happens if we have just one node?

src/Raven.Server/ServerWide/Commands/Indexes/PutStaggeredIndexCommand.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        public override string UpdateDatabaseRecord(DatabaseRecord record, long etag)
+        {
+            if (record.StaggeredIndexes.TryGetValue(IndexName, out var staggeredIndex ) == false)
+            {
+                return null; // was already removed
+            }
+
+            // What happens here though? No memebers.. do we resend the command? How do we make sure this will run again?
```
ayende: You need to designate someone else, I think.
Alternatively, because we don't modify the database record, the next indexing run will send it again?

src/Raven.Server/ServerWide/Commands/Indexes/PutStaggeredIndexCommand.cs
```diff
@@ -0,0 +1,90 @@
...
+            if (record.Topology.Members.Count == 0)
+                return null;
+
+            var reversedMembers = record.Topology.Members;
+            reversedMembers.Reverse();
```
ayende: Reverse for is better, no?

src/Raven.Server/ServerWide/Commands/Indexes/PutStaggeredIndexCommand.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            // Check if we have new members, coming back from rehab or perhaps completely new
+            foreach (var member in reversedMembers)
+            {
+                if (staggeredIndex.ContainsKey(member) == false)
```
ayende: Why bother with the check, and why does the order matter?
`TryAdd` will skip the addition if it is there, can just scan through this.

src/Raven.Server/ServerWide/Commands/Indexes/PutStaggeredIndexCommand.cs
```diff
@@ -0,0 +1,90 @@
...
+                    staggeredIndex.TryAdd(member, StaggeredIndexStatus.Pending);
+            }             
+            
+            var currentRunningNode = staggeredIndex.Single(x => x.Value == StaggeredIndexStatus.Running).Key;
```
ayende: What happens if we don't have one with running mode?

src/Raven.Server/ServerWide/Commands/Indexes/PutStaggeredIndexCommand.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            if (string.IsNullOrEmpty(currentRunningNode))
```
ayende: This says that you expect it to not fail, but return null, but `Single` will throw if it doesn't find one.
Also, the use of `Single` means that you force a single running instance, we may want to allow more.

src/Raven.Server/ServerWide/Commands/Indexes/PutStaggeredIndexCommand.cs
```diff
@@ -0,0 +1,90 @@
...
+                {
+                    nextNode = member;
+                    break;
+                }
+            }
+
+            if (nextNode != null)
+                staggeredIndex[nextNode] = StaggeredIndexStatus.Running;
+
+            return null;
```
ayende: What happens if this _is_ null?

test/FastTests/Server/Documents/Indexing/Staggered/StaggeredIndexesClusterTests.cs
```diff
@@ -0,0 +1,68 @@
...
+                await leaderStore.Operations.SendAsync(new PatchByQueryOperation(@"FROM Orders UPDATE { for(var i=0; i<100; i++) put(""Orders/"", this); }"));
+
+                var indexDefinition = new IndexDefinition
+                {
+                    Name = "order_companies",
+                    Maps = { "from order in docs.Orders select new { company = order.Company }" },
+                    Staggered = true
+                };
+
+                var db = await cluster.Leader.ServerStore.DatabasesLandlord.TryGetOrCreateResourceStore(dbName);
```
ayende: You aren't actually testing this?

src/Raven.Server/Documents/Indexes/Index.cs
karmeli87: We don't need to materialize the entire db record, but only the `StaggeredIndexes` part.
Take a look at `RawDatabaseRecord` usages.

src/Raven.Server/Documents/Indexes/Index.cs
karmeli87: It would be nice to avoid spamming the state machine and skip the `Task` creation if we don't have leader.
But it should be able to handle running this multiple times.

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1465,6 +1478,37 @@ protected void ExecuteIndexing()
...
+
+            if (staggeredIndexStatus != StaggeredIndexStatus.Running)
+                return;
+
+            // todo We need to make sure we only run this command once for each finished node
+            Task.Run(async () =>
+                {
+                    try
+                    {
+                        await DocumentDatabase.ServerStore.SendToLeaderAsync(new PutStaggeredIndexCommand(record.DatabaseName, Definition.Name, RaftIdGenerator.NewId()))
```
karmeli87: Maybe the `unique raft id` can something like: `<index-name>/<node>/<state>`
IIRC replacement index has the suffix `ReplacementOf`, so this should not collide.

src/Raven.Server/Documents/Indexes/Index.cs
karmeli87: This exception is under the `Task` so it should be fine, I think.
we also ignore it with `IgnoreUnobservedExceptions()`.

src/Raven.Server/Documents/Indexes/IndexDefinitionBase.cs
```diff
@@ -27,6 +27,8 @@ public abstract class IndexDefinitionBase
 
         public IndexPriority Priority { get; set; }
 
+        public bool Staggered { get; set; } = true;
```
karmeli87: Later we need also to consider what should happened in a mix version cluster.

src/Raven.Server/Documents/Indexes/IndexStore.cs
karmeli87: Not sure it can happened TBH, but I'm fine with returning `false` here.

src/Raven.Server/Documents/Indexes/IndexStore.cs
karmeli87: returning `false` here will start the index, no?

src/Raven.Server/ServerWide/ClusterCommandsVersionManager.cs
```diff
@@ -114,7 +114,8 @@ public static class ClusterCommandsVersionManager
             [nameof(ToggleDatabasesStateCommand)] = 42_005,
             [nameof(EditTimeSeriesConfigurationCommand)] = 50_000,
             [nameof(DeleteExpiredCompareExchangeCommand)] = 50_000,
-            [nameof(EditDocumentsCompressionCommand)] = 50_000
+            [nameof(EditDocumentsCompressionCommand)] = 50_000,
+			[nameof(PutStaggeredIndexCommand)] = 51_001
```
karmeli87: I think that we touched the "normal" put index command, since we added the `Staggered` property, which is `true` by default.

src/Raven.Server/ServerWide/Commands/Indexes/PutIndexCommand.cs
karmeli87: The order of the raft commands is guaranteed to be the same on all nodes.
So the order of the members will be the same on each state machine when executing this command.

I think we do want to use a node which isn't the primary to avoid changing the primary frequently if this index will consume a lot of resources.

src/Raven.Server/ServerWide/Commands/Indexes/PutIndexCommand.cs
```diff
@@ -50,6 +58,31 @@ public override string UpdateDatabaseRecord(DatabaseRecord record, long etag)
...
+            var lastMember = record.Topology.Members.Last();
+
+            foreach (var member in record.Topology.Members)
```
karmeli87: What about `Rehab` and `Promotable` ? I think we need to populate them with `Running` ? or is it taken care of in `ShouldSkipNode`?

karmeli87: One thing to consider about `Rehab`/`Promotable` it might be that their mentor doesn't start to run this index, so we promote this node to member. 
Not sure if we want to say that if we have a staggered index, we want to wait for its completion before promoting.

test/FastTests/Server/Documents/Indexing/Staggered/StaggeredIndexesClusterTests.cs
iftahbe: No, this is just a draft and it's not ready yet.
I just wanted to see if the general approach is fine before I write proper tests and error handling.

src/Raven.Server/Documents/Indexes/Index.cs
karmeli87: Do you need to pass the `nodeTag` to the command?

src/Raven.Server/ServerWide/Commands/Indexes/PutStaggeredIndexCommand.cs
```diff
@@ -0,0 +1,90 @@
...
+        public string IndexName { get; set; }
+
+        public PutStaggeredIndexCommand()
+        {
+            // for deserialization
+        }
+
+        public PutStaggeredIndexCommand(string databaseName, string indexName, string uniqueRequestId) 
+            : base(databaseName, uniqueRequestId)
+        {
```
karmeli87: I would expect to see a `NodeTag` here of the finished node.

test/FastTests/Server/Documents/Indexing/Staggered/StaggeredIndexesClusterTests.cs
ayende: Understood, and perfectly fine for draft PR / review / in progress.

Testing this might mean that we insert some hooks to enable that, though.

karmeli87:
@iftahbe I think you have most of the pieces in place. well done :-)

src/Raven.Server/Documents/Indexes/Index.cs
ayende: You generally can't tell if you have a leader or not.
But we don't need a `Task.Run`, do we? We can just run the `SendToLeaderAsync` and not wait on the task?
But we need to handle two error locations then.

src/Raven.Server/Documents/Indexes/IndexDefinitionBase.cs
ayende: We should disable staggered indexes in such a model. 
Also, we need to make this a licensable feature.

src/Raven.Server/Documents/Indexes/IndexStore.cs
ayende: Yes, but the comment should be clearer

src/Raven.Server/ServerWide/Commands/Indexes/PutIndexCommand.cs
ayende: But Rehab / Promotable aren't "live", shouldn't they start running all the indexes ASAP?
And yes, we need to be careful about not moving the mentor to rehab because of that.

src/Raven.Server/ServerWide/Commands/Indexes/PutIndexCommand.cs
arekpalinski: I also think that's not correct. Since we're in initialization method of staggered deployment then we should always modify the status to an initial value

src/Raven.Server/ServerWide/Commands/Indexes/PutStaggeredIndexCommand.cs
```diff
@@ -0,0 +1,90 @@
...
+                return null;
+
+            staggeredIndex[currentRunningNode] = StaggeredIndexStatus.Done;
+            
+            // check if this is the last node
+            if (staggeredIndex.All(x => x.Value == StaggeredIndexStatus.Done))
+            {
+                record.StaggeredIndexes.Remove(IndexName);
+                if (record.Indexes.TryGetValue(IndexName, out var index))
+                    index.Staggered = false;
```
arekpalinski: I wonder if setting it back to `false` is correct. Consider the scenario where additional 2 nodes are be added or the database replication factor is changed from 1 to 3 (but `Staggered` was `true`). I would expect that it will still process the index in staggered manner.

Also I think that this is property of an index which should be exposed e.g. the Studio so I don't think it should be modified afterwards. It's very important information that this index was deployed in staggered manner e.g. foe debugging purposes when analyzing debug package info.

arekpalinski: That also makes me think about having an option to look for history if a staggered index deployment. So we could check that an index was processed on node A, C and then B. 

@karmeli87 What do you think?

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1243,6 +1248,14 @@ protected void ExecuteIndexing()
                                         ResetErrors();
                                         _hadRealIndexingWorkToDo.Raise();
                                     }
+                                    else
+                                    {
+                                        if (Definition.Staggered)
```
arekpalinski: Here you rely on `didWork == false` as an indicator that the indexing is completed, don't you? That's not enough. You need to check the staleness - see `ShouldReplace()` method

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -322,6 +325,9 @@ private void HandleChangesForStaticIndexes(DatabaseRecord record, long index, Li
 
                 try
                 {
+                    if (definition.Staggered && ShouldSkipThisNode(record, name))
```
arekpalinski: It might be better to move the check `definition.Staggered` to the `ShouldSkipThisNode` method. Right now `ShouldSkipThisNode` is meant to be called _only_ if `Staggered` is `true` - that's a bit hidden and implicit. 

Including all such checks in a single method e.g. `ShouldSkip(IndexDefinition def, DatabaseRecord dbRec)` might allow to include more logic in the future if needed - not only for staggered indexes.

src/Raven.Server/ServerWide/Commands/Indexes/PutIndexCommand.cs
```diff
@@ -40,7 +41,14 @@ public override string UpdateDatabaseRecord(DatabaseRecord record, long etag)
                 {
                     throw new InvalidOperationException($"Can not add index: {Definition.Name} because an index with the same name but different casing already exist");
                 }
+
                 record.AddIndex(Definition, Source, CreatedAt, etag);
+
+                if (Definition.Staggered)
+                {
+                    if (InitializeStaggeredDeployment(record) == false) 
+                        return null;
```
arekpalinski: Can `InitializeStaggeredDeployment` method be a `void` method? We return `null` anyway here

src/Raven.Server/ServerWide/Commands/Indexes/PutStaggeredIndexCommand.cs
iftahbe: Yes, it makes sense that the index should remain staggered if it was defined like that. I'll change that.
About keeping history, do you mean to save the order of nodes that the index was deployed on?

src/Raven.Server/ServerWide/Commands/Indexes/PutIndexCommand.cs
iftahbe: Yes.

src/Raven.Server/Documents/Indexes/IndexStore.cs
iftahbe: Yes ok. I actually separated the check from the ShouldSkip function so that it will be easy to understand that it only runs on staggered. But I can also change the name of the function to ShouldSkipThisNodeWhenStaggered.

src/Raven.Server/ServerWide/Commands/Indexes/PutIndexCommand.cs
iftahbe: My initial thinking was if the index was defined to be staggered, it's probably a heavy index. 
Why are nodes in rehab? Probably because they had performance or connection trouble. Should we take these "weak" nodes and throw the index at them? Don't we want to let them recover first, become members, and only then let them start the indexing?
It will be much easier to have them run the index immediately but doesn't that defeat the purpose?

src/Raven.Server/Documents/Indexes/Index.cs
iftahbe: Can you please elaborate on "two error locations"?

src/Raven.Server/ServerWide/Commands/Indexes/PutStaggeredIndexCommand.cs
iftahbe: Yes, the next indexing run will send it again. So the command should not be limited to run only once.

src/Raven.Server/ServerWide/Commands/Indexes/PutIndexCommand.cs
ayende: The problem is that they now go to full members, but don't have this index deployed, that will surprise the users,likely.

src/Raven.Server/Documents/Indexes/Index.cs
ayende: One for sync failure and one for async failure.

src/Raven.Server/ServerWide/Commands/Indexes/PutStaggeredIndexCommand.cs
arekpalinski: Yes, that's what I meant regarding the history. Not sure if we really need this but I was thinking that it might be useful information for debugging.

src/Raven.Server/ServerWide/ClusterCommandsVersionManager.cs
ppekrol: Also it is a 5.2 feature. Make the version 52_000

src/Raven.Server/ServerWide/Commands/Indexes/PutIndexCommand.cs
iftahbe: I still believe that running rehabs immediately defeats the purpose of deploying in a staggered manner and contradicts the "guarantee" of this feature to have each node run separately to maximize the cluster's high availability.

I suggest that if we have rehabs/promotables we can also add them to the staggered dictionary in a pending state but they will be executed last after all the members are done. So in each stage, the finished node chooses a member and if none are left, a rehab node will be chosen. In this way, we don't promote to be a full member until the index is done so no surprise there.

iftahbe:
@ayende here is the progress so far.

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -166,6 +171,22 @@ public abstract class Index : ITombstoneAware, IDisposable, ILowMemoryHandler
...
+        private string _lastPendingReason;
+
+        public bool IsPending
+        {
+            get
+            {
+                if (State == IndexState.Normal)
+                    return DocumentDatabase.IndexStore.ShouldSkipThisNodeWhenRolling(this, out _lastPendingReason);
+
+                _lastPendingReason = $"Not pending due to state: {State}";
```
arekpalinski: Pending reason is 'not pending'?

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1528,6 +1566,46 @@ protected void ExecuteIndexing()
...
+        }
+
+        internal TestingStuff ForTestingPurposes { get; set; }
+
+        internal class TestingStuff
+        {
+            internal Action<RawDatabaseRecord> OnRollingIndexNodeFinished;
+        }
+
+        public bool IsStale()
```
arekpalinski: Why not to use existing `IsStale(...)` methods?

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1528,6 +1566,46 @@ protected void ExecuteIndexing()
...
+        {
+            using (var queryOperationContext = QueryOperationContext.Allocate(DocumentDatabase, this))
+            using (_contextPool.AllocateOperationContext(out TransactionOperationContext indexContext))
+            {
+                using (indexContext.OpenReadTransaction())
+                using (queryOperationContext.OpenReadTransaction())
+                {
+                    indexContext.IgnoreStalenessDueToReduceOutputsToDelete = true;
```
arekpalinski: Why this is _default_ behavior here?

src/Raven.Server/ServerWide/Commands/Indexes/PutRollingIndexCommand.cs
```diff
@@ -0,0 +1,138 @@
...
+            
+            rollingIndex.ActiveDeployments[chosenNode].State = RollingIndexState.Running;
+            rollingIndex.ActiveDeployments[chosenNode].StartedAt = FinishedAt;
+        }
+
+        private static void MaybeAddNewNodesToRollingDeployment(DatabaseRecord record, Dictionary<string, RollingIndexDeployment> rollingIndex)
+        {
+            var allNodes = record.Topology.Members.Select(m => m)
+                .Concat(record.Topology.Promotables.Select(p => p))
+                .Concat(record.Topology.Rehabs.Select(r => r));
```
arekpalinski: Don't we have `topology.AllNodes` for that?

src/Raven.Server/ServerWide/Commands/RemoveNodeFromDatabaseCommand.cs
```diff
@@ -37,6 +39,19 @@ public override void UpdateDatabaseRecord(DatabaseRecord record, long etag)
...
             }
+
+            if (record.RollingIndexes == null)
+                return;
+
+            foreach (var rollingIndex in record.RollingIndexes)
+            {
+                if (rollingIndex.Value.ActiveDeployments.TryGetValue(NodeTag, out var deployment))
+                {
+                    var dummy = new PutRollingIndexCommand(DatabaseName, rollingIndex.Key, NodeTag, DateTime.Now, "dummy update");
```
arekpalinski: DateTime.Now -> DateTime.UtcNow ?

arekpalinski: Why do we actually need this "dummy update"?

src/Raven.Server/Web/Studio/StudioIndexHandler.cs
```diff
@@ -14,6 +14,29 @@ namespace Raven.Server.Web.Studio
 {
     public class StudioIndexHandler : DatabaseRequestHandler
     {
+
+        [RavenAction("/databases/*/studio/index-defaults", "GET", AuthorizationStatus.ValidUser, EndpointType.Read)]
```
arekpalinski, line 18: What do we use this endpoint for?

src/Raven.Server/Documents/Indexes/Index.cs
karmeli87: changed to `_lastPendingStatus`

src/Raven.Server/Documents/Indexes/Index.cs
karmeli87: right, used the the `IsStaleInternal` instead

src/Raven.Server/Documents/Indexes/Index.cs
karmeli87: not relevant, since I use now the `IsStaleInternal` instead

src/Raven.Server/ServerWide/Commands/Indexes/PutRollingIndexCommand.cs
karmeli87: yes

src/Raven.Server/ServerWide/Commands/RemoveNodeFromDatabaseCommand.cs
karmeli87: we use a dummy command to update the record as if the indexing on the removed node was completed.
(added this as a comment as well)

src/Raven.Server/Web/Studio/StudioIndexHandler.cs
karmeli87: this is used by the studio. to get the defaults for the rolling configuration.

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1416,6 +1416,11 @@ internal HttpRequestMessage CreateRequest<TResult>(JsonOperationContext ctx, Ser
                     }
 
                     await ExecuteAsync(node, index, context, command, shouldRetry: true, sessionInfo: sessionInfo, token: token).ConfigureAwait(false);
+                    if (response.Headers.Contains("Rolling-Index") && nodeIndex.HasValue)
+                    {
+                        _nodeSelector.RestoreNodeIndex(nodeIndex.Value);
+                    }
+
```
ayende: I really don't like that this is here. This is mixing of responsibilities.
The request executer shouldn't be aware of such details.

src/Raven.Client/ServerWide/DatabaseRecord.cs
```diff
@@ -203,32 +205,118 @@ public void AddIndex(IndexDefinition definition, string source, DateTime created
...
+                    InitializeRollingDeployment(definition.Name, createdAt, raftIndex);
+            }
+        }
+
+        internal static bool IsRolling(bool? fromDefinition, bool fromSetting)
+        {
+            if (fromDefinition == false)
+                return false;
+
+            return fromSetting || fromDefinition == true;
```
ayende: This looks awkward, what is going on here?

src/Raven.Client/ServerWide/DatabaseRecord.cs
```diff
@@ -203,32 +205,118 @@ public void AddIndex(IndexDefinition definition, string source, DateTime created
...
+        }
+
+        private void InitializeRollingDeployment(string indexName, DateTime createdAt, long raftIndex)
+        {
+            // todo
```
ayende: ??

src/Raven.Client/ServerWide/DatabaseRecord.cs
```diff
@@ -203,32 +205,118 @@ public void AddIndex(IndexDefinition definition, string source, DateTime created
...
+                return false;
+
+            return fromSetting || fromDefinition == true;
+        }
+
+        private void InitializeRollingDeployment(string indexName, DateTime createdAt, long raftIndex)
+        {
+            // todo 
+            // check cluster command versions
+            // what happens if the definition changed but this node didn't start yet? Now we'll process this index but still need to do the replacement. ==> We can think about this later as an optimization
```
ayende: What would happen now?

src/Raven.Client/ServerWide/StaggeredIndex.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
+using System.Collections.Generic;
+using Sparrow.Json.Parsing;
+
+namespace Raven.Client.ServerWide
```
ayende: Need to rename file

src/Raven.Server/Documents/Handlers/IndexHandler.cs
```diff
@@ -48,15 +50,16 @@ public Task Replace()
...
             {
                 Database.IndexStore.ReplaceIndexes(name, newIndex.Name, token.Token);
             }
 
-            return NoContent();
+            if (newIndex.IsRolling)
+            {
+                var command = new PutRollingIndexCommand(Database.Name, name, ServerStore.NodeTag, DateTime.UtcNow, RaftIdGenerator.NewId());
+                await ServerStore.SendToLeaderAsync(command);
+            }
```
ayende, line 69: No need to await for it to be applied?

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -166,6 +171,22 @@ public abstract class Index : ITombstoneAware, IDisposable, ILowMemoryHandler
 
         private bool _didWork;
         private bool _isReplacing;
+        public bool IsRolling => Definition.Rolling ?? DocumentDatabase.Configuration.Indexing.Rolling;
+        public string NormalizedName => Name.Replace(Constants.Documents.Indexing.SideBySideIndexNamePrefix, string.Empty);
+
+        private string _lastPendingStatus;
```
ayende: what is this for?

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -795,13 +816,21 @@ public virtual void Start()
...
+            
             using (DrainRunningQueries())
             {
                 StartIndexingThread();
             }
         }
 
+        private readonly ManualResetEventSlim _rollingEvent = new ManualResetEventSlim();
+
+        public void RollIfNeeded()
```
ayende: when is this called, and why?

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -328,6 +347,17 @@ internal static AutoIndexDefinitionBase CreateAutoDefinition(AutoIndexDefinition
...
         }
 
+        private void RollIndexesIfNeeded()
+        {
+            if (_indexes == null)
+                return;
+
+            foreach (var index in _indexes)
+            {
+                index.RollIfNeeded();
```
ayende, line 361: So the process it that on any document record change, we'll "wake" the pending indexes that are no longer pending?

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -373,6 +407,103 @@ private void HandleChangesForStaticIndexes(DatabaseRecord record, long index, Li
             }
         }
 
+        private void RaiseRollingChangeNotificationIfNeeded(DatabaseRecord record, long index, string name)
+        {
+            if (record.RollingIndexes.TryGetValue(name, out var rollingIndex) && rollingIndex.RaftIndexChange == index)
```
ayende: I'm not following what you are doing here? Why are you checking the `index` ?

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -373,6 +407,103 @@ private void HandleChangesForStaticIndexes(DatabaseRecord record, long index, Li
...
+                return false;
+            }
+
+            if (index.Name == index.NormalizedName)
+            {
+                var latestDef = GetDefinitionFromRecord(record, index.NormalizedName);
+
+                if (index.Definition.Compare(latestDef) == IndexDefinitionCompareDifferences.None)
+                {
+                    reason = "It isn't my turn to be deployed";
```
ayende: What does this mean?

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -629,6 +762,105 @@ public Index GetIndex(string name)
...
+                return; 
+
+            if (index.IsStale())
+                return;
+
+            try
+            {
+                // We may send the command multiple times so we need a new Id every time.
+                var command = new PutRollingIndexCommand(_documentDatabase.Name, definition.Name, nodeTag, DateTime.UtcNow, RaftIdGenerator.NewId());
+                _serverStore.SendToLeaderAsync(command).IgnoreUnobservedExceptions();
```
ayende: Don't we also need logging here if the async failed?

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -731,7 +963,7 @@ public async Task<Index> CreateIndex(IndexDefinitionBase definition, string raft
             long index = 0;
             try
             {
-                index = (await _serverStore.SendToLeaderAsync(command)).Index;
+                index = (await _serverStore.SendToLeaderAsync(command).ConfigureAwait(false)).Index;
```
ayende, line 968: Not needed server side.

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceIndexWriteOperation.cs
```diff
@@ -60,7 +60,7 @@ public override void Commit(IndexingStatsScope stats)
         private async Task CommitOutputReduceToCollection()
         {
             foreach (var command in _outputReduceToCollectionCommandBatcher.CreateCommands())
-                await DocumentDatabase.TxMerger.Enqueue(command);
+                await DocumentDatabase.TxMerger.Enqueue(command).ConfigureAwait(false);
```
ayende, line 63: no need server side

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -1171,7 +1171,21 @@ public void CompleteDeletionIfNeeded(CancellationTokenSource cts)
...
+                            try
+                            {
+                                _server.DatabasesLandlord.DeleteDatabase(dbName, deletionInProgress[_server.NodeTag], record);
+                            }
+                            catch (Exception e)
+                            {
+                                if (_log.IsOperationsEnabled)
+                                {
+                                    _log.Operations("Unexpected error during database deletion from replication loader",e);
+                                }
```
ayende, line 1185: what can cause this?

src/Raven.Server/RavenServerStartup.cs
```diff
@@ -377,6 +379,13 @@ private static void MaybeSetExceptionStatusCode(HttpContext httpContext, ServerS
                 return;
             }
 
+            if (exception is PendingRollingIndexException)
+            {
+                response.StatusCode = (int)HttpStatusCode.Gone;
+                response.Headers["Rolling-Index"] = "true";
+                return;
+            }
+
```
ayende: why do we need a special reply for this?

src/Raven.Server/ServerWide/ClusterCommandsVersionManager.cs
```diff
@@ -78,7 +78,7 @@ public static class ClusterCommandsVersionManager
             [nameof(UpdateSnmpDatabasesMappingCommand)] = Base40CommandsVersion,
             [nameof(DeleteIndexCommand)] = Base40CommandsVersion,
             [nameof(PutAutoIndexCommand)] = Base40CommandsVersion,
-            [nameof(PutIndexCommand)] = Base40CommandsVersion,
+            [nameof(PutIndexCommand)] = Base40CommandsVersion, //todo
```
ayende: ??

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -1173,6 +1174,19 @@ private void RemoveNodeFromCluster(ClusterOperationContext context, BlittableJso
...
                         }
 
+                        if (record.RollingIndexes != null)
+                        {
+                            foreach (var rollingIndex in record.RollingIndexes)
+                            {
+                                if (rollingIndex.Value.ActiveDeployments.TryGetValue(removed, out var deployment))
+                                {
+                                    var dummy = new PutRollingIndexCommand(record.DatabaseName, rollingIndex.Key, removed, DateTime.Now, "dummy update");
+                                    dummy.UpdateDatabaseRecord(record, index);
```
ayende, line 1188: what about adding node to the cluster?

src/Raven.Server/ServerWide/Commands/Indexes/PutRollingIndexCommand.cs
```diff
@@ -0,0 +1,136 @@
...
+
+            var chosenNode = ChooseNextNode(record, rollingIndex.ActiveDeployments);
+            if (chosenNode == null)
+            {
+                if (rollingIndex.ActiveDeployments.All(x => x.Value.State == RollingIndexState.Done))
+                {
+                    if (record.IndexesHistory.TryGetValue(IndexName, out var indexHistoryEntries))
+                    {
+                        if (indexHistoryEntries.Count > 0)
+                            indexHistoryEntries[0].RollingDeployment = rollingIndex.ActiveDeployments;
```
ayende, line 96: what are you doing here?

src/Raven.Server/ServerWide/Commands/Indexes/PutRollingIndexCommand.cs
```diff
@@ -0,0 +1,136 @@
...
+        public string IndexName { get; set; }
+        public string FinishedNodeTag { get; set; }
+        public DateTime FinishedAt { get; set; }
+
+        public PutRollingIndexCommand()
+        {
+            // for deserialization
+        }
+
+        public PutRollingIndexCommand(string databaseName, string indexName, string finishedNodeTag, DateTime finishedAt, string uniqueRequestId)
```
ayende: this is called whenever a rolling index changes, right?

src/Raven.Server/ServerWide/Commands/RemoveNodeFromDatabaseCommand.cs
```diff
@@ -37,6 +39,21 @@ public override void UpdateDatabaseRecord(DatabaseRecord record, long etag)
...
+
+            if (record.RollingIndexes == null)
+                return;
+
+            foreach (var rollingIndex in record.RollingIndexes)
+            {
+                if (rollingIndex.Value.ActiveDeployments.ContainsKey(NodeTag))
+                {
+                    // we use a dummy command to update the record as if the indexing on the removed node was completed
+                    // note: we don't care here that the 'DateTime.UtcNow' will be different on each server
```
ayende: why don't we care about it?
Can we use the leader ticks, instead?

src/Raven.Client/ServerWide/DatabaseRecord.cs
karmeli87: We can set rolling globally or per index definition, which is nullable.
If the definition is set (not null) it override any global setting.
otherwise we use the global setting.

src/Raven.Client/ServerWide/DatabaseRecord.cs
karmeli87: removed

src/Raven.Client/ServerWide/DatabaseRecord.cs
karmeli87: the rolling deployment will start-over with the new definition.
On each node the index is replaced immediately after the it is done the indexing of the side-by-side index, we don't wait for the entire rollout to complete to do the replacement.

src/Raven.Client/ServerWide/StaggeredIndex.cs
karmeli87: done

src/Raven.Server/Documents/Handlers/IndexHandler.cs
karmeli87: I don't think it matters... we replace the index locally anyway.. even we fail here we should notify the cluster about the this from the indexing code at `MaybeFinishRollingDeployment`

src/Raven.Server/Documents/Indexes/Index.cs
karmeli87: debug

src/Raven.Server/Documents/Indexes/IndexStore.cs
karmeli87: yes, it is the easiest way.

src/Raven.Server/Documents/Indexes/IndexStore.cs
karmeli87: This is basically to notify when a deployment of a node was completed.
How would I know on node A that the deployment of node B was done?

src/Raven.Server/Documents/Indexes/IndexStore.cs
karmeli87: we call this directly from a test, which might deadlock the xunit

src/Raven.Server/RavenServerStartup.cs
karmeli87: The idea is to fail-over to a node that is deployed or deploying this index... but on the client we don't want to mark this entire node as faulted, I think.
So I use this header to restore that node from failed state on the client

src/Raven.Server/ServerWide/ClusterStateMachine.cs
karmeli87: haven't checked that yet, it might be interesting with the dynamic node distribution.

src/Raven.Server/ServerWide/Commands/Indexes/PutRollingIndexCommand.cs
karmeli87: every time we change index definition we put it in the indexes history,
here I'm update that history with the deployment.

src/Raven.Server/ServerWide/Commands/Indexes/PutRollingIndexCommand.cs
karmeli87: This is called whenever an index finish its deployment

src/Raven.Server/ServerWide/Commands/RemoveNodeFromDatabaseCommand.cs
karmeli87: leader ticks is the leadership duration. 
we don't care about it (I think) because we removing this node anyway.

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
karmeli87: Anything that can fail the deletion of the database... e.g. server is already disposed at the point, `AccessDenied`, etc..

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceIndexWriteOperation.cs
karmeli87: we call this directly from a test, which might deadlock the xunit

src/Raven.Client/Documents/Indexes/AutoIndexDefinition.cs
```diff
@@ -22,6 +22,8 @@ public AutoIndexDefinition()
 
         public string Collection { get; set; }
 
+        public bool? Rolling { get; set; } = true;
```
ppekrol: Why do we have it in AutoIndexDefinition?

src/Raven.Client/Documents/Indexes/IndexDefinition.cs
```diff
@@ -560,6 +566,7 @@ public enum IndexDefinitionCompareDifferences
         AdditionalSources = 1 << 10,
         AdditionalAssemblies = 1 << 11,
 
-        All = Maps | Reduce | Fields | Configuration | LockMode | Priority | State | AdditionalSources | AdditionalAssemblies
+        All = Maps | Reduce | Fields | Configuration | LockMode | Priority | State | AdditionalSources | AdditionalAssemblies,
+        ReIndexRequiredMask = Maps | Reduce | Fields | Configuration | AdditionalSources | AdditionalAssemblies
```
ppekrol: I would rather avoid this. Its public and more complicated that this. If needed move this to a private field somewhere.

src/Raven.Client/Documents/Indexes/IndexProgress.cs
```diff
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using Raven.Client.ServerWide;
```
ppekrol: It suggests that something is not in an appropriate namespace.

src/Raven.Client/Documents/Indexes/IndexProgress.cs
```diff
@@ -18,6 +19,8 @@ internal class IndexProgress
 
         public double ProcessedPerSecond { get; set; }
 
+        public RollingIndex RollingProgress { get; set; }
```
ppekrol: RollingStatus?

src/Raven.Client/ServerWide/StaggeredIndex.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+    public class RollingIndex : IDynamicJson
+    {
+        public Dictionary<string, RollingIndexDeployment> ActiveDeployments = new Dictionary<string, RollingIndexDeployment>();
```
ppekrol: Probably not needed, but maybe StringComparer.OridinalIgnoreCase?

src/Raven.Client/ServerWide/StaggeredIndex.cs
ppekrol: This should probably be moved into indexes namespace

src/Raven.Server/Documents/Handlers/IndexHandler.cs
```diff
@@ -48,15 +50,16 @@ public Task Replace()
             if (oldIndex == null && newIndex == null)
                 throw new IndexDoesNotExistException($"Could not find '{name}' and '{replacementName}' indexes.");
 
-            if (newIndex == null)
```
ppekrol: It suggests that you can get NRE in ReplaceIndexes few lines below?

src/Raven.Server/Documents/Indexes/Auto/AutoIndexDefinitionBase.cs
```diff
@@ -10,7 +10,7 @@ namespace Raven.Server.Documents.Indexes.Auto
     public abstract class AutoIndexDefinitionBase : IndexDefinitionBase<AutoIndexField>
     {
         protected AutoIndexDefinitionBase(string indexName, string collection, AutoIndexField[] fields, long? indexVersion = null)
-            : base(indexName, new [] { collection }, IndexLockMode.Unlock, IndexPriority.Normal, IndexState.Normal, fields, indexVersion ?? IndexVersion.CurrentVersion)
+            : base(indexName, new [] { collection }, IndexLockMode.Unlock, IndexPriority.Normal, IndexState.Normal, fields, indexVersion ?? IndexVersion.CurrentVersion, rolling: true)
```
ppekrol: Shouldn't this be taken from a configuration?

src/Raven.Server/Documents/Indexes/Auto/AutoMapIndexDefinition.cs
```diff
@@ -84,7 +84,7 @@ public override IndexDefinitionCompareDifferences Compare(IndexDefinitionBase ot
 
         public override IndexDefinitionCompareDifferences Compare(IndexDefinition indexDefinition)
         {
-            return IndexDefinitionCompareDifferences.All;
+            return GetOrCreateIndexDefinitionInternal().Compare(indexDefinition);
```
ppekrol: Why suddenly we need to make this change?

src/Raven.Server/Documents/Indexes/Errors/FaultyAutoIndexDefinition.cs
```diff
@@ -12,7 +12,7 @@ public class FaultyAutoIndexDefinition : IndexDefinitionBase<IndexField>
         public readonly AutoIndexDefinitionBase Definition;
 
         public FaultyAutoIndexDefinition(string name, HashSet<string> collections, IndexLockMode lockMode, IndexPriority priority, IndexState state, IndexField[] mapFields, AutoIndexDefinitionBase definition)
-            : base(name, collections, lockMode, priority, state, mapFields, definition.Version)
+            : base(name, collections, lockMode, priority, state, mapFields, definition.Version, rolling: true)
```
ppekrol: Shouldn't this come from a configuration?

src/Raven.Server/Documents/Indexes/Index.cs
ppekrol: We can remove this at the end?

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1528,6 +1566,23 @@ protected void ExecuteIndexing()
...
+        }
+
+        internal TestingStuff ForTestingPurposes { get; set; }
+
+        internal class TestingStuff
+        {
+            internal Action<RawDatabaseRecord> OnRollingIndexNodeFinished;
+        }
+
+        public bool IsStale() => IsStaleInternal();
```
ppekrol, line 1674: Where this is used?

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -1870,7 +2105,7 @@ public void ReplaceIndexes(string oldIndexName, string replacementIndexName, Can
                                 {
                                     using (IndexLock(newIndex.Name))
                                     {
-                                        newIndex.Stop(disableIndex: true);
+                                        newIndex.Stop(disableIndex: false);
```
ppekrol, line 2130: Why we switched here?

src/Raven.Server/Exceptions/PendingStaggeredIndexException.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+
+namespace Raven.Server.Exceptions
```
ppekrol: This does not have to go to Client?

src/Raven.Server/RavenServerStartup.cs
ppekrol: Do we even want this behavior?

src/Raven.Server/Exceptions/PendingStaggeredIndexException.cs
ppekrol: Also file name is incorrect

src/Raven.Server/ServerWide/Commands/Indexes/PutAutoIndexCommand.cs
```diff
@@ -15,23 +15,26 @@ namespace Raven.Server.ServerWide.Commands.Indexes
 {
     public class PutAutoIndexCommand : UpdateDatabaseCommand
     {
+        public DateTime CreatedAt { get; set; }
```
ppekrol: convert to field?

src/Raven.Server/ServerWide/Commands/Indexes/PutAutoIndexCommand.cs
```diff
@@ -15,23 +15,26 @@ namespace Raven.Server.ServerWide.Commands.Indexes
...
+            CreatedAt = createdAt ?? DateTime.MinValue;
         }
 
         public override void UpdateDatabaseRecord(DatabaseRecord record, long etag)
         {
             try
             {
-                record.AddIndex(Definition);
+                var setting = IndexStore.GetGlobalRollingSetting(record);
+                record.AddIndex(Definition, CreatedAt, etag, setting);
```
ppekrol: Old commands will have DateTime.MinValue, it looks like it is fine, but double-checking.

src/Raven.Client/ServerWide/DatabaseRecord.cs
```diff
@@ -203,32 +205,118 @@ public void AddIndex(IndexDefinition definition, string source, DateTime created
...
+            if (IsRolling(definition.Rolling, globalRollingSetting))
+            {
+                if (differences == null || (differences.Value & IndexDefinitionCompareDifferences.ReIndexRequiredMask) != 0)
+                    InitializeRollingDeployment(definition.Name, createdAt, raftIndex);
+            }
         }
 
         public void AddIndex(AutoIndexDefinition definition)
         {
+            AddIndex(definition, DateTime.UtcNow, 0, false);
```
ppekrol: SystemTime

src/Raven.Server/Documents/Handlers/IndexHandler.cs
```diff
@@ -48,15 +50,16 @@ public Task Replace()
...
-
             using (var token = CreateOperationToken(TimeSpan.FromMinutes(15)))
             {
                 Database.IndexStore.ReplaceIndexes(name, newIndex.Name, token.Token);
             }
 
-            return NoContent();
+            if (newIndex.IsRolling)
+            {
+                var command = new PutRollingIndexCommand(Database.Name, name, ServerStore.NodeTag, DateTime.UtcNow, RaftIdGenerator.NewId());
```
ppekrol: Database.Time.GetUtcNow()

src/Raven.Server/Documents/Indexes/IndexStore.cs
```diff
@@ -629,6 +762,105 @@ public Index GetIndex(string name)
...
+            if (index.Definition.Compare(latestDef) != IndexDefinitionCompareDifferences.None)
+                return; 
+
+            if (index.IsStale())
+                return;
+
+            try
+            {
+                // We may send the command multiple times so we need a new Id every time.
+                var command = new PutRollingIndexCommand(_documentDatabase.Name, definition.Name, nodeTag, DateTime.UtcNow, RaftIdGenerator.NewId());
```
ppekrol: Database.Time.GetUtcNow()

src/Raven.Server/ServerWide/Commands/Indexes/PutAutoIndexCommand.cs
```diff
@@ -43,13 +46,14 @@ public override void UpdateDatabaseRecord(DatabaseRecord record, long etag)
...
             json[nameof(Definition)] = TypeConverter.ToBlittableSupportedType(Definition);
+            json[nameof(CreatedAt)] = TypeConverter.ToBlittableSupportedType(CreatedAt);
         }
 
         public static PutAutoIndexCommand Create(AutoIndexDefinitionBase definition, string databaseName, string raftRequestId)
         {
             var indexType = GetAutoIndexType(definition);
 
-            return new PutAutoIndexCommand(GetAutoIndexDefinition(definition, indexType), databaseName, raftRequestId);
+            return new PutAutoIndexCommand(GetAutoIndexDefinition(definition, indexType), databaseName, raftRequestId, DateTime.UtcNow);
```
ppekrol: SystemTime.UtcNow

src/Raven.Client/ServerWide/StaggeredIndex.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        public long RaftIndexChange { get; internal set; }
```
ppekrol: LastRaftIndex ?

src/Raven.Server/ServerWide/Commands/Indexes/PutRollingIndexCommand.cs
```diff
@@ -0,0 +1,136 @@
...
+        private static void MaybeAddNewNodesToRollingDeployment(DatabaseRecord record, Dictionary<string, RollingIndexDeployment> rollingIndex)
+        {
+            var allNodes = record.Topology.AllNodes;
+
+            foreach (var node in allNodes)
+            {
+                rollingIndex.TryAdd(node, new RollingIndexDeployment
+                {
+                    State = RollingIndexState.Pending, 
+                    CreatedAt = DateTime.UtcNow
```
ppekrol: SystemTime.UtcNow

src/Raven.Server/ServerWide/Commands/RemoveNodeFromDatabaseCommand.cs
```diff
@@ -37,6 +39,21 @@ public override void UpdateDatabaseRecord(DatabaseRecord record, long etag)
...
+                    var dummy = new PutRollingIndexCommand(DatabaseName, rollingIndex.Key, NodeTag, DateTime.UtcNow, "dummy update");
```
ppekrol: SystemTime.UtcNow

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -535,7 +535,11 @@ public void Initialize()
             StorageEnvironmentOptions options;
             if (Configuration.Core.RunInMemory)
             {
-                options = StorageEnvironmentOptions.CreateMemoryOnly();
+                options = StorageEnvironmentOptions.CreateMemoryOnly(
```
ppekrol: Why we need this?

src/Raven.Server/Web/Studio/StudioIndexHandler.cs
ppekrol: Don't we have a configuration endpoint exactly for this?

src/Raven.Client/ServerWide/DatabaseRecord.cs
ayende: So if we have a rolling index *update*, the process goes like this:

* Rolling side-by-side update
* Replaced immediately on each node as it is completed.

We are not doing rolling side by side, right?

src/Raven.Server/Documents/Indexes/Index.cs
ayende: Better to remove it, yes.

src/Raven.Server/Documents/Indexes/IndexStore.cs
ayende: The question is, why are you looking at the `index` of the notification?
What happens if we skipped notifying that particular update and you handle this in a batch?

src/Raven.Server/RavenServerStartup.cs
ayende: No, I don't want that. This is a fairly big violation of the separation of concerns.

I'm fine with having a way to tell the client: "This isn't a real failure, it is transient, don't consider me failed."

But not fitting that particular method.

We can use `421 Misdirected Request` or `425 Too Early` errors here, and teach the client about them.

src/Raven.Server/ServerWide/Commands/Indexes/PutRollingIndexCommand.cs
ayende: why are you overwriting that value?

src/Raven.Server/ServerWide/Commands/RemoveNodeFromDatabaseCommand.cs
ayende: We care that this would be consistent, though.

src/Raven.Server/RavenServerStartup.cs
karmeli87: wanted to use `I'M A TEAPOT` (418) but the dotnet seems to miss that one :)

`Processing` (201) / MisdirectedRequest is not available in standard2.0 
425 is also missing.
Misdirected Request / TemporaryRedirect scares me that the studio may go frenzy with redirections..
Until we find a better one will go with `TemporaryRedirect `

ayende: You can force that using a cast of an int, no?

ayende: `302` is handled by browsers, and sometimes `http clients`, but we want to handle that directly.
We want to have an error code that our code will process

src/Raven.Client/ServerWide/DatabaseRecord.cs
karmeli87: Rolling index update has few scenarios:

1. Old index already deployed:
We will start the side-by-side rolling deployment and as soon a node is finished, we will replace it.

2. The old index is still rolling:
If the old index is pending and hasn't started yet, we will replace it immediately with the new index (whether the new index is still pending or not)  
The running indexes will retain the current logic of side-by-side.

src/Raven.Server/Documents/Indexes/IndexStore.cs
karmeli87: done

src/Raven.Server/ServerWide/Commands/Indexes/PutRollingIndexCommand.cs
karmeli87: I'm not overwriting it... it is simply populates the `RollingDeployment` field of the index history

src/Raven.Server/ServerWide/Commands/RemoveNodeFromDatabaseCommand.cs
karmeli87: okay, will pass a `null` there (since we didn't actually finished it)

src/Raven.Client/Documents/Indexes/AutoIndexDefinition.cs
karmeli87: removed

src/Raven.Client/Documents/Indexes/IndexDefinition.cs
karmeli87: done

karmeli87: done

src/Raven.Client/Documents/Indexes/IndexProgress.cs
karmeli87: indeed. mvoed

src/Raven.Client/ServerWide/StaggeredIndex.cs
karmeli87: done

src/Raven.Client/ServerWide/StaggeredIndex.cs
karmeli87: moved

src/Raven.Server/Documents/Handlers/IndexHandler.cs
karmeli87: reverted this

src/Raven.Server/Documents/Indexes/Auto/AutoIndexDefinitionBase.cs
karmeli87: done

src/Raven.Server/Documents/Indexes/Errors/FaultyAutoIndexDefinition.cs
karmeli87: done

src/Raven.Server/Documents/Indexes/IndexStore.cs
karmeli87: this is core code for this variable:
```
if (disableIndex)
{
    lock (_disablingIndexLock)
    {
        _indexDisabled = true;
        _mre.Set(ignoreThrottling: true);
    }
}
else
{
    _indexingProcessCancellationTokenSource?.Cancel();
}
```

Upon replacing it looks reasonable to me to cancel the thread rather than to disable it.
Also, for pending indexes I _have_ to cancel it, since I'm waiting on a different MRE

src/Raven.Server/Exceptions/PendingStaggeredIndexException.cs
karmeli87: done

src/Raven.Server/ServerWide/Commands/Indexes/PutAutoIndexCommand.cs
karmeli87: done

src/Raven.Server/Web/Studio/StudioIndexHandler.cs
karmeli87: cc @ml054

src/Raven.Server/ServerWide/ServerStore.cs
karmeli87: Otherwise all nodes in the in-memory cluster will use the same dir, which might be a problem when deleting.

src/Raven.Client/Exceptions/PendingRollingIndexException.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+
+namespace Raven.Client.Exceptions
+{
+    public class PendingRollingIndexException : Exception
```
ppekrol: inherit from RavenException?

src/Raven.Client/Documents/Indexes/RollingIndex.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Collections.Generic;
+using Sparrow.Json.Parsing;
+
+namespace Raven.Client.Documents.Indexes
+{
+    public class RollingIndex : IDynamicJson
+    {
+        public Dictionary<string, RollingIndexDeployment> ActiveDeployments = new Dictionary<string, RollingIndexDeployment>(StringComparer.OrdinalIgnoreCase);
+
+        public long RaftIndexChange { get; internal set; }
```
ppekrol: We usually name it `Last...`

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
```diff
@@ -72,6 +73,16 @@ public virtual bool RunInMemory
         [ConfigurationEntry("Indexing.Disable", ConfigurationEntryScope.ServerWideOrPerDatabase)]
         public virtual bool Disabled { get; protected set; }
 
+        [DefaultValue(IndexDeploymentMode.Rolling)]
+        [IndexUpdateType(IndexUpdateType.None)]
+        [ConfigurationEntry("Indexing.StaticIndexDeploymentMode", ConfigurationEntryScope.ServerWideOrPerDatabase)]
```
ppekrol: Indexing.Static.DeploymentMode

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
```diff
@@ -72,6 +73,16 @@ public virtual bool RunInMemory
...
+        public IndexDeploymentMode StaticIndexDeploymentMode { get; protected set; }
+
+        [DefaultValue(IndexDeploymentMode.Rolling)] // TODO: for testing
+        [IndexUpdateType(IndexUpdateType.None)]
+        [ConfigurationEntry("Indexing.AutoIndexDeploymentMode", ConfigurationEntryScope.ServerWideOrPerDatabase)]
```
ppekrol: Indexing.Auto.DeploymentMode

src/Raven.Server/Documents/Indexes/IndexStore.cs
karmeli87: changed the logic, this isn't used any longer.

src/Raven.Server/ServerWide/ClusterCommandsVersionManager.cs
karmeli87: removed

src/Raven.Client/Documents/Indexes/IndexProgress.cs
karmeli87: done

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
karmeli87: done

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
karmeli87: done

src/Raven.Client/ServerWide/DatabaseRecord.cs
karmeli87: done

src/Raven.Server/ServerWide/Commands/RemoveNodeFromDatabaseCommand.cs
karmeli87: passing `null` here

src/Raven.Server/ServerWide/Commands/Indexes/PutRollingIndexCommand.cs
karmeli87: done

src/Raven.Server/Documents/Indexes/IndexStore.cs
karmeli87: done

src/Raven.Server/ServerWide/Commands/Indexes/PutAutoIndexCommand.cs
karmeli87: done

src/Raven.Server/Web/Studio/StudioIndexHandler.cs
ml054: looks like we can use: http://live-test.ravendb.net/databases/db1/admin/configuration/settings

so I can remove endpoint I've added.

src/Raven.Server/Documents/Indexes/Auto/AutoMapIndexDefinition.cs
karmeli87: changed the logic, we can revert this now.

src/Raven.Server/Documents/Handlers/IndexHandler.cs
karmeli87: Now indexes replacement doesn't imply finish the rollout.
To finish the rollout for the giving node we need explicitly to ask it, and in that case we will wait for it to be applied.

src/Raven.Client/Documents/Indexes/RollingIndex.cs
karmeli87: I've remove that entirely

src/Raven.Server/ServerWide/ServerStore.cs
karmeli87: reverted

src/Raven.Client/ServerWide/StaggeredIndex.cs
karmeli87: remove that entirely

src/Raven.Server/Documents/Indexes/IndexStore.cs
karmeli87: remove that entirely.

src/Raven.Server/Web/Studio/StudioIndexHandler.cs
ml054: done

PR #11627, opened 2021-01-03
Feature/sharding

ayende:


src/Raven.Studio/typescript/common/serverNotificationCenterClient.ts
```diff
@@ -1,4 +1,4 @@
-/// <reference path="../../typings/tsd.d.ts" />
+Topl/// <reference path="../../typings/tsd.d.ts" />
```
ml054, line 1: ?

ayende: No idea how it go here :-)

PR #11629, opened 2021-01-03
RavenDB-15978: wait for topology to apply on all nodes on cluster creation in test

garayx:


test/Tests.Infrastructure/ClusterTestBase.cs
```diff
@@ -576,6 +576,8 @@ protected async Task DisposeAndRemoveServer(RavenServer serverToDispose)
                 }
             }
 
+            WaitForClusterTopologyOnAllNodes(clusterNodes);
```
ppekrol: Shouldn't this be an async method? We do have WaitForValueAsync, right?

garayx: done

PR #11630, opened 2021-01-03
RavenDB-16030 Stats View: Fix Change Vector Tooltip

Danielle9897:


src/Raven.Studio/wwwroot/Content/css/icons.less
```diff
@@ -459,7 +459,7 @@
 
 /* MENU AND TOPBAR ICON COLORS*/
 
-#main-menu, .inputs-container, .stats {
+#main-menu, .inputs-container {
```
ml054: why?

src/Raven.Studio/typescript/viewmodels/database/status/statistics.ts
```diff
@@ -42,20 +42,26 @@ class statistics extends viewModelBase {
...
-        popoverUtils.longWithHover($(".js-identities-header"),
+        const cvTooltip = this.stats().databaseChangeVector.map(cv => `<small>${cv.fullFormat}</small>`)
+            .join("<br>");
+
+        popoverUtils.longWithHover($(".cv-tooltip"),
+            {
+                content: `<div>${cvTooltip}</div>`
+            });
+        
+        popoverUtils.longWithHover($(".identities-header"),
```
ml054: why it doesn't work? Old selector seems to be valid.

src/Raven.Studio/wwwroot/App/views/database/status/statistics.html
```diff
@@ -18,7 +18,7 @@ <h2 class="on-base-background">General Database Stats
                 </div>
             </div>
             <div class="col-sm-6 col-lg-4 col-xl-3">
-                <div class="stats-item js-size-tooltip">
+                <div class="stats-item size-tooltip">
```
ml054: why to removed js- prefix?

src/Raven.Studio/wwwroot/Content/css/icons.less
Danielle9897: This fixes the problem that the (i) was purple
With this fix, the icon is back to blue, and it is only purple in the main 'About'

**See comment in**
https://issues.hibernatingrhinos.com/issue/RavenDB-16067#focus=Comments-67-54400.0-0


image

src/Raven.Studio/typescript/viewmodels/database/status/statistics.ts
Danielle9897: Yes, old selector was valid.

I removed it (see line 213) because there is already a selector there (`identities-header`)
that can be used for the tooltip. 
No need to have 2 classes (both `js-identities-header` & `identities-header`) in line 213
So I chose to remove it and use `identities-header` for the tooltip.

src/Raven.Studio/wwwroot/App/views/database/status/statistics.html
Danielle9897: This was not mandatory as well 
I did that only because I removed the `js-` prefix from the class selector in the _previous comment_

src/Raven.Studio/wwwroot/Content/css/icons.less
Danielle9897: So I reverted this change and will update the comment (delete the comment) in issue RavenDB-16067 which still needs to be fixed

ml054 (approved):

PR #11634, opened 2021-01-05
Remote changes base concurrent

danielmarbach:
Alternative to #11578 that would allow concurrent sends

### DIFF
--- a/Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
+++ b/Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
@@ -73,7 +73,7 @@ namespace Raven.Client.Changes
                         {
                             task.AssertNotFailed();
                             return this as TChangesApi;
-                        });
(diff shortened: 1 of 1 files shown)


Raven.Client.Lightweight/Changes/RemoteChangesClientBase.cs
```diff
@@ -201,45 +201,39 @@ private bool DisposeHeartbeatTimer()
...
+                var sendTask = request.ExecuteRequestAsync().ObserveException();
 
-                    return lastSendTask.ContinueWith(task =>
-                    {
-                        lastSendTask = null;
-                        request.Dispose();
-                    });
-                }
-                catch (Exception e)
+                return sendTask.ContinueWith(task =>
```
ayende: Can change this to async and simplify further, no?

danielmarbach: Sure. I was assuming you deliberately wanted to avoid the statemachine on this path

ayende (approved):

PR #11635, opened 2021-01-05
Remote changes base single send simplified

danielmarbach:
Alternative to #11578 that would allow one concurrent send only


ayende (approved):
I like this approach

ayende:
Test failures looks like related to an uninstalled feature on the machine? Will re-run

PR #11636, opened 2021-01-05
RavenDB-16078 Projection with streaming gets us an NRE

ayende:


src/Raven.Server/Documents/Queries/Results/QueryResultRetrieverBase.cs
```diff
@@ -675,7 +675,11 @@ protected bool TryGetValue(FieldsToFetch.FieldToFetch fieldToFetch, Document doc
                 if (_loadedDocuments.TryGetValue(docId, out var doc) == false)
                 {
                     using (_loadScope = _loadScope?.Start() ?? _projectionScope?.For(nameof(QueryTimingsScope.Names.Load)))
+                    {
                         _loadedDocuments[docId] = doc = LoadDocument(docId);
+                        if (doc != null)
+                            doc.IgnoreDispose = true;
```
ppekrol, line 681: I have a feeling that this is too broad. Where is the code that is disposing them? Are we disposing somewhere else the `_loadedDocuments` ? Can it cause some memory pressure now because of lack of dispose?

ayende: That is the responsability of the context, that will be disposed when they are done.

ppekrol: Not always, sometimes we are releasing sooner. Where it was previously disposed?

ayende: This was disposed in the streaming, after we run output the document, we'll dicpose it.

arekpalinski: More precisely it was disposed here:

https://github.com/ravendb/ravendb/blob/d8d91a5ebe53350ac26970ce615fe3ba51337206/src/Raven.Server/Documents/Queries/StreamDocumentQueryResult.cs#L16-L17

PR #11642, opened 2021-01-08
RavenDB-16086  Recovery needs to take into account that a modified pa…

arekpalinski:
…ge could be freed and overwritten in later transactions by a page that was _also_ freed and overwritten and so on. We didn't consider the case that this can happen multiple times when dealing with large overflows.


arekpalinski:
Additional case like the one fixed long time ago in https://github.com/ravendb/ravendb/pull/8313

ayende (approved):

PR #11651, opened 2021-01-11
RavenDB-16000: Mark Changes API connection state as faulted if failed…

garayx:
… to setup Changes API

### DIFF
--- a/src/Raven.Client/Documents/Changes/DatabaseChanges.cs
+++ b/src/Raven.Client/Documents/Changes/DatabaseChanges.cs
@@ -371,9 +371,15 @@ namespace Raven.Client.Documents.Changes
                 return new DatabaseConnectionState(OnConnect, OnDisconnect);
             });
 
-            // try to reconnect
-            if (newValue && Volatile.Read(ref _immediateConnection) != 0)
+            if (_tcs.Task.IsFaulted)
+            {
+                counter.Set(_tcs.Task);
+            }
+            else if (newValue && Volatile.Read(ref _immediateConnection) != 0)
+            {
+                // try to reconnect
                 counter.Set(counter.OnConnect());
+            }
 
             return counter;
         }
--- a/test/SlowTests/Issues/RavenDB-13697.cs
+++ b/test/SlowTests/Issues/RavenDB-13697.cs
@@ -3,6 +3,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using FastTests;
 using Raven.Client.Documents;
+using Raven.Client.Documents.Changes;
 using Raven.Client.Exceptions.Database;
 using Raven.Client.Extensions;
 using Raven.Client.ServerWide;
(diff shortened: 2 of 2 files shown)


src/Raven.Client/Documents/Changes/DatabaseChanges.cs
```diff
@@ -371,9 +371,15 @@ async Task OnConnect()
                 return new DatabaseConnectionState(OnConnect, OnDisconnect);
             });
 
-            // try to reconnect
-            if (newValue && Volatile.Read(ref _immediateConnection) != 0)
+            if (_tcs.Task.IsFaulted)
+            {
+                counter.Set(_tcs.Task);
```
ppekrol, line 376: What is the implication of doing that? Will it reconnect later on?

garayx: in case if there was a failure on setting up database changes (this is only called in ctor), `_tcs.TrySetException(e);` is called:
https://github.com/ravendb/ravendb/blob/5742373db23b1d3800dd332e02ee0c6e2a0b4066/src/Raven.Client/Documents/Changes/DatabaseChanges.cs#L455

then we should set this exception for every connection added for this database changes, otherwise it will hang on `EnsureSubscribedNow`

No it cannot reconnect, because there was exception in setting up the database changes

ppekrol: My assumption would be that we would reconnect, or is it invalid?

garayx: for such exception its invalid and should not reconnect, like in this test the DB doesn't exists.

ppekrol: EnsureSubscribedNow should throw in such case, but in general the Changes API should reconnect by itself? Imagine a user  application that would have to be restarted because of that.

ayende: Yes, we should remove the cached connection from the document store and the next call to Changes should retry.

garayx: this is exactly what happens now with the fix (EnsureSubscribedNow throws) and doesnt hang forever like it was before.

in general yes, but this thrown before the actual connection occurred

ayende: You have the `_changes` field on the `DocumentStore`, but that is only removed when the instance is disposed, but I don't think you are disposing it in this case?

ppekrol: After a discussion with @garayx and @grisha-kotler we think that:
- `EnsureConnectedNow` should throw if there is a problem e.g. DatabaseDoesNotExist (what @garayx did in this PR)
- if `EnsureConnectedNow` is not executed, but Changes API is created, it should reconnect after a while (separate issue - change of behavior from what we have now).

@ayende Thoughts?

ppekrol: Actually if we will throw DatabaseDoesNotExist then user can catch it and react if needed. It is a very rare case. I would leave it be.

PR #11663, opened 2021-01-13
RavenDB-15900 Have a way to remove entry from the raft log

efratshenhar:


src/Raven.Server/Rachis/RachisStateMachine.cs
```diff
@@ -17,7 +17,7 @@ public abstract class RachisStateMachine : IDisposable
         protected TransactionContextPool ContextPoolForReadOnlyOperations;
         protected RachisConsensus _parent;
         public RachisVersionValidation Validator;
-
+        public long skipIndex;
```
ayende: Naming
```suggestion
        public long SkipIndex;
```

karmeli87 (approved):

ayende (changes requested):
I don't understand what it is that you are trying to do.
Note that allowing to skip a single entry is probably a Bad Thing.
To start with, it would be different on each node. 

What is the scenario you are trying to enable? Shouldn't we allow some sort of a "Trim Log" or "Reset Log" ?

What happens if the log was already applied?

karmeli87:
The idea is to have an emergency operation to allow the state machine to skip an entry upon the apply if that command is for some reason is faulted and *jam* the entire state machine across the entire cluster.

Note that this an expert method that the user will have manually to call on all nodes via the admin console.

At this point (after the commit and before the apply) it is guaranteed that the index is identical on every node and executes the same command.

Trimming or resetting is a scary option, since you are rolling back the raft index, and I don't want to find out the implications of that :-)
If the log already applied, trimming or resetting wouldn't help either (you can't undo a command).

ayende:
Try to imagine explaining how to do this for a user?
I don't like it.
And in general, if one command is aborted, all the next one should be.

We can handle that by overwriting all of them with `noop` commands, instead.

Also, there are timing issues to deal with here.

karmeli87:
From the user point of view, It is not different from any other admin operations via the console.

If you need to abort more than one command it has to be done manually, yes.
I don't see a reason to abort the next commands as well..

About timing issues.. if the bad command reached node A & B and was skipped there... and reached only on later stage to node C, then yes, it has to be skipped there as well.

We can overwrite the command, I think, but what would that give us?

karmeli87:
@efratshenhar After discussing with Oren we have decided to replace the command with a `Noop` instead of skipping it.
So newly added nodes will not have to deal with the faulted command.

Also instead of doing it via the JS admin console, we need to create an endpoint for the user, that will send this command to all nodes. (but we need to keep the console option just in case)

ayende (commented):
Shouldn't we also have an endpoint that will hit the other nodes as well?

src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
```diff
@@ -672,5 +668,93 @@ public async Task DemoteNode()
...
+                    {
+                        if (node.Value == Server.WebUrl)
+                        {
+                            ServerStore.Engine.RemoveEntryFromRaftLog(index);
+                            nodeList.Add(node.Key);
+                            continue;
+                        }
+
+                        var cmd = new RemoveEntryFromRaftLogCommand(index);
+                        var requestExecutor = ClusterRequestExecutor.CreateForSingleNode(node.Value, Server.Certificate.Certificate);
```
karmeli87: I think we need to dispose the `requestExecutor`

src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
```diff
@@ -672,5 +668,95 @@ public async Task DemoteNode()
...
+                            nodeList.AddRange(cmd.Result);
+                        }
+                    }
+                }
+                else
+                {
+                    var removed = ServerStore.Engine.RemoveEntryFromRaftLog(index);
+                    if (removed)
+                        nodeList.Add(ServerStore.NodeTag);
+                }
```
ayende: Can simplify the code here.

```suggestion
                    var removed = ServerStore.Engine.RemoveEntryFromRaftLog(index);
                    if (removed)
                          nodeList.Add(ServerStore.NodeTag);

                if (first)
                {

                    foreach (var node in ServerStore.GetClusterTopology(context).AllNodes)
                    {
                        if (node.Value == Server.WebUrl)
                        {
                            continue;
                        }

                        var cmd = new RemoveEntryFromRaftLogCommand(index);
                        using (var requestExecutor = ClusterRequestExecutor.CreateForSingleNode(node.Value, Server.Certificate.Certificate))
                        {
                            await requestExecutor.ExecuteAsync(cmd, context);
                            nodeList.AddRange(cmd.Result);
                        }
                    }
                }
```

src/Raven.Server/Documents/Handlers/Admin/RachisAdminHandler.cs
```diff
@@ -672,5 +668,95 @@ public async Task DemoteNode()
...
+                        if (firstNode)
+                        {
+                            writer.WriteString(nodeTag);
+                            firstNode = false;
+                            continue;
+                        }
+                        writer.WriteComma();
+                        writer.WriteString(nodeTag);
+                    }
+                    writer.WriteEndArray();
```
ayende: Don't we have?

```suggestion
                    writer.WriteArray(nodeList); 
```

ayende (approved):

PR #11666, opened 2021-01-14
RavenDB-16112 clone ReplicationItem properly

karmeli87:


src/Raven.Server/Documents/Replication/ReplicationItems/AttachmentReplicationItem.cs
```diff
@@ -121,13 +122,20 @@ protected override ReplicationBatchItem CloneInternal(JsonOperationContext conte
...
-                Key = Key,
+                ContentType = ContentType.Clone(context),
+                Name = Name.Clone(context),
                 Stream = stream
             };
 
+            var baseMem = Base64Hash.CloneToJsonContext(context, out item.Base64Hash);
+            var keyMem = Key.CloneToJsonContext(context, out item.Key);
+
+            ToDispose(new DisposableAction(() =>
```
ppekrol: Shouldn't this be attached to item.Dispose?

karmeli87: I think it is..

PR #11673, opened 2021-01-17
RavenDB-16110 Export & download in Firefox - file saved as .json

Danielle9897:


src/Raven.Server/Smuggler/Documents/Handlers/SmugglerHandler.cs
```diff
@@ -136,7 +136,8 @@ public async Task PostExport()
 
                 var contentDisposition = "attachment; filename=" + Uri.EscapeDataString(fileName) + ".ravendbdump";
                 HttpContext.Response.Headers["Content-Disposition"] = contentDisposition;
-
+                HttpContext.Response.Headers["Content-Type"] = "text/ravendbdump";
```
ml054: did we already use this somewhere or you create this content type? I wonder if it is more safe to use some generic content type?

Danielle9897: done - changed to `application/octet-stream`

ml054 (approved):

ayende (approved):

PR #11678, opened 2021-01-19
RavenDB-15551 

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-15551

### DIFF
--- a/src/Raven.Client/Documents/Operations/Revisions/GetRevisionsOperation.cs
+++ b/src/Raven.Client/Documents/Operations/Revisions/GetRevisionsOperation.cs
@@ -9,7 +9,6 @@ using System.Collections.Generic;
 using System.Net.Http;
 using Raven.Client.Documents.Commands;
 using Raven.Client.Documents.Conventions;
-using Raven.Client.Extensions;
 using Raven.Client.Http;
 using Raven.Client.Json.Serialization;
 using Sparrow.Json;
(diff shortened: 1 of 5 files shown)


src/Raven.Client/Documents/Operations/Revisions/GetRevisionsOperation.cs
```diff
@@ -0,0 +1,77 @@
...
+using Raven.Client.Json.Serialization.NewtonsoftJson.Internal;
+using Sparrow.Json;
+
+namespace Raven.Client.Documents.Operations.Revisions
+{
+    public class GetRevisionsOperation<T> : IOperation<RevisionsResult<T>>
+    {
+        private readonly GetRevisionsResultCommand _command;
+
+        public GetRevisionsOperation(string id, int? start = 0, int? pageSize = int.MaxValue, bool metadataOnly = false)
```
ppekrol: create two ctors:
1. (string id, int? start = 0, int? pageSize = int.MaxValue) - for convenience
2. (Parameters parameters) - nested class - check other operations how this is done

Introduce validation of the parameters e.g. can id be null?

src/Raven.Client/Documents/Operations/Revisions/GetRevisionsOperation.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            _command = new GetRevisionsResultCommand(id, start, pageSize, metadataOnly);
+        }
+
+        public RavenCommand<RevisionsResult<T>> GetCommand(IDocumentStore store, DocumentConventions conventions, JsonOperationContext context, HttpCache cache)
+        {
+            return _command;
```
ppekrol: I do not recall a single command returning always the same instance. What are the implications? I think it would be better to just create new instance here.

src/Raven.Client/Documents/Operations/Revisions/GetRevisionsOperation.cs
```diff
@@ -0,0 +1,77 @@
...
+                response.TryGet(nameof(RevisionsResult<T>.TotalResults), out int total);
+
+                var results = new List<T>(revisions.Length);
+                foreach (BlittableJsonReaderObject revision in revisions)
+                {
+                    if (revision == null)
+                        continue;
+                    
+                    revision.TryGetId(out var id);
+                    var entity = Converter.FromBlittable<T>(revision, id);
```
ppekrol: That is incorrect. It should use the converter from the store, because user might have overriden the serialization for that particular entity. You have access to that in GetCommand

src/Raven.Client/Documents/Operations/Revisions/GetRevisionsOperation.cs
ayende: Why both default params and nullables?

src/Raven.Client/Documents/Operations/Revisions/GetRevisionsOperation.cs
ayende: Yes

ppekrol (approved):

PR #11680, opened 2021-01-20
RavenDB-16139

ppekrol:


src/Voron/Impl/EncryptionBuffersPool.cs
```diff
@@ -150,11 +146,17 @@ public void Return(byte* ptr, long size, NativeMemory.ThreadStats allocatingThre
             var success = _items[index].TryPush(allocation);
 
             if (success)
+            {
+                // updating the thread allocations since we released the memory back to the pool
+                NativeMemory.UpdateMemoryStatsForThread(allocatingThread, size);
```
grisha-kotler, line 152: Can we have a test for this?

ppekrol: Done. Can you re-check?

grisha-kotler: LGTM

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
```diff
@@ -202,7 +202,7 @@ public virtual bool RunInMemory
         public Size? TransactionSizeLimit { get; protected set; }
 
         [Description("Transaction size limit, for encrypted database only, after which an index will stop and complete the current batch")]
-        [DefaultValue(64)]
+        [DefaultValue(96)]
```
arekpalinski, line 205: Do we want to backport this to 4.2?

ppekrol: Yes, part of https://github.com/ravendb/ravendb/pull/11681

PR #11681, opened 2021-01-21
RavenDB-16139

ppekrol:
- temporary disable Encryption Buffers Pool
- increase Indexing.Encrypted.TransactionSizeLimitInMb to 96


ayende:
Why disable _and_ increase the size limit?

arekpalinski:
It's not necessary to increase the size limit. It's just in case someone have it enabled explicitly.

PR #11686, opened 2021-01-21
RavenDB-16140: Don't redirect to node with failures, add current node…

garayx:
… tag to exception message, increase timeout and wait for responsible node in test


ayende:
What is the scenario where this is relevant?

garayx:
> What is the scenario where this is relevant?

If there is a node with failures (for example offline node), we don't want to redirect the subscription to it

ayende:
Is there an existing test for this?

ppekrol (approved):

PR #11687, opened 2021-01-21
RavenDB-16023 Fixing unable to decrypt page

arekpalinski:


arekpalinski:
Details in https://issues.hibernatingrhinos.com/issue/RavenDB-16023

Sample case (test `RavenDB_16023.RecoveryOfEncryptedStorageNeedsToTakeIntoAccountFreedOverflowPages`):

1. Modify pages in a specific order, they need to be overflows:

  - tx1: allocate #21 (overflow - 10 pages)
  - tx2: free #21
  - tx3: allocate #20 (overflow - 10 pages)
  - tx4: free #20
  - tx5: allocate #21

2. Database restart

3. Recovery on startup: 
  - JournalReader has CryptoPagerTransactionState which has _loadedBuffers dictionary
  - when reading journal we add items to that dictionary
  - on JournalReader.Dispose we encrypt data from the buffers and store it on disk
  - PROBLEM: iterating over dictionary didn't take into account that some of the added overflow pages could be freed in later transactions. so their encryption buffers should have SkipOnTxCommit: true

FIX (1st commit here):
  - On JournalReader.Dispose let's sort the modified pages and iterate over them in reversed order to detect overflow pages that overlap another valid pages modified by later transactions (so same behavior like during checksum validation after the recovery when encryption is turned off - https://github.com/ravendb/ravendb/blob/446de5132301d0816b9cf4f92cc2d514f97b5931/src/Voron/Impl/Journal/WriteAheadJournal.cs#L262-L338)
  
---

Regarding 2nd commit, it fixes the problem that we were encrypting pages even though they were already freed. Potentially it could result in similar overwrite of valid page by an overflow page like above.

arekpalinski:
Referencing https://github.com/ravendb/ravendb/pull/11642 - similar case for non encrypted dbs.

ayende (approved):

PR #11696, opened 2021-01-24
RavenDB-16079 normalize new line characters when comparing maps or reduce

ppekrol:


src/Raven.Client/Util/NewLineInsensitiveStringComparer.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+
+namespace Raven.Client.Util
+{
+    internal sealed class NewLineInsensitiveStringComparer : IEqualityComparer<string>
```
ayende: I don't like this. This allocates multiple times during the equality process.
We are better off just normalizing the values once in `Compare`

ayende (changes requested):

test/SlowTests/Issues/RavenDB_16079.cs
```diff
@@ -0,0 +1,101 @@
...
+using Raven.Client.Documents.Indexes;
+using Raven.Client.Documents.Operations.Indexes;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace SlowTests.Issues
+{
+    public class RavenDB_16079 : RavenTestBase
+    {
+        public RavenDB_16079(ITestOutputHelper output) : base(output)
```
ml054, line 11: I wonder if Git normalizes new line endings during checkout process, so we test X vs X here? Maybe we should replace new line endings on the fly during test execution and then compare?

ml054: https://github.com/ravendb/ravendb/blob/v5.1/.gitattributes#L5

src/Raven.Client/Util/NewLineInsensitiveStringComparer.cs
ppekrol: Simplified

test/SlowTests/Issues/RavenDB_16079.cs
ppekrol: I think that it only normalizes line endings at EOL not in the content?

src/Raven.Client/Documents/Indexes/IndexDefinition.cs
```diff
@@ -17,6 +18,8 @@ namespace Raven.Client.Documents.Indexes
     /// </summary>
     public class IndexDefinition
     {
+        private static readonly Regex _regex = new Regex("\r*\n", RegexOptions.Compiled);
```
ayende:
```suggestion
        private static readonly Regex _regex = new Regex("\r?\n", RegexOptions.Compiled);
```

That is a more correct way to do that, no?

ppekrol: In theory yes, in practice... maybe :) Pushed.

test/SlowTests/Issues/RavenDB_16079.cs
ml054: oh you are right - we use " and \r \n inside stings. Ignore my comment then.

grisha-kotler:
Which index will be saved in the database record?
Old one or new one?

ml054:
I guess old one as new one with equal - no there is "no change" ?

ppekrol:
Correct

PR #11697, opened 2021-01-25
RavenDB-15405 Register load configurations of time series policy runner and the dismiss of it's alert to the license change

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15405

### DIFF
--- a/src/Raven.Server/Documents/DocumentDatabase.cs
+++ b/src/Raven.Server/Documents/DocumentDatabase.cs
@@ -9,6 +9,7 @@ using System.Threading.Tasks;
 using Raven.Client.Documents.Operations.Backups;
 using Raven.Client.Documents.Operations.Configuration;
 using Raven.Client.Documents.Smuggler;
+using Raven.Client.Exceptions.Commercial;
 using Raven.Client.Exceptions.Database;
 using Raven.Client.Extensions;
 using Raven.Client.ServerWide;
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -1358,9 +1364,22 @@ private void InitializeFromDatabaseRecord(DatabaseRecord record)
             DocumentsStorage.DocumentPut.InitializeFromDatabaseRecord(record);
             ExpiredDocumentsCleaner = ExpiredDocumentsCleaner.LoadConfigurations(this, record, ExpiredDocumentsCleaner);
             TimeSeriesPolicyRunner = TimeSeriesPolicyRunner.LoadConfigurations(this, record, TimeSeriesPolicyRunner);
+            _serverStore.LicenseManager.LicenseChanged += LoadTimeSeriesPolicyRunnerConfigurations;
```
ppekrol: How many times this is called? Shouldn't this be registered only once in a lifetime of DocumentDatabase?

ayende: This worries me with the chance of having a reference held here.

haludi: Yes, I moved it to `DocumentDatabase.Initialize`

PR #11702, opened 2021-01-26
RavenDB-14744: Don't move loading idle database to rehab in cluster o…

garayx:
…bserver

### DIFF
--- a/src/Raven.Server/Documents/DatabasesLandlord.cs
+++ b/src/Raven.Server/Documents/DatabasesLandlord.cs
@@ -76,6 +76,8 @@ namespace Raven.Server.Documents
         internal class TestingStuff
         {
             internal Action<ServerStore> BeforeHandleClusterDatabaseChanged;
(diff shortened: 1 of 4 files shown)


grisha-kotler (approved):

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -742,6 +744,10 @@ void AddToInitLog(string txt)
 
                 var sp = Stopwatch.StartNew();
                 documentDatabase = new DocumentDatabase(config.ResourceName, config, _serverStore, AddToInitLog);
+
+                if (ForTestingPurposes?.HoldDocumentDatabaseCreation != null)
+                    Thread.Sleep(ForTestingPurposes.HoldDocumentDatabaseCreation.Value);
```
karmeli87, line 749: This is async, please use `await Task.Delay()`

karmeli87 (approved):

PR #11704, opened 2021-01-26
RavenDB-16172 fix NRE during CompleteDeletionIfNeeded

karmeli87:


src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -863,9 +863,10 @@ public void EnsureNotDeleted(string node)
 
         public void CompleteDeletionIfNeeded(CancellationTokenSource cts)
         {
+            var dbName = Database.Name;
```
ayende, line 866: Do we need to check for null here?

karmeli87: The race is that we execute on the `ThreadPool` after the dispose happened where we set the `Database` to null.
We don't need to check for `null` here because we wait for this thread to end before setting the `Database` to null.

ppekrol (approved):

PR #11705, opened 2021-01-27
RavenDB-15905 Using aggressively cache when load lazily

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15905

### DIFF
--- a/test/SlowTests/Bugs/Caching/UseCachingInLazyTests.cs
+++ b/test/SlowTests/Bugs/Caching/UseCachingInLazyTests.cs
@@ -43,9 +43,9 @@ namespace SlowTests.Bugs.Caching
         {
             using var store = GetDocumentStore();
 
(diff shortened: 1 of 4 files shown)


grisha-kotler (approved):

test/SlowTests/Bugs/Caching/UseCachingInLazyTests.cs
```diff
@@ -0,0 +1,227 @@
...
+            if (createVersion2)
+            {
+                using (var session = store.OpenAsyncSession())
+                {
+                    await session.StoreAsync(new Doc {Version = "2"}, docId);
+                    await session.SaveChangesAsync();
+                }
+            }
+
+            var expected = requestExecutor.NumberOfServerRequests;
```
ppekrol: I think it will be better if you will take the value from session.Advanced.NumberOfRequests, here you might get a topology update so it is racy IMO

PR #11710, opened 2021-01-27
RavenDB-13103 move things around and complete the 'ShardedDocumentHan…

karmeli87:
…dler'

### DIFF
--- a/src/Raven.Server/Documents/ShardedHandlers/ShardedCommands/ShardedBaseCommand.cs
+++ b/src/Raven.Server/Documents/ShardedHandlers/ShardedCommands/ShardedBaseCommand.cs
@@ -1,41 +1,53 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.IO;
 using System.Net;
 using System.Net.Http;
 using System.Threading.Tasks;
 using Raven.Client.Http;
+using Raven.Server.Documents.Sharding;
 using Sparrow.Json;
 
-namespace Raven.Server.Documents.Handlers
+namespace Raven.Server.Documents.ShardedHandlers.ShardedCommands
 {
-    internal class ShardedCommand : RavenCommand<BlittableJsonReaderObject>
+    public abstract class ShardedBaseCommand<T> : RavenCommand<T>
     {
-        public BlittableJsonReaderObject Content;
-        public Dictionary<string, string> Headers = new Dictionary<string, string>();
+        public readonly BlittableJsonReaderObject Content;
+        public readonly Dictionary<string, string> Headers = new Dictionary<string, string>();
         public string Url;
-        public HttpMethod Method;
+        public readonly HttpMethod Method;
 
         public HttpResponseMessage Response;
-        public IDisposable Disposable;
-        public List<int> PositionMatches;
+        
 
         public override bool IsReadRequest => false;
+
+        public ShardedBaseCommand(ShardedRequestHandler handler, Headers headers, BlittableJsonReaderObject content = null)
+        {
+            Method = handler.Method;
+            Url = handler.RelativeShardUrl;
+            Content = content;
+
+            handler.AddHeaders(this, headers);
+        }
+
         public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, ServerNode node, out string url)
         {
             url = $"{node.Url}/databases/{node.Database}{Url}";
             var message = new HttpRequestMessage
             {
-                Method = Method,
+                Method = Method, 
                 Content = Content == null ? null : new BlittableJsonContent(Content),
             };
             foreach ((string key, string value) in Headers)
             {
-                message.Headers.Add(key, value);
+                if (value == null) //TODO sharding: make sure it is okay to skip null
+                    continue;
+
+                message.Headers.TryAddWithoutValidation(key, value);
             }
+
             return message;
         }
-
 
         public override Task<ResponseDisposeHandling> ProcessResponse(JsonOperationContext context, HttpCache cache, HttpResponseMessage response, string url)
         {
@@ -43,10 +55,6 @@ namespace Raven.Server.Documents.Handlers
             return base.ProcessResponse(context, cache, response, url);
         }
 
-        public override void SetResponse(JsonOperationContext context, BlittableJsonReaderObject response, bool fromCache)
-        {
-            Result = response;
-        }
         internal class BlittableJsonContent : HttpContent
         {
             private readonly BlittableJsonReaderObject _data;
@@ -54,7 +62,7 @@ namespace Raven.Server.Documents.Handlers
             public BlittableJsonContent(BlittableJsonReaderObject data)
             {
                 _data = data;
-                
+
             }
 
             protected override Task SerializeToStreamAsync(Stream stream, TransportContext context)
@@ -69,5 +77,12 @@ namespace Raven.Server.Documents.Handlers
                 return false;
             }
         }
+    }
+    
+    public enum Headers
+    {
+        None,
+        IfMatch,
+        IfNonMatch,
     }
 }
--- a/src/Raven.Server/Documents/ShardedHandlers/ShardedDocumentHandler.cs
+++ b/src/Raven.Server/Documents/ShardedHandlers/ShardedDocumentHandler.cs
@@ -1,23 +1,23 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Net;
-using System.Net.Http;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Primitives;
 using Raven.Client;
 using Raven.Client.Documents.Changes;
 using Raven.Client.Documents.Commands;
+using Raven.Server.Documents.ShardedHandlers.ShardedCommands;
 using Raven.Server.Documents.Sharding;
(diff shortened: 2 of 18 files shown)


src/Raven.Server/Documents/ShardedHandlers/ShardedCommands/ShardedBaseCommand.cs
```diff
@@ -70,4 +78,11 @@ protected override bool TryComputeLength(out long length)
             }
         }
     }
+    
+    public enum Headers
+    {
+        None,
+        IfMatch,
+        IfNonMatch,
```
ayende, line 86:
```suggestion
        IfNoneMatch,
```

src/Raven.Server/Documents/ShardedHandlers/ShardedDocumentHandler.cs
```diff
@@ -1,49 +1,60 @@
...
+        public async Task GetDocSize()
+        {
+            var id = GetQueryStringValueAndAssertIfSingleAndNotEmpty("id");
+
+            using (ContextPool.AllocateOperationContext(out TransactionOperationContext context))
+            {
+                var index = ShardedContext.GetShardIndex(context, id);
+
+                var cmd = new ShardedCommand(this, Headers.None);
+                await ShardedContext.RequestExecutors[index].ExecuteAsync(cmd, context);
```
ayende, line 51: I wonder if we need to worry about race conditions with the number of shard changing midway?

src/Raven.Server/Documents/ShardedHandlers/ShardedDocumentHandler.cs
```diff
@@ -58,20 +69,9 @@ public async Task Put()
...
-                {
-                    Method = HttpMethod.Put,
-                    Url = $"/docs?id={Uri.EscapeUriString(id)}",
-                    Content = doc,
-                    Headers =
-                    {
-                        ["If-Match"] = changeVector
-                    }
-                };
+                var cmd = new ShardedCommand(this, Headers.None, content: doc);
```
ayende, line 74: Should `IfMatch` here, no?

src/Raven.Server/Documents/ShardedHandlers/ShardedDocumentHandler.cs
```diff
@@ -124,16 +113,100 @@ public async Task Get()
...
+            var includePaths = GetStringValuesQueryString("include", required: false);
+            var etag = GetStringFromHeaders("If-None-Match");
+
+            using (ContextPool.AllocateOperationContext(out TransactionOperationContext context))
+            {
+                var docs = await context.ReadForMemoryAsync(RequestBodyStream(), "docs");
+                if (docs.TryGet("Ids", out BlittableJsonReaderArray array) == false)
+                    ThrowRequiredPropertyNameInRequest("Ids");
+
+                var ids = new string[array.Length];
```
ayende, line 136: Can we avoid this allocation?

src/Raven.Server/Documents/ShardedHandlers/ShardedDocumentHandler.cs
```diff
@@ -124,16 +113,100 @@ public async Task Get()
...
+                    HttpContext.Response.StatusCode = (int)HttpStatusCode.NotFound;
+                    return;
+                }
+
+                switch (lang)
+                {
+                    case "csharp":
+                        break;
+                    default:
+                        throw new NotImplementedException($"Document code generator isn't implemented for {lang}");
```
ayende, line 192: Shouldn't we just reflect the error back to the user that we got from the destination?

src/Raven.Server/Documents/Sharding/ShardedContext.cs
```diff
@@ -19,6 +19,7 @@ public unsafe class ShardedContext
 
         public ShardedContext(ServerStore server, DatabaseRecord record)
         {
+            //TODO: reduce the record to the needed fields
```
ayende, line 22: Why does this matter?

src/Raven.Server/Documents/Sharding/ShardedContext.cs
```diff
@@ -27,6 +28,7 @@ public ShardedContext(ServerStore server, DatabaseRecord record)
             {
                 var allNodes = server.GetClusterTopology().AllNodes;
                 var urls = record.Shards[i].AllNodes.Select(tag => allNodes[tag]).ToArray();
+                // TODO: pool request executors?
```
ayende, line 31: Don't think we need to. The context is shared among all calls, no?

src/Raven.Server/Documents/Sharding/ShardedRequestHandler.cs
```diff
@@ -28,7 +33,25 @@ public override void Init(RequestHandlerContext context)
             var clientConfigurationEtag = GetLongFromHeaders(Constants.Headers.ClientConfigurationEtag);
             if (clientConfigurationEtag.HasValue && ShardedContext.HasClientConfigurationChanged(clientConfigurationEtag.Value))
                 context.HttpContext.Response.Headers[Constants.Headers.RefreshClientConfiguration] = "true";
-            
+
+            var request = HttpContext.Request;
+            var url = request.Path.Value;
+            var relativeIndex = url.IndexOf('/', 11); //start after "/databases/" and skip the database name
+
+            RelativeShardUrl = url.Substring(relativeIndex) + request.QueryString;
```
ayende, line 41: Don't like this very much. Magic number and string contact?

src/Raven.Server/Extensions/HttpExtensions.cs
```diff
@@ -24,7 +24,7 @@ public static string GetFullUrl(this HttpRequest request)
                 throw new InvalidOperationException("Missing Host");
 
             string path = (request.PathBase.HasValue || request.Path.HasValue) ? (request.PathBase + request.Path).ToString() : "/";
-            return request.Scheme + "://" + request.Host + path + request.Query;
+            return request.Scheme + "://" + request.Host + path + request.QueryString;
```
ayende, line 27: UriBuilder, maybe?

src/Raven.Server/Utils/ShardHelper.cs
```diff
@@ -0,0 +1,48 @@
...
+            int shardIndex = name.IndexOf('$');
+            if (shardIndex != -1)
+            {
+                var slice = name.AsSpan().Slice(shardIndex + 1);
+                if (int.TryParse(slice, out shardIndex) == false)
+                    throw new ArgumentNullException(nameof(name), "Unable to parse sharded database name: " + name);
+            }
+
+            return shardIndex;
+        }
```
ayende, line 32: Duplicate of the above, just have it as another overload that call this

ayende (changes requested):

src/Raven.Server/Documents/ShardedHandlers/ShardedCommands/ShardedBaseCommand.cs
karmeli87: done

src/Raven.Server/Documents/ShardedHandlers/ShardedDocumentHandler.cs
karmeli87: I think, this should be taken care as part of the re-sharding in the replication level? But it is a good point that we need to keep in mind

src/Raven.Server/Documents/ShardedHandlers/ShardedDocumentHandler.cs
karmeli87: yes, done.

src/Raven.Server/Documents/Sharding/ShardedContext.cs
karmeli87: Yes, you are correct.
So we need to update the database record upon each request, I think (or make it lazy?).

ayende: What? Why update it per request?

src/Raven.Server/Documents/ShardedHandlers/ShardedDocumentHandler.cs
ayende: The problem is that we have to take that into account right now
Otherwise, it will mean another set of bugs

Basically, we need to be able to get the request execs and the index as an atomic operation.

karmeli87: I think we need to accept the fact that it is possible to send a document to the wrong shard and handle that properly.
What if you already sent the request and the node re-sharded? we need to handle that as well.

ayende: No, that is something that we'll be handled just fine, I think. We will need two state approach here.
But the process of updates is the hard part, so we need to make sure that this works.

Having to be prepare for a shift each time is too complex and hard.

src/Raven.Server/Documents/Sharding/ShardedContext.cs
karmeli87: I might rely on the recent state of the database record. Right now we materializing it once.. and that's it.

ayende: IIRC, this is updated whenever the database record is.

PR #11714, opened 2021-01-28
RavenDB-15683 Avoid generating index error as a result of server shut down

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-15683

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -1114,7 +1114,8 @@ namespace Raven.Server.Documents.Indexes
                                 }
                                 catch (IndexWriteException iwe)
                                 {
-                                    HandleWriteErrors(scope, iwe);
+                                    if(DocumentDatabase.DatabaseShutdown.IsCancellationRequested == false)
+                                        HandleWriteErrors(scope, iwe);
                                 }
                                 catch (IndexAnalyzerException iae)
                                 {
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -1114,8 +1114,7 @@ namespace Raven.Server.Documents.Indexes
                                 }
                                 catch (IndexWriteException iwe)
                                 {
-                                    if(DocumentDatabase.DatabaseShutdown.IsCancellationRequested == false)
-                                        HandleWriteErrors(scope, iwe);
+                                    HandleWriteErrors(scope, iwe);
                                 }
                                 catch (IndexAnalyzerException iae)
                                 {
@@ -1135,6 +1134,8 @@ namespace Raven.Server.Documents.Indexes
                                 }
                                 catch (OperationCanceledException)
                                 {
+                                    Debug.Assert(_indexingProcessCancellationTokenSource.IsCancellationRequested, $"Got {nameof(OperationCanceledException)} while the index was not canceled");
+
                                     // We are here only in the case of indexing process cancellation.
                                     scope.RecordMapCompletedReason("Operation canceled.");
                                     return;
--- a/src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceIndexWriteOperation.cs
+++ b/src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceIndexWriteOperation.cs
@@ -49,7 +49,7 @@ namespace Raven.Server.Documents.Indexes.MapReduce.OutputToCollection
             }
             catch (ObjectDisposedException e) when (DocumentDatabase.DatabaseShutdown.IsCancellationRequested)
             {
-                throw new TaskCanceledException("A task was canceled", e);
+                throw new TaskCanceledException("The operation of writing output reduce documents was cancelled because of database shutdown", e);
             }
             catch (Exception e)
             {
--- a/src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceIndexWriteOperation.cs
+++ b/src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceIndexWriteOperation.cs
@@ -43,6 +43,14 @@ namespace Raven.Server.Documents.Indexes.MapReduce.OutputToCollection
                     enqueue.GetAwaiter().GetResult();
                 }
             }
+            catch (TaskCanceledException)
+            {
+                throw;
+            }
+            catch (ObjectDisposedException e) when (DocumentDatabase.DatabaseShutdown.IsCancellationRequested)
+            {
+                throw new TaskCanceledException("A task was canceled", e);
+            }
             catch (Exception e)
             {
                 throw new IndexWriteException("Failed to save output reduce documents to disk", e);


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1114,7 +1114,8 @@ protected void ExecuteIndexing()
                                 }
                                 catch (IndexWriteException iwe)
                                 {
-                                    HandleWriteErrors(scope, iwe);
+                                    if(DocumentDatabase.DatabaseShutdown.IsCancellationRequested == false)
```
ppekrol: They way to handle that would be to not wrap OperationCanceledException with IndexWriteException. From my brief look that this only can happen for OutputMapToReduce because we are using TxMerger there (2 places).

```
                            try
                            {
                                enqueue.GetAwaiter().GetResult();
                            }
                            catch (OperationCanceledException)
                            {
                                throw;
                            }
                            catch (Exception e)
                            {
                                throw new IndexWriteException("", e);
                            }
```

The other advantage of doing that is that index will not continue working (it will close down the indexing thread immediately - in your case you will write indexing errors and other stats - which is not what we probably want, etc)

Also please add Debug.Assert to:
```
                                catch (OperationCanceledException)
                                {
                                    Debug.Assert(_indexingProcessCancellationTokenSource.IsCancellationRequested, "msg");

                                    // We are here only in the case of indexing process cancellation.
                                    scope.RecordMapCompletedReason("Operation canceled.");
                                    return;
                                }
```

haludi: There can be index error also for `ObjectDisposedException` when the `OutputReduceIndexWriteOperation` try to add operation to disposed `TransactionOperationsMerger`

ppekrol: Indexing should be stopped way before TxMerger is disposed.

haludi: To get exception from awating to `CommitOutputReduceToCollection` here - https://github.com/ravendb/ravendb/blob/d22e2b6c20e5e989f9eba4fe434047a5864cd911/src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceIndexWriteOperation.cs#L43
there are two options:
1. The last `OutputReduceToCollectionCommand` in progress canceled - this will throw `TaskCanceledException`.
2. When trying to enque a new `OutputReduceToCollectionCommand` - this for now throwing `ObjectDisposedException`

The `TransactionOperationsMerher.Enque` throw `ObjectDisposedException` when the `_concurrentOperations` is zero.
In the `TransactionOperationsMerher.Dispose` we first decrement `_concurrentOperations` to zero and only then iterate all the running operations in the `TransactionOperationsMerher` and canceling them.

I can add in `CommitOutputReduceToCollection` a check before enqueuing a new `OutputReduceToCollectionCommand` if the `ServerShutdown` canceled but there is still an option the code will pass the check and then `_concurrentOperations` will decrement to zero.

Another option is to throw 'TaskCanceledException` from `TransactionOperationsMerher.Enque`.

grisha-kotler: It's impossible to get the `ObjectDisposed` exception since we are waiting for the indexing thread to finish before we dispose the `TxMerger`.
https://github.com/ravendb/ravendb/blob/e817bffab9ad18de7ecbfac1efb6a0f6f1f553fc/src/Raven.Server/Documents/Indexes/Index.cs#L298
https://github.com/ravendb/ravendb/blob/e817bffab9ad18de7ecbfac1efb6a0f6f1f553fc/src/Raven.Server/Documents/Indexes/Index.cs#L309

haludi: Actually, when I tried to reproduce it I found it more difficult to get `TaskCanceledException`.
Just create an index with enough documents for multiple batches of output reduce to collection.
Put breakpoint here https://github.com/ravendb/ravendb/blob/d22e2b6c20e5e989f9eba4fe434047a5864cd911/src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceIndexWriteOperation.cs#L43
and when it got there just press 'q' and 'y' in the console to shut down the server, continue the debugger and you will get `ObjectDisposedException`.

haludi: I added the RavenDB dump file to reproduce if you like https://issues.hibernatingrhinos.com/issue/RavenDB-15683#focus=Comments-67-55584.0-0

haludi: The main point here is the `ObjectDisposedException` is thrown base on `_concurrentOperations` and we use it to block enqueuing more operations to `TransactionOperationsMerher`.

arekpalinski: Interesting. Can you share the actual stacktrace of `ObjectDisposedExecption`? Maybe we have invalid order of disposals

haludi:
```
Raven.Server.dll!Raven.Server.Documents.TransactionOperationsMerger.ThrowTxMergerWasDisposed() Line 150
Raven.Server.dll!Raven.Server.Documents.TransactionOperationsMerger.Enqueue(Raven.Server.Documents.TransactionOperationsMerger.MergedTransactionCommand cmd) Line 129
Raven.Server.dll!Raven.Server.Documents.Indexes.MapReduce.OutputToCollection.OutputReduceIndexWriteOperation.CommitOutputReduceToCollection() Line 59	
```

ppekrol: It looks like the order of disposal is invalid. We first are disposing the merger, then subsystems that are actually using it. From my brief look we should dispose merger last (just before _databaseShutdown.Dispose)

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceIndexWriteOperation.cs
```diff
@@ -60,7 +56,10 @@ public override void Commit(IndexingStatsScope stats)
         private async Task CommitOutputReduceToCollection()
         {
             foreach (var command in _outputReduceToCollectionCommandBatcher.CreateCommands())
+            {
+                DocumentDatabase.DatabaseShutdown.ThrowIfCancellationRequested();
```
arekpalinski: Do we need to call it explicitly here? Won't we get `TaskCancelledException` from `DocumentDatabase.TxMerger.Enqueue(command);` anyway?

haludi: But we are disposing the "TxMerger" last so why will we get `TaskCancelledException`?

arekpalinski: Yeah, we're disposing it last but the database shutdown event is raised first:

https://github.com/ravendb/ravendb/blob/07f16e35386b93ddef7149b746e57734aec753b6/src/Raven.Server/Documents/DocumentDatabase.cs#L610

so I expected that an already enqueued command will result in the operation cancelled exception anyway (I haven't checked the code).

arekpalinski (approved):

src/Raven.Server/Documents/Indexes/Index.cs
haludi: https://issues.hibernatingrhinos.com/issue/RavenDB-16243

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceIndexWriteOperation.cs
haludi: https://issues.hibernatingrhinos.com/issue/RavenDB-16243

src/Raven.Server/Documents/Indexes/MapReduce/OutputToCollection/OutputReduceIndexWriteOperation.cs
```diff
@@ -43,6 +43,14 @@ public override void Commit(IndexingStatsScope stats)
                     enqueue.GetAwaiter().GetResult();
                 }
             }
+            catch (TaskCanceledException)
+            {
+                throw;
+            }
+            catch (ObjectDisposedException e) when (DocumentDatabase.DatabaseShutdown.IsCancellationRequested)
+            {
+                throw new TaskCanceledException("A task was canceled", e);
```
arekpalinski: Can have something meaningful here - e.g. "The operation of writing output reduce documents was cancelled because of database shutdown"?

arekpalinski (approved):

ppekrol (approved):

PR #11715, opened 2021-01-28
RavenDB-16178: fix test

garayx:


ppekrol:
@garayx It still fails

PR #11726, opened 2021-02-01
RavenDB-16170: log UpdatePeriodicBackupStatusCommand failure to BackupResult

garayx:


src/Raven.Server/Documents/PeriodicBackup/BackupTask.cs
```diff
@@ -876,6 +876,8 @@ public static void SaveBackupStatus(PeriodicBackupStatus status, DocumentDatabas
 
                 if (logger.IsOperationsEnabled)
                     logger.Operations(message, e);
+
+                backupResult?.AddError($"{message}{Environment.NewLine}{e}");
```
ayende, line 880: Is this thread safe?

ppekrol: It does not have to be? cc @garayx

ayende: Don't we have the potential for reading while this is going on?

garayx: yes it should be added under lock, as we do here:
https://github.com/ravendb/ravendb/blob/v5.1/src/Raven.Server/Documents/PeriodicBackup/BackupTask.cs#L717-L724

ayende: Consider the case where this is being _read_ at the same time, no?

ppekrol: The question is if this can happen. @garayx @grisha-kotler ?

grisha-kotler: It can be read at the same time:
https://github.com/ravendb/ravendb/blob/v5.1/src/Raven.Client/Documents/Smuggler/SmugglerResult.cs#L86

This is an issue for import and export as well

PR #11727, opened 2021-02-01
RavenDB-14748 Race condition in the test fixed and asserts changed to be more specific specific

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-14748

Without waiting to index with documents to index 
the assert can be called before the first assignment of `LastIndexingTime`  in `Index.ExecuteIndexing`


ppekrol:
@haludi Can you check failing test?

PR #11728, opened 2021-02-01
RavenDB-16176 Remote Database Info in Databases List View

Danielle9897:


src/Raven.Studio/typescript/viewmodels/resources/databases.ts
```diff
@@ -689,7 +692,7 @@ class databases extends viewModelBase {
         return true; // don't prevent default action as we have links inside links
     }
 
-    createIsLocalDatabaseObservable(dbName: string) {
+    isLocalDatabase(dbName: string) {
```
ml054: leave method name but call isLocalDatabase inside, and use isLocalDatabase https://github.com/ravendb/ravendb/pull/11728/files#diff-8664c88b6171062c2f59d23aadc2e5898cd1cd14c864a3c1efbdc54cfec87612R119 here instead of creating purecomputed and calling this.

Danielle9897: done

ml054 (approved):

PR #11736, opened 2021-02-03
RavenDB-16170: Use ConcurrentQueue in SmugglerResult

garayx:
…pResult

### DIFF
--- a/src/Raven.Server/Documents/PeriodicBackup/BackupTask.cs
+++ b/src/Raven.Server/Documents/PeriodicBackup/BackupTask.cs
@@ -738,11 +738,8 @@ namespace Raven.Server.Documents.PeriodicBackup
 
         private void AddInfo(string message)
         {
-            lock (this)
-            {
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Documents/PeriodicBackup/BackupTask.cs
```diff
@@ -745,6 +747,15 @@ private void AddInfo(string message)
             }
         }
 
+        private void AddError(string message)
+        {
+            lock (this)
```
grisha-kotler: There is no need for the lock here.
The reason that we added the lock in `AddInfo` is that there could be concurrent writers (concurrent writes to the cloud).
At this point, there is only a single thread that can write the errors.

ayende: what about reads?

garayx: changed List to ConcurrentQueue

PR #11744, opened 2021-02-04
RavenDB-13103 sharded bulk docs

karmeli87:


src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -44,87 +41,91 @@ public class BatchHandler : DatabaseRequestHandler
...
+                    await commandBuilder.ParseMultipart(context, RequestBodyStream(), HttpContext.Request.ContentType);
                 }
                 else
                     ThrowNotSupportedType(contentType);
 
                 if (TrafficWatchManager.HasRegisteredClients)
                 {
-                    BatchTrafficWatch(command.ParsedCommands);
+                    var log = BatchTrafficWatch(commandBuilder.Commands);
+                    // add sb to httpContext
```
ayende, line 68: TODO missing here?

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
```diff
@@ -0,0 +1,355 @@
...
+                }
+
+                using (var batch = await commandBuilder.GetCommand(context))
+                {
+                    if (batch.IsClusterTransaction)
+                    {
+                        throw new NotSupportedException("Cluster transactions are currently not supported in a sharded database.");
+                    }
+
+                    var shardedBatchCommands = new Dictionary<int, SingleNodeShardedBatchCommand>();
```
ayende: I wonder if we can / should cache these.

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
```diff
@@ -0,0 +1,355 @@
...
+                        if (shardedBatchCommands.ContainsKey(shardIndex) == false)
+                        {
+                            shardedBatchCommands.Add(shardIndex, new SingleNodeShardedBatchCommand(this, requestExecutor.ContextPool));
+                        }
+
+                        var shardedBatchCommand = shardedBatchCommands[shardIndex];
+                        shardedBatchCommand.AddCommand(command);
+                    }
+
+                    var tasks = shardedBatchCommands.Select(x => ShardedContext.RequestExecutors[x.Key].ExecuteAsync(x.Value, x.Value.Context));
```
ayende: Let's avoid Linq in this code path

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
```diff
@@ -0,0 +1,355 @@
...
+                    }
+
+                    var shardedBatchCommands = new Dictionary<int, SingleNodeShardedBatchCommand>();
+                    foreach (var command in batch)
+                    {
+                        var id = command.Id;
+                        var shardIndex = ShardedContext.GetShardIndex(context, id);
+                        var requestExecutor = ShardedContext.RequestExecutors[shardIndex];
+
+                        if (shardedBatchCommands.ContainsKey(shardIndex) == false)
```
ayende: TryGetValue, instead of two calls

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
```diff
@@ -0,0 +1,355 @@
...
+
+                        var shardedBatchCommand = shardedBatchCommands[shardIndex];
+                        shardedBatchCommand.AddCommand(command);
+                    }
+
+                    var tasks = shardedBatchCommands.Select(x => ShardedContext.RequestExecutors[x.Key].ExecuteAsync(x.Value, x.Value.Context));
+                    await Task.WhenAll(tasks);
+
+                    HttpContext.Response.StatusCode = (int)HttpStatusCode.Created;
+                    using (var writer = new BlittableJsonTextWriter(context, ResponseBodyStream()))
```
ayende, line 87: How do we ensure that we get the results in the same order that the user asked for them?

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
```diff
@@ -0,0 +1,355 @@
...
+            if (command.AttachmentStream != null)
+            {
+                var stream = command.AttachmentStream;
+                if (_attachmentStreams == null)
+                {
+                    _attachmentStreams = new List<Stream>();
+                    _uniqueAttachmentStreams = new HashSet<Stream>();
+                }
+
+                PutAttachmentCommandHelper.ValidateStream(stream);
```
ayende: Why is this required?

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
```diff
@@ -0,0 +1,355 @@
...
+        public ShardedBatchCommandBuilder(RequestHandler handler, string database, char identityPartsSeparator) :
+            base(handler, database, identityPartsSeparator)
+        {
+        }
+
+        public override async Task SaveStream(JsonOperationContext context, Stream input)
+        {
+            Streams ??= new List<Stream>();
+
+            var ms = new MemoryStream();
```
ayende: What if this is a big attachments? IIRC, we are writing these to disk , no?

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
karmeli87: This is actually a good point. I know it is a requirement for `Load` (need to check if we do that properly there as well)
But is that a requirement for bulk docs as well?

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
karmeli87: Yes, in the normal code path it is coupled with a specific database, so we use `_database.DocumentsStorage.AttachmentsStorage.GetTempFile` to store it.
Here we don't have it, so we should use the System temp path?

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
karmeli87: It doesn't cost us anything and I think we actually need it to be able to put it on the network stream.
```
public static void ValidateStream(Stream stream)
        {
            if (stream.CanRead == false)
                ThrowNotReadableStream();
            if (stream.CanSeek == false)
                ThrowNotSeekableStream();
            if (stream.Position != 0)
                ThrowPositionNotZero(stream.Position);
        }
```

src/Raven.Server/Documents/Handlers/BatchHandler.cs
karmeli87: No, this is the normal batch handler.

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
karmeli87: will add a todo

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
karmeli87: done

src/Raven.Server/Documents/Handlers/BatchHandler.cs
ayende: So why added the comment?

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
ayende: It is a *hard* requirement for bulk docs as well.
On the client side, we are assuming that the order of results match the order of commands.

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
ayende: That ends up here: 
`_documentsStorage.Environment.Options.DataPager.Options.TempPath.Combine(name);`

So we either have a `TempPath` in the database configuration or use the default one.
Note that we want one per sharded db, so we'll create a (mostly) empty directory for that.

karmeli87: I think we might not have this info, since we need to the `StorageConfiguration` which we might not have.
So I put it in the server temp folder.

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
karmeli87: done by marking for each request the position in the response

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
```diff
@@ -0,0 +1,621 @@
...
+
+                        if (shardedBatchCommands.TryGetValue(shardIndex, out var shardedBatchCommand) == false)
+                        {
+                            shardedBatchCommand = new SingleNodeShardedBatchCommand(this, requestExecutor.ContextPool);
+                            shardedBatchCommands.Add(shardIndex, shardedBatchCommand);
+                        }
+
+                        shardedBatchCommand.AddCommand(command);
+                    }
+
```
ayende, line 77: Let's create an issue here to add an option to reject multi shard transactions

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
```diff
@@ -0,0 +1,621 @@
...
+                    var tasks = new List<Task>();
+                    foreach (var command in shardedBatchCommands)
+                    {
+                        tasks.Add(ShardedContext.RequestExecutors[command.Key].ExecuteAsync(command.Value, command.Value.Context));
+                    }
+
+                    await Task.WhenAll(tasks);
```
ayende, line 84: Let's create an issue here, we need to have better error handling in case of partial success, we want to report that to the user.

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
```diff
@@ -0,0 +1,621 @@
...
+            await attachment.FlushAsync(Handler.AbortRequestToken);
+            Streams.Add(attachment);
+        }
+
+        public StreamsTempFile GetServerTempFile(string prefix)
+        {
+            var name = $"{_shardedContext.DatabaseName}.attachment.{Guid.NewGuid():N}.{prefix}";
+            var tempPath = ServerStore._env.Options.DataPager.Options.TempPath.Combine(name);
+            
+            return new StreamsTempFile(tempPath.FullPath, _encrypted);
```
ayende, line 334: Let's create an issue to worry about encrypted sharded databases here, we need to make sure that we'll handle this properly and not write plain text data here. Would also give us a chance to have a temp directory for the sharded db. 

Another thought, do we have a way to control the _paths_ of the sharded dbs?

ayende (approved):

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
karmeli87: https://issues.hibernatingrhinos.com/issue/RavenDB-16276

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
karmeli87: same issue as above?
https://issues.hibernatingrhinos.com/issue/RavenDB-16276

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
karmeli87: Right now the shards inherit the path from database record.

Opened an issue about encryption https://issues.hibernatingrhinos.com/issue/RavenDB-16277
Not sure what should we consider here.

PR #11748, opened 2021-02-06
RavenDB-16209 Avoid calling TryGetServerVersion every time.

redknightlois:
The call to TryGetServerVersion is a bit expensive, we are trying to avoid calling it as much as possible.

### DIFF
--- a/src/Raven.Client/Http/RequestExecutor.cs
+++ b/src/Raven.Client/Http/RequestExecutor.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
@@ -82,6 +82,8 @@ namespace Raven.Client.Http
 
         private ServerNode _topologyTakenFromNode;
 
+        internal string LastServerVersion { get; private set; }
+
         private HttpClient _httpClient;
 
         public HttpClient HttpClient
@@ -131,8 +133,6 @@ namespace Raven.Client.Http
         protected bool _disableTopologyUpdates;
 
         protected bool _disableClientConfigurationUpdates;
-
-        internal string LastServerVersion;
 
         public TimeSpan? DefaultTimeout
         {
@@ -1051,8 +1051,18 @@ namespace Raven.Client.Http
 
             var response = await preferredTask.ConfigureAwait(false);
 
-            if (TryGetServerVersion(response, out var serverVersion))
-                LastServerVersion = serverVersion;
+            // PERF: The reason to avoid rechecking every time is that servers wont change so rapidly
+            //       and therefore we dimish its cost by orders of magnitude just doing it
+            //       once in a while. We dont care also about the potential race conditions that may happen
+            //       here mainly because the idea is to have a lax mechanism to recheck that is at least
+            //       orders of magnitude faster than currently. 
+            if (chosenNode.ShouldUpdateServerVersion())
+            {
+                if (TryGetServerVersion(response, out var serverVersion))
+                    chosenNode.UpdateServerVersion(serverVersion);                    
+            }
+
+            LastServerVersion = chosenNode.LastServerVersion;
 
             if (sessionInfo?.LastClusterTransactionIndex != null)
             {
@@ -1060,7 +1070,7 @@ namespace Raven.Client.Http
                 // Since the current executed command can be dependent on that, we have to wait for the cluster transaction.
                 // But we can't do that if the server is an old one.
 
-                if (serverVersion == null || string.Compare(serverVersion, "4.1", StringComparison.Ordinal) < 0)
+                if (LastServerVersion == null || string.Compare(LastServerVersion, "4.1", StringComparison.Ordinal) < 0)
                 {
                     using (response)
                     {
@@ -1475,6 +1485,9 @@ namespace Raven.Client.Http
                 SpawnHealthChecks(chosenNode, nodeIndex.Value);
                 return false;
             }
+
+            // As the server is down, we discard the server version to ensure we update when it goes up. 
+            chosenNode.DiscardServerVersion();
 
             _nodeSelector.OnFailedRequest(nodeIndex.Value);
 
--- a/src/Raven.Client/Http/ServerNode.cs
+++ b/src/Raven.Client/Http/ServerNode.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using static Sparrow.Hashing;
 
 namespace Raven.Client.Http
 {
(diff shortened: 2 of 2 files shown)


src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1051,16 +1052,23 @@ private Task RefreshIfNeeded(ServerNode chosenNode, HttpResponseMessage response
 
             var response = await preferredTask.ConfigureAwait(false);
 
-            if (TryGetServerVersion(response, out var serverVersion))
-                LastServerVersion = serverVersion;
+            // PERF: The reason to avoid rechecking every time is that servers wont change so rapidly
+            //       and therefore we dimish its cost by several orders of magnitude just doing it
+            //       once in a while. 
+            if (LastServerVersion == null || _lastServerVersionCheck > (1 << 18))
```
ayende: Huh? Why is the `1<<18` for?

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1051,16 +1052,23 @@ private Task RefreshIfNeeded(ServerNode chosenNode, HttpResponseMessage response
...
+            {
+                if (TryGetServerVersion(response, out var serverVersion))
+                    LastServerVersion = serverVersion;                    
+            }
+            _lastServerVersionCheck++;
```
ayende: This isn't thread safe.

ayende (changes requested):

src/Raven.Client/Http/RequestExecutor.cs
redknightlois: That's 32K, yeah probably better to use a constant there... I forgot that we have the Constants class to build such things.

src/Raven.Client/Http/RequestExecutor.cs
redknightlois: I know and we don't care.

src/Raven.Client/Http/RequestExecutor.cs
ppekrol: Or you can write 32_000 instead here, because you do not need it elsewhere.

ppekrol: Also where are you resetting _lastServerVersionCheck? And 32k is too much. Consider using 128.

ayende: We want to reduce this as much as possible. I would prefer to do a time based value.
Something like every 5 - 15 minutes or so.

In fact, I'm also fine with doing this _once_ for all a node. The chance that they'll revert back a node while the app is running is small.

ppekrol: Time-based sounds like a good approach IMO.

redknightlois: Time-based requires accessing the clock which is not that good. We can time it under stress but the idea is to diminish this enough so that it does not register (3 orders of magnitude is huge) --- so I changed it to 1K. The point there is that this only register under stress anyways, so diminishing it 3 orders of magnitude (at the current optimization level) is negligible for the client.

ayende: @ppekrol Why bother to refresh anyway?

ppekrol: Best option would be to set `LastServerVersion` to null when we are switching nodes and then extract the server version when it is null. This would consume minimal amount of compute power. Thoughts?

ayende: What do you mean, switching nodes?

ppekrol: 1. When you are talking with node A and LastServerVersion is null, set it. 
2. If you will switch for any reason to node B, set LastServerVersion to null and go to point 1.

ayende: Won't work when we have failover, or when we use round robin, though.

ayende: In general, going back is such a rare event, we can just do it once and be done with it.
No extra checks needed.

redknightlois: The big question is: Should we want a reliability problem (even if rare) when the cost for it will become effectively 3 orders of magnitude less aka virtually 0?

ppekrol: It's not `3 orders of magnitude` issue - that is micro view on the problem. From macro overview this is 1.82% (probably even less) with the reliability issue. I do not see us stepping back from path once taken. Doing that every `mod 10` call or on failover decreases the problem to neglectable level.

redknightlois: I agree but right now the level of optimization of the client is from what I observed something that can be improved a lot and that will eventually will arise again, and we will need again to decide what to do. That 1.86% is a lie caused by the blocking call inside the `SendAsync` synchronization, it is roughly around 5% (still small in the grand scheme of things today) where 1 order of magnitude push it to 0.5% and 2 to 0.05% (1 in 10 vs 1 in 100). 

I chose 3 to avoid having to revisit this :D

ppekrol: For some apps 100 is a lot IMO.

ayende: @redknightlois  @ppekrol I'm missing something. This isn't a reliability issue.
This is used for one purpose, to check if the server is 4.2, which is pretty static.
We don't expect to replace the node versions (and backward) very often, so why do the additional check?

ppekrol: I'm worried that server version will change and Client API will not pick it up because if we will be doing that one time then it will require the restart of the client application. Plus there is always an option to have mixed clusters during update.

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -1476,6 +1486,9 @@ public static async Task<Stream> ReadAsStreamUncompressedAsync(HttpResponseMessa
                 return false;
             }
 
+            // As the server is down, we discard the server version to ensure we update when it goes up. 
+            chosenNode.DiscardServerVersion();
```
redknightlois, line 1490: I am not entirely sure this is the proper place to discard the Server Version on failure.

ayende: Looks like it should do the trick, yes.

src/Raven.Client/Http/ServerNode.cs
```diff
@@ -36,10 +39,35 @@ public override int GetHashCode()
...
             unchecked
             {
-                var hashCode = Url?.GetHashCode() ?? 0;
-                hashCode = (hashCode * 397) ^ (Database?.GetHashCode() ?? 0);
-                return hashCode;
+                return HashCombiner.CombineInline(Url?.GetHashCode() ?? _emptyStringHash, Database?.GetHashCode() ?? _emptyStringHash);
             }
         }
+
+        private int _lastServerVersionCheck = 0;
```
ayende, line 46: Can kill this, I think. The failover discard should be enough

redknightlois: Yes, but we negotiated a tradeoff with @ppekrol about how to deal with that... for now we will keep it and if eventually, we find we don't need the 'rechecks' then we will remove that part of the checks.

ppekrol:
@redknightlois can you move to 5.1?

PR #11749, opened 2021-02-08
RavenDB-16147 ability to change the default analyzers

ppekrol:


src/Raven.Client/Constants.cs
```diff
@@ -301,6 +301,19 @@ private Spatial()
...
                     public const double MilesToKm = 1.60934;
                 }
+
+                internal class Analyzers
+                {
+                    private Analyzers()
+                    {
+                    }
+
+                    public const string Default = "LowerCaseKeywordAnalyzer";
```
ppekrol, line 311: Use nameof

src/Raven.Server/Config/Categories/IndexingConfiguration.cs
```diff
@@ -259,6 +263,30 @@ public virtual bool RunInMemory
         [ConfigurationEntry("Indexing.NuGetPackageSourceUrl", ConfigurationEntryScope.ServerWideOnly)]
         public string NuGetPackageSourceUrl { get; set; }
 
+        [Description("TODO")]
```
ppekrol: Write description

PR #11751, opened 2021-02-08
RavenDB-13107 - Basic support for sharded queries

ayende:
This is still very raw and hacky, but it shows the overall structure that I think we'll have.

### DIFF
--- a/src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
@@ -148,7 +148,7 @@ namespace Raven.Server.Documents.Handlers
                     writer.WriteComma();
                     writer.WritePropertyName("Includes");
                     var includes = new List<Document>();
-                    includeCmd.Fill(includes);
+                    includeCmd.Fill(includes, false);
                     writer.WriteIncludes(context, includes);
                     writer.WriteEndObject();
                 }
--- a/src/Raven.Server/Documents/Queries/GraphQueryRunner.cs
+++ b/src/Raven.Server/Documents/Queries/GraphQueryRunner.cs
@@ -182,7 +182,7 @@ namespace Raven.Server.Documents.Queries
                         }
                     }
 
-                    idc.Fill(final.Includes);
+                    idc.Fill(final.Includes, false);
 
                     final.TotalResults = final.Results.Count;
 
--- a/src/Raven.Server/Documents/Queries/IndexQueryServerSide.cs
+++ b/src/Raven.Server/Documents/Queries/IndexQueryServerSide.cs
@@ -29,6 +29,9 @@ namespace Raven.Server.Documents.Queries
     public class IndexQueryServerSide : IndexQuery<BlittableJsonReaderObject>
     {
         [JsonDeserializationIgnore]
+        public bool ReturnMissingIncludeAsNull;
+
+        [JsonDeserializationIgnore]
         public int? Offset;
 
         [JsonDeserializationIgnore]
--- a/src/Raven.Server/Documents/Queries/IndexQueryServerSide.cs
+++ b/src/Raven.Server/Documents/Queries/IndexQueryServerSide.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Text;
 using Microsoft.AspNetCore.Http;
 using Raven.Client;
@@ -13,8 +15,14 @@ using Raven.Server.Documents.Queries.Timings;
 using Raven.Server.Extensions;
 using Raven.Server.Json;
 using Raven.Server.NotificationCenter;
+using Raven.Server.ServerWide;
+using Raven.Server.ServerWide.Context;
 using Raven.Server.TrafficWatch;
+using Sparrow;
 using Sparrow.Json;
+using Sparrow.Json.Parsing;
+using Sparrow.Server;
+using Voron;
 
 namespace Raven.Server.Documents.Queries
 {
@@ -84,6 +92,29 @@ namespace Raven.Server.Documents.Queries
 
         public bool IsStream;
 
+        private BlittableJsonReaderObject _asJson;
+
+        public BlittableJsonReaderObject ToJson(JsonOperationContext context)
+        {
+            if (_asJson != null) 
+                return _asJson;
+
+            _asJson = context.ReadObject(new DynamicJsonValue
+            {
+                [nameof(Query)] = Query,
+                [nameof(Start)] = Start,
+                [nameof(PageSize)] = PageSize,
+                [nameof(QueryParameters)] = QueryParameters,
+                [nameof(WaitForNonStaleResults)] = WaitForNonStaleResults,
+                [nameof(WaitForNonStaleResultsTimeout)] = WaitForNonStaleResultsTimeout,
+                [nameof(SkipDuplicateChecking)] = SkipDuplicateChecking,
+                [nameof(ProjectionBehavior)] = ProjectionBehavior,
+                
+            }, "query");
+            
+            return _asJson;
+        }
+
         public IndexQueryServerSide(string query, BlittableJsonReaderObject queryParameters = null)
         {
             Query = Uri.UnescapeDataString(query);
@@ -102,6 +133,7 @@ namespace Raven.Server.Documents.Queries
             try
             {
                 result = JsonDeserializationServer.IndexQuery(json);
+                result._asJson = json;
 
                 if (result.PageSize == 0 && json.TryGet(nameof(PageSize), out int _) == false)
                     result.PageSize = int.MaxValue;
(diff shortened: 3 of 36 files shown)


src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
```diff
@@ -148,7 +148,7 @@ public async Task Try()
                     writer.WriteComma();
                     writer.WritePropertyName("Includes");
                     var includes = new List<Document>();
-                    includeCmd.Fill(includes);
+                    includeCmd.Fill(includes, false);
```
karmeli87, line 151: issue to handle that?

src/Raven.Server/Documents/Queries/GraphQueryRunner.cs
```diff
@@ -182,7 +182,7 @@ public async Task WriteDetailedQueryResult(IndexQueryServerSide indexQuery, Quer
                         }
                     }
 
-                    idc.Fill(final.Includes);
+                    idc.Fill(final.Includes, false);
```
karmeli87, line 185: Also an issue?

src/Raven.Server/Documents/Queries/IndexQueryServerSide.cs
```diff
@@ -84,6 +95,29 @@ public IndexQueryServerSide(QueryMetadata metadata)
 
         public bool IsStream;
 
+        private BlittableJsonReaderObject _asJson;
+
+        public BlittableJsonReaderObject ToJson(JsonOperationContext context)
+        {
+            if (_asJson != null) 
+                return _asJson;
```
karmeli87, line 103: Can it move between contexts? Just wonder if `Clone` here is needed.

src/Raven.Server/Documents/ShardedHandlers/ShardedQueriesHandler.cs
```diff
@@ -0,0 +1,622 @@
...
+            {
+                var tasks = new List<Task>();
+
+                foreach (var (shardIndex, cmd) in _commands)
+                {
+                    var task = _parent.ShardedContext.RequestExecutors[shardIndex].ExecuteAsync(cmd, cmd.Context);
+                    tasks.Add(task);
+                }
+
+                await Task.WhenAll(tasks);
```
karmeli87, line 319: Also here we need some handling for partial failures?

src/Raven.Server/Documents/ShardedHandlers/ShardedQueriesHandler.cs
```diff
@@ -0,0 +1,622 @@
...
+                                return -1;
+                            return AlphaNumericFieldComparator.StringAlphanumComparer.Instance.Compare(xVal, yVal);
+                        }
+                    case Queries.AST.OrderByFieldType.Random:
+                        return (_random ??= new Random()).Next(0, int.MaxValue);
+                 
+                    case Queries.AST.OrderByFieldType.Custom:
+                        throw new NotSupportedException("Custom sorting is not supported in sharding as of yet");
+                    case Queries.AST.OrderByFieldType.Score:
+                    case Queries.AST.OrderByFieldType.Distance:
```
karmeli87, line 492: TODO?

src/Raven.Server/Documents/ShardedHandlers/ShardedQueriesHandler.cs
```diff
@@ -0,0 +1,622 @@
...
+                            //   support including from another shard, so we'll need to do that again for missing includes
+                            //   That means also recording the include() call from JS on missing values that we'll need to rerun on
+                            //   other shards
+                            var includeTask = queryProcessor.HandleIncludes();
+                            if (includeTask.IsCompleted == false)
+                            {
+                                await includeTask.AsTask();
+                            }
+                            
+                            queryProcessor.MergeResults();
```
karmeli87, line 105: We can have partial errors in `ExecuteShardedOperations` **and** / **or** in `HandleIncludes`, not sure how to handle all the permutations here.
Also need to consider those scenarios with respect to the reduce result.

src/Raven.Server/Documents/ShardedHandlers/ShardedQueriesHandler.cs
```diff
@@ -0,0 +1,622 @@
...
+                        return;
+                    }
+                }
+
+                
+                // * For paging queries, we modify the limits on the query to include all the results from all
+                //   shards if there is an offset. But if there isn't an offset, we can just get the limit from
+                //   each node and then merge them
+                if (_query.Offset > 0)
+                {
```
karmeli87, line 240: I think, it would be nice to see here maybe some performance hint, to suggest to avoid this and use `startWith` instead?

src/Raven.Server/Documents/ShardedHandlers/ShardedQueriesHandler.cs
```diff
@@ -0,0 +1,622 @@
...
+                    }
+                    else
+                    {
+                        queryTemplate.Modifications[nameof(IndexQuery.QueryParameters)] = modifiedArgs = new DynamicJsonValue();
+                    }
+
+                    int limit = ((_query.Limit  ?? 0) + (_query.Offset ?? 0))  * _parent.ShardedContext.NumberOfShardNodes;
+                    if (limit < 0) // overflow
+                        limit = int.MaxValue;
+                    modifiedArgs["__raven_limit"] = limit;
```
karmeli87, line 262: not sure where `__raven_limit` is wired and how it translated to RQL `limit`..

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -2134,9 +2134,20 @@ private void UpdateDatabase(ClusterOperationContext context, string type, Blitta
 
                     if (databaseRecordJson == null)
                     {
-                        if (updateCommand.ErrorOnDatabaseDoesNotExists)
-                            throw DatabaseDoesNotExistException.CreateWithMessage(databaseName, $"Could not execute update command of type '{type}'.");
-                        return;
+                        var shardIndex = valueNameLowered.AsSpan().LastIndexOf((byte)'$');
+                        if (shardIndex != -1) // try loading the sharded version, instead
```
karmeli87, line 2138: Why? I don't think it is a valid scenario and we should handle that higher in the stack?

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -2377,6 +2389,8 @@ private void SqueezeDatabasesToSingleNodeCluster(ClusterOperationContext context
                 using (var rawRecord = ReadRawDatabaseRecord(context, name))
                 {
                     var topology = rawRecord.Topology;
+                    if (topology == null)
```
karmeli87, line 2392: Is that a TODO?

src/Raven.Studio/wwwroot/version.txt
```diff
@@ -1 +1 @@
-﻿{ "Version": "5.1.0-custom-51" }
+﻿{ "Version": "5.1.5-custom-51" }
```
karmeli87, line 1: why the change?

karmeli87: and we need to change to 6.0 :-)

src/Sparrow/Json/JsonDeserializationBase.cs
```diff
@@ -375,6 +375,9 @@ private static ParameterExpression GetParameter(Type type, Dictionary<Type, Para
         private static Dictionary<string, T> ToDictionaryOfPrimitive<T>(BlittableJsonReaderObject json, string name, JsonDeserializationStringDictionaryAttribute jsonDeserializationDictionaryAttribute)
             where T : struct
         {
+            if (json == null)
```
karmeli87, line 378: why we added all those `null` checks?

src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
ayende: https://issues.hibernatingrhinos.com/issue/RavenDB-16279

src/Raven.Server/Documents/Queries/GraphQueryRunner.cs
ayende: Not planning to support graph queries on sharded dbs at this point at all

src/Raven.Server/Documents/Queries/IndexQueryServerSide.cs
ayende: It shouldn't, a query has a single context for its duration, unless I missed something.

src/Raven.Server/ServerWide/ClusterStateMachine.cs
ayende: Because in this way, pretty much all our existing infrastructure behaves in the same manner. We just create "virtual" sharded dbs, but we have to specialize a lot less to support it.

src/Raven.Server/ServerWide/ClusterStateMachine.cs
ayende: Not quite, this can happen when you are reading sharded database, since the root doesn't have a topology.

ayende: https://issues.hibernatingrhinos.com/issue/RavenDB-16280

src/Raven.Studio/wwwroot/version.txt
ayende: Accidental

src/Sparrow/Json/JsonDeserializationBase.cs
ayende: Because you might actually get nulls here, and we do when working with null topology, for example.
This is deserialization code and should handle this scenario.

PR #11755, opened 2021-02-10
RavenDB-16229 - Memory leak when patching a large collection

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-16229

### DIFF
--- a/src/Raven.Server/Documents/CollectionRunner.cs
+++ b/src/Raven.Server/Documents/CollectionRunner.cs
@@ -116,6 +116,8 @@ namespace Raven.Server.Documents
                                     break;
                                 }
 
+                                context.Transaction.ForgetAbout(document);
+
                                 ids.Enqueue(document.Id);
                             }
                         }


src/Raven.Server/Documents/CollectionRunner.cs
```diff
@@ -116,6 +116,8 @@ public virtual Task<IOperationResult> ExecuteDelete(string collectionName, long
                                     break;
                                 }
 
+                                context.Transaction.ForgetAbout(document);
```
karmeli87, line 119: Should we use the `PulsedTransaction` here?

ayende: That is relevant for compression only, on

ppekrol: Now I'm wondering. Shouldn't this be after we materialize the Id in ids.Enqueue? We are releasing the buffer before we materialize the Id? Maybe it is not a problem?

ayende: Yes, need to reverse this, to be safe

ppekrol: Pushed: https://github.com/ravendb/ravendb/commit/14dc3a84c8d7256374ff3383723e45d367052779

PR #11758, opened 2021-02-10
RavenDB-16047 /databases endpoint - return database topology on an error as well

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-16047

### DIFF
--- a/src/Raven.Server/Web/System/DatabasesHandler.cs
+++ b/src/Raven.Server/Web/System/DatabasesHandler.cs
@@ -297,50 +297,25 @@ namespace Raven.Server.Web.System
         private void WriteDatabaseInfo(string databaseName, BlittableJsonReaderObject dbRecordBlittable,
             TransactionOperationContext context, AbstractBlittableJsonTextWriter writer)
         {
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Web/System/DatabasesHandler.cs
```diff
@@ -297,50 +297,25 @@ private Task DbInfo(string dbName)
...
-                
-                var nodesTopology = new NodesTopology();
 
                 var statuses = ServerStore.GetNodesStatuses();
                 if (topology != null)
                 {
                     nodesTopology.PriorityOrder = topology.PriorityOrder;
-                    
+
+                    var clusterTopology = ServerStore.GetClusterTopology(context);
```
ppekrol, line 316: This is only gathered if `topology` is not null? Previously there was no relation between `topology` and `clusterTopology` I think?

haludi: We only use it to `GetUrlFromTag` and we do that only if there is topology

PR #11762, opened 2021-02-11
RDBS-124 avoid recreating index write upon every commit

karmeli87:


src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexPersistence.cs
```diff
@@ -198,6 +198,10 @@ public void Clean(CleanupMode mode)
             _converter?.Clean();
             _indexSearcherHolder.Cleanup(_index._indexStorage.Environment().PossibleOldestReadTransaction(null), mode);
 
+            var indexWriter = _indexWriter;
+            if (indexWriter != null)
+                indexWriter.RecreateAfterCommit = true;
```
ayende: Why it this being set here only? When is this running? Shouldn't this be on deep mode, then?

karmeli87: In general:
```
low mem -> regular clean up
timeToWaitForMemoryCleanup -> deep if no query for 10 min otherwise regular;
idleness -> can be deep if no work was for 5 min otherwise regular
```

Since this is a index _writer_ we clean it every time we reduce the mem. (right now we were clean it upon each commit)

ayende: If no query happened in the last ten minutes, we lose the writer?
Same for idle for 5 minutes? 

That means that a write every 15 minutes would be very expensive, no?

Why are we cleaning this up if there is no pressure?

karmeli87: We clean the writer if 
1. under low mem
2. idle run every 10 min (which is wired by its own, it seem we always clean every 10 min regardless)
3. no new docs for 5 seconds.

I think we don't need to over allocate here and clean up in reasonable time-frame and not wait for low mem to happened.
(mem pressure isn't the only reason for cleaning up, high number of managed objects is something we don't want either)

src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexWriter.cs
```diff
@@ -80,7 +82,12 @@ public void Commit(IState state, IndexingStatsScope commitStats)
             finally
             {
                 _indexWriter.SetCommitStats(null);
-                RecreateIndexWriter(state);
+
+                if (RecreateAfterCommit)
+                {
+                    RecreateAfterCommit = false;
```
arekpalinski: In addition to `LuceneIndexWriter` we also have `LuceneSuggestionIndexWriter` where also `RecreateIndexWriter()` is called
 
 https://github.com/ravendb/ravendb/blob/a868f1422b7edc40064728e4d583585209cf899f/src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneSuggestionIndexWriter.cs#L204
 
 So I believe it needs to follow the same logic as here

src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexPersistence.cs
ayende: The problem is that we will have high CPU after idle time, no? 
And 5 sec with no docs happens a _lot_.
Imagine a document coming in every 6 seconds. Then you are back to high CPU .

karmeli87: I agree it might be a problem, the question is, when do we want to cleanup the writer.
should we return the previous behavior of ~250K writes?

ayende: I would prefer to do that on low memory only.
250K writes isn't ideal, I think. We throw a lot of memory away then, which we then reuse?
Maybe after we run a merge, instead, since we'll need to reopen the new segment anyway?

karmeli87: If I understand correctly, upon merge we always drop, so we shouldn't worry about that.

My concern is that we are not playing nice with the memory and will be forced to hit low-mem some day, instead of doing that per index once in a while and avoid the high costs of the low-mem event (which has a global impact)

ayende: The "once in a while" thing worries me.
I guess we could set it to be something like: "after 10 minutes of no writes (needs to be configurable)", dispose the writer.

src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexPersistence.cs
```diff
@@ -166,7 +170,9 @@ IndexSearcher CreateIndexSearcher(IState state)
                             {
                                 var newReader = reader.Reopen(state);
                                 if (newReader != reader)
-                                    reader.DecRef(state);
+                                {
+                                    reader.Dispose();
```
ayende: How do you know that the old reader isn't still in use?

src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -2389,17 +2389,17 @@ private static bool DatabaseNeedsToRunIdleOperations(DocumentDatabase database,
 
             if ((now - maxLastWork).TotalMinutes > 5)
```
ayende, line 2390: maxLastWork here is the last write on any env, right?

karmeli87: yes

src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexPersistence.cs
karmeli87: I was looking at the dispose: 
https://github.com/ravendb/lucenenet/blob/3506f5028832315a4d6f8aeababcdd96b87bbac1/src/Lucene.Net/Index/IndexReader.cs#L1145

Which seems to be a "single-shot", so if we dispose twice and the ref count drop to 1 only on the second time, we will not do the actual dispose.

which make me think that this is wrong ?
https://github.com/ravendb/ravendb/blob/3e8d206e3dc7a9e611a1f253c68df6d1883ea613/src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexPersistence.cs#L205

ayende: What happens when you have a _query_ that is running while this is also running.
The `readerLock` you have isn't covered by those, and you may dispose the reader while queries are also running, no?
That is why we have the `DecRef` there.

src/Raven.Server/ServerWide/ServerStore.cs
ayende: https://issues.hibernatingrhinos.com/issue/RavenDB-16252

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1431,7 +1431,10 @@ protected void ExecuteIndexing()
                                 // so this is a good time to release resources we won't need
                                 // anytime soon
 
-                                var mode = NoQueryInLast10Minutes() ? CleanupMode.Deep : CleanupMode.Regular;
+                                var mode = IndexCleanup.Basic;
+                                if (NoQueryInLast10Minutes())
```
ayende, line 1435: I think we should make this configurable, no?

src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexPersistence.cs
```diff
@@ -65,6 +66,9 @@ public class LuceneIndexPersistence : IDisposable
 
         internal LuceneVoronDirectory LuceneDirectory => _directory;
 
+        private readonly object _readersLock = new object();
```
ayende, line 69: What is this used for? You are already using `DecRef` for concurrency, no?

src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexPersistence.cs
```diff
@@ -65,6 +66,9 @@ public class LuceneIndexPersistence : IDisposable
...
+        private readonly object _writersLock = new object();
```
ayende: Why is this required? Writers are only created on the indexer thread, no?

src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexPersistence.cs
```diff
@@ -518,18 +533,21 @@ public bool ContainsField(string field)
 
         public void DisposeWriters()
```
ayende: Is this called concurrently from outside the indexer thread?

src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexPersistence.cs
karmeli87: This used to sync the writes to `_lastReader`.
Since now we have 2 code paths that might modify it.

src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexPersistence.cs
karmeli87: yes, upon disable or idle clean up.

src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexPersistence.cs
arekpalinski: Answer is below - https://github.com/ravendb/ravendb/pull/11762/files#r579855989

src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexPersistence.cs
arekpalinski: Is it possible that we'll execute index cleanup and trigger indexing work at the same time? I wonder if we can dispose the writer that might be in use. Any chance for that?

ayende (approved):

src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexPersistence.cs
karmeli87: I have change the code, so now only one thread should modify the index writer.

arekpalinski:
Changes looks good to me. I have scheduled additional 3 more build on Jenkins:
https://jenkins.hibernatingrhinos.com/job/PR_Tests_v5.1/552/
https://jenkins.hibernatingrhinos.com/job/PR_Tests_v5.1/553/
https://jenkins.hibernatingrhinos.com/job/PR_Tests_v5.1/554/

@karmeli87 Please check the conflict

karmeli87:
@arekpalinski It's the conflict on the Lucene version which I will replace anyway after https://github.com/ravendb/lucenenet/pull/29

arekpalinski:
Oh, okay

ppekrol:
@karmeli87 @arekpalinski are we ready to merge this?

arekpalinski:
In general yes but Karmel needs to bump to valid version of Lucene.Net

karmeli87:
Closing in favor of 4.2 (https://github.com/ravendb/ravendb/pull/11824)

PR #11771, opened 2021-02-15
RavenDB-16250 Avoid CanConvert allocations

redknightlois:


ayende (approved):
Do we have perf numbers on this change?

redknightlois:
Preliminary yes.
image

Running time (CPU) -> Before: 686,958 After 463464  (~30% improvement)
Running time (wall clock) -> Before: 14m40s - After: 12m01s (edited)

src/Raven.Client/Json/Serialization/NewtonsoftJson/Internal/Converters/JsonDictionaryDateTimeKeysConverter.cs
```diff
@@ -16,6 +17,8 @@ internal sealed class JsonDictionaryDateTimeKeysConverter : JsonConverter
 
         public static readonly JsonDictionaryDateTimeKeysConverter Instance = new JsonDictionaryDateTimeKeysConverter();
 
+        private readonly ConcurrentDictionary<Type, bool> _canConvertCache = new ConcurrentDictionary<Type, bool>();
```
ppekrol: Isn't it better to use regular dictionary the way we use it here https://github.com/ravendb/ravendb/blob/v5.1/src/Raven.Client/Json/Serialization/NewtonsoftJson/Internal/Converters/CachingJsonConverter.cs ?

ayende: Yes, probably better, the number of entries is small, so will give better perf.

redknightlois: No, you can't, it will die on concurrent accesses. I tried locking and it is good, but so far not better than the concurrent dictionary.

ppekrol: Concurrent reads are not a problem IIRC, check how this is done in the link that I have send you.

ayende: There are no concurrent writes or writes & reads.
We create a new instance each time, because the number of times involved is very small.

redknightlois: It is a switch-on-change, that would work too. My question here is: Do we want to complicate it when the performance impact is negligible?

ayende: For the common case, for reads only, it is cheaper than concurrent dictionary, yes.

redknightlois:
@ppekrol Do we want to apply this to 5.1 branch?

ayende (approved):

ppekrol:
@redknightlois yes, move it to 5.1 (I will do a 5.1 to 5.2 merge)

ppekrol (approved):

ppekrol:
@redknightlois can you move it to 5.1?

PR #11778, opened 2021-02-17
RavenDB-15262 Use DisableCertificateDownloads when building the chain

ppekrol:


src/Raven.Server/RavenServer.cs
```diff
@@ -1653,7 +1654,10 @@ private bool CertHasKnownIssuer(X509Certificate2 userCertificate, X509Certificat
             X509Certificate2 issuerCertificate = null;
 
             var userChain = new X509Chain();
+            userChain.ChainPolicy.DisableCertificateDownloads = true;
+
             var knownCertChain = new X509Chain();
+            knownCertChain.ChainPolicy.DisableCertificateDownloads = true;
```
ayende: This is the one place that I'm not certain what we should do.
Here we _need_ to insect the whole chain. This is also likely to be relatively rare.
But we need to add a comment explaining why not.

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -4062,40 +4062,8 @@ public bool IsReplicationCertificate(TransactionOperationContext context, string
...
-                for (var i = 0; i < knownCertChain.ChainElements.Count; i++)
-                {
-                    // We walk the chain and compare the user certificate vs one of the existing certificate with same pinning hash
-                    var currentElementPinningHash = userChain.ChainElements[i].Certificate.GetPublicKeyPinningHash();
-                    if (currentElementPinningHash != knownCertChain.ChainElements[i].Certificate.GetPublicKeyPinningHash())
-                    {
-                        return false;
-                    }
-                }
+                if (CertificateUtils.CertHasKnownIssuer(userCert, knownCert, securityConfiguration, out var _) == false)
```
ppekrol, line 4065: Couple of things that I'm not sure about:
1. We are iterating here through a list of certificates, but old code when there was an exception or in few other places was returning immediately `false`. Shouldn't we go over the whole list anyway?
2. I've moved `CertHasKnownIssuer` to CertificateUtils, but it know also checks WellKnownIssuerHashes and we were not doing it here. Is this correct?

CC @ayende @iftahbe

iftahbe: 1. For the pinning hash to be accepted, we compare the entire chain, and if even one element is different we return false and don't trust it. No point in checking the next elements.

iftahbe: 2. I think you are correct and it should have also checked the WellKnownIssuerHashes. It probably was a mistake.

iftahbe (approved):

src/Raven.Server/Utils/CertificateUtils.cs
```diff
@@ -13,17 +13,85 @@
...
+
+        private static readonly Logger Logger = LoggingSource.Instance.GetLogger("Server", typeof(CertificateUtils).FullName);
+
+        internal static bool CertHasKnownIssuer(X509Certificate2 userCertificate, X509Certificate2 knownCertificate, SecurityConfiguration securityConfiguration, out string issuerPinningHash)
+        {
+            issuerPinningHash = null;
+            X509Certificate2 issuerCertificate = null;
+
+            var userChain = new X509Chain();
+            var knownCertChain = new X509Chain();
```
ayende: I think we need a comment here explaining that we explicitly don't set DisableCertificateDownloads

PR #11793, opened 2021-02-20
RavenDB-16273: Improved codegen for the Blittable JSON Document Builder

redknightlois:


src/Raven.Client/Json/Serialization/NewtonsoftJson/Internal/BlittableJsonConverterBase.cs
```diff
@@ -240,11 +240,36 @@ private static bool ShouldSimplifyJsonBasedOnType(string typeValue)
             return typeof(IEnumerable).IsAssignableFrom(type.GetGenericTypeDefinition());
         }
 
+        internal class TypeCacheItem : IEquatable<TypeCacheItem>
```
ayende, line 243: Should this be a struct?

src/Raven.Client/Json/Serialization/NewtonsoftJson/Internal/BlittableJsonConverterBase.cs
```diff
@@ -240,11 +240,36 @@ private static bool ShouldSimplifyJsonBasedOnType(string typeValue)
...
+
         internal static Type GetPropertyType(string propertyName, Type rootType)
         {
             if (rootType == null)
                 return null;
 
+            Type type = null;
+            var item = new TypeCacheItem() { Property = propertyName, Type = rootType };
+            if (_propertyTypeCache.TryGetValue(item, out var result))
+                return type;
```
ayende: Why are you returning `null` here (the only option for `type` if you _got_ a value? Shouldn't this be `result`?
```suggestion
                return result;
```

src/Sparrow/Collections/FastStack.cs
```diff
@@ -208,10 +217,26 @@ public void Push(T item)
...
             return;
 
-            Grow:
+        Grow:
             PushUnlikely(item);
         }
 
+        // Pushes an item to the top of the stack.
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public ref T PushByRef()
```
ayende, line 226: What is the use case here, it looks like a pretty strange API, no?

src/Sparrow/Json/ManualBlittableJsonDocumentBuilder.cs
```diff
@@ -32,27 +33,24 @@ public sealed class ManualBlittableJsonDocumentBuilder<TWriter> : AbstractBlitta
 
         public void StartWriteObjectDocument()
         {
-            _continuationState.Push(new BuildingState
-            {
-                State = ContinuationState.ReadObjectDocument
-            });
+            ref var state = ref _continuationState.PushByRef();
+            state.State = ContinuationState.ReadObjectDocument;
```
ayende, line 37: Wouldn't it be easier to have the `Push` accept the value, instead of a `ref return`?
A `ref` param here to save copying would have the same impact, no?

src/Raven.Client/Json/Serialization/NewtonsoftJson/Internal/BlittableJsonConverterBase.cs
redknightlois: No. If we use structs the actual cost of dealing with them explodes.

src/Raven.Client/Json/Serialization/NewtonsoftJson/Internal/BlittableJsonConverterBase.cs
redknightlois: Good catch. I actually push that one for Jenkins to run the slow tests. Thats why I havent seen improvements on that one when I run the profiling test.

src/Sparrow/Collections/FastStack.cs
redknightlois: It is. But that strange API allows the JIT to generate a great assembler. The reason why that API is useful is that you use the knowledge those objects have been already created in bulk and you just write it's values. Big differences in speed and CPU use

src/Sparrow/Json/ManualBlittableJsonDocumentBuilder.cs
redknightlois: Nope. That will force emitting a constructor and a copy.

src/Sparrow/Json/LazyStringValue.cs
```diff
@@ -196,15 +196,19 @@ public bool Equals(string other)
             if (_string != null)
                 return string.Equals(_string, other, StringComparison.Ordinal);
 
-            var sizeInBytes = Encodings.Utf8.GetMaxByteCount(other.Length);
 
-            if (_lazyStringTempComparisonBuffer == null || _lazyStringTempComparisonBuffer.Length < other.Length)
-                _lazyStringTempComparisonBuffer = new byte[Bits.PowerOf2(sizeInBytes)];
+            if (_lazyStringTempComparisonBuffer == null || _lazyStringTempComparisonBuffer.Length < other.Length * 5)
+            {
+                var sizeInBytes = Encodings.Utf8.GetMaxByteCount(other.Length);
```
ayende, line 202: Why bother with this entirely? Can't we just specify that the size is *5 and be done with it entirely?

redknightlois: Yes, but how much do we grow it afterward. I prefer to have a tighter number, which that calls provide us that.

ppekrol:
@redknightlois can you rebase?

redknightlois:
@ppekrol No, I will create a new PR with only the safe changes.

src/Sparrow/Json/BlittableJsonTextWriter.cs
```diff
@@ -94,52 +94,62 @@ public abstract unsafe class AbstractBlittableJsonTextWriter : IDisposable
         private const byte Comma = (byte)',';
         private const byte Quote = (byte)'"';
         private const byte Colon = (byte)':';
-        public static readonly byte[] NaNBuffer = { (byte)'"', (byte)'N', (byte)'a', (byte)'N', (byte)'"' };
-        public static readonly byte[] PositiveInfinityBuffer =
+        public static ReadOnlySpan<byte> NaNBuffer => new byte[] { (byte)'"', (byte)'N', (byte)'a', (byte)'N', (byte)'"' };
+        public static ReadOnlySpan<byte> PositiveInfinityBuffer => new byte[]
```
ayende, line 98: This create _properties_,  which allocates a new value each time.
Pretty sure that isn't what you intended to do.

src/Sparrow/Json/BlittableJsonTextWriter.cs
```diff
@@ -94,52 +94,62 @@ public abstract unsafe class AbstractBlittableJsonTextWriter : IDisposable
...
         {
             (byte)'"', (byte)'I', (byte)'n', (byte)'f', (byte)'i', (byte)'n', (byte)'i', (byte)'t', (byte)'y', (byte)'"'
-        };
-        public static readonly byte[] NegativeInfinityBuffer =
+        };        
+
+        public static ReadOnlySpan<byte> NegativeInfinityBuffer => new byte[]
```
ayende, line 103: Property instead of static field here.

src/Sparrow/Json/BlittableJsonTextWriter.cs
```diff
@@ -94,52 +94,62 @@ public abstract unsafe class AbstractBlittableJsonTextWriter : IDisposable
...
         };
         public static readonly byte[] NullBuffer = { (byte)'n', (byte)'u', (byte)'l', (byte)'l', };
         public static readonly byte[] TrueBuffer = { (byte)'t', (byte)'r', (byte)'u', (byte)'e', };
         public static readonly byte[] FalseBuffer = { (byte)'f', (byte)'a', (byte)'l', (byte)'s', (byte)'e', };
 
-        private static readonly byte[] EscapeCharacters;
-        public static readonly byte[][] ControlCodeEscapes;
+        private static ReadOnlySpan<byte> EscapeCharacters => InitializeEscapeCharacters();
+
+        public static ReadOnlySpan<int> ControlCodeEscapes => InitializeControlCodeEscapes();
```
ayende, line 113: Properties, computed each time

src/Sparrow/Json/BlittableJsonTextWriter.cs
redknightlois: Nope, those are subjected to a special optimization at the JIT level. What happens there is that those values are baked in at the dll .text section. It is a neat trick :D
https://vcsjones.dev/csharp-readonly-span-bytes-static/

ayende: I remember reading that, but I though it was for span fields, not for properties

src/Raven.Client/Json/Serialization/NewtonsoftJson/Internal/BlittableJsonWriter.cs
```diff
@@ -25,6 +30,10 @@ internal class BlittableJsonWriter : JsonWriter, IJsonWriter
             _manualBlittableJsonDocumentBuilder.StartWriteObjectDocument();
             _documentInfo = documentInfo;
             _first = true;
+
+            _metadataKey = context.GetLazyStringForFieldWithCaching(Constants.Documents.Metadata.Key);
+            _metadataCollection = context.GetLazyStringForFieldWithCaching(Constants.Documents.Metadata.Collection);
+            _metadataId = context.GetLazyStringForFieldWithCaching(Constants.Documents.Metadata.Id);
```
arekpalinski, line 36: Shouldn't we dispose them on writer dispose?

ayende: No, note that this is used with caching, so it will live for the life time of the context.

PR #11794, opened 2021-02-21
RavenDB-16264 Don't user batch delete for deleting multiple blobs

garayx:


src/Raven.Server/Documents/PeriodicBackup/Azure/RavenAzureClient.cs
```diff
@@ -590,6 +590,10 @@ public void DeleteMultipleBlobs(List<string> blobs)
             if (blobs.Count == 0)
                 return;
 
+            DeleteBlobsWithSasToken(blobs);
+
+            //TODO: RavenDB-16264
+            /*
```
ayende, line 596: Why do we keep so much dead code?

ayende (changes requested):

src/Raven.Server/Documents/PeriodicBackup/Azure/RavenAzureClient.cs
ppekrol: This is not entirely dead code. We want to support multi-delete, but it requires further investigation so for now it is better to go back to a one by one call as I understood from @garayx and @grisha-kotler

ppekrol:
This should go to 4.2?

garayx:
> This should go to 4.2?

yes: https://github.com/ravendb/ravendb/pull/11800

PR #11802, opened 2021-02-23
RavenDB-15747 prepare deb packaging

gregolsky:
On how it's supposed to work: https://github.com/ravendb/ravendb/blob/f59849b7a11f0615c9aec3c8621e7c88390541fa/scripts/linux/pkg/linux-packaging.md

I also added a config key to specify where does the Setup save the server certificate after it's been done. By default it was the working dir, which won't do in this scenario.


scripts/linux/pkg/deb/assets/DEBIAN/postinst
```diff
@@ -0,0 +1,84 @@
...
+if command -v systemctl &> /dev/null; then
+    cp /usr/local/ravendb/ravendb.service /etc/systemd/system/ravendb.service 
+    chmod 444 /etc/systemd/system/ravendb.service
+
+    systemctl daemon-reload 
+    systemctl reenable ravendb
+    systemctl restart ravendb
+fi
+
+# CopyPalFile() workaround - 4.2 only
```
gregolsky: @ppekrol Do you recall any other files being copied in this manner in runtime? We need to handle this ahead of time on installation when we're `root`.

scripts/linux/pkg/deb/assets/DEBIAN/changelog
```diff
@@ -0,0 +1,4 @@
+ravendb ($RAVENDB_VERSION-1) stable; urgency=medium
+
+Please visit for full set of changes:
+https://ravendb.net/docs/article-page/5.1/csharp/start/whats-new
```
ayende: This will need updating each time

```suggestion
https://ravendb.net/docs/article-page/latest/csharp/start/whats-new
```

scripts/linux/pkg/deb/assets/DEBIAN/control
```diff
@@ -0,0 +1,13 @@
+Package: ravendb
+Version: $RAVENDB_VERSION
+Section: database
+Priority: optional
+Depends: $DEB_DEPS 
+Architecture: amd64
```
ayende: we need one for ARM as well (32 & 64, I think)

scripts/linux/pkg/deb/assets/DEBIAN/control
```diff
@@ -0,0 +1,13 @@
...
+Maintainer: $DEBFULLNAME <$DEBEMAIL>
+Homepage: https://ravendb.net
+Vcs-Browser: https://github.com/ravendb/ravendb
+Vcs-Git: https://github.com/ravendb/ravendb.git
+Description: RavenDB NoSQL Database
+ A NoSQL Database that's fully transactional. 
+ Allows 1 million reads and 150000 writes per second.
```
ayende: Do we need this?

scripts/linux/pkg/deb/assets/DEBIAN/copyright
```diff
@@ -0,0 +1,29 @@
...
+
+Files: *
+Copyright: 2021 Hibernating Rhinos LTD <support@ravendb.net>
+License: https://raw.githubusercontent.com/ravendb/ravendb/$RAVENDB_VERSION/LICENSE.txt
+
+# If you want to use GPL v2 or later for the /debian/* files use
+# the following clauses, or change it to suit. Delete these two lines
+Files: debian/*
+Copyright: 2021 Hibernating Rhinos LTD <support@ravendb.net>
+License: GPL-2+
```
ayende: Shouldn't this be?

```suggestion
License: AGPL-3+
```

scripts/linux/pkg/deb/assets/DEBIAN/changelog
ayende: Or use:

```suggestion
https://ravendb.net/docs/article-page/$RAVENDB_VERSION/csharp/start/whats-new
```

lukaszdobrzynski (approved):

scripts/linux/pkg/deb/assets/DEBIAN/postinst
```diff
@@ -0,0 +1,84 @@
...
+chmod 770 /var/log/ravendb/{logs,audit} /var/lib/ravendb/{data,security,nuget} 
+
+find /usr/local/ravendb -type d -exec chmod 555 {} + 
+find /usr/local/ravendb -type f -exec chmod 444 {} + 
+
+chown root:ravendb /usr/local/ravendb/server/{Raven.Server,rvn}
+chmod 550 /usr/local/ravendb/server/{Raven.Server,rvn}
+
+if command -v setcap &> /dev/null; then
+    setcap CAP_NET_BIND_SERVICE=+eip /usr/local/ravendb/server/Raven.Server
```
ayende: Need to also setup the debug permissions, no?

scripts/linux/pkg/deb/assets/DEBIAN/postinst
ayende: I think we have zstd and sodium as well

scripts/linux/pkg/deb/assets/DEBIAN/postrm
```diff
@@ -0,0 +1,55 @@
...
+# the debian-policy package
+
+
+case "$1" in
+    remove|upgrade|failed-upgrade|abort-install|abort-upgrade|disappear)
+    ;;
+
+    purge)
+    
+    rm -v /etc/ravendb/settings.json
```
ayende: Remove the certificate as well?

scripts/linux/pkg/deb/assets/DEBIAN/prerm
```diff
@@ -0,0 +1,51 @@
...
+esac
+
+
+if command -v systemctl &> /dev/null && systemctl cat -- ravendb.service &> /dev/null; then
+    systemctl disable ravendb
+    systemctl stop ravendb
+fi
+
+# 4.2 only
+rm -vf /usr/local/ravendb/server/librvnpal.so
```
ayende: zstd & sodium as well
But aren't we removing the directory entirely?

scripts/linux/pkg/deb/assets/build-deb.sh
```diff
@@ -0,0 +1,76 @@
...
+
+python3 - <<SCRIPT || exit 1
+import json
+import codecs
+
+defaultSettingsPath = '$BUILD_DIR/RavenDB/Server/settings.default.json'
+with codecs.open(defaultSettingsPath, 'r', 'utf-8-sig') as f:
+    settings = json.load(f);
+    
+del settings['DataDir']
```
ayende: I think we should default to `/data/ravendb` and create that in the install phase

scripts/linux/pkg/deb/assets/wrapper.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash -x
+
+# This is a wrapper allowing to run RavenDB by regular users *wherever* (directory-wise) they like
```
ayende: You are making this exposed as `ravendb` command, which is likely not ideal.
maybe `ravendb-local` ?

scripts/linux/pkg/linux-packaging.md
```diff
@@ -0,0 +1,88 @@
...
+- install systemd service
+
+- add soft links to binaries
+
+- add `ravendb` user and group
+
+- set file permissions properly
+
+- setcap on the Raven.Server
+
```
ayende, line 16: `setcap`  to allow debugging

scripts/linux/pkg/linux-packaging.md
```diff
@@ -0,0 +1,88 @@
...
+### Ran under systemd
+The default is when RavenDB is used as a systemd daemon and controlled with systemd.
+The following directory layout (initialized by the package installation process) and permissions is used in this scenario:
+
+#### Filesystem locations and permissions
+
+- settings - `/etc/ravendb/settings.json`
+
+- data - `/var/lib/ravendb/data`
```
ayende: Default path in `/data/ravendb` ? Should we go for the layout consistency or assume that this is the main purpose of the instance and make the data location obvious?

scripts/linux/pkg/linux-packaging.md
```diff
@@ -0,0 +1,88 @@
...
+
+- security (e.g. certificate)  - `/var/lib/ravendb/security`
```
ayende: Who can read into this? Also, do we write the secret key into this location as well or to `$HOME/.secrets` ?

scripts/linux/pkg/linux-packaging.md
ayende: Thinking this over, MySQL & Postgres all store their data in similar paths, so `/var/lib` it is.

scripts/linux/pkg/linux-packaging.md
```diff
@@ -0,0 +1,88 @@
...
+
+- logs  - `/var/log/ravendb/logs`
+
+- audit logs  - `/var/log/ravendb/audit`
+
+- binaries (current Server dir) - `/usr/local/ravendb`
+
+- rvn link - `/usr/bin/rvn -> /usr/local/ravendb/rvn`
+
+- ravendb - `/usr/bin/ravendb -> /usr/local/ravendb/Raven.Server`
```
ayende: The ability to run this independently as a stand alone server isn't really attractive, what scenario are we trying to enable here?

scripts/linux/pkg/linux-packaging.md
```diff
@@ -0,0 +1,88 @@
...
+| `/etc/ravendb/settings.json` | root | ravendb | 440 |
+| `/var/lib/ravendb/data` | root | ravendb | 770 |
+| `/var/lib/ravendb/security` | root | ravendb | 770 |
+| `/var/log/ravendb/logs` | root | ravendb | 770 |
+| `/var/log/ravendb/audit` | root | ravendb | 770 |
+
+## Running as a regular user
+
+The idea is for the users to be able to run multiple instances of RavenDB on a multi-user system, keeping the data and the settings in their home directory.
+When run as a regular user we start on a random port without settings, producing RavenData dir (and all the other dirs) in the working dir.
```
ayende: I can't really think of a good scenario for this. If user have installed us, they are likely going to be wanting to use the dameon.

scripts/linux/pkg/linux-packaging.md
```diff
@@ -0,0 +1,88 @@
...
+
+Default settings should include proper retention policies for all logs.
+Turn on audit logs.
+
+## Package lifecycle
+
+### Installation
+
+When installed and run the first time we start in `SetupMode` `Initial`.
+
```
ayende, line 85: Need to consider this scenario. Assume that I'm using SSH to connect to the server.
How do I do the setup?

We can either make it work via command line browser (doubtful) or need to emit the right command to show SSH tunneling and connect to remotely to the local port.

ayende (changes requested):

scripts/linux/pkg/deb/assets/DEBIAN/changelog
gregolsky: This one actually depends on the version we're releasing. This is a PR to 4.2 so it should be 4.2 etc.

scripts/linux/pkg/deb/assets/DEBIAN/control
gregolsky: first need to figure out using arm docker images on x64 - I think it's called `buildx`. Anyway, yes - will log that.

scripts/linux/pkg/deb/assets/DEBIAN/control
gregolsky: Some package description would be nice. That's also what's shown in `apt` when you `show` the package. Would you like something else here?

scripts/linux/pkg/deb/assets/DEBIAN/changelog
gregolsky: yeah but we'd need to cut that to first 2 numbers. Will do

scripts/linux/pkg/deb/assets/DEBIAN/copyright
gregolsky: I'm not really an expert here when it comes to licensing.. This is basically about stuff in `debian` dir - the packaging, not RavenDB.

scripts/linux/pkg/linux-packaging.md
gregolsky: Yes

scripts/linux/pkg/linux-packaging.md
gregolsky: Only `root` and `ravendb` group users. Also what determines where the `.secrets` is kept? Also it it does keep it 'at home' then it's `ravendb` user's home.

scripts/linux/pkg/linux-packaging.md
gregolsky: Well, I thought about things like a university setting up a multi-user server for all students and then they have the ability to run RavenDB for their purposes. Scrap it?

scripts/linux/pkg/linux-packaging.md
gregolsky: CLI browser is out of question. They are not that powerful to run Studio or even the setup part - too complex. As far as tunneling is concerned it's probably their thing to do it - we can put something in the docs and in the package readme, but I would not invest there. 

Main question here is how do we set it up when having just the terminal. Ideas for the future:

- a CLI setup wizard script (drawing some menus and all), sending Setup Wizard payloads via curl, doing LE and puff - it's there. It's feasible for sure. 

- simpler and cheaper, no wizard, but elaborate `ravendb-setup` for which you'd pass few to multiple switches (I mean options elaboration level certbot) and they would init the settings file
```bash
$ sudo ravendb-setup \
-o raven-setup-package.zip \
--license ./license.json \
--accept-eula \
--agree-le-tos \
--node-ip 1.2.3.4 \
--node-ip 2.2.3.4 \
--node-ip 1.6.6.7 \
--domain ravendb.com
# probably few more here
```
or
```
ravendb-setup --localhost --unsecured --yes-i-accept-its-unsecured
```
People are getting used to this kind of commands nowadays - mentioned `certbot` or `docker` have this kind of interface and it works for them.

Without a tool like this we can:
- start in setup as we do by default - document accessing it the first time if set up on a "headless" machine
- start unsecured on localhost, which is not ideal

Thoughts?

scripts/linux/pkg/deb/assets/DEBIAN/postinst
gregolsky: yes

scripts/linux/pkg/deb/assets/DEBIAN/prerm
gregolsky: `apt` lists the files in the package and removes them one by one. Then if the dir is empty removes that dir. If something's copied manually it will prevent directory removal - so we need to remove manually the stuff we manually added.

scripts/linux/pkg/deb/assets/DEBIAN/postrm
gregolsky: yes

scripts/linux/pkg/deb/assets/build-deb.sh
gregolsky: We do that, create /var/lib/ravendb/data

scripts/linux/pkg/deb/assets/wrapper.sh
gregolsky: OK, or do we scrap it altogether? Please see my other comment on this one.

scripts/linux/pkg/deb/assets/DEBIAN/postrm
gregolsky: Actually since it's in `/var/lib/ravendb/security` it's removed a line below.

scripts/linux/pkg/deb/assets/DEBIAN/control
ayende: Why is docker invovled here?

scripts/linux/pkg/deb/assets/DEBIAN/control
ayende: Yes, send that to David :-)

scripts/linux/pkg/deb/assets/DEBIAN/copyright
ayende: I think that this is about our license, not theirs.

scripts/linux/pkg/deb/assets/wrapper.sh
ayende: yes

scripts/linux/pkg/linux-packaging.md
ayende: It's either on `$HOME/.ravendb/secret.key` or you can specify `Security.MasterKey.Path` - I think we want the later option

scripts/linux/pkg/linux-packaging.md
ayende: Yes, in almost all such cases, docker containers would be better

scripts/linux/pkg/linux-packaging.md
ayende: the `ravendb-setup` command might be the best option, yes. But also needs to be interactive, I think

That said, I think that setting up the system to allow SSH tunneling is easiest to do and give us all the benefits.
We just need to provide the command and ensure that the user is aware of that.

By default, we use dynamic port, which isn't ideal, though.

scripts/linux/pkg/deb/assets/DEBIAN/copyright
gregolsky: Our stuff license is above. It was just how it was in template. I can remove this `debian/*` clause and wrap everything in one - https://github.com/ravendb/ravendb/pull/11802/files/d53312532f519c94e7ec3fef786663baef2acd96#diff-82262bb7894e7f610c8c2626336d25575097bd90860600953b81b40ef1a2c2d8R7

scripts/linux/pkg/deb/assets/DEBIAN/control
gregolsky: We built package for 18.04 on 18.04 image and test its installation there, then proceed with 20.04. Actually, you're right, I mixed those 2, but I'll need to run arm on my local to test it - but that works out-of-the-box on Docker for Desktop. 

Anyway I'll add a variable for architecture.

ayende: Creating the deb can be done on WSL / etc, it is just a packaging format, not related to where / how we run it.

ayende: Note that we already do the cross compliation, etc.

gregolsky: Yes, but on CI we're gonna run that pipeline on Ubuntu machine on Docker.

scripts/linux/pkg/linux-packaging.md
gregolsky: So I presume we want that like the rest of security - `/var/lib/ravendb/security/secret.key`.

ayende: Yes

gregolsky: You can specify that if it's there. If a new one needs to be generated it'll die with "no such file" https://github.com/ravendb/ravendb/blob/42bd9f403e176081040712826fb8e7b217b1afa7/src/Raven.Server/ServerWide/SecretProtection.cs#L726

But.. if it has length 0 it'll generate it. I'll just `touch` it in the script then.

scripts/linux/pkg/linux-packaging.md
gregolsky: We don't have a good place to show such command when installing the package - server is not running yet and we don't know the port. Even if we did, we don't know how does the user connect to this server in the first place (the key/password, tunneling the ssh itself through some bastion) since we would need to mimic that. I don't think we're able to do that.

solutions:
- `ravendb-setup` mentioned above
- include `open-ravendb-setup-to-the-world.sh` in the package
- add explicit info in the docs on how to setup on headless/ssh-only machines - we already have some piece here sort of https://ravendb.net/docs/article-page/5.1/Csharp/start/installation/setup-examples/aws-linux-vm
- not use dynamic port for setup? (still can't show exact command but can print that we're gonna wait for you on `localhost:53780`

ayende: Yes, we will pretty much have to run a fixed port. We can use a high one, which is unlikely to be busy.

gregolsky: Added a message like this during the install:
image

ayende: Please add the `ssh` command to do the port forwarding as well here.

gregolsky: Since we don't know the external hostname the best I think we can do is something like this:
image

Does it work for you?

ayende: Can't we use: `dig +short myip.opendns.com @resolver1.opendns.com ` ?

gregolsky: OK, sure. Might cover most cases. Here's what I have now:
```
$ sudo dpkg -i ravendb_4.2.112-0_amd64.deb
Selecting previously unselected package ravendb.
(Reading database ... 70102 files and directories currently installed.)
Preparing to unpack ravendb_4.2.112-0_amd64.deb ...
Unpacking ravendb (4.2.112-0) ...
Setting up ravendb (4.2.112-0) ...
Adjust createdump binary permissions...
Adjust Raven.Debug binary permissions...
###
#
#  Please navigate to http://127.0.0.1:53700 in your web browser to complete setting up RavenDB.
#  If you set up the server through SSH, you can tunnel RavenDB setup port and proceed with the setup on your local.
#
#  For public address:   ssh -N -L localhost:8080:localhost:53700 ubuntu@54.198.78.116
#
#  For internal address: ssh -N -L localhost:8080:localhost:53700 ubuntu@ip-10-0-2-76
#
###
Processing triggers for libc-bin (2.27-3ubuntu1) ...
```

Sufficient? Still don't know if somebody logged in with password or key so we might be wrong with regards to auth method.

ayende: Looks really nice, yes.

ayende: It is enough that we provide this, the actual auth is on them.

We can't handle everything.

For example, we might not be able to handle it if they force to bind to localhost and use another port, etc. 
Not on us in this case.

gregolsky:
Applied Paweł's suggestions. I think it's ready to be merged - further work will come in 5.1 - there's `amd64` profile to add and probably some script tweaks to be made.

scripts/linux/pkg/deb/assets/build.sh
```diff
@@ -0,0 +1,81 @@
...
+echo "Package dependencies: $DEB_DEPS"
+
+case $RAVEN_PLATFORM in
+
+    linux-x64)
+        export DEB_ARCHITECTURE="amd64"
+        export RAVEN_SO_ARCH_SUFFIX="linux.x64"
+        ;;
+
+    raspberry-pi)
```
ayende, line 44: what about arm 64?

scripts/linux/pkg/deb/assets/ravendb/debian/ravendb.service
```diff
@@ -0,0 +1,26 @@
...
+StartLimitBurst=0
+Restart=on-failure
+Type=simple
+TimeoutStopSec=300
+Environment="RAVEN_DataDir=/var/lib/ravendb/data"
+Environment="RAVEN_Indexing_NugetPackagesPath=/var/lib/ravendb/nuget"
+Environment="RAVEN_Logs_Path=/var/log/ravendb/logs"
+Environment="RAVEN_Security_AuditLog_FolderPath=/var/log/ravendb/audit"
+Environment="RAVEN_Security_MasterKey_Path=/etc/ravendb/security/master.key"
+Environment="RAVEN_Setup_Certificate_Path=/etc/ravendb/security/server.pfx"
```
ayende, line 21: Who owns this directoyr?

gregolsky: root:ravendb

scripts/linux/pkg/deb/assets/build.sh
gregolsky: will come with 5.1 - unsupported in 4.2

scripts/linux/pkg/deb/assets/ravendb/debian/ravendb.service
gregolsky:
```
/etc/ravendb:                                                
total 8                                                      
drwxrwx--- 2 root ravendb 4096 Apr 21 13:00 security         
-rw-rw---- 1 root ravendb  239 Apr 21 13:00 settings.json    
                                                             
/etc/ravendb/security:                                       
total 0                                                      
-rw-rw---- 1 root ravendb 0 Apr 21 13:00 master.key          
```

PR #11811, opened 2021-02-26
RavenDB-15897 support for read-only certificates at database level

ppekrol:
- all endpoints were marked with either Read or Write flag, but only database endpoints with ValidUser are affected

### DIFF
--- a/src/Raven.Client/ServerWide/Operations/Certificates/CertificateDefinition.cs
+++ b/src/Raven.Client/ServerWide/Operations/Certificates/CertificateDefinition.cs
@@ -56,8 +56,9 @@ namespace Raven.Client.ServerWide.Operations.Certificates
 
     public enum DatabaseAccess
     {
-        ReadWrite,
-        Admin
+        Read = 0,
+        ReadWrite = 1,
+        Admin = 2
     }
 
     public enum SecurityClearance
--- a/src/Raven.Server/Dashboard/AbstractDashboardNotification.cs
+++ b/src/Raven.Server/Dashboard/AbstractDashboardNotification.cs
@@ -4,12 +4,14 @@ using Sparrow.Json.Parsing;
 
 namespace Raven.Server.Dashboard
 {
+    public delegate bool CanAccessDatabase(string databaseName, bool requiresWrite);
+
     public abstract class AbstractDashboardNotification : IDynamicJson
     {
         public abstract DashboardNotificationType Type { get; }
-        
+
         public DateTime Date => SystemTime.UtcNow;
-        
+
         public virtual DynamicJsonValue ToJson()
         {
             return new DynamicJsonValue
@@ -23,6 +25,6 @@ namespace Raven.Server.Dashboard
         /// Convert object to DynamicJsonValue but using provided filter
         /// Return null when filter matches nothing - such message will be skipped
         /// </summary>
-        public abstract DynamicJsonValue ToJsonWithFilter(Func<string, bool> filter);
+        public abstract DynamicJsonValue ToJsonWithFilter(CanAccessDatabase filter);
     }
 }
--- a/src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
@@ -11,7 +11,7 @@ namespace Raven.Server.Documents.Handlers.Debugging
 {
     public class QueriesDebugHandler : DatabaseRequestHandler
     {
-        [RavenAction("/databases/*/debug/queries/kill", "POST", AuthorizationStatus.ValidUser)]
+        [RavenAction("/databases/*/debug/queries/kill", "POST", AuthorizationStatus.ValidUser, EndpointType.Write)]
         public Task KillQuery()
         {
             var name = GetQueryStringValueAndAssertIfSingleAndNotEmpty("indexName");
@@ -31,7 +31,7 @@ namespace Raven.Server.Documents.Handlers.Debugging
             return NoContent();
         }
 
-        [RavenAction("/databases/*/debug/queries/running", "GET", AuthorizationStatus.ValidUser, IsDebugInformationEndpoint = true)]
+        [RavenAction("/databases/*/debug/queries/running", "GET", AuthorizationStatus.ValidUser, EndpointType.Read, IsDebugInformationEndpoint = true)]
         public async Task RunningQueries()
         {
             using (ContextPool.AllocateOperationContext(out DocumentsOperationContext context))
@@ -67,7 +67,7 @@ namespace Raven.Server.Documents.Handlers.Debugging
             }
         }
 
-        [RavenAction("/databases/*/debug/queries/cache/list", "GET", AuthorizationStatus.ValidUser, IsDebugInformationEndpoint = true)]
+        [RavenAction("/databases/*/debug/queries/cache/list", "GET", AuthorizationStatus.ValidUser, EndpointType.Read, IsDebugInformationEndpoint = true)]
         public async Task QueriesCacheList()
         {
             using (ContextPool.AllocateOperationContext(out DocumentsOperationContext context))
--- a/src/Raven.Server/Documents/Handlers/Debugging/ServerWideDebugInfoPackageHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/Debugging/ServerWideDebugInfoPackageHandler.cs
@@ -199,7 +199,7 @@ namespace Raven.Server.Documents.Handlers.Debugging
             var contentDisposition = $"attachment; filename={DateTime.UtcNow:yyyy-MM-dd H:mm:ss} - Node [{ServerStore.NodeTag}].zip";
             HttpContext.Response.Headers["Content-Disposition"] = contentDisposition;
             HttpContext.Response.Headers["Content-Type"] = "application/zip";
-            
+
             using (ServerStore.ContextPool.AllocateOperationContext(out JsonOperationContext context))
             {
                 using (var ms = new MemoryStream())
--- a/src/Raven.Server/NotificationCenter/Handlers/ServerNotificationCenterHandler.cs
+++ b/src/Raven.Server/NotificationCenter/Handlers/ServerNotificationCenterHandler.cs
@@ -6,6 +6,7 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features.Authentication;
 using Raven.Client.Util;
+using Raven.Server.Dashboard;
 using Raven.Server.NotificationCenter.Notifications;
 using Raven.Server.Routing;
 using Raven.Server.Web;
@@ -14,7 +15,7 @@ namespace Raven.Server.NotificationCenter.Handlers
 {
     public class ServerNotificationCenterHandler : ServerNotificationHandlerBase
     {
-        [RavenAction("/server/notification-center/watch", "GET", AuthorizationStatus.ValidUser)]
+        [RavenAction("/server/notification-center/watch", "GET", AuthorizationStatus.ValidUser, EndpointType.Read)]
         public async Task Get()
         {
             using (var webSocket = await HttpContext.WebSockets.AcceptWebSocketAsync())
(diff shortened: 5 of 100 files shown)


src/Raven.Client/ServerWide/Operations/Certificates/CertificateDefinition.cs
```diff
@@ -56,8 +56,9 @@ public DynamicJsonValue ToJson()
 
     public enum DatabaseAccess
     {
-        ReadWrite,
-        Admin
+        Read = 0,
```
ayende, line 59: To note: Making `Read` the default means that if we don't have anything, it will be _less_ privileges then now.

src/Raven.Server/Dashboard/AbstractDashboardNotification.cs
```diff
@@ -23,6 +23,6 @@ public virtual DynamicJsonValue ToJson()
         /// Convert object to DynamicJsonValue but using provided filter
         /// Return null when filter matches nothing - such message will be skipped
         /// </summary>
-        public abstract DynamicJsonValue ToJsonWithFilter(Func<string, bool> filter);
+        public abstract DynamicJsonValue ToJsonWithFilter(Func<string, bool, bool> filter);
```
ayende: Probably time to create a dedicated delegate with param names.

src/Raven.Server/Documents/Handlers/Admin/AdminConfigurationHandler.cs
```diff
@@ -18,7 +18,7 @@ namespace Raven.Server.Documents.Handlers.Admin
 {
     public class AdminConfigurationHandler : DatabaseRequestHandler
     {
-        [RavenAction("/databases/*/admin/configuration/settings", "GET", AuthorizationStatus.DatabaseAdmin, IsDebugInformationEndpoint = true)]
+        [RavenAction("/databases/*/admin/configuration/settings", "GET", AuthorizationStatus.DatabaseAdmin, EndpointType.Read, IsDebugInformationEndpoint = true)]
```
ayende: Should we allow this? I mean, this is a read operation, but admin only for a reason.
Making sure that Admin + Read aren't allowed.

src/Raven.Server/Documents/Handlers/Admin/AdminLogsHandler.cs
```diff
@@ -13,7 +12,7 @@ namespace Raven.Server.Documents.Handlers.Admin
 {
     public class AdminLogsHandler : ServerRequestHandler
     {
-        [RavenAction("/admin/logs/configuration", "GET", AuthorizationStatus.Operator)]
+        [RavenAction("/admin/logs/configuration", "GET", AuthorizationStatus.Operator, EndpointType.Read)]
```
ayende: Same note on operator + read

src/Raven.Server/Documents/Handlers/Admin/AdminRevisionsHandler.cs
```diff
@@ -15,7 +15,7 @@ namespace Raven.Server.Documents.Handlers.Admin
 {
     public class AdminRevisionsHandler : DatabaseRequestHandler
     {
-        [RavenAction("/databases/*/admin/revisions", "DELETE", AuthorizationStatus.DatabaseAdmin)]
+        [RavenAction("/databases/*/admin/revisions", "DELETE", AuthorizationStatus.DatabaseAdmin, EndpointType.Write)]
```
ayende: In general, I don't like the `Write` option. It is not needed. If the value isn't set, it isn't allowed for read is a better mode.

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
```diff
@@ -31,7 +31,7 @@ public Task KillQuery()
             return NoContent();
         }
 
-        [RavenAction("/databases/*/debug/queries/running", "GET", AuthorizationStatus.ValidUser, IsDebugInformationEndpoint = true)]
+        [RavenAction("/databases/*/debug/queries/running", "GET", AuthorizationStatus.ValidUser, EndpointType.Read, IsDebugInformationEndpoint = true)]
```
ayende, line 34: That gives you the option to see the queries for other users, though. 
I wonder if we should allow it for read only users

src/Raven.Server/Documents/Handlers/Debugging/ServerWideDebugInfoPackageHandler.cs
```diff
@@ -60,7 +58,7 @@ await transactionOperationContext.ReadForMemoryAsync(HttpContext.Request.Body, "
             }
         }
 
-        [RavenAction("/admin/debug/cluster-info-package", "GET", AuthorizationStatus.Operator, IsDebugInformationEndpoint = true)]
+        [RavenAction("/admin/debug/cluster-info-package", "GET", AuthorizationStatus.Operator, EndpointType.Read, IsDebugInformationEndpoint = true)]
```
ayende: That contains all sort of private information, I don't like it marked as read, even with Operator marker. 
We should whitelist the endpoints, and by default block them for this

src/Raven.Server/NotificationCenter/Handlers/ServerNotificationCenterHandler.cs
```diff
@@ -113,23 +112,25 @@ public abstract class ServerNotificationHandlerBase : RequestHandler
...
-                    return s => false; // deny everything
+                    return (_, __) => false; // deny everything
 
                 case RavenServer.AuthenticationStatus.Operator:
                 case RavenServer.AuthenticationStatus.ClusterAdmin:
                     return null;
 
                 case RavenServer.AuthenticationStatus.Allowed:
-                    return database =>
+                    return (database, requireWrite) =>
```
ayende, line 123: I don't understand under what scenario `requireWrite` would be true?

src/Raven.Server/ServerWide/DebugInfoPackageUtils.cs
```diff
@@ -57,18 +57,21 @@ public static IEnumerable<RouteInformation> GetAuthorizedRoutes(RavenServer.Auth
...
                         if (route.AuthorizationStatus != AuthorizationStatus.ClusterAdmin)
                             authorized = true;
                         break;
+
                     case RavenServer.AuthenticationStatus.Allowed:
                         if (route.AuthorizationStatus == AuthorizationStatus.ClusterAdmin || route.AuthorizationStatus == AuthorizationStatus.Operator)
                             break;
                         if (route.TypeOfRoute == RouteInformation.RouteType.Databases
-                            && (db == null || authenticateConnection.CanAccess(db, route.AuthorizationStatus == AuthorizationStatus.DatabaseAdmin) == false))
+                            && (db == null || authenticateConnection.CanAccess(db, requireAdmin: route.AuthorizationStatus == AuthorizationStatus.DatabaseAdmin, requireWrite: route.IsReadOnlyEndpoint) == false))
```
ayende: ` requireWrite: route.IsReadOnlyEndpoint` ?? That seems like the other way around.

ppekrol:
@ayende I've added some tests + normalized access to database endpoints (for DatabaseAdmin mostly). Could you review that part carefully?

src/Raven.Server/NotificationCenter/Handlers/ServerNotificationCenterHandler.cs
ppekrol: Dismiss and Postpone

src/Raven.Server/Documents/Handlers/Debugging/ServerWideDebugInfoPackageHandler.cs
ppekrol: This requries Operator. For Operator and ClusterAdmin EndpointType is actually ignored (not introducing at the moment read-only Operator or read-only Cluster Admin permissions)

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
ppekrol: This is GET endpoints. The purpose of this feature is not to isolate one database user from another, just to not allow to make any modifications.

src/Raven.Server/Documents/Handlers/Admin/AdminRevisionsHandler.cs
ppekrol: I prefer this to be explicit. Less error prone.

src/Raven.Server/Documents/Handlers/Admin/AdminLogsHandler.cs
ppekrol: I prefer this to be explicit. Less error prone.

src/Raven.Server/Documents/Handlers/Admin/AdminConfigurationHandler.cs
ppekrol: If endpoint has DatabaseAdmin requirement, the endpoint type is not checked.

src/Raven.Client/ServerWide/Operations/Certificates/CertificateDefinition.cs
ppekrol: Not sure if this is a problem? I can move it to be last.

src/Raven.Server/Documents/Handlers/Debugging/ServerWideDebugInfoPackageHandler.cs
ayende: My point exactly, This is really confusing when reading it. 
We should go the other way, explicitly allow endpoints.

src/Raven.Server/Documents/Handlers/Debugging/QueriesDebugHandler.cs
ayende: You are allowing a read only user (rare, limited) to see queries from admin users.
That is going to be surprising.

ppekrol: You have the same issue with ReadWrite

src/Raven.Server/Documents/Handlers/Debugging/ServerWideDebugInfoPackageHandler.cs
ppekrol: Not following

ayende: You made the `EndpointType` a required value.
And then we have `Read` endpoint with operator access.

Instead, make it so `AllowedReadOnlyAccess` is the optional property, and we explicitly allow certain endpoints.

Debug Package, running queries, etc - are not allowed for read only certs.
Basically, we just want them to be able to use basic functions is studio and read documents.

ppekrol: The purpose of this access level should be to disallow changing (writing) in the database. Why someone with ReadWrite can see e.g. the running queries for all users, but someone with Read cannot? It does not make sense to me.

ayende: It is a matter of use cases.

* ReadWrite - current model
* ReadOnly - user has explicitly asked to do something that limit permissions. That is likely the result of wanting to give some limited access.  At this point, we have to consider this as a security boundary. And we want to reduce access as much as possible. 

It is more the case of I can't think of a good reason _to_ allow them to see all the running queries, for example.

ayende: It is much easier to add access down the line, harder to limit it.

ppekrol: No, I think you are talking here more about new user type, not a Read permission for a database. That in my opinion would be something different e.g `AuthorizationStatus.ValidReadOnlyUser`

ayende: I don't think that a user would make any such distinction. The _purpose_ of the feature it to allow limited access scenario, and we should default to as close as we can be.

PR #11821, opened 2021-03-02
RavenDB-16089 Add more time to wait for time series deletion

efratshenhar:


test/StressTests/Client/TimeSeries/TimeSeriesStress.cs
```diff
@@ -84,7 +84,7 @@ public async Task RapidRetention()
                 var check = true;
                 while (check)
                 {
-                    Assert.True(sp.Elapsed < ((TimeSpan)retention).Add((TimeSpan)retention / 5),
+                    Assert.True(sp.Elapsed < ((TimeSpan)retention).Add((TimeSpan)retention / 2),
```
karmeli87: From tests on local machine we saw the min time for the test to complete is about 2;12 min and the max time is 2:46 setting the timeout here to 3 min, I think it is too close.

If this is only a timeout issue I don't want false alarm here so let's set to 4m e.g. `((TimeSpan)retention).Add((TimeSpan)retention)`.

PR #11822, opened 2021-03-02
RavenDB-16099 Database remove/add while subscribe to changes

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-16099

### DIFF
--- a/src/Raven.Client/Http/NodeSelector.cs
+++ b/src/Raven.Client/Http/NodeSelector.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Threading;
 using Raven.Client.Exceptions;
+using Raven.Client.Exceptions.Database;
 
 namespace Raven.Client.Http
 {
@@ -78,7 +79,7 @@ namespace Raven.Client.Http
             }
 
             if (state.Nodes.Count == 0)
-                throw new AllTopologyNodesDownException("There are no nodes in the topology.");
+                throw new DatabaseDoesNotExistException("There are no nodes in the topology.");
 
             throw new RequestedNodeUnavailableException($"Could not find requested node {nodeTag}.");
         }
@@ -118,7 +119,7 @@ namespace Raven.Client.Http
             // if there are all marked as failed, we'll chose the first
             // one so the user will get an error (or recover :-) );
             if (state.Nodes.Count == 0)
-                throw new AllTopologyNodesDownException("There are no nodes in the topology at all");
+                throw new DatabaseDoesNotExistException("There are no nodes in the topology at all");
 
             return (0, state.Nodes[0]);
         }
--- a/src/Raven.Server/Web/System/DatabasesHandler.cs
+++ b/src/Raven.Server/Web/System/DatabasesHandler.cs
@@ -90,14 +90,13 @@ namespace Raven.Server.Web.System
 
             using (ServerStore.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
             {
-                var dbId = Constants.Documents.Prefix + name;
                 using (context.OpenReadTransaction())
-                using (var dbBlit = ServerStore.Cluster.Read(context, dbId, out long _))
+                using (var rawRecord = ServerStore.Cluster.ReadRawDatabaseRecord(context, name))
                 {
                     if (TryGetAllowedDbs(name, out var _, requireAdmin: false) == false)
                         return Task.CompletedTask;
 
-                    if (dbBlit == null)
+                    if (rawRecord == null)
                     {
                         // here we return 503 so clients will try to failover to another server
                         // if this is a newly created db that we haven't been notified about it yet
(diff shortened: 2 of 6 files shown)


src/Raven.Client/Http/NodeSelector.cs
```diff
@@ -78,7 +79,7 @@ public bool OnUpdateTopology(Topology topology, bool forceUpdate = false)
             }
 
             if (state.Nodes.Count == 0)
-                throw new AllTopologyNodesDownException("There are no nodes in the topology.");
+                throw new DatabaseDoesNotExistException("There are no nodes in the topology.");
```
ppekrol, line 82: Is this correct?

src/Raven.Client/Http/NodeSelector.cs
```diff
@@ -118,7 +119,7 @@ private static (int Index, ServerNode Node) GetPreferredNodeInternal(NodeSelecto
             // if there are all marked as failed, we'll chose the first
             // one so the user will get an error (or recover :-) );
             if (state.Nodes.Count == 0)
-                throw new AllTopologyNodesDownException("There are no nodes in the topology at all");
+                throw new DatabaseDoesNotExistException("There are no nodes in the topology at all");
```
ppekrol, line 122: Is this correct?

src/Raven.Server/Web/System/DatabasesHandler.cs
```diff
@@ -89,67 +89,76 @@ public Task GetTopology()
             }
 
             using (ServerStore.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
+            using (context.OpenReadTransaction())
```
ppekrol: @haludi can you point me to the change in this code and the reasoning behind it?

haludi: In general, if we try to get the topology of deleted database we will get an error with `["Database-Missing"]` in the header.
There is a race condition when the database is in delete progress so we can get empty topology.
I changed it to return `["Database-Missing"]` in the header in this situation as well.
```
if (dbRecord.Topology.Members.Count == 0 && dbRecord.Topology.Rehabs.Count == 0 && dbRecord.DeletionInProgress.Any())
{
    // The database at deletion progress from all nodes
    HttpContext.Response.StatusCode = (int)HttpStatusCode.ServiceUnavailable;
    HttpContext.Response.Headers["Database-Missing"] = name;
    using var writer = new BlittableJsonTextWriter(context, HttpContext.Response.Body);
    context.Write(writer, new DynamicJsonValue
        {
            ["Type"] = "Error",
            ["Message"] = "Database " + name + " was deleted"
        });
}
```

src/Raven.Client/Http/NodeSelector.cs
haludi: The `Nodes` collection built from `Topology.Members` and `Topology.Rehabs` so to get this empty those two need to be empty.
As I understand it, this situation only possible if we in a delete process.
The server fix in the database handler is to prevent the client from getting empty topology

src/Raven.Server/Web/System/DatabasesHandler.cs
```diff
@@ -123,9 +123,22 @@ public Task GetTopology()
...
+                    {
+                        // The database at deletion progress from all nodes
+                        HttpContext.Response.StatusCode = (int)HttpStatusCode.ServiceUnavailable;
+                        HttpContext.Response.Headers["Database-Missing"] = name;
+                        using var writer = new BlittableJsonTextWriter(context, HttpContext.Response.Body);
+                        context.Write(writer, new DynamicJsonValue
+                        {
+                            ["Type"] = "Error", 
+                            ["Message"] = "Database " + name + " was deleted"
+                        });
```
ppekrol, line 135: shouldn't we return here?

src/Raven.Client/Http/NodeSelector.cs
ppekrol: cc @karmeli87

src/Raven.Server/Web/System/DatabasesHandler.cs
haludi: Yes

src/Raven.Server/Web/System/DatabasesHandler.cs
```diff
@@ -123,9 +123,23 @@ public Task GetTopology()
                         return Task.CompletedTask;
                     }
 
+                    var dbRecord = JsonDeserializationCluster.DatabaseRecord(dbBlit);
```
ppekrol: Can we switch the code in this method to use RawDatabaseRecord?

src/Raven.Client/Http/NodeSelector.cs
karmeli87: Looks fine to me. Have you ever hit that exception?

haludi: In my test, it hit that consistently

ppekrol (approved):

PR #11823, opened 2021-03-02
RavenDB-15746 Added configuration option which allows to control ind…

arekpalinski:
…ex throttling. That allows to delay indexing for frequently updated sources - e.g. time series

https://issues.hibernatingrhinos.com/issue/RavenDB-15746

### DIFF
--- a/src/Raven.Server/Utils/ThrottledManualResetEventSlim.cs
+++ b/src/Raven.Server/Utils/ThrottledManualResetEventSlim.cs
@@ -1,0 +1,224 @@
+﻿using System;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+using Sparrow.Threading;
+using Sparrow.Utils;
+
+namespace Raven.Server.Utils
+{
+    public class ThrottledManualResetEventSlim : IDisposable
+    {
+        public enum ThrottlingBehavior
+        {
+            Automatic,
+            ManualManagement
+        }
+
+        private readonly ManualResetEventSlim _mre;
+        private readonly MultipleUseFlag _setCalled = new MultipleUseFlag();
+        private readonly ThrottlingBehavior _throttlingBehavior;
+        private readonly CancellationToken _token;
+        private bool _throttlingStarted;
+        private TimeSpan? _throttlingInterval;
+        private Task _timerTask;
+        private CancellationTokenSource _timerCts;
+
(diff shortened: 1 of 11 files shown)


src/Raven.Server/Utils/ThrottledManualResetEventSlim.cs
```diff
@@ -0,0 +1,103 @@
...
+using System.Threading;
+using Sparrow.Threading;
+
+namespace Raven.Server.Utils
+{
+    public class ThrottledManualResetEventSlim : IDisposable
+    {
+        private readonly ManualResetEventSlim _mre;
+        private readonly MultipleUseFlag _setCalled = new MultipleUseFlag();
+        private Timer _timer;
```
arekpalinski: TODO: Instead of creating separate timers for each `ThrottledManualResetEventSlim ` instance we might use pool of timers having same interval and just register there

src/Raven.Server/Utils/ThrottledManualResetEventSlim.cs
```diff
@@ -0,0 +1,103 @@
...
+        private TimeSpan? _throttlingTimeInterval;
+        private bool _throttlingEnabled;
+
+        public ThrottledManualResetEventSlim(TimeSpan? throttlingTime, bool initialState = false)
+        {
+            _mre = new ManualResetEventSlim(initialState);
+
+            if (throttlingTime != null)
+                EnableThrottling(throttlingTime.Value);
```
arekpalinski: TODO: we spawn a timer immediately after creating the instance - while the index might be disabled. The timer should be active only if an index is enabled

src/Raven.Server/Utils/ThrottledManualResetEventSlim.cs
ayende: We already have TimeoutManager, no? Exactly for that.

arekpalinski: Initially I was thinking about creating something dedicated here - similar impl to what we have in `TimeoutManager`, but indeed it looks we can just spawn a task for that purpose and use `TimeoutManager` directly. Something like:

```
_timerTask = Task.Run(async () =>
{
	while (_token.IsCancellationRequested == false)
	{
		await TimeoutManager.WaitFor(_throttlingTimeInterval.Value, _token);

		if (_setCalled.Lower())
		{
			_mre.Set();
		}
	}
}, _token);
```

ppekrol (approved):

ayende (approved):

PR #11828, opened 2021-03-03
RavenDB-16253 : Cannot append timeseries with DateTime.MinValue

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-16253

### DIFF
--- a/src/Raven.Server/Documents/TimeSeries/TimeSeriesStats.cs
+++ b/src/Raven.Server/Documents/TimeSeries/TimeSeriesStats.cs
@@ -210,6 +210,13 @@ namespace Raven.Server.Documents.TimeSeries
 
             void HandleLiveSegment()
             {
+                if (segment.NumberOfEntries == 1 && start > end)
+                {
+                    // new series
+                    start = end = baseline;
+                    return;
+                }
+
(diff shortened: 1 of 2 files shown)


src/Raven.Server/Documents/TimeSeries/TimeSeriesStats.cs
```diff
@@ -210,6 +210,13 @@ public long UpdateStats(DocumentsOperationContext context, TimeSeriesSliceHolder
 
             void HandleLiveSegment()
             {
+                if (segment.NumberOfEntries == 1 && start > end)
```
ayende, line 213: Under what scenario do you have `not (start > end) ` ?

ayende (approved):

src/Raven.Server/Documents/TimeSeries/TimeSeriesStats.cs
aviv86: `start > end` only when table doesn't exist (`start` is initialized to `DateTime.MaxValue` and `end` to `DateTime.MinValue`)
if the table exists `start <= end`

PR #11832, opened 2021-03-04
RavenDB-14939 custom analyzers

ppekrol:


src/Raven.Server/Documents/Handlers/Admin/AdminAnalyzersHandler.cs
```diff
@@ -0,0 +1,83 @@
...
+                    command.Analyzers.Add(analyzerDefinition);
+                }
+
+                var index = (await ServerStore.SendToLeaderAsync(command)).Index;
+
+                await Database.RachisLogIndexNotifications.WaitForIndexNotification(index, ServerStore.Engine.OperationTimeout);
+
+                NoContentStatus();
+
+                HttpContext.Response.StatusCode = (int)HttpStatusCode.Created;
```
ayende: Why both these lines?

src/Raven.Server/Documents/Indexes/Analysis/AnalyzerCompilationCache.cs
```diff
@@ -0,0 +1,158 @@
...
+
+namespace Raven.Server.Documents.Indexes.Analysis
+{
+    public static class AnalyzerCompilationCache
+    {
+        private static readonly ConcurrentDictionary<CacheKey, Lazy<Type>> AnalyzersCache = new ConcurrentDictionary<CacheKey, Lazy<Type>>();
+
+        private static readonly ConcurrentDictionary<CacheKey, Lazy<Type>> AnalyzersPerDatabaseCache = new ConcurrentDictionary<CacheKey, Lazy<Type>>();
+
+        public static Type GetAnalyzerType(string name, string databaseName)
```
ayende, line 16: I think it would be better to use the hash of the analyzer text here, instead.
These are going to live forever, and if we have multiple databases, an analzyer is going to be used across many of them.

For that matter, can I define a default custom analyzer?

ayende: That would also save the compilation costs, the need to remove analyzers, etc.

test/SlowTests/Issues/RavenDB_14939.cs
```diff
@@ -0,0 +1,252 @@
...
+            public MyIndex()
+            {
+                Map = customers => from customer in customers
+                                   select new
+                                   {
+                                       Name = customer.Name
+                                   };
+
+                Indexes.Add(x => x.Name, FieldIndexing.Search);
+                Analyzers.Add(x => x.Name, "MyAnalyzer");
```
ayende, line 242: How is the index name determined?

test/SlowTests/Issues/RavenDB_14939.cs
```diff
@@ -0,0 +1,252 @@
...
+                Fill(store);
+
+                var errors = WaitForIndexingErrors(store);
+                Assert.Equal(1, errors.Length);
+                Assert.Equal(1, errors[0].Errors.Length);
+                Assert.Contains("Cannot find analyzer type 'MyAnalyzer' for field: @default", errors[0].Errors[0].Error);
+
+                store.Maintenance.Send(new PutAnalyzersOperation(new AnalyzerDefinition
+                {
+                    Name = "MyAnalyzer",
```
ayende, line 175: If I changed to to `MyFooBar`, how would this impact other things?

src/Raven.Server/Documents/Handlers/Admin/AdminAnalyzersHandler.cs
ppekrol: We want 'Created', but we also want to remove the 'Content-Type' from the headers. It is done like that in bunch of places. Let me refactor.

src/Raven.Server/Documents/Indexes/Analysis/AnalyzerCompilationCache.cs
ppekrol: There are two caches here. One global-level to avoid compiling same thing all over. Second per database so users cannot use analyzers that were not meant to be used in this database. Identical mechanism for custom sorters.

Yes, you can use configuration setting to define a default analyzer for a given database/server. Now when I'm thinking about that, you cannot define sorters/analyzers server-wide. You are suggesting to add this ability?

test/SlowTests/Issues/RavenDB_14939.cs
ppekrol: New analyzer registered under `MyFooBar`

test/SlowTests/Issues/RavenDB_14939.cs
ppekrol:
```csharp
        public static Type GetAnalyzerType(string name, string analyzerTypeAsString, string databaseName)
        {
            var analyzerType = LuceneAssembly.GetType(analyzerTypeAsString) ??
                               Type.GetType(analyzerTypeAsString) ??
                               Type.GetType("Raven.Server.Documents.Indexes.Persistence.Lucene.Analyzers." + analyzerTypeAsString) ??
                               LuceneAssembly.GetType("Lucene.Net.Analysis." + analyzerTypeAsString) ??
                               LuceneAssembly.GetType("Lucene.Net.Analysis.Standard." + analyzerTypeAsString) ??
                               AnalyzerCompilationCache.GetAnalyzerType(analyzerTypeAsString, databaseName); // <- added this

            if (analyzerType == null)
                throw new InvalidOperationException($"Cannot find analyzer type '{analyzerTypeAsString}' for field: {name}");

            return analyzerType;
        }
```

src/Raven.Server/Documents/Indexes/Analysis/AnalyzerCompilationCache.cs
ayende: The issue is that we now have a static cache that never expires.
When are you removing from `AnalyzersPerDatabaseCache ` ?

ppekrol: This is handled in `public static void AddAnalyzers(DatabaseRecord databaseRecord)` we are removing analyzers that are no longer in the DatabaseRecord there.

test/SlowTests/Issues/RavenDB_14939.cs
ayende: `MyAnalyzer` comes from where?

src/Raven.Server/Documents/Indexes/Analysis/AnalyzerCompilationCache.cs
ayende: And when the database is removed?

test/SlowTests/Issues/RavenDB_14939.cs
ppekrol: From the compilation cache? Not sure if I understand you?

ayende (approved):
We still need to handle global default custom analyzer , no?

ppekrol:
Yes, but I will handle that as a separate issue. Would like to handle also CustomSorters.

ppekrol:
Created an issue for that: https://issues.hibernatingrhinos.com/issue/RavenDB-16328

PR #11833, opened 2021-03-04
RavenDB-1234 - log replication connection error as Operations

grisha-kotler:


src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -349,8 +349,8 @@ private ReplicationLatestEtagRequest IncomingInitialHandshake(TcpConnectionOptio
             }
             catch (Exception e)
             {
-                if (_log.IsInfoEnabled)
-                    _log.Info($"Connection from [{connectionInfo}] is rejected.", e);
+                if (_log.IsOperationsEnabled)
+                    _log.Operations($"Connection from [{connectionInfo}] is rejected.", e);
```
ayende, line 354: This may very well flood the log, no? 
That is _very_ common error if the other side isn't up. For example, we may setup replication to a dev laptop, then they go off for a few weeks. This will fill up the logs with errors that aren't really that interesting to the user.

ayende (changes requested):

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
ppekrol: Why you want to make this change?

grisha-kotler: This is called when we have an incoming connection.
We had missing documents that faiiled to replicate and there was nothing in the log files (default is Operations).
We waited for a bit in the Admin Logs in the studio but could have missed it (depends on the retry mechanism)

ayende: Still has a problem that this will flood the logs, no?

ppekrol:
Closing due to inactivity and incorrect issue #

PR #11840, opened 2021-03-07
RavenDB-16155 Test - modify wait for replication

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-16155

### DIFF
--- a/test/SlowTests/Client/Attachments/AttachmentsReplication.cs
+++ b/test/SlowTests/Client/Attachments/AttachmentsReplication.cs
@@ -1572,10 +1572,7 @@ namespace SlowTests.Client.Attachments
                     await session.StoreAsync(new User { Name = "EGR" }, "users/1");
                     await session.SaveChangesAsync();
 
-                    using (var a1 = new MemoryStream(new byte[] { 1, 2, 3 }))
-                    {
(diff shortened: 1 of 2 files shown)


test/SlowTests/Client/Attachments/AttachmentsReplication.cs
```diff
@@ -1564,136 +1564,111 @@ public async Task AttachmentWithSameStreamSameNameAndDifferentContentTypeShouldB
         [Fact]
         public async Task ResolvedToLatestOnAttachmentConflictShouldRemoveDuplicateAttachment()
         {
-            using (var store1 = GetDocumentStore())
-            using (var store2 = GetDocumentStore())
+            using var store1 = GetDocumentStore();
+            using var store2 = GetDocumentStore();
```
ppekrol: This single change is causing wall of pink. Please revert this and avoid doing this kind of refactoring in future. I'm unable to track changes here.

haludi: In this case, I actually rearrange the whole test.
If I will separate this in a different commit?

ppekrol: Sure

PR #11847, opened 2021-03-08
RavenDB-16273: Only the safe changes. 

redknightlois:


src/Sparrow/Json/Parsing/ObjectJsonParser.cs
```diff
@@ -577,24 +577,30 @@ private void SetStringBuffer(string str)
...
-            var size = byteCount + escapePositionsSize;
-            if (_currentStateBuffer == null || _currentStateBuffer.SizeInBytes < size)
+            int byteCount = str.Length * 5 + escapePositionsSize;
+            if (_currentStateBuffer == null || _currentStateBuffer.SizeInBytes < byteCount)
             {
-                if (_currentStateBuffer != null)
-                    _ctx.ReturnMemory(_currentStateBuffer);
-                _currentStateBuffer = _ctx.GetMemory(size);
-                Debug.Assert(_currentStateBuffer != null && _currentStateBuffer.Address != null);
+                byteCount = Encodings.Utf8.GetMaxByteCount(str.Length);
```
ayende, line 586: why is this line needed?

ayende (approved):

src/Sparrow/Json/Parsing/ObjectJsonParser.cs
redknightlois: Because we are playing safe with 5.2, the fallback is still the same as it was before. Will probably remove it entirely on 6.0 but not for now. [Only the safe changes]

PR #11848, opened 2021-03-08
RavenDB-16226: Avoid repeated type checks. Type objects are unique.

redknightlois:


src/Raven.Server/Utils/TypeConverter.cs
```diff
@@ -390,42 +391,57 @@ internal static IEnumerable<object> Flatten(IEnumerable items)
...
         {
             if (value == null)
                 return DynamicNullObject.ExplicitNull;
 
-            if (value is DynamicNullObject)
-                return value;
+            Type objectType = value.GetType();
 
-            if (value is IDynamicMetaObjectProvider)
+            if (objectType == DynamicRules[(int)DynamicRuleTypeLocation.DynamicNullObject] ||
```
ayende: I thought that `type == `typeof(Foo)` is the cheapest way to do that, no?
Turned into direct constant by the JIT?

redknightlois: Apparently, it does not where there are so many of them. Even though in this case, I believe this is an error, because that is an interface we will need the `is` anyways... 

I have a few tests to perform yet on it, but this is why I believe this is faster: 
image
As you can see, if you are jumping around to get the object type through the pointers table every time; if we can skip that (which we did) we can win a lot.

ayende: `is` vs. `==` is different.
`f is Type` vs. `f.GetType() == typeof(Type)` - this is doing exact match, not for the child classes

redknightlois:
There were multiple checks that could be avoided and some other could be made faster via exact match semantics for type checks.

0.00026805 --- reference ToDynamicType
0.00007309 --- optimized ToDynamicType (-72%)

0.00036740 --- reference ToBlittableSupportedType
0.00030613 --- optimized ToBlittableSupportedType (-16.6%)

src/Raven.Server/Utils/TypeConverter.cs
redknightlois: That is correct. However, that check is actually not needed at all.

src/Sparrow/Utils/TypeCache.cs
```diff
@@ -0,0 +1,92 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Threading.Tasks;
+using Sparrow.Collections;
+
+namespace Sparrow.Utils
+{
+    public sealed class TypeCache<T>
```
ppekrol: Can you make it internal?

src/Sparrow/Json/BlittableJsonReaderArray.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace Sparrow.Json
 {
-    public unsafe class BlittableJsonReaderArray : BlittableJsonReaderBase, IEnumerable<object>, IDisposable
+    public sealed unsafe class BlittableJsonReaderArray : BlittableJsonReaderBase, IEnumerable<object>, IDisposable
```
ppekrol, line 12: We have a warning here:
```
'BlittableJsonReaderArray.Equals(BlittableJsonReaderArray)': new protected member declared in sealed type
```

Can you take a look?

ppekrol (approved):

PR #11851, opened 2021-03-09
RavenDB-16334: Consider the index referenced collections on WaitForIn…

garayx:
…dexesAfterSaveChanges

### DIFF
--- a/src/Raven.Server/Documents/Handlers/BatchHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/BatchHandler.cs
@@ -90,7 +90,7 @@ namespace Raven.Server.Documents.Handlers
                 }
 
                 if (waitForIndexesTimeout != null)
-                    command.ModifiedCollections = new HashSet<string>();
+                    command.ModifiedCollections = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
                 try
                 {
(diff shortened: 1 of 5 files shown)


src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -411,6 +411,16 @@ private static List<Index> GetImpactedIndexesToWaitForToBecomeNonStale(DocumentD
                 }
             }
             return indexesToCheck;
+
+            bool IndexContainsModifiedCollections(Index index)
+            {
+                return index.GetReferencedCollections().Any(kvp =>
```
ppekrol: LINQ here will allocate. Please avoid.

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -411,6 +411,16 @@ private static List<Index> GetImpactedIndexesToWaitForToBecomeNonStale(DocumentD
...
+                {
+                    var collections = kvp.Value.Select(x => x.Name).ToHashSet();
```
ppekrol: This will allocate (per each SaveChanges). Better to introduce virtual method in Index e.g. Index.WorksOnAnyCollection(collections). Implement it in base and override in for indexes that have references collections (the ones that override the GetReferencedCollections).

ayende: Wouldn't it be better to just add a `HashSet` to the `Index` directly? As well as a flag if it works on all collections?

ayende (approved):

PR #11855, opened 2021-03-10
RavenDB-16185 Add global full change vector for replication handshake

efratshenhar:


src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -367,7 +367,7 @@ private ReplicationLatestEtagRequest IncomingInitialHandshake(TcpConnectionOptio
                 using (documentsOperationContext.OpenReadTransaction())
                 using (configurationContext.OpenReadTransaction())
                 {
-                    var changeVector = DocumentsStorage.GetDatabaseChangeVector(documentsOperationContext);
+                    var changeVector = DocumentsStorage.GetFullDatabaseChangeVector(documentsOperationContext);
```
ayende, line 370: Shouldn't we default to the database change vector if empty?

efratshenhar: if full CV is empty so does regular database CV

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -515,6 +519,31 @@ public void SetDatabaseChangeVector(DocumentsOperationContext context, string ch
...
         }
 
+        public static string GetFullDatabaseChangeVector(DocumentsOperationContext context)
+        {
+            var tx = context.Transaction.InnerTransaction;
+            var tree = tx.ReadTree(GlobalTreeSlice);
+            var val = tree.Read(GlobalFullChangeVectorSlice);
+            if (val == null)
+            {
+                return string.Empty;
```
karmeli87: Let's return the normal database change vector here.
Upgrading an existing server will have null as full database change vector..

karmeli87 (changes requested):

src/Raven.Server/Documents/DocumentsStorage.cs
efratshenhar: ok

karmeli87 (approved):

ayende (approved):

PR #11861, opened 2021-03-11
RavenDB-16338: Use sql connection from SqlAwareTestBase in SqlEtlTests

garayx:


ppekrol:
This should be in 4.2, right?

arekpalinski:
@garayx Can you explain what was the underlying issue?

garayx:
> @garayx Can you explain what was the underlying issue?

`WithMsSqlDatabase` uses guid as DB name, so we can now check if the underlying issue is that the DB already exists (and modified to  SINGLE_USER) on the SQL server.

Also now we can use those tests to test the other SQL DBs we support

arekpalinski:
OK

arekpalinski (approved):

PR #11866, opened 2021-03-14
RavenDB-16352 - add id() to the list of index terms

grisha-kotler:


ppekrol (commented):
What about other types of indexes? What about map-reduce?

grisha-kotler:
Map-reduce doesn't have an id field

ppekrol:
But there are additional fields like:
```
                    public const string DocumentIdFieldName = "id()";

                    public const string SourceDocumentIdFieldName = "sourceDocId()";

                    public const string ReduceKeyHashFieldName = "hash(key())";

                    public const string ReduceKeyValueFieldName = "key()";

                    public const string ValueFieldName = "value()";
```

why we want to add only `id()` ? What about other fields?

grisha-kotler:
I'm looking for stored fields I can search on.

PR #11871, opened 2021-03-15
RavenDB-15951 added GC.GetGCMemoryInfo endpoint

ppekrol:


ppekrol:
Is there anything here we would want to expose via SNMP? cc @gregolsky @karmeli87 @ayende

karmeli87:
If I understand the format correctly maybe the (`Index` ,`PauseDurations`) of the `FullBlocking` will be the most useful?

gregolsky:
I presume we would not monitor all these by default, but turn them on if needed. Preferably put into an additional monitoring template.
With that said, would it hurt to expose all of it? We can then pick interesting ones without any additional code changes.

ayende:
Is there significant difference in the mode between each of those? Is this expensive to query?

ppekrol:
I would have to profile it and check. Will do that a bit later.

When it comes to the difference between modes (kinds). It shows last of the requested kind. This is why I thought it would be better to expose all of them at once, including 'any'.

ppekrol:
This is interesting: https://devblogs.microsoft.com/dotnet/the-updated-getgcmemoryinfo-api-in-net-5-0-and-how-it-can-help-you/

ppekrol:
I do not think this is that expensive:
image

PR #11877, opened 2021-03-16
RavenDB-16364 updated Jint

ppekrol:
Sync: https://github.com/ravendb/jint/commit/b27153726faf4158b8aa8001d630861076340a93


src/Raven.Server/Documents/Patch/ReflectionDescriptor.cs
```diff
@@ -0,0 +1,206 @@
+﻿using System;
+using System.Globalization;
+using System.Reflection;
+using Jint;
+using Jint.Native;
+using Jint.Runtime.Descriptors;
+
+namespace Raven.Server.Documents.Patch
+{
+    internal sealed class ReflectionDescriptor : PropertyDescriptor
```
lahma: I was hoping that you could get away with using `ObjectWrapper.GetPropertyDescriptor` when I tested compilation against RavenDB code base with the reflection changes, did you happen to notice that?

ppekrol: Thanks @lahma 

I did not notice that those are available. You definitely know Jint better than I do! Fixed.

lahma: Well I added the method after noticing that you need it for your use case 😉 Great if it works for you!

PR #11882, opened 2021-03-17
RavenDB-16363 Fixing unhandled case of splitting a compressed leaf page. 

arekpalinski:
On the update of the last entry in a compressed page we might not have enough space to put updated value (that's fine). So we did a page splitting which effectively removed the updated entry from existing page (leaving it empty) and put the values to a _new_ (right) page. In result we had a tree with empty page.

### DIFF
--- a/src/Voron/Data/BTrees/TreePageSplitter.cs
+++ b/src/Voron/Data/BTrees/TreePageSplitter.cs
@@ -66,6 +66,40 @@ namespace Voron.Data.BTrees
         {
             using (DisableFreeSpaceUsageIfSplittingRootTree())
             {
+                if (_page.IsLeaf)
+                {
+                    _tree.ClearPagesCache();
+                }
+
+                if (_page.IsCompressed)
+                {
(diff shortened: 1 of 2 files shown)


src/Voron/Data/BTrees/TreePageSplitter.cs
```diff
@@ -66,6 +66,40 @@ private FreeSpaceHandlingDisabler DisableFreeSpaceUsageIfSplittingRootTree()
...
+                    _pageDecompressed.Search(_tx, _newKey);
+
+                    if (_pageDecompressed.LastMatch == 0)
+                    {
+                        // we are going to insert the value in a bit, but it might have 
+                        // been in the compressed portion and not removed by the calling
+                        // code
+                        _tree.RemoveLeafNode(_pageDecompressed);
+
+                        if (_pageDecompressed.NumberOfEntries == 0)
```
arekpalinski, line 86: Here is the fix. I just had to move entire section of code dealing with a compressed page up because we might not need to do any allocation of a new page (`rightPage`) or splitting a root page.

PR #11884, opened 2021-03-18
RavenDB-16373 - throw a better exception

grisha-kotler:


src/Raven.Server/Documents/Indexes/Persistence/Lucene/LuceneIndexWriter.cs
```diff
@@ -83,10 +82,7 @@ public void Commit(IState state)
         public static void TryThrowingBetterException(SystemException e, LuceneVoronDirectory directory)
         {
             if (e.Message.StartsWith("this writer hit an OutOfMemoryError"))
-                ThrowOutOfMemoryException();
-
-            if (e is Win32Exception win32Exception && win32Exception.IsOutOfMemory())
```
grisha-kotler, line 88: This was removed since it will be handled here: 
https://github.com/ravendb/ravendb/blob/v4.2/src/Raven.Server/Documents/Indexes/Index.cs#L1140

PR #11888, opened 2021-03-19
RavenDB-16381: Optimize ToBlittableSupportedType

redknightlois:


src/Raven.Server/Utils/TypeConverter.cs
```diff
@@ -184,150 +184,199 @@ public static object ToBlittableSupportedType(object value, bool flattenArrays =
...
-            if (recursiveLevel > MaxAllowedRecursiveLevelForType)
-                NestingLevelTooDeep(root);
 
-            if (value is JsValue js)
-            {
-                if (js.IsNull())
-                {
-                    if (forIndexing && js is DynamicJsNull dynamicJsNull)
-                        return dynamicJsNull.IsExplicitNull ? DynamicNullObject.ExplicitNull : DynamicNullObject.Null;
+        private enum BlittableSupportedReturnType : int
```
ayende, line 188: Why the duplication with TypeConverter on the client?

src/Raven.Client/Json/TypeConverter.cs
```diff
@@ -7,108 +7,141 @@
...
+            Runtime = 3,
+            Dictionary = 5,
+            GenericDictionary = 6,
+            Enumerable = 7,
+            String = 8
+        }
 
-            if (value is BlittableJsonReaderObject)
-                return value;
+        private static readonly TypeCache<BlittableSupportedReturnType> _supportedTypeCache = new(512);
```
ayende, line 28: What is this class?

src/Raven.Server/Utils/TypeConverter.cs
redknightlois: Because it was like that. The code is actually slightly different, so I didn't dare to unify them.

src/Raven.Client/Json/TypeConverter.cs
redknightlois: It's an optimized open hash table for types. Freaking simple and fast. Designed so that eventually I could do SIMD lookups if needed.

ayende: where is the code for that?

redknightlois:
It is already in use from an earlier PR.

ppekrol:
@redknightlois Can you check failing tests?

redknightlois:
Yes. As soon as I come back from taking the kids to school. These PRs have
priority.

PR #11890, opened 2021-03-19
RavenDB-16383: Avoid having to retrieve from dictionary the id, metadata and collection fields

redknightlois:
This access to the dictionaries is done in the millions when indexing, therefore not having to look for them is a massive improvement in the CPU overhead of the process.

### DIFF
--- a/src/Raven.Client/Json/Serialization/NewtonsoftJson/Internal/BlittableJsonWriter.cs
+++ b/src/Raven.Client/Json/Serialization/NewtonsoftJson/Internal/BlittableJsonWriter.cs
@@ -23,7 +23,8 @@ namespace Raven.Client.Json.Serialization.NewtonsoftJson.Internal
         private readonly DocumentInfo _documentInfo;
 
         public BlittableJsonWriter(JsonOperationContext context, DocumentInfo documentInfo = null,
-            BlittableJsonDocumentBuilder.UsageMode? mode = null, BlittableWriter<UnmanagedWriteBuffer> writer = null)
+            BlittableJsonDocumentBuilder.UsageMode? mode = null, BlittableWriter<UnmanagedWriteBuffer> writer = null,
+            LazyStringValue idField = null, LazyStringValue keyField = null, LazyStringValue collectionField = null)
         {
             _manualBlittableJsonDocumentBuilder = new ManualBlittableJsonDocumentBuilder<UnmanagedWriteBuffer>(context, mode ?? BlittableJsonDocumentBuilder.UsageMode.None, writer);
             _manualBlittableJsonDocumentBuilder.Reset(mode ?? BlittableJsonDocumentBuilder.UsageMode.None);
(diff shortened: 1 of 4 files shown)


src/Raven.Client/Json/Serialization/NewtonsoftJson/Internal/BlittableJsonWriter.cs
```diff
@@ -31,9 +31,13 @@ internal class BlittableJsonWriter : JsonWriter, IJsonWriter
...
-            _metadataKey = context.GetLazyStringForFieldWithCaching(Constants.Documents.Metadata.Key);
-            _metadataCollection = context.GetLazyStringForFieldWithCaching(Constants.Documents.Metadata.Collection);
-            _metadataId = context.GetLazyStringForFieldWithCaching(Constants.Documents.Metadata.Id);
+            _metadataId = context.MetadataIdFieldCache;
+            _metadataKey = context.MetadataKeyFieldCache;
+            _metadataCollection = context.MetadataCollectionFieldCache;
+
+            //_metadataKey = context.GetLazyStringForFieldWithCaching(Constants.Documents.Metadata.Key);
+            //_metadataCollection = context.GetLazyStringForFieldWithCaching(Constants.Documents.Metadata.Collection);
+            //_metadataId = context.GetLazyStringForFieldWithCaching(Constants.Documents.Metadata.Id);
```
ayende: Remove commented code.

src/Sparrow/Json/JsonOperationContext.cs
```diff
@@ -160,6 +165,10 @@ void TryExecute(Action a)
             if (_perCoreLazyStringValuesList.TryPull(out _allocateStringValues) == false)
                 _allocateStringValues = new FastList<LazyStringValue>(256);
 
+            MetadataIdFieldCache = GetLazyStringForFieldWithCaching("@id");
+            MetadataKeyFieldCache = GetLazyStringForFieldWithCaching("@metadata");
+            MetadataCollectionFieldCache = GetLazyStringForFieldWithCaching("@collection");
```
ayende: Are we always going to want to do that? Wouldn't it make more sense to do this lazily?

redknightlois:
For some reason it doesn't allow me to comment on your comment. I decided to go against check if null and lazy do it, to avoid the comparison altogether... those things are small, they are used everywhere and we can just do it at the construction.

src/Sparrow/Json/JsonOperationContext.cs
ppekrol: I do not like this change, especially because we are clearly mixing layers here (pushing to Sparrow strings from upper layers - clearly visible, because Constants cannot be used). We need to find a better way to do this IMO.

redknightlois:
I completely agree with that. I don't like it either, I just couldn't find
any reasonable alternative, that is why I wanted for you both to take a
look and suggest how to do so.

ayende:
You are putting that in JsonContext, but we can put that in DocumetnOperationContext, no?

ppekrol:
@redknightlois ping me later and we will find a good spot for this

ppekrol (approved):

PR #11904, opened 2021-03-23
RavenDB-16378 Fixing restore of tombstones from a backup:

arekpalinski:
- making sure that tombstones of artificial documents aren't restored (they are going to be created in the result of processing tombstone of source documents)
- making sure that flags of tombstones are preserved after the restore
- applyting same logic of dealing with tombstone flags as we do with flags of documents


karmeli87 (commented):
LGTM and we definitely need this fix.

But I would like first to reproduce the NRE issue that we had (and have a test) before merging this PR, to make sure we fix that case as well and not mask it with this fix.

PR #11926, opened 2021-03-30
RavenDB-16418 IAsyncDocumentSession.StreamAsync method forcedly cast …

### LINKED ISSUE #11914: IAsyncDocumentSession.StreamAsync method forcedly cast IAsyncDocumentQuery<T> to concreate AsyncDocumentQuery<T> class implementation.
Hi Acid Raven team, :)

We are experiencing the issue with our **MiniProfiler** implementation which is based on implementation of **RavenDB** async interfaces like **IAsyncDocumentSession, IAsyncDocumentQuery** etc..
Let's say we have something like: **class ProfilerAsyncDocumentQuery : IAsyncDocumentQuery {...}**.

If we call in this class **IAsyncDocumentSession.Task<IAsyncEnumerator<StreamResult<T>>> StreamAsync<T>(IAsyncDocumentQuery<T> query, CancellationToken token = default)** metohd.
we are getting the **InvalidCastException**, because of this line:
https://github.com/ravendb/ravendb/blob/649dc64525966b03e2f1e5c226bcb03b1ba60e99/src/Raven.Client/Documents/Session/AsyncDocumentSession.Stream.cs#L173

So, can you please take a look on that? From my point of view there is a design issue - if I see an interface within the method signature, I don't expect that somewhere inside of this method it would be casted to any concreate implementation. 

Thanks!

P.S. RavenDB Rocks! ;)

Best regards,
Dmitriy

ayende:
…IAsyncDocumentQuery<T> to concreate AsyncDocumentQuery<T> class implementation

Fixes #11914

### DIFF
--- a/src/Raven.Client/Documents/Session/AsyncDocumentSession.Stream.cs
+++ b/src/Raven.Client/Documents/Session/AsyncDocumentSession.Stream.cs
@@ -70,13 +70,13 @@ namespace Raven.Client.Documents.Session
         public class YieldStream<T> : AbstractYieldStream<StreamResult<T>>
         {
             private readonly AsyncDocumentSession _parent;
-            private readonly AsyncDocumentQuery<T> _query;
+            private readonly IAbstractDocumentQueryImpl<T> _query;
             private readonly FieldsToFetchToken _fieldsToFetch;
 
             internal YieldStream(
-                AsyncDocumentSession parent,
+                AsyncDocumentSession parent, 
                 StreamOperation.YieldStreamResults enumerator,
-                AsyncDocumentQuery<T> query,
+                IAbstractDocumentQueryImpl<T> query, 
                 FieldsToFetchToken fieldsToFetch,
                 CancellationToken token) :
                 base(enumerator, token)
@@ -89,7 +89,8 @@ namespace Raven.Client.Documents.Session
             internal override StreamResult<T> ResultCreator(StreamOperation.YieldStreamResults asyncEnumerator)
             {
                 var current = asyncEnumerator.Current;
-                _query?.InvokeAfterStreamExecuted(current);
+                if (_query is IAsyncDocumentQuery<T> q)
+                    q.InvokeAfterStreamExecuted(current);
                 return _parent.CreateStreamResult<T>(current, _fieldsToFetch, _query?.IsProjectInto ?? false);
             }
         }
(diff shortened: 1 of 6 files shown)


src/Raven.Client/Documents/Session/IAsyncDocumentQueryImpl.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    /// <summary>
+    /// This is used as an abstraction for the implementation
+    /// of an aysnc query when passing to other parts of the
+    /// query infrastructure. Meant to be internal only, making
+    /// this public to allow mocking / instrumentation. 
+    /// </summary>
+    public interface IAsyncDocumentQueryImpl<T> : IAsyncDocumentQuery<T>
+    {
+        FieldsToFetchToken FieldsToFetchToken { get; set; }
```
ppekrol: `get` is enough here?

src/Raven.Client/Documents/Session/IAsyncDocumentQueryImpl.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Raven.Client.Documents.Session
+{
+    /// <summary>
+    /// This is used as an abstraction for the implementation
+    /// of an aysnc query when passing to other parts of the
+    /// query infrastructure. Meant to be internal only, making
+    /// this public to allow mocking / instrumentation. 
+    /// </summary>
+    public interface IAsyncDocumentQueryImpl<T> : IAsyncDocumentQuery<T>
```
ppekrol: I think we need exactly the same thing for sync session

src/Raven.Client/Documents/Session/AsyncDocumentSession.Stream.cs
```diff
@@ -170,7 +170,7 @@ private async Task<IAsyncEnumerator<StreamResult<T>>> PerformQueryStreamOperatio
         {
             using (AsyncTaskHolder())
             {
-                var documentQuery = (AsyncDocumentQuery<T>)query;
+                var documentQuery = (IAsyncDocumentQueryImpl<T>)query;
```
ppekrol: few lines below we are casting again
```
var queryOperation = ((AsyncDocumentQuery<T>)query).InitializeQueryOperation();
```

PR #11932, opened 2021-03-31
RavenDB-16419 Fix error handling of IdleFlushTimer 

arekpalinski:


ppekrol (approved):

src/Raven.Server/Documents/Handlers/Debugging/StorageHandler.cs
```diff
@@ -18,6 +18,52 @@ namespace Raven.Server.Documents.Handlers.Debugging
...
+            var env = Database.GetAllStoragesEnvironment()
+                .FirstOrDefault(x => string.Equals(x.Name, name, StringComparison.OrdinalIgnoreCase) && x.Type == type);
+
+            if (env == null)
+            {
+                HttpContext.Response.StatusCode = (int)HttpStatusCode.NotFound;
+                return Task.CompletedTask;
+            }
+
+            env.Environment.SuggestSyncDataFile();
```
ayende: Should be force here, not suggest

src/Voron/StorageEnvironment.cs
```diff
@@ -158,9 +161,6 @@ public StorageEnvironment(StorageEnvironmentOptions options)
                 else // existing db, let us load it
                     LoadExistingDatabase();
 
-                if (_options.ManualFlushing == false)
-                    Task.Run(IdleFlushTimer);
-
```
ayende: Why did you remove this?

src/Raven.Server/Documents/Handlers/Debugging/StorageHandler.cs
arekpalinski: Fixed in https://github.com/ravendb/ravendb/pull/11932/commits/bdf631647072d6fc969bf85e3ab27f0a6d352ac6

src/Voron/StorageEnvironment.cs
arekpalinski: Since we moved the creation of the task to `NewLowLevelTransaction()` then at this point it's already created. We create a new transaction when opening or creating the storage.

Added `Debug.Assert` in https://github.com/ravendb/ravendb/pull/11932/commits/6d7e32b3c1f836bc0871e7ba5e1bee94248057d8

PR #11938, opened 2021-04-01
RavenDB-16412 Add the ability to limit number of indexes that can exe…

arekpalinski:
…cute at the same time

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -202,8 +202,6 @@ namespace Raven.Server.Documents.Indexes
 
         private static readonly Size DefaultMaximumMemoryAllocation = new Size(32, SizeUnit.Megabytes);
 
-        private static FifoSemaphore _concurrentlyRunningIndexes;
-
         public long? LastTransactionId => _environment?.CurrentReadTransactionId;
 
         internal bool IsLowMemory => _lowMemoryFlag.IsRaised();
@@ -1035,7 +1033,7 @@ namespace Raven.Server.Documents.Indexes
 
                         IndexingStatsAggregator stats = null;
                         
-                        _concurrentlyRunningIndexes?.Acquire(_indexingProcessCancellationTokenSource.Token);
+                        DocumentDatabase.ServerStore.ServerWideConcurrentlyRunningIndexesLock?.Acquire(_indexingProcessCancellationTokenSource.Token);
 
                         try
                         {
@@ -1217,7 +1215,7 @@ namespace Raven.Server.Documents.Indexes
                         }
                         finally
                         {
-                            _concurrentlyRunningIndexes?.Release();
+                            DocumentDatabase.ServerStore.ServerWideConcurrentlyRunningIndexesLock?.Release();
 
                             stats?.Complete();
                         }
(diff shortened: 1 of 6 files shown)


src/Sparrow/Utils/FifoSemaphore.cs
```diff
@@ -0,0 +1,115 @@
...
+    /// in first out basis. However, one should note that a thread is considered "in" not by when
+    /// it calls any of the acquire methods, but by when it acquires the internal lock inside 
+    /// any of the acquire methods. This is not normally an issue one need worry about.
+    /// </para>
+    /// <para>
+    /// This class supports the safe use of interrupts. An interrupt that occurs within a method of
+    /// this class results in the action performed by the method not occurring.
+    /// </para>
+    /// </remarks>
+    public class FifoSemaphore
```
ppekrol: internal

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -4094,5 +4103,10 @@ public void AssertNotDisposed()
             if (_disposeOne.Disposed)
                 ThrowObjectDisposed();
         }
+
+        public static void LimitNumberOfConcurrentlyRunningIndexes(int maxNumberOfConcurrentlyRunningIndexes)
```
ayende: Can this be called concurrently with indexing? If so, we may end up with releasing when we didn't acuqire.

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -202,6 +202,8 @@ public abstract class Index : ITombstoneAware, IDisposable, ILowMemoryHandler
 
         private static readonly Size DefaultMaximumMemoryAllocation = new Size(32, SizeUnit.Megabytes);
 
+        private static FifoSemaphore _concurrentlyRunningIndexes;
```
ayende: Not a good idea, we may run into issues with concurrency here because of tests using different ServerStores.
Better to simply set it on the ServerStore directly.

src/Sparrow/Utils/FifoSemaphore.cs
```diff
@@ -0,0 +1,115 @@
...
+    {
+        private readonly Queue<OneTimeWaiter> _waitQueue;
+
+        protected readonly object _Lock;
+
+        protected int _Tokens;
+
+        public FifoSemaphore(int tokens)
+        {
```
ayende, line 38: Verify that this is > 0

src/Sparrow/Utils/FifoSemaphore.cs
```diff
@@ -0,0 +1,115 @@
...
+            {
+                _mre.Set();
+            }
+
+            public void Dispose()
+            {
+                _mre?.Dispose();
+            }
+        }
+    }
```
ayende, line 117: This is pretty elegant impl. I like it much better than the originla

ayende (changes requested):

src/Sparrow/Utils/FifoSemaphore.cs
arekpalinski: done

src/Raven.Server/Documents/Indexes/Index.cs
arekpalinski: No. Anyway moved the lock to `ServerStore` so this is no longer relevant

src/Raven.Server/Documents/Indexes/Index.cs
arekpalinski: Very good point. Moved to `ServerStore`

src/Sparrow/Utils/FifoSemaphore.cs
arekpalinski: Done

ppekrol (approved):

PR #11943, opened 2021-04-02
RavenDB-16446 Avoid using index temp files

ayende:


src/Raven.Server/Indexing/TempFileCache.cs
```diff
@@ -0,0 +1,126 @@
...
+        private const long MaxFileSizeToReduce = 16 * Constants.Size.Megabyte;
+        private const int MaxFilesInCache = 32;
+
+        public TempFileCache(StorageEnvironmentOptions options)
+        {
+            _options = options;
+            string path = _options.TempPath.FullPath;
+            if (Directory.Exists(path) == false)
+                Directory.CreateDirectory(path);
+            foreach (string file in Directory.GetDirectories(path,  "*" + StorageEnvironmentOptions.DirectoryStorageEnvironmentOptions.TempFileExtension))
```
ppekrol: This is the same directory that Voron is using. Can we have a conflict with some file that voron is using?
Also... Voron clears that directory on startup, will we have any files here?

src/Raven.Server/Indexing/TempFileCache.cs
```diff
@@ -0,0 +1,126 @@
...
+    /// <summary>
+    /// This assume single threaded and is meant to be used for indexing only
+    /// </summary>
+    public sealed class TempFileCache : IDisposable
+    {
+        private readonly StorageEnvironmentOptions _options;
+        private readonly Queue<FileStream> _files = new Queue<FileStream>();
+        private readonly Queue<MemoryStream> _ms = new Queue<MemoryStream>();
+
+        private const long MaxFileSizeToReduce = 16 * Constants.Size.Megabyte;
```
ppekrol: MaxFileSizeToKeepInBytes

src/Raven.Server/Indexing/TempFileCache.cs
```diff
@@ -0,0 +1,126 @@
...
+            {
+                var info = new FileInfo(file);
+                if (info.Length > MaxFileSizeToReduce || 
+                    _files.Count >= MaxFilesInCache)
+                {
+                    File.Delete(file);
+                }
+                else
+                {
+                    _files.Enqueue(new FileStream(file, FileMode.OpenOrCreate));
```
ppekrol: FileMode.Open?

src/Raven.Server/Indexing/TempFileCache.cs
```diff
@@ -0,0 +1,126 @@
...
+        {
+            FileStream stream;
+            if (_files.Count > 0)
+            {
+                stream = _files.Dequeue();
+            }
+            else
+            {
+                stream = SafeFileStream.Create(
+                    Path.Combine(_options.TempPath.FullPath, Guid.NewGuid() +  StorageEnvironmentOptions.DirectoryStorageEnvironmentOptions.TempFileExtension),
```
ppekrol: Can it be `Lucene_{GUID}.tmp` ?

src/Raven.Server/Indexing/TempFileCache.cs
```diff
@@ -0,0 +1,126 @@
...
+                TempCryptoStream tcs => tcs.InnerStream,
+                FileStream fs => fs,
+                _ => throw new ArgumentOutOfRangeException()
+            };
+
+            if (s.Length > MaxFileSizeToReduce || 
+                _files.Count >= MaxFilesInCache)
+            {
+                var fileName = s.Name;
+                s.Dispose();
```
ppekrol: We had `DisposeFile` method. Probably better to have sth like that here as well.

src/Raven.Server/Indexing/TempFileCache.cs
```diff
@@ -0,0 +1,126 @@
...
+        }
+        
+        public void Dispose()
+        {
+            foreach (FileStream file in _files)
+            {
+                string fileName = null;
+                try
+                {
+                    fileName = file.Name;
```
ppekrol: We had `DisposeFile` method. Probably better to have sth like that here as well.

src/Raven.Server/Indexing/TempFileCache.cs
```diff
@@ -0,0 +1,126 @@
...
+    /// This assume single threaded and is meant to be used for indexing only
+    /// </summary>
+    public sealed class TempFileCache : IDisposable
+    {
+        private readonly StorageEnvironmentOptions _options;
+        private readonly Queue<FileStream> _files = new Queue<FileStream>();
+        private readonly Queue<MemoryStream> _ms = new Queue<MemoryStream>();
+
+        private const long MaxFileSizeToReduce = 16 * Constants.Size.Megabyte;
+        private const int MaxFilesInCache = 32;
```
ppekrol: MaxFilesToKeepInCache

src/Raven.Server/Indexing/VoronIndexOutput.cs
```diff
@@ -18,60 +22,51 @@ public class VoronIndexOutput : BufferedIndexOutput
...
+                {
+                    if (_ms.Length + len <= _ms.Capacity)
+                    {
+                        _ms.Write(b, offset, len);
+                        return;
+                    }
+                    // too big, copy the buffer to the file
+                    _file = _fileCache.RentFileStream();
+                    _ms.CopyTo(_file);
+                    _fileCache.ReturnMemoryStream(_ms);
```
ppekrol, line 61: _ms should be set to `null` afterwards?

src/Raven.Server/Indexing/VoronIndexOutput.cs
```diff
@@ -85,21 +80,27 @@ public override void Seek(long pos)
...
-        public override long Length => _file.Length;
+        public override long Length => _ms?.Length ?? _file.Length;
 
         public override void SetLength(long length)
         {
             try
             {
-                _file.SetLength(length);
+                if(_ms != null)
+                    _ms.SetLength(length);
```
ppekrol, line 96: what if that length exceeds the capacity?

src/Raven.Server/Indexing/LuceneVoronDirectory.cs
```diff
@@ -14,6 +14,7 @@ public sealed unsafe class LuceneVoronDirectory : Lucene.Net.Store.Directory
         private readonly StorageEnvironment _environment;
         private readonly string _name;
         private readonly IndexOutputFilesSummary _indexOutputFilesSummary;
+        private readonly TempFileCache _tmpFileCache;
```
ppekrol, line 17: The problem that I can see here is that size of this can be 32 * 128kb = 4mb

Now if we will take that 4MB and multiply that by the amount of indexes then we are getting quite a lot of memory. Are we OK with that?
Do we want to subscribe to low memory notifications and clear the cache then?

src/Raven.Server/Indexing/TempFileCache.cs
ayende: That is why I'm using the temp extension.
I'm not _expecting_ to have any files here, but this is there to handle the case of us leaving files around because we crashes, so we'll clean them up.

src/Raven.Server/Indexing/TempFileCache.cs
ayende: Not critical either way

src/Raven.Server/Indexing/TempFileCache.cs
ayende: Yes

src/Raven.Server/Indexing/TempFileCache.cs
ppekrol: Ok, but can we have a conflict with some of the files that Voron is using? I do not recall, but double-checking.

src/Raven.Server/Indexing/VoronIndexOutput.cs
ayende: Yes!

src/Raven.Server/Indexing/LuceneVoronDirectory.cs
ayende: Good point, I'll add handling to that.

src/Raven.Server/Indexing/VoronIndexOutput.cs
ayende: Cannot happen, this is used only to trim the file, not to extend it.
At any rate, that would increased the MemoryStream size.

ppekrol (approved):

src/Raven.Server/Indexing/TempFileCache.cs
```diff
@@ -0,0 +1,156 @@
...
+        private const long MaxFileSizeToReduceInBytes = 16 * Constants.Size.Megabyte;
+        private const int MaxFilesToKeepInCache = 32;
+
+        public TempFileCache(StorageEnvironmentOptions options)
+        {
+            _options = options;
+            string path = _options.TempPath.FullPath;
+            if (Directory.Exists(path) == false)
+                Directory.CreateDirectory(path);
+            foreach (string file in Directory.GetDirectories(path,  "lucene-*" + StorageEnvironmentOptions.DirectoryStorageEnvironmentOptions.TempFileExtension))
```
arekpalinski: There shouldn't be any such temp files at this point because we delete them on storage startup. See https://github.com/ravendb/ravendb/pull/11497/files (https://issues.hibernatingrhinos.com/issue/RavenDB-15911)

ayende: Yes, we don't expect that. However, I'm including this here in case the server crashed.
If that happened, we don't want the old files to still remain there (nothing will remove them).
For that purpose, we scan and remove / reuse them as needed.

ayende:
Probably need to merge that to 4.2 LTS as well

PR #11944, opened 2021-04-02
RavenDB-16251 support for native DLLs from NuGet

ppekrol:
TODO:
- [x] Linux support


src/Raven.Server/Documents/Indexes/Static/IndexCompiler.cs
```diff
@@ -468,6 +464,24 @@ static Assembly LoadAssembly(string path)
...
+                    return;
+
+                foreach (string library in package.Libraries)
+                {
+                    var assembly = LoadAssembly(library);
+                    references.Add(RegisterAssembly(assembly));
+                }
+
+                if (userDefined)
+                    NuGetNativeLibraryResolver.RegisterPath(package.NativePath);
```
ppekrol, line 504: Need to make a decision if we want to register only native DLLs from packages that user defined or also from dependencies. I'm a bit afraid to 'let it loose' and register everything.

PR #11957, opened 2021-04-07
RavenDB-16367 implemented LockMode in DatabaseRecord

ppekrol:


src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -717,48 +721,61 @@ public async Task Delete()
...
+                            if (rawRecord == null)
+                                continue;
+
+                            switch (rawRecord.LockMode)
                             {
-                                if (rawRecord == null)
+                                case DatabaseLockMode.Unlock:
+                                    databasesToDelete.Add(databaseName);
+                                    break;
+                                case DatabaseLockMode.PreventDeletesIgnore:
```
ayende, line 744: Add a note to the log? Maybe an alert?
For that matter, if this is set, we probably want to record who is trying the delete (ip, cert, etc, no)?

ppekrol: We are recording that beforehand (auditlog). Maybe we actually should record that after (or both?) ?

ayende: Audit log may not be configured.

src/Raven.Server/Web/System/AdminDatabasesHandler.cs
```diff
@@ -1393,6 +1419,51 @@ await using (var writer = new AsyncBlittableJsonTextWriter(context, ResponseBody
...
+                        databasesToUpdate.Add(databaseName);
+                    }
+                }
+
+                if (databasesToUpdate.Count > 0)
+                {
+                    long index = 0;
+                    foreach (var databaseName in databasesToUpdate)
+                    {
+                        //var result = await ServerStore.WriteDatabaseRecordAsync(kvp.DatabaseRecord.DatabaseName, kvp.DatabaseRecord, kvp.Index, raftRequestId);
```
ml054: remove?

ml054:
@ppekrol I'm waiting for icons - once I get them will apply them and it will ready for merge.

ml054:
@ppekrol icons are in place so everything is ready on my side.

PR #11959, opened 2021-04-08
RavenDB-16464 Fixing unreleased allocations in scratch buffers which made that scratch files were never cleaned up.

arekpalinski:
The issue was that we didn't clean all pages in scrach files because potentially there could be some read transactions still looking at those pages but at the same time we decided that related journals are unused so we removed them from in-memory list of journals. That resulted in the bug that we never freed some allocations in the scratch buffer files.

### DIFF
--- a/src/Voron/Impl/Journal/WriteAheadJournal.cs
+++ b/src/Voron/Impl/Journal/WriteAheadJournal.cs
@@ -643,10 +643,10 @@ namespace Voron.Impl.Journal
                         if (oldestActiveTransaction != 0)
                             maxTransactionId = Math.Min(oldestActiveTransaction - 1, maxTransactionId);
 
-                        foreach (var modifedPagesInTx in journalFile.PageTranslationTable.GetModifiedPagesForTransactionRange(
+                        foreach (var modifiedPagesInTx in journalFile.PageTranslationTable.GetModifiedPagesForTransactionRange(
                             lastFlushed.TransactionId, maxTransactionId))
                         {
-                            foreach (var pagePosition in modifedPagesInTx)
+                            foreach (var pagePosition in modifiedPagesInTx)
                             {
                                 if (pagePosition.Value.IsFreedPageMarker)
                                 {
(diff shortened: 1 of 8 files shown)


src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -826,52 +828,93 @@ public void ApplyLogsToDataFile(CancellationToken token, TimeSpan timeToWait)
...
-                var lastSyncedTransactionId =
-                    Math.Min(Math.Min(lastFlushedTransactionId, _waj._env.CurrentReadTransactionId - 1), txw.Id - 1);
+                            var oldestActiveTransactionAfterPreventingNewReadTransactions = _waj._env.ActiveTransactions.OldestTransaction;
 
-                // we have to free pages of the unused journals before the remaining ones that are still in use
-                // to prevent reading from them by any read transaction (read transactions search journals from the newest
-                // to read the most updated version)
+                            if (oldestActiveTransactionAfterPreventingNewReadTransactions == txw.Id)
+                            {
+                                lastFlushedTransactionIdThatWontReadFromJournal = Math.Min(lastFlushedTransactionId, txw.Id - 1);
```
arekpalinski, line 859: We're doing this specifically to delete the last journal (potentially large) in order to avoid applying it after the restart. See https://github.com/ravendb/ravendb/pull/8326 for reference.

src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -749,17 +747,21 @@ public void ApplyLogsToDataFile(CancellationToken token, TimeSpan timeToWait)
...
-                            UpdateJournalStateUnderWriteTransactionLock(txw, lastProcessedJournal, lastFlushedTransactionId, unusedJournals);
+                            UpdateJournalStateUnderWriteTransactionLock(txw, journalSnapshots, lastProcessedJournal, lastFlushedTransactionId);
+
+                            if (_waj._logger.IsInfoEnabled)
+                                _waj._logger.Info($"Updated journal state under write tx lock (txId: {txw.Id}) after waiting for {sp.Elapsed}");
                         }
                         catch (Exception e)
                         {
+                            if (_waj._logger.IsOperationsEnabled)
+                                _waj._logger.Operations($"Failed to update journal state under write tx lock (waited - {sp.Elapsed})");
```
ayende, line 758: Missing the actual exception in the logger.

arekpalinski: Fixed in https://github.com/ravendb/ravendb/pull/11972

PR #11971, opened 2021-04-12
RavenDB-16450 Import MultiGetCommand from 5.0 and set result to  use cloned Blittable instead of Blittable based on cached memory

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-16450

### DIFF
--- a/src/Raven.Client/Documents/Commands/MultiGet/MultiGetCommand.cs
+++ b/src/Raven.Client/Documents/Commands/MultiGet/MultiGetCommand.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Buffers;
 using System.Collections.Generic;
 using System.IO;
 using System.Net;
@@ -12,14 +13,15 @@ using Sparrow.Json.Parsing;
 
 namespace Raven.Client.Documents.Commands.MultiGet
 {
-    public class MultiGetCommand : RavenCommand<GetResponse[]>
+    public class MultiGetCommand : RavenCommand<GetResponse[]>, IDisposable
     {
         private readonly RequestExecutor _requestExecutor;
         private readonly HttpCache _cache;
         private readonly List<GetRequest> _commands;
 
         private string _baseUrl;
-
+        private (HttpCache.ReleaseCacheItem release, BlittableJsonReaderObject cached)[] _cachedValues;
+        
         internal bool AggressivelyCached;
 
         public MultiGetCommand(RequestExecutor requestExecutor, List<GetRequest> commands)
@@ -34,8 +36,6 @@ namespace Raven.Client.Documents.Commands.MultiGet
         {
             _baseUrl = $"{node.Url}/databases/{node.Database}";
             url = $"{_baseUrl}/multi_get";
-
-            Result = new GetResponse[_commands.Count];
 
             if (MaybeReadAllFromCache(ctx, _requestExecutor.AggressiveCaching.Value))
             {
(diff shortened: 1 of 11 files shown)


haludi:
https://github.com/ravendb/ravendb/blob/45f33a7c78954bdb731df961ecf298bb49c2268a/src/Raven.Client/Http/HttpCache.cs#L132

I wonder if we should add by `mem.SizeInBytes` instead of `result.Size` since this is the actual allocated size
```
var mem = _unmanagedBuffersPool.Allocate(result.Size);
result.CopyTo(mem.Address);
if (Interlocked.Add(ref _totalSize, result.Size) > _maxSize)
// if (Interlocked.Add(ref _totalSize, mem.SizeInBytes) > _maxSize)
{
    if (_isFreeSpaceRunning == false)
        Task.Run(FreeSpace);
}
```

ayende:
That might be a good idea, yes

src/Raven.Client/Documents/Commands/MultiGet/MultiGetCommand.cs
```diff
@@ -12,29 +13,35 @@
 
 namespace Raven.Client.Documents.Commands.MultiGet
 {
-    public class MultiGetCommand : RavenCommand<List<GetResponse>>
+    public class MultiGetCommand : RavenCommand<GetResponse[]>, IDisposable
```
ppekrol: This is a breaking change, we cannot switch from List to Array

src/Raven.Client/Documents/Commands/MultiGet/MultiGetCommand.cs
```diff
@@ -12,29 +13,35 @@
...
     {
+        private readonly RequestExecutor _requestExecutor;
         private readonly HttpCache _cache;
         private readonly List<GetRequest> _commands;
 
         private string _baseUrl;
+        private (HttpCache.ReleaseCacheItem release, BlittableJsonReaderObject cached)[] _cachedValues;
```
ppekrol: should start from uppercase: `Release` `Cached`

ppekrol (commented):
Ping me, I want to go over this together

src/Raven.Client/Documents/Commands/MultiGet/MultiGetCommand.cs
```diff
@@ -296,9 +315,30 @@ private void MaybeSetCache(GetResponse getResponse, GetRequest command)
...
+            public readonly (HttpCache.ReleaseCacheItem Release, BlittableJsonReaderObject Cached)[] Values;
+
+            public Cached(int size)
+            {
+                Values = ArrayPool<(HttpCache.ReleaseCacheItem, BlittableJsonReaderObject)>.Shared.Rent(size);
+            }
+
+            public void Dispose()
+            {
+                for (int i = 0; i < Values.Length; i++)
```
ppekrol: Values.Length might be much greater than requested size. Won't you get NRE here?

src/Raven.Client/Documents/Commands/MultiGet/MultiGetCommand.cs
```diff
@@ -296,9 +315,30 @@ private void MaybeSetCache(GetResponse getResponse, GetRequest command)
...
+        private class Cached : IDisposable
+        {
+            public readonly (HttpCache.ReleaseCacheItem Release, BlittableJsonReaderObject Cached)[] Values;
+
+            public Cached(int size)
+            {
+                Values = ArrayPool<(HttpCache.ReleaseCacheItem, BlittableJsonReaderObject)>.Shared.Rent(size);
+            }
+
+            public void Dispose()
```
ppekrol, line 336: It does not check if it was disposed before. Will it result in double disposal since we are disposing it earlier as well?

src/Raven.Client/Documents/Commands/MultiGet/MultiGetCommand.cs
haludi: Basically, it is struct so the dispose in `HttpCache.ReleaseCacheItem` will do nothing since its item is null.
But it is more appropriate to use the initialize size so I will change it

src/Raven.Client/Documents/Commands/MultiGet/MultiGetCommand.cs
haludi: I set the Cache to null after disposing but I will do additional check here

PR #11993, opened 2021-04-16
RavenDB-15988 support for FindIndex and FindLastIndex in indexes

ppekrol:


src/Raven.Server/Documents/Indexes/Static/DynamicArray.cs
```diff
@@ -138,6 +138,42 @@ public dynamic All(Func<dynamic, bool> predicate)
             return Enumerable.All(this, predicate);
         }
 
+        public int FindIndex(Predicate<object> match)
+        {
+            var items = Enumerable.ToList(this);
+            return items.FindIndex(match);
+        }
+
+        public int FindIndex(int startIndex, Predicate<object> match)
```
ayende:
```suggestion
        public int FindIndex(int startIndex, Predicate<dynamic> match)
```

ayende: And for the rest as well

ppekrol: Fixed. My bad.

PR #12008, opened 2021-04-20
RavenDB-16507 Making the acquirement and usage of a pager state more robust by:

arekpalinski:
- using EnsurePagerStateReference method when we want to get a pager state - it guarantees that we'll get the latest state and will add ref atomically
- releasing the pager state if any exception happens - it's very impornant because we already called AddRef()
- passing state paramter as ref to EnsurePagerStateReference because we might have further usage of this variable and inside EnsurePagerStateReference we could get newest state (while the one passed as parameter could be disposed)

https://issues.hibernatingrhinos.com/issue/RavenDB-16507
https://issues.hibernatingrhinos.com/issue/RavenDB-16128

### DIFF
--- a/src/Voron/Impl/Paging/AbstractPager.cs
+++ b/src/Voron/Impl/Paging/AbstractPager.cs
@@ -287,13 +287,13 @@ namespace Voron.Impl.Paging
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private byte* AcquirePagePointerInternal(IPagerLevelTransactionState tx, long pageNumber, PagerState pagerState)
         {
+            var state = pagerState;
+
             if (DisposeOnceRunner.Disposed)
                 goto AlreadyDisposed;
 
             if (pageNumber > NumberOfAllocatedPages || pageNumber < 0)
                 goto InvalidPageNumber;
-
-            var state = pagerState;
 
             if (state == null)
             {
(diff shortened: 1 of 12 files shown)


ayende (approved):

src/Voron/Impl/Paging/AbstractPager.cs
```diff
@@ -303,14 +303,21 @@ public int NodeMaxSize
                     goto AlreadyDisposed;
             }
 
-            tx?.EnsurePagerStateReference(state);
+            tx?.EnsurePagerStateReference(ref state);
+            
+            if (state._released)
+                goto InvalidPagerState;
```
arekpalinski, line 309: @redknightlois Please have a look at this change. I followed the way we throw exceptions in this method and used `goto` statement as in other two places here.

src/Voron/Impl/Paging/AbstractPager.cs
```diff
@@ -303,14 +303,21 @@ public int NodeMaxSize
                     goto AlreadyDisposed;
             }
 
-            tx?.EnsurePagerStateReference(state);
+            tx?.EnsurePagerStateReference(ref state);
```
arekpalinski, line 306: This is the most important change - passing `state` using `ref` so we can be sure that below `state.MapBase ...` usage is valid. The issue was that `state` could be disposed during `EnsurePagerStateReference` but we were still using it below. See the test which proves that can be an issue.

That could cause AVE (like here - https://issues.hibernatingrhinos.com/issue/RavenDB-16128) and also `ObjectDisposedException` (original problem that caused the investigation - https://issues.hibernatingrhinos.com/issue/RavenDB-16507) if such a state was passed to the ctor of LowLevelTransaction used for the transaction cloning feature.

ppekrol (approved):

PR #12016, opened 2021-04-20
RavenDB-15910 Creating a debug package should show progress

ml054:


src/Raven.Server/Documents/Handlers/Debugging/ServerWideDebugInfoPackageHandler.cs
```diff
@@ -67,98 +67,107 @@ public async Task GetClusterWideInfoPackage()
 
             HttpContext.Response.Headers["Content-Disposition"] = contentDisposition;
 
-            using (ServerStore.ContextPool.AllocateOperationContext(out TransactionOperationContext transactionOperationContext))
-            using (ServerStore.ContextPool.AllocateOperationContext(out JsonOperationContext jsonOperationContext))
-            using (transactionOperationContext.OpenReadTransaction())
+            var token = CreateOperationToken();
+            var operationId = GetLongQueryString("operationId", false) ?? ServerStore.Operations.GetNextOperationId();
+
+            await ServerStore.Operations.AddOperation(null, "Created debug package for all cluster nodes", Operations.Operations.OperationType.DebugPackage, async _ =>
```
ayende, line 73: You aren't actually reporting any progress, however?

ml054: Not sure if we want to report progress or result here (used nulls for both). 

The most important thing is we know when operation starts and ends, so we can block UI or show information after package is collected.

PR #12019, opened 2021-04-21
RavenDB-16516 dispose TcpConnectionOptions when doing TestConnection or Ping to avoid getting into CLOSE_WAIT state

ppekrol:


src/Raven.Server/Documents/TcpHandlers/TcpConnectionOptions.cs
```diff
@@ -67,6 +68,10 @@ public void Dispose()
             if (_isDisposed)
                 return;
 
+#if !RELEASE
```
ayende, line 71: Do we actually set this, I thought we had just `DEBUG`

ppekrol: I copied this from HttpCache: https://github.com/ravendb/ravendb/blob/6af136312be49c596af32d3d175d289477bd50a1/src/Raven.Client/Http/HttpCache.cs#L105

DEBUG and RELEASE AFAIK are built-in?

src/Raven.Server/Documents/TcpHandlers/TcpConnectionOptions.cs
```diff
@@ -93,6 +98,13 @@ public void Dispose()
             // a possible race condition on dispose
         }
 
+#if !RELEASE
+        ~TcpConnectionOptions()
+        {
+            throw new LowMemoryException($"Detected a leak on TcpConnectionOptions ('{ToString()}') when running the finalizer.");
```
karmeli87, line 104: Not sure it is a good idea to throw in the finalizer... if this for test running, better is simply to write to console.

ppekrol: This is only in `!RELEASE`, we are already doing this in HttpCache.

ayende: And no one read the console logs

ayende (approved):

PR #12020, opened 2021-04-21
RavenDB-14324 - Don't delete the entire folder when we delete a database

shiranshalom:
https://issues.hibernatingrhinos.com/issue/RavenDB-14324


src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -347,6 +350,34 @@ public void DeleteDatabase(string dbName, DeletionInProgressStatus deletionInPro
...
+                                        throw new IntersectionInPaths($"Cannot delete database {dbName} from {parentDirectory}. " +
+                                                                      $"There is an intersection with database {currRecord.DatabaseName} located in {currDirectory}.");
+                            }
+                        }
+                    }
+                    catch (IntersectionInPaths e)
+                    {
+                        if (_logger.IsInfoEnabled)
+                            _logger.Info(e.Message);
+                        throw;
```
ayende: Consider the ergonomics of this.  We already have error handling, why are you making it so?

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -347,6 +350,34 @@ public void DeleteDatabase(string dbName, DeletionInProgressStatus deletionInPro
...
+                                var currConfiguration = CreateDatabaseConfiguration(currRecord.DatabaseName, ignoreDisabledDatabase: true, 
+                                    ignoreBeenDeleted: true, ignoreNotRelevant: true, databaseRecord: currRecord);
+
+                                var currDirectory = currConfiguration.Core.DataDirectory.FullPath;
+
+                                if (PathUtil.IsSubDirectory(currDirectory, parentDirectory))
+                                        throw new IntersectionInPaths($"Cannot delete database {dbName} from {parentDirectory}. " +
+                                                                      $"There is an intersection with database {currRecord.DatabaseName} located in {currDirectory}.");
+                            }
+                        }
```
ayende: This should be in a separate function

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -1053,5 +1084,32 @@ private void CompleteDatabaseUnloading(DocumentDatabase database)
 
             database.DatabaseShutdownCompleted.Set();
         }
+
+        private class IntersectionInPaths : Exception
```
ayende: Why create a special exception for this?

ayende (changes requested):

src/Raven.Server/Documents/DatabasesLandlord.cs
shiranshalom: If I understand correctly, the reason for adding the new catch is that I don't want the exception thrown to reach:        
image
(I want to reach directly the "finally" section). 
I removed the exception I created and change the exception to "InvalidOperationException", hope it's fine.

src/Raven.Server/Documents/DatabasesLandlord.cs
shiranshalom: Done

src/Raven.Server/Documents/DatabasesLandlord.cs
shiranshalom: Done

src/Raven.Server/Documents/DatabasesLandlord.cs
ayende: Why don't you want it to reach this exception? What is the new error handling that you are implementing here?

shiranshalom: This exception does not include any _throw_ statement, therefore we will reach this code:
image
which include operations that cause file deletion & other inadvisable operations.

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -347,6 +350,12 @@ public void DeleteDatabase(string dbName, DeletionInProgressStatus deletionInPro
                     {
                         configuration = CreateDatabaseConfiguration(dbName, ignoreDisabledDatabase: true, ignoreBeenDeleted: true, ignoreNotRelevant: true,
                             databaseRecord: record);
+
+                        CheckDatabasePathsIntersection(dbName, configuration);
+                    }
+                    catch (InvalidOperationException)
+                    {
+                        throw;
                     }
```
ayende: Avoid the `catch` here put the `CheckDatabasePathsIntersection` after the try,

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -1053,5 +1062,30 @@ private void CompleteDatabaseUnloading(DocumentDatabase database)
...
+                var parentDir = configuration.Core.DataDirectory.FullPath;
+
+                foreach (var currRecord in allDatabasesRecords)
+                {
+                    if (currRecord.DatabaseName.Equals(databaseName))
+                        continue;
+
+                    var currConfiguration = CreateDatabaseConfiguration(currRecord.DatabaseName, ignoreDisabledDatabase: true,
+                        ignoreBeenDeleted: true, ignoreNotRelevant: true, databaseRecord: currRecord);
+
```
ayende: Here, you need a `try`, to avoid errors in creating _another_ database configuration.
In that case, just ignore it.

ayende (changes requested):

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -1053,5 +1058,41 @@ private void CompleteDatabaseUnloading(DocumentDatabase database)
...
+                    RavenConfiguration currConfiguration;
+                    try
+                    {
+                        currConfiguration = CreateDatabaseConfiguration(currRecord.DatabaseName, ignoreDisabledDatabase: true, 
+                            ignoreBeenDeleted: true, ignoreNotRelevant: true, databaseRecord: currRecord);
+                    }
+                    catch (Exception e)
+                    {
+                        currConfiguration = null;
```
ayende:
```suggestion
                        continue;
```

We can't figure out anything here, anyway, so let's skip this explicitly.

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -1053,5 +1058,41 @@ private void CompleteDatabaseUnloading(DocumentDatabase database)
...
+                        if (_logger.IsInfoEnabled)
+                            _logger.Info("Could not create database configuration", e);
+                    }
+
+                    var currDir = currConfiguration?.Core.DataDirectory.FullPath;
```
ayende: Also check the TempPath and JournalsPath

src/Raven.Server/Documents/DatabasesLandlord.cs
shiranshalom: Done.

src/Raven.Server/Documents/DatabasesLandlord.cs
shiranshalom: Done.

ayende (approved):

src/Raven.Server/Documents/DatabasesLandlord.cs
shiranshalom: Done

src/Raven.Server/Documents/DatabasesLandlord.cs
shiranshalom: Done

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -355,6 +358,8 @@ public void DeleteDatabase(string dbName, DeletionInProgressStatus deletionInPro
                             _logger.Info("Could not create database configuration", ex);
                     }
 
+                    CheckDatabasePathsIntersection(dbName, configuration);
```
ppekrol: There is a `if` statement bellow that says that configuration might be null. Shouldn't this be moved inside that `if` ?

shiranshalom: Done.

ppekrol:
@shiranshalom can you check failing tests?

PR #12029, opened 2021-04-22
RavenDB-16519 If we cancel on attemt to get a lock from FifoSemaphore then we must not leave the waiter in the queue.

arekpalinski:


src/Sparrow/Utils/FifoSemaphore.cs
```diff
@@ -56,19 +56,44 @@ public bool TryAcquire(TimeSpan timeout, CancellationToken token)
...
+                _waitQueue.Add(waiter);
             }
 
             using (waiter)
             {
-                return waiter.TryWait(timeout, token);
+                try
+                {
+
+                    return waiter.TryWait(timeout, token);
```
ppekrol: What if this returns false? Same issue?

PR #12033, opened 2021-04-23
RavenDB-15747 add arm64, small deb build script tweaks

gregolsky:


scripts/linux/pkg/deb/assets/test.sh
```diff
@@ -24,7 +24,7 @@ rvnpid=\$!
 
 trap "rc=\$?; pgrep \$rvnpid && kill -9 \$rvnpid 2>&1 >/dev/null; echo \$rc; exit \$rc" EXIT
 
-sleep 5 
+sleep 7
```
ayende, line 27: That looks suspicious and likely to fail, the `--rerty` isn't enough?

gregolsky: `arm64` docker host via qemu is just slower than the other ones I presume. Also that's just my test code to validate the build - we won't be running this on CI.

PR #12034, opened 2021-04-23
RavenDB-16536 Making sure that we'll flush everything even if there is no activity on a database

arekpalinski:


src/Voron/Impl/Journal/WriteAheadJournal.cs
```diff
@@ -576,6 +576,11 @@ public void OnTransactionCommitted(LowLevelTransaction tx)
             public long LastFlushedTransactionId => _lastFlushed.TransactionId;
             public long LastFlushedJournalId => _lastFlushed.JournalId;
             public long TotalWrittenButUnsyncedBytes => Interlocked.Read(ref _totalWrittenButUnsyncedBytes);
+
+            internal int TotalCommittedSinceLastFlushPages;
+            internal bool ShouldFlush => TotalCommittedSinceLastFlushPages != 0 || _lastFlushed.TransactionId != _waj._env.CurrentReadTransactionId; // TODO arek - double check the second condition
```
ayende: Why the second part?

test/SlowTests/Voron/Issues/RavenDB_16536.cs
```diff
@@ -0,0 +1,77 @@
...
+            Assert.True(Env.Journal.Applicator.ShouldFlush);
+
+            onLogsAppliedMre.Reset();
+
+
+            ownFlusher.ForTestingPurposesOnly().AddEnvironmentToFlushQueue(Env);
+
+            // below two lines are necessary
+            Thread.Sleep(1000);
+            Env.TimeToSyncAfterFlushInSec = 0;
```
ayende: Why is this?

test/SlowTests/Voron/Issues/RavenDB_16536.cs
```diff
@@ -0,0 +1,77 @@
...
+
+                    tx.Commit();
+                }
+            }
+
+            var onLogsAppliedMre = new ManualResetEventSlim(false);
+
+            Env.OnLogsApplied += () => onLogsAppliedMre.Set();
+
+            var ownFlusher = new GlobalFlushingBehavior();
```
ayende, line 45: why do you create an instance of this?

src/Voron/Impl/Journal/WriteAheadJournal.cs
arekpalinski: The problem was described here - https://issues.hibernatingrhinos.com/issue/RavenDB-16536

TLDR we had this integer - `SizeOfUnflushedTransactionsInJournalFile` (which I renamed to `TotalCommittedSinceLastFlushPages`) which was used to track if we need to flush or not but it was decreased even before we actually started the flush operation. It didn't take into account that we might not be able to release all committed pages because there might be read transactions running during the flush.

I'm a bit afraid of tracking _exactly_ how many of modified pages has been flushed (there could be some overwrites of modified pages so we're filtering them out etc) so decided that the easiest would be to check if we still have something to flush by tracking the last flushed transaction ID. As commented need to carefully verify that.

test/SlowTests/Voron/Issues/RavenDB_16536.cs
arekpalinski: Sorry, forgot to finish the comments. We have internal checks which makes sure we don't flush too often. Anyway it should be set in the test during `Configure()` call

test/SlowTests/Voron/Issues/RavenDB_16536.cs
arekpalinski: Because I want to manage and control it explicitly in the test

src/Voron/Impl/Journal/WriteAheadJournal.cs
ayende: okay, understood

test/SlowTests/Voron/Issues/RavenDB_16536.cs
arekpalinski: No, we cannot set it via options because options set it to 30 seconds at minimum. Added the comment which explains why we have this.

ppekrol (approved):

PR #12036, opened 2021-04-24
RavenDB-16537 added OnSessionDisposing event

ppekrol:


src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -1340,6 +1342,8 @@ private void Dispose(bool isDisposing)
             if (_isDisposed)
                 return;
 
+            OnSessionDisposing?.Invoke(this, new SessionDisposingEventArgs(this));
```
ayende: What happens if this throws?
What happens if this is an async session and the user may want to invoke something async?

ayende (changes requested):

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
ppekrol: 1. Wrapped in `tryf` (both exceptions)
2. We do not support async event yet: https://issues.hibernatingrhinos.com/issue/RavenDB-14316 added that to 6.0 effort.

ppekrol: 1. Actually I think the 'asyncTasksCounter' exception should be checked before anything else and not in `tryf` block.

ayende: No, I think it should be before `OnDisposing`
Because you may want to complete stuff on there.
If we leak a task from `OnDisposing`, that should be immediately failed.

ppekrol: Changed.

PR #12038, opened 2021-04-26
RavenDB-16538 Fixing unhandled error in DetermineServerIp kills the server with bad error:

arekpalinski:
- Let's make sure that we see the acual host name which we failed to get the IP during Dns.GetHostAddressesAsync
- DetermineServerIp() is going to be called only if server is unsecured because it's needed only then so a failure there won't prevent a secured server to start
- Logging an exception thrown by the validation of the security configuration

### DIFF
--- a/src/Raven.Server/Config/Categories/SecurityConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/SecurityConfiguration.cs
@@ -187,10 +187,6 @@ namespace Raven.Server.Config.Categories
 
                 var isServerUrlHttps = uri.Scheme.Equals("https", StringComparison.OrdinalIgnoreCase);
 
-                var serverAddresses = DetermineServerIp(uri);
-                var unsecuredAccessAddressRange = configuration.Security.UnsecuredAccessAllowed;
-                var serverIsWithinUnsecuredAccessRange = serverAddresses.Any(x => SecurityUtils.IsUnsecuredAccessAllowedForAddress(unsecuredAccessAddressRange, x));
-
                 if (configuration.Security.AuthenticationEnabled)
                 {
                     if (isServerUrlHttps == false)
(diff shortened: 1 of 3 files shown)


src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -229,7 +230,14 @@ private static IPAddress[] DetermineServerIp(Uri serverUri)
                 if (getHostAddressesTask.Wait(TimeSpan.FromSeconds(30)) == false)
                     throw new InvalidOperationException($"Could not obtain IP address from DNS {serverUri.DnsSafeHost} for 30 seconds.");
```
ayende, line 231: This will also kill us, no?

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
```diff
@@ -203,6 +199,11 @@ internal static void Validate(RavenConfiguration configuration)
                     if (isServerUrlHttps)
                         throw new InvalidOperationException($"Configured server address { string.Join(", ", configuration.Core.ServerUrls) } requires HTTPS. Please set up certification information under { RavenConfiguration.GetKey(x => x.Security.CertificatePath) } configuration key.");
 
+                    var serverAddresses = DetermineServerIp(uri);
```
ayende, line 202: This can also throw, no?

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
arekpalinski: Yes. We're throwing in multiple places during `SecurityConfiguration.Validate`. IMO this is not bad, we're preventing a server from running with invalid security settings.

src/Raven.Server/Config/Categories/SecurityConfiguration.cs
arekpalinski: Yes. But I don't think we cannot just ignore this. We want to use the returned IP to verify it using `SecurityUtils.IsUnsecuredAccessAllowedForAddress` here:

https://github.com/ravendb/ravendb/blob/5213dbcc9cbe2b6e24960ce1fac4281bbc69211e/src/Raven.Server/Config/Categories/SecurityConfiguration.cs#L194

PR #12046, opened 2021-04-27
RavenDB-16565 Capture stack traces for entire cluster doesn't work

ml054:


src/Raven.Server/ServerWide/DebugInfoPackageUtils.cs
```diff
@@ -28,8 +28,8 @@ public static string GetOutputPathFromRouteInformation(RouteInformation route, s
             if (path.StartsWith("."))
                 path = path.Substring(1);
 
-            path = string.IsNullOrWhiteSpace(prefix) == false ?
-                Path.Combine(prefix, $"{path}.json") :
+            path = string.IsNullOrWhiteSpace(prefix) == false ?
```
ppekrol, line 30: This needs probably to go into 4.2?

ml054: @ppekrol you're right!

ml054: done

PR #12052, opened 2021-04-27
RavenDB-16357 Support read-only Certificates in Studio - Part 2

Danielle9897:


src/Raven.Studio/typescript/common/shell/accessManager.ts
```diff
@@ -26,6 +26,11 @@ class accessManager {
 
     isValidUser = ko.pureComputed<boolean>(() => this.securityClearance() === "ValidUser");
 
+
+    isAdminByDbName(dbName: string) {
```
ml054, line 30: rename to: isDatabaseAdminOrAbove ?

src/Raven.Studio/typescript/common/shell/accessManager.ts
```diff
@@ -26,6 +26,11 @@ class accessManager {
...
+        return this.getDatabaseAccessLevelTextByDbName(dbName) === "Admin";
```
ml054: use getEffectiveDatabaseAccessLevel here.

Danielle9897: done

ml054 (approved):

PR #12091, opened 2021-05-02
RavenDB-16450 MultiGetCommand - handle cache & respond of not found

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-16450

### DIFF
--- a/src/Raven.Client/Http/HttpCache.cs
+++ b/src/Raven.Client/Http/HttpCache.cs
@@ -14,6 +14,8 @@ namespace Raven.Client.Http
 {
     public class HttpCache : IDisposable, ILowMemoryHandler
     {
+        public const string NotFoundResponse = "404 Response";
+
         private static readonly Logger Logger = LoggingSource.Instance.GetLogger<HttpCache>("Client");
 
         private readonly long _maxSize;
@@ -164,7 +166,7 @@ namespace Raven.Client.Http
             var flag = aggressivelyCached ? ItemFlags.AggressivelyCached : ItemFlags.None;
             var httpCacheItem = new HttpCacheItem
             {
-                ChangeVector = "404 Response",
+                ChangeVector = NotFoundResponse,
                 Ptr = null,
                 Size = 0,
                 Allocation = null,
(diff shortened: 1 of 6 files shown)


src/Raven.Server/Documents/Handlers/DocumentHandler.cs
```diff
@@ -222,7 +223,9 @@ private async Task GetDocumentsByIdAsync(DocumentsOperationContext context, Micr
                 }
                 if (document == null && ids.Count == 1)
                 {
-                    HttpContext.Response.StatusCode = (int)HttpStatusCode.NotFound;
+                    HttpContext.Response.StatusCode = GetStringFromHeaders("If-None-Match") == HttpCache.NotFoundResponse
```
haludi, line 226: When requesting multiple documents that do not exist we respond with `NotModified` (if the previous response was the same) but when requesting only one we respond with `NotFound`.

test/SlowTests/Bugs/Caching/UseCachingInLazyTests.cs
```diff
@@ -422,6 +421,66 @@ await using (var bulk = store.BulkInsert())
...
+                using (var session = store.OpenAsyncSession())
+                {
+                    //Add "NotExistDocId" to cache
+                    await session.Advanced.Lazily.LoadAsync<TestObj>(notExistDocId).Value;
+                    Assert.Equal(0, session.Advanced.NumberOfRequests);
+                }
+            }
+
+#if DEBUG
+            await Task.Delay(10000);
```
haludi: I found it the best way to test it without modifying the production code for the test and without relying too much on the internal implementation.

src/Raven.Client/Http/HttpCache.cs
```diff
@@ -14,6 +14,8 @@ namespace Raven.Client.Http
 {
     public class HttpCache : IDisposable, ILowMemoryHandler
     {
+        public const string NotFoundResponse = "404 Response";
```
ppekrol: private or internal

test/SlowTests/Bugs/Caching/UseCachingInLazyTests.cs
ppekrol: Let's remove this. The whole test suite will die anyway.

Issue #83, opened 2010-05-22
Update client docs to include Linq


ayende:
Right now, it isn't there, and that mislead people

PR #12094, opened 2021-05-03
RavenDB-16357 Support read-only Certificates in Studio - Part 4

Danielle9897:


src/Raven.Studio/typescript/common/shell/accessManager.ts
```diff
@@ -113,19 +113,27 @@ class accessManager {
...
         if (!activeDatabaseAccess) {
             return false;
         }
 
-        if (activeDatabaseAccess === 'Read' &&
-            (requiredAccess === "DatabaseReadWrite" || requiredAccess === "DatabaseAdmin")) {
+        return this.canHandleOperationDatabaseLevelCheck(activeDatabaseAccess, requiredAccess as databaseAccessLevel);
+    }
+    
+    canHandleOperationDatabaseLevelCheck(actualLevel: databaseAccessLevel, requiredLevel: accessLevel) {
```
ml054: do we need separeate function?

src/Raven.Studio/typescript/common/shell/accessManager.ts
```diff
@@ -113,19 +113,27 @@ class accessManager {
...
+
         const activeDatabaseAccess = this.activeDatabaseAccessLevel();
 
         if (!activeDatabaseAccess) {
             return false;
         }
 
-        if (activeDatabaseAccess === 'Read' &&
-            (requiredAccess === "DatabaseReadWrite" || requiredAccess === "DatabaseAdmin")) {
+        return this.canHandleOperationDatabaseLevelCheck(activeDatabaseAccess, requiredAccess as databaseAccessLevel);
```
ml054: can we avoid op operator?

src/Raven.Studio/typescript/models/auth/certificateModel.ts
```diff
@@ -96,8 +96,8 @@ class certificateModel {
                 dbAccessInfo = certificateDefinition.Permissions;
         }
         
-        const dbAccessArray = _.map(dbAccessInfo, (accessLevel: Raven.Client.ServerWide.Operations.Certificates.DatabaseAccess, dbName: string) => 
-            ({ accessLevel: accessLevel,  dbName: dbName }));
+        const dbAccessArray = _.map(dbAccessInfo, (accessLevel: databaseAccessLevel, dbName: string) =>
+            ({ accessLevel: `Database${accessLevel}` as databaseAccessLevel,  dbName: dbName }));
```
ml054, line 100: you pass databaseAccessLevel then you add word database and cast? Looks weird.

src/Raven.Studio/typescript/viewmodels/viewModelBase.ts
```diff
@@ -66,7 +66,21 @@ class viewModelBase {
...
-        return this.databasesManager.activateBasedOnCurrentUrl();
+        const dbNameFromUrl = appUrl.getDatabaseNameFromUrl();
+        if (dbNameFromUrl) {
+            const dbAccessLevel = accessManager.default.getEffectiveDatabaseAccessLevel(dbNameFromUrl);
+            const requiredAccessForView = router.activeInstruction().config.requiredAccess;
+            const isViewAllowed = this.allowToAccessView(dbAccessLevel, requiredAccessForView);
+            
+            return this.databasesManager.activateBasedOnCurrentUrl(dbNameFromUrl, isViewAllowed);
+        } else {
+            return true;
```
ml054: shouldn't we check here access as well?

src/Raven.Studio/typings/_studio/customTypings.d.ts
```diff
@@ -315,6 +315,7 @@ interface DurandalRouteConfiguration {
     tooltip?: string;
     dynamicHash?: KnockoutObservable<string> | (() => string);
     tabName?: string;
+    requiredAccess: accessLevel;
```
ml054: optional ??

src/Raven.Studio/typescript/viewmodels/viewModelBase.ts
```diff
@@ -66,7 +66,21 @@ class viewModelBase {
...
+        }
+    }
+
+    protected allowToAccessView(dbAccessLevel: databaseAccessLevel, requiredAccessForView: accessLevel) {
```
ml054: I think we shoulnd't have database access level here, but generic access level
+ database name as param?

src/Raven.Studio/typescript/common/shell/databasesManager.ts
```diff
@@ -44,7 +44,8 @@ class databasesManager {
     init(): JQueryPromise<Raven.Client.ServerWide.Operations.DatabasesInfo> {
         return this.refreshDatabases()
             .done(() => {
-                this.activateBasedOnCurrentUrl();
+                const dbNameFromUrl = appUrl.getDatabaseNameFromUrl();
+                this.activateBasedOnCurrentUrl(dbNameFromUrl, true);
```
ml054: why we always pass true here?

src/Raven.Studio/typescript/common/shell/databasesManager.ts
```diff
@@ -62,22 +63,29 @@ class databasesManager {
...
+    activateBasedOnCurrentUrl(dbName: string, isViewAllowed: boolean): JQueryPromise<canActivateResultDto> | boolean {
+        if (dbName) {
+            const db = this.getDatabaseByName(dbName);
+            return this.activateIfDifferent(db, dbName, isViewAllowed);
         }
         return true;
     }
 
-    private activateIfDifferent(db: database, dbName: string): JQueryPromise<canActivateResultDto> {
+    private activateIfDifferent(db: database, dbName: string, isViewAllowed: boolean): JQueryPromise<canActivateResultDto> {
```
ml054: can we check if view is allowed inside this method, instead of passing this around?

src/Raven.Studio/typescript/common/shell/databasesManager.ts
```diff
@@ -62,22 +63,29 @@ class databasesManager {
...
+        const databasesListView = appUrl.forDatabases();
 
         const incomingDatabaseName = db ? db.name : undefined;
 
         this.initialized.done(() => {
             const currentActiveDatabase = this.activeDatabaseTracker.database();
+
+            if (!isViewAllowed) {
+                messagePublisher.reportError("Redirecting. Invalid access level.");
+                task.resolve({ redirect: appUrl.forDatabases() });
```
ml054: use databasesListView here?

src/Raven.Studio/typescript/common/shell/databasesManager.ts
```diff
@@ -62,22 +63,29 @@ class databasesManager {
...
         const task = $.Deferred<canActivateResultDto>();
+        const databasesListView = appUrl.forDatabases();
 
         const incomingDatabaseName = db ? db.name : undefined;
 
         this.initialized.done(() => {
             const currentActiveDatabase = this.activeDatabaseTracker.database();
+
+            if (!isViewAllowed) {
+                messagePublisher.reportError("Redirecting. Invalid access level.");
```
ml054: ==> Forbidden access.

Danielle9897: Changing to:
`Redirecting. Access is forbidden.`

ok ?

src/Raven.Studio/typescript/common/shell/databasesManager.ts
Danielle9897: done

src/Raven.Studio/typings/_studio/customTypings.d.ts
Danielle9897: done

src/Raven.Studio/typescript/common/shell/accessManager.ts
Danielle9897: I separated this because it is called from 2 places:
1. from `canHandleOperation` (in accessManager.ts)
2. from `allowToAccessView` (in viewModelBase.ts Line #83)

src/Raven.Studio/typescript/models/auth/certificateModel.ts
Danielle9897: changed to:

```
const dbAccessArray = _.map(dbAccessInfo, (accessLevel: Raven.Client.ServerWide.Operations.Certificates.DatabaseAccess, dbName: string) =>
   ({ accessLevel: `Database${accessLevel}` as databaseAccessLevel,  dbName: dbName }));
```

src/Raven.Studio/typescript/common/shell/accessManager.ts
Danielle9897: removed the 'as' operator

src/Raven.Studio/typescript/common/shell/databasesManager.ts
Danielle9897: Don't we want the check to be made after `this.initialized.done` ?

src/Raven.Studio/typescript/common/shell/accessManager.ts
ml054: ok but I'd like to avoid having separate functions per global and per database. 

I think we should have single function: 

canHandleOperation(level: accessLevel) 

and then we have all magic inside. 

Optionally: we could pass **optional** database name. 

So those function should work for **All** types not only for database level. 

Unless I'm missing something.

src/Raven.Studio/typescript/common/shell/accessManager.ts
ml054: great - we should avoid as + control flow in typescript should automatically infer based on code flow that it has to be databaseAccessLevel. That's the beauty of typescript.

src/Raven.Studio/typescript/models/auth/certificateModel.ts
ml054: yes now it makes more sense. ;-)

src/Raven.Studio/typescript/viewmodels/viewModelBase.ts
ml054: I'm not sure why we return true here w/o any checks. 

For views which are not database specific I assume we should check if we have required clearance or redirect to db lists.

src/Raven.Studio/typescript/viewmodels/viewModelBase.ts
ml054: I don't think we should have required access a parameter of this method...

method should go more or less like this:
protected canAccessView(accessLevel: accessLevel): boolean {
   // in base class fetch required access level for view and check if access is possible
   // but in base classes we can have different logic - all we have in question is: can user with provided (via accessLevel params) access given view? 
  // this method should get all deps like required level - we shouldn't pass that from outside. 
}

src/Raven.Studio/typescript/common/shell/databasesManager.ts
ml054: not sure I follow

src/Raven.Studio/typescript/common/shell/databasesManager.ts
ml054: I'd start with cause then with reaction. 
Access is forbidden. Redirecting to databases view. 

But in this case I'm not sure second part is needed, since the most important thing is: You don't have an access. 
How we "heal" this state is not so important.

Danielle9897: ok - done

src/Raven.Studio/typescript/common/shell/accessManager.ts
Danielle9897: done

src/Raven.Studio/typescript/common/shell/databasesManager.ts
Danielle9897: Removed this part for now - seems to work w/o this

src/Raven.Studio/typescript/common/shell/databasesManager.ts
Danielle9897: `isViewAllowed `is now calculated outside, before calling this method

src/Raven.Studio/typescript/viewmodels/viewModelBase.ts
Danielle9897: Access is now checked in the `canAccessView `method

src/Raven.Studio/typescript/viewmodels/viewModelBase.ts
Danielle9897: We now have:  `protected canAccessView(dbName: string) `

src/Raven.Studio/typescript/common/extensions.ts
```diff
@@ -284,7 +284,8 @@ class extensions {
                             const shouldBeVisibleByKo = visibleValue && !hiddenValue;
                             const isElementVisible = element.style.display !== "none";
 
-                            if (accessManager.default.canHandleOperation(requiredAccessLevel)) {
+                            const activeDbName = activeDatabaseTracker.default.database().name;
```
ml054: activeDatabase.name

src/Raven.Studio/typescript/common/extensions.ts
```diff
@@ -306,7 +307,8 @@ class extensions {
                             const shouldBeEnabledByKo = !disableValue && enableValue;
                             const isElementDisabled = element.hasAttribute("disabled");
 
-                            if (accessManager.default.canHandleOperation(requiredAccessLevel)) {
+                            const activeDbName = activeDatabaseTracker.default.database().name;
```
ml054: same here

src/Raven.Studio/typescript/models/auth/certificateModel.ts
ml054: please avoid as operator

src/Raven.Studio/typescript/viewmodels/viewModelBase.ts
```diff
@@ -66,7 +67,23 @@ class viewModelBase {
...
+            const task = $.Deferred<canActivateResultDto>();
+            messagePublisher.reportError("Access is forbidden. Redirecting to databases view.");
+            task.resolve({ redirect: appUrl.forDatabases() });
+            return task;
+        }
+        
+        return this.databasesManager.activateBasedOnCurrentUrl(dbNameFromUrl);
+    }
+
+    protected canAccessView(dbName: string) {
```
ml054: make db name optional

src/Raven.Studio/typescript/common/extensions.ts
Danielle9897: done

src/Raven.Studio/typescript/common/extensions.ts
Danielle9897: done

src/Raven.Studio/typescript/viewmodels/viewModelBase.ts
Danielle9897: done

ml054 (approved):

PR #12096, opened 2021-05-03
RavenDB-16619 added the ability to set CheckCertificateRevocationList in the Client API

ppekrol:


src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -327,8 +327,14 @@ private static bool ShouldRemoveHttpClient(SocketException exception)
 
         private static readonly Exception ServerCertificateCustomValidationCallbackRegistrationException;
 
+        private static readonly bool? CheckCertificateRevocationList;
+
         static RequestExecutor()
         {
+            var checkCertificateRevocationListAsString = Environment.GetEnvironmentVariable("RAVEN_CHECK_CERTIFICATE_REVOCATION_LIST");
+            if (string.IsNullOrWhiteSpace(checkCertificateRevocationListAsString) == false)
+                CheckCertificateRevocationList = bool.Parse(checkCertificateRevocationListAsString);
```
ayende, line 336: TryGet, to be more robust?

src/Raven.Client/Documents/Conventions/DocumentConventions.cs
```diff
@@ -237,6 +237,7 @@ public DocumentConventions()
         private OperationStatusFetchMode _operationStatusFetchMode;
         private string _topologyCacheLocation;
         private Version _httpVersion;
+        private bool? _checkCertificateRevocationList;
```
ayende, line 240: Why make it nullable? We want to default to `false` always, no?

src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -327,8 +327,14 @@ private static bool ShouldRemoveHttpClient(SocketException exception)
 
         private static readonly Exception ServerCertificateCustomValidationCallbackRegistrationException;
 
+        private static readonly bool? CheckCertificateRevocationList;
+
         static RequestExecutor()
         {
+            var checkCertificateRevocationListAsString = Environment.GetEnvironmentVariable("RAVEN_CHECK_CERTIFICATE_REVOCATION_LIST");
```
gregolsky, line 334: Why not make it part of the Raven Configuration - `SecurityConfiguration`? Then you'd get parsing env vars, cli params and settings all for free.

gregolsky: Umm nvm, that's client-side

ppekrol:
In .NET 4.7.2+ and .NET Core this is set to 'false' which is what we actually want.
In .NET 4.7.1 and before this property throws, the only way to change it is via ServicePointManager.

Closing it down.

PR #12100, opened 2021-05-04
RavenDB-16378 / RavenDB-16401 Failing test which is based on a snapsh…

arekpalinski:
…ot backup made using version 5.0 (so we cannot restore it on 4.2 branch)


test/SlowTests/Issues/RavenDB_16378_Replication_NRE.cs
```diff
@@ -0,0 +1,74 @@
...
+                        session.SaveChanges();
+                    }
+
+                    await SetupReplicationAsync(sourceOfRestoredDb, destination);
+
+                    WaitForDocument(destination, "marker");
+
+                    var stats = destination.Maintenance.Send(new GetStatisticsOperation());
+
+                    Assert.Equal(6, stats.CountOfDocuments);
```
arekpalinski, line 69: Some explanation - here we're having 6 documents because some artificials has been migrated as well. Although there isn't `orders/2` document because the replication batch contained the following items:

```
(etag: 133) - orders/2 (Document) flags - HasRevisions, DeleteRevision - 3/24/2021 6:57:17 PM
(etag: 137) - orders/2 (Document) flags - HasRevisions, FromReplication - 3/24/2021 6:55:10 PM
(etag: 138) - orders/2 (Document) flags - HasRevisions, Revision, FromReplication - 3/24/2021 6:55:10 PM
```

As the result of resolve to latest conflict resolution strategy the one from `3/24/2021 6:57:17 PM` has won.

Again, the snapshot isn't valid but it reproduced real issue.

PR #12106, opened 2021-05-05
RavenDB-16617 Creating a dedicated endpoint for proxy to be sure that…

arekpalinski:
… we'll get info about databases that are accessible using the certificate using the orginal connection

### DIFF
--- a/src/Raven.Server/NotificationCenter/Handlers/ClusterDashboardHandler.cs
+++ b/src/Raven.Server/NotificationCenter/Handlers/ClusterDashboardHandler.cs
@@ -6,11 +6,13 @@
 
 using System;
 using System.IO;
-using System.Linq;
 using System.Net.WebSockets;
 using System.Threading.Tasks;
 using Raven.Client.Http;
+using Raven.Client.ServerWide.Operations.Certificates;
+using Raven.Server.Dashboard;
 using Raven.Server.Dashboard.Cluster;
+using Raven.Server.Json;
 using Raven.Server.Routing;
 using Raven.Server.ServerWide.Context;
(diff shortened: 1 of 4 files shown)


ppekrol (approved):

src/Raven.Server/NotificationCenter/Handlers/ClusterDashboardHandler.cs
```diff
@@ -75,34 +75,101 @@ public async Task Get()
...
                 {
-                    if (Logger.IsInfoEnabled)
-                        Logger.Info("Error encountered in cluster dashboard handler", ex);
+                    await HandleException(ex, webSocket);
+                }
+            }
+        }
 
-                    try
+        [RavenAction("/cluster-dashboard/remote/watch", "GET", AuthorizationStatus.ClusterAdmin, SkipUsagesCount = true)]
```
ayende: Need `admin` in the link, no?

ayende: What happens if we aren't secured?

ppekrol: IIRC everyone is cluster admin in unsecured?

arekpalinski: 1. Added `admin` in https://github.com/ravendb/ravendb/pull/12106/commits/36c23bea0dd0abcccd10625f9588b51b8260a40d
2. Correct. Then we're ClusterAdmin by default

PR #12118, opened 2021-05-07
RavenDB-16635 Making sure that a newly created transaction won't get pager states of already disposed scratches.

arekpalinski:
We must not allow to run the scratch buffer pool cleanup and create transactions at the same time.

See https://issues.hibernatingrhinos.com/issue/RavenDB-16635 for detailed explanation.

The issue has been revealed by changes made in PR https://github.com/ravendb/ravendb/pull/12008

### DIFF
--- a/src/Voron/Impl/LowLevelTransaction.cs
+++ b/src/Voron/Impl/LowLevelTransaction.cs
@@ -187,12 +187,7 @@ namespace Voron.Impl
 
             try
             {
-                foreach (var scratchOrDataPagerState in previous._pagerStates)
-                {
-                    var pagerState = scratchOrDataPagerState;
-
-                    EnsurePagerStateReference(ref pagerState);
-                }
+                CopyPagerStatesFromPreviousTx(previous);
 
                 _pageLocator = transactionPersistentContext.AllocatePageLocator(this);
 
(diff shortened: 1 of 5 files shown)


src/Voron/Impl/LowLevelTransaction.cs
```diff
@@ -187,12 +187,7 @@ public LowLevelTransaction(LowLevelTransaction previous, TransactionPersistentCo
 
             try
             {
-                foreach (var scratchOrDataPagerState in previous._pagerStates)
-                {
-                    var pagerState = scratchOrDataPagerState;
-
-                    EnsurePagerStateReference(ref pagerState);
-                }
+                CopyPagerStatesFromPreviousTx(previous);
```
arekpalinski, line 190: Here we're in the ctor used to clone reading transaction. I'm pretty sure we want to have the same logic like we have in the ctor used by transaction merging feature. Then we'll also avoid dragging pager states of non active scratches. 

Hence moved the logic to single method and used in both places.

src/Voron/Impl/Scratch/ScratchBufferPool.cs
```diff
@@ -503,18 +504,29 @@ public void Cleanup()
...
+
+                // additionally we must not allow any transaction (even read one) to start because it uses GetPagerStatesOfAllScratches() which
+                // returns _pagerStatesAllScratchesCache  we're updating here
+
+                IDisposable exitPreventNewTransactions = null;
+
                 try
                 {
-                    using (_env.WriteTransaction(context: byteStringContext))
+                    if (_env.TryPreventNewReadTransactions(TimeSpan.FromSeconds(30), out exitPreventNewTransactions))
```
ayende: Doesn't this now expose us to a _write_ transaction doing this?

arekpalinski: No. The method is actually misleading - it prevents from creating _any_ new transaction. I have renamed it and force pushed changes

ppekrol (approved):

ayende (approved):

PR #12121, opened 2021-05-08
RavenDB-16614 Atomic read write in cluster wide transaction without c…

ayende:
…onflicts.

Not done yet - allowing the user to change from the default.
Properly testing deletes.


src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
```diff
@@ -190,15 +200,73 @@ public List<string> ExecuteCompareExchangeCommands(ClusterOperationContext conte
...
+                            }
+
+                            ClusterCommands.Add(new ClusterTransactionDataCommand
+                            {
+                                Type = CommandType.CompareExchangePUT,
+                                Id = atomicGuardKey,
+                                Index = putIndex,
+                                Document = context.ReadObject(new DynamicJsonValue {["Id"] = docId}, "cmp-xchg-content")
+                            });
+                            if (metadata != null) // just to be on the safe side, should never happen
```
karmeli87: why this should never happen ?

karmeli87: What would happened if the `metadata` is `null`, should we create a new `metadata`?

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -66,6 +66,7 @@ public async Task BulkDocs()
                     BatchTrafficWatch(command.ParsedCommands);
                 }
 
+                var disableAtomicDocumentWrites = GetBoolValueQueryString("disableAtomicDocumentWrites", required: false) ?? false;
```
karmeli87: How does the user set the `disableAtomicDocumentWrites` option?
to maintain compatibility, maybe we need to allow setting this globally on the `DocumentStore`?

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
```diff
@@ -190,15 +200,73 @@ public List<string> ExecuteCompareExchangeCommands(ClusterOperationContext conte
...
+
+                            break;
+                        case nameof(CommandType.DELETE):
+                            if (document.TryGetWithoutThrowingOnError(Constants.Documents.Metadata.ClusterTransactionIndex,
+                                out long deleteIndex))
+                            {
+                                // this can happen if the user tried to delete without first loading the document
+                                // we can safely assume that such a blind write doesn't care about the state of the document
+                                // and just proceed normally
+                                deleteIndex = -1;
```
karmeli87: This is actually a problem.. I think.

If I delete and modify concurrently I might get successful response for both, while the document might or might not exist and we back to the original issue.

I think it would be better to enforce loading the document first (check on the client side as well?).

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
ayende: If this happens, that means that the user submitted a document without metadata.
The only way that can happen is if they crafted a special request. I'm including this check here to avoid NRE in that case, but this isn't something likely.

src/Raven.Server/Documents/Handlers/BatchHandler.cs
ayende: Currently, they don't, haven't gotten to it

ayende: And yes, need to enable that per session & globally.

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
ayende: If you call `Delete(string id)`, that is a good indication that you don't care what the current value is.

ppekrol:
Rebased

karmeli87 (commented):
LGTM, but what will happened if...

1. mixing cluster-tx with normal I think it shouldn't be an issue.. IIRC we always prefer the document of the cluster-tx
2. conflicts or updates because of an external replication... we are replicating the document but not the compare exchange values.. which might be a problem?

ayende:
The document contains a `@cluster-tx-id` field, which is set to the index of the compare exchange value.
In such a case, we'll have a problem, yes. I'm not sure what would be a good way to resolve it.

If the field is now out of sync of the compare exchange, what should we do?

For that matter, we probably need to strip this value on external replication, because that is a per cluster setting.

We can try to do a repair on the cluster transaction value, I guess? But I'm not sure how viable that would be.

karmeli87:
Maybe we can add the TopologyId to the compare-exchange name and to the metadata to make it some sort of multi-cluster change vector.
`@multi-cluster-change-vector`:`<topology1>:<raft-index>, <topology2>:<raft-index>`

One issue that I can think with this approach is in import to new database, I think.

ayende:
That wouldn't help the scenario of misaligned updates, thoughs.

ayende:
Consider this scenario:

* I load a document with `cluster-tx-id` of 7 in tx A.
* Modify that in a cluster tx B.
* The compare exchange value is now 8.
* I call `SaveChanges()` on tx A, overwriting the old value and returning the *`cluster-tx-id` to be. 
* All cluster wide operations will fail now on that document.

ryanheath:
Nice to see this is coming to RavenDB, making cluster wide transactions a whole lot easier to work with!
Mad props @ayende !

test/SlowTests/Issues/RavenDB-16614.cs
```diff
@@ -0,0 +1,179 @@
...
+            }
+
+            using (var session = store.OpenAsyncSession(new SessionOptions { TransactionMode = TransactionMode.ClusterWide }))
+            {
+                var user = await session.LoadAsync<User>("users/arava");
+                user.Name = "Arava Eini";
+                var user2 = await session.LoadAsync<User>("users/phoebe");
+                user2.Name = "Phoebe Eini";
+
+                using (var conflictedSession = store.OpenAsyncSession(new SessionOptions {TransactionMode = TransactionMode.ClusterWide}))
```
ryanheath, line 286: How can we make sure that this session is hitting another node? Or is this done implicit by the test runner?
If implicit, how can we make sure the test also succeeds with hitting one node each time?

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
```diff
@@ -190,15 +200,84 @@ public List<string> ExecuteCompareExchangeCommands(ClusterOperationContext conte
...
+
         public unsafe void SaveCommandsBatch(ClusterOperationContext context, long index)
         {
             if (HasDocumentsInTransaction == false)
                 return;
 
             var items = context.Transaction.InnerTransaction.OpenTable(ClusterStateMachine.TransactionCommandsSchema, ClusterStateMachine.TransactionCommands);
             var commandsCountPerDatabase = context.Transaction.InnerTransaction.ReadTree(ClusterStateMachine.TransactionCommandsCountPerDatabase);
-            var commands = SerializedDatabaseCommands.Clone(context);
+            var commands = context.ReadObject(SerializedDatabaseCommands, "serialized-tx-commands");
```
ppekrol, line 279: Is this change needed?

ayende: Yes, we are mutating the `SerializedDatabaseCommands` to add artifical commands here.

test/SlowTests/Issues/RavenDB-16614.cs
ayende: We aren't trying to do that. Cluster Wide transactions are going to go to the leader anyway, so it doesn't matter from our perspective.

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -58,6 +58,8 @@ public abstract partial class InMemoryDocumentSessionOperations : IDisposable
 
         private BatchOptions _saveChangesOptions;
 
+        public bool DisableAtomicDocumentWritesInClusterWideTransaction;
```
ppekrol: internal readonly?

src/Raven.Server/Documents/TcpHandlers/TcpConnectionOptions.cs
```diff
@@ -114,9 +114,10 @@ public void Dispose()
         }
 
 #if !RELEASE
+        private string ss = Environment.StackTrace;
```
ppekrol: Can we remove this?

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
```diff
@@ -69,6 +70,7 @@ public class ClusterTransactionOptions : IDynamicJson
             public TimeSpan? WaitForIndexesTimeout;
             public bool WaitForIndexThrow;
             public List<string> SpecifiedIndexesQueryString;
+            public bool DisableAtomicDocumentWrites;
```
ppekrol, line 79: This will be false for legacy commands? Is this OK?

ppekrol (commented):
@ayende can you check failing tests?

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
ayende: No, this is so the user can disable that for a single session.

src/Raven.Server/Documents/TcpHandlers/TcpConnectionOptions.cs
ayende: Yes, I added this to see if it would fail with additional information.

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
ayende: I don't think that this is a scenario that we need to worry about.

ayende:
Here is what the document will look like:

```
{
    "Name": "arava",
    "@metadata": {
        "@collection": "Users",
        "Raven-Clr-Type": "SlowTests.Issues.RavenDB_16614+User, SlowTests",
        "@flags": "FromClusterTransaction",
        "@cluster-transaction-index": {
            "Au3CeLCT8kWAU1UB4QcDuQ": 11
        }
    }
}
```

src/Raven.Client/Documents/Commands/Batches/BatchCommand.cs
```diff
@@ -15,10 +15,21 @@ namespace Raven.Client.Documents.Commands.Batches
...
         {
+            DisableAtomicDocumentWrites = disableAtomicDocumentsWrites;
+        }
+
+        protected override void AppendOptions(StringBuilder sb)
+        {
+            base.AppendOptions(sb);
+            if (DisableAtomicDocumentWrites)
+            {
+                sb.Append("?disableAtomicDocumentWrites=true");
```
garayx: should be `&disableAtomicDocumentWrites=true` if ` _options` are not null?

ayende: I don't follow?

ppekrol: At the end if null then we should not send anything (and then use what s configured in the database), if not null then send whatever is configured because client will override it.

ayende:
Adding the newest commits just to have a place for them. I'll do a clean PR tomoor.w

ayende:
Closing in favor of a cleanup PR

PR #12140, opened 2021-05-12
RavenDB-16681 OLAP stats colors

Danielle9897:


ml054 (changes requested):
load/file/* is missing

ml054 (approved):

PR #12145, opened 2021-05-13
RavenDB-16620: prevent possible double dispose of return context

garayx:


src/Raven.Client/Documents/Smuggler/DatabaseSmuggler.cs
```diff
@@ -269,7 +270,12 @@ private async Task<Operation> ImportInternalAsync(DatabaseSmugglerImportOptions
             }
             catch (Exception e)
             {
-                returnContext?.Dispose();
+                if (requestTask == null)
+                {
+                    // handle the possible double dispose of return context
+                    returnContext?.Dispose();
```
ayende, line 289: Double dispose should always be safe, no?

karmeli87: double dispose of the context is safe, but not the dispose of the `ReturnRequestContext`, even if the dispose is not concurrent (it isn't thread safe) you might hit different dispose path if the low mem event going on and off

ayende: There should be no case that the dispose here is concurrent. 
And the `ReturnRequestContext` should be made for repeated calls.

ppekrol (approved):

src/Raven.Client/Documents/Smuggler/DatabaseSmuggler.cs
```diff
@@ -127,6 +133,13 @@ private async Task<Operation> ExportAsync(DatabaseSmugglerExportOptions options,
                     getOperationIdCommand.NodeTag,
                     additionalTask);
             }
+            catch (Exception e)
+            {
+                if (requestTask == null)
+                    returnContext?.Dispose();
+
+                throw e.ExtractSingleInnerException();
```
ayende, line 141: Why are you throwing it like this? We are losing the stack trace in this manner.

ayende, line 141:
```suggestion
                throw;
```

ppekrol: It is the same in every place, probably to avoid throwing AggregateException

PR #12153, opened 2021-05-13
RavenDB-16661 added /databases/*/admin/studio-tasks/folder-path-options endpoint

ppekrol:


ppekrol:
Requires Studio work. cc @ml054

ml054:
@ppekrol  studio part ready.

PR #12157, opened 2021-05-13
RavenDB-16688 Fix some of SonarQube's "Duplicate casts should not be …

marodev:
Hello,

This pull request is _very simple_ and aims to reduce the technical debt. It specifically targets the "Duplicate casts should not be made" [rule](https://rules.sonarsource.com/csharp/RSPEC-3247) in SonarQube.


ppekrol:
Thanks!

PR #12159, opened 2021-05-14
RavenDB-16702 / RavenDB-16635 Make sure that'we trhowing ObjectDisposedException instead of NRE when trying to acquire page pointer on already disposed pager

arekpalinski:
https://issues.hibernatingrhinos.com/issue/RavenDB-16702

### DIFF
--- a/src/Voron/Impl/Paging/RvnMemoryMapPager.cs
+++ b/src/Voron/Impl/Paging/RvnMemoryMapPager.cs
@@ -97,7 +97,7 @@ namespace Voron.Impl.Paging
         {
             var state = pagerState ?? _pagerState;
 
-            if (CanPrefetch.Value)
+            if (CanPrefetch.Value && _pagerState != null)
             {
                 if (_pagerState.ShouldPrefetchSegment(pageNumber, out var virtualAddress, out var bytes))
                     rvn_prefetch_virtual_memory(virtualAddress, bytes, out _); // ignore if unsuccessful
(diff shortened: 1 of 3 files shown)


src/Voron/Impl/Paging/RvnMemoryMapPager.cs
```diff
@@ -97,7 +97,7 @@ protected override bool CanPrefetchQuery()
         {
             var state = pagerState ?? _pagerState;
 
-            if (CanPrefetch.Value)
+            if (CanPrefetch.Value && _pagerState != null)
```
ayende, line 100: Shouldn't we use `state`, instead of `_pageState` here? The value may be different, no?

src/Voron/Platform/Win32/WindowsMemoryMapPager.cs
```diff
@@ -141,7 +141,7 @@ public struct SplitValue
             // We need to decide what pager we are going to use right now or risk inconsistencies when performing prefetches from disk.
             var state = pagerState ?? _pagerState;
 
-            if (this.CanPrefetch.Value)
+            if (this.CanPrefetch.Value && _pagerState != null)
```
ayende, line 144: Same here.

src/Voron/Impl/Paging/RvnMemoryMapPager.cs
arekpalinski: Also looked a bit weird to me but I didn't wanted to change it here. Asked @redknightlois to double check that.

arekpalinski: Federico just confirmed. I will create an issue and send a PR

arekpalinski: Fixed in https://github.com/ravendb/ravendb/pull/12160

PR #12168, opened 2021-05-17
[WIP] RavenDB-16705

ppekrol:
- [x] fixed NRe in BlittableJsonReaderObject.AssertNoModifications
- [x] fixed issue with replication a document tombstone with DeleteRevision flag
- [ ] write tests
- [ ] apply to 4.2

### DIFF
--- a/src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
+++ b/src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
@@ -17,7 +17,6 @@ using Raven.Server.Utils;
 using Sparrow;
 using Sparrow.Json.Parsing;
 using Sparrow.Logging;
-using Sparrow.Server.Utils;
 using Voron;
 
 namespace Raven.Server.Documents.Replication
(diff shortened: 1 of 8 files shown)


src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
```diff
@@ -373,12 +374,13 @@ private bool CanContinueBatch(ReplicationState state, ref long next)
         {
             if (MissingAttachmentsInLastBatch)
             {
-                state.MissingTxMarkers.Remove(state.LastTransactionMarker);
+                // we do have missing attachments but we haven't gathered yet any of the missing hashes
+                if (state.MissingAttachmentBase64Hashes == null)
```
ayende, line 378: I don't understand how that scenario can be?

test/SlowTests/Issues/EncryptedDatabaseGroup.cs
```diff
@@ -28,12 +28,14 @@ public async Task AddingNodeToEncryptedDatabaseGroupShouldThrow()
             var options = new Options
             {
                 Server = leader,
-                ReplicationFactor = 2,
+                ReplicationFactor = 1,
```
ayende, line 31: was this intentional?

ppekrol: That's for debugging, all of those fixes need to be applied to 4.2 anyway.

src/Raven.Server/Documents/Replication/ReplicationDocumentSender.cs
ppekrol: We are iterating items by etag, documents can be later on, but I do not want to break the batch before that? I think it should not be a case BTW, but prefer to stay on the safe side, I would like to not break the batch to early.

ayende: attachments will _always_ happen before their document.

PR #12174, opened 2021-05-18
RavenDB-16493 update docker tags

gregolsky:


docker/common.ps1
```diff
@@ -35,15 +35,19 @@ function GetUbuntuImageTags($repo, $version, $arch) {
     switch ($arch) {
         "x64" { 
             return @(
-                "$($repo):$($version)-ubuntu.20.04-x64",
-                "$($repo):5.2-ubuntu-latest"
+                "$($repo):latest",
+                "$($repo):latest-lts",
+                "$($repo):ubuntu-latest",
```
ppekrol, line 40: should we consider `ubuntu-latest-lts` ?

docker/publish-nanoserver.ps1
```diff
@@ -42,7 +42,7 @@ function PushImages($imageTags) {
 }
 function GetImageTags($repo, $version) {
         return @(
-            #"$($repo):windows-nanoserver-latest",
+            "$($repo):windows-nanoserver-latest",
```
ppekrol: Should we consider `windows-nanoserver-latest-lts` ?

gregolsky: No harm in adding it I guess.
BTW this reminds we need to update the docker readme to reflect these changes.

Also I thought we could shorten it to `windows-latest` and `windows-latest-lts` maybe?

docker/common.ps1
gregolsky: How about `ubuntu-lts`?

ppekrol (approved):

PR #12176, opened 2021-05-18
RavenDB-16696 OLAP ETL to Local destination: Create Sub Folders according to Partitions

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-16696


ppekrol (approved):
@aviv86 can you rebase?

PR #12181, opened 2021-05-18
RavenDB-16677  Disable dynamic node distribution for encrypted databases

Danielle9897:


src/Raven.Studio/typescript/viewmodels/resources/manageDatabaseGroup.ts
```diff
@@ -39,6 +39,7 @@ class manageDatabaseGroup extends viewModelBase {
     private sortable: any;
 
     private graph = new databaseGroupGraph();
+    backupsOnly = false;
```
ml054, line 42: ?

Danielle9897: the fix for the ongoing tasks graph legend...

ml054: ok

ml054 (approved):

PR #12193, opened 2021-05-19
RavenDB-16724: fixed - number of request incrementation. 

omerlewitz:
https://issues.hibernatingrhinos.com/issue/RavenDB-16724

### DIFF
--- a/src/Raven.Client/Documents/Session/Operations/GetRevisionOperation.cs
+++ b/src/Raven.Client/Documents/Session/Operations/GetRevisionOperation.cs
@@ -42,6 +42,7 @@ namespace Raven.Client.Documents.Session.Operations
 
         public GetRevisionsCommand CreateRequest()
         {
+            _session.IncrementRequestCount();
             return _command;
         }
 
--- a/test/SlowTests/Issues/RavenDB-16724.cs
+++ b/test/SlowTests/Issues/RavenDB-16724.cs
@@ -1,5 +1,7 @@
 using System;
+using System.IO;
 using System.Linq;
+using System.Text;
 using System.Threading.Tasks;
 using FastTests;
 using FastTests.Utils;
@@ -31,6 +33,15 @@ namespace SlowTests.Issues
                             Name = "Hibernating",
                         },
                         id);
+
+                    byte[] byteArray = Encoding.UTF8.GetBytes("Rhinos");
+                    
+                    await session.SaveChangesAsync();
+                     
+                    session.Advanced.Attachments.Store("users/Hibernating",
+                        "invoice.pdf",
+                        new MemoryStream(byteArray),
+                        "application/pdf");
                     
                     await session.SaveChangesAsync();
                 }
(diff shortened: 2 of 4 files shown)


ayende (approved):

ppekrol:
@omerlewitz Can you check failing tests?

src/Raven.Client/Documents/Session/Operations/GetRevisionOperation.cs
```diff
@@ -42,6 +42,7 @@ public GetRevisionOperation(InMemoryDocumentSessionOperations session, IEnumerab
 
         public GetRevisionsCommand CreateRequest()
         {
+            _session.IncrementRequestCount();
```
karmeli87, line 45: I wonder if we don't want to have `IncrementRequestCount` in a more common place like in the `RequestExectuor.SendRequestToServer`, it is very easy to forget this increment.
For instance, does getting attachments increase the number of requests?

karmeli87: cc @ppekrol  thoughts?

omerlewitz: What about thread-safety if we change the implementation of the incrementation to `RequestExectuor.Increment..`

ayende: `RequestExectuor` isn't scoped to a session, it is global, so you cannot do that there.
We can't modify the API as well.

omerlewitz:
> @omerlewitz Can you check failing tests?

Karmel suggested deleting the conditional statement of the `MaxNumbrOfRequests` because at `SlowTests.Cluster.ClusterTransactionTests.ClusterTransactionRequestWithRevisions` there's `while(true)` loop which breaks it.

src/Raven.Client/Documents/Session/Operations/GetRevisionOperation.cs
karmeli87: you do have the `SessionInfo`

ayende: Yes, that might work

ppekrol: That is a bigger change. Out of scope for this task. We can move it as a part of 5.3 effort.

omerlewitz: ok :)

test/SlowTests/Issues/RavenDB-16724.cs
```diff
@@ -0,0 +1,119 @@
...
+                    Assert.Equal(3, session.Advanced.NumberOfRequests);
+                }   
+            }
+        }
+        private class User
+        {
+            public string Name { get; set; }
+        }
+        
+        public class AttachmentResult
```
ppekrol: You do have those classes available in Raven.Client, no need to copy them?

PR #12197, opened 2021-05-19
RavenDB-16705

ppekrol:
- fixed NRE in BlittableJsonReaderObject.AssertNoModifications when data is null
- stripping DeleteRevision flag to avoid NRE in IncomingReplicationHandler
- fixed possible endless replication loop when we do have missing attachments but max batch size is limiting the batch


ayende (approved):

test/SlowTests/Issues/RavenDB_16705.cs
```diff
@@ -0,0 +1,99 @@
...
+                    tx.Commit();
+                }
+
+                await SetupReplicationAsync(source, destination);
+
+                Assert.NotNull(WaitForDocumentWithAttachmentToReplicate<User>(destination, documentId1, attachmentName1, 15 * 1000));
+                Assert.NotNull(WaitForDocumentWithAttachmentToReplicate<User>(destination, documentId1, attachmentName2, 15 * 1000));
+                Assert.NotNull(WaitForDocumentWithAttachmentToReplicate<User>(destination, documentId2, attachmentName1, 15 * 1000));
+                Assert.NotNull(WaitForDocumentWithAttachmentToReplicate<User>(destination, documentId2, attachmentName2, 15 * 1000));
+            }
```
karmeli87, line 88: Let's put here `EnsureNoReplicationLoop`

karmeli87 (approved):

PR #12200, opened 2021-05-20
RavenDB-16694 :  Edit OLAP configuration not working

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-16694

### DIFF
--- a/src/Raven.Client/Documents/Operations/ETL/OLAP/OlapEtlConfiguration.cs
+++ b/src/Raven.Client/Documents/Operations/ETL/OLAP/OlapEtlConfiguration.cs
@@ -60,18 +60,16 @@ namespace Raven.Client.Documents.Operations.ETL.OLAP
             if (OlapTables.Count != other.OlapTables.Count)
                 return false;
 
-            bool equalTables = true;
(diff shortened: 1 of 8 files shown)


src/Raven.Client/Documents/Operations/ETL/OLAP/OlapEtlConfiguration.cs
```diff
@@ -45,6 +45,32 @@ public override DynamicJsonValue ToJson()
...
             return json;
         }
+
+        internal bool Equals(OlapEtlConfiguration other)
+        {
+            if (other == null || RunFrequency != other.RunFrequency ||
+                CustomPartitionValue != other.CustomPartitionValue )
+                return false;
+
+            return EqualTables(other);
```
ppekrol: Can we use https://github.com/ravendb/ravendb/blob/v5.1/src/Raven.Client/Extensions/EnumerableExtension.cs#L23 ?

aviv86: done

PR #12205, opened 2021-05-20
RavenDB-16736: dispose skipped result and projection result

garayx:


src/Raven.Server/Documents/Queries/Results/QueryResultRetrieverBase.cs
```diff
@@ -250,32 +250,36 @@ public Document GetProjectionFromDocument(Document doc, Lucene.Net.Documents.Doc
 
         private Document GetProjectionFromDocumentInternal(Document doc, Lucene.Net.Documents.Document luceneDoc, Lucene.Net.Search.ScoreDoc scoreDoc, FieldsToFetch fieldsToFetch, JsonOperationContext context, IState state)
         {
-            var result = new DynamicJsonValue();
-
-            foreach (var fieldToFetch in fieldsToFetch.Fields.Values)
+            using (doc)
```
grisha-kotler, line 253: we dispose the document but are using it for caching in the projections

arekpalinski: Note that we even have a code which explicitly ignores the dispose so we can do a projection from the same document multiple times:

https://github.com/ravendb/ravendb/blob/d8d91a5ebe53350ac26970ce615fe3ba51337206/src/Raven.Server/Documents/Queries/Results/QueryResultRetrieverBase.cs#L684

I have just added it even in a one more place - https://github.com/ravendb/ravendb/pull/12210

src/Raven.Server/Documents/Queries/Results/QueryResultRetrieverBase.cs
```diff
@@ -306,41 +310,36 @@ private Document CreateNewDocument(Document doc, string key, object fieldVal)
                 case BlittableJsonReaderObject nested:
                     return new Document
                     {
-                        Id = doc.Id,
+                        Id = _context.GetLazyString(doc.Id),
                         ChangeVector = doc.ChangeVector,
                         Data = nested,
```
arekpalinski, line 315: `nested` comes from the original document which you're about to dispose so it will invalidate it as well

src/Raven.Server/Documents/Queries/Results/QueryResultRetrieverBase.cs
```diff
@@ -610,7 +609,7 @@ protected bool TryGetValue(FieldsToFetch.FieldToFetch fieldToFetch, Document doc
 
             //_loadedDocuments.Clear(); - explicitly not clearing this, we want to cache this for the duration of the query
 
-            _loadedDocuments[document.Id ?? string.Empty] = document;
+            _loadedDocuments[document.Id ?? string.Empty] = document.Clone(_context);
```
arekpalinski, line 612: We shouldn't clone it. For that purpose we have `IgnoreDispose`. See my above comment

arekpalinski (commented):
The fix should be just `result.Dispose();` when we're skipping the document because of distinct. That has been proven that it fixes the memory leak. Other changes aren't relevant and cause test failures.

arekpalinski:
I have opened another PR: https://github.com/ravendb/ravendb/pull/12212

PR #12211, opened 2021-05-21
RavenDB-16736 Fixing memory leak when running distinc query by disposing results that are skipped

arekpalinski:
https://issues.hibernatingrhinos.com/issue/RavenDB-16736


arekpalinski:
Closing this one and will open another PR for 5.1. This is because of the following changes https://github.com/ravendb/ravendb/pull/11636 and  https://github.com/ravendb/ravendb/pull/12210 were applied only to 5.1. `IgnoreDispose` added there will make sure that we won't dispose a loaded document even if it's skipped due to `distinct` usage.

arekpalinski:
New PR: https://github.com/ravendb/ravendb/pull/12212

PR #12238, opened 2021-05-23
RavenDB-11014 reduce the index triggering

karmeli87:


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -900,13 +900,25 @@ public void RollIfNeeded()
...
 
-                if (GetNumberOfDeployedNodes() != -1 && IsStaleInternal() == false)
-                    // this for the case when we removed a rolling side-by-side, we need the original node to complete the deployment
-                    // if the original has no work to do, we need to force it.
-                    _mre?.Set(ignoreThrottling: true);
+                CompleteIfSideBySideRemoved();
             }
         }
 
+        private void CompleteIfSideBySideRemoved()
```
arekpalinski: CompleteIfSideBySideRemoved -> CompleteIfRollingSideBySideRemoved ?

In order to make it explicit that this is about rolling side by side

PR #12248, opened 2021-05-25
RavenDB-11014 add tests and bug fixes

karmeli87:


src/Raven.Client/Http/RequestExecutor.cs
```diff
@@ -522,7 +522,17 @@ protected void DisposeAllFailedNodesTimers()
 
             if (topologyUpdate != null && topologyUpdate.Status == TaskStatus.RanToCompletion || _disableTopologyUpdates)
             {
-                var (nodeIndex, chosenNode) = ChooseNodeForRequest(command, sessionInfo);
+                int nodeIndex;
+                ServerNode chosenNode;
+                try
+                {
+                    (nodeIndex, chosenNode) = ChooseNodeForRequest(command, sessionInfo);
```
ppekrol: Better to update the topology in test before doing that, this is very unlikely.

src/Raven.Server/Documents/Handlers/IndexHandler.cs
```diff
@@ -378,7 +378,9 @@ await using (var writer = new AsyncBlittableJsonTextWriter(context.Documents, Re
 
                                     var state = x.State;
 
-                                    if (e.IsOutOfMemory() == false && e.IsRavenDiskFullException() == false)
+                                    if (e.IsOutOfMemory() == false && 
+                                        e.IsRavenDiskFullException() == false && 
+                                        e is OperationCanceledException == false) // we probably closed the indexing thread
```
ppekrol: Probably better to handle that ex explicitly and not raise any alerts, etc

src/Raven.Client/Http/RequestExecutor.cs
karmeli87: done

src/Raven.Server/Documents/Handlers/IndexHandler.cs
karmeli87: done

PR #12254, opened 2021-05-25
RavenDB-16705 fixed issue with StreamsTempFile

ppekrol:
- if we would not read all the streams till the end and Reset was called then when TempCryptoStream was used underneath it would cause write exceptions until the StreamsTempFile was disposed - this could cause replication to hang because there is a single instance of StreamsTempFile there

### DIFF
--- a/src/Raven.Server/Documents/StreamsTempFile.cs
+++ b/src/Raven.Server/Documents/StreamsTempFile.cs
@@ -1,5 +1,8 @@
 ﻿using System;
 using System.IO;
+using System.Runtime.CompilerServices;
+using Raven.Client.Util;
+using Raven.Server.Indexing;
 using Raven.Server.ServerWide;
 using Raven.Server.Utils;
 using Sparrow.Utils;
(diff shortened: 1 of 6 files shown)


src/Raven.Server/Documents/StreamsTempFile.cs
```diff
@@ -60,18 +72,105 @@ public void Dispose()
...
+                AssertGeneration();
+
+                _stream.Dispose();
+            }
+
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            private void AssertGeneration()
+            {
+                if (_parent.Generation != _generation)
+                    throw new NotSupportedException();
```
arekpalinski: Let's include actual generation numbers here. Also wouldn't `InvalidOperationException` suit better here?

src/Raven.Server/Documents/StreamsTempFile.cs
```diff
@@ -28,24 +32,32 @@ public Stream StartNewStream()
...
+                Generation++;
+                _reading = false;
+                _file.ResetLength();
+
+                const int _128mb = 128 * 1024 * 1024;
+                if (_file.InnerStream.Length > _128mb)
+                {
+                    _file.InnerStream.SetLength(_128mb);
+                    _file.SetLength(_128mb);
+                }
```
arekpalinski: I find this piece of code difficult to understand. Let's talk about that to see if we can simplify it somehow.

arekpalinski (approved):

PR #12255, opened 2021-05-25
RavenDB-16100 Cluster certificate -> Server certificate

Danielle9897:


src/Raven.Server/Web/Authentication/AdminCertificatesHandler.cs
```diff
@@ -752,7 +752,7 @@ public Task GetClusterCertificates()
 
             var pfx = collection.Export(X509ContentType.Pfx);
 
-            var contentDisposition = "attachment; filename=ClusterCertificatesCollection.pfx";
+            var contentDisposition = "attachment; filename=ServerCertificatesCollection.pfx";
```
ml054, line 755: @gregolsky Can it breaks something in cloud? Do you really on that name?

ml054: any - it might be risky change IMHO

gregolsky: No, we do not rely on it. I'll share more info in a PM to you.

ml054 (approved):

PR #12292, opened 2021-05-30
RavenDB-16782 - support DateTime in the format of "yyyy'-'MM'-'dd'T'HH':'mm':'ss'+'dd':'dd'"

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-16782

### DIFF
--- a/test/SlowTests/Issues/RavenDB-16782.cs
+++ b/test/SlowTests/Issues/RavenDB-16782.cs
@@ -1,0 +1,47 @@
+﻿using System.Linq;
+using System.Threading.Tasks;
+using FastTests;
+using Raven.Client.Documents;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace SlowTests.Issues
+{
+    public class RavenDB_16782 : RavenTestBase
+    {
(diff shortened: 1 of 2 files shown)


ppekrol (approved):

test/SlowTests/Issues/RavenDB-16782.cs
```diff
@@ -0,0 +1,47 @@
...
+
+                    result = await session.Query<User>().Where(x => x.DateTime == "2015-10-17T13:28:17-04:00").ToListAsync();
+                    Assert.Equal(0, result.Count);
+                }
+            }
+        }
+
+        private class User
+        {
+            public string DateTime { get; set; }
```
ryanheath, line 44: Maybe I'm missing something here, but shouldn't this be DateTime type?
And if it is supposed to be a string type, how is this test related to DateTime? All I see are string comparisons, no?

ppekrol: We are skipping here Client's internal behavior where DateTimeOffset is sent using `o` format instead of `ISO-8601` for which we are adding support in this PR.

arekpalinski (approved):

PR #12297, opened 2021-05-31
RavenDB-16795 add logging

gregolsky:
Don't know if it's the only place there, but it'll do for starters.
We should improve it when doing RavenDB-16792


ppekrol (approved):

gregolsky:
added missing `'`

PR #12301, opened 2021-06-01
RavenDB-16759  Skipping the test which made that not awaited async calls (intentional) can cause heap corruption

arekpalinski:
https://issues.hibernatingrhinos.com/issue/RavenDB-16759

### DIFF
--- a/test/SlowTests/Issues/RavenDB_14235.cs
+++ b/test/SlowTests/Issues/RavenDB_14235.cs
@@ -164,7 +164,7 @@ namespace SlowTests.Issues
             }
         }
 
-        [Fact]
+        [Fact(Skip = "Running this test makes that not awaited async calls (intentional) can cause heap corruption. See https://issues.hibernatingrhinos.com/issue/RavenDB-16759")]
         public async Task ShouldThrowOnNotAwaitedAsyncMethods()
         {
             const string name = "1";


garayx (approved):

test/SlowTests/Issues/RavenDB_14235.cs
```diff
@@ -164,7 +164,7 @@ public async Task CombiningEscapedCharactersAsCollectionName(int size)
             }
         }
 
-        [Fact]
+        [Fact(Skip = "Running this test makes that not awaited async calls (intentional) can cause heap corruption. See https://issues.hibernatingrhinos.com/issue/RavenDB-16759")]
```
ayende, line 167: Oh, this is pretty bad, yes.
The problem is that we dispose of the session (and maybe the document store) before we complete the operation.
We aren't always able to recognize that in time, and we then access the invalid memory, which result in heap corruption.

We can (and should) fix that by waiting for the task to complete (via failure?) after the failure. of the session dispose, I think.

ayende: The actual heap corruption is because of the dispose of the document store, I believe.

arekpalinski: I have created the following draft PR - https://github.com/ravendb/ravendb/pull/12313 

I'm not fully convinced that it's worth to unskip this test. I'm afraid it can somehow can cause problems in the future

PR #12313, opened 2021-06-02
RavenDB-16759 Making sure that not awaited async calls (intentional) …

arekpalinski:
…won't cause heap corruption

### DIFF
--- a/src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
+++ b/src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
@@ -1322,7 +1322,11 @@ more responsive application.
 
             var asyncTasksCounter = Interlocked.Read(ref _asyncTasksCounter);
             if (asyncTasksCounter != 0)
+            {
+                _forTestingPurposes?.OnSessionDisposeAboutToThrowDueToRunningAsyncTask?.Invoke();
+
                 throw new InvalidOperationException($"Disposing session with active async task is forbidden, please make sure that all asynchronous session methods returning Task are awaited. Number of active async tasks: {asyncTasksCounter}");
+            }
 
             _isDisposed = true;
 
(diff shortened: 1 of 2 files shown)


arekpalinski:
I'm not fully convinced that it's worth to unskip this test. I'm afraid it can somehow can cause problems in the future

src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs
```diff
@@ -1320,24 +1320,32 @@ private void Dispose(bool isDisposing)
...
-                GC.SuppressFinalize(this);
+                var asyncTasksCounter = Interlocked.Read(ref _asyncTasksCounter);
+                if (asyncTasksCounter != 0)
+                    throw new InvalidOperationException($"Disposing session with active async task is forbidden, please make sure that all asynchronous session methods returning Task are awaited. Number of active async tasks: {asyncTasksCounter}");
 
-                _releaseOperationContext?.Dispose();
             }
-            else
+            finally
             {
```
ayende: I don't understand why put that in a try / finally?

ayende: We can just let the GC handle that, no?

arekpalinski: Why do you think that GC should handle the release of the context when we're throwing "Disposing session with active async task is forbidden ..."? Is it because the context might have invalid memory and we should rather just dispose it instead of returning to the pool? (there might be still running async operation using it)

BTW We already have the same try/finally in 5.1:
https://github.com/ravendb/ravendb/blob/be27481a063f91b7e7620165834743018711ef83/src/Raven.Client/Documents/Session/InMemoryDocumentSessionOperations.cs#L1345-L1370

ayende: This code (in both versions) is dangerous. Because we _know_ we have an async operation going, and we close the context.
That can cause us data corruption. I would rather do nothing, let the GC finalizer run (in which case we _know_ that there are no active references) and then we can close the context.

arekpalinski: I agree. I will adjust this PR and make sure that we will not have this dangerous code in 5.1 and above as well.

arekpalinski: 1. Force pushed changes here
2. Created https://issues.hibernatingrhinos.com/issue/RavenDB-16833 to ensure we don't have `try / finally` in 5.1 code

PR #12315, opened 2021-06-03
RavenDB-16777: Fix creation of empty folder when database name start …

garayx:
…with / in serverwide backup


ppekrol:
@garayx Can you check failing tests?

PR #12321, opened 2021-06-06
RavenDB-16828: try to return the new leader if it changes while cluster creation in tests

garayx:


test/Tests.Infrastructure/ClusterTestBase.cs
```diff
@@ -607,12 +608,28 @@ protected async Task DisposeAndRemoveServer(RavenServer serverToDispose)
 
             // ReSharper disable once PossibleNullReferenceException
             var condition = await leader.ServerStore.WaitForState(RachisState.Leader, CancellationToken.None).WaitAsync(numberOfNodes * _electionTimeoutInMs * 5);
-            var states = string.Empty;
+            var states = "The leader has changed while waiting for cluster to become stable. All nodes status: ";
             if (condition == false)
             {
-                states = GetLastStatesFromAllServersOrderedByTime();
+                InvalidOperationException e = null;
+                if (actualLeaderIndex == null)
```
karmeli87, line 707: why do we need all this if we can simply return `await ActionWithLeader(_ => Task.CompletedTask, clusterNodes);` ?

garayx: I want to maintain the long timeout `numberOfNodes * _electionTimeoutInMs * 5`

karmeli87 (approved):

PR #12324, opened 2021-06-06
RavenDB-16503  Add max cores to cluster - value not stored in textbox…

Danielle9897:
… when toggle state is changed


src/Raven.Studio/typescript/models/database/cluster/addClusterNodeModel.ts
```diff
@@ -14,12 +14,10 @@ class addClusterNodeModel {
     });
 
     constructor() {
-        this.usaAvailableCores.subscribe(newValue => {
+        this.useAvailableCores.subscribe(newValue => {
             this.maxUtilizedCores.clearError();
             if (!newValue)
                 return;
-
-            this.maxUtilizedCores(undefined);
```
ml054, line 22: do we need to remove that? I assume that if we enter 7 as cores then disable and enable then we can persist value, but when toggle is disabled we don't send 7 to server. 

Validation also shouldn't be a problem as it is conditional

ml054 (approved):

PR #12333, opened 2021-06-08
HRINT-2001 support custom api.ravendb.net url

gregolsky:
see the original ticket for reference

### DIFF
--- a/src/Raven.Server/Commercial/ApiHttpClient.cs
+++ b/src/Raven.Server/Commercial/ApiHttpClient.cs
@@ -8,7 +8,20 @@ namespace Raven.Server.Commercial
 {
     public static class ApiHttpClient
     {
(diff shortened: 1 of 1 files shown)


src/Raven.Server/Commercial/ApiHttpClient.cs
```diff
@@ -8,7 +8,14 @@ namespace Raven.Server.Commercial
 {
     public static class ApiHttpClient
     {
-        public const string ApiRavenDbNet = "https://api.ravendb.net";
+        public static string ApiRavenDbNet 
+        {
+            get
+            {
+                var envValue = Environment.GetEnvironmentVariable("API_RAVENDB_NET_URL");
```
ppekrol: I do not like it. Better to have `RAVEN_API_MODE` with `Stable` and `Test` values and hardcode the URLs.

PR #12337, opened 2021-06-08
RavenDB-16846 Default index deployment mode

Danielle9897:


src/Raven.Studio/typescript/viewmodels/database/indexes/editIndex.ts
```diff
@@ -195,7 +195,7 @@ class editIndex extends viewModelBase {
                         });
                     return canActivateResult;
                 } else {
-                    this.editedIndex(indexDefinition.empty());
+                    this.editedIndex(indexDefinition.empty(this.defaultDeploymentMode()));
```
ml054: like we talked I think we have race here.

src/Raven.Studio/typescript/models/database/index/indexDefinition.ts
```diff
@@ -380,15 +380,15 @@ class indexDefinition {
         }
     }
     
-    static empty(): indexDefinition {
+    static empty(defaultDeploymentMode: Raven.Client.Documents.Indexes.IndexDeploymentMode): indexDefinition {
```
ml054: rename param

src/Raven.Studio/typescript/viewmodels/database/indexes/editIndex.ts
Danielle9897: removed

src/Raven.Studio/typescript/models/database/index/indexDefinition.ts
Danielle9897: removed

ml054 (approved):

PR #12357, opened 2021-06-10
HRINT-2184 Try to include the blittable json in the exception message when a deserialization error happens

arekpalinski:


src/Raven.Client/Json/Serialization/NewtonsoftJson/Internal/SessionBlittableJsonConverter.cs
```diff
@@ -71,9 +72,44 @@ public object FromBlittable(Type type, ref BlittableJsonReaderObject json, strin
...
+
+            string TryReadBlittableAsString(BlittableJsonReaderObject jsonToRead)
+            {
+                var jsString = string.Empty;
+
+                using (var memoryStream = new MemoryStream())
+                {
+                    try
+                    {
+                        jsonToRead.WriteJsonToAsync(memoryStream).AsTask().ConfigureAwait(false).GetAwaiter().GetResult();
```
ayende, line 89: why async here?

arekpalinski: There is no sync method there since we moved to async - https://github.com/ravendb/ravendb/pull/11321

PR #12376, opened 2021-06-14
RavenDB-16827 Large raft command might kill the cluster

efratshenhar:


test/SlowTests/Issues/RavenDB-16827.cs
```diff
@@ -0,0 +1,117 @@
...
+                    }, false, 500);
+
+                    Assert.False(res);
+                }
+
+
+            }
+        }
+
+        public class Lines
```
ppekrol: We do have those in Tests.Infrastructure IIRC

test/SlowTests/Issues/RavenDB-16827.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            public float Discount { get; set; }
+            public float PricePerUnit { get; set; }
+            public string Product { get; set; }
+            public string ProductName { get; set; }
+            public int Quantity { get; set; }
+        }
+
+        public class Order
```
ppekrol: We do have those in Tests.Infrastructure IIRC

ppekrol (approved):

karmeli87 (approved):

src/Raven.Server/Rachis/Follower.cs
```diff
@@ -110,7 +110,14 @@ private void FollowerSteadyState()
                     {
                         for (int i = 0; i < appendEntries.EntriesCount; i++)
                         {
-                            entries.Add(_connection.ReadRachisEntry(context));
+                            using (var cts = new CancellationTokenSource())
+                            {
+                                var task = Concurrent_SendAppendEntriesPendingToLeaderAsync(cts, _term, appendEntries.PrevLogIndex);
```
ayende: Can we do this for the entire batch of commands, rather than for each single one?

test/SlowTests/Issues/RavenDB-16827.cs
```diff
@@ -0,0 +1,117 @@
...
+                            Product = "4",
+                            ProductName = "4",
+                            Quantity = 4
+                        }
+                    }
+                };
+                for (int i = 0; i < 12; i++)
+                {
+                    order.Lines.AddRange(order.Lines);
+                }
```
ayende, line 74: How much do we end up with here? Can we simple do it as a single unit instead of this manner?

ayende (changes requested):

src/Raven.Server/Rachis/Follower.cs
efratshenhar: yes

test/SlowTests/Issues/RavenDB-16827.cs
efratshenhar: No. the problem is with a lot of objects  and not a large document

PR #12381, opened 2021-06-14
RavenDB-16768: fix failing test, don't wait for WaitForExecutionOnSpecificNode on single node

garayx:


src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
```diff
@@ -499,11 +499,15 @@ private async Task CreateInternal(BlittableJsonReaderObject bjro, SubscriptionCr
             using (ServerStore.ContextPool.AllocateOperationContext(out TransactionOperationContext serverContext))
             using (serverContext.OpenReadTransaction())
             {
-                // need to wait on the relevant remote node
-                var node = Database.SubscriptionStorage.GetResponsibleNode(serverContext, name);
-                if (node != null && node != ServerStore.NodeTag)
+                var topology = ServerStore.GetClusterTopology(serverContext);
+                if (topology.AllNodes.Count != 1)
```
ppekrol: Shouldn't we use Members here? cc @karmeli87

src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
```diff
@@ -499,11 +499,15 @@ private async Task CreateInternal(BlittableJsonReaderObject bjro, SubscriptionCr
...
                 {
-                    await WaitForExecutionOnSpecificNode(serverContext, ServerStore.GetClusterTopology(serverContext), node, subscriptionId);
+                    // need to wait on the relevant remote node
+                    var node = Database.SubscriptionStorage.GetResponsibleNode(serverContext, name);
+                    if (node != null && node != ServerStore.NodeTag)
+                    {
+                        await WaitForExecutionOnSpecificNode(serverContext, topology, node, subscriptionId);
```
ppekrol: Don't we need at least to execute `await ServerStore.Cluster.WaitForIndexNotification(index);` ?

garayx: no if its single node we waited inside PutSubscription

src/Raven.Server/Documents/Handlers/SubscriptionsHandler.cs
garayx: this check is only to know if its single node cluster

karmeli87: What will happened if you create a database in cluster of 3 nodes, but with replication factor of 1?
node A executes the command, but the database is created on node B.
Would that work as you expect?

karmeli87: okay... this is the _cluster_ topology.

karmeli87:
I don't understand why this change can fix anything?

garayx:
> I don't understand why this change can fix anything?

1st commit is optimization the actual fix for test is in 2nd commit

PR #12388, opened 2021-06-15
RavenDB-14301 Show unsupported browser warning for IE in setup wizard

Danielle9897:


src/Raven.Studio/wwwroot/App/views/shell.html
```diff
@@ -1,27 +1,5 @@
...
-        <button class="btn btn-default" data-bind="click: browserAlertContinue">Continue</button>
-        <div class="checkbox">
-            <input id="dontShowAgainBrowser" class="styled" type="checkbox" data-bind="checked: dontShowBrowserAlertAgain">
-            <label for="dontShowAgainBrowser">
-                Don't show again
-            </label>
-        </div>
-    </div>
-    <div class="backdrop show"></div>
+<div data-bind="visible: browserAlert()">
```
ml054: can we move that logic inside and compose viewmodel instead of html file?

Danielle9897: done

src/Raven.Studio/typescript/viewmodels/common/detectBrowser.ts
```diff
@@ -0,0 +1,43 @@
+import viewModelBase = require("viewmodels/viewModelBase");
+import studioSettings = require("common/settings/studioSettings");
+
+class detectBrowser extends viewModelBase {
```
ml054: do we want to extend from view model base?

src/Raven.Studio/typescript/viewmodels/common/detectBrowser.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    showBrowserAlert = ko.observable<boolean>(false);
+    dontShowBrowserAlertAgain = ko.observable<boolean>(false);
+    
+    constructor(private settingsHasInfo: boolean) {
```
ml054: can we call that: rememberSettings or persistChoice, etc?

src/Raven.Studio/typescript/viewmodels/shell.ts
```diff
@@ -148,7 +150,7 @@ class shell extends viewModelBase {
             (settings, db) => new saveStudioConfigurationCommand(settings, db).execute()
         );
         
-        this.detectBrowser();
+        this.browserAlert = new detectBrowser(true);
```
ml054, line 153: join declaration and assignment (just like in welcome.ts file)

src/Raven.Studio/typescript/viewmodels/common/detectBrowser.ts
Danielle9897: done - removed

src/Raven.Studio/typescript/viewmodels/common/detectBrowser.ts
Danielle9897: How about: `infoPersistedInSettings `

src/Raven.Studio/typescript/viewmodels/shell.ts
Danielle9897: Can't - has to be after `studioSettings.default.configureLoaders`
otherwise I get => `Global settings loader not found`

Unless we move the `configureLoaders `also outside of constructor...

ml054: ok so let's leave it

ml054 (approved):

PR #12389, opened 2021-06-15
RavenDB-16802: log database wake up caller for test

garayx:


ppekrol:
moved to 5.1: https://github.com/ravendb/ravendb/pull/12389

PR #12390, opened 2021-06-15
RavenDB-16882 - log only when we deleted the notification

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-16882

### DIFF
--- a/src/Raven.Server/NotificationCenter/NotificationsStorage.cs
+++ b/src/Raven.Server/NotificationCenter/NotificationsStorage.cs
@@ -226,22 +226,26 @@ namespace Raven.Server.NotificationCenter
 
         public bool Delete(string id, RavenTransaction existingTransaction = null)
         {
-            if (Logger.IsInfoEnabled)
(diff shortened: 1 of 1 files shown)


src/Raven.Server/NotificationCenter/NotificationsStorage.cs
```diff
@@ -226,23 +226,27 @@ private NotificationTableValue Get(string id, JsonOperationContext context, Rave
...
+                using (_contextPool.AllocateOperationContext(out TransactionOperationContext context))
+                using (var tx = context.OpenWriteTransaction())
+                {
+                    deleteResult = DeleteFromTable(tx);
+                    tx.Commit();
+                }
             }
 
+            if (deleteResult && Logger.IsInfoEnabled)
+                Logger.Info($"Deleted notification '{id}'.");
```
ayende, line 246: Shouldn't we log a bit more here? What _was_ the notification, who deleted it, etc?
Why are we recording this, anyway?

PR #12391, opened 2021-06-16
RavenDB-16447:  Wrap query with clauses dynamically

omerlewitz:
https://issues.hibernatingrhinos.com/issue/RavenDB-16447


omerlewitz:
Making some changes

src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
```diff
@@ -932,7 +932,57 @@ public void AndAlso()
...
+                tokens.AddLast(QueryOperatorToken.Or);
+            }
+            else
+            { 
+                tokens.AddFirst(OpenSubclauseToken.Create());
+                tokens.AddLast(CloseSubclauseToken.Create());
+                tokens.AddLast(QueryOperatorToken.Or);
+            }
+        }
+
```
ayende: I don't think we need this

src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
```diff
@@ -932,7 +932,57 @@ public void AndAlso()
...
+            
+            if (wrapPreviousQueryClauses == false)
+            {
+                tokens.AddLast(QueryOperatorToken.And);
+            }
+            else
+            { 
+                tokens.AddFirst(OpenSubclauseToken.Create());
+                tokens.AddLast(CloseSubclauseToken.Create());
+                tokens.AddLast(QueryOperatorToken.And);
```
ayende, line 949: Wouldn't it be better to add this code to `WrapQuery()` directly ? 
Or, just call this before we invoke `OnBeforeQuery` ?

test/FastTests/Client/Queries/QueryTests.cs
```diff
@@ -32,6 +33,237 @@ private class B
...
+                }
+                
+                using (var session = store.OpenAsyncSession())
+                {
+                    
+                    var query =  session.Advanced.AsyncDocumentQuery<Article>()
+                        .Search(article => article.Title, "foo")
+                        .Search(article => article.Description, "bar", @operator: SearchOperator.Or);
+
+                    var result =  await query.OrElse(wrapPreviousQueryClauses:true).WhereEquals(x => x.IsDeleted, false).ToListAsync();
```
ayende: This is a problem, that doesn't do what the user likely think it does

ayende (changes requested):

test/FastTests/Client/Queries/QueryTests.cs
ppekrol: Could you elaborate?

ayende: Original query is: `from Articles where search(Title, "foo") or search(Description, "bar")`

With the modification is turned into:

 `from Articles where (search(Title, "foo") or search(Description, "bar")) or IsDeleted = false`

I don't think that the user is likely to be expecting this sort of behavior. It will now return unrelated articles, as well as deleted ones.

src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
ppekrol: From binary logic point of view probably not, but wondering if there is a reason not to have it. Sometimes people organize their intent by wrapping some binary blocks into subclauses.

src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
ppekrol: It needs to be in the DocumentQuery interface

ayende: Why? We can handle that completely internally.
But I'm fine with adding that to the interface

src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
ayende: It is more about how likely this is to cause people to do the wrong thing.

ppekrol: Let's remove it then, I do not insist to have it.

src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
```diff
@@ -932,7 +932,32 @@ public void AndAlso()
 
             tokens.AddLast(QueryOperatorToken.And);
```
ppekrol: let's remove code redundancy. We can execute here `AndAlso(wrapPreviousQueryClauses: false)` right?

test/FastTests/Client/Queries/QueryTests.cs
```diff
@@ -32,6 +33,134 @@ private class B
...
+                            Description = "bar",
+                            IsDeleted = true
+                        },
+                        id2);
+                    session.SaveChanges();
+                }
+                
+                using (var session = store.OpenAsyncSession())
+                {
+                    var query = await session.Advanced.AsyncDocumentQuery<Article>()
```
ppekrol: There is no assertion in this test.

Do 2 things:
1. ToString() and assert if generated RQL query is correct
2. ToListAsync and check if returned results are correct

test/FastTests/Client/Queries/QueryTests.cs
```diff
@@ -32,6 +33,134 @@ private class B
...
+                }
+                
+                using (var session = store.OpenAsyncSession())
+                {
+                    
+                    var query = session.Advanced.AsyncDocumentQuery<Article>()
+                        .Search(article => article.Title, "foo")
+                        .Search(article => article.Description, "bar", @operator: SearchOperator.Or);
+
+                    var result = await query.AndAlso(wrapPreviousQueryClauses:true).WhereEquals(x => x.IsDeleted, false).ToListAsync();
```
ppekrol: No assertion here + please also test a case with OnBeforeQuery event

src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs
omerlewitz: indeed

ppekrol (approved):

ppekrol (approved):

PR #12400, opened 2021-06-17
RavenDB-16858 Fixing race condition betweenc ancelling the waiter in the queue and calling Release() on it.

arekpalinski:
I have changed and approach and get rid of RemoveWaiter() method. If the waiter gets cancelled then next Release() call will do the cleanup of waiters that already got cancelled.

https://issues.hibernatingrhinos.com/issue/RavenDB-16858


arekpalinski:
This is related to changes we introduced in https://github.com/ravendb/ravendb/pull/12029

PR #12408, opened 2021-06-18
RavenDB-16894 / RavenDB-16736 Fixing NRE that was intruduced in memory leak fix (https://github.com/ravendb/ravendb/pull/12212)

arekpalinski:
https://issues.hibernatingrhinos.com/issue/RavenDB-16894


arekpalinski:
Regression introduced in https://github.com/ravendb/ravendb/pull/12212

PR #12410, opened 2021-06-20
RavenDB-16900 - Double dispose in DisposeAsync in DisposeOnce

shiranshalom:


arekpalinski (approved):

karmeli87:
Do we want to have a test for that?

PR #12412, opened 2021-06-20
RavenDB-15906  Show single entry when the Client certificate used is …

Danielle9897:
…the Server certificate


src/Raven.Studio/typescript/viewmodels/manage/certificates.ts
```diff
@@ -409,9 +411,10 @@ class certificates extends viewModelBase {
                 const clientCert = mergedCertificates.find(x => _.includes(x.Thumbprints, this.clientCertificateThumbprint()));
                 const orderedCertificates = mergedCertificates.filter(x => x !== serverCert && x !== clientCert);
                 
-                if (clientCert) {
+                if (clientCert && serverCert && clientCert.Thumbprint != serverCert.Thumbprint) {
```
ml054: use !==

Danielle9897: done

ml054 (approved):

PR #12413, opened 2021-06-20
RavenDB-16769

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-16769

### DIFF
--- a/src/Raven.Server/Documents/ETL/Providers/OLAP/OlaptEtl.cs
+++ b/src/Raven.Server/Documents/ETL/Providers/OLAP/OlaptEtl.cs
@@ -144,35 +144,64 @@ namespace Raven.Server.Documents.ETL.Providers.OLAP
         protected override int LoadInternal(IEnumerable<OlapTransformedItems> records, DocumentsOperationContext context, OlapEtlStatsScope scope)
         {
             var count = 0;
-
+            var uploadedFiles = new List<UploadInfo>();
+            var localFiles = new List<string>();
             var outerScope = scope.For(EtlOperations.LoadLocal, start: false);
-
-            foreach (var transformed in records)
+            try
             {
-                outerScope.NumberOfFiles++;
-                string localPath = null;
-                try
+                foreach (var transformed in records)
                 {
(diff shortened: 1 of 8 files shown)


src/Raven.Server/Documents/ETL/Providers/OLAP/OlaptEtl.cs
```diff
@@ -144,35 +144,64 @@ protected override IEnumerator<ToOlapItem> ConvertTimeSeriesDeletedRangeEnumerat
...
 
-                    if (AnyRemoteDestinations)
-                        UploadToServer(localPath, folderName, fileName, safeFolderName, scope);
+                foreach (var uploadInfo in uploadedFiles)
+                {
+                    TryDeleteFromRemoteDestinations(uploadInfo);
                 }
-                finally
+
+                foreach (var file in localFiles)
```
karmeli87, line 186: If you have local setting you still will be deleting those?

karmeli87: Also, how the local file structure will look like after few failures here?

ppekrol: > If you have local setting you still will be deleting those?

Yes, because we will be regenerating them during next run

ppekrol: > Also, how the local file structure will look like after few failures here?

This is just a cleanup, in theory we do not need to clean anything since e.g. Athena queries should be handling duplicates anyway

karmeli87: That will mean, that if we configured _both_ S3 and local... non of them will make progress if the upload is failing.
I think a nicer approach will be to continue the ETL and generate more local files, and try to upload them later..

ppekrol: This is not how ETL works :)

PR #12424, opened 2021-06-23
RavenDB-16920 - Argument Null Exception when storing empty string

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-16920

### DIFF
--- a/src/Sparrow/Extensions/EncodingsExtensions.cs
+++ b/src/Sparrow/Extensions/EncodingsExtensions.cs
@@ -9,6 +9,9 @@ namespace Sparrow
     {
         public static unsafe int GetBytes(this UTF8Encoding encoding, ReadOnlySpan<char> chars, Span<byte> bytes)
         {
+            if (chars.IsEmpty)
+                return 0;
+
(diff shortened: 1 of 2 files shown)


src/Sparrow/Extensions/EncodingsExtensions.cs
```diff
@@ -9,6 +9,9 @@ internal static class EncodingsExtensions
     {
         public static unsafe int GetBytes(this UTF8Encoding encoding, ReadOnlySpan<char> chars, Span<byte> bytes)
         {
+            if (chars.IsEmpty)
```
ppekrol, line 12: Do we need to do the same in `GetString` ?

ppekrol (commented):
Maybe better to move this to 5.1 as well?

src/Sparrow/Extensions/EncodingsExtensions.cs
grisha-kotler: Moved PR to 5.1

grisha-kotler:
https://github.com/ravendb/ravendb/pull/12430

PR #12426, opened 2021-06-23
RavenDB-16836: Fixed

omerlewitz:
https://issues.hibernatingrhinos.com/issue/RavenDB-16836

### DIFF
--- a/src/Raven.Client/Documents/Session/DocumentSession.Load.cs
+++ b/src/Raven.Client/Documents/Session/DocumentSession.Load.cs
@@ -25,7 +25,7 @@ namespace Raven.Client.Documents.Session
         /// <inheritdoc />
         public T Load<T>(string id)
         {
-            if (id == null)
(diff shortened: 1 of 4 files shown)


src/Raven.Client/Documents/Session/DocumentSession.Load.cs
```diff
@@ -25,7 +25,7 @@ public partial class DocumentSession
         /// <inheritdoc />
         public T Load<T>(string id)
         {
-            if (id == null)
+            if (string.IsNullOrWhiteSpace(id))
```
ppekrol, line 28: I think you need to add this to `AsyncDocumentSession.LoadAsync` as well

ppekrol (approved):

src/Raven.Client/Documents/Session/DocumentSession.Load.cs
omerlewitz: of course...haste makes waste

PR #12427, opened 2021-06-23
RavenDB-16923

ppekrol:
- Elector should set thread priority to AboveNormal like Follower
- Background worker should self dispose like in Follower to avoid TCP connection leak
- Registering elector in engine disposables to avoid background worker leak which could result in TCP connection leak

### DIFF
--- a/src/Raven.Server/Rachis/Elector.cs
+++ b/src/Raven.Server/Rachis/Elector.cs
@@ -386,8 +386,6 @@ namespace Raven.Server.Rachis
 
             if (_electorLongRunningWork != null && _electorLongRunningWork.ManagedThreadId != Thread.CurrentThread.ManagedThreadId)
                 _electorLongRunningWork.Join(int.MaxValue);
-
-            _engine.InMemoryDebug.RemoveRecorderOlderThan(DateTime.UtcNow.AddMinutes(-5));
         }
     }
 }
--- a/src/Raven.Server/Rachis/Elector.cs
+++ b/src/Raven.Server/Rachis/Elector.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading;
 using Raven.Client.Http;
 using Raven.Client.ServerWide;
 using Raven.Server.Rachis.Remote;
@@ -7,8 +8,10 @@ using Raven.Server.Utils;
 
 namespace Raven.Server.Rachis
 {
-    public class Elector
+    public class Elector : IDisposable
     {
+        private const int TimeToWaitForElectorToFinishInMs = 60 * 1000;
+
         private readonly RachisConsensus _engine;
         private readonly RemoteConnection _connection;
         private PoolOfThreads.LongRunningWork _electorLongRunningWork;
@@ -20,9 +23,14 @@ namespace Raven.Server.Rachis
             _connection = connection;
         }
 
-        public void Run()
+        public void RunAndWait()
         {
-            _electorLongRunningWork = PoolOfThreads.GlobalRavenThreadPool.LongRunning(x => HandleVoteRequest(), null, $"Elector for candidate {_connection.Source}");
+            var name = $"Elector for candidate {_connection.Source}";
+
+            _electorLongRunningWork = PoolOfThreads.GlobalRavenThreadPool.LongRunning(x => HandleVoteRequest(), null, name);
+
+            if (_electorLongRunningWork.Join(TimeToWaitForElectorToFinishInMs) == false)
+                throw new InvalidOperationException($"{name} did not finish processing in {TimeToWaitForElectorToFinishInMs}ms."); // throwing will dispose the elector
         }
 
         public override string ToString()
@@ -30,11 +38,23 @@ namespace Raven.Server.Rachis
             return $"Elector {_engine.Tag} for {_connection.Source}";
         }
 
-        public void HandleVoteRequest()
+        private void HandleVoteRequest()
         {
             try
             {
-                while (true)
+                try
+                {
+                    Thread.CurrentThread.Priority = ThreadPriority.AboveNormal;
+                }
+                catch (Exception e)
+                {
+                    if (_engine.Log.IsInfoEnabled)
+                    {
+                        _engine.Log.Info("Elector was unable to set the thread priority, will continue with the same priority", e);
+                    }
+                }
+
+                while (_engine.IsDisposed == false)
                 {
                     using (_engine.ContextPool.AllocateOperationContext(out TransactionOperationContext context))
                     {
@@ -44,7 +64,7 @@ namespace Raven.Server.Rachis
                         {
                             var election = rv.IsTrialElection ? "Trial" : "Real";
                             _engine.Log.Info($"Received ({election}) 'RequestVote' from {rv.Source}: Election is {rv.ElectionResult} in term {rv.Term} while our current term is {_engine.CurrentTerm}, " +
-                                             $"Forced election is {rv.IsForcedElection}. (Sent from:{rv.SendingThread})");
+                                $"Forced election is {rv.IsForcedElection}. (Sent from:{rv.SendingThread})");
                         }
 
                         //We are getting a request to vote for our known leader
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Rachis/Elector.cs
```diff
@@ -20,21 +23,38 @@ public Elector(RachisConsensus engine, RemoteConnection connection)
...
 
-        public void Run()
+        public void RunAndWait()
         {
-            _electorLongRunningWork = PoolOfThreads.GlobalRavenThreadPool.LongRunning(x => HandleVoteRequest(), null, $"Elector for candidate {_connection.Source}");
+            var name = $"Elector for candidate {_connection.Source}";
+
+            _electorLongRunningWork = PoolOfThreads.GlobalRavenThreadPool.LongRunning(x => HandleVoteRequest(), null, name);
+
+            if (_electorLongRunningWork.Join(TimeToWaitForElectorToFinishInMs) == false)
```
ayende: You are now holding the calling thread, which is not expected.

src/Raven.Server/Rachis/RachisConsensus.cs
```diff
@@ -1172,8 +1173,8 @@ public void AcceptNewConnection(Stream stream, Action disconnect, EndPoint remot
                     switch (initialMessage.InitialMessageType)
                     {
                         case InitialMessageType.RequestVote:
-                            var elector = new Elector(this, remoteConnection);
-                            elector.Run();
+                            using (var elector = new Elector(this, remoteConnection))
+                                elector.RunAndWait();
```
ayende: This is wrong, you are not allowing us to spin a new election cycle quickly enough.
If we haven't won, we need to continue.

ayende: This is the thread that _accepts new TCP connections_, you must never block it.

ayende (changes requested):

src/Raven.Server/Rachis/RachisConsensus.cs
ppekrol: No, it is a separate one, this is why I'm blocking.

ayende: I believe that we have this stack:

* Raven.Server.ListenToNewTcpConnection - start a new task, which is in a loop
* RavenServer.DispatchTcpConnection
* RavenServer.DispatchServerWideTcpConnection
* ServerStore.ClusterAcceptNewConnection
* RachisConsensus.AcceptNewConnection
* Elector.Run

There are four of them (`ListenToNewTcpConnection` tasks), so you are blocking a very critical limited resource that you are blocking here.

ayende (changes requested):

src/Raven.Server/Rachis/RachisConsensus.cs
ppekrol: I think we are spawning a new one before dispatching https://github.com/ravendb/ravendb/blob/7e1121ef6bf2b4f5fd518145ed51745441cba013/src/Raven.Server/RavenServer.cs#L1881 ?

ayende: You are correct, I though it was a `while`, but it isn't.
What is the reasoning behind wanting to block here? 

I still don't like it (convoy on the thread pool issue)

ppekrol: Previously we were not waiting anywhere for the Elector at all. I cannot use subscribe to the current disposal mechanism because it relies on order of subscribers and Electors were never taken into account there. Separate mechanism could be added just for electors, but @karmeli87 was not a fan of this.

karmeli87: We can maybe limit one (or two electors) per source? 
e.g. Node A can have a single elector talking to B at a time.

src/Raven.Server/Rachis/Elector.cs
```diff
@@ -30,249 +34,271 @@ public override string ToString()
...
+                                {
+                                    _electionWon = true;
+                                    try
+                                    {
+                                        var follower = new Follower(_engine, negotiation.Term, _connection);
+                                        follower.AcceptConnection(negotiation);
+                                    }
+                                    catch
+                                    {
+                                        Debugger.Launch();
```
karmeli87: :)

src/Raven.Server/Rachis/Elector.cs
```diff
@@ -30,249 +34,271 @@ public override string ToString()
...
-                                Term = _engine.CurrentTerm,
-                                VoteGranted = false,
-                                Message = $"My last log term is {lastLogTerm} and higher than yours {rv.LastLogTerm}"
-                            });
-                            _connection.Dispose();
-                            return;
-                        }
+                                if (Follower.CheckIfValidLeader(_engine, _connection, out var negotiation))
+                                {
+                                    _electionWon = true;
```
karmeli87, line 126: move this below `follower.AcceptConnection`

src/Raven.Server/Rachis/Elector.cs
```diff
@@ -30,249 +34,271 @@ public override string ToString()
...
-                            _engine.Log.Info($"Received ({election}) 'RequestVote' from {rv.Source}: Election is {rv.ElectionResult} in term {rv.Term} while our current term is {_engine.CurrentTerm}, " +
-                                             $"Forced election is {rv.IsForcedElection}. (Sent from:{rv.SendingThread})");
-                        }
+                        _engine.Log.Info("Elector was unable to set the thread priority, will continue with the same priority", e);
+                    }
+                }
 
-                        //We are getting a request to vote for our known leader
-                        if (_engine.LeaderTag == rv.Source)
+                using (this)
```
karmeli87, line 53: You still need to dispose elector externally if you have this?

ppekrol: This is a background task. Same situation with Follower.

src/Raven.Server/Rachis/Elector.cs
ppekrol: Ups

src/Raven.Server/Rachis/Elector.cs
ppekrol: I cannot, because AcceptConnection is changing the state which is disposing the Elector

src/Raven.Server/Rachis/RachisConsensus.cs
ayende: I don't like that idea. Electors needs to be able to run independently, otherwise we may have them stuck and unable to progress.

src/Raven.Server/Rachis/Elector.cs
```diff
@@ -357,5 +373,21 @@ private HandleVoteResult ShouldGrantVote(TransactionOperationContext context, lo
...
+
+            if (_engine.Log.IsInfoEnabled)
+            {
+                _engine.Log.Info($"{ToString()}: Disposing");
+            }
+
+            if (_electorLongRunningWork != null && _electorLongRunningWork.ManagedThreadId != Thread.CurrentThread.ManagedThreadId)
+                _electorLongRunningWork.Join(int.MaxValue);
+
+            _engine.InMemoryDebug.RemoveRecorderOlderThan(DateTime.UtcNow.AddMinutes(-5));
```
ayende: why do we need this?

ppekrol: Copied from Follower, not needed, but it is there as well.

src/Raven.Server/RavenServer.cs
```diff
@@ -1979,7 +1979,7 @@ private void ListenToNewTcpConnection(TcpListener listener)
                         _tcpLogger.Info("Failure when processing tcp connection", e);
                     }
                 }
-            });
+            }, TaskCreationOptions.LongRunning);
```
ayende: Why have this? In most cases, we will shift to our own thread. Even for something like Elector, they are commonly relatively short. Here you are forcing new thread creation on each connection

ppekrol: Reverted. I've added that when Elector was sync.

ayende (approved):

PR #12455, opened 2021-06-29
RavenDB-16848 Highlight & Scroll to the relevant index on indexes vie…

Danielle9897:
…w after index creation + A similar fix for databases view

### DIFF
--- a/src/Raven.Studio/typescript/common/generalUtils.ts
+++ b/src/Raven.Studio/typescript/common/generalUtils.ts
@@ -693,6 +693,23 @@ class genUtils {
 
         return result;
     }
+    
+    static scrollToElement(element: Element) {
+        const containerSelector = ".js-scroll-container";
+        
+        const containerRect = document.querySelector(containerSelector).getBoundingClientRect();
+        const rectTop = containerRect.top;
+        const rectBottom = rectTop + containerRect.height;
+
+        const elementRect = element.getBoundingClientRect();
+        const elementTop = elementRect.top;
+        const elementBottom = elementTop + elementRect.height;
+
+        // Scroll vertically only if element is outside of viewport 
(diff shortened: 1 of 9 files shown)


src/Raven.Studio/typescript/common/generalUtils.ts
```diff
@@ -693,6 +693,23 @@ class genUtils {
 
         return result;
     }
+    
+    static scrollToElement(element: Element) {
+        const containerSelector = ".js-scroll-container";
```
ml054: can we move this as second parameter with default value?

src/Raven.Studio/wwwroot/App/views/database/indexes/indexes.html
```diff
@@ -183,7 +184,7 @@ <h2 class="on-base-background" data-bind="text: entityName, visible: !groupHidde
 </div>
 
 <script type="text/html" id="index-template">
-    <div class="padding padding-sm">
+    <div data-bind="attr: { class: 'padding padding-sm js-index-template ' + ($root.shouldHighlightIndex(name) ? 'blink-style-basic' : ''), id: 'id_' + name }">
```
ml054: 1. can't we avoid having: id_ in id? It isn't meaningful.
2. What if index name contains characters which are invalid for id?

src/Raven.Studio/typescript/common/generalUtils.ts
Danielle9897: done

src/Raven.Studio/wwwroot/App/views/database/indexes/indexes.html
Danielle9897: rename ` id_` to `index_`

ml054 (approved):

PR #12479, opened 2021-07-06
RavenDB-16793 Add exit confirmation popup when creating debug package

Danielle9897:


src/Raven.Studio/typescript/viewmodels/manage/infoPackage.ts
```diff
@@ -47,23 +60,58 @@ class infoPackage extends viewModelBase {
...
                     });
             });
     }
+
+    private confirmLeavingPage() {
+        const abort = "Leave and Abort";
+        const stay = "Stay on this page";
+        const abortResult = $.Deferred<confirmDialogResult>();
+
+        const confirmation = this.confirmationMessage("Abort Debug Package Creation", "Leaving this page will abort package creation.<br>How do you want to proceed?", {
```
ml054, line 87: where do we have code which aborts package creation?

Danielle9897: The cancelation token contains the `HttpContext.RequestAborted` param
Once leaving the page, `HttpContext.RequestAborted` becomes true

ml054 (approved):

PR #12484, opened 2021-07-07
RavenDB-16968 - Avoid calling DiscardVirtualMemory on Windows by default

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-16968

### DIFF
--- a/src/Raven.Server/Config/Categories/StorageConfiguration.cs
+++ b/src/Raven.Server/Config/Categories/StorageConfiguration.cs
@@ -2,11 +2,24 @@ using System.ComponentModel;
 using Raven.Server.Config.Attributes;
 using Raven.Server.Config.Settings;
 using Sparrow;
+using Sparrow.Platform;
 
 namespace Raven.Server.Config.Categories
 {
     public class StorageConfiguration : ConfigurationCategory
     {
+        public StorageConfiguration()
+        {
+            // On Windows, DiscardVirtualMemory can result in high CPU usage due to contention on the
+            // PTE entry (in Win Server 2016 and Win Server 2019), we want to avoid it by default.
+            DiscardVirtualMemory = PlatformDetails.RunningOnPosix;
+        }
+
+        [Description("You can use this setting to specify whether to disable or enable discard virtual memory. By default, on windows, it will be disabled.")]
+        [DefaultValue(DefaultValueSetInConstructor)]
+        [ConfigurationEntry("Storage.DiscardVirtualMemory", ConfigurationEntryScope.ServerWideOrPerDatabase)]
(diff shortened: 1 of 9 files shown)


src/Voron/Impl/Paging/AbstractPager.cs
```diff
@@ -26,7 +23,7 @@ namespace Voron.Impl.Paging
     public abstract unsafe class AbstractPager : IDisposable, ILowMemoryHandler
     {
         public readonly Logger Log = LoggingSource.Instance.GetLogger<AbstractPager>("AbstractPager");
-        private readonly StorageEnvironmentOptions _options;
+        public readonly StorageEnvironmentOptions _options;
```
ayende: Revert this, please

grisha-kotler: done

src/Voron/StorageEnvironmentOptions.cs
```diff
@@ -1233,7 +1233,8 @@ public int TimeToSyncAfterFlushInSec
         public bool SkipChecksumValidationOnDatabaseLoading { get; set; }
 
         public int MaxNumberOfRecyclableJournals { get; set; } = 32;
-
+        public bool DiscardVirtualMemory { get; set; }
```
ppekrol: Maybe set default to true? Some tests are creating StorageEnv directly. Should not matter much, but I think its better to have it turn on.

grisha-kotler: done

arekpalinski (commented):
Asked some questions in the issue https://issues.hibernatingrhinos.com/issue/RavenDB-16968

src/Raven.Server/Config/Categories/StorageConfiguration.cs
```diff
@@ -2,11 +2,24 @@
...
+using Sparrow.Platform;
 
 namespace Raven.Server.Config.Categories
 {
     public class StorageConfiguration : ConfigurationCategory
     {
+        public StorageConfiguration()
+        {
+            // On Windows, DiscardVirtualMemory can result in high CPU usage due to contention on the
+            // PTE entry (in Win Server 2016 and Win Server 2019), we want to avoid it by default.
```
arekpalinski, line 14: This is specific to WS2016 and WS2019, right? But we're disabling it on all Windows versions. Is it really what we want?

ayende: We saw that in those two, and we don't _need_ to have it. Better to have more consistent behavior than try to test across everything that exists.

PR #12485, opened 2021-07-08
RavenDB-16950 Added CompareValue case for two LazyNumber using AlmostEquals instead string comparison.

maciejaszyk:


src/Raven.Client/Json/BlittableOperation.cs
```diff
@@ -223,6 +224,14 @@ private static bool CompareValues(BlittableJsonReaderObject.PropertyDetails oldP
                 return @double % 1 == 0 && @long.Equals((long)@double);
             }
 
+            if (oldProp.Token == BlittableJsonToken.LazyNumber && newProp.Token == BlittableJsonToken.LazyNumber)
+            {
```
ayende, line 228: Can we check the raw string value first, to avoid the allocations?

maciejaszyk: I think it's already done by first condition in 102 line.  It's first comparing length of string and when new and old are equal run Memory.CompareInline method.

ayende: Correct

ayende (approved):

test/SlowTests/Issues/RavenDB_16950.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+        }
+
+        [Fact]
+        public void CheckIfDecimalValueNotChangesAfterComputation()
+        {
+            int changesCounter;
+            using (var store = GetDocumentStore())
+            {
+                store.Maintenance.Send(new CreateSampleDataOperation());
```
ppekrol, line 20: Let's avoid doing that if possible. Takes a while and you need only 1 entry in the test.

test/SlowTests/Issues/RavenDB_16950.cs
```diff
@@ -0,0 +1,51 @@
...
+                    Order order = session.Load<Order>("orders/823-A");
+                    order.ShipTo.Location.Latitude = order.ShipTo.Location.Latitude;
+                    var changes = session.Advanced.WhatChanged();
+                    changesCounter = changes.Count;
+                }
+            }
+            Assert.Equal(0, changesCounter);
+        }
+
+        public class Address
```
ppekrol: private

test/SlowTests/Issues/RavenDB_16950.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            public Location Location { get; set; }
+        }
+
+        public class Location
```
ppekrol: private

test/SlowTests/Issues/RavenDB_16950.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            public double Latitude { get; set; }
+            public double Longitude { get; set; }
+        }
+
+        public class Order
```
ppekrol: private

PR #12488, opened 2021-07-11
RavenDB-16949 Limit revisions cleaning in a single transaction

efratshenhar:


src/Raven.Server/Config/Categories/DatabaseConfiguration.cs
```diff
@@ -99,5 +99,10 @@ public DatabaseConfiguration(bool forceUsing32BitsPager)
         [SizeUnit(SizeUnit.Megabytes)]
         [ConfigurationEntry("Databases.PulseReadTransactionLimitInMb", ConfigurationEntryScope.ServerWideOrPerDatabase)]
         public Size PulseReadTransactionLimit { get; set; }
+
+        [Description("Max number of revision to delete upon document update after revision configuration changed ")]
+        [DefaultValue(int.MaxValue)]
```
ayende: That is pretty high, don't we want to have a more conservative value? 
Like 2048 or something like that, so we'll clean things iteratively, rather than all at once.

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -544,7 +544,8 @@ private BlittableJsonReaderObject RecreateCountersIfNeeded(DocumentsOperationCon
             long numberOfRevisionsToDelete;
             if (configuration.MinimumRevisionsToKeep.HasValue)
             {
-                numberOfRevisionsToDelete = revisionsCount - configuration.MinimumRevisionsToKeep.Value;
+                numberOfRevisionsToDelete = Math.Min(revisionsCount - configuration.MinimumRevisionsToKeep.Value,
```
karmeli87: We need to take into account also age change.

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -1224,7 +1225,7 @@ private long EnforceConfigurationFor(DocumentsOperationContext context, string i
 
                 var prevRevisionsCount = GetRevisionsCount(context, id);
                 var configuration = GetRevisionsConfiguration(collectionName.Name);
-
+                configuration.MaxRevisionsToDeleteUponDocumentUpdate = 0;
```
karmeli87: Actually, when we enforce the configuration and we have a lot to delete, we might hold the write lock for a long time.
So let's remove that line, but need to make sure that enforcing will eventually clean all unwanted revisions.

ayende (approved):

src/Raven.Client/Documents/Operations/Revisions/RevisionsCollectionConfiguration.cs
```diff
@@ -13,12 +13,15 @@ public class RevisionsCollectionConfiguration
 
         public bool PurgeOnDelete { get; set; }
 
+        public int MaxRevisionsToDeleteUponDocumentUpdate { get; set; }
```
karmeli87: let's make it a nullable so the default value will be `null` and not 0.

karmeli87 (changes requested):

karmeli87 (approved):

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -531,35 +531,42 @@ private BlittableJsonReaderObject RecreateCountersIfNeeded(DocumentsOperationCon
...
             if (configuration.MinimumRevisionsToKeep.HasValue == false &&
                 configuration.MinimumRevisionAgeToKeep.HasValue == false)
-                return;
+                return false;
 
             long numberOfRevisionsToDelete;
             if (configuration.MinimumRevisionsToKeep.HasValue)
             {
-                numberOfRevisionsToDelete = revisionsCount - configuration.MinimumRevisionsToKeep.Value;
+                numberOfRevisionsToDelete = Math.Min(revisionsCount - configuration.MinimumRevisionsToKeep.Value,
```
ayende: Can this be negative?

efratshenhar: yes

ayende: isn't that a problem?

efratshenhar: we check if it is negative a few lines later and then return false and don't delete

src/Raven.Client/Documents/Operations/Revisions/RevisionsCollectionConfiguration.cs
```diff
@@ -13,12 +13,15 @@ public class RevisionsCollectionConfiguration
 
         public bool PurgeOnDelete { get; set; }
 
+        public int? MaxRevisionsToDeleteUponDocumentUpdate { get; set; }
```
ppekrol: For consistency with other properties it should be `Maximum` not `Max`

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -531,35 +531,42 @@ private BlittableJsonReaderObject RecreateCountersIfNeeded(DocumentsOperationCon
...
 
             var deletedRevisionsCount = DeleteRevisions(context, table, prefixSlice, collectionName, numberOfRevisionsToDelete, configuration.MinimumRevisionAgeToKeep, changeVector, lastModifiedTicks);
+
             Debug.Assert(numberOfRevisionsToDelete >= deletedRevisionsCount);
             IncrementCountOfRevisions(context, prefixSlice, -deletedRevisionsCount);
+            return ((moreRevisionToDelete) && (deletedRevisionsCount == numberOfRevisionsToDelete));
         }
 
-        public void DeleteRevisionsFor(DocumentsOperationContext context, string id)
+        public void DeleteRevisionsFor(DocumentsOperationContext context, string id, int? maxRevisionsToDeleteUponDocumentUpdate = null)
```
ppekrol: long maxRevisionsToDeleteUponDocumentUpdate = long.MaxValue ?

src/Raven.Client/Documents/Operations/Revisions/RevisionsCollectionConfiguration.cs
ppekrol: long

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -1130,7 +1137,7 @@ public async Task<IOperationResult> EnforceConfiguration(Action<IOperationProgre
             var result = new EnforceConfigurationResult();
             var ids = new List<string>();
             var sw = Stopwatch.StartNew();
-
+            var hasMoreList = new HashSet<string>();
```
ppekrol: What we are doing with this?

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -531,35 +531,42 @@ private BlittableJsonReaderObject RecreateCountersIfNeeded(DocumentsOperationCon
...
-                return;
+                return false;
 
             long numberOfRevisionsToDelete;
             if (configuration.MinimumRevisionsToKeep.HasValue)
             {
-                numberOfRevisionsToDelete = revisionsCount - configuration.MinimumRevisionsToKeep.Value;
+                numberOfRevisionsToDelete = Math.Min(revisionsCount - configuration.MinimumRevisionsToKeep.Value,
+                        configuration.MaxRevisionsToDeleteUponDocumentUpdate ?? long.MaxValue);
+                moreRevisionToDelete = ((configuration.MaxRevisionsToDeleteUponDocumentUpdate.HasValue) &&
```
ppekrol: Please simplify this:
```
numberOfRevisionsToDelete = revisionsCount - configuration.MinimumRevisionsToKeep.Value;
if (numberOfRevisionsToDelete > 0 && configuration.MaxRevisionsToDeleteUponDocumentUpdate.HasValue && configuration.MaxRevisionsToDeleteUponDocumentUpdate.Value > numberOfRevisionsToDelete)
{
    numberOfRevisionsToDelete = configuration.MaxRevisionsToDeleteUponDocumentUpdate.Value;
    moreRevisionToDelete  = true;
}
```

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -531,35 +531,42 @@ private BlittableJsonReaderObject RecreateCountersIfNeeded(DocumentsOperationCon
...
                 // delete all revisions which age has passed
-                numberOfRevisionsToDelete = long.MaxValue;
+                numberOfRevisionsToDelete = configuration.MaxRevisionsToDeleteUponDocumentUpdate ?? long.MaxValue;
             }
 
             var deletedRevisionsCount = DeleteRevisions(context, table, prefixSlice, collectionName, numberOfRevisionsToDelete, configuration.MinimumRevisionAgeToKeep, changeVector, lastModifiedTicks);
+
             Debug.Assert(numberOfRevisionsToDelete >= deletedRevisionsCount);
             IncrementCountOfRevisions(context, prefixSlice, -deletedRevisionsCount);
+            return ((moreRevisionToDelete) && (deletedRevisionsCount == numberOfRevisionsToDelete));
```
ppekrol: a lot of unnecessary parenthesis

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -1204,7 +1216,7 @@ bool CanContinueBatch(List<string> idsToCheck, TimeSpan elapsed, JsonOperationCo
             MinimumRevisionsToKeep = 0
         };
 
-        private long EnforceConfigurationFor(DocumentsOperationContext context, string id)
+        private long EnforceConfigurationFor(DocumentsOperationContext context, string id, List<string> hasMoreIdsList)
```
ppekrol: This adds unnecessary complexity to a very rare case. Let's just track in the command if there is more work (bool). If there is, just fire command with the same set of Ids.

PR #12494, opened 2021-07-12
RavenDB-16982 IndexDefinition State comparison not commutative

efratshenhar:


test/SlowTests/Issues/RavenDB-15890.cs
```diff
@@ -638,6 +639,24 @@ public async Task LocalStateDisabledCanBeChangedClusterWide()
             }
         }
 
+        [Fact]
+        public async Task RavemDB_16982()
```
ml054: can test name reflect what we are actually testing?

PR #12499, opened 2021-07-12
RavenDB-16944 Add disabled property to CreateSubscriptionCommand 

poissoncorp:
https://issues.hibernatingrhinos.com/issue/RavenDB-16944

### DIFF
--- a/src/Raven.Client/Documents/Commands/CreateSubscriptionCommand.cs
+++ b/src/Raven.Client/Documents/Commands/CreateSubscriptionCommand.cs
@@ -15,29 +15,19 @@ namespace Raven.Client.Documents.Commands
     {
         private readonly SubscriptionCreationOptions _options;
         private readonly string _id;
-        private readonly bool _disabled;
 
         public CreateSubscriptionCommand(DocumentConventions conventions, SubscriptionCreationOptions options, string id = null)
         {
             _options = options ?? throw new ArgumentNullException(nameof(options));
             _id = id;
-            _disabled = false;
         }
-        public CreateSubscriptionCommand(DocumentConventions conventions, SubscriptionCreationOptions options, bool disabled, string id = null)
-        {
-            _options = options ?? throw new ArgumentNullException(nameof(options));
-            _id = id;
-            _disabled = disabled;
-        }
-       
 
         public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, ServerNode node, out string url)
         {
             url = $"{node.Url}/databases/{node.Database}/subscriptions";
-            url += "?disabled=" + Uri.EscapeDataString(_disabled.ToString());
             if (_id != null)
-                url += "&id=" + Uri.EscapeDataString(_id);
-            
+                url += "?id=" + Uri.EscapeDataString(_id);
+
             var request = new HttpRequestMessage
             {
                 Method = HttpMethod.Put,
--- a/src/Raven.Client/Documents/Commands/CreateSubscriptionCommand.cs
+++ b/src/Raven.Client/Documents/Commands/CreateSubscriptionCommand.cs
@@ -15,18 +15,29 @@ namespace Raven.Client.Documents.Commands
     {
         private readonly SubscriptionCreationOptions _options;
         private readonly string _id;
+        private readonly bool _disabled;
 
         public CreateSubscriptionCommand(DocumentConventions conventions, SubscriptionCreationOptions options, string id = null)
         {
             _options = options ?? throw new ArgumentNullException(nameof(options));
             _id = id;
+            _disabled = false;
         }
+        public CreateSubscriptionCommand(DocumentConventions conventions, SubscriptionCreationOptions options, bool disabled, string id = null)
+        {
+            _options = options ?? throw new ArgumentNullException(nameof(options));
+            _id = id;
+            _disabled = disabled;
+        }
+       
 
         public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, ServerNode node, out string url)
         {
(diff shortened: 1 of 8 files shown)


src/Raven.Client/Documents/Commands/CreateSubscriptionCommand.cs
```diff
@@ -15,18 +15,29 @@ public class CreateSubscriptionCommand : RavenCommand<CreateSubscriptionResult>,
     {
         private readonly SubscriptionCreationOptions _options;
         private readonly string _id;
+        private readonly bool _disabled;
 
         public CreateSubscriptionCommand(DocumentConventions conventions, SubscriptionCreationOptions options, string id = null)
```
ml054: can first constructor call second?

src/Raven.Client/Documents/Commands/CreateSubscriptionCommand.cs
```diff
@@ -15,18 +15,29 @@ public class CreateSubscriptionCommand : RavenCommand<CreateSubscriptionResult>,
...
+            _options = options ?? throw new ArgumentNullException(nameof(options));
+            _id = id;
+            _disabled = disabled;
+        }
+       
 
         public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, ServerNode node, out string url)
         {
             url = $"{node.Url}/databases/{node.Database}/subscriptions";
+            url += "?disabled=" + Uri.EscapeDataString(_disabled.ToString());
```
ml054: no need to escape?

test/SlowTests/Client/Subscriptions/RavenDB_16944.cs
```diff
@@ -0,0 +1,82 @@
...
+using FastTests;
+using Raven.Client.Documents.Commands;
+using Raven.Client.Documents.Subscriptions;
+using Raven.Tests.Core.Utils.Entities;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace SlowTests.Client.Subscriptions
+{
+    public class RavenDB_16944:RavenTestBase
```
ml054: spaces around : `:`

test/SlowTests/Client/Subscriptions/RavenDB_16944.cs
```diff
@@ -0,0 +1,82 @@
...
+                {
+                    var entity = new User
+                    {
+                        Name = "Ed"
+                    };
+                    session.Store(entity);
+                    session.SaveChanges();
+                }
+
+                using (store.GetRequestExecutor().ContextPool.AllocateOperationContext(out var context))
```
ml054: can we use store.subscriptions.create here? higher level api

test/SlowTests/Client/Subscriptions/RavenDB_16944.cs
```diff
@@ -0,0 +1,82 @@
...
+                    store.GetRequestExecutor()
+                        .Execute(
+                            new CreateSubscriptionCommand(store.Conventions,
+                                new SubscriptionCreationOptions() { Name = "10", Query = "From Users" }, true, "10"), context);
+                }
+
+                using (var session = store.OpenSession())
+                {
+                    var subs = store.Subscriptions.GetSubscriptions(0, 10);
+                    var item = subs.FirstOrDefault(x => x.SubscriptionName == "10" && x.Disabled);
```
ml054: maybe find by name and then assert that is disabled.

test/SlowTests/Client/Subscriptions/RavenDB_16944.cs
```diff
@@ -0,0 +1,82 @@
...
+                {
+                    var entity = new User
+                    {
+                        Name = "Ed"
+                    };
+                    session.Store(entity);
+                    session.SaveChanges();
+                }
+
+                using (store.GetRequestExecutor().ContextPool.AllocateOperationContext(out var context))
```
ml054: same here

src/Raven.Client/Documents/Commands/CreateSubscriptionCommand.cs
ayende: there should be no escaping here.

ayende: But can only add that if it is et to true, no?

src/Raven.Client/Documents/Commands/CreateSubscriptionCommand.cs
```diff
@@ -26,7 +26,8 @@ public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, Serve
         {
             url = $"{node.Url}/databases/{node.Database}/subscriptions";
             if (_id != null)
-                url += "?id=" + Uri.EscapeDataString(_id);
+                url += "&id=" + Uri.EscapeDataString(_id);
```
ppekrol: should be `?`, right?

test/SlowTests/Client/Subscriptions/RavenDB_16944.cs
```diff
@@ -0,0 +1,70 @@
...
+                    session.SaveChanges();
+                }
+
+                store.Subscriptions.Create(new SubscriptionCreationOptions() { Name = "10", Query = "From Users"}, store.Database);
+
+                using (var session = store.OpenSession())
+                {
+                    var subs = store.Subscriptions.GetSubscriptions(0, 10);
+                    var item = subs.First(x => x.SubscriptionName == "10");
+                    Assert.True(!item.Disabled);
```
ppekrol: try to avoid this notation. use `== false`

PR #12503, opened 2021-07-14
RavenDB-13114 : Sharded Hilo

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-13114


src/Raven.Client/Documents/Commands/HiLoReturnCommand.cs
```diff
@@ -10,8 +10,9 @@ public class HiLoReturnCommand : RavenCommand
         private readonly string _tag;
         private readonly long _last;
         private readonly long _end;
+        private readonly int? _shardIndex;
 
-        public HiLoReturnCommand(string tag, long last, long end)
+        public HiLoReturnCommand(string tag, long last, long end, int? shardIndex = null)
```
ayende: Create another ctor, backward compact.

src/Raven.Client/Documents/Commands/HiLoReturnCommand.cs
```diff
@@ -21,11 +22,12 @@ public HiLoReturnCommand(string tag, long last, long end)
             _tag = tag ?? throw new ArgumentNullException(nameof(tag));
             _last = last;
             _end = end;
+            _shardIndex = shardIndex;
         }
 
         public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, ServerNode node, out string url)
         {
-            url = $"{node.Url}/databases/{node.Database}/hilo/return?tag={_tag}&end={_end}&last={_last}";
+            url = $"{node.Url}/databases/{node.Database}/hilo/return?tag={_tag}&end={_end}&last={_last}&shardIndex={_shardIndex}";
```
ayende: Only if `_shardIndex` is not null

src/Raven.Client/Documents/Commands/NextHiLoCommand.cs
```diff
@@ -14,27 +16,37 @@ public class NextHiLoCommand : RavenCommand<HiLoResult>
         private readonly DateTime _lastRangeAt;
         private readonly char _identityPartsSeparator;
         private readonly long _lastRangeMax;
+        private readonly int? _shardIndex;
 
-        public NextHiLoCommand(string tag, long lastBatchSize, DateTime lastRangeAt, char identityPartsSeparator, long lastRangeMax)
+        public NextHiLoCommand(string tag, long lastBatchSize, DateTime lastRangeAt, char identityPartsSeparator, long lastRangeMax, int? shardIndex)
```
ayende: different ctor

src/Raven.Client/Documents/Commands/NextHiLoCommand.cs
```diff
@@ -14,27 +16,37 @@ public class NextHiLoCommand : RavenCommand<HiLoResult>
...
         }
 
         public override HttpRequestMessage CreateRequest(JsonOperationContext ctx, ServerNode node, out string url)
         {
-            var path = $"hilo/next?tag={Uri.EscapeDataString(_tag)}&lastBatchSize={_lastBatchSize}&lastRangeAt={_lastRangeAt:o}&identityPartsSeparator={Uri.EscapeDataString(_identityPartsSeparator.ToString())}&lastMax={_lastRangeMax}";
+            var pathBuilder = new StringBuilder();
 
-            var request = new HttpRequestMessage
+            pathBuilder.Append($"{node.Url}/databases/{node.Database}/hilo/next")
+                .Append($"?tag={Uri.EscapeDataString(_tag)}")
```
ayende: If you are using StringBuilder, don't user interpolation, write it directly to avoid creating temporaries

src/Raven.Client/Documents/Identity/AsyncHiLoIdGenerator.cs
```diff
@@ -67,15 +73,12 @@ public RangeValue(long min, long max)
...
-        private Lazy<Task> _nextRangeTask = new Lazy<Task>(() => Task.CompletedTask);
-        protected string ServerTag;
-
         /// <summary>
         /// Generates the document ID.
         /// </summary>
         /// <param name="entity">The entity.</param>
         /// <returns></returns>
-        public async Task<string> GenerateDocumentIdAsync(object entity)
+        public async Task<string> GenerateDocumentIdAsync()
```
ayende: You are changing the public API here, why?

src/Raven.Client/Documents/Identity/AsyncMultiTypeHiLoIdGenerator.cs
```diff
@@ -48,14 +48,14 @@ public async Task<string> GenerateDocumentIdAsync(object entity)
             var tag = Conventions.TransformTypeCollectionNameToDocumentIdPrefix(typeTagName);
             if (_idGeneratorsByTag.TryGetValue(tag, out var value))
             {
-                return await value.GenerateDocumentIdAsync(entity).ConfigureAwait(false);
```
ayende: revert this

src/Raven.Server/Documents/Handlers/HiLoHandler.cs
```diff
@@ -120,8 +124,10 @@ protected override long ExecuteCmd(DocumentsOperationContext context)
 
                     if (hiloDocReader == null)
                     {
+                        var newDoc = new DynamicJsonValue();
+
                         OldMax = LastRangeMax;
-                        newDoc["Max"] = OldMax + Capacity;
+                        newDoc[Max] = OldMax + Capacity;
```
ayende: `nameof` ?

src/Raven.Server/Documents/Handlers/HiLoHandler.cs
```diff
@@ -98,12 +100,14 @@ internal class MergedNextHiLoCommand : TransactionOperationsMerger.MergedTransac
             public string Prefix;
             public long OldMax;
 
+            private const string Max = "Max";
```
ayende: replace with `nameof` ?

src/Raven.Server/Documents/ShardedHandlers/ShardedHiLoHandler.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            using (ContextPool.AllocateOperationContext(out TransactionOperationContext context))
+            {
+                var tag = GetQueryStringValueAndAssertIfSingleAndNotEmpty("tag");
+                var numOfShards = ShardedContext.RequestExecutors.Length;
+                var shardIndex = ShardedContext.GetShardIndex(context, DateTime.UtcNow.Ticks.ToString());
+
+                for (int i = 0; i < numOfShards; i++)
+                {
+                    var cmd = new ShardedCommand(this, Headers.IfMatch);
```
ayende: `IfMatch`? we don't do any caching here, no?

src/Raven.Server/Documents/ShardedHandlers/ShardedHiLoHandler.cs
```diff
@@ -0,0 +1,88 @@
...
+{
+    public class ShardedHiLoHandler : ShardedRequestHandler
+    {
+        [RavenShardedAction("/databases/*/hilo/next", "GET")]
+        public async Task GetNextHiLo()
+        {
+            using (ContextPool.AllocateOperationContext(out TransactionOperationContext context))
+            {
+                var tag = GetQueryStringValueAndAssertIfSingleAndNotEmpty("tag");
+                var numOfShards = ShardedContext.RequestExecutors.Length;
```
ayende: That isn't number of shards, that is the number of _nodes_.
Note that this _can_ change.

src/Raven.Server/Documents/ShardedHandlers/ShardedHiLoHandler.cs
```diff
@@ -0,0 +1,88 @@
...
+
+                    try
+                    {
+                        await ShardedContext.RequestExecutors[shardIndex].ExecuteAsync(cmd, context);
+                    }
+                    catch (Exception)
+                    {
+                        // failed to reach shard
+                        // try the next one
+                        shardIndex = (shardIndex + 1) % numOfShards;
```
ayende: I don't think we can rely on this being stable given resharding. We need a stable approach.

src/Raven.Server/Documents/ShardedHandlers/ShardedHiLoHandler.cs
```diff
@@ -0,0 +1,88 @@
...
+
+                    return;
+                }
+            }
+        }
+
+        [RavenShardedAction("/databases/*/hilo/return", "PUT")]
+        public async Task HiLoReturn()
+        {
+            var shardIndex = GetLongQueryString("shardIndex", required: true);
```
ayende: I think a better option would be to use sequential shard ids (0..1M) and use their mapping.
In other words, we'll align that to a shard (and ping between nodes as needed), based to the shard topology.
So the shard index won't be the position here, which may change.

ayende (changes requested):

src/Raven.Client/Documents/Identity/AsyncHiLoIdGenerator.cs
aviv86: `entity` was not used
reverted this

src/Raven.Server/Documents/ShardedHandlers/ShardedHiLoHandler.cs
karmeli87: Another solution might be to move the HiLo doc to be at the system storage of the node.
So it will be node specific only and not shard-specific. All we care here is about non-conflicted ids.
If we do go this way, we need to handle backup/restore properly.

ayende: That would conflict between different sharded databases, and it doesn't match the way we do everything else.
Better to just split it to a shard like we are always doing.

karmeli87: It doesn't match the way we do everything else indeed :-)
but it will not conflict, since it will be done under write lock of the _server_.

Split to shards might be tricky, since they can move from node and from database.

ayende: That is the _point_, you select a shard id, and return to the shard id.
And the rest is just dispatching to the right node, like we do everywhere here.

test/FastTests/Client/HiLo.cs
```diff
@@ -119,7 +118,7 @@ public void Capacity_Should_Double()
                     session.SaveChanges();
 
                     for (var i = 0; i < 32; i++)
-                        hiLoKeyGenerator.GenerateDocumentIdAsync(new User()).GetAwaiter().GetResult();
+                        hiLoKeyGenerator.GenerateDocumentIdAsync(null).GetAwaiter().GetResult();
```
karmeli87: why this change?

src/Raven.Client/Documents/Identity/HiLoResult.cs
```diff
@@ -29,5 +29,12 @@ public class HiLoResult
 
         public DateTime LastRangeAt { get; set; }
 
+        public int? ShardIndex { get; set; }
```
karmeli87: Do we use this?

test/FastTests/Client/HiLo.cs
aviv86: reverted this

src/Raven.Client/Documents/Identity/HiLoResult.cs
aviv86: no we don't
removed

ayende (approved):

test/FastTests/Client/HiLo.cs
```diff
@@ -129,7 +128,7 @@ public void Capacity_Should_Double()
                     Assert.Equal(max, 96);
 
                     //we should be receiving a range of 64 now
-                    hiLoKeyGenerator.GenerateDocumentIdAsync(new User()).GetAwaiter().GetResult();
+                    hiLoKeyGenerator.GenerateDocumentIdAsync(null).GetAwaiter().GetResult();
```
karmeli87: revert this as well

src/Raven.Client/Documents/Identity/HiLoResult.cs
```diff
@@ -30,4 +29,9 @@ public class HiLoResult
         public DateTime LastRangeAt { get; set; }
 
     }
+
+    public class HiloDocument
```
ppekrol, line 33: can we make it internal?

ppekrol (approved):

PR #12504, opened 2021-07-14
RavenDB-16993 - Make CheckReferenceBecauseOfDocumentUpdate duration configurable

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-16993

### DIFF
--- a/Raven.Database/Actions/IndexActions.cs
+++ b/Raven.Database/Actions/IndexActions.cs
@@ -125,7 +125,7 @@ namespace Raven.Database.Actions
         {
             Stopwatch sp = null;
             var count = 0;
-
+            
             using (Database.DocumentLock.Lock())
             using (Database.TransactionalStorage.DisableBatchNesting())
             {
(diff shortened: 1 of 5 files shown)


ayende (approved):

danielmarbach:
@ayende @grisha-kotler Why has the single patch tx limit been introduced?

https://github.com/ravendb/ravendb/commit/8e148dda4cd5f8136d95cc88456c0fc10c83b0c9

It is not clear to be based on the issue description. Just curious what the effect of this change would be

grisha-kotler:
@danielmarbach We introduced it in order to prevent the patch operation from holding the write transaction for too long (which prevents any write operations while the patch is running).
If you have a popular referenced document that is being patched we must run `CheckReferenceBecauseOfDocumentUpdate` for it and if you have a lot of those, the patch operation is going to run for a very long time.

Did you encounter an issue with the patch operation?

PR #12506, opened 2021-07-15
RavenDB-15312 In Windows, take into account Job specific memory limits

ayende:


src/Raven.Server/Documents/Handlers/Debugging/MemoryDebugHandler.cs
```diff
@@ -318,6 +318,10 @@ private static void WriteMemoryStats(AsyncBlittableJsonTextWriter writer, JsonOp
                 [nameof(MemoryInfo.HighMemSinceStartup)] = memoryUsageRecords.High.SinceStartup.ToString(),
                 [nameof(MemoryInfo.LowMemSinceStartup)] = memoryUsageRecords.Low.SinceStartup.ToString(),
             };
+            if (memInfo.Remarks != null)
```
grisha-kotler, line 321: Should add this for Linux as well (cgroup)

ayende: Good point, yes.

src/Sparrow.Server/LowMemory/MemoryInformation.cs
```diff
@@ -524,14 +608,59 @@ private static unsafe MemoryInfoResult GetMemoryInfoWindows(Process process, boo
...
+                
+                if (limits.JobMemoryLimit != UIntPtr.Zero)
+                {
+                    maxSize = Math.Min(maxSize, (long)limits.ProcessMemoryLimit);
+                }
+
+                if (maxSize != long.MaxValue)
+                {
+                    long workingSet64 = Process.GetCurrentProcess().WorkingSet64;
+                    availableMemoryForProcessingInBytes = maxSize - workingSet64;
```
arekpalinski, line 648: Potentially this can result in negative number, can't it?

src/Sparrow.Server/LowMemory/MemoryInformation.cs
```diff
@@ -524,14 +608,59 @@ private static unsafe MemoryInfoResult GetMemoryInfoWindows(Process process, boo
...
+                    availPageFile = maxSize - workingSet64;
```
arekpalinski, line 649: Same here

src/Sparrow.Server/LowMemory/MemoryInformation.cs
arekpalinski: @ayende ?

ayende: I don't believe so, no. That means that the process memory size is larger than the limit, and that should not be possible.

ayende: I'm fine with doing `Math.Max(0,);`, though

arekpalinski: We'll be returning the limited available amount to `MemoryInfoResult` but I don't think this will guarantee that the process won't exceed that limit in some nasty case. Anyway, to be on the safe side I opened the following PR: https://github.com/ravendb/ravendb/pull/12554

PR #12510, opened 2021-07-18
RavenDB-17007 LoadDocument should handle null as it handles dynamic n…

ayende:
…ull values


ppekrol (approved):
Can we add a test for this?

PR #12513, opened 2021-07-19
Raven db 17001 Abort write transaction on concurrency failing

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-17001

### DIFF
--- a/src/Raven.Server/Documents/Handlers/AttachmentHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/AttachmentHandler.cs
@@ -237,7 +237,6 @@ namespace Raven.Server.Documents.Handlers
                     };
                     stream.Flush();
                     await Database.TxMerger.Enqueue(cmd);
-                    cmd.ExceptionDispatchInfo?.Throw();
                     result = cmd.Result;
                 }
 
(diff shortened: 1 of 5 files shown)


src/Raven.Server/Documents/TransactionOperationsMerger.cs
```diff
@@ -1079,12 +1080,21 @@ private void NotifyOnThreadPool(List<MergedTransactionCommand> cmds)
...
-                foreach (var op in pendingOps)
+                int lastIndex = pendingOps.Count - 1;
+                for (int i = 0; i < pendingOps.Count; i++)
                 {
+                    if (i == lastIndex && exception is ConcurrencyException)
+                    {
+                        pendingOps[i].Exception = exception;
+                        continue;
+                    }
+
```
ayende: I don't understand this?

karmeli87: I guess it is an optimization, to say that if we already failed on `ConcurrencyException` no need to re-run, because we assume we will fail again.

haludi: Ok, so the optimization is not worth the special treatment.
I will revert that commit

PR #12517, opened 2021-07-19
RavenDB-14634 avoid blocking record change

karmeli87:


src/Raven.Client/Documents/DocumentStore.cs
```diff
@@ -360,6 +360,7 @@ private void ListenToChangesAndUpdateTheCache(string database)
                     () => new EvictItemsFromCacheBasedOnChanges(this, database)));
             }
             GC.KeepAlive(lazy.Value); // here we force it to be evaluated
+            lazy.Value.EnsureConnected();
```
ppekrol, line 363: Can you explain why we need this?

src/Raven.Client/Documents/Operations/Replication/InternalReplication.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System;
+using Raven.Client.Documents.Replication;
+
+namespace Raven.Client.Documents.Operations.Replication
+{
+    public class InternalReplication : ReplicationNode
```
ppekrol, line 6: internal?

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -235,19 +235,30 @@ private void CreatePullReplicationAsHub(TcpConnectionOptions tcpConnectionOption
                 PullReplicationDefinitionName = initialRequest.PullReplicationDefinitionName
             };
 
+            if (_outgoing.TryAdd(outgoingReplication) == false)
+            {
+                outgoingReplication.Dispose();
```
ppekrol: better to wrap those in usings (mind the disposal order then)

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
```diff
@@ -235,19 +235,30 @@ private void CreatePullReplicationAsHub(TcpConnectionOptions tcpConnectionOption
...
-        public void RunPullReplicationAsSink(TcpConnectionOptions tcpConnectionOptions, JsonOperationContext.ManagedPinnedBuffer buffer, PullReplicationAsSink destination)
+        public void RunPullReplicationAsSink(
+            TcpConnectionOptions tcpConnectionOptions, 
+            JsonOperationContext.ManagedPinnedBuffer buffer, 
+            PullReplicationAsSink destination, 
+            OutgoingReplicationHandler source)
         {
             var newIncoming = CreateIncomingReplicationHandler(tcpConnectionOptions, buffer, destination.HubDefinitionName);
             newIncoming.Failed += RetryPullReplication;
+            _outgoing.TryRemove(source);
```
ppekrol: can you explain why?

src/Raven.Server/Documents/Replication/ResolveConflictOnReplicationConfigurationChange.cs
```diff
@@ -37,31 +37,68 @@ public ResolveConflictOnReplicationConfigurationChange(ReplicationLoader replica
...
+        {
+            await _runOnce.WaitAsync(_database.DatabaseShutdown);
+
+            return new DisposableAction(() => _runOnce.Release());
+        }
+
+        public void WaitForBackgroundResolveTask()
         {
-            UpdateScriptResolvers();
+            _runOnce.Wait(TimeSpan.FromSeconds(60));
```
ppekrol: why 60 and what if this returns false?

src/Raven.Client/Documents/Operations/Replication/InternalReplication.cs
ppekrol: where it is actually used in the Client API?

src/Sparrow/Threading/DisposeOnce.cs
```diff
@@ -38,18 +38,67 @@ public void LeaveDispose()
...
+            _action = action;
+        }
+
+        public void Dispose(T state)
+        {
+            DisposeInternal(() => _action(state));
+        }
+
+        [Obsolete("Please use the overload that accepts a parameter")]
+        public override void Dispose() => throw new NotSupportedException("Please use the overload that accepts a parameter");
```
ppekrol: You have a warning here

src/Sparrow.Server/AsyncManualResetEvent.cs
```diff
@@ -9,18 +9,20 @@ namespace Sparrow.Server
     public class AsyncManualResetEvent : IDisposable
```
ppekrol, line 9: Can you explain why this class was changed?

src/Raven.Server/ServerWide/Commands/UpdateTopologyCommand.cs
```diff
@@ -1,24 +1,29 @@
...
+﻿using System;
+using Raven.Client.ServerWide;
 using Sparrow.Json.Parsing;
 
 namespace Raven.Server.ServerWide.Commands
 {
     public class UpdateTopologyCommand : UpdateDatabaseCommand
     {
         public DatabaseTopology Topology;
+        public DateTime At;
```
ppekrol, line 10: this can be DateTime.Min for old commands?

src/Raven.Client/Documents/DocumentStore.cs
karmeli87: It relates to the commit `fixes a race in LazyLoadResultShouldBeUpToDateEvenIfAggressiveCacheIsEnabled`
We await on the task of creation the initial connection to the server before relaying on the `EvictItemsFromCacheBasedOnChanges` to increment the generation.

src/Raven.Client/Documents/Operations/Replication/InternalReplication.cs
karmeli87: I move this `as is` to a dedicated class from `src/Raven.Client/ServerWide/DatabaseTopology.cs`
We don't use this in the client at all!

src/Raven.Server/Documents/Replication/ResolveConflictOnReplicationConfigurationChange.cs
karmeli87: It is during the dispose, so I give it a grace time of 60 seconds, otherwise I close it anyway.

src/Sparrow.Server/AsyncManualResetEvent.cs
karmeli87: The reason is that disposing the `AsyncManualResetEvent` should cancel any waiting tasks and not leave them hanging

src/Sparrow/Threading/DisposeOnce.cs
karmeli87: Couldn't find a better way to hide the inherited `Dispose`.
Reverted it. This functionality is currently not needed.

src/Raven.Server/Documents/Replication/ReplicationLoader.cs
karmeli87: We are pulling and therefore incoming, upon failure 'RetryPullReplication' will put us back as an outgoing

src/Raven.Server/ServerWide/Commands/UpdateTopologyCommand.cs
karmeli87: yes, handle this case in the `ClusterObserver`

PR #12533, opened 2021-07-21
RavenDB-17010: fix Memory leak in smuggler with transform script 2

garayx:


src/Raven.Server/Smuggler/Documents/DatabaseSmuggler.cs
```diff
@@ -733,7 +733,7 @@ await foreach (DocumentItem item in _source.GetDocumentsAsync(_options.Collectio
 
                     if (_patcher != null)
                     {
-                        item.Document = _patcher.Transform(item.Document, actions.GetContextForNewDocument());
+                        item.Document = _patcher.Transform(item.Document, _source.GetContext());
```
ayende: IIRC, we explicitly wanted that to be a new context, no?

garayx: The name could be missleading but this is the context that used for destination, it's not new a new one each time.

ppekrol: I think this is incorrect. The context on which the document is built needs to be exactly the one that is used by the command in the destination, we cannot use source context here

PR #12544, opened 2021-07-22
RavenDB-17033 - Add the response size to the traffic watch

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-17033


src/Raven.Studio/typescript/viewmodels/manage/trafficWatch.ts
```diff
@@ -405,16 +405,23 @@ class trafficWatch extends viewModelBase {
                         extraClass: rowHighlightRules,
                         sortable: "number"
                 }),
+                new textColumn<Raven.Client.Documents.Changes.TrafficWatchChangeBase>(grid,
+                    x => trafficWatch.isHttpItem(x) ? x.ResponseSizeInBytes : "n/a",
```
ml054, line 409: why can't we format to bytes directly ?

ml054: w/o using transform value

grisha-kotler: I need to sort by bytes but display the formatted value (which cannot be sorted by string).
`30 Bytes` > `1 Kbytes` when sorting by string

ml054 (approved):

PR #12547, opened 2021-07-25
RavenDB-16744 Consider using POST for large queries instead of GET

Danielle9897:


src/Raven.Studio/typescript/commands/commandBase.ts
```diff
@@ -41,6 +41,23 @@ class commandBase {
         }
     }
 
+    protected post<T>(relativeUrl: string, args: any, db?: database, options?: JQueryAjaxSettings, timeToAlert: number = 9000, baseUrl?: string, resultsSelector?: (results: any, xhr: JQueryXHR) => T ): JQueryPromise<any> {
```
ml054: please remove that

Danielle9897: done

src/Raven.Studio/typescript/commands/database/query/queryCommand.ts
```diff
@@ -98,22 +108,29 @@ f();
 `;
         let parameters = eval(parametersJs);
         const rql = queryText.substring(match.index);
-        return [parameters, rql];
+        return [JSON.parse(parameters), rql];
```
ml054, line 111: why we parse here something that was just stringified? 

Can't we return object from getPayload function?

ml054 (approved):

PR #12550, opened 2021-07-25
[RavenDB-11275] Corax

redknightlois:


Corax.sln
```diff
@@ -0,0 +1,173 @@
+﻿
```
ppekrol: 2nd sln file is a bit problematic four our CI. Can we merge that to RavenDB.sln?

redknightlois: Is it back? I have killed it. Let me check

redknightlois: Forgot to push that one 🤦‍♂️

PR #12563, opened 2021-07-28
RavenDB-16277 Sharded encryption

aviv86:
https://issues.hibernatingrhinos.com/issue/RavenDB-16277


src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -136,8 +135,18 @@ public DocumentDatabase(string name, RavenConfiguration configuration, ServerSto
...
                             if (isEncrypted && MasterKey == null)
-                                throw new InvalidOperationException($"Attempt to create encrypted db {Name} without supplying the secret key");
+                            {
+                                // check if sharded 
+                                string originalName = Name;
+                                var index = ShardHelper.TryGetShardIndexAndDatabaseName(ref originalName);
+                                if (index != -1)
+                                    MasterKey = serverStore.GetSecretKey(ctx, originalName);
+
+                                if (MasterKey == null)
```
ayende, line 147: What this means is that each of the shards on the node are using the same key, and that there is no way to avoid that.
I'm fine with that (this is how we are setup on different nodes with the same key by default), but we have to make sure that we are documenting this.

PR #12567, opened 2021-07-29
RavenDB-17010

ppekrol:


src/Raven.Server/Documents/Patch/JavaScriptUtils.cs
```diff
@@ -63,7 +63,8 @@ internal JsValue GetMetadata(JsValue self, JsValue[] args)
             if (boi.Distance != null)
                 metadata.Modifications[Constants.Documents.Metadata.SpatialResult] = boi.Distance.Value.ToJson();
 
-            using (var old = metadata)
+            // we cannot dispose the metadata here because the BOI is accessing blittable directly using the .Blittable property
+            //using (var old = metadata)
```
ayende, line 67: Isn't here specifically to handle memory leak in smuggler transformers?
If not here, this should be disposed after the script has run, no?

garayx: yes this is handled by reset method

PR #12571, opened 2021-07-29
RavenDB-17044

ppekrol:


arekpalinski (approved):

src/Raven.Server/Documents/ETL/Providers/Raven/RavenEtlDocumentTransformer.cs
```diff
@@ -247,25 +247,28 @@ public override void Transform(RavenEtlItem item, EtlStatsScope stats, EtlProces
...
-                                break;
-                            
-                            if (_script.TryGetLoadCounterBehaviorFunctionFor(item.Collection, out var counterFunction))
-                            {
-                                var counterGroups = GetCounterGroupsFor(item);
-                                if (counterGroups != null)
-                                    AddCounters(item.DocumentId, counterGroups, counterFunction);    
-                            }
-                            if (_script.TryGetLoadTimeSeriesBehaviorFunctionFor(item.Collection, out var timeSeriesLoadBehaviorFunc))
+                            using (DocumentScript.Run(Context, Context, "execute", new object[] {Current.Document}))
```
arekpalinski, line 250: Note: the fix is here that instead of calling `DocumentScript.Run().Dispose()` immediately we wrap it by `using()`

PR #12573, opened 2021-07-29
RavenDB-17061 - Projection returns an empty result when the document is missing

grisha-kotler:
https://issues.hibernatingrhinos.com/issue/RavenDB-17061

### DIFF
--- a/src/Raven.Server/Documents/Queries/Results/QueryResultRetrieverBase.cs
+++ b/src/Raven.Server/Documents/Queries/Results/QueryResultRetrieverBase.cs
@@ -177,6 +177,13 @@ namespace Raven.Server.Documents.Queries.Results
 
                 if (doc == null)
                 {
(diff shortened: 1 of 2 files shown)


src/Raven.Server/Documents/Queries/Results/QueryResultRetrieverBase.cs
```diff
@@ -177,6 +177,13 @@ protected Document GetProjection(Lucene.Net.Documents.Document input, Lucene.Net
 
                 if (doc == null)
                 {
+                    if (documentLoaded)
```
ppekrol: What will happen if 1st field is projected from index and 2nd from document?

grisha-kotler: The result will be skipped

arekpalinski (approved):

PR #12576, opened 2021-07-30
RavenDB-17046 Added bugfix for paging on unbounded set.

maciejaszyk:
https://issues.hibernatingrhinos.com/issue/RavenDB-17046

### DIFF
--- a/src/Raven.Server/Documents/DocumentsStorage.cs
+++ b/src/Raven.Server/Documents/DocumentsStorage.cs
@@ -726,38 +726,54 @@ namespace Raven.Server.Documents
             return lastEtag;
         }
 
-        public IEnumerable<Document> GetDocumentsStartingWith(DocumentsOperationContext context, string idPrefix, string matches, string exclude, string startAfterId,
-            long start, long take, string collection, DocumentFields fields = DocumentFields.All)
+        public IEnumerable<Document> GetDocumentsStartingWith(DocumentsOperationContext context, string idPrefix,  string startAfterId,
+            long start, long take, string collection, Reference<long> skippedResults, DocumentFields fields = DocumentFields.All)
         {
+            int alreadyReturnedDocumentsCount = 0;
+            int lastLoadedDocumentsCount;
             var isAllDocs = collection == Constants.Documents.Collections.AllDocumentsCollection;
             var requestedDataField = fields.HasFlag(DocumentFields.Data);
             if (isAllDocs == false && requestedDataField == false)
                 fields |= DocumentFields.Data;
-
-            foreach (var doc in GetDocumentsStartingWith(context, idPrefix, matches, exclude, startAfterId, start, take, fields: fields))
+            
+            do
             {
-                if (isAllDocs)
+                lastLoadedDocumentsCount = 0;
+                foreach (var doc in GetDocumentsStartingWith(context, idPrefix, null, null, startAfterId, start, take, fields: fields))
                 {
+                    lastLoadedDocumentsCount++;
+                    if (isAllDocs)
+                    {
+                        yield return doc;
+                        continue;
+                    }
+
+                    if (IsCollectionMatch(doc, collection) == false)
+                    {
+                        skippedResults.Value++;
+                        doc.Dispose();
+                        continue;
+                    }
+
+                    if (requestedDataField == false)
+                    {
+                        doc.Data.Dispose();
+                        doc.Data = null;
+                    }
+
+                    alreadyReturnedDocumentsCount++;
                     yield return doc;
-                    continue;
+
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -727,37 +727,53 @@ public static long ReadLastEtag(Transaction tx)
...
             if (isAllDocs == false && requestedDataField == false)
                 fields |= DocumentFields.Data;
-
-            foreach (var doc in GetDocumentsStartingWith(context, idPrefix, matches, exclude, startAfterId, start, take, fields: fields))
+            
+            do
             {
-                if (isAllDocs)
+                lastLoadedDocumentsCount = 0;
+                foreach (var doc in GetDocumentsStartingWith(context, idPrefix, matches, exclude, startAfterId, start, take, fields: fields))
```
arekpalinski: This overload of `GetDocumentsStartingWith` accepts `Reference<long> skip = null` parameter and uses is under the covers. Please pass there the newly added `skippedResults`.

As I see we don't have the usage where we pass `string matches` parameter but we might have one in the future so this way we'll have this case covered properly.

test/SlowTests/Issues/RavenDB_17046.cs
```diff
@@ -0,0 +1,105 @@
...
+            string fullStatement = fromStatement + $" where startsWith(id(),'{startsWithParam}')";
+            List<T> wholeCollection;
+            List<T> wholeCollectionByPaging;
+
+            using (var store = GetDocumentStore())
+            {
+                PrepareDatabaseForTest(store);
+                using (var session = store.OpenSession())
+                {
+                    wholeCollection = session.Advanced.RawQuery<T>(fromStatement).ToList();
```
arekpalinski: As we discussed  - for assertion purposes it's better to check it against the values generated inside the test instead of relaying on the results of another query (even if it's raw query).

maciejaszyk: Done

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -771,9 +785,9 @@ static bool IsMatch(Document doc, string collection)
                 return true;
             }
         }
-
+        
         public IEnumerable<Document> GetDocumentsStartingWith(DocumentsOperationContext context, string idPrefix, string matches, string exclude, string startAfterId,
-            long start, long take, Reference<long> skip = null, DocumentFields fields = DocumentFields.All)
+            long start, long take, Reference<long> skip = null, DocumentFields fields = DocumentFields.All, Reference<long> unmatchedSkipped = null)
```
arekpalinski: Why adding additional parameter? We already have `Reference<long> skip` for that purpose, don't we?

test/SlowTests/Issues/RavenDB_17046.cs
```diff
@@ -0,0 +1,187 @@
...
+                }
+            }
+        }
+        private static int CompaniesSortById(Company a, Company b)
+        {
+            return a.Id.CompareTo(b.Id);
+        }
+    }
+
+    public class RavenDB_17046_LowLevelTests : RavenLowLevelTestBase
```
arekpalinski: Let's move it to separate file. This way it will be easier to find and navigate there once any test fails here.

test/SlowTests/Issues/RavenDB_17046.cs
```diff
@@ -0,0 +1,187 @@
...
+        
+        public void Initialize([CallerMemberName] string caller = null)
+        {
+            _disposeDatabase = CreatePersistentDocumentDatabase(NewDataPath(prefix: "DocumentsCrud"), out _documentDatabase, caller);
+        }
+
+        [Fact]
+        public void CheckIfNotMatchedItemCountIntoSkippedValues()
+        {
+            Initialize();
```
arekpalinski: Minor (no need to make any code changes) - you have only one test here. It would be simpler to just put:

`using (var db = CreatePersistentDocumentDatabase(...)))`

so you wouldn't need to override the `Dispose()`.

src/Raven.Server/Documents/DocumentsStorage.cs
maciejaszyk: Sure, but it's used in a different case (with streams and replacing start parameter) and doesn't work in our case. We need only information when it doing skip for unmatched items so I introduced new parameter.

test/SlowTests/Issues/RavenDB_17046.cs
arekpalinski: Eventually this test got removed

arekpalinski (approved):

PR #12590, opened 2021-08-02
RavenDB-16117 Database Storage Report: Combine items to Grouped Data

Danielle9897:


src/Raven.Studio/typescript/models/database/status/storageReportItem.ts
```diff
@@ -8,6 +8,7 @@ class storageReportItem {
     type: string;
     internalChildren: storageReportItem[];
     size?: number;
+    length?: number;
```
ml054, line 11: what this length represents?

src/Raven.Studio/typescript/viewmodels/database/status/storageReport.ts
```diff
@@ -18,6 +18,7 @@ type positionAndSizes = {
 class storageReport extends viewModelBase {
 
     static readonly animationLength = 200;
+    static readonly maxChildrenToShow = 100; // todo - make this 1000 ?
```
ml054: yes

src/Raven.Studio/typescript/viewmodels/database/status/storageReport.ts
```diff
@@ -172,28 +174,74 @@ class storageReport extends viewModelBase {
         };
         return preAllocatedBuffers;
     }
+    
+    private groupChildren(items: storageReportItem[]): storageReportItem[] {
+        
+        if (items.length > storageReport.maxChildrenToShow) {
+            let clonedItems = _.cloneDeep(items);
```
ml054: why you deep clone here? Can we avoid that?

src/Raven.Studio/typescript/viewmodels/database/status/storageReport.ts
```diff
@@ -172,28 +174,74 @@ class storageReport extends viewModelBase {
...
+                    lazyLoadChildren: false,
+                    type: sampleItem.type,
+                    parent: sampleItem.parent,
+                    showType: sampleItem.showType,
+                    formatSize: sampleItem.formatSize,
+                    formatPercentage: sampleItem.formatPercentage,
+                    recyclableJournal: sampleItem.recyclableJournal,
+                    customSizeProvider: sampleItem.customSizeProvider,
+                    isStorageEnvironment: sampleItem.isStorageEnvironment
+                })
```
ml054: this looks very suspicious - what do you want to achieve here?

src/Raven.Studio/typescript/models/database/status/storageReportItem.ts
```diff
@@ -19,15 +20,18 @@ class storageReportItem {
     numberOfEntries: number = null;
     lazyLoadChildren = false;
     customSizeProvider: (header: boolean) => string;
-
+    isGrouped: boolean;
+    
     recyclableJournal = false;
 
-    constructor(name: string, type: string, showType: boolean, size: number, internalChildren: storageReportItem[] = null) {
+    constructor(name: string, type: string, showType: boolean, size: number, length: number, internalChildren: storageReportItem[] = null, isGrouped: boolean = false) {
```
ml054: if we need length param but it is null 90% of use cases then don't put that as constructor param - set this using property - only when needed.

Danielle9897: done

src/Raven.Studio/typescript/models/database/status/storageReportItem.ts
Danielle9897: Length is used to calculate the length for '_stream_'
Needed for the `customSizeProvider` (inside method` mapStreams()`)

src/Raven.Studio/typescript/viewmodels/database/status/storageReport.ts
Danielle9897: changed to `_.clone()`

src/Raven.Studio/typescript/viewmodels/database/status/storageReport.ts
Danielle9897: just needed to provide an initial `storageReportItem `object for the reduce

src/Raven.Studio/typescript/models/database/status/storageReportItem.ts
ml054: Can't we use size?

src/Raven.Studio/typescript/viewmodels/database/status/storageReport.ts
ml054: Can we avoid clone at all?

src/Raven.Studio/typescript/models/database/status/storageReportItem.ts
Danielle9897: For streams, the server returns both `Length` & `AllocatedSpaceInBytes`
which is used here:
https://github.com/ravendb/ravendb/blob/7f32742ee1fe90c30b9d4242cdb3234a3b589268/src/Raven.Studio/typescript/viewmodels/database/status/storageReport.ts#L225

src/Raven.Studio/typescript/viewmodels/database/status/storageReport.ts
Danielle9897: yes - removed

ml054 (approved):

PR #12597, opened 2021-08-04
RavenDB-17074 Query results: Fix the Show Raw Output

Danielle9897:


ml054:
I wonder if we should use GET for raw query url - it is easier to copy/retry/etc?

Danielle9897:
Possibe to use  GET - but - didn't we want this: [https://github.com/ravendb/ravendb/discussions/12216](https://github.com/ravendb/ravendb/discussions/12216)
?

src/Raven.Studio/typescript/commands/database/query/queryCommand.ts
```diff
@@ -124,17 +124,34 @@ f();
         return JSON.stringify(payload);
     }
     
-    getUrl() {
+    getUrl(GETRequest: boolean = false) {
```
ml054: paramaters names in ts starts with lowercase, maybe let's do:
getUrl(method: "POST" | "GET" = "POST") ?

src/Raven.Studio/typescript/commands/database/query/queryCommand.ts
```diff
@@ -124,17 +124,34 @@ f();
...
-        });
+            metadataOnly: typeof(criteria.metadataOnly()) !== 'undefined' ? criteria.metadataOnly() : undefined
+        };
+        
+        let urlArgs = this.urlEncodeArgs(argsForPOST);
+        
+        if (GETRequest) {
+            const [parameters, rql] = this.getQueryText();
+            
+            const argsForGET = {...argsForPOST, ...{
```
ml054: no need to explode this twice:
`
{ 
  ... oldObject,
  newProp: value1
}
`

src/Raven.Studio/typescript/commands/database/query/queryCommand.ts
ml054: then we avoid having: 
queryCmd.getUrl(true)
since no one knows what this true means.

ml054 (changes requested):

src/Raven.Studio/typescript/commands/database/query/queryCommand.ts
Danielle9897: done

src/Raven.Studio/typescript/commands/database/query/queryCommand.ts
Danielle9897: done

src/Raven.Studio/typescript/commands/database/query/queryCommand.ts
```diff
@@ -138,16 +138,16 @@ f();
         
         let urlArgs = this.urlEncodeArgs(argsForPOST);
         
-        if (GETRequest) {
+        if (method === "GET") {
             const [parameters, rql] = this.getQueryText();
             
-            const argsForGET = {...argsForPOST, ...{
+            const argsForGET = {...argsForPOST,
```
ml054: please fix indent:
```
{
    ...argsForPost,
    query: rql, 
    other props
```

Danielle9897: done

ml054 (approved):

PR #12598, opened 2021-08-04
RavenDB-17072 fix patching cause `UseAfterFree detected`

karmeli87:
The root cause was that we disposed an object that was "held" by an already disposed context

- Better error when patch failed
- Make sure we scope and dispose `ScriptRunnerResult` properly
- return results should clone to external context first

### DIFF
--- a/src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
+++ b/src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
@@ -62,7 +62,8 @@ namespace Raven.Server.Documents.Patch
             if (runIfMissing != null)
                 runIfMissing.DebugMode = _debugMode;
 
-            var originalDocument = _database.DocumentsStorage.Get(context, id);
+            var originalDocument = GetCurrentDocument(context, id);
+
             if (expectedChangeVector != null)
             {
                 if (originalDocument == null)
@@ -109,9 +110,8 @@ namespace Raven.Server.Documents.Patch
                     };
                 }
 
-                object documentInstance;
+                object documentInstance = null;
                 var args = _patch.Args;
-                BlittableJsonReaderObject originalDoc = null;
                 if (originalDocument == null)
                 {
                     if (_createIfMissing == null)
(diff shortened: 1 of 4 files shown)


src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
```diff
@@ -101,17 +103,15 @@ protected PatchResult ExecuteOnDocument(DocumentsOperationContext context, strin
             }
 
             {
+
```
ppekrol: remove those

src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
```diff
@@ -101,17 +103,15 @@ protected PatchResult ExecuteOnDocument(DocumentsOperationContext context, strin
...
+
                 if (originalDocument == null && runIfMissing == null && _createIfMissing == null)
                 {
-                    return new PatchResult
-                    {
-                        Status = PatchStatus.DocumentDoesNotExist
-                    };
+                    return new PatchResult {Status = PatchStatus.DocumentDoesNotExist};
```
ppekrol: formatting

src/Raven.Server/Documents/Patch/PatchDocumentCommand.cs
```diff
@@ -149,11 +147,12 @@ protected PatchResult ExecuteOnDocument(DocumentsOperationContext context, strin
...
                     var modifiedDoc = ExecuteScript(context, id, run, patchContext, documentInstance, args);
-                    var result = new PatchResult 
+
+                    var result = new PatchResult
                     {
                         Status = PatchStatus.NotModified,
-                        OriginalDocument = _isTest == false ? null : originalDoc?.Clone(context),
-                        ModifiedDocument = modifiedDoc
+                        OriginalDocument = _isTest == false ? null : originalDocumentData?.Clone(_externalContext ?? context),
+                        ModifiedDocument = modifiedDoc?.Clone(_externalContext ?? context)
```
ppekrol: I'm worried that this might decrease the performance

src/Raven.Server/Documents/Patch/ScriptRunner.cs
```diff
@@ -1891,7 +1891,7 @@ private Client.Exceptions.Documents.Patching.JavaScriptException CreateFullError
                 else
                     msg = e.Error.ToString();
 
-                msg = "At " + e.Column + ":" + e.LineNumber + " " + msg;
+                msg = $"An Error occurred for document '{documentId}' at {e.Column}:{e.LineNumber} {msg}";
```
ppekrol: Error -> error

ppekrol: Also this is an end of exception. I would expect to see that at the beginning?

ppekrol (approved):

PR #12604, opened 2021-08-05
RavenDB-17070 Added abstraction of indexing classes.

maciejaszyk:


src/Raven.Server/Documents/Indexes/Persistence/IndexReadOperationBase.cs
```diff
@@ -0,0 +1,70 @@
...
+using Raven.Server.ServerWide.Context;
+using Raven.Server.Utils;
+using Sparrow.Json;
+using Sparrow.Logging;
+
+namespace Raven.Server.Documents.Indexes.Persistence
+{
+    public abstract class IndexReadOperationBase : IndexOperationBase
+    {
+        protected static readonly Sort SortByFieldScore = new Sort(SortField.FIELD_SCORE);
```
ppekrol: probably should go to Lucene

ppekrol (approved):

src/Raven.Server/Documents/Indexes/Persistence/IndexReadOperationBase.cs
maciejaszyk: You're right. I moved it.

PR #12607, opened 2021-08-05
RavenDB-11275 - Support Sorting

redknightlois:


src/Corax/IndexEntryBuilder.cs
```diff
@@ -141,7 +129,7 @@ public void Write(int field, ReadOnlySpan<byte> value, long longValue, double do
             int dataLocation = _dataIndex;
 
             // Write known field pointer.
-            _knownFieldsLocations[field] = dataLocation | IsTypedMask[sizeof(int)];
+            _knownFieldsLocations[field] = dataLocation | unchecked((int)0x80000000);
```
ayende, line 132: Can you define constants here, instead of hard coded values?

src/Corax/IndexSearcher.cs
```diff
@@ -187,6 +196,71 @@ public MultiTermMatch StartWithQuery(string field, string startWith)
...
+                delegate*<ReadOnlySpan<byte>, ReadOnlySpan<byte>, int> compareSequenceFunc,
+                MatchCompareFieldType entryFieldType = MatchCompareFieldType.Sequence, 
+                int take = -1)
+            where TInner : IQueryMatch     
+        {
+            // Federico: I don't even really know if we are going to find a use case for this. However, it was built for the purpose
+            //           of showing that it is possible to build any custom group of functions. Why would we want to do this instead
+            //           of just building a TComparer, I dont know. But for now the `CustomMatchComparer` can be built like this from
+            //           static functions. 
+            return Create(new SortingMatch<TInner, CustomMatchComparer>(
```
ayende, line 212: Agreed, that it not something that we'll likely want and is candidate for removal.

src/Corax/Queries/SortingMatch.Comparers.cs
```diff
@@ -0,0 +1,305 @@
...
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            internal static int CompareAscending<T>(T x, T y)
+            {
+                if (typeof(T) == typeof(long))
+                {
+                    return Math.Sign((long)(object)y - (long)(object)x);
+                }
+                else if (typeof(T) == typeof(double))
+                {
+                    return Math.Sign((double)(object)y - (double)(object)x);
```
ayende, line 49: Why not call `double.CompareTo`, etc?

src/Corax/Queries/SortingMatch.Comparers.cs
```diff
@@ -0,0 +1,305 @@
...
+                {
+                    var readerX = comparer._searcher.GetReaderFor(x);
+                    var readX = readerX.Read(comparer._fieldId, out var resultX);
+
+                    var readerY = comparer._searcher.GetReaderFor(y);
+                    var readY = readerY.Read(comparer._fieldId, out var resultY);
+
+                    if (readX && readY)
+                    {
+                        return comparer.CompareSequence(resultX, resultY);
```
ayende, line 254: shouldn't this be the other way around?

ayende, line 254:
```suggestion
                        return comparer.CompareSequence(resultY, resultX);
```

ayende: Also, why is that a dedicated comparer instead of just applying `*-1` ?

src/Corax/Queries/SortingMatch.cs
```diff
@@ -0,0 +1,325 @@
...
+            private const int CacheSize = 4096;
+
+            public HashCacheMatchComparer(IndexSearcher searcher, in T comparer)
+            {
+                _searcher = searcher;
+                _comparer = comparer;
+                _fieldId = comparer.FieldId;
+
+                var hash = ArrayPool<Item>.Shared.Rent(CacheSize);
+                for (int i = 0; i < hash.Length; i++)
```
ayende: `Array.Clear` would be faster here?

src/Corax/Queries/SortingMatch.cs
```diff
@@ -0,0 +1,325 @@
...
+                        readX = reader.Read<double>(_fieldId, out var value);
+                        ix.Value = (W)(object)new NumericalItem<double>(value);
+                    }
+                    else
+                        throw new NotSupportedException("Not supported.");
+
+                    ix.Match = (long)((ulong)x | (readX ? 0 : NoMatch));
+                }
+
+                bool readY;
```
ayende: Lots of code duplication here, can we do aggresive inline function instead?

src/Corax/Queries/SortingMatch.cs
```diff
@@ -0,0 +1,325 @@
...
+                MatchCompareFieldType.Floating => Fill<NumericalItem<double>>(matches),
+                MatchCompareFieldType.Integer => Fill<NumericalItem<long>>(matches),
+                MatchCompareFieldType.Sequence => Fill<SequenceItem>(matches),
+            };
+        }
+
+        [SkipLocalsInit]
+        private int Fill<W>(Span<long> matches) where W : struct
+        {
+            int take = _take <= 0 ? matches.Length : _take;
```
ayende, line 144: There are scenarios where users will send `0` as the take value. For example, when they just care about the number of results.
In this case, why bother sorting?

src/Corax/Queries/SortingMatch.cs
```diff
@@ -0,0 +1,325 @@
...
+                tmpMatches = matches.Slice(totalMatches);
+            }
+
+            // We sort the first batch
+            using var hashCacheComparer = new HashCacheMatchComparer<TComparer, W>(_searcher, _comparer);
+            var sorter = new Sorter<long, HashCacheMatchComparer<TComparer, W>>(hashCacheComparer);
+            sorter.Sort(matches.Slice(0, totalMatches));
+
+            Span<long> a = stackalloc long[matches.Length];
+            Span<long> b = stackalloc long[matches.Length];
```
ayende: I think we shouldn't be cleaning this up, if `Length` is high (16K), that can end up being *very* large.
Better to have an "arena" allocation and use that instead, no?

src/Corax/Queries/SortingMatch.cs
ayende: What happens when Take is greater than Length?

src/Corax/Queries/SortingMatch.cs
```diff
@@ -0,0 +1,325 @@
...
+{
+    public unsafe partial struct SortingMatch<TInner, TComparer> : IQueryMatch
+        where TInner : IQueryMatch
+        where TComparer : struct, IMatchComparer
+    {
+        private readonly IndexSearcher _searcher;
+        private readonly IQueryMatch _inner;        
+        private readonly TComparer _comparer;
+        private readonly int _take;
+        private unsafe struct SequenceItem
```
ayende, line 18: Same as `UnmanagedSpan`, no?

src/Voron/Data/Containers/Container.cs
```diff
@@ -457,17 +462,16 @@ public static Item Get(LowLevelTransaction llt, long id)
...
         }
 
         private Span<byte> Get(int offset)
         {
-            ref var metadata = ref Offsets[OffsetToIndex(offset)];
-            Debug.Assert(metadata.Size != 0);
-            return Span.Slice(metadata.Offset, metadata.Size);
+            ItemMetadata* metadata = (ItemMetadata*)(_page.Pointer + offset);
+            Debug.Assert(metadata->Size != 0);
+            return Span.Slice(metadata->Offset, metadata->Size);
```
ayende, line 492: there is a perf difference here?

src/Corax/IndexEntryBuilder.cs
redknightlois: Yes, I was fighting the JIT there... this is a candidate for a constant.

src/Corax/Queries/SortingMatch.Comparers.cs
redknightlois: Good point, will see if it generated the proper code.

src/Corax/Queries/SortingMatch.Comparers.cs
redknightlois: Originally it was using the same code for both, that is a leftover of that.

redknightlois: Because we need the type anyways... Will see if it is doable without losing on the codegen side to just use an instance of an `AscendingMatchComparer` inside.

src/Corax/Queries/SortingMatch.cs
redknightlois: That should not be the case by design. If we allow `Take` to be longer that `Length` then we need an entirely different approach.

src/Corax/Queries/SortingMatch.cs
redknightlois: Yes, I plan to use a pattern there that essentially is something like:
```
Span<long> a = matches.Length < 4096 ? stackalloc long[matches.Length] : ArrayPool<long>.Shared.Rent(matches.Length); 
```

src/Corax/Queries/SortingMatch.cs
ayende: What happens when `Take` is 20K? Or `int.MaxValue` ?

src/Voron/Data/Containers/Container.cs
redknightlois: Yes, huge.

src/Corax/Queries/SortingMatch.cs
ayende: No, why bother with the pool, just allocate a thread local buffer and get from it

redknightlois: Yes, that is an implementation detail. Could be that of basically anything we choose to support that.

src/Corax/Queries/SortingMatch.cs
redknightlois: For unconstrained `Take` we need a version that would take all the potential values and sort afterwards. Different implementation of Sort. This implementation is constrained, and therefore more performant than the unconstrained.

redknightlois:
Updated with the latest vectorized operations. @ayende

redknightlois:
Include latest improvements that pushes sorting cost to roughly 1/4th of the cost in Lucene.

```
|              Method | BufferSize | TakeSize |     Mean |     Error |    StdDev |
|-------------------- |----------- |--------- |---------:|----------:|----------:|
| OrderByRuntimeQuery |       1024 |       16 | 3.356 ms | 0.0122 ms | 0.0114 ms |
| OrderByRuntimeQuery |       2048 |       16 | 2.188 ms | 0.0067 ms | 0.0060 ms |
| OrderByRuntimeQuery |       4096 |       16 | 1.692 ms | 0.0104 ms | 0.0087 ms |
| OrderByRuntimeQuery |      16384 |       16 | 1.223 ms | 0.0032 ms | 0.0026 ms |
```

And the improvements that push the AndOr benchmark below the 3 digits in microseconds. 
```
|                    Method | BufferSize |      Mean |    Error |   StdDev | Code Size |
|-------------------------- |----------- |----------:|---------:|---------:|----------:|
|              RuntimeQuery |         16 | 437.73 us | 1.995 us | 1.769 us |   2,600 B |
| RuntimeQueryOnlyIteration |         16 | 406.00 us | 1.682 us | 1.574 us |   2,216 B |
|               ParserQuery |         16 | 436.47 us | 1.721 us | 1.526 us |   7,980 B |
|  ParserQueryOnlyIteration |         16 | 409.87 us | 3.075 us | 2.401 us |   7,580 B |
|              RuntimeQuery |         64 | 169.24 us | 1.995 us | 1.666 us |   2,600 B |
| RuntimeQueryOnlyIteration |         64 | 143.23 us | 0.608 us | 0.539 us |   2,216 B |
|               ParserQuery |         64 | 174.60 us | 2.531 us | 3.630 us |   7,980 B |
|  ParserQueryOnlyIteration |         64 | 145.56 us | 0.889 us | 0.694 us |   7,580 B |
|              RuntimeQuery |        256 | 101.01 us | 2.008 us | 2.391 us |   2,600 B |
| RuntimeQueryOnlyIteration |        256 |  78.33 us | 0.682 us | 0.638 us |   2,216 B |
|               ParserQuery |        256 |  97.91 us | 0.274 us | 0.214 us |   7,980 B |
|  ParserQueryOnlyIteration |        256 |  79.26 us | 0.593 us | 0.525 us |   7,580 B |
|              RuntimeQuery |       1024 |  79.01 us | 0.302 us | 0.236 us |   2,600 B |
| RuntimeQueryOnlyIteration |       1024 |  61.38 us | 0.800 us | 0.668 us |   2,216 B |
|               ParserQuery |       1024 |  84.32 us | 0.691 us | 0.613 us |   7,980 B |
|  ParserQueryOnlyIteration |       1024 |  59.26 us | 0.279 us | 0.233 us |   7,580 B |
```

PR #12612, opened 2021-08-08
RavenDB-17076 Import - check ClusterTransactionId for null

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-17076

### DIFF
--- a/src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
+++ b/src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
@@ -1417,9 +1417,13 @@ namespace Raven.Server.Smuggler.Documents
                         }
                     }
 
-                    var trxn = ChangeVectorUtils.GetEtagById(oldChangeVector, _database.ClusterTransactionId);
-                    if (trxn > 0)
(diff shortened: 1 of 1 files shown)


haludi:
I wonder if it would be better to check and set Topology.ClusterTransactionIdBase64 when loading the database.

ayende (approved):

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
```diff
@@ -1417,9 +1417,13 @@ private void AddTrxnIfNeeded(DocumentsOperationContext context, string id, ref s
                         }
                     }
 
-                    var trxn = ChangeVectorUtils.GetEtagById(oldChangeVector, _database.ClusterTransactionId);
-                    if (trxn > 0)
-                        changeVector += $",TRXN:{trxn}-{_database.ClusterTransactionId}";
+                    //The ClusterTransactionId can be null if the database was migrated from version smaller then v5.2 
+                    if (_database.ClusterTransactionId != null)
```
karmeli87, line 1421: This is happening when importing to a database that was created before the upgrade?

haludi: Yes

PR #12614, opened 2021-08-08
RavenDB-17088 RemoveInactiveScratches may be called from a write tran…

ayende:
…saction


arekpalinski:
I have opened another PR against 4.2 where also added a test for that case:

https://github.com/ravendb/ravendb/pull/12618

PR #12615, opened 2021-08-08
RavenDB-15926 Wrong number when doing delete by collection query

Danielle9897:


src/Raven.Studio/typescript/viewmodels/database/query/query.ts
```diff
@@ -1382,20 +1389,19 @@ class query extends viewModelBase {
...
         this.runQuery();
-        this.queryFetcher()(0, 1)
-            .done((results) => {
-                if (results.totalResultCount === 0) {
-                    app.showBootstrapMessage("There are no documents matching your query.", "Nothing to do");
-                } else {
-                    this.promptDeleteDocsMatchingQuery(results.totalResultCount);
-                }
-            });
+
```
ml054: do we need runQuery call if we don't call queryFetcher?

Danielle9897: refactored...

ml054 (approved):

PR #12616, opened 2021-08-09
RavenDB-17090 schedule index run only if clean is really needed

karmeli87:


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -1453,19 +1457,26 @@ private void ReduceMemoryUsage(StorageEnvironment environment, IndexCleanup mode
...
+
+                var allocatedBeforeCleanup = indexingStats.TotalAllocated;
                 if (allocatedBeforeCleanup == _allocatedAfterPreviousCleanup)
                     return;
 
                 DocumentDatabase.DocumentsStorage.ContextPool.Clean();
                 _contextPool.Clean();
-                ByteStringMemoryCache.CleanForCurrentThread();
+                
+                if (_indexingThread?.ManagedThreadId == Thread.CurrentThread.ManagedThreadId)
```
ppekrol: Better to create a method `IsInsideIndexingThread` or sth like that.

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -994,10 +996,14 @@ public long GetLastMappedEtagFor(string collection)
             }
         }
 
+        private NativeMemory.ThreadStats _indexingThreadStats;
```
ayende, line 999: What is this for?

karmeli87: To hold a reference to the indexing stats for the indexing thread.
We can call `ReduceMemoryUsage` from outside of the indexing thread, but we need to get the stats of the index.

ayende (approved):

PR #12609, opened 2021-08-08
RavenDB-17079 fix - `Attachments.GetNames` missing retrievement data

omerlewitz:
Break statement caused the issue

https://issues.hibernatingrhinos.com/issue/RavenDB-17079

PR #12620, opened 2021-08-09
RavenDB-17079 fix - `Attachments.GetNames ` not returning the correct number of attachment after deletion 

omerlewitz:
`GetNames` should return the appropriate number of attachments after deletion

The cause was _break statment_


omerlewitz:
https://github.com/ravendb/ravendb/pull/12609

PR #12622, opened 2021-08-09
RavenDB-16961 Enforce revision handle document tombstone

efratshenhar:


src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -1249,6 +1249,11 @@ private long EnforceConfigurationFor(DocumentsOperationContext context, string i
                     changeVector, lastModifiedTicks);
 
                 var currentRevisionsCount = GetRevisionsCount(context, id);
+
+                if (currentRevisionsCount == 0)
+                {
+                    _documentsStorage.Delete(context, lowerId, id, null, fromEnforceConfiguration: true);
```
karmeli87: What if we have document?

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -830,7 +830,7 @@ private static void DeleteCountOfRevisions(DocumentsOperationContext context, Sl
                     return;
                 }
 
-                if (configuration.Disabled == false && configuration.PurgeOnDelete)
+                if (configuration.Disabled == false && configuration. PurgeOnDelete)
```
karmeli87: revert this

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1451,7 +1451,7 @@ public static Tombstone TableValueToTombstone(JsonOperationContext context, ref
         public DeleteOperationResult? Delete(DocumentsOperationContext context, Slice lowerId, string id,
             LazyStringValue expectedChangeVector, long? lastModifiedTicks = null, string changeVector = null,
             CollectionName collectionName = null, NonPersistentDocumentFlags nonPersistentFlags = NonPersistentDocumentFlags.None,
-            DocumentFlags documentFlags = DocumentFlags.None)
+            DocumentFlags documentFlags = DocumentFlags.None, bool fromEnforceConfiguration = false)
```
karmeli87: use `ByEnforceRevisionConfiguration` flag

karmeli87 (changes requested):

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1485,19 +1485,27 @@ public static Tombstone TableValueToTombstone(JsonOperationContext context, ref
...
+                    local.Tombstone.Flags.Contain(DocumentFlags.HasRevisions))
+                {
+                    flags = local.Tombstone.Flags & ~DocumentFlags.HasRevisions;
+                }
+                else
                 {
-                    var revisionsStorage = DocumentDatabase.DocumentsStorage.RevisionsStorage;
-                    if (nonPersistentFlags.Contain(NonPersistentDocumentFlags.FromReplication) == false &&
-                        (revisionsStorage.Configuration != null || flags.Contain(DocumentFlags.Resolved)))
+                    var localFlags = local.Tombstone.Flags.Strip(DocumentFlags.FromClusterTransaction);
```
karmeli87: I think we need this as well?

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -1249,6 +1249,16 @@ private long EnforceConfigurationFor(DocumentsOperationContext context, string i
                     changeVector, lastModifiedTicks);
 
                 var currentRevisionsCount = GetRevisionsCount(context, id);
+
+                if (currentRevisionsCount == 0)
+                {
+                    var tombstoneTable = context.Transaction.InnerTransaction.OpenTable(TombstonesSchema, collectionName.GetTableName(CollectionTableType.Tombstones));
```
karmeli87: Please use `GetDocumentOrTombstone` in the same code path where we do the document check

karmeli87 (approved):

src/Raven.Server/Documents/DocumentsStorage.cs
```diff
@@ -1485,19 +1485,26 @@ public static Tombstone TableValueToTombstone(JsonOperationContext context, ref
...
-
+                DocumentFlags flags;
                 var localFlags = local.Tombstone.Flags.Strip(DocumentFlags.FromClusterTransaction);
-                var flags = localFlags | documentFlags;
-
-                if (collectionName.IsHiLo == false &&
-                    (flags & DocumentFlags.Artificial) != DocumentFlags.Artificial)
+                if (nonPersistentFlags.Contain(NonPersistentDocumentFlags.ByEnforceRevisionConfiguration))
+                {
+                    flags = localFlags.Strip(DocumentFlags.HasRevisions);
```
ppekrol, line 1492: Can we write a comment here explaining why this is required?

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -1249,16 +1249,19 @@ private long EnforceConfigurationFor(DocumentsOperationContext context, string i
                     changeVector, lastModifiedTicks);
 
                 var currentRevisionsCount = GetRevisionsCount(context, id);
+
                 if (needToDeleteMore && currentRevisionsCount > 0)
                     moreWork = true;
 
                 if (currentRevisionsCount == 0)
                 {
+                    var res = _documentsStorage.GetDocumentOrTombstone(context, lowerId, throwOnConflict: false);
```
ppekrol, line 1258: It appears to be a change. Previosuly method could throw in there was a conflict. Do we want that? If yes, do we have a test for that?

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
```diff
@@ -1249,16 +1249,19 @@ private long EnforceConfigurationFor(DocumentsOperationContext context, string i
...
                     // need to strip the HasRevisions flag from the document
-                    var document = _documentsStorage.Get(context, id);
+                    if (res.Tombstone != null)
+                        _documentsStorage.Delete(context, lowerId, id, null, nonPersistentFlags: NonPersistentDocumentFlags.ByEnforceRevisionConfiguration);
 
-                    if (document != null)
-                        _documentsStorage.Put(context, id, null, document.Data.Clone(context),
+                    if (res.Document != null)
```
ppekrol, line 1263: Can Tombstone and Document be not null at the same time?

karmeli87: nope

src/Raven.Server/Documents/Revisions/RevisionsStorage.cs
karmeli87: Actually I'm better without throwing.. it will simply return `null` for both and nothing will happened until the conflict will be resolved.
But yes, having a test to verify the correct behavior here is a good idea. 
Let's do it in the next round of maintenance day?

PR #12631, opened 2021-08-10
RavenDB-16082 - add and endpoint for connectivity check to api.ravendb.net

grisha-kotler:


src/Raven.Server/Web/Studio/LicenseHandler.cs
```diff
@@ -134,5 +135,44 @@ public async Task RenewLicense()
                 context.Write(writer, renewLicense.ToJson());
             }
         }
+
+        [RavenAction("/license-server/connectivity", "GET", AuthorizationStatus.ValidUser)]
```
ayende: should this also be a debug endpoint?

ppekrol: We can add it to the debug package, but the timout needs to be small, 3-5 seconds max IMO.

Do we also want to have access to this during setup?

src/Raven.Server/Web/Studio/LicenseHandler.cs
```diff
@@ -134,5 +135,44 @@ public async Task RenewLicense()
...
+        public async Task CheckConnectivityToLicenseServer()
+        {
+            var result = new ConnectivityToLicenseServer();
+
+            try
+            {
+                var response = await ApiHttpClient.Instance.GetAsync("/Subscription.svc");
```
ppekrol: You do not define a timeout which means it will take up to 100 seconds (default)

ppekrol: I do not think that `/Subscription.svc` is an appropriate endpoint

First of all this is 3.5 endpoint, second this endpoint is not a lightweight endpoint. There should be a dedicated one e.g. `/alive` or sth like that

grisha-kotler: The `/Subscription.svc` endpoint is lightweight (`GET` request) and was added for v3.5.
I'll add a new endpoint for this.

src/Raven.Server/Web/Studio/LicenseHandler.cs
grisha-kotler: done

src/Raven.Server/Web/Studio/LicenseHandler.cs
grisha-kotler: done

PR #12644, opened 2021-08-11
RavenDB-17104 fix remove encrypted node with wrong key

karmeli87:


src/Raven.Server/ServerWide/ServerStore.cs
```diff
@@ -1608,6 +1608,15 @@ public byte[] GetSecretKey(TransactionOperationContext context, string name)
 
         }
 
+        public byte[] GetSecretKey(string name)
```
ppekrol: what is `name` ? Database name? also where this is used?

karmeli87: yes.
It is mainly needed for the JS admin console to get the stored key when the database itself it not loaded.

ppekrol: lets rename it to `databaseName` then, so in half a year we will remember what needs to be passed there exactly

PR #12656, opened 2021-08-16
RavenDB-15890 more info to test

efratshenhar:


test/SlowTests/Issues/RavenDB-15890.cs
```diff
@@ -571,22 +579,32 @@ public async Task LastSetStateDetemineTheState()
 
 
                 index0 = documentDatabase.IndexStore.GetIndex(indexName);
-                index0.SetState(IndexState.Idle);
+                index0.SetState(IndexState.Normal);
```
ppekrol, line 582: isn't that changing the conditions of a test? Do we want that? Previosuly we were checking for Idle, now for Normal

efratshenhar: IndexState.Idle is for auto index and not static. I want to see if the idle causing the problem or that there is a problem with changing the state

PR #12660, opened 2021-08-16
RavenDB-13111 Shard Cluster wide transaction 

haludi:
https://issues.hibernatingrhinos.com/issue/RavenDB-13111

### DIFF
--- a/src/Raven.Client/ServerWide/DatabaseRecord.cs
+++ b/src/Raven.Client/ServerWide/DatabaseRecord.cs
@@ -474,6 +474,12 @@ namespace Raven.Client.ServerWide
 
             return set.ToList();
         }
+
+        public string GetClusterTransactionId()
+        {
+            Debug.Assert(Shards.All(s => s.ClusterTransactionIdBase64.Equals(Shards[0].ClusterTransactionIdBase64)));
+            return Shards[0].ClusterTransactionIdBase64;
+        }
     }
 
     public class IndexHistoryEntry
--- a/src/Raven.Server/Documents/DatabasesLandlord.cs
+++ b/src/Raven.Server/Documents/DatabasesLandlord.cs
@@ -348,11 +348,10 @@ namespace Raven.Server.Documents
 
             // We materialize the values here because we close the read transaction
             var record = rawRecord.MaterializedRecord;
-            
-            if (index == -1)
-                context.CloseTransaction(); // other shards use this context we can't close it
 
-            DeleteDatabase(dbName, deletionInProgress, record);
+            context.Transaction.InnerTransaction.LowLevelTransaction.OnDispose += _ =>
+                DeleteDatabase(dbName, deletionInProgress, record);
+            
             return true;
         }
 
@@ -853,7 +852,7 @@ namespace Raven.Server.Documents
                             if (rawRecord == null)
                                 return;
 
-                        ShouldDeleteDatabase(context, databaseName.Value, rawRecord, fromReplication);
+                            ShouldDeleteDatabase(context, databaseName.Value, rawRecord, fromReplication);
                         }
                     }
                 }
--- a/src/Raven.Server/Documents/DocumentDatabase.cs
+++ b/src/Raven.Server/Documents/DocumentDatabase.cs
@@ -35,6 +35,7 @@ using Raven.Server.Documents.TimeSeries;
 using Raven.Server.Json;
 using Raven.Server.NotificationCenter.Notifications;
 using Raven.Server.NotificationCenter.Notifications.Details;
+using Raven.Server.Rachis;
 using Raven.Server.Routing;
 using Raven.Server.ServerWide;
 using Raven.Server.ServerWide.Commands;
@@ -92,8 +93,6 @@ namespace Raven.Server.Documents
         public string DatabaseGroupId;
         public string ClusterTransactionId;
         public string ShardedDatabaseId;
-
-        public readonly ClusterTransactionWaiter ClusterTransactionWaiter;
 
         private readonly Lazy<RequestExecutor> _requestExecutor;
 
@@ -164,7 +163,6 @@ namespace Raven.Server.Documents
                     }
                 }
 
-                ClusterTransactionWaiter = new ClusterTransactionWaiter(this);
                 QueryMetadataCache = new QueryMetadataCache();
                 IoChanges = new IoChangesNotifications
                 {
--- a/src/Raven.Server/Documents/DocumentDatabase.cs
+++ b/src/Raven.Server/Documents/DocumentDatabase.cs
@@ -527,13 +527,15 @@ namespace Raven.Server.Documents
             }
         }
 
-        struct ClusterTransactionBatch : IDisposable
+        private struct ClusterTransactionBatchCollector : IDisposable
         {
+            private readonly int _maxSize;
             private readonly ClusterTransactionCommand.SingleClusterDatabaseCommand[] _data;
             public int Count = 0;
 
-            public ClusterTransactionBatch(int maxSize)
+            public ClusterTransactionBatchCollector(int maxSize)
             {
+                _maxSize = maxSize;
                 _data = ArrayPool<ClusterTransactionCommand.SingleClusterDatabaseCommand>.Shared.Rent(maxSize);
             }
 
@@ -548,7 +550,12 @@ namespace Raven.Server.Documents
                 }
             }
             
-            public void Add(ClusterTransactionCommand.SingleClusterDatabaseCommand command) => _data[Count++] = command;
+            public void Add(ClusterTransactionCommand.SingleClusterDatabaseCommand command)
+            {
+                if (Count >= _maxSize)
+                    throw new ArgumentOutOfRangeException($"Tried to add beyong the capacity - {nameof(Count)}:{Count}, {nameof(_maxSize)}:{_maxSize}");
+                _data[Count++] = command;
+            }
 
             public void Dispose() => ArrayPool<ClusterTransactionCommand.SingleClusterDatabaseCommand>.Shared.Return(_data);
         }
@@ -557,14 +564,15 @@ namespace Raven.Server.Documents
         {
             const int takeCount = 256;
 
-            using var batch = new ClusterTransactionBatch(takeCount);
+            using var batchCollector = new ClusterTransactionBatchCollector(takeCount);
 
             long maxRaftIndex = -1;
+            long maxCommandCount = -1;
             if (IsShard == false)
             {
                 var readCommands = ClusterTransactionCommand.ReadCommandsBatch(context, Name, fromCount: _nextClusterCommand, take: takeCount);
-                batch.Add(readCommands);
-                if (batch.Count == 0)
+                batchCollector.Add(readCommands);
+                if (batchCollector.Count == 0)
                     return;
             }
             else
@@ -574,18 +582,20 @@ namespace Raven.Server.Documents
                 foreach (var command in readCommands)
                 {
                     maxRaftIndex = command.Index; 
+                    maxCommandCount = command.PreviousCount + command.Commands.Length; 
                     if(command.ShardNumber == _shardProperties.Number)
-                        batch.Add(command);
+                        batchCollector.Add(command);
                 }
                 
-                if (batch.Count == 0)
+                if (batchCollector.Count == 0)
                 {
-                    NotifyIndexIfNeeded();
+                    NotifyIndexIfNeeded(maxRaftIndex, maxCommandCount);
                     return;
                 }
             }
 
-            var mergedCommands = new BatchHandler.ClusterTransactionMergedCommand(this, batch.GetData());
+            var batch = batchCollector.GetData();
+            var mergedCommands = new BatchHandler.ClusterTransactionMergedCommand(this, batch);
             try
             {
                 //If we get a database shutdown while we process a cluster tx command this
@@ -595,26 +605,27 @@ namespace Raven.Server.Documents
             catch (Exception e)
             {
                 if (_logger.IsInfoEnabled)
-                    _logger.Info($"Failed to execute cluster transaction batch (count: {batch.Count}), will retry them one-by-one.", e);
+                    _logger.Info($"Failed to execute cluster transaction batch (count: {batchCollector.Count}), will retry them one-by-one.", e);
 
-                if (await ExecuteClusterTransactionOneByOne(batch.GetData()))
-                    NotifyIndexIfNeeded();
+                if (await ExecuteClusterTransactionOneByOne(batch))
+                    NotifyIndexIfNeeded(maxRaftIndex, maxCommandCount);
                 return;
             }
-            foreach (var command in batch.GetData())
+            foreach (var command in batch)
             {
                 OnClusterTransactionCompletion(command, mergedCommands);
             }
             
-            NotifyIndexIfNeeded();
-
-            void NotifyIndexIfNeeded()
+            NotifyIndexIfNeeded(maxRaftIndex, maxCommandCount);
+        }
+        void NotifyIndexIfNeeded(long index, long count)
+        {
+            if (index >= 0)
             {
-                if (maxRaftIndex >= 0)
-                    RachisLogIndexNotifications.NotifyListenersAbout(maxRaftIndex, null);
+                RachisLogIndexNotifications.NotifyListenersAbout(index, null);
+                _nextClusterCommand = count;
             }
         }
-
         private async Task<bool> ExecuteClusterTransactionOneByOne(ArraySegment<ClusterTransactionCommand.SingleClusterDatabaseCommand> batch)
         {
             for (int i = 0; i < batch.Count; i++)
--- a/src/Raven.Server/Documents/DocumentDatabase.cs
+++ b/src/Raven.Server/Documents/DocumentDatabase.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Buffers;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
@@ -115,6 +116,8 @@ namespace Raven.Server.Documents
 
             Is32Bits = PlatformDetails.Is32Bits || Configuration.Storage.ForceUsing32BitsPager;
 
+            CheckIfShardAndFindItsNumber();
+            
             _disposeOnce = new DisposeOnce<SingleAttempt>(DisposeInternal);
 
             try
@@ -201,6 +204,22 @@ namespace Raven.Server.Documents
             }
         }
 
+        private void CheckIfShardAndFindItsNumber()
+        {
+            int separationIndex = Name.IndexOf('$');
+            if (separationIndex < 0)
+                return;
+
+            var start = separationIndex + 1;
+            int.TryParse(Name.AsSpan(start, Name.Length - start), out var shardNumber);
+            
+            _shardProperties = new ShardProperties
+            {
+                Number = shardNumber,
+                DatabaseName = Name[..separationIndex]
+            };
+        }
+
         public void SetIds(RawDatabaseRecord record) => SetIds(record.Topology, record.ShardedDatabaseId);
         public void SetIds(DatabaseRecord record) => SetIds(record.Topology, record.ShardedDatabaseId);
 
@@ -230,6 +249,15 @@ namespace Raven.Server.Documents
 
         public string Name { get; }
 
+        private class ShardProperties
+        {
+            public int Number;
+            public string DatabaseName;
+        }
+
+        private ShardProperties _shardProperties;
+        private bool IsShard => _shardProperties != null;
+        
         public Guid DbId => DocumentsStorage.Environment?.DbId ?? Guid.Empty;
 
         public string DbBase64Id => DocumentsStorage.Environment?.Base64Id ?? "";
@@ -499,15 +527,65 @@ namespace Raven.Server.Documents
             }
         }
 
+        struct ClusterTransactionBatch : IDisposable
+        {
+            private readonly ClusterTransactionCommand.SingleClusterDatabaseCommand[] _data;
+            public int Count = 0;
+
+            public ClusterTransactionBatch(int maxSize)
+            {
+                _data = ArrayPool<ClusterTransactionCommand.SingleClusterDatabaseCommand>.Shared.Rent(maxSize);
+            }
+
+            public ArraySegment<ClusterTransactionCommand.SingleClusterDatabaseCommand> GetData()
+                => new ArraySegment<ClusterTransactionCommand.SingleClusterDatabaseCommand>(_data, 0, Count);
+
+            public void Add(IEnumerable<ClusterTransactionCommand.SingleClusterDatabaseCommand> data)
+            {
+                foreach (var command in data)
+                {
+                    Add(command);
+                }
+            }
+            
+            public void Add(ClusterTransactionCommand.SingleClusterDatabaseCommand command) => _data[Count++] = command;
+
+            public void Dispose() => ArrayPool<ClusterTransactionCommand.SingleClusterDatabaseCommand>.Shared.Return(_data);
+        }
+        
         public async Task ExecuteClusterTransaction(TransactionOperationContext context)
         {
-            var batch = new List<ClusterTransactionCommand.SingleClusterDatabaseCommand>(
-                ClusterTransactionCommand.ReadCommandsBatch(context, Name, fromCount: _nextClusterCommand, take: 256));
+            const int takeCount = 256;
 
-            if (batch.Count == 0)
-                return;
+            using var batch = new ClusterTransactionBatch(takeCount);
 
-            var mergedCommands = new BatchHandler.ClusterTransactionMergedCommand(this, batch);
+            long maxRaftIndex = -1;
+            if (IsShard == false)
+            {
+                var readCommands = ClusterTransactionCommand.ReadCommandsBatch(context, Name, fromCount: _nextClusterCommand, take: takeCount);
+                batch.Add(readCommands);
+                if (batch.Count == 0)
+                    return;
+            }
+            else
+            {
+                var readCommands = ClusterTransactionCommand.ReadCommandsBatch(context, _shardProperties.DatabaseName, fromCount: _nextClusterCommand, take: takeCount);
+                
+                foreach (var command in readCommands)
+                {
+                    maxRaftIndex = command.Index; 
+                    if(command.ShardNumber == _shardProperties.Number)
+                        batch.Add(command);
+                }
+                
+                if (batch.Count == 0)
+                {
+                    NotifyIndexIfNeeded();
+                    return;
+                }
+            }
+
+            var mergedCommands = new BatchHandler.ClusterTransactionMergedCommand(this, batch.GetData());
             try
             {
                 //If we get a database shutdown while we process a cluster tx command this
@@ -517,26 +595,32 @@ namespace Raven.Server.Documents
             catch (Exception e)
             {
                 if (_logger.IsInfoEnabled)
-                {
                     _logger.Info($"Failed to execute cluster transaction batch (count: {batch.Count}), will retry them one-by-one.", e);
-                }
-                await ExecuteClusterTransactionOneByOne(batch);
+
+                if (await ExecuteClusterTransactionOneByOne(batch.GetData()))
+                    NotifyIndexIfNeeded();
                 return;
             }
-            foreach (var command in batch)
+            foreach (var command in batch.GetData())
             {
                 OnClusterTransactionCompletion(command, mergedCommands);
             }
+            
+            NotifyIndexIfNeeded();
+
+            void NotifyIndexIfNeeded()
+            {
+                if (maxRaftIndex >= 0)
+                    RachisLogIndexNotifications.NotifyListenersAbout(maxRaftIndex, null);
+            }
         }
 
-        private async Task ExecuteClusterTransactionOneByOne(List<ClusterTransactionCommand.SingleClusterDatabaseCommand> batch)
+        private async Task<bool> ExecuteClusterTransactionOneByOne(ArraySegment<ClusterTransactionCommand.SingleClusterDatabaseCommand> batch)
         {
-            foreach (var command in batch)
+            for (int i = 0; i < batch.Count; i++)
             {
-                var singleCommand = new List<ClusterTransactionCommand.SingleClusterDatabaseCommand>
-                {
-                    command
-                };
+                var command = batch[i];
+                var singleCommand = batch.Slice(i, 1);
                 var mergedCommand = new BatchHandler.ClusterTransactionMergedCommand(this, singleCommand);
                 try
                 {
@@ -562,40 +646,58 @@ namespace Raven.Server.Documents
                     await Task.Delay(_clusterTransactionDelayOnFailure, DatabaseShutdown);
                     _clusterTransactionDelayOnFailure = Math.Min(_clusterTransactionDelayOnFailure * 2, 15000);
 
-                    return;
+                    return false;
                 }
             }
+
+            return true;
         }
 
         private void OnClusterTransactionCompletion(ClusterTransactionCommand.SingleClusterDatabaseCommand command,
-            BatchHandler.ClusterTransactionMergedCommand mergedCommands, Exception exception = null)
+            BatchHandler.ClusterTransactionMergedCommand mergedCommands)
         {
             try
             {
                 var index = command.Index;
                 var options = mergedCommands.Options[index];
-                if (exception == null)
+                Task indexTask = null;
+                if (options.WaitForIndexesTimeout != null)
                 {
-                    Task indexTask = null;
-                    if (options.WaitForIndexesTimeout != null)
-                    {
-                        indexTask = BatchHandler.WaitForIndexesAsync(DocumentsStorage.ContextPool, this, options.WaitForIndexesTimeout.Value,
-                            options.SpecifiedIndexesQueryString, options.WaitForIndexThrow,
-                            mergedCommands.LastChangeVector, mergedCommands.LastTombstoneEtag, mergedCommands.ModifiedCollections);
-                    }
-
-                    var result = new BatchHandler.ClusterTransactionCompletionResult
-                    {
-                        Array = mergedCommands.Replies[index],
-                        IndexTask = indexTask,
-                    };
-                    ClusterTransactionWaiter.SetResult(options.TaskId, index, result);
-                    _nextClusterCommand = command.PreviousCount + command.Commands.Length;
-                    _lastCompletedClusterTransaction = _nextClusterCommand.Value - 1;
-                    return;
+                    indexTask = BatchHandler.WaitForIndexesAsync(DocumentsStorage.ContextPool, this, options.WaitForIndexesTimeout.Value,
+                        options.SpecifiedIndexesQueryString, options.WaitForIndexThrow,
+                        mergedCommands.LastChangeVector, mergedCommands.LastTombstoneEtag, mergedCommands.ModifiedCollections);
                 }
 
-                ClusterTransactionWaiter.SetException(options.TaskId, index, exception);
+                var result = new ClusterTransactionCompletionResult
+                {
+                    Array = mergedCommands.Replies[index],
+                    IndexTask = indexTask,
+                };
+                RachisLogIndexNotifications.NotifyListenersAbout(index, null);
+                ServerStore.Cluster.ClusterTransactionWaiter.TrySetResult(options.TaskId, result);
+                _nextClusterCommand = command.PreviousCount + command.Commands.Length;
+                _lastCompletedClusterTransaction = _nextClusterCommand.Value - 1;
+            }
+            catch (Exception e)
+            {
+                // nothing we can do
+                if (_logger.IsInfoEnabled)
+                {
+                    _logger.Info($"Failed to notify about transaction completion for database '{Name}'.", e);
+                }
+            }
+        }
+
+        private void OnClusterTransactionCompletion(ClusterTransactionCommand.SingleClusterDatabaseCommand command,
+            BatchHandler.ClusterTransactionMergedCommand mergedCommands, Exception exception)
+        {
+            try
+            {
+                var index = command.Index;
+                var options = mergedCommands.Options[index];
+
+                RachisLogIndexNotifications.NotifyListenersAbout(index, exception);
+                ServerStore.Cluster.ClusterTransactionWaiter.TrySetException(options.TaskId, exception);
             }
             catch (Exception e)
             {
--- a/src/Raven.Server/Documents/Handlers/BatchHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/BatchHandler.cs
@@ -81,7 +81,11 @@ namespace Raven.Server.Documents.Handlers
                 {
                     if (command.IsClusterTransaction)
                     {
-                        var clusterTransactionHandler = new ClusterTransactionRequestProcessor(this, Database.Name, Database.IdentityPartsSeparator);
+                        var topology = ServerStore.LoadDatabaseTopology(Database.Name);
+                        if (topology.Promotables.Contains(ServerStore.NodeTag))
+                            throw new DatabaseNotRelevantException("Cluster transaction can't be handled by a promotable node.");
+                        
+                        var clusterTransactionHandler = new ClusterTransactionRequestProcessor(this, Database.Name, Database.IdentityPartsSeparator, topology);
                         await clusterTransactionHandler.Process(context, command.ParsedCommands);
                         return;
                     }
--- a/src/Raven.Server/Documents/Handlers/BatchHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/BatchHandler.cs
@@ -81,8 +81,8 @@ namespace Raven.Server.Documents.Handlers
                 {
                     if (command.IsClusterTransaction)
                     {
-                        var clusterTransactionHandler = new ClusterTransactionRequestHandler(this, Database.Name, Database.IdentityPartsSeparator);
-                        await clusterTransactionHandler.Handle(context, command.ParsedCommands);
+                        var clusterTransactionHandler = new ClusterTransactionRequestProcessor(this, Database.Name, Database.IdentityPartsSeparator);
+                        await clusterTransactionHandler.Process(context, command.ParsedCommands);
                         return;
                     }
 
--- a/src/Raven.Server/Documents/Handlers/BatchHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/BatchHandler.cs
@@ -128,23 +128,6 @@ namespace Raven.Server.Documents.Handlers
             }
         }
 
-        internal static void CheckBackwardCompatibility(HttpContext httpContext, ref bool disableAtomicDocumentWrites)
-        {
-            if (disableAtomicDocumentWrites)
-                return;
-
-            if (RequestRouter.TryGetClientVersion(httpContext, out var clientVersion) == false)
-            {
-                disableAtomicDocumentWrites = true;
-                return;
-            }
-
-            if (clientVersion.Major < 5 || (clientVersion.Major == 5 && clientVersion.Minor < 2))
-            {
-                disableAtomicDocumentWrites = true;
-            }
-        }
-
         public static string BatchTrafficWatch(ArraySegment<BatchRequestParser.CommandData> parsedCommands)
         {
             var sb = new StringBuilder();
@@ -388,11 +371,11 @@ namespace Raven.Server.Documents.Handlers
 
         public class ClusterTransactionMergedCommand : TransactionMergedCommand
         {
-            private readonly List<ClusterTransactionCommand.SingleClusterDatabaseCommand> _batch;
+            private readonly ArraySegment<ClusterTransactionCommand.SingleClusterDatabaseCommand> _batch;
             public readonly Dictionary<long, DynamicJsonArray> Replies = new Dictionary<long, DynamicJsonArray>();
             public readonly Dictionary<long, ClusterTransactionCommand.ClusterTransactionOptions> Options = new Dictionary<long, ClusterTransactionCommand.ClusterTransactionOptions>();
 
-            public ClusterTransactionMergedCommand(DocumentDatabase database, List<ClusterTransactionCommand.SingleClusterDatabaseCommand> batch) : base(database)
+            public ClusterTransactionMergedCommand(DocumentDatabase database, ArraySegment<ClusterTransactionCommand.SingleClusterDatabaseCommand> batch) : base(database)
             {
                 _batch = batch;
             }
@@ -419,15 +402,13 @@ namespace Raven.Server.Documents.Handlers
                     {
                         ModifiedCollections = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
                     }
-                    (long, string)? clusterGuardAddition = options.DisableAtomicDocumentWrites == true ? null
-                        : (command.Index, Database.ClusterTransactionId);
 
                     if (commands != null)
                     {
                         foreach (BlittableJsonReaderObject blittableCommand in commands)
                         {
                             count++;
-                            var changeVector = GetClusterWideChangeVector(Database.DatabaseGroupId, count, clusterGuardAddition);
+                            var changeVector = GetClusterWideChangeVector(Database.DatabaseGroupId, count, options.DisableAtomicDocumentWrites == false, command.Index, Database.ClusterTransactionId);
                             var cmd = JsonDeserializationServer.ClusterTransactionDataCommand(blittableCommand);
 
                             switch (cmd.Type)
@@ -574,21 +555,20 @@ namespace Raven.Server.Documents.Handlers
                 };
             }
 
-            public static string GetClusterWideChangeVector(string databaseId, long prevCountPerShard, (long Index, string ClusterId)? clusterAddition)
+            public static string GetClusterWideChangeVector(string databaseId, long prevCountPerShard, bool addTrxAddition, long index, string clusterTransactionId)
             {
                 var stringBuilder = new StringBuilder(ChangeVectorParser.RaftTag)
                     .Append(':').Append(prevCountPerShard)
                     .Append('-').Append(databaseId);
-                if (clusterAddition.HasValue)
+                if (addTrxAddition)
                 {
                     stringBuilder
                         .Append(',').Append(ChangeVectorParser.TrxnTag)
-                        .Append(':').Append(clusterAddition.Value.Index)
-                        .Append('-').Append(clusterAddition.Value.ClusterId);
+                        .Append(':').Append(index)
+                        .Append('-').Append(clusterTransactionId);
                 }
                 return stringBuilder.ToString();
             }
-
         }
 
         public class MergedBatchCommand : TransactionMergedCommand, IDisposable
@@ -1084,7 +1064,7 @@ namespace Raven.Server.Documents.Handlers
 
     public class ClusterTransactionMergedCommandDto : TransactionOperationsMerger.IReplayableCommandDto<BatchHandler.ClusterTransactionMergedCommand>
     {
-        public List<ClusterTransactionCommand.SingleClusterDatabaseCommand> Batch { get; set; }
+        public ArraySegment<ClusterTransactionCommand.SingleClusterDatabaseCommand> Batch { get; set; }
 
         public BatchHandler.ClusterTransactionMergedCommand ToCommand(DocumentsOperationContext context, DocumentDatabase database)
         {
--- a/src/Raven.Server/Documents/Handlers/BatchHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/BatchHandler.cs
@@ -8,6 +8,7 @@ using System.Runtime.ExceptionServices;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
 using Raven.Client.Documents.Attachments;
 using Raven.Client.Documents.Changes;
 using Raven.Client.Documents.Commands.Batches;
@@ -68,18 +69,18 @@ namespace Raven.Server.Documents.Handlers
                 }
                 finally
                 {
-                if (TrafficWatchManager.HasRegisteredClients)
-                {
-                    var log = BatchTrafficWatch(commandBuilder.Commands);
-                    // add sb to httpContext
-                    AddStringToHttpContext(log, TrafficWatchChangeType.BulkDocs);
-                }
+                    if (TrafficWatchManager.HasRegisteredClients)
+                    {
+                        var log = BatchTrafficWatch(commandBuilder.Commands);
+                        // add sb to httpContext
+                        AddStringToHttpContext(log, TrafficWatchChangeType.BulkDocs);
+                    }
                 }
                 
                 var disableAtomicDocumentWrites = GetBoolValueQueryString("disableAtomicDocumentWrites", required: false) ??
                                                   Database.Configuration.Cluster.DisableAtomicDocumentWrites;
 
-                CheckBackwardCompatibility(ref disableAtomicDocumentWrites);
+                CheckBackwardCompatibility(HttpContext, ref disableAtomicDocumentWrites);
 
                 using (var command = await commandBuilder.GetCommand(context))
                 {
@@ -87,7 +88,7 @@ namespace Raven.Server.Documents.Handlers
                     {
                         ValidateCommandForClusterWideTransaction(command);
 
-                        using (Database.ClusterTransactionWaiter.CreateTask(out var taskId))
+                        using (ServerStore.Cluster.ClusterTransactionWaiter.CreateTask(out var taskId))
                         {
                             // Since this is a cluster transaction we are not going to wait for the write assurance of the replication.
                             // Because in any case the user will get a raft index to wait upon on his next request.
@@ -143,12 +144,12 @@ namespace Raven.Server.Documents.Handlers
             }
         }
 
-        private void CheckBackwardCompatibility(ref bool disableAtomicDocumentWrites)
+        internal static void CheckBackwardCompatibility(HttpContext httpContext, ref bool disableAtomicDocumentWrites)
         {
-            if (disableAtomicDocumentWrites) 
+            if (disableAtomicDocumentWrites)
                 return;
 
-            if (RequestRouter.TryGetClientVersion(HttpContext, out var clientVersion) == false)
+            if (RequestRouter.TryGetClientVersion(httpContext, out var clientVersion) == false)
             {
                 disableAtomicDocumentWrites = true;
                 return;
@@ -160,39 +161,10 @@ namespace Raven.Server.Documents.Handlers
             }
         }
 
-        private static void ValidateCommandForClusterWideTransaction(MergedBatchCommand command)
+        internal static void ValidateCommandForClusterWideTransaction(MergedBatchCommand command)
         {
-            foreach (var document in command.ParsedCommands)
-            {
-                switch (document.Type)
-                {
-                    case CommandType.PUT:
-                    case CommandType.DELETE:
-                    case CommandType.CompareExchangePUT:
-                    case CommandType.CompareExchangeDELETE:
-                        if (document.Type == CommandType.PUT)
-                        {
-                            if (document.SeenAttachments)
-                                throw new NotSupportedException($"The document {document.Id} has attachments, this is not supported in cluster wide transaction.");
-
-                            if (document.SeenCounters)
-                                throw new NotSupportedException($"The document {document.Id} has counters, this is not supported in cluster wide transaction.");
-
-                            if (document.SeenTimeSeries)
-                                throw new NotSupportedException($"The document {document.Id} has time series, this is not supported in cluster wide transaction.");
-                        }
-                        break;
-
-                    default:
-                        throw new ArgumentOutOfRangeException($"The command type {document.Type} is not supported in cluster transaction.");
-                }
-            }
-        }
-
-        public class ClusterTransactionCompletionResult
-        {
-            public Task IndexTask;
-            public DynamicJsonArray Array;
+            var commandParsedCommands = command.ParsedCommands;
+            ClusterTransactionCommand.ValidateCommands(commandParsedCommands);
         }
 
         public static string BatchTrafficWatch(ArraySegment<BatchRequestParser.CommandData> parsedCommands)
@@ -248,7 +220,7 @@ namespace Raven.Server.Documents.Handlers
                 using (var timeout = new CancellationTokenSource(ServerStore.Engine.OperationTimeout))
                 using (var cts = CancellationTokenSource.CreateLinkedTokenSource(timeout.Token, HttpContext.RequestAborted))
                 {
-                    reply = (ClusterTransactionCompletionResult)await Database.ClusterTransactionWaiter.WaitForResults(options.TaskId, cts.Token);
+                    reply = await ServerStore.Cluster.ClusterTransactionWaiter.WaitForResults(options.TaskId, cts.Token);
                 }
                 if (reply.IndexTask != null)
                 {
@@ -530,17 +502,15 @@ namespace Raven.Server.Documents.Handlers
                     {
                         ModifiedCollections = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
                     }
+                    (long, string)? clusterGuardAddition = options.DisableAtomicDocumentWrites == true ? null
+                        : (command.Index, Database.ClusterTransactionId);
 
                     if (commands != null)
                     {
                         foreach (BlittableJsonReaderObject blittableCommand in commands)
                         {
                             count++;
-                            var changeVector = $"{ChangeVectorParser.RaftTag}:{count}-{Database.DatabaseGroupId}";
-                            if (options.DisableAtomicDocumentWrites == false)
-                            {
-                                changeVector += $",{ChangeVectorParser.TrxnTag}:{command.Index}-{Database.ClusterTransactionId}";
-                            }
+                            var changeVector = GetClusterWideChangeVector(Database.DatabaseGroupId, count, clusterGuardAddition);
                             var cmd = JsonDeserializationServer.ClusterTransactionDataCommand(blittableCommand);
 
                             switch (cmd.Type)
@@ -686,6 +656,22 @@ namespace Raven.Server.Documents.Handlers
                     Batch = _batch
                 };
             }
+
+            public static string GetClusterWideChangeVector(string databaseId, long prevCountPerShard, (long Index, string ClusterId)? clusterAddition)
+            {
+                var stringBuilder = new StringBuilder(ChangeVectorParser.RaftTag)
+                    .Append(':').Append(prevCountPerShard)
+                    .Append('-').Append(databaseId);
+                if (clusterAddition.HasValue)
+                {
+                    stringBuilder
+                        .Append(',').Append(ChangeVectorParser.TrxnTag)
+                        .Append(':').Append(clusterAddition.Value.Index)
+                        .Append('-').Append(clusterAddition.Value.ClusterId);
+                }
+                return stringBuilder.ToString();
+            }
+
         }
 
         public class MergedBatchCommand : TransactionMergedCommand, IDisposable
--- a/src/Raven.Server/Documents/Handlers/BatchHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/BatchHandler.cs
@@ -54,18 +54,18 @@ namespace Raven.Server.Documents.Handlers
                 var contentType = HttpContext.Request.ContentType;
                 try
                 {
-                if (contentType == null ||
-                    contentType.StartsWith("application/json", StringComparison.OrdinalIgnoreCase))
-                {
-                    await commandBuilder.BuildCommandsAsync(context, RequestBodyStream());
-                }
-                else if (contentType.StartsWith("multipart/mixed", StringComparison.OrdinalIgnoreCase) ||
-                         contentType.StartsWith("multipart/form-data", StringComparison.OrdinalIgnoreCase))
-                {
-                    await commandBuilder.ParseMultipart(context, RequestBodyStream(), HttpContext.Request.ContentType);
-                }
-                else
-                    ThrowNotSupportedType(contentType);
+                    if (contentType == null ||
+                        contentType.StartsWith("application/json", StringComparison.OrdinalIgnoreCase))
+                    {
+                        await commandBuilder.BuildCommandsAsync(context, RequestBodyStream());
+                    }
+                    else if (contentType.StartsWith("multipart/mixed", StringComparison.OrdinalIgnoreCase) ||
+                             contentType.StartsWith("multipart/form-data", StringComparison.OrdinalIgnoreCase))
+                    {
+                        await commandBuilder.ParseMultipart(context, RequestBodyStream(), HttpContext.Request.ContentType);
+                    }
+                    else
+                        ThrowNotSupportedType(contentType);
                 }
                 finally
                 {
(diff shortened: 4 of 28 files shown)


src/Raven.Server/Documents/ClusterTransactionWaiter.cs
```diff
@@ -9,17 +9,47 @@ public class ClusterTransactionWaiter
...
+            TaskCompletionSource<object> created = null;
+            while (true)
+            {
+                if (_results.TryGetValue(id, out var exist))
+                    return exist;
+
+                created ??= new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
+                if (_results.TryAdd(id, created))
+                {
+                    _results.TryAdd(id, created);
```
ppekrol: why this is added twice?

src/Raven.Server/Documents/ClusterTransactionWaiter.cs
```diff
@@ -9,17 +9,47 @@ public class ClusterTransactionWaiter
...
+            CreateOrGetTaskInternal(id);
+            return new DisposableRemoveTask(this, id);
+        }
+
+        public TaskCompletionSource<object> CreateOrGetTaskInternal(string id)
+        {
+            TaskCompletionSource<object> created = null;
+            while (true)
+            {
+                if (_results.TryGetValue(id, out var exist))
```
ppekrol: isn't it easier to use GetOrAdd?

src/Raven.Server/Documents/ClusterTransactionWaiter.cs
```diff
@@ -9,17 +9,47 @@ public class ClusterTransactionWaiter
...
+                    return exist;
+
+                created ??= new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
+                if (_results.TryAdd(id, created))
+                {
+                    _results.TryAdd(id, created);
+                    _queue.Enqueue((DateTime.UtcNow, id));
```
ppekrol: if you would use get or add then inside the 'add' you can update this?

test/SlowTests/Cluster/ShardedClusterTransactionTests.cs
```diff
@@ -0,0 +1,122 @@
...
+                await session.SaveChangesAsync();
+            }
+        }
+
+        [Theory]
+        [InlineData(1, 1)]
+        [InlineData(2, 1)]
+        [InlineData(1, 2)]
+        [InlineData(2, 2)]
+        public async Task ShardedClusterTransaction_WhenAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA(int numberOfNodes, int numberOfShards)
```
ppekrol: AAA? :)

src/Raven.Server/Documents/ClusterTransactionWaiter.cs
ayende: There is also a race here with adding the item and then adding to the queue? Also, why is that _on_ a queue?

src/Raven.Server/Documents/ClusterTransactionWaiter.cs
```diff
@@ -31,10 +61,8 @@ public TaskCompletionSource<object> Get(string id)
         public void SetResult(string id, long index, object result)
         {
             Database.RachisLogIndexNotifications.NotifyListenersAbout(index, null);
-            if (_results.TryGetValue(id, out var task))
-            {
-                task.SetResult(result);
-            }
+            var task = CreateOrGetTaskInternal(id);
+            task.SetResult(result);
```
ayende: Can this be null?

src/Raven.Server/Documents/ClusterTransactionWaiter.cs
```diff
@@ -77,5 +100,24 @@ await using (token.Register(() => task.TrySetCanceled()))
                 return await task.Task.ConfigureAwait(false);
             }
         }
+
+        public void CleanOld()
+        {
+            while (_queue.TryPeek(out var item))
+            {
+                //TODO To make time configurable
```
ayende: I really don't like this approach, and I don't think any of that is needed.
We can tell what the value is upfront and save this complexlity.

src/Raven.Server/Documents/ShardedHandlers/ClusterTransactionResultsHandler.cs
```diff
@@ -0,0 +1,74 @@
...
+using Sparrow.Json.Parsing;
+
+namespace Raven.Server.Documents.ShardedHandlers
+{
+    //TODO To think where this endpoint should be
+
+    public class ClusterTransactionResultsHandler : DatabaseRequestHandler
+    {
+        //TODO To check AuthorizationStatus
+        [RavenAction("/databases/*/cluster-transaction/wait-for-result", "GET", AuthorizationStatus.ValidUser, EndpointType.Read)]
```
ayende: We aren't changing the URLs to support sharding. This isn't required.

ayende (changes requested):

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -145,10 +146,17 @@ await using (var writer = new AsyncBlittableJsonTextWriter(context, ResponseBody
 
         private void CheckBackwardCompatibility(ref bool disableAtomicDocumentWrites)
```
karmeli87: remove this?

src/Raven.Server/Documents/ClusterTransactionWaiter.cs
haludi: It should be `_results.TryAdd(id, created) == false`.
But it is not relevant anymore.
I changed the implementation to get the result from the cluster state machine

src/Raven.Server/Documents/Sharding/ShardLocator.cs
```diff
@@ -13,7 +13,7 @@ public static class ShardLocator
             for (var i = 0; i < ids.Count; i++)
             {
                 var id = ids[i];
-                var shardId = shardedContext.GetShardId(context, id);
+                var shardId = ShardedContext.GetShardId(context, id);
```
karmeli87, line 31: is this the same as `shardedContext`?

src/Raven.Server/Documents/Sharding/ShardedContext.cs
```diff
@@ -51,7 +51,7 @@ public ShardedContext(ServerStore server, DatabaseRecord record)
         /// per TB of overall db size. That means that even for *very* large databases, the
         /// size of the shard is still going to be manageable.
         /// </summary>
-        public int GetShardId(TransactionOperationContext context, string key)
+        public static int GetShardId<TTransaction>(TransactionOperationContext<TTransaction> context, string key) where TTransaction : RavenTransaction
```
karmeli87, line 94: why?

src/Raven.Server/Documents/ClusterTransactionWaiter.cs
haludi: It is not relevant anymore.
I changed the implementation to get the result from the cluster state machine

src/Raven.Server/Documents/ClusterTransactionWaiter.cs
haludi: It is not relevant anymore.
I changed the implementation to get the result from the cluster state machine

src/Raven.Server/Documents/ClusterTransactionWaiter.cs
haludi: It is not relevant anymore.
I changed the implementation to get the result from the cluster state machine

src/Raven.Server/Documents/ClusterTransactionWaiter.cs
haludi: It is not relevant anymore.
I changed the implementation to get the result from the cluster state machine

src/Raven.Server/Documents/ShardedHandlers/ClusterTransactionResultsHandler.cs
haludi: It is not relevant anymore.
I changed the implementation to get the result from the cluster state machine

src/Raven.Server/Rachis/ClusterTransactionWaiter.cs
```diff
@@ -2,56 +2,60 @@
 using System.Collections.Concurrent;
 using System.Threading;
 using System.Threading.Tasks;
+using Sparrow.Json.Parsing;
 
-namespace Raven.Server.Documents
+namespace Raven.Server.Rachis
 {
-    public class ClusterTransactionWaiter
+    public class ClusterTransactionWaiter : AsyncWaiter<ClusterTransactionCompletionResult>
```
karmeli87, line 9: do we need this class?

test/SlowTests/Cluster/ShardedClusterTransactionTests.cs
haludi: Temporary name...

src/Raven.Server/Rachis/ClusterTransactionWaiter.cs
```diff
@@ -2,56 +2,60 @@
...
         }
 
-        public void SetResult(string id, long index, object result)
+        public void TrySetResult(string id, T result)
         {
-            Database.RachisLogIndexNotifications.NotifyListenersAbout(index, null);
             if (_results.TryGetValue(id, out var task))
             {
-                task.SetResult(result);
+                task.TrySetResult(result);
```
karmeli87, line 41: why `TrySetResult` and not `SetResult`? 
`Set` will fail only if the `TCS` was previously set.. for me it was helpful to find debugs of an unexpected behavior.
Like trying to set a result after an exception was already set..

src/Raven.Server/ServerWide/ClusterStateMachine.cs
```diff
@@ -918,38 +922,37 @@ private List<string> ExecuteClusterTransaction(ClusterOperationContext context,
             }
         }
 
-        private DatabaseTopology UpdateDatabaseRecordId(ClusterOperationContext context, long index, ClusterTransactionCommand clusterTransaction)
+        private void UpdateDatabaseAndClusterTransactionForBackwardCompatibility(ClusterOperationContext context, RawDatabaseRecord rawRecord, long index,
```
karmeli87: can we not change this function, but check in the caller whether it is a sharded db or not

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
```diff
@@ -117,8 +120,13 @@ public DynamicJsonValue ToJson()
             ClusterTransactionOptions options, string uniqueRequestId) : base(uniqueRequestId)
         {
             DatabaseName = databaseName;
-            DatabaseRecordId = topology.DatabaseTopologyIdBase64 ?? Guid.NewGuid().ToBase64Unpadded();
-            ClusterTransactionId = topology.ClusterTransactionIdBase64 ?? Guid.NewGuid().ToBase64Unpadded();
+
+            //It is irrelevant in shard database
```
karmeli87: would that make sense to have a different ctor?

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
```diff
@@ -295,28 +332,136 @@ public static string GetAtomicGuardKey(string docId)
...
+            if (SerializedDatabaseCommands == null)
+                return;
+
+            if (rawRecord.IsSharded())
+            {
+                if (SerializedDatabaseCommands.TryGet(nameof(DatabaseCommands), out BlittableJsonReaderArray commands) == false)
+                    throw new InvalidOperationException($"Cluster {nameof(SerializedDatabaseCommands)} don't include the actual commands : {SerializedDatabaseCommands}");
+
+                //TODO Can I relay on the shards to be numbered consecutively from 0 ???
+                var shardCount = rawRecord.MaterializedRecord.ShardAllocations.Count;
```
karmeli87: why do you need to materialize the record here?

karmeli87: why `ShardAllocations` and not `Shards` ?

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
```diff
@@ -295,28 +332,136 @@ public static string GetAtomicGuardKey(string docId)
...
+                var shardIndexPerId = new (string Id, int Index)[commands.Length];
+                var prevCountPerShard = new long[shardCount];
+
+                var commandsItr = 0;
+                foreach (BlittableJsonReaderObject command in commands)
+                {
+                    if (command.TryGet(nameof(ClusterTransactionDataCommand.Id), out string id) == false)
+                        throw new InvalidOperationException($"Got cluster transaction database command without an id: {command}");
+
+                    var shardIndex = ShardedContext.GetShardIndex(context, rawRecord.MaterializedRecord, id);
```
karmeli87: can we avoid the MaterializedRecord ?

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
```diff
@@ -295,28 +332,136 @@ public static string GetAtomicGuardKey(string docId)
...
+                        [nameof(Options)] = Options.ToJson()
+                    }, "serialized-database-commands");
+
+                    var shard = rawRecord.DatabaseName + "$" + i;
+                    SaveCommandBatch(context, index, shard, commandsCountPerDatabase, items, toSave, out var prevCount);
+                    prevCountPerShard[i] = prevCount;
+                }
+
+                var result = GetDatabaseCommandsResult(rawRecord, index, shardIndexPerId, prevCountPerShard);
+                context.Transaction.InnerTransaction.LowLevelTransaction.OnDispose += tx =>
```
karmeli87, line 500: what if we fail to commit?

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
```diff
@@ -295,28 +332,136 @@ public static string GetAtomicGuardKey(string docId)
...
+            {
+                if (SerializedDatabaseCommands.TryGet(nameof(DatabaseCommands), out BlittableJsonReaderArray commands) == false)
+                    throw new InvalidOperationException($"Cluster {nameof(SerializedDatabaseCommands)} don't include the actual commands : {SerializedDatabaseCommands}");
+
+                //TODO Can I relay on the shards to be numbered consecutively from 0 ???
+                var shardCount = rawRecord.MaterializedRecord.ShardAllocations.Count;
+                
+                var commandsPreShard = new DynamicJsonArray[shardCount];
+                var shardIndexPerId = new (string Id, int Index)[commands.Length];
+                var prevCountPerShard = new long[shardCount];
```
karmeli87: Can we combine those into a single entity with
`Id`, `Shard`, `PrevCount` ?

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -527,17 +507,16 @@ protected override long ExecuteCmd(DocumentsOperationContext context)
...
                     {
                         foreach (BlittableJsonReaderObject blittableCommand in commands)
                         {
                             count++;
-                            var changeVector = $"{ChangeVectorParser.RaftTag}:{count}-{Database.DatabaseGroupId}";
-                            if (options.DisableAtomicDocumentWrites == false)
-                            {
-                                changeVector += $",{ChangeVectorParser.TrxnTag}:{command.Index}-{Database.ClusterTransactionId}";
-                            }
+                            var changeVector = GetClusterWideChangeVector(Database.DatabaseGroupId, count, clusterGuardAddition);
```
karmeli87: cosmetics, but let's make it
`string GetClusterWideChangeVector(string databaseGroupId, long count, string clusterTransactionId, long index)`
and for legacy
`string GetClusterWideChangeVector(string databaseGroupId, long count)`

src/Raven.Server/Documents/Handlers/BatchHandler.cs
haludi: Done

src/Raven.Server/Documents/Sharding/ShardLocator.cs
haludi: I turned it to be static so I can use this logic in the cluster state machine.

src/Raven.Server/Documents/Sharding/ShardedContext.cs
haludi: I want to use it in the cluster state machine.
Is there a problem with that?

src/Raven.Server/Rachis/ClusterTransactionWaiter.cs
haludi: What is the question?
If we need the inheritance?
The question is if we want to separate the result from the `DocumentDatabase` to the result from the `ClusterStateMachine`.
For now, I used the same collection for both of them so actually, I can remove the inheritance and define directly the tasks to return  `ClusterTransactionCompletionResult` (anyway I think it is prettier like that but it is not important if you don't)

Anyway, as it is implemented now I have to use `TrySet` at least in the `DocumentDatabase` because there can be two attempts to set the result. 
Or I can conditionally set it only if it is not a shared database. does it sound good?

src/Raven.Server/Rachis/ClusterTransactionWaiter.cs
haludi: https://github.com/ravendb/ravendb/pull/12660#discussion_r698228174

src/Raven.Server/ServerWide/ClusterStateMachine.cs
haludi: yes, but I do need the `rawRecord` outside and because I need to dispose it I prefer to get it outside.
Is it ok?

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
haludi: Yes, it will be nicer :)

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
haludi: No need here (so I will switch it with `Shards`)
But later I need the `ShardAllocations` for `ShardedContext.GetShardIndex`

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
haludi: Yes, you mean to deserialize only the `ShardAllocations`, right?

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
haludi: I guess we will try again, no?
I can set the result anyway but then I will have to use `TrySetResult` to at least check if the result was set before.

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
haludi: I do need to think about how to polish this but I wanted to be sure the main implementation concept is acceptable for you.
Anyway `Id` & `Shard` has a different length than `PrevCount`

src/Raven.Server/Documents/Handlers/BatchHandler.cs
haludi: I like cosmetics but I prefer to avoid reallocate the cluster addition

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -160,39 +161,10 @@ private void CheckBackwardCompatibility(ref bool disableAtomicDocumentWrites)
...
-                        break;
-
-                    default:
-                        throw new ArgumentOutOfRangeException($"The command type {document.Type} is not supported in cluster transaction.");
-                }
-            }
-        }
-
-        public class ClusterTransactionCompletionResult
+        internal static void ValidateCommandForClusterWideTransaction(MergedBatchCommand command)
```
karmeli87: what is the difference between this and `ValidateCommands`?

src/Raven.Server/Documents/Handlers/BatchHandler.cs
haludi: Done

src/Raven.Client/ServerWide/DatabaseRecord.cs
```diff
@@ -451,6 +451,12 @@ public IEnumerable<string> GetTopologyMembers(Func<DatabaseTopology, List<string
 
             return set.ToList();
         }
+
+        public string GetClusterTransactionId()
```
haludi, line 478: The decision were to continue using the `ClusterTransactionIdBase64` from the topology this value must be the same on all shard topologies

src/Raven.Server/Smuggler/Documents/DatabaseDestination.cs
```diff
@@ -621,33 +621,29 @@ private async ValueTask SendClusterTransactionsAsync(JsonOperationContext contex
                 if (_clusterTransactionCommands.Length == 0)
                     return;
 
-                using (_database.ClusterTransactionWaiter.CreateTask(out var taskId))
```
haludi: This change is not relevant for this PR but now I understand there is no need for retrieving the result mechanism here
so I made a little cleaning

test/SlowTests/Cluster/ShardedClusterTransactionTests.cs
```diff
@@ -0,0 +1,102 @@
...
+                    await session.StoreAsync(testObj, $"TestObjs/{i}");
+                }
+                
+                await session.SaveChangesAsync();
+
+                // WaitForUserToContinueTheTest(store);
+
+                //TODO To remove waiting
+                await Task.Delay(2000);
+                using var session2 = store.OpenAsyncSession(new SessionOptions
```
haludi, line 66: For now the `Load` does not wait on the raft index

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
```diff
@@ -97,6 +103,82 @@ await using (var writer = new AsyncBlittableJsonTextWriter(context, ResponseBody
...
+                }
+                await ServerStore.WaitForCommitIndexChange(RachisConsensus.CommitIndexModification.GreaterOrEqual, clusterTransactionCommandResult.Index);
+                
+                DynamicJsonArray result;
+                if (clusterTransactionCommand.DatabaseCommands.Count > 0)
+                {
+                    using var timeout = new CancellationTokenSource(ServerStore.Engine.OperationTimeout);
+                    using var cts = CancellationTokenSource.CreateLinkedTokenSource(timeout.Token, HttpContext.RequestAborted);
+                    var databaseResult = await ServerStore.Cluster.ClusterTransactionWaiter.WaitForResults(taskId, cts.Token);
+                    result = databaseResult.Array;
```
ayende: Isn't this a race condition? When we have many concurrent transactions and the buffer is cleared?

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
ayende: You need to check here that the transaction committed.

test/SlowTests/Cluster/ShardedClusterTransactionTests.cs
ayende: what do you mean?

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -559,18 +557,20 @@ private async Task ExecuteClusterTransactionOneByOne(List<ClusterTransactionComm
...
 
-                    var result = new BatchHandler.ClusterTransactionCompletionResult
+                    var result = new ClusterTransactionCompletionResult
                     {
                         Array = mergedCommands.Replies[index],
                         IndexTask = indexTask,
                     };
-                    ClusterTransactionWaiter.SetResult(options.TaskId, index, result);
+                    RachisLogIndexNotifications.NotifyListenersAbout(index, null);
+                    ServerStore.Cluster.ClusterTransactionWaiter.TrySetResult(options.TaskId, result);
```
ayende: what if the current node doesn't contains any of the shards ?

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
haludi: Are you talking about `ClusterTransactionWaiter._results`?
Only the handler removes the result (by dispose the return value of `CreateTask`).

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
haludi: Done

test/SlowTests/Cluster/ShardedClusterTransactionTests.cs
haludi: As I understand it, for now, we don't read `Constants.Headers.LastKnownClusterTransactionIndex` in a shard database handler (in the `RequestRouter`).
So the client doesn't wait for the cluster transaction to be executed and the test got null here.

src/Raven.Server/Documents/DocumentDatabase.cs
haludi: In a sharded database, the results are already set by the cluster state machine so `TrySetResult` will "fail" if there are shards in the node that handles the request

test/SlowTests/Cluster/ShardedClusterTransactionTests.cs
ayende: Why aren't we just forwarding the value to the shard instances? Then it will Just Work, no?

haludi: We need to add that to the header of the requests sent from the shard handler.
I can add it here for `Load` but I guess it will be more appropriate to implement something more generic.
I will start to check that but maybe it is worth separating it to another issue.

ayende: That should be passed along always, yes. Can add that in the shard handler base class, I think

src/Raven.Server/Documents/Handlers/BatchRequestParser.cs
```diff
@@ -584,6 +584,33 @@ public Stream GetBlob(long blobSize)
                             // we need this only for identities and we expect to have always null
                             bufferedCommand.ChangeVectorPosition = checked((int)(commandCopy.Position + parser.BufferOffset - position - 4));
                         }
+                        break;
```
ayende, line 603: Check with @karmeli87 why we need a cloned method here.

src/Raven.Server/Documents/ShardedHandlers/ShardedBatchHandler.cs
```diff
@@ -97,6 +103,82 @@ await using (var writer = new AsyncBlittableJsonTextWriter(context, ResponseBody
                 }
             }
         }
+
+        private async Task HandleClusterTransaction(TransactionOperationContext context, ShardedBatchCommand batch)
```
ayende: duplicate with the non sharded version
ideally we want a single copy

src/Raven.Server/Rachis/ClusterTransactionWaiter.cs
```diff
@@ -2,56 +2,60 @@
...
             }
         }
 
-        public void SetException(string id, long index, Exception e)
+        public void TrySetException(string id, Exception e)
```
ayende, line 45: Remember to actually handle this scenario as well

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
```diff
@@ -295,28 +333,130 @@ public static string GetAtomicGuardKey(string docId)
...
+            else
+            {
+                var commands = context.ReadObject(SerializedDatabaseCommands, "serialized-tx-commands");
+                SaveCommandBatch(context, index, DatabaseName, commandsCountPerDatabase, items, commands, out _);
+            }
+        }
+
+        private static long[] GetPrevCountPerShard(ClusterOperationContext context, Tree commandsCountPerDatabase, RawDatabaseRecord rawRecord)
+        {
+            long[] prevCountPerShard = new long[rawRecord.Shards.Length];
```
ayende: try to avoid this allocation

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
```diff
@@ -295,28 +333,130 @@ public static string GetAtomicGuardKey(string docId)
...
+                    if (commandsPreShard[i] == null)
+                        continue;
+
+                    var toSave = context.ReadObject(new DynamicJsonValue
+                    {
+                        [nameof(DatabaseCommands)] = commandsPreShard[i],
+                        [nameof(Options)] = Options.ToJson()
+                    }, "serialized-database-commands");
+
+                    var shard = rawRecord.DatabaseName + "$" + i;
```
ayende: try to avoid this allocation

ppekrol:
https://jenkins.hibernatingrhinos.com/job/PR_Tests_Features/163/console

```
11:18:28  Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
11:18:28     at Sparrow.Json.Parsing.DynamicJsonValue.Remove(System.String)
11:18:28     at Sparrow.Json.Parsing.DynamicJsonValue.set_Item(System.String, System.Object)
11:18:28     at Raven.Server.ServerWide.RawDatabaseRecord.GetShardedDatabaseRecord(Int32)
11:18:28     at Raven.Server.ServerWide.RawDatabaseRecord+<GetShardedDatabaseRecords>d__33.MoveNext()
11:18:28     at Raven.Server.Documents.DatabasesLandlord+<HandleClusterDatabaseChanged>d__20.MoveNext()
11:18:28     at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[[Raven.Server.Documents.DatabasesLandlord+<HandleClusterDatabaseChanged>d__20, Raven.Server, Version=5.3.0.0, Culture=neutral, PublicKeyToken=null]](<HandleClusterDatabaseChanged>d__20 ByRef)
11:18:28     at Raven.Server.Documents.DatabasesLandlord.HandleClusterDatabaseChanged(System.String, Int64, System.String, ClusterDatabaseChangeType, System.Object)
11:18:28     at Raven.Server.ServerWide.ClusterChanges+<OnDatabaseChanges>d__14.MoveNext()
11:18:28     at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[[Raven.Server.ServerWide.ClusterChanges+<OnDatabaseChanges>d__14, Raven.Server, Version=5.3.0.0, Culture=neutral, PublicKeyToken=null]](<OnDatabaseChanges>d__14 ByRef)
11:18:28     at Raven.Server.ServerWide.ClusterChanges.OnDatabaseChanges(System.String, Int64, System.String, ClusterDatabaseChangeType, System.Object)
11:18:28     at Raven.Server.ServerWide.ClusterStateMachine+<>c__DisplayClass85_0+<<ExecuteAsyncTask>b__0>d.MoveNext()
11:18:28     at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[[Raven.Server.ServerWide.ClusterStateMachine+<>c__DisplayClass85_0+<<ExecuteAsyncTask>b__0>d, Raven.Server, Version=5.3.0.0, Culture=neutral, PublicKeyToken=null]](<<ExecuteAsyncTask>b__0>d ByRef)
11:18:28     at Raven.Server.ServerWide.ClusterStateMachine+<>c__DisplayClass85_0.<ExecuteAsyncTask>b__0()
11:18:28     at System.Threading.Tasks.Task`1[[System.__Canon, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].InnerInvoke()
11:18:28     at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(System.Threading.Thread, System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)
11:18:28     at System.Threading.Tasks.Task.ExecuteWithThreadLocal(System.Threading.Tasks.Task ByRef, System.Threading.Thread)
11:18:28     at System.Threading.ThreadPoolWorkQueue.Dispatch()
```

haludi:
> https://jenkins.hibernatingrhinos.com/job/PR_Tests_Features/163/console
> 
> ```
> 11:18:28  Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
> 11:18:28     at Sparrow.Json.Parsing.DynamicJsonValue.Remove(System.String)
> 11:18:28     at Sparrow.Json.Parsing.DynamicJsonValue.set_Item(System.String, System.Object)
> 11:18:28     at Raven.Server.ServerWide.RawDatabaseRecord.GetShardedDatabaseRecord(Int32)
> 11:18:28     at Raven.Server.ServerWide.RawDatabaseRecord+<GetShardedDatabaseRecords>d__33.MoveNext()
> 11:18:28     at Raven.Server.Documents.DatabasesLandlord+<HandleClusterDatabaseChanged>d__20.MoveNext()
> 11:18:28     at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[[Raven.Server.Documents.DatabasesLandlord+<HandleClusterDatabaseChanged>d__20, Raven.Server, Version=5.3.0.0, Culture=neutral, PublicKeyToken=null]](<HandleClusterDatabaseChanged>d__20 ByRef)
> 11:18:28     at Raven.Server.Documents.DatabasesLandlord.HandleClusterDatabaseChanged(System.String, Int64, System.String, ClusterDatabaseChangeType, System.Object)
> 11:18:28     at Raven.Server.ServerWide.ClusterChanges+<OnDatabaseChanges>d__14.MoveNext()
> 11:18:28     at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[[Raven.Server.ServerWide.ClusterChanges+<OnDatabaseChanges>d__14, Raven.Server, Version=5.3.0.0, Culture=neutral, PublicKeyToken=null]](<OnDatabaseChanges>d__14 ByRef)
> 11:18:28     at Raven.Server.ServerWide.ClusterChanges.OnDatabaseChanges(System.String, Int64, System.String, ClusterDatabaseChangeType, System.Object)
> 11:18:28     at Raven.Server.ServerWide.ClusterStateMachine+<>c__DisplayClass85_0+<<ExecuteAsyncTask>b__0>d.MoveNext()
> 11:18:28     at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[[Raven.Server.ServerWide.ClusterStateMachine+<>c__DisplayClass85_0+<<ExecuteAsyncTask>b__0>d, Raven.Server, Version=5.3.0.0, Culture=neutral, PublicKeyToken=null]](<<ExecuteAsyncTask>b__0>d ByRef)
> 11:18:28     at Raven.Server.ServerWide.ClusterStateMachine+<>c__DisplayClass85_0.<ExecuteAsyncTask>b__0()
> 11:18:28     at System.Threading.Tasks.Task`1[[System.__Canon, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].InnerInvoke()
> 11:18:28     at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(System.Threading.Thread, System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)
> 11:18:28     at System.Threading.Tasks.Task.ExecuteWithThreadLocal(System.Threading.Tasks.Task ByRef, System.Threading.Thread)
> 11:18:28     at System.Threading.ThreadPoolWorkQueue.Dispatch()
> ```

@ppekrol, I don't think it is related to my PR.
I can get it failed in my machine and the test doesn't use ClusterTransactio and even batch command.

ayende:
*can* get, or *can't* get?
At any rate, need to test it again.

The stack trace is in managed stack, and isn't touching native memory, so likely heap corruption?

ppekrol:
Please mind that tests are running concurrently and it happened 2 out of 2 retires of this PR and it does not happen in the main branch. Scheduled another 2 builds.

src/Raven.Server/Documents/Handlers/ClusterTransactionRequestHandler.cs
```diff
@@ -0,0 +1,128 @@
...
+using Raven.Server.ServerWide;
+using Raven.Server.ServerWide.Commands;
+using Raven.Server.ServerWide.Context;
+using Raven.Server.Web;
+using Sparrow.Json;
+using Sparrow.Json.Parsing;
+
+namespace Raven.Server.Documents.Handlers
+{
+    public class ClusterTransactionRequestHandler
```
karmeli87: Usually the convention `...RequestHandler` is to handle web routes (and inherit for the `RequestHandler`), let's rename this class to something like `ClusterTransactionRequestProcessor`

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -143,12 +124,12 @@ await using (var writer = new AsyncBlittableJsonTextWriter(context, ResponseBody
             }
         }
 
-        private void CheckBackwardCompatibility(ref bool disableAtomicDocumentWrites)
+        internal static void CheckBackwardCompatibility(HttpContext httpContext, ref bool disableAtomicDocumentWrites)
```
karmeli87: Do we need this? since you moved it to a dedicated class.

src/Raven.Server/Documents/Handlers/BatchHandler.cs
haludi: `ValidateCommandForClusterWideTransaction` was changed to `ClusterTransactionCommand.ValidateCommands`

src/Raven.Server/Documents/Handlers/ClusterTransactionRequestHandler.cs
haludi: done

src/Raven.Server/Documents/DocumentDatabase.cs
ayende: I don't understand this? You mean we won't set the value? How do we get it back to the caller, then?

Consider a sharded db on B & C.
We go to node A to run the action, how do we get the result back?

src/Raven.Server/Documents/Handlers/BatchRequestParser.cs
ayende: Reminder

src/Raven.Server/Documents/DocumentDatabase.cs
haludi: I will be set by the cluster state machine (so even if it will get here the result will be already set).
https://github.com/haludi/ravendb/blob/RavenDB-13111/src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs#L391

src/Raven.Server/Documents/Handlers/BatchRequestParser.cs
haludi: I need to refactor the code to use the same path in both case

src/Raven.Server/Rachis/ClusterTransactionWaiter.cs
haludi: While I don't wait for the database command to be executed I can't get this exception.
And the exception from the compare exchange I will get from the cluster state machine
https://github.com/haludi/ravendb/blob/RavenDB-13111/src/Raven.Server/Documents/Handlers/ClusterTransactionRequestProcessor.cs#L68

src/Raven.Server/Documents/Handlers/BatchRequestParser.cs
haludi: https://issues.hibernatingrhinos.com/issue/RavenDB-17370

src/Raven.Server/Documents/DatabasesLandlord.cs
```diff
@@ -341,9 +341,10 @@ public bool ShouldDeleteDatabase(TransactionOperationContext context, string dbN
 
             // We materialize the values here because we close the read transaction
             var record = rawRecord.MaterializedRecord;
-            context.CloseTransaction();
 
-            DeleteDatabase(dbName, deletionInProgress, record);
+            context.Transaction.InnerTransaction.LowLevelTransaction.OnDispose += _ =>
```
haludi, line 352: We close the transaction here so it will not be opened while actually deleting the database (can take a long time).
In shard database, the raw database record of the shard database is built from the main raw database record.
On closing the transaction here we also disposing the main raw database record so in the next iteration we read from a freed memory.
I think the fact we close the transaction here is surprising while we call this function from inside scope of `using` of the transaction.
So I think it would be more clear to keep the transaction closing on the closing of the `using` scope and just register the delete with the relevant data.
The only thing that bothers me is the name of the event `OnDispose` even though it called in the end (and actually after) of transaction dispose so the transaction is already closed while doing the actual delete.

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
```diff
@@ -346,30 +384,171 @@ public static string GetAtomicGuardKey(string docId)
...
+
+        private void InsertCommandBatch(ClusterOperationContext context, long index, string databaseName, Tree commandsCountPerDatabase, Table items,
+            BlittableJsonReaderObject commands, long commandsCount)
+        {
+            SaveCommandBatch(context, index, databaseName, commandsCountPerDatabase, items, commands, commandsCount, ItemsInsert);
+        }
+        private void SetCommandBatch(ClusterOperationContext context, long index, string databaseName, Tree commandsCountPerDatabase, Table items,
+            BlittableJsonReaderObject commands, long commandsCount)
+        {
+            SaveCommandBatch(context, index, databaseName, commandsCountPerDatabase, items, commands, commandsCount, ItemsSet);
```
haludi: When sending a database request we send the last known raft command index (saved in the document store level).
Sending cluster-wide transactions with a write request to let say only $0 will update the client with the new raft index of the cluster transaction command.
If then the client sends another database request (line `Load`) it will contain the new index.
If that database request is then directed to the same shard the directed request will contain the new index to wait for that index to be applied on the shard so the client will not get an old result.
But if the `Load` request is directed to another shard (let's say $1) it will wait for the index to be updated but the origin command is not relevant for him.
This is only relevant if another raft command didn't apply on that shard (like creating/modifying index) and updated the index to be higher than the cluster transaction command index.
I included here a way to solve it.
The cluster state machine will send empty command if no database command is related to it.
That will throw an exception in the previous implementation because the `prevCount` is the same so the key is the same and we use `Inset` (NewOnly).
I know @karmeli87 like "guard" like that to catch potential bugs (And I like the idea) so I modified only the shard implementation to use `set`.

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -53,55 +54,39 @@ public async Task BulkDocs()
...
-                                    WaitForIndexesTimeout = waitForIndexesTimeout,
-                                    WaitForIndexThrow = waitForIndexThrow,
-                                    SpecifiedIndexesQueryString = specifiedIndexesQueryString.Count > 0 ? specifiedIndexesQueryString.ToList() : null
-                                };
-                            await HandleClusterTransaction(context, command, options);
-                        }
-
+                        var topology = ServerStore.LoadDatabaseTopology(Database.Name);
+                        if (topology.Promotables.Contains(ServerStore.NodeTag))
+                            throw new DatabaseNotRelevantException("Cluster transaction can't be handled by a promotable node.");
```
karmeli87, line 86: Does this relevant for the sharded cluster tx?

src/Raven.Server/Documents/Handlers/ClusterTransactionRequestProcessor.cs
```diff
@@ -0,0 +1,173 @@
...
+                }
+                await _handler.ServerStore.WaitForCommitIndexChange(RachisConsensus.CommitIndexModification.GreaterOrEqual, clusterTransactionCommandResult.Index);
+
+                DynamicJsonArray result;
+                if (clusterTransactionCommand.DatabaseCommands.Count > 0)
+                {
+                    using var timeout = new CancellationTokenSource(_handler.ServerStore.Engine.OperationTimeout);
+                    using var cts = CancellationTokenSource.CreateLinkedTokenSource(timeout.Token, _handler.HttpContext.RequestAborted);
+                    var databaseResult = await _handler.ServerStore.Cluster.ClusterTransactionWaiter.WaitForResults(taskId, cts.Token);
+                    result = databaseResult.Array;
```
karmeli87, line 76: We need to wait on the database as well no?

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
```diff
@@ -346,30 +384,171 @@ public static string GetAtomicGuardKey(string docId)
...
+            {
+                if (SerializedDatabaseCommands.TryGet(nameof(DatabaseCommands), out BlittableJsonReaderArray commands) == false)
+                    throw new InvalidOperationException($"Cluster {nameof(SerializedDatabaseCommands)} don't include the actual commands : {SerializedDatabaseCommands}");
+
+                var commandsPreShard = new DynamicJsonArray[rawRecord.Shards.Length];
+                var shardNames = GetShardNames(rawRecord.DatabaseName, rawRecord.Shards.Length);
+                var prevCountPerShard = GetPrevCountPerShard(context, commandsCountPerDatabase, shardNames);
+
+                (long, string)? clusterGuardAddition = Options.DisableAtomicDocumentWrites == true ? null
+                    : (index, rawRecord.GetClusterTransactionId());
```
karmeli87: let's make it nicer.

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -530,17 +419,15 @@ protected override long ExecuteCmd(DocumentsOperationContext context)
                     {
                         ModifiedCollections = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
                     }
+                    (long, string)? clusterGuardAddition = options.DisableAtomicDocumentWrites == true ? null
+                        : (command.Index, Database.ClusterTransactionId);
```
karmeli87: pass the `options`  and the `index` to `GetClusterWideChangeVector` and assemble it there

src/Raven.Server/Documents/Sharding/ShardedRequestHandler.cs
```diff
@@ -54,14 +54,19 @@ public override void Init(RequestHandlerContext context)
...
-            if (header.HasFlag(Headers.IfNoneMatch))
+            }
+            else if(header.HasFlag(Headers.IfNoneMatch))
+            {
                 command.Headers["If-None-Match"] = GetStringFromHeaders("If-None-Match");
+            }
+
+            var lastKnownClusterTransactionIndex = GetStringFromHeaders(Constants.Headers.LastKnownClusterTransactionIndex);
+            if (lastKnownClusterTransactionIndex != null)
+                command.Headers[Constants.Headers.LastKnownClusterTransactionIndex] = lastKnownClusterTransactionIndex;
```
karmeli87: What happened if:
- you reach an EP of `/databases/Northwind` ? (and it may or may not have any shard of this database)
- you reach an EP of `/databases/Northwind$0` while the actual db tx was done on ` Northwind$2`?

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
```diff
@@ -346,30 +384,171 @@ public static string GetAtomicGuardKey(string docId)
...
+                }
+
+                context.Transaction.InnerTransaction.LowLevelTransaction.OnDispose += tx =>
+                {
+                    if (context.Transaction.InnerTransaction.LowLevelTransaction.Committed == false)
+                        return;
+
+                    clusterTransactionWaiter.TrySetResult(Options.TaskId, new ClusterTransactionCompletionResult
+                    {
+                        Array = result
```
karmeli87: does your result here contain blittables? if yes it might be invalidated out side of this tx.

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
```diff
@@ -346,30 +384,171 @@ public static string GetAtomicGuardKey(string docId)
...
+
+        private string[] GetShardNames(string rawRecordDatabaseName, int shardsLength)
+        {
+            var shardNames = new string[shardsLength];
+            for (int i = 0; i < shardsLength; i++)
+            {
+                shardNames[i] = rawRecordDatabaseName + '$' + i;
+            }
+            return shardNames;
+        }
```
karmeli87: Move this to ShardedHelper

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
karmeli87: wouldn't it be easier that each database on each shard will read the command and update the notifier regardless if this related to him or not?

src/Raven.Server/Documents/Sharding/ShardedRequestHandler.cs
karmeli87: discussion is here: https://github.com/ravendb/ravendb/pull/12660/files#r811975137

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
```diff
@@ -346,30 +384,171 @@ public static string GetAtomicGuardKey(string docId)
...
+            var changeVector = BatchHandler.ClusterTransactionMergedCommand.GetClusterWideChangeVector(databaseId, currentCommandCount, clusterGuardAddition);
+
+            result[Constants.Documents.Metadata.ChangeVector] = changeVector;
+            return result;
+        }
+
+        private void InsertCommandBatch(ClusterOperationContext context, long index, string databaseName, Tree commandsCountPerDatabase, Table items,
+            BlittableJsonReaderObject commands, long commandsCount)
+        {
+            SaveCommandBatch(context, index, databaseName, commandsCountPerDatabase, items, commands, commandsCount, ItemsInsert);
```
karmeli87: instead of `ItemsInsert` simply `isNew: true/false`

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
karmeli87: ping me about that I want to better understand this

src/Raven.Server/ServerWide/RawDatabaseRecord.cs
```diff
@@ -231,33 +258,44 @@ public bool IsSharded()
...
-                [nameof(DatabaseRecord.ShardAllocations)] = null,
-                [nameof(DatabaseRecord.Shards)] = null,
-                [nameof(DatabaseRecord.Settings)] = DynamicJsonValue.Convert(settings)
-            };
-
-            return new RawDatabaseRecord(_context, _context.ReadObject(_record, shardName));
+                Console.WriteLine(shardName);
+                if (e is AccessViolationException)
+                    throw new AccessViolationException($"Temp Debug Info {shardName}", e);
+                throw;
```
karmeli87: ?

test/RachisTests/DatabaseCluster/AtomicClusterReadWriteTests.cs
```diff
@@ -628,4 +599,102 @@ private static IDisposable LocalGetDocumentStores(string[] urls, string database
...
+            await base.ClusterWideTransaction_WhenLoadAndUpdateInParallel_ShouldSucceedOnlyInTheFirst();
+        }
+
+        [Fact]
+        public async Task ClusterWideTransaction_WhenLoadAndDeleteWhileUpdated_ShouldFailDeletion()
+        {
+            await base.ClusterWideTransaction_WhenLoadAndDeleteWhileUpdated_ShouldFailDeletion();
+        }
+
+        [Fact(Skip = "Should complete shard implementation")]
```
karmeli87: please unskip those when `https://github.com/ravendb/ravendb/pull/13671` merged

src/Raven.Server/Documents/Handlers/BatchHandler.cs
haludi: Removed.
Rebase's residues, thanks

src/Raven.Server/Documents/Handlers/BatchHandler.cs
haludi: No, but this is the regular database endpoint

src/Raven.Server/Documents/Handlers/ClusterTransactionRequestProcessor.cs
haludi: `WaitForResults` handle that.
On not sharded database the `DocumentDatabase` set the result.

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
haludi: Done

src/Raven.Server/Documents/Handlers/BatchHandler.cs
haludi: Done

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
ayende: The database _should_ update it for newer values, yes.

src/Raven.Server/Documents/Sharding/ShardedRequestHandler.cs
haludi: Are you talking here about the raft index issue we spoke about in the presentation or it is another issue?

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
haludi: Done

src/Raven.Server/ServerWide/Commands/ClusterTransactionCommand.cs
haludi: Not relevant anymore

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -223,6 +240,15 @@ private void SetIds(DatabaseTopology topology, string shardedDatabaseId)
...
 
+        private class ShardProperties
+        {
+            public int Number;
+            public string DatabaseName;
+        }
+
+        private ShardProperties _shardProperties;
+        private bool IsShard => _shardProperties != null;
+
```
karmeli87, line 258: @ppekrol those keep repeat it self...
I wonder if we should go the path of having `ShardedDocumentDatabase : DocumentDatabase`

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -510,26 +586,32 @@ public async Task ExecuteClusterTransaction(TransactionOperationContext context)
...
             }
-            foreach (var command in batch)
+            foreach (var command in batch.GetData())
             {
                 OnClusterTransactionCompletion(command, mergedCommands);
             }
+            
+            NotifyIndexIfNeeded();
+
+            void NotifyIndexIfNeeded()
```
karmeli87: let's pass the raftIndex here and move this method outside

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -492,15 +518,65 @@ private async Task ExecuteClusterTransactionTask()
             }
         }
 
+        struct ClusterTransactionBatch : IDisposable
+        {
+            private readonly ClusterTransactionCommand.SingleClusterDatabaseCommand[] _data;
+            public int Count = 0;
```
karmeli87, line 532: private

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -492,15 +518,65 @@ private async Task ExecuteClusterTransactionTask()
...
+
+            public void Add(IEnumerable<ClusterTransactionCommand.SingleClusterDatabaseCommand> data)
+            {
+                foreach (var command in data)
+                {
+                    Add(command);
+                }
+            }
+            
+            public void Add(ClusterTransactionCommand.SingleClusterDatabaseCommand command) => _data[Count++] = command;
```
karmeli87: do you need to check boundaries here?

karmeli87: if you try to add more than `maxSize` ?

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -492,15 +518,65 @@ private async Task ExecuteClusterTransactionTask()
...
 
-            if (batch.Count == 0)
-                return;
+            using var batch = new ClusterTransactionBatch(takeCount);
 
-            var mergedCommands = new BatchHandler.ClusterTransactionMergedCommand(this, batch);
+            long maxRaftIndex = -1;
+            if (IsShard == false)
+            {
+                var readCommands = ClusterTransactionCommand.ReadCommandsBatch(context, Name, fromCount: _nextClusterCommand, take: takeCount);
```
karmeli87, line 571: `_nextClusterCommand` is now common to all shards? do we need to update it accordingly even if we skip?

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -492,15 +518,65 @@ private async Task ExecuteClusterTransactionTask()
...
+                }
+                
+                if (batch.Count == 0)
+                {
+                    NotifyIndexIfNeeded();
+                    return;
+                }
+            }
+
+            var mergedCommands = new BatchHandler.ClusterTransactionMergedCommand(this, batch.GetData());
```
karmeli87: can we call `batch.GetData()` once?

src/Raven.Server/Documents/DocumentDatabase.cs
haludi: yes

haludi: Yes, I will update it

src/Raven.Server/Documents/DocumentDatabase.cs
haludi: Why?
It will do shallow copy anyway.
I don't think it is relevant for performance (since it is struct)
Do you want to call it once because of how it is looked?

src/Raven.Server/Documents/DocumentDatabase.cs
haludi: I think it is unnecessary to check here while it was implemented for this specific task and not for general purposes.
The max is 256 and we read up to 256.
But if you think it is required I will add

src/Raven.Server/Documents/DocumentDatabase.cs
haludi: done

src/Raven.Server/Documents/DocumentDatabase.cs
haludi: done

src/Raven.Server/Documents/DocumentDatabase.cs
karmeli87: > Do you want to call it once because of how it is looked?

yes

src/Raven.Server/Documents/DocumentDatabase.cs
karmeli87: let's add it.
better safe than sorry here

src/Raven.Server/Documents/DocumentDatabase.cs
haludi: Done

src/Raven.Server/Documents/DocumentDatabase.cs
haludi: Done

src/Raven.Server/Documents/DocumentDatabase.cs
haludi: Done

src/Raven.Server/Documents/DocumentDatabase.cs
arekpalinski: Was it pushed? It's still `public`

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -555,40 +648,58 @@ private async Task ExecuteClusterTransactionOneByOne(List<ClusterTransactionComm
...
+                };
+                RachisLogIndexNotifications.NotifyListenersAbout(index, null);
+                ServerStore.Cluster.ClusterTransactionWaiter.TrySetResult(options.TaskId, result);
+                _nextClusterCommand = command.PreviousCount + command.Commands.Length;
+                _lastCompletedClusterTransaction = _nextClusterCommand.Value - 1;
+            }
+            catch (Exception e)
+            {
+                // nothing we can do
+                if (_logger.IsInfoEnabled)
```
arekpalinski, line 693: I wonder if `Info` is sufficient? It won't be visible when using our default logging settings - is that ok? Not sure how bad is that

src/Raven.Server/Documents/Handlers/BatchHandler.cs
```diff
@@ -686,6 +554,21 @@ public override TransactionOperationsMerger.IReplayableCommandDto<TransactionOpe
                     Batch = _batch
                 };
             }
+
+            public static string GetClusterWideChangeVector(string databaseId, long prevCountPerShard, bool addTrxAddition, long index, string clusterTransactionId)
```
arekpalinski, line 558: Looks like a good candidate to be part of `ChangeVectorUtils`

src/Raven.Server/Documents/Handlers/ClusterTransactionRequestProcessor.cs
```diff
@@ -0,0 +1,173 @@
...
+            }
+
+            if (clientVersion.Major < 5 || (clientVersion.Major == 5 && clientVersion.Minor < 2))
+            {
+                disableAtomicDocumentWrites = true;
+            }
+        }
+    }
+
+    class ClusterTransactionRequestProcessor : BaseClusterTransactionRequestProcessor
```
arekpalinski, line 130: Let's move it to separate file. Having the classes defined in dedicated files makes that it's easy to find it on Github. We use it very often.

src/Raven.Server/Documents/Handlers/ClusterTransactionRequestProcessor.cs
```diff
@@ -0,0 +1,173 @@
...
+                Database,
+                IdentitySeparator,
+                _topology,
+                parsedCommands,
+                options,
+                raftRequestId);
+        }
+    }
+    
+    class ShardClusterTransactionRequestProcessor : BaseClusterTransactionRequestProcessor
```
arekpalinski, line 154: Same here

test/RachisTests/DatabaseCluster/AtomicClusterReadWriteTests.cs
arekpalinski: This got merged

src/Raven.Client/ServerWide/DatabaseRecord.cs
```diff
@@ -474,6 +474,12 @@ public IEnumerable<string> GetTopologyMembers(Func<DatabaseTopology, List<string
...
```
ppekrol, line 478: internal

src/Raven.Server/Documents/DatabasesLandlord.cs
ppekrol: @arekpalinski can you verify this?

src/Raven.Server/Documents/DocumentDatabase.cs
```diff
@@ -196,6 +197,22 @@ public DocumentDatabase(string name, RavenConfiguration configuration, ServerSto
             }
         }
 
+        private void CheckIfShardAndFindItsNumber()
+        {
+            int separationIndex = Name.IndexOf('$');
```
ppekrol, line 207: isn't shardHelper doing same thing?

src/Raven.Server/Documents/DocumentDatabase.cs
ppekrol: k

src/Raven.Server/ServerWide/RawDatabaseRecord.cs
```diff
@@ -232,9 +259,8 @@ public bool IsSharded()
         public RawDatabaseRecord GetShardedDatabaseRecord(int index)
         {
             _record.TryGet(nameof(DatabaseRecord.Shards), out BlittableJsonReaderArray array);
-            var name = DatabaseName;
             var shardedTopology = (BlittableJsonReaderObject)array[index];
-            var shardName = name + "$" + index;
+            var shardName = DatabaseName + "$" + index;
```
ppekrol, line 263: ShardHelper?

test/RachisTests/DatabaseCluster/ShardedAtomicClusterReadWriteTests.cs
```diff
@@ -0,0 +1,124 @@
...
+
+namespace RachisTests.DatabaseCluster
+{
+    public class ShardedAtomicClusterReadWriteTests : AtomicClusterReadWriteTestsBase
+    {
+        public ShardedAtomicClusterReadWriteTests(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        protected override DocumentStore GetDocumentStore(Options options = null, [CallerMemberName] string caller = null)
```
ppekrol, line 18: remove, use .Sharding.GetDocumentStore
base method should have IDocumentStore as a parameter

ppekrol: Actually they should use RavenData and RavenTheory

src/Raven.Server/Documents/DatabasesLandlord.cs
arekpalinski: LGTM. Yes, `OnDispose` is called after we return the write transaction lock. @haludi do you think we should backport this change to 5.2?

test/Tests.Infrastructure/RavenTestBase.cs
```diff
@@ -1389,9 +1389,9 @@ private Options(bool frozen)
                 _frozen = frozen;
             }
 
-            public static Options ForMode(RavenDatabaseMode mode)
+            public static Options ForMode(RavenDatabaseMode mode, Options options = null)
```
ppekrol, line 1392: revert

test/SlowTests/Cluster/ShardedClusterTransactionTests.cs
```diff
@@ -0,0 +1,131 @@
...
+
+namespace SlowTests.Cluster
+{
+    public class ShardedClusterTransactionTests : ClusterTransactionTests
+    {
+        public ShardedClusterTransactionTests(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        protected override DocumentStore GetDocumentStore(Options options = null, [CallerMemberName] string caller = null)
```
ppekrol, line 22: revert

ppekrol:
https://jenkins.hibernatingrhinos.com/job/PR_Tests/928/testReport/junit/SlowTests.Issues/RavenDB_7322/SlowTests_Issues_RavenDB_7322_TestClassesShouldNotInheritFromOtherTestClassesToNotMultiplyTests/

src/Raven.Server/Documents/DocumentDatabase.cs
haludi: For catastrophic failure, I agree but for `TimeoutException` form `BatchHandler.WaitForIndexesAsync` I think the info is enough.
Or actually, we should handle that totally different in this case...
But this is relevant for V5.2 so may be worth opening a dedicated issue for that.
I opened an issue about `WaitForIndexesAsync` while using cluster-wide transaction so we can handle that there
https://issues.hibernatingrhinos.com/issue/RavenDB-18373
@arekpalinski, what do you think?

src/Raven.Server/Documents/Handlers/BatchHandler.cs
haludi: Good point, thanks

src/Raven.Server/Documents/Handlers/ClusterTransactionRequestProcessor.cs
haludi: Done

src/Raven.Server/Documents/Handlers/ClusterTransactionRequestProcessor.cs
haludi: Done

src/Raven.Client/ServerWide/DatabaseRecord.cs
haludi: Done

src/Raven.Server/Documents/DocumentDatabase.cs
haludi: Yes, changed

src/Raven.Server/ServerWide/RawDatabaseRecord.cs
haludi: Done

test/RachisTests/DatabaseCluster/ShardedAtomicClusterReadWriteTests.cs
haludi: I think this is not relevant anymore after @karmeli87 's commit but I changed the implementation of the overridden `GetDocumentStore` to use `Sharding.GetDocumentStore`

src/Raven.Server/Documents/DocumentDatabase.cs
haludi: Not relevant anymore

test/Tests.Infrastructure/RavenTestBase.cs
haludi: Done

test/SlowTests/Cluster/ShardedClusterTransactionTests.cs
haludi: Don't you like overriding the `GetDocumentStore`?
The alternative is to have a function test for sharded and unsharded with `DocumentStore` as a parameter.
But that means I need to create the cluster and the document store separately for sharded and unsharded tests.
I think the "override" is cleaner? Or did I miss the meaning of your comment?

src/Raven.Server/Documents/DatabasesLandlord.cs
haludi: It doesn't harm anything right now in a regular database but I think it makes much more sense.
It really surprising to find that the transaction was closed in inner function while its `using` scope is still "active"
So in the bottom line, yes I think we should fix it on v5.2 as well

src/Raven.Server/Documents/DocumentDatabase.cs
arekpalinski: Catastrophic failures are handled by dedicated `CatastrophicFailureHandler` - so it will be recorded to the log.

Not sure what you mean by "we should handle that totally different in this case...". Can you elaborate?

haludi: I understood it mistakenly, the catch here doesn't catch the `TimeoutException`.
I don't think there is something that can fail here right now but we can change the log to operation.
I will add it to [RavenDB-18373](https://issues.hibernatingrhinos.com/issue/RavenDB-18373) so it will be fixed for v5.2 as well.

arekpalinski: ok, great

src/Raven.Server/Documents/DocumentDatabase.cs
haludi: @karmeli87 , did you reffer to `ClusterTransactionBatch` or `Count`?

PR #12688, opened 2021-08-23
RavenDB-17048 - Make `/admin/debug/memory/gc` EP more human readable

lastav5:
https://issues.hibernatingrhinos.com/issue/RavenDB-17048


ayende (changes requested):
No, create a *separate* set of values here, which will be humane.
Don't mess the existing values. Users may rely on them not changing

PR #12698, opened 2021-08-26
RavenDB-17134 Allow a query result to output 0..N results to the quer…

ayende:
…y output

### Issue link

https://issues.hibernatingrhinos.com/issue/RavenDB-17134

### Additional description

This change allows RavenDB to output flatten multiple values per result from the index. The intent is to allow more complex queries and projections out to the user. 

### Type of change

- New feature

### How risky is the change?

- Moderate risk - we are changing the query engine. I doubt there will be any regression that won't be caught by the existing tests.

### Backward compatibility

- Non breaking change

The change make what was previously an error into a valid operation.
### Documentation update

- This change requires a documentation update. Please mark the issue on YouTrack using `Documentation Required` tag.

Will be done after review.

### Testing 

- Tests have been added that prove the fix is effective or that the feature works

Waiting for tests to see if the approach is valid or if there are other options.

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexReadOperation.cs
+++ b/src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexReadOperation.cs
@@ -155,6 +155,7 @@ namespace Raven.Server.Documents.Indexes.Persistence.Lucene
                             continue;
                         }
 
+                        bool markedAsSkipped = false;
                         var r = retriever.Get(document, scoreDoc, _state);
                         if (r.Document != null)
                         {
@@ -180,7 +181,12 @@ namespace Raven.Server.Documents.Indexes.Persistence.Lucene
                             {
                                 d?.Dispose();
 
-                                skippedResults.Value++;
+                                if (markedAsSkipped == false)
+                                {
+                                    skippedResults.Value++;
+                                    markedAsSkipped = true;
+                                }
+
                                 return default;
                             }
 
--- a/src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexReadOperation.cs
+++ b/src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexReadOperation.cs
@@ -166,12 +166,19 @@ namespace Raven.Server.Documents.Indexes.Persistence.Lucene
                         }
                         else if (r.List != null)
                         {
+                            int numberOfProjectedResults = 0;
                             foreach (Document item in r.List)
                             {
                                 var qr = CreateQueryResult(item);
                                 if(qr.Result == null)
                                     continue;
                                 yield return qr;
+                                numberOfProjectedResults++;
+                            }
+
+                            if (numberOfProjectedResults > 1)
+                            {
+                                totalResults.Value += numberOfProjectedResults - 1;
                             }
                         }
 
--- a/src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexReadOperation.cs
+++ b/src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexReadOperation.cs
@@ -155,44 +155,61 @@ namespace Raven.Server.Documents.Indexes.Persistence.Lucene
                             continue;
                         }
 
-                        var result = retriever.Get(document, scoreDoc, _state);
-                        if (scope.TryIncludeInResults(result) == false)
+                        var r = retriever.Get(document, scoreDoc, _state);
+                        if (r.Document != null)
                         {
-                            result?.Dispose();
-
-                            skippedResults.Value++;
-                            continue;
+                            var qr = CreateQueryResult(r.Document);
+                            if(qr.Result == null)
+                                continue;
+                            yield return qr;
+                        }
+                        else if (r.List != null)
+                        {
+                            foreach (Document item in r.List)
+                            {
+                                var qr = CreateQueryResult(item);
+                                if(qr.Result == null)
+                                    continue;
+                                yield return qr;
+                            }
                         }
 
-                        returnedResults++;
-
-                        if (isDistinctCount == false)
+                        QueryResult CreateQueryResult(Document d)
                         {
-                            Dictionary<string, Dictionary<string, string[]>> highlightings = null;
-                            if (query.Metadata.HasHighlightings)
+                            if (scope.TryIncludeInResults(d) == false)
                             {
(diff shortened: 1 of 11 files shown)


src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexReadOperation.cs
```diff
@@ -155,44 +155,61 @@ public IEnumerable<QueryResult> Query(IndexQueryServerSide query, QueryTimingsSc
...
                         {
-                            Dictionary<string, Dictionary<string, string[]>> highlightings = null;
-                            if (query.Metadata.HasHighlightings)
+                            if (scope.TryIncludeInResults(d) == false)
                             {
-                                using (highlightingScope?.Start())
-                                    highlightings = GetHighlighterResults(query, _searcher, scoreDoc, result, document, documentsContext);
+                                d?.Dispose();
+
+                                skippedResults.Value++;
```
ppekrol: is this valid for lists? shouldn't we only increase that by 1 for each list? that raises me a question how paging works here?

src/Raven.Server/Documents/Queries/CollectionQueryEnumerable.cs
```diff
@@ -306,15 +347,44 @@ private long Initialize()
                     {
                         count++;
 
-                        var doc = _fieldsToFetch.IsProjection
-                            ? _resultsRetriever.GetProjectionFromDocument(document, null, QueryResultRetrieverBase.ZeroScore, _fieldsToFetch, _context, null)
-                            : _inner.Current;
+                        Document doc;
```
ppekrol: not used variable

src/Raven.Server/Documents/Queries/Results/QueryResultRetrieverBase.cs
```diff
@@ -291,26 +293,53 @@ protected Document AddProjectionToResult(Document doc, Lucene.Net.Search.ScoreDo
...
+                        {
+                            foreach (var document in result.List)
+                            {
+                                results.Add(document);
+                            }
+                        }
+                    }
+
+                    return (null, results);
+                    break;
```
ppekrol: unreachable code

src/Raven.Server/Documents/Queries/CollectionQueryEnumerable.cs
```diff
@@ -306,15 +347,44 @@ private long Initialize()
...
+                                {
+                                    if (IsStartingPoint(item))
+                                    {
+                                        match = true;
+                                        break;
+                                    }
+                                }
 
-                        if (doc.Data.Count <= 0)
-                            continue;
```
arekpalinski, line 314: Previously we did `continue`. In current code I don't see it anymore - is that okay?

src/Raven.Server/Documents/Queries/CollectionQueryEnumerable.cs
```diff
@@ -306,15 +347,44 @@ private long Initialize()
...
+                            {
+                                break;
+                            }
+                        }
 
-                        if (_alreadySeenProjections.Add(doc.DataHash) && _alreadySeenProjections.Count == _query.Start)
-                            break;
+                        bool IsStartingPoint(Document d)
+                        {
+                            return d.Data.Count > 0 && _alreadySeenProjections.Add(d.DataHash) && _alreadySeenProjections.Count == _query.Start;
```
arekpalinski, line 385: It's a little bit tricky that `IsStartingPoint` calls `Add()` under the covers and modifies `_alreadySeenProjections`. It's a local method so not a big issue.

src/Raven.Server/Documents/Indexes/Persistence/Lucene/IndexReadOperation.cs
ayende: Paging work before this is called. So that isn't accounted for this.
For multiple results, you are correct, this should be once per document, not result.

src/Raven.Server/Documents/Queries/CollectionQueryEnumerable.cs
ayende: Was moved to `IsStartingPoint`

src/Raven.Server/Documents/Queries/CollectionQueryEnumerable.cs
ayende: The idea was to wrap the logic to call it from both locations, yes.

src/Raven.Server/Documents/Queries/CollectionQueryEnumerable.cs
arekpalinski: `doc.Data.Count <= 0` was moved to `IsStartingPoint` but my question was that now we never have `continue` in the code. 

In old code we had `continue` so we iterated to next document in the loop:

`foreach (var document in _documents.GetDocumentsFrom(_context, _collection, 0, start, _query.PageSize))`

now we have only `break` so I believe we might break the loop too early

ayende: No, this is fine. Previously, we would continue if `Count == 0`, now we break if `Count > 0`.
Otherwise, we just go back to the loop again. The logic was reversed.

src/Raven.Server/Documents/Queries/CollectionQueryEnumerable.cs
ayende: And yes, the fact that this is a local method and is right there was a factor

ayende:
Take a look in particular at this query:

```
declare timeseries time_walked_by_day(e){
    from e.Walks group by 1d select sum()
}
declare function project(e){
    return time_walked_by_day(e).Results;
}
from Users as e
where e.Dogs  == null
select project(e)
```

The idea is that we should be able to feed this into something like Grafana directly, with nicer input.

I do think that we'll need nicer syntax for this, but at least now we can push the results into Grafana compatible format in a much easier manner.

arekpalinski (commented):
Changes looks good to me. Although I'd like to see tests which returns results from multiple documents (in addition to current ones where we have just a single user)

arekpalinski (approved):

ppekrol (approved):
LGTM. @ayende can you check failing tests?

ayende:
looking

ayende:
Should all work now

PR #12700, opened 2021-08-26
RavenDB-17123 - Query over index with spatial field in studio is showing __ignored for that field

shiranshalom:


src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/AnonymousLuceneDocumentConverter.cs
```diff
@@ -99,7 +99,12 @@ protected override int GetFields<T>(T instance, LazyStringValue key, LazyStringV
 
                 if (storedValue != null && shouldSkip == false)
                 {
-                    storedValue[property.Key] = TypeConverter.ToBlittableSupportedType(value, flattenArrays: true);
+                    var blittableValue = TypeConverter.ToBlittableSupportedType(value, flattenArrays: true);
```
ppekrol: probably it would be better to pass parameter e.g. `ignoreFieldables` or sth like that?

ppekrol (commented):
A test would be great.

src/Raven.Server/Documents/Indexes/Persistence/Lucene/Documents/AnonymousLuceneDocumentConverter.cs
arekpalinski: @ppekrol 
Such parameter wouldn't help much I think - we need to return an object anyway (`null` is still a valid value)

@shiranshalom 
Why not to use the approach we talked about - creating an overload of `ToBlittableSupportedType()` that returns `out BlittableSupportedReturnType`. Then we could filter out the value if we have `BlittableSupportedReturnType.Ignored`

arekpalinski:
We need to have tests for both cases: when using `CreateField()` and `CreateSpatialField()`

arekpalinski:
Why the PR is missing our PR template (https://github.com/ravendb/ravendb/blob/v5.2/.github/pull_request_template.md)? It should look like here - https://github.com/ravendb/ravendb/pull/12699

arekpalinski:
It was closed in favor of https://github.com/ravendb/ravendb/pull/12704

PR #12703, opened 2021-08-29
RavenDB-17115 have an exclusive access to lock the memory

karmeli87:
### Issue link

https://issues.hibernatingrhinos.com/issue/RavenDB-17115

### Additional description

There are 2 fixes here. 

~~1 - if the database was already removed, avoid re-adding a faulty task to the cache.
Since it might end-up throwing an exception upon actually adding that database.~~

2 - Have an exclusive access to lock the memory
Might be a race when one thread increase the WS but another one locks the memory.

### Type of change

- Bug fix
- Optimization

### How risky is the change?

- Moderate 

- The memory locking is relevant only for windows

### Testing 

- It has been verified by manual testing 
Run in a loop creation of an encrypted database on 7 nodes (reset the WS each iteration)

- No

### DIFF
--- a/src/Raven.Server/Documents/Handlers/Debugging/MemoryDebugHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/Debugging/MemoryDebugHandler.cs
@@ -260,7 +260,7 @@ namespace Raven.Server.Documents.Handlers.Debugging
                 [nameof(MemoryInfo.UnmanagedAllocations)] = Size.Humane(totalUnmanagedAllocations),
                 [nameof(MemoryInfo.EncryptionBuffersInUse)] = Size.Humane(encryptionBuffers.CurrentlyInUseSize),
                 [nameof(MemoryInfo.EncryptionBuffersPool)] = Size.Humane(encryptionBuffers.TotalPoolSize),
-                [nameof(MemoryInfo.EncryptionLockedMemory)] = Sodium.LockedMemory.ToString(),
+                [nameof(MemoryInfo.EncryptionLockedMemory)] = Size.Humane(Sodium.LockedBytes),
                 [nameof(MemoryInfo.MemoryMapped)] = Size.Humane(totalMapping),
                 [nameof(MemoryInfo.ScratchDirtyMemory)] = memInfo.TotalScratchDirtyMemory.ToString(),
                 [nameof(MemoryInfo.IsHighDirty)] = dirtyMemoryState.IsHighDirty,
--- a/src/Raven.Server/Documents/Handlers/Debugging/MemoryDebugHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/Debugging/MemoryDebugHandler.cs
@@ -14,6 +14,7 @@ using Sparrow.Json.Parsing;
 using Sparrow.LowMemory;
 using Sparrow.Platform;
 using Sparrow.Platform.Posix;
+using Sparrow.Server;
 using Sparrow.Server.Platform.Win32;
 using Sparrow.Utils;
 using Size = Raven.Client.Util.Size;
@@ -271,6 +272,7 @@ namespace Raven.Server.Documents.Handlers.Debugging
                 [nameof(MemoryInfo.LowMemLastFiveMinute)] = memoryUsageRecords.Low.LastFiveMinutes.ToString(),
                 [nameof(MemoryInfo.HighMemSinceStartup)] = memoryUsageRecords.High.SinceStartup.ToString(),
                 [nameof(MemoryInfo.LowMemSinceStartup)] = memoryUsageRecords.Low.SinceStartup.ToString(),
+                [nameof(MemoryInfo.LockedMemory)] = Sodium.LockedMemory.ToString(),
             };
 
             writer.WritePropertyName(nameof(MemoryInformation));
@@ -475,6 +477,7 @@ namespace Raven.Server.Documents.Handlers.Debugging
             public string LowMemLastFiveMinute { get; set; }
             public string HighMemSinceStartup { get; set; }
             public string LowMemSinceStartup { get; set; }
+            public string LockedMemory { get; set; }
             public MemoryInfoMappingItem[] Mappings { get; set; }
         }
 
--- a/src/Raven.Server/Documents/Handlers/Debugging/MemoryDebugHandler.cs
+++ b/src/Raven.Server/Documents/Handlers/Debugging/MemoryDebugHandler.cs
@@ -260,6 +260,7 @@ namespace Raven.Server.Documents.Handlers.Debugging
                 [nameof(MemoryInfo.UnmanagedAllocations)] = Size.Humane(totalUnmanagedAllocations),
                 [nameof(MemoryInfo.EncryptionBuffersInUse)] = Size.Humane(encryptionBuffers.CurrentlyInUseSize),
                 [nameof(MemoryInfo.EncryptionBuffersPool)] = Size.Humane(encryptionBuffers.TotalPoolSize),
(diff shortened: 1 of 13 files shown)


src/Raven.Server/Documents/Handlers/Debugging/MemoryDebugHandler.cs
```diff
@@ -475,6 +477,7 @@ internal class MemoryInfo
             public string LowMemLastFiveMinute { get; set; }
             public string HighMemSinceStartup { get; set; }
             public string LowMemSinceStartup { get; set; }
+            public string LockedMemory { get; set; }
```
arekpalinski: I wonder if adding "encryption" to the name wouldn't make it more clear e.g. `EncryptionLockedMemory`

src/Raven.Server/Documents/ResourceCache.cs
```diff
@@ -165,20 +165,10 @@ public IDisposable RemoveLockAndReturn(string databaseName, Action<TResource> on
 
                 task.IgnoreUnobservedExceptions();
 
-                bool found = false;
-                while (found == false)
```
arekpalinski: I was trying to find the reason why we had it handled this way. I found this PR: https://github.com/ravendb/ravendb/pull/4997 but I'm not sure how that was meant to fix the issue https://issues.hibernatingrhinos.com/issue/RavenDB-10054.

@karmeli87 effectively your changes gets the code back to previous version? Can you think about any possible problems now?

src/Raven.Server/Documents/ResourceCache.cs
```diff
@@ -165,20 +165,10 @@ public IDisposable RemoveLockAndReturn(string databaseName, Action<TResource> on
...
+                if (_caseInsensitive.TryGetValue(databaseName, out current) == false)
                 {
-                    found = _caseInsensitive.TryGetValue(databaseName, out current);
-                    if (found == false)
-                    {
-                        resource = default(TResource);
-                        if (_caseInsensitive.TryAdd(databaseName, task) == false)
```
arekpalinski, line 175: What is the result of not adding it at all?

src/Raven.Server/Documents/ResourceCache.cs
karmeli87: Not exactly... in the PR you showed, we still added that faulted task (line 140)
TBH, I can't see a reason why to add that task if the cache is empty..

src/Raven.Server/Documents/ResourceCache.cs
karmeli87: Not sure why we added that in the first place...

src/Raven.Server/Documents/Handlers/Debugging/MemoryDebugHandler.cs
karmeli87: done

ppekrol (approved):

arekpalinski (approved):

src/Raven.Server/Documents/ResourceCache.cs
ayende: The issue is that if there isn't anything in the cache, you still need to add a *lock* to prevent it from being added, no?

src/Voron/Impl/Paging/AbstractPager.cs
```diff
@@ -105,58 +94,91 @@ public void SetPagerState(PagerState newState)
             }
         }
 
+        protected void Lock(byte* address, long sizeToLock, TransactionState state)
+        {
+            if (PlatformDetails.RunningOnPosix == false)
+                // when running on linux we can't do anything from within the process, so let's avoid the locking entirely
+                WorkingSetIncreaseLocker.EnterUpgradeableReadLock();
```
ayende: Why bother? There can only be a single one here anything, and there are no readers anyway

src/Voron/Impl/Paging/AbstractPager.cs
```diff
@@ -27,7 +28,7 @@ public abstract unsafe class AbstractPager : IDisposable, ILowMemoryHandler
 
         public static ConcurrentDictionary<string, uint> PhysicalDrivePerMountCache = new ConcurrentDictionary<string, uint>();
 
-        protected static readonly object WorkingSetIncreaseLocker = new object();
+        protected static readonly ReaderWriterLockSlim WorkingSetIncreaseLocker = new ReaderWriterLockSlim(LockRecursionPolicy.NoRecursion);
```
ayende: How much contention are we likely to actually get? Not worth the trouble , I thik

src/Voron/Impl/Paging/AbstractPager.cs
karmeli87: `WorkingSetIncreaseLocker` is `static`.. the important thing here, is if we need to increase the WS, no other thread should try and lock the memory until the increasing thread is done.

src/Voron/Impl/Paging/AbstractPager.cs
karmeli87: Startup of many encrypted databases can have a lot of contention... but for most of the normal operations, there shouldn't be much.

arekpalinski: Recently we saw the issue in our own system for that scenario - after the restart of the production server, once databases were loaded

src/Raven.Server/Documents/ResourceCache.cs
karmeli87: What race we want to solve here?

The only one that I can think of is adding a database before the `_caseInsensitive.TryGetValue` is executed.. 
Which would resolve into either throwing because the database is loading, or removing the database after it was loaded.. both are expected and ok.. I think.

ayende: Consider removing a database that is unloaded, and at the same time trying to load it.
Because it was unloaded, you didn't take the lock, and then you have a race.

src/Voron/Impl/Paging/AbstractPager.cs
ayende: The problem is that you are not actually using the reader / writer mode. 
You have a single entry at any given point anyway. Better to use a simpler lock

src/Voron/Impl/Paging/AbstractPager.cs
ayende: You are entering here with a single thread at a time, you should not use a reader / writer

karmeli87:
I have reverted the changes related to the `ResoruceCache`, will handle that in a separate PR.

test/SlowTests/Issues/EncryptedDatabaseGroup.cs
```diff
@@ -22,6 +22,30 @@ public class EncryptedDatabaseGroup : ClusterTestBase
         public EncryptedDatabaseGroup(ITestOutputHelper output) : base(output)
         {
         }
+        
+        [Fact]
```
ppekrol, line 26: Stress and x64 only?

karmeli87: I actually expect it to work on 32bit

src/Sparrow.Server/Sodium.cs
```diff
@@ -270,5 +272,31 @@ static Sodium()
 
         [DllImport(LIBSODIUM)]
         public static extern int sodium_mlock(byte* addr, UIntPtr len);
+
+        private static long _lockedBytes;
+
+        public static Size LockedMemory => new Size(_lockedBytes, SizeUnit.Bytes);
```
ayende: This is only used in the debug endpoint. Better to expose it directly and use the `new Size()` there for better code symmetry

karmeli87: done

ppekrol (approved):

PR #12710, opened 2021-08-31
RavenDB-16656 and RavenDB-16726

arekpalinski:
### Issue link

https://issues.hibernatingrhinos.com/issue/RavenDB-16656
https://issues.hibernatingrhinos.com/issue/RavenDB-16726

### Additional description

This PR contains 2 improvements:

1. RavenDB-16656: It exposes the details of References phase during indexing. In particular the number of processed references:
image

2. RavenDB-16726: It adds a performance hint which warns users about referencing a single document from multiple ones during indexing. If the reference documents are frequently updated then it might trigger heavy indexing work.

image

### Type of change

- Usability

### How risky is the change?

- Low 

### Backward compatibility

- Not relevant

- No

### Documentation update

- This change requires a documentation update. YouTrack issue has been tagged by `Documentation Required`. 

### Testing 

- Tests have been added that prove the fix is effective or that the feature works

- No

### UI work

- RavenDB-16656: Handled
- RavenDB-16726: The alert is shown in the Studio but it requires to show alert details

### DIFF
--- a/src/Raven.Server/Documents/Indexes/Index.cs
+++ b/src/Raven.Server/Documents/Indexes/Index.cs
@@ -2228,7 +2228,7 @@ namespace Raven.Server.Documents.Indexes
                                     entriesCount = writeOperation.Value.EntriesCount();
                                 }
 
-                                UpdateThreadAllocations(indexContext, null, null, false);
+                                UpdateThreadAllocations(indexContext, null, null, IndexingWorkType.None);
                             }
 
                             IndexFieldsPersistence.Persist(indexContext);
@@ -4135,7 +4135,7 @@ namespace Raven.Server.Documents.Indexes
             RenewTransaction
         }
 
-        public CanContinueBatchResult CanContinueBatch(IndexingStatsScope stats, QueryOperationContext queryContext, TransactionOperationContext indexingContext,
+        public CanContinueBatchResult CanContinueBatch(IndexingStatsScope stats, IndexingWorkType workType, QueryOperationContext queryContext, TransactionOperationContext indexingContext,
             Lazy<IndexWriteOperation> indexWriteOperation, long currentEtag, long maxEtag, long count,
             Stopwatch sw, ref TimeSpan maxTimeForDocumentTransactionToRemainOpen)
         {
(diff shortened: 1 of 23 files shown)


src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -2283,6 +2287,18 @@ public bool DoIndexingWork(IndexingStatsScope stats, CancellationToken cancellat
...
         }
 
+        private void HandleReferences(RavenTransaction tx)
+        {
+            _indexStorage.WriteReferences(CurrentIndexingScope.Current, tx);
+
+            if (_updateReferenceLoadWarning == false)
+                return;
+            
+            DocumentDatabase.NotificationCenter.Indexing.AddWarning(Name, _referenceLoadWarning);
```
ayende, line 2297: We need to add the document id here, I think

src/Raven.Server/Documents/Indexes/Index.cs
```diff
@@ -4135,7 +4162,7 @@ public enum CanContinueBatchResult
             RenewTransaction
         }
 
-        public CanContinueBatchResult CanContinueBatch(IndexingStatsScope stats, QueryOperationContext queryContext, TransactionOperationContext indexingContext,
+        public CanContinueBatchResult CanContinueBatch(IndexingStatsScope stats, IndexingWorkType workType, QueryOperationContext queryContext, TransactionOperationContext indexingContext,
```
ayende: At this point, better to create a struct and pass it using `in`

src/Raven.Server/Documents/Indexes/Index.cs
arekpalinski: We already have the ID. See `ReferenceId` in the following class:

```
public class LoadedReference
{
	public string ReferenceId { get; set; }

	public int NumberOfLoads { get; set; }
}
```

ayende: You aren't showing that in the alert, however

src/Raven.Server/Documents/Indexes/Index.cs
arekpalinski: I don't follow. Use struct instead of enum? What's the reasoning?

src/Raven.Server/Documents/Indexes/Index.cs
arekpalinski: It's going to be available in the details of the alert. It requires changes on Studio side.

arekpalinski: See comment: https://issues.hibernatingrhinos.com/issue/RavenDB-16726#focus=Comments-67-90182.0-0

src/Raven.Server/Documents/Indexes/Index.cs
arekpalinski: You mean to create a struct like `CanContinueBatchParameters` and put all options there?

ayende: Yes, for the parameters.

ml054 (approved):

src/Raven.Server/Documents/Indexes/Index.cs
arekpalinski: Introduced in https://github.com/ravendb/ravendb/pull/12710/commits/3463878bcdcbf4e66654f135e28da980bfa579b3

ppekrol (approved):

PR #12716, opened 2021-08-31
RavenDB-11275: Fixes ordering issues.

redknightlois:


redknightlois:
cc @maciejaszyk

src/Corax/Queries/SortingMatch.cs
```diff
@@ -135,15 +142,34 @@ public int Fill(Span<long> matches)
...
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private int Fill<W>(Span<long> matches) where W : struct
-        {
-            int take = _take <= 0 ? matches.Length : _take;
+        {             
+            // Important: If you are going to request a massive take like 20K you need to pass at least a 20K size buffer to work with.
+            //            The rationale for such behavior is that sorting has to find among the candidates the order between elements,
+            //            and it can't do so without checking every single element found. If you fail to do so, your results may not be
+            //            correct. 
+            Debug.Assert(_take <= matches.Length);
```
ayende, line 155: That should be a runtime check, no?

src/Corax/Queries/SortingMatch.cs
```diff
@@ -135,15 +142,34 @@ public int Fill(Span<long> matches)
...
+
+            Span<MatchComparer<TComparer, W>.Item> matchesKeys;
+            var matchesKeysHolder = MatchComparer<TComparer, W>.MatchKeysHolder;
+            if (matchesKeysHolder != null && matchesKeysHolder.Length > matches.Length)
+                matchesKeys = matchesKeysHolder.AsSpan(0, matches.Length);
+            else
+                matchesKeys = new MatchComparer<TComparer, W>.Item[matches.Length].AsSpan();
```
ayende, line 162: PowerOfTwo here?
Or use a ArrayPool?

src/Corax/Queries/SortingMatch.cs
```diff
@@ -135,15 +142,34 @@ public int Fill(Span<long> matches)
...
+
+            Span<MatchComparer<TComparer, W>.Item> bKeys;
+            var bKeysHolder = MatchComparer<TComparer, W>.BKeysHolder;
+            if (bKeysHolder != null && bKeysHolder.Length > matches.Length)
+                bKeys = matchesKeysHolder.AsSpan(0, matches.Length);
+            else
+                bKeys = new MatchComparer<TComparer, W>.Item[matches.Length].AsSpan();
```
ayende, line 169: PowerOfTwo here?
Or use a ArrayPool?

ayende: Also, what happen if the user specify a *very* large take? 
We have scenarios where it would be `int.MaxValue` here, then it is just stuck in the static holder for a very long time.

src/Corax/Queries/SortingMatch.cs
ayende: Another important factor, we can get the total number of entries in the index, and limit to that if the take is higher.

src/Corax/Queries/SortingMatch.cs
redknightlois: Yes. Question is how we do that. I am of the opinion that too big and unbounded take should use a special algorithm optimized for that case instead.

src/Corax/Queries/SortingMatch.cs
redknightlois: I would have choose ArrayPool but you said to use a threadstatic, which is fast. I have no opinion on what it is best yet.

src/Corax/Queries/SortingMatch.cs
redknightlois: I debated myself about that too, and decided against. That is a contract, the query planner must not break the contract, a debug check to catch errors should be enough.

src/Corax/Queries/SortingMatch.cs
ayende: The concern with a thread static is that we'll have that buffer stuck for quite some time. ArrayPool might be easier / better. IIRC, they use thread static internally to ensure this. Might want to use a local one, instead of the shared?

src/Corax/Queries/SortingMatch.cs
ayende: What _would_ you do here? 

`from Users order by CreatedAt desc limit 10_000_000` 

That is a reasonable thing that a user may want to do. And that can also use streaming.

I can't think of any way to handle that without actually sorting through all of them, however.

src/Corax/Queries/SortingMatch.cs
redknightlois: If we go `ArrayPool` I believe we need a query array pool that should have high traffic anyways.

src/Corax/Queries/SortingMatch.cs
redknightlois: We need to do that with a buffer of 10M, there is no good alternative really.

ayende: The question then becomes if we want to keep that around or not.
We can also do a two stage approach.
A thread static value that will cover the < 1024 values.
A query array pool that will have items that are larger than that.

The issue is that if we have queries for 10M, we'll likely have lots of that, so we'll benefit from the pool

PR #12718, opened 2021-09-01
RavenDB-17158 increased timeout

ppekrol:
### Issue link

https://issues.hibernatingrhinos.com/issue/RavenDB-17158

### Additional description

- Increased the timeout for AWS, Azure and GCP libraries so huge backups will not be timing out

### Type of change

- Regression bug fix

### How risky is the change?

- Low 

### Backward compatibility

- Not relevant

- No

### Documentation update

- Documentation update is needed. Issue was marked with Documentation Required tag.

- No

### UI work

- No UI work is needed


src/Raven.Server/Documents/PeriodicBackup/Aws/RavenAwsGlacierClient.cs
```diff
@@ -50,7 +51,12 @@ public RavenAwsGlacierClient(GlacierSettings glacierSettings, Progress progress
             else
                 credentials = new SessionAWSCredentials(glacierSettings.AwsAccessKey, glacierSettings.AwsSecretKey, glacierSettings.AwsSessionToken);
 
-            _client = new AmazonGlacierClient(credentials, region);
+            _client = new AmazonGlacierClient(credentials, new AmazonGlacierConfig
+            {
+                RegionEndpoint = region,
+                Timeout = RequestExecutor.GlobalHttpClientTimeout
```
ayende: How is the user supposed to be able to configure that? Do we allow that?

ayende: What if they need more? This is a `static readonly` value, and we might need a configurable option.

ayende (changes requested):

src/Raven.Server/Documents/PeriodicBackup/Aws/RavenAwsGlacierClient.cs
ppekrol: We never allowed that, but I've added a configuration option now.

ayende (approved):

arekpalinski (approved):

ppekrol:
Renamed the option to 'Backup.CloudStorageOperationTimeoutInMin' because this is for all the operations

